PR #1, opened 2015-03-06
Updated RTPS SEDP to process all QoS policies even if some are disabled locally due to profiles

mitza-oci:
Avoids reading uninitialized memory when minimum profile is in effect, may fix built-in topics for those cases.


ChuckRenner:
Test

PR #30, opened 2015-05-07
Memory Pool Implementation

iamtheschmitzer:


dds/DCPS/MemoryPool.cpp
```diff
@@ -0,0 +1,699 @@
...
+  if (largest_free_ && (largest_free_->size() >= size)) {
+    result = largest_free_;
+
+    // Look from largest to smallest
+    while (index >= nodes_) {
+      if (index->ptr() && index->ptr()->size() >= size) {
+        result = index->ptr();
+      }
+
+      if (index->size() < size) {
```
mitza-oci: check limit of index node instead of size of index node

dds/DCPS/MemoryPool.cpp
```diff
@@ -0,0 +1,699 @@
...
+
+  // Insert and replace alloc if necessary
+  free_index_.add(freed);
+}
+
+FreeHeader*
+MemoryPool::find_free_block(size_t req_size)
+{
+  // If larger than final index
+  if (largest_free_ && req_size >= largest_free_->size()) {
```
mitza-oci: Review this, dummy

dds/DCPS/MemoryPool.cpp
iamtheschmitzer: actually, the code is correct.  The check is after the node is used, and before the node is decremented.

PR #80, opened 2015-06-24
Fixed function to call in lines 119:120

AndroidDev77:


AndroidDev77:
@AndroidDev77 	Fixed function to call in lines 119:120

mitza-oci:
I don't understand why this change is proposed.  Could you provide more background info?

I've seen very few scenarios where I would recommend mpc.pl over mwc.pl.  In the MPC documentation, it says "The most common way to use the Make Project Creator is to run the workspace generator (mwc.pl)."

iamtheschmitzer:
Did you run into an issue running mwc.pl?

AndroidDev77:
You must create a workspace first. The instructions only says to create a
project file. If you run mwc with no workspace file it errs out telling you
that you have no workspace.

mitza-oci:
Every directory in the filesystem is an mpc workspace.  The default workspace is the current directory, which contains an *.mpc file that's the (only) member of the workspace.

mitza-oci:
I would like to try running this myself to make sure the directions are correct.  AndroidDev77, please specify which platform you're building on and version numbers of the C++ compiler and JDK.

PR #91, opened 2015-07-06
Configure fix for LD_LIBRARY_PATH

iamtheschmitzer:


configure
```diff
@@ -622,7 +634,7 @@ sub write_platform_macros {
       $plat = $pi->{'aceplatform'};
       $plat =~ s/\$COMP/($opts{'compiler'} =~ m!CC!) ? 'sunc++' : 'g++'/e;
       $plat =~ s/\$NONSTDCOMP/($opts{'compiler'} =~ m!clang!) ? 'clang' : ''/e;
-      $plat =~ s/_$//e; # Trailing _
+      $plat =~ s/_//; # Trailing _
```
mitza-oci: If It is only trailing _ (as comment indicates), use $ in regex.

iamtheschmitzer: Oh yes, this was a temporary change due to a warning.  Reverted back to
_$...

PR #123, opened 2015-08-03
TS API Update: Message Instance Guid support

oschwaldp-oci:


dds/FACE/FaceTSS.cpp
```diff
@@ -747,9 +790,12 @@ void populate_header_received(const FACE::CONNECTION_ID_TYPE& connection_id,
   }
   FACE::TS::MessageHeader& header = readers[connection_id]->last_msg_header;
 
-  //TODO: Populate other header fields appropriately
   header.platform_view_guid = Entities::instance()->connections_[connection_id].platform_view_guid;
-  //  header.message_instance_guid = sinfo.instance_handle;
+  OpenDDS::DCPS::DomainParticipantImpl* dpi =
+    dynamic_cast<OpenDDS::DCPS::DomainParticipantImpl*>(readers[connection_id]->dr->get_subscriber()->get_participant());
```
mitza-oci: Use smart pointer types to avoid reference count leaks here (both get_* methods).

PR #146, opened 2015-08-19
Fix issues with IPV6 enabled builds on Windows.

oschwaldp-oci:


dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -85,6 +86,23 @@ OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr)
...
+#ifdef ACE_HAS_IPV6
+        //front load IPV6 addresses to give preference to IPV6 interfaces
+        ACE_INET_Addr temp;
+        size_t index_last_non_ipv6 = 0;
+        for (size_t i = 0; i < addr_count; i++) {
+          if (addr_array[i].get_type() == AF_INET6) {
+            if (i == index_last_non_ipv6) {
+              ++index_last_non_ipv6;
+            } else {
+              temp = addr_array[i];
```
mitza-oci: use std::swap() here?

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -167,5 +185,45 @@ OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr)
   return fullname;
 }
 
+bool set_socket_ttl(const ACE_SOCK_Dgram& unicast_socket, const char& ttl)
```
mitza-oci: The name 'unicast_socket' came from how it was used in RtpsUdpTransport/DataLink.  That name may just be confusing here.

dds/DCPS/transport/framework/NetworkAddress.h
```diff
@@ -63,9 +65,15 @@ struct OpenDDS_Dcps_Export NetworkAddress {
...
 /// warning is logged. If there is no any name discovered from network interfaces,
 /// an error is logged.
+/// If ACE_HAS_IPV6, will give priority to IPV6 interfaces
 extern OpenDDS_Dcps_Export
 OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr = 0);
 
+/// Helper function to set the ttl on a socket appropriately
+/// given whether it is IPV4 or IPV6
+extern OpenDDS_Dcps_Export
+bool set_socket_ttl(const ACE_SOCK_Dgram& unicast_socket, const char& ttl);
```
mitza-oci: I think the use of 'char' for TTL has been a mistake from the start (it could just be int), but we need to consult documentation for ACE, linux manpages and MSDN.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -158,7 +164,15 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
     }
   }
 
-  /// FUTURE: Set TTL on the alternate sockets.
+  if (!OpenDDS::DCPS::set_socket_ttl(ipv6_alternate_socket_, ttl)) {
```
mitza-oci: Use this same helper function everywhere else multicast TTL is set in the codebase (also, rename to have "multicast" in the name?)

dds/DCPS/RTPS/BaseMessageUtils.h
```diff
@@ -139,15 +142,23 @@ locator_to_address(ACE_INET_Addr& dest, const OpenDDS::DCPS::Locator_t& locator)
     break;
 #endif
   case LOCATOR_KIND_UDPv4:
+#if !defined (ACE_HAS_IPV6) && !defined (IPV6_V6ONLY)
```
mitza-oci: On this line, replace && with || (inverse logic of the #if below).
Other lines with ACE_UNUSED_ARG should not be needed.

dds/DCPS/RTPS/ParameterListConverter.h
```diff
@@ -18,10 +18,10 @@ namespace ParameterListConverter {
                     ParameterList& param_list);
   OpenDDS_Rtps_Export
   int to_param_list(const OpenDDS::DCPS::DiscoveredWriterData& writer_data,
-                    ParameterList& param_list);
+                    ParameterList& param_list, bool map = 0 /*map IPV4 to IPV6 addr*/);
```
mitza-oci: Use 'false' as the constant for a bool type (here and line 24).

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -268,6 +269,50 @@ Sedp::init(const RepoId& guid, const RtpsDiscovery& disco,
...
+    RTPS::address_to_bytes(locators[idx].address,
+      rtps_inst->multicast_group_address_);
+  }
+
+  if (rtps_inst->local_address_config_str_.empty()) {
+    typedef OPENDDS_VECTOR(ACE_INET_Addr) AddrVector;
+    AddrVector addrs;
+    OpenDDS::DCPS::get_interface_addrs(addrs);
+    AddrVector::iterator adr_it = addrs.begin();
+    while (adr_it != addrs.end()) {
```
mitza-oci: Looks like a for loop to me.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -62,13 +62,7 @@ Spdp::Spdp(DDS::DomainId_t domain, RepoId& guid,
   sedp_.init(guid_, *disco, domain_);
 
   { // Append metatraffic unicast locator
```
mitza-oci: Extra local scope { } doesn't appear to be doing anything here.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -768,20 +775,12 @@ Spdp::SpdpTransport::open_unicast_socket(u_short port_common,
...
-                              &ttl,
-                              sizeof(ttl))) {
+  if (!OpenDDS::DCPS::set_socket_multicast_ttl(unicast_socket_, outer_->disco_->ttl())) {
     ACE_DEBUG((LM_ERROR,
                ACE_TEXT("(%P|%t) ERROR: Spdp::SpdpTransport::open_unicast_socket() - ")
-               ACE_TEXT("failed to set TTL value to %d ")
-               ACE_TEXT("for port:%hd %p\n"),
-               outer_->disco_->ttl(), uni_port,
-               ACE_TEXT("ACE_OS::setsockopt(TTL)")));
+               ACE_TEXT("failed to set TTL value to %C ")
```
mitza-oci: I don't think %C is the right formatter for this (not a null-terminated byte string).  Also please include some representation of errno (probably using %p) in the log message.  We should be able to assume that errno hasn't been changed since the system call.

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -167,5 +183,281 @@ OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr)
...
+#endif
+}
+
+bool set_socket_multicast_ttl(const ACE_SOCK_Dgram& socket, const unsigned char& ttl)
+{
+  ACE_HANDLE handle = socket.get_handle();
+#if defined (ACE_HAS_IPV6)
+  ACE_INET_Addr local_addr;
+  if (0 != socket.get_local_addr(local_addr)) {
+  VDBG((LM_WARNING, "(%P|%t) set_socket_ttl: "
```
mitza-oci: Indent VDBG within 'if'.

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -167,5 +183,281 @@ OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr)
...
+  return true;
+}
+
+bool open_dual_stack_socket(ACE_SOCK_Dgram& socket, const ACE_INET_Addr& local_address)
+{
+#if defined (ACE_HAS_IPV6) && defined (IPV6_V6ONLY)
+  int protocol_family = ACE_PROTOCOL_FAMILY_INET;
+  int protocol = 0;
+  int reuse_addr = 0;
+  if ((ACE_Addr)local_address != ACE_Addr::sap_any) {
```
mitza-oci: Is C-style cast needed here?

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -167,5 +183,281 @@ OPENDDS_STRING get_fully_qualified_hostname(ACE_INET_Addr* addr)
...
+    VDBG_LVL((LM_WARNING, "(%P|%t) WARNING: open_dual_stack_socket: "
+                          "failed to bind address to socket\n"), 2);
+    return false;
+  }
+  return true;
+#else
+  ACE_UNUSED_ARG(socket);
+  ACE_UNUSED_ARG(local_address);
+#endif
+  return false;
```
mitza-oci: The 'then' side of the #if ends in 'return true', which means that the 'return false' should be before the #endif.

dds/DCPS/transport/multicast/MulticastInst.cpp
```diff
@@ -167,7 +167,7 @@ MulticastInst::dump_to_str()
   os << formatNameForDump("nak_delay_intervals") << this->nak_delay_intervals_ << std::endl;
   os << formatNameForDump("nak_max")             << this->nak_max_ << std::endl;
   os << formatNameForDump("nak_timeout")         << this->nak_timeout_.msec() << std::endl;
-  os << formatNameForDump("ttl")                 << int(this->ttl_) << std::endl;
+  os << formatNameForDump("ttl")                 << this->ttl_ << std::endl;
```
mitza-oci: Keep the cast to an integer, to avoid printing an ASCII character.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -268,6 +269,50 @@ Sedp::init(const RepoId& guid, const RtpsDiscovery& disco,
...
+      idx = locators.length();
+      locators.length(idx + 1);
+      locators[idx].kind = address_to_kind(*adr_it);
+      locators[idx].port = rtps_inst->local_address_.get_port_number();
+      RTPS::address_to_bytes(locators[idx].address,
+        *adr_it);
+      ++adr_it;
+    }
+  }
+  else {
```
mitza-oci: Please check for coding style issues like this one: the two lines 305 and 306 should be combined.

PR #200, opened 2015-10-29
Multicast reliability, association, and message order guarantee changes.  Change LargeSample test to check for in-order messages.

oschwaldp-oci:


dds/DCPS/transport/framework/TransportClient.cpp
```diff
@@ -496,7 +496,8 @@ TransportClient::PendingAssoc::initiate_connect(TransportClient* tc,
           data_.remote_reliable_, data_.remote_durable_};
 
         TransportImpl::AcceptConnectResult res;
-
+        GuidConverter tmp_local(tc->repo_id_);
+        GuidConverter tmp_remote(this->data_.remote_id_);
```
mitza-oci: Why were these moved?

dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -295,6 +295,10 @@ SingleSendBuffer::resend(const SequenceRange& range, DisjointSequence* gaps)
 bool
 SingleSendBuffer::resend_i(const SequenceRange& range, DisjointSequence* gaps)
 {
+  //Special case, nak to make sure it has all history
+  SequenceNumber lowForAllResent;
+  range.first == SequenceNumber() ? lowForAllResent = low() : lowForAllResent = range.first;
```
mitza-oci: This seems overly complex... how about
const SequenceNumber lowForAllResent = range.first == SequenceNumber() ? low() : range.first;

dds/DCPS/transport/multicast/MulticastDataLink.cpp
```diff
@@ -296,10 +297,10 @@ MulticastDataLink::sample_received(ReceivedDataSample& sample)
 
       for (MulticastSessionMap::iterator it(temp_sessions.begin());
           it != temp_sessions.end(); ++it) {
-        it->second->control_received(sample.header_.submessage_id_,
-            sample.sample_);
+        bool ctrl_rcvd = it->second->control_received(sample.header_.submessage_id_,
+                                                      sample.sample_);
         // reset read pointer
-        if (ptr) {
+        if (ctrl_rcvd && ptr) {
```
mitza-oci: What's the rationale behind not resetting rd_ptr when control_received returns false?  Maybe this deserves clarification in the comment on ln 302 (or I'm missing something obvious).

dds/DCPS/transport/multicast/MulticastDataLink.cpp
```diff
@@ -315,10 +333,89 @@ MulticastDataLink::sample_received(ReceivedDataSample& sample)
...
+      this->session_lock_, false);
+  SequenceNumber seq = data.header_.sequence_;
+  MulticastSessionMap::iterator session_it = sessions_.find(source);
+  if (session_it != sessions_.end()) {
+    bool reliable = session_it->second->is_reliable();
+    if (!reliable) {
+      return true;
+    }
+    MulticastSession_rch session(session_it->second);
+    DisjointSequence recvd = this->data_samples_seen_[data.header_.publication_id_];
```
mitza-oci: Is a copy of the DisjointSequence needed here?  That could be expensive (heap alloc).

dds/DCPS/transport/framework/TransportClient.cpp
oschwaldp-oci: tc is a simple TransportClient*, and during the connect_datalink call (inside tc->initiate_connect_i) the guard is released at which time the TransportClient could go away and the connect fails.  This path is for failure to connect/initiate_connect_i path and thus should not access the tc pointer after this call if AcceptConnectResult returned was false.  This allows the stack to unwind without accessing anything inside tc.

dds/DCPS/transport/framework/TransportSendBuffer.cpp
oschwaldp-oci: That's more concise.  Thanks.

dds/DCPS/transport/multicast/MulticastDataLink.cpp
oschwaldp-oci: Indeed. Thanks for the catch.

PR #206, opened 2015-11-05
Fix check for link being nil

oschwaldp-oci:


dds/DCPS/transport/multicast/MulticastTransport.cpp
```diff
@@ -360,7 +360,7 @@ MulticastTransport::shutdown_i()
   for (link = this->client_links_.begin();
        link != this->client_links_.end();
        ++link) {
-    if (!::CORBA::is_nil(link->second)) {
+    if (! ::CORBA::is_nil(link->second.in())) {
```
mitza-oci: I'm not sure this will fix it.  Don't use any CORBA:: namespace functions unless you're dealing with types defined in IDL.  This is an _rch type.  The simplest code that might work is `if (link->second.in())` because _rch has no implicit conversion to raw pointer (like _var does).

PR #228, opened 2015-11-17
Wait 60 seconds for other processes, don't say unreadable publisher_t…

iamtheschmitzer:
…imeouts is an error


iamtheschmitzer:
Also wait_for_acks in RtiSerialization

PR #240, opened 2015-11-20
Updated PerlDDS/Run_Test.pm tracking of temp files.

mitza-oci:


bin/PerlDDS/Run_Test.pm
```diff
@@ -830,7 +834,7 @@ sub _temporary_file_path {
     if ($flag) {
```
iamtheschmitzer: Rather than passing a flag, you should check to see if target TEST_ROOT and TEST_FSROOT are set, as is done elsewhere.

mitza-oci: That's something for @jrw972 to address if he has time.  It's not changing in this PR.

PR #246, opened 2015-11-25
RtiSerialization: don't do infinite duration wait_for_acks, StaticDiscoveryalter pool size per configuration

iamtheschmitzer:


iamtheschmitzer:
and alter StaticDiscovery pool size per configuration

PR #297, opened 2016-09-01
SPDP: make network interface for spdp's unicast socket configurable

mitza-oci:
using SedpLocalAddress as the config parameter for now, even though the naming doesn't make sense


mitza-oci:
This ACE_INET_Addr method isn't doing what I thought it would.  And it seems ACE provides no convenient method for setting the address (and maintaining the port) from a text representation (like inet_pton from POSIX).

I think we we will need to call one of the set() methods which will require a bit more adaptation.

mitza-oci:
See #298

PR #300, opened 2016-09-09
Fixes for Java bindings docs and configure, and build on Mac OS X El Capitan

huangminghuang:


java/INSTALL
```diff
@@ -103,18 +108,25 @@ general overview of OpenDDS.
      use paths with no spaces (true on Unix too but spaces are rarely seen).
      The top level of the OpenDDS distribution is a directory that contains
      the subdirectory named dds.
+     
+     
+  3. Find the full path to JDK
+  
+     Linix    => export JAVA_HOME=$(readlink -f /usr/bin/javac | sed "s:/bin/javac::")
```
mitza-oci: typo: Linux

Should we put some of this logic in the configure script?  That way a user could just have "--java" and it would be located based on finding javac in the path?  That could work for all platforms.  If the user supplies "--java=/path" it would override the automatic detection.

java/idl2jni/runtime/idl2jni_runtime.h
```diff
@@ -89,7 +89,7 @@ void copyToCxx(JNIEnv *jni,
 template <typename Traits>
 void copyToJava(JNIEnv *jni, jobject &target,
   const TAO::details::object_reference_const_sequence_element<Traits> &source,
-  bool createNewObject = false)
+  bool /* createNewObject = false */)
```
mitza-oci: I think this was supposed to match another usage or something that the idlj2ni compiler generates.  We can avoid the warning with `bool /*createNewObject*/ = false)`

But if everything works with your change it may be that it's just never called with 3 parameters.

java/INSTALL
huangminghuang: > On Sep 9, 2016, at 1:37 PM, Adam Mitz <notifications@github.com> wrote:
> 
> Should we put some of this logic in the configure script? That way a user could just have "--java" and it would be located based on finding javac in the path? That could work for all platforms. If the user supplies "--java=/path" it would override the automatic detection.
> 
> 

The problem is I don’t see anyway to detect $JAVA_HOME for windows if it’s not set. Maybe we can assume JAVA_HOME should always set by user. If that’s the case, we can embed the logic to the configure file.

Huang-Ming

mitza-oci: If the user has supplied "--java" on the command line (with no =), I think there are a few cases we can handle (regardless of platform):

1. User has set JAVA_HOME in the environment -> use it
2. javac is found on PATH -> use location of bin/javac (after resolving symlinks on Unix) as JAVA_HOME
3. Error -> message about supplying full path to the JDK installation or setting up PATH to include javac

java/idl2jni/runtime/idl2jni_runtime.h
huangminghuang: Based on my search. copyToJava() are generated in *SupportJC.cpp file for some custom types and some of the generated function use the 3rd parameter. The one I changed is just an overloaded version for some unknown types.


Huang-Ming

java/INSTALL
huangminghuang: OK, I’ll work on it based on your proposal.

configure
```diff
@@ -808,10 +810,34 @@ $opts{'boost'} = '' if exists $opts{'boost-version'} && !exists $opts{'boost'};
...
 
+
+sub default_java_home {
+  my $host = $opts{'host'};
+  my $result = '';
+  if ($host eq 'macosx') {
+     $result = `/usr/libexec/java_home`;
+  }
+  elsif ($host ne 'win32') {
+     $result = `readlink -f /usr/bin/javac | sed "s:/bin/javac::"`;
```
mitza-oci: javac (or the user's preferred javac) may not be in /usr/bin, you can use the existing function "which()" to see if javac is on the PATH

configure
```diff
@@ -808,10 +810,34 @@ $opts{'boost'} = '' if exists $opts{'boost-version'} && !exists $opts{'boost'};
...
 for my $key (keys %optdep) {
   if (exists $opts{$key} && $opts{$key} eq '') {
     if (defined $optdep{$key}->[0] && $ENV{$optdep{$key}->[0]}) {
       $opts{$key} = $ENV{$optdep{$key}->[0]};
+      
+      if ($key eq 'java') {
+        ## when the environmental variable JAVA_HOME is set to a JRE location, 
+        ## try to resolve JAVA_HOME based on the location of javac
+        my $java_home = default_java_home();
+        if (!-r $targetEnv{$optdep{$key}->[1]} . '/' . $optdep{$key}->[2] and $java_home ne '') {
```
mitza-oci: This doesn't look right, doesn't "$optdep{'java'}->[1]" evaluate to "include/jni.h"?

mitza-oci:
Is this ready to merge?  Do you want to commit changes to TAO 2.2a first?  You can also create a pull request for DOCGroup/ACE_TAO.

huangminghuang:
I have already commit all the changes to TAO 2.2a and I just created a new pull request for DDS. You can merge it if you don’t see any problem. 

From I can see from the User guide, there’s nothing need to be changed for the java configuration because it already refers to the INSTALL file instead of repeating the content. Is there anything else you think need to be modified?

Huang-Ming

mitza-oci:
I didn't see a new pull request, please post a link to it.

The other source of "documentation" is the web site on the gh-pages branch.  Take a look at the getting started and building pages and see if they need to be brought up to date with these changes and also make sense from a Java user's point of view.

huangminghuang:
> On Sep 13, 2016, at 11:05 AM, Adam Mitz <notifications@github.com> wrote:
> 
> I didn't see a new pull request, please post a link to it.
> 
> 

You should see it now. The problem was I have to close the old one in order to create a new pull request.
> The other source of "documentation" is the web site on the gh-pages branch. Take a look at the getting started and building pages and see if they need to be brought up to date with these changes and also make sense from a Java user's point of view.
> 
> 

I am still working on that with Philip. I have no problem with modify the content of gh-pages. I have problem to verify the website before it being published. There’s a lot of steps to do the view the modified content locally and I keep getting some errors in the process.

Huang-Ming

PR #302, opened 2016-09-13
Fixes for Java bindings docs and configure, and build on Mac OS X El Capitan 

huangminghuang:


java/INSTALL
```diff
@@ -103,16 +108,16 @@ general overview of OpenDDS.
...
 
-     Unix =>
-       ./configure --java=<full path to the JDK>
-     Windows => 
-       configure --java=<full path to the JDK>
-
+     Unix => cd $DDS_ROOT
+             ./configure --java
+             make       
+     Windows => configure --java=<full path to the JDK>
```
mitza-oci: I don't know why this requires platform-specific behavior.  Can't we look for javac on the PATH (or presence of JAVA_HOME) just as easily on Windows?

huangminghuang: The problem is that JAVA_HOME is not set automatically during JDK installation. If 
it is not set, there is no logic in configure script to detect where is JDK based
on the location of javac currently. 

Huang-Ming

mitza-oci: Just like other environment variables that appear in "configure --help", it would be up to the user to set.  Find the location of javac on the path (on all platforms) using the which() function of the configure script.

huangminghuang: Are you referring  that which() function is a perl subroutine instead of a shell function I am currently using now?

Huang-Ming

huangminghuang: OK, I finally find documentation for the perl module “File::Which”. The problem is that the module is not installed by default on my Mac or CentOS7 as far as I tested. Not sure if it’s installed by default on Windows as part of ActivePerl. I don’t have fresh windows to test it.

mitza-oci: Search for the string "sub which {" in the configure script.  You can call this on all platforms.

PR #307, opened 2016-09-23
Fix ManyToMany test infinite wait issue

huangminghuang:


mitza-oci:
The test was observed running indefinitely (until I killed a process)
> auto_run_tests_finished: tests/DCPS/ManyToMany/run_test.pl multicast 1to1 small Time:54720s Result:256

oschwaldp-oci:
The overall test run time should be governed by wait_to_kill in the run_test.pl, which is set by a calculation in the run_test.pl for total_duration_msec.  Might want to check to make sure that is being set/calculated properly.

huangminghuang:
It’s set correctly to 120 sec. However, the perl process couldn’t kill it’s subprocess for unknown reason.

Huang-Ming

oschwaldp-oci:
As many of the tests in the suite rely on the test->finish(wait)
functionality as their fail safe to avoid the infinite test run time, is
this something that needs to be investigated more?

huangminghuang:
From I traced so far, most run_test.pl  has a line $test->finish($timeout) at the end. 
Inside finish(), it can kill a named process if it is  supplied as the second parameter of finish(), like
$test->finish($time, “process_to_kill”). If the second parameter is not supplied, no process would
be killed. Unfortunately, the majority of the test scripts just supply the first parameter. Therefore,
the timeout has no effect at all. 


Huang-Ming

oschwaldp-oci:
I believe the second parameter is supposed to be optional and is just a
suggestion for which process to kill first, if it was passed in.  Otherwise
it just cycles through all of its processes killing each one.

sub stop_processes {
my $self = shift;
my $timed_wait = shift;
# if passed, this will be the first processed WaitKilled
my $name = shift;

brianjohnson5972:
I probably shouldn't add my 2 cents since I am remote and not looking at
the code, but assuming that it hasn't changed since I wrote this part of
the framework, I think it kills things in reverse order of starting, or
else it kills the process indicated as the second parameter and then kills
everything else like it would have.

mitza-oci:
Was there a consensus reached on this one?  What are the pros/cons of merging this change?

oschwaldp-oci:
I agree with Brian on the intended behavior in the framework code.

pros:

1) Fixes temporary problem seen in one test.

cons:

1) May mask actual problem in test framework.
2) Changes behavior of DDS such that scaled up versions of this test may
stop trying to associate prematurely. This test has been used at high
volumes of associations to test the limits in a large system and that's why
the wait_to_kill time was allowed to be scaled up and down depending on the
test case.  This led to the decision to have the test kill it if hung
instead of setting a limit on association time within DDS itself.
3) With the change in DDS configuration, test may signal failure when it
was just experiencing some latency in a large scale test.

PR #324, opened 2016-11-14
Initial pr to add support for versioned namespaces

jwillemsen:
Initial pull request to add support for versioned namespaces. With these changes I have been able to compile and run the ishapes DCPS example. This depends on some necessary changes in TAO_IDL that have been made after the TAO 2.4.1 release.


ChuckRenner:

jwillemsen:
After last push everything compiles locally (using GCC 6.2.1 and ACETAO master), waiting for CI builds to finish

mitza-oci:
ok to test

Johnny, FYI - that line above enables our Jenkins CI that builds PRs and runs all tests (Travis doesn't have enough time).

Are the TAO_IDL changes only required when versioned namespaces are enabled?

Are you adding a scoreboard build that enables versioned namespaces?

jwillemsen:
The TAO_IDL changes are only required when versioned namespaces are enabled. After we release a new TAO micro I will work on extending the travis-ci file to also compile with versioned namespaces.

jwillemsen:
Could someone tell me why the internal OCI build is failing?

mitza-oci:
Why did dds/DCPS/BitPubListenerImpl.h gain two extra #includes?  They don't seem to be needed.  Same question for DataDurabilityCache.cpp (though not as important).

For GuardCondition, doesn't the traits specialization need to be visible at the point of instantiation GuardCondition_var?  I thought that was why it was at the end of the file.

The change in GuidGenerator.h doesn't seem necessary to me.

Service_Participant.cpp comments-out some code, it probably needs to be uncommented or deleted.

This sort of change isn't compliant with OpenDDS's style guide, and I'd prefer not to have churn in the git history if it's not changing anything.
```
-  const CORBA::ULong len = seq.length();
+  CORBA::ULong const len = seq.length();
```

In WaitSet.h, the following is not valid C++ because identifiers starting with _ and a capital letter are reserved for the implementation.  A few lines down, the change introduces a space unnecessarily.
```
#if !defined (_DDS_WAITSET__TRAITS_)
```

dds/DdsDcps.mpc contains a change that converts tabs (needed for verbatim makefile text) to spaces.

I'm not sure why the change is dds/DdsDcpsInfrastructure.idl is necessary (related to DdsDcpsTopic.idl).

I need to spend more time looking at the change in dds/idl/keys_generator.cpp: why does it matter if OpenDDSGenerated comes before or after the user's module?

mitza-oci:
This is the "initial PR" for versioned namespaces.  What else is needed to have functioning versioned namespaces?

jwillemsen:
I am working right now on your comments. This is all part of my work on DDSX11, for that I also need to reduce the amount of visible internal classes.

The change for key_generator was needed because of namespace lookup issues at the moment we add a versioned namespace

jwillemsen:
Pushed some changes

mitza-oci:
Why does dds/Versioned_Namespace.h line 21 set OPENDDS_HAS_VERSIONED_NAMESPACE to 0 instead of 1?  What does a user need to do to enable this, just set the MPC feature versioned_namespace=1?

jwillemsen:
I have disabled the feature by default, will later make this dependent on the MPC feature, also need to make a new TAO release. In order to test everything I first need more work in DDSX11 which puts this feature to another level

mitza-oci:
In order to integrate this PR into OpenDDS, it needs to be a feature that stands on its own merits, not one that depends on some external proprietary code.  I'd like to enable the versioned namespace in my own build to see if it works.  Of course I have latest ACE_TAO master with versioned_namespace=1.

mitza-oci:
Build error of latest revision with versioned_namespace:
```
g++ -fvisibility=hidden -fvisibility-inlines-hidden -Wnon-virtual-dtor -ggdb -pthread -fno-strict-aliasing -Wall -W -Wpointer-arith -pipe -D_GNU_SOURCE   -I/tao_builds/mitza/DOC/ACE -I/tao_builds/mitza/versioned-OpenDDS -D__ACE_INLINE__ -I/tao_builds/mitza/DOC/ACE -I/tao_builds/mitza/DOC/TAO -I.. -I../dds -DACE_HAS_VERSIONED_NAMESPACE=1 -DNOMINMAX -DOPENDDS_DCPS_BUILD_DLL  -c -fPIC -o .shobj/DdsDcpsCoreTypeSupportImpl.o DdsDcpsCoreTypeSupportImpl.cpp
In file included from DdsDcpsCoreTypeSupportImpl.cpp:3:0:
DdsDcpsCoreTypeSupportImpl.h:1629:75: error: template argument 1 is invalid
 typedef OpenDDS::DCPS::DataWriterImpl_T<DDS::ParticipantBuiltinTopicData> ParticipantBuiltinTopicDataDataWriterImpl;
(many more errors follow)
```

jwillemsen:
I will work on the reduction of includes which helps with these errors, I would prefer to get these already big set on master and work first further on reduction of the includes before testing the last issues

jwillemsen:
Fyi #330 removes the typedefs that you listed

mitza-oci:
In my forked repo I now have a branch called "versioned" which has your changes from both this PR and #330.

```
DdsDcpsCoreTypeSupportImpl.h:1627:76: error: template argument 1 is invalid
   typedef OpenDDS::DCPS::TypeSupportImpl_T<DDS::ParticipantBuiltinTopicData> ParticipantBuiltinTopicDataTypeSupportImpl;
 ```

jwillemsen:
Please check #330 , will pick this up again later this week

PR #329, opened 2016-11-15
When C++11 is enabled make use of C++11 shared_ptr and to_string instead…

jwillemsen:
… of using boost

    * examples/DCPS/ishapes/Circle.cpp:
    * examples/DCPS/ishapes/Circle.hpp:
    * examples/DCPS/ishapes/DDSShapeDynamics.cpp:
    * examples/DCPS/ishapes/DDSShapeDynamics.hpp:
    * examples/DCPS/ishapes/Shape.hpp:
    * examples/DCPS/ishapes/ShapesDialog.cpp:
    * examples/DCPS/ishapes/ShapesWidget.cpp:
    * examples/DCPS/ishapes/ShapesWidget.hpp:
    * examples/DCPS/ishapes/Square.cpp:
    * examples/DCPS/ishapes/Square.hpp:
    * examples/DCPS/ishapes/Triangle.cpp:
    * examples/DCPS/ishapes/Triangle.hpp:

### DIFF
--- a/examples/DCPS/ishapes/Circle.cpp
+++ b/examples/DCPS/ishapes/Circle.cpp
@@ -4,7 +4,7 @@
 
 
 Circle::Circle(const QRect& bounds,
-               boost::shared_ptr<ShapeDynamics> dynamics,
+               shared_ptr<ShapeDynamics> dynamics,
                const QPen& pen,
                const QBrush& brush,
                bool targeted)
--- a/examples/DCPS/ishapes/Circle.hpp
+++ b/examples/DCPS/ishapes/Circle.hpp
@@ -1,7 +1,6 @@
 #ifndef _CIRCLE_HPP
 #define _CIRCLE_HPP
 
-#include <boost/shared_ptr.hpp>
 #include <Shape.hpp>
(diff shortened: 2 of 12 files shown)


ChuckRenner:

mitza-oci:
With these changes, the MPC project for the shapes demo no longer strictly "require boost".  It would be good to handle this in the MPC file and/or configure script so that the project is still included even without boost as long as the compiler is new enough and (for GCC) has -std=<something>.

jwillemsen:
There is no MPC feature for C++11 so it is not easy to handle this in the MPC file

jwillemsen:
Maybe first merge this PR so that as a second step we can see how we can get the MPC file reworked or do something with configure

mitza-oci:
I'll work on adding support to configure/MPC.  Since that will build on this PR, there's no need to merge at this time.

jwillemsen:
Ok, this one has for me lower priority. The other pull requests are more important because they are part of my proof of concept work to integrate OpenDDS into DDSX11 which is part of AXCIOMA

Issue #320, opened 2016-11-08
Too many internal headers pulled into user code


jwillemsen:
When working on the versioned namespace support (issue #316) I noticed that when using OpenDDS a lot of headers are pulled into user code. A lot of coming through the type specific data reader, data writer, and topic. I think it would help our users when we can reduce the amount of headers they pull in when using OpenDDS. I would expect they only need the IDL generated types and stubs and not a lot of the internal headers.

mitza-oci:
I wouldn't be opposed to improving this.  Do you have some examples?  We can examine why it's set up that way or attempt to change it.

jwillemsen:
I encountered this when working on the versioned namespace support but I first need to rework a few things in DDSX11 before I can back to the OpenDDS part. Starting point to look at is the generated TypeSupportImpl.h, that pulls in several templates which pull in way more. Implementing the typed reader/writer/typesupport as templates saves a few lines of code when generating, but does result in a lot of code being included.

mitza-oci:
Doesn't versioned namespace need to apply to all (exported) symbols regardless of where they are in the headers?  I'm not sure exactly what the connection is, or was it just that you were working on it at the time?

We made a deliberate decision to use C++ templates.  It's a good fit because we'd like to have the exact same behavior parameterized only on a data type.  There are some drawbacks with the inclusion model, but there are also advantages in having the IDL compiler(s) generate as little as possible.

jwillemsen:
It was just something I encountered when working on the versioned namespace support.

There are advantages and disadvantages for the usage of C++ templates. It is easy for the IDL compiler, but does result in exposing a lot of internals to user code and probably less efficient in terms of footprint. Personally I think too much is exposed to the user code, something to look at to my idea when there is some time.

jwillemsen:
Implemented first change, see pr #330

jwillemsen:
Looking further I think that a lot of headers can be hided when instead of using OpenDDS::DCPS::TypeSupportImpl_T a type specific factory is generated in TypeSupportImpl.h where the implementation of the factory methods is in the TypeSupportImpl.cpp. At that moment much less headers are exposed to the user code, more hidden through TypeSupportImpl.cpp. That way a type specific datareader/writer can be implemented using templates. 

@mitza-oci What do you think about this approach?

mitza-oci:
Is there more to do on this one, or should we close it?

jwillemsen:
[...]

PR #330, opened 2016-11-16
Remove the type specific datareader/writer implementation typedef from the user type…

jwillemsen:
… related namespace. This only pollutes the user type namespace and isn't used in general. Within the discovery the build in data is used, added typedefs for the necessary readers within the OpenDDS namespace, this is not for the user

    * dds/DCPS/DiscoveryBase.h:
    * dds/DCPS/RTPS/Sedp.cpp:
    * dds/DCPS/RTPS/Sedp.h:
    * dds/DCPS/RTPS/Spdp.cpp:
    * dds/DCPS/RTPS/Spdp.h:
    * dds/DCPS/StaticDiscovery.cpp:
    * dds/DCPS/StaticDiscovery.h:
    * dds/DCPS/TypeSupportImpl_T.h:
    * dds/idl/ts_generator.cpp:

### DIFF
--- a/dds/DCPS/TypeSupportImpl_T.h
+++ b/dds/DCPS/TypeSupportImpl_T.h
@@ -39,7 +39,7 @@ namespace OpenDDS {
     {
       typedef DataWriterImpl_T<MessageType> DataWriterImplType;
 
-      DataWriterImplType* writer_impl;
+      DataWriterImplType* writer_impl = 0;
       ACE_NEW_RETURN(writer_impl,
                      DataWriterImplType(),
                      ::DDS::DataWriter::_nil());
--- a/performance-tests/DCPS/TCPProfilingTest/DataReaderListener.cpp
+++ b/performance-tests/DCPS/TCPProfilingTest/DataReaderListener.cpp
@@ -20,7 +20,7 @@ int read (::DDS::DataReader_ptr reader, bool useZeroCopy)
       ACE_OS::sleep (delay);
     }
 
-   ::CORBA::Long const max_read_samples = 100;
+  const ::CORBA::Long max_read_samples = 100;
   int samples_recvd = 0;
   // initialize to zero.
 
--- a/performance-tests/DCPS/TCPProfilingTest/DataReaderListener.cpp
+++ b/performance-tests/DCPS/TCPProfilingTest/DataReaderListener.cpp
@@ -10,15 +10,8 @@ extern long subscriber_delay_msec; // from common.h
 
 int read (::DDS::DataReader_ptr reader, bool useZeroCopy)
 {
-  // TWF: There is an optimization to the test by
-  // using a pointer to the known servant and
-  // static_casting it to the servant
   ::profilingTest::testMsgDataReader_var var_dr
     = ::profilingTest::testMsgDataReader::_narrow(reader);
-
-  ::profilingTest::testMsgDataReader_ptr pt_dr = var_dr.ptr();
-  ::profilingTest::testMsgDataReaderImpl* dr_servant =
-      dynamic_cast< ::profilingTest::testMsgDataReaderImpl*>(pt_dr);
 
   if (subscriber_delay_msec)
     {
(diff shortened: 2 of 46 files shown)


ChuckRenner:

jwillemsen:
ok to test

mitza-oci:
ok to test

jwillemsen:
@mitza-oci is there a way to see the results from the OCI build somewhere?

mitza-oci:
It's behind the firewall.  I'll see if I can set up an rsync to the public server.

Here's the current error:
```
In file included from DdsDcpsGuidC.cpp:30:0:
DdsDcpsGuidC.h:60:42: fatal error: dds/Versioned_Namespace.h: No such file or directory
 #include /**/ "dds/Versioned_Namespace.h"
                                          ^
compilation terminated.

```

jwillemsen:
That is strange, that is a new file from #324 which is not part of this PR

mitza-oci:
I removed the workspace and restarted it.  I don't think the manually started build makes the calls to GitHub's API to update the status here.

So it seems there is a problem with how this job is set up.  I didn't set it up so I need to see how it's configured in Jenkins.  It may need an explicit 'remove workspace' step.

jwillemsen:
You could you this plugin https://wiki.jenkins-ci.org/display/JENKINS/Workspace+Cleanup+Plugin

mitza-oci:
The restarted Jenkins build completed without error.  I still need to review this one.

jwillemsen:
In short this PR removes the type specific datareader/writer typedef from the generated code because these are internal to OpenDDS and shouldn't be used by user code, the user code does need to use the IDL generated stubs. I had to rework a few tests to not use the implementation types but just the stubs (which makes the tests also more portable). In some places I added a manual typedef of the typedef because it was required for some core functionality or a very special unit test.

jwillemsen:
When this PR and #324 are merged I can work on the proposal as part of #320 for the type support.

mitza-oci:
test this please
(this should trigger Jenkins)

mitza-oci:
test this please
(this should trigger Jenkins)

dds/DCPS/TypeSupportImpl_T.h
```diff
@@ -39,7 +39,7 @@ namespace OpenDDS {
     {
       typedef DataWriterImpl_T<MessageType> DataWriterImplType;
 
-      DataWriterImplType* writer_impl;
+      DataWriterImplType* writer_impl = 0;
```
mitza-oci: This doesn't seem to be necessary since ACE_NEW_RETURN will always assign a value, writer_impl will never be uninitialized.  I don't mind leaving the change in though.

performance-tests/DCPS/TCPProfilingTest/DataReaderListener.cpp
```diff
@@ -10,32 +10,24 @@ extern long subscriber_delay_msec; // from common.h
...
-
   if (subscriber_delay_msec)
     {
       ACE_Time_Value delay ( subscriber_delay_msec / 1000,
                             (subscriber_delay_msec % 1000) * 1000);
       ACE_OS::sleep (delay);
     }
 
-  const ::CORBA::Long max_read_samples = 100;
+   ::CORBA::Long const max_read_samples = 100;
```
mitza-oci: Seem comments in other pull request about this one... let's not introduce changes that don't change anything.

mitza-oci:
I built this locally and it looks like a good change to have.  One thing that had me confused was that the subject line referred to "type specific DataReader" and not DataReaderImpl.  User code shouldn't be referring to DataReaderImpl (aka DataReader_T) so this is fine.  It would be nice if all the Impl classes could be hidden but the DDS spec requires the user to interact with TypeSupportImpl.

One the Jenkins build has finished and the minor notes above are addressed I'll merge it.

jwillemsen:
Thanks for the review, I think that by reworking the TypeSupportImpl to a class we can reduce a lot of implicit includes for the user code, will try that at the moment we have all ready work integrated and stable

PR #339, opened 2016-11-18
Add ACE_TAO as include, is the default directory name for the ACE_TAO…

jwillemsen:
… repository

    * DDS.mwc:

### DIFF
--- a/DDS.mwc
+++ b/DDS.mwc
@@ -12,6 +12,7 @@ workspace {
     ACE_wrappers
     // For the case where the user clones git repository
     ATCD
+    ACE_TAO
 
     contrib/wrapper
     // Run tools/modeling/tests/setup.pl to generate


DDS.mwc
```diff
@@ -12,6 +12,7 @@ workspace {
     ACE_wrappers
     // For the case where the user clones git repository
     ATCD
+    ACE_TAO
```
mitza-oci: Looks like this didn't get updated when the ATCD git repo name changed.  I think we can replace the ATCD line.

jwillemsen: At the moment an user has still a local git clone from before the rename he could still use ATCD as name, that is why I didn't remove the ATCD line

mitza-oci: That's fine.  I'll merge this when I merge something else, I don't want to kick off builds just for this.

jwillemsen: Setup two new builds from our side for the scoreboard, first builds will run tonight

Issue #345, opened 2016-11-22
Rework DataReaderImpl::auto_return_loan


jwillemsen:
Currently ZeroCopyDataSeq::~ZeroCopyDataSeq pulls in the full DataReaderImpl.h just to call auto_return_loan, maybe move this to a seperate class so that the zero sequence implementation doesn't pull in the full datareaderimpl with everything it uses

jwillemsen:
@mitza-oci Can you assign this one to me?

jwillemsen:
Already implemented, working on the last errors because we now pull in much less headers which causes some compile errors in tests

jwillemsen:
Implemented

Issue #316, opened 2016-10-17
Add support for versioned namespaces


jwillemsen:
ACE and TAO have support to put everything into a versioned namespace. Would be useful for IDL2C++11 when at least the IDL generated code can support versioned namespaces.

jwillemsen:
Working on some initial changes for this enhancement

mitza-oci:
See pull request #324

jwillemsen:
Only FACE is not done yet, need to revisit that again

jwillemsen:
Support for FACE is implemented, working on a new PR

jwillemsen:
Implemented support for versioned namespaces

PR #347, opened 2016-11-23
Combined PR for various changes

jwillemsen:
1. Support for versioned namespaces for FACE, last part for issue #316 
2. Reduce includes pulled in by the zero copy sequences issue #345 
3. Add includes to various test files because we pull in much less headers into user code
4. Make sure we include the necessary ACE condition specializations when we use these


jwillemsen:
Looks the OCI build has one test error which doesn't seem related to our changes

PR #372, opened 2017-01-03
Liveliness qos updates

oschwaldp-oci:


jwillemsen:
Looks some Windows have some issues, see for example http://scoreboard.ociweb.com/dds_doc_master_cricket_win2012r2-x64_vc10_d1o0/2017_01_04_17_23_Brief.html#section_5

mitza-oci:
That kind of error occurs when two projects have rules for running a code generator and both run in parallel.

PR #375, opened 2017-01-04
Fix issues reported by Coverity.

oschwaldp-oci:


dds/DCPS/WriteDataContainer.cpp
```diff
@@ -642,6 +642,7 @@ WriteDataContainer::data_delivered(const DataSampleElement* sample)
                  OPENDDS_STRING(converter).c_str()));
     }
     release_buffer(stale);
+    stale = NULL;
```
jwillemsen: Shouldn't this be 0? NULL is normally not used in ACE/TAO and I assume also OpenDDS.

oschwaldp-oci: Correct. Fixed.

jwillemsen:
FYI, our Coverity build will run each Saturday morning so new results tomorrow

Issue #268, opened 2016-04-15
Time-Based Filter QoS


mitza-oci:
The spec requires a Time-Based Filter reader to eventually see the latest sample per instance. In cases where that latest sample would normally be suppressed by the TBFilter, the current implementation does not eventually deliver that later sample. This isn't an issue for periodic writers.

In terms of implementation, consider reusing much of the code for reader-side Deadline (timer support, etc.) to make the most recent sample available if the writer stops updating the instance.

_This issue was originally on the SourceForge tracker, which we are no longer using._

brianjohnson5972:
I think this is the important part of the spec:

"In the case where the reliability QoS kind is RELIABLE then in steady-state, defined as the situation where the DataWriter does not write new samples for a period “long” compared to the minimum_separation, the system should guarantee delivery the last sample to the DataReader."

So we only need to guarantee that the sample is re-sent if the DataReader has a reliability QOS kind of RELIABLE.

Also, the following:

" The HISTORY and RELIABILITY QoS affect the behavior of the middleware with respect to the samples that have been determined to be of interest to the DataReader, that is, they apply after the TIME_BASED_FILTER has been applied. "

indicates that filtering should prevent samples from being added to the history (which we already do), but seems to be ambiguous as to whether the Reliably delivered sample should be applied to the History QOS. But I think that is the way it has to work, since we are reporting a sample and a subsequent request to get the last sample should return that sample.

mitza-oci:
The phrase "sample is re-sent" seems to imply that there's an actual network transaction here.  That might not be the case.  It's an implementation detail whether or not the writer actually sends a sample.  The reader could just cache it.

brianjohnson5972:
Poor word choice, edited original comment

brianjohnson5972:
New Test scenario for Time Based Filter with Reliability type of Reliable
![timebasedfilter](https://cloud.githubusercontent.com/assets/10270043/19978696/cc189d38-a1c5-11e6-98e0-7018a60f773f.JPG)

brianjohnson5972:
Not really part of this problem, but in the affected code:
In set_instance_state, store_instance_data is being called and the filtered flag is ignored and notify_read_conditions() is called even if there was no sample added to the history(filtered).  It isn't producing a wrong state, since those listening for the read condition should see that they already saw the last instance sample in history, but this will cause spurious read conditions.  Will clean this up with the change to avoid signalling there is new data when there is none.

[...]

PR #377, opened 2017-01-05
Time Based Filter QoS (support for reliable readers) Issue #268

brianjohnson5972:
Time Based Filter Qos support is currently just filtering samples to prevent more than 1 sample for an instance being delivered by the data reader in the time based filter period.  Need to add support for Data readers with a Reliability Qos of Reliable, to deliver the last filtered sample after the time based filter period has expired.

### DIFF
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -7,7 +7,11 @@
 #include "dds/DCPS/BuiltInTopicUtils.h"
 #include "dds/DCPS/Util.h"
 #include "dds/DCPS/TypeSupportImpl.h"
+#include "dds/DCPS/Watchdog.h"
 #include "dcps_export.h"
+
+#include "ace/Bound_Ptr.h"
+#include "ace/Time_Value.h"
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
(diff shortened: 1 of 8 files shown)


dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1476,196 +1521,297 @@ void store_instance_data(
...
+          inst->insert(typename InstanceMap::value_type(*instance_data,
+            handle));
+        if (bpair.second == false)
+        {
           ACE_ERROR ((LM_ERROR,
                       ACE_TEXT("(%P|%t) ")
                       ACE_TEXT("%CDataReaderImpl::")
                       ACE_TEXT("store_instance_data, ")
-                      ACE_TEXT("release instance_lock failed. \n"), TraitsType::type_name()));
+                      ACE_TEXT("insert to participant scope %s failed. \n"), TraitsType::type_name(), TraitsType::type_name()));
```
mitza-oci: Why does the log message use the same parameter twice?  Also %C is probably what you want instead of %s.

brianjohnson5972: I can change this, but this code was only re-alligned.  The diff logic is matching the very similar ACE_ERROR messages.

jwillemsen:
Our fedora 24 build has some new warnings in the new test

PR #381, opened 2017-01-09
Coverity issues fixes

oschwaldp-oci:


tests/DCPS/LivelinessKeepAliveTest/subscriber.cpp
```diff
@@ -169,6 +169,13 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
     AlertDataReaderListenerImpl* drl_servant =
       dynamic_cast<AlertDataReaderListenerImpl*>(alert_listener.in());
 
+    if (CORBA::is_nil(drl_servant))
```
jwillemsen: You are checking a regular C++ pointer here, so shouldn't this be `if (!drl_servant)`

PR #383, opened 2017-01-10
Make use of OPENDDS_DES_FREE_THIS

jwillemsen:
Make use of the new OPEN_DES_FREE_THIS to resolve compile warnings.

### DIFF
--- a/dds/DCPS/Definitions.h
+++ b/dds/DCPS/Definitions.h
@@ -40,10 +40,12 @@
   ACE_UNIMPLEMENTED_FUNC(CLASS& operator=(const CLASS&))
 #endif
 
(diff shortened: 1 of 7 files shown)


dds/DCPS/Definitions.h
```diff
@@ -40,6 +40,18 @@
   ACE_UNIMPLEMENTED_FUNC(CLASS& operator=(const CLASS&))
 #endif
 
+#if !defined  (ACE_DES_FREE_THIS)
```
mitza-oci: #ifdef ACE_DES_FREE_THIS
#define OPENDDS_DES_FREE_THIS ACE_DES_FEE_THIS
#else
#define OPENDDS_DES_FREE_THIS //definition
#endif

PR #387, opened 2017-01-12
Add read/take_instance_w_condition and some other changes

jwillemsen:


mitza-oci:
Test failure "tests/DCPS/FooTest5/run_test.pl blocking" is expected at this point.  It also shows up in master and we currently working on fixing it.

mitza-oci:
@jwillemsen please merge from master to get this branch ready to merge

jwillemsen:
@mitza-oci Merged with master

jwillemsen:
@mitza-oci anything holding the merge?

mitza-oci:
Once the scoreboard is back to green I'll merge this one.

mitza-oci:
How do the versioned namespace macros interact with the XSC-generated code?

jwillemsen:
I put the external classes into the versioned namespace, works without problems from our X11 side, now able to use these without namespace conflicts

mitza-oci:
However, a user would not be able to use two different versions both with the XML handler library.  I don't mind leaving it like this, but it seems like a problem to have these names in non-versioned namespaces.  On a side note, it would be good to have them in an OpenDDS namespace to avoid potential conflicts with user code.

jwillemsen:
I can see the issue when an user wants to load two versions, but that is something I haven't tried at all, I have a DDS namespace in C++11 with C++11 types, all DDS types from C++ are now in the versioned namespace, that now works, even got the QoS XML in and it all compiled, next step is to test it.

PR #389, opened 2017-01-12
Fix issues with running test on static builds.

oschwaldp-oci:


oschwaldp-oci:
Tested on local static build without issue.  travis, appveyor, and continuous-integration won't catch any failures for these changes due to none being static builds.

Issue #368, opened 2016-12-22
Extend OpenDDS TypeSupport with an unregister_type operation


jwillemsen:
Currently OpenDDS doesn't provide the functionality to unregister a type from its type support. Other DDS vendors provide this and as far as I remember there should be an issue on the DDS specification to add this.

mitza-oci:
I could not find an OMG issue for it here: http://issues.omg.org/issues/lists/dds-rtf5

What is the use case for unregsitering?  I agree that it's a good idea in general, but as long as there is a small number of types the overhead is small.

Is this for unloading and then loading a modified version of the type?  That would be difficult to coordinate across the network.

jwillemsen:
It is for unloading components and its associated connectors from a running lwccm application, at that moment we also want to unload the dds type specific dll

mitza-oci:
In this case is the DomainParticipant object deleted?  If so, the types are effectively unregistered anyway.  In the DDS API, types are registered "belonging" to a specific participant.

jwillemsen:
Not always, we try to limit the amount of dp instances as much as we can

mitza-oci:
You can add to the OpenDDS-specific interface in $DDS_ROOT/dds/DdsDcpsTypeSupportExt.idl.

mathersm:
Working with containers we also came across this issue. Another place where it raises its head is with the interface for the OpenDDS::DCPS::Data_Types_Register. There is no unregister_type method available on the interface for that singleton. When the singleton itself goes to unregister all its registered types if any component has been unloaded (that also unloads the final reference to a type library) it will lead to access violations.

mitza-oci:
@mathersm please clarify what you mean by "containers" here?

Data_Types_Register is an internal class (as you can see by the namespace), it only exists to meet the requirements of the spec.  If/when there is an enhancement made to DdsDcpsTypeSupportExt.idl, it can be changed as well.

Are you planning on contributing that change?

mathersm:
@mitza-oci I'm referring to component containers. Probably why we've noted similar issues to @jwillemsen.

I have a branch with a simple change to the Data_Types_Register in a forked repo (https://github.com/mathersm/OpenDDS/tree/Issue-368_Extend_OpenDDS_TypeSupport_with_an_unregister_type_operation). I couldn't see an existing branch in the main repo to address this issue so I'm not sure how you would like to manage pull requests etc.

jwillemsen:
[...]

PR #393, opened 2017-01-13
Issue 368 extend open dds type support with an unregister type operation

mathersm:
The changes added here should allow for user-initiated unregistering of types from a domain participant. This has been included as a type support extension as unregister_type is not required by the DDS specification.

It might be worth considering alternatives to reporting DDS::RETCODE_ERROR for the case where a type name is supplied that has not been previously registered. DDS::RETCODE_BAD_PARAMETER is reported for the case of an empty or null string.

I am not that familiar with the automated test suite for OpenDDS so I'm happy to accept advice on better ways to structure the supplied simple test (UnregisterTypeTest project).


ChuckRenner:

dds/DCPS/Registered_Data_Types.cpp
```diff
@@ -59,6 +59,32 @@ DDS::ReturnCode_t Data_Types_Register::register_type(
...
 
+DDS::ReturnCode_t Data_Types_Register::unregister_type(
+    DDS::DomainParticipant_ptr domain_participant,
+    const char* type_name,
+    TypeSupport_ptr the_type)
+{
+  ACE_GUARD_RETURN(ACE_SYNCH_MUTEX, guard, lock_, DDS::RETCODE_ERROR);
+
+  TypeSupportMap& tsm = participants_[domain_participant];
+  const TypeSupport_var typeSupport = TypeSupport::_duplicate(the_type);
```
jwillemsen: Do we need typeSupport here, can't we just use the_type below?

jwillemsen:
In tests/DCPS/unit/ there seems to be a manual implemented type support, that has to be updated also for this change

mitza-oci:
ok to test

@mathersm thanks for your contribution.  The phrase above will trigger our Jenkins build to run the entire regression test suite on your branch (actually on the merge commit of your branch with master).  Once that build is done, you can see the full results at http://scoreboard.ociweb.com/OpenDDS/

mitza-oci:
references issue #368

dds/DCPS/Registered_Data_Types.cpp
mathersm: Added that change

jwillemsen:
No remarks on the code, only the new test has to be added to bin/dcps_tests.lst so that it runs as part of the regression test suite

mathersm:
Can someone review the failed test from the OCI CI tools? Through my own testing I'm not able to reproduce an error condition when running under a Linux environment for the Reconnect test (restart_sub).

mitza-oci:
I'm not concerned with the Reconnect test.  There are some issues with that test where it may be making some assumptions regarding timing.  We'll work on that test when we can.

dds/DdsDcpsTypeSupportExt.idl
```diff
@@ -39,6 +39,11 @@ module OpenDDS {
   #endif
 
       boolean has_dcps_key();
+
+      ::DDS::ReturnCode_t unregister_type(
+          in ::DDS::DomainParticipant domain,
+          in string type_name);
```
mitza-oci: Is there a need for the type_name parameter here?  Couldn't the receiving object just be unregistered from the domain participant?  It already knows its own type_name.

jwillemsen: The same type support can be registered using multiple typenames?

mitza-oci: Interesting, that does seem to be allowed by the spec.  I wonder why it doesn't just use two different objects (of the same class) for that?  Perhaps something that was "lost in translation" when they attempted to model static member functions in IDL.  In any case, we're stuck with it now.

Given that, the implementation of TypeSupportImpl::get_type_name() should probably work slightly differently.  But it would probably only matter in very strange use cases.

jwillemsen: Maybe make a separate issue for reworking the get_type_name()?

tests/DCPS/UnregisterType/run_test.pl
```diff
@@ -0,0 +1,37 @@
...
+my $options = '';
+my $DCPSREPO;
+
+# The test results in error output from attempting to register
+# more instances than allowed. Capture it to prevent false alarms.
+my $logfile = 'output.log';
+
+unlink $logfile;
+
+$options = "-DCPSConfigFile rtps_disc.ini -ORBLogFile $logfile";
```
mitza-oci: The test shouldn't be hard-coded to use an RTPS Discovery configuration.  Since the test is independent of discovery mechanism, just use the defaults.  The problem with RTPS here is that it requires multicasting which can interfere with other builds unless certain steps are taken.  That is why tests that use RTPS as marked with an RTPS tag in dcps_tests.lst.

tests/DCPS/UnregisterType/run_test.pl
```diff
@@ -10,22 +10,12 @@
...
-
-# The test results in error output from attempting to register
-# more instances than allowed. Capture it to prevent false alarms.
-my $logfile = 'output.log';
-
-unlink $logfile;
-
-$options = "-DCPSConfigFile rtps_disc.ini -ORBLogFile $logfile";
+my $status = 0;
+my $options = "-DcpsBits 0";
```
mitza-oci: This should be "-DCPSBit 0"

PR #397, opened 2017-01-16
When we have C++11 we also use noexcept(false)

jwillemsen:
* tools/modeling/codegen/model/Sync.h:


mitza-oci:
Do we no longer need this PR now that PR #387 contains these changes?

jwillemsen:
Not really, but when you merge this one I think merging #387 after that will be no problem

jwillemsen:
With this change and the changes made to ACE/TAO master today we should get warning free compilation on fc24 with C++11 enabled

PR #420, opened 2017-01-19
Extend travis ci config

oschwaldp-oci:


.travis.yml
```diff
@@ -1,17 +1,68 @@
...
-  - clang
 
 env:
+  matrix:
+    - FUZZ=1
+    - NO_INLINE=1
+    - NO_BIT=1
+    - STATIC=1 NO_DEBUG=1
+    - SAFETY_PROFILE=1
+    - MOD_DEFAULT_FEATURES=1 USES_WCHAR=1
```
mitza-oci: Why is MOD_DEFAULT_FEATURES needed?

.travis.yml
```diff
@@ -1,17 +1,68 @@
...
+      env: FUZZ=1 USE_JAVA=1
+    - compiler: clang
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.9
+      env: USE_ALT_COMPILER=1 COMPILER=g++-4.9
```
mitza-oci: I don't think we need to test with GCC 4.9 if we also have a newer GCC working.

.travis.yml
```diff
@@ -1,17 +1,68 @@
...
+      env: USE_ALT_COMPILER=1 COMPILER=g++-5
+    - compiler: clang
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-precise-3.6
+          packages:
+            - clang++-3.6
+      env: USE_ALT_COMPILER=1 COMPILER=clang++-3.6
```
mitza-oci: I don't think we need to test with Clang 3.6 if we also have a newer Clang working.

.travis.yml
```diff
@@ -1,17 +1,68 @@
...
+    - compiler: clang
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-precise-3.7
+          packages:
+            - clang++-3.7
+      env: USE_ALT_COMPILER=1 COMPILER=clang++-3.7
```
mitza-oci: Is it possible to get the current clang release (3.7.x?) or is Travis's Ubuntu too old for it?

.travis.yml
```diff
@@ -1,17 +1,68 @@
...
+            - g++-4.9
+      env: USE_ALT_COMPILER=1 COMPILER=g++-4.9
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-5
+      env: USE_ALT_COMPILER=1 COMPILER=g++-5
```
mitza-oci: Is it possible to use the latest GCC release (6.3)?

.travis.yml
```diff
@@ -1,17 +1,68 @@
...
     - LD_LIBRARY_PATH=$ACE_ROOT/lib:$DDS_ROOT/lib:$LD_LIBRARY_PATH
 
+matrix:
+  include:
+    - compiler: gcc
+      addons:
+        apt:
+          packages:
+            - openjdk-7-jdk
+      env: FUZZ=1 USE_JAVA=1
```
mitza-oci: Does the Java build require the older GCC, or is this just for coverage?

.travis.yml
oschwaldp-oci: Was using this flag to simplify the construction of all the ADDL_DEFAULT_FEATURES additions that would need to be included after the --features="..." config param.  If we eventually want to do more than a wchar build here this would be beneficial.

.travis.yml
oschwaldp-oci: Set this up before confirming that newer compilers were going to work.  Was also under the impression we were trying to somewhat mirror builds we maintain so that we could confirm when it was merged into the main branch we might see test failures on those builds, but at least builds should work.

.travis.yml
oschwaldp-oci: That's fine.  We can remove this compiler.  Once again added it for coverage of the compiler set we might be supporting.

.travis.yml
oschwaldp-oci: That's fine.  We can remove this compiler.  Once again added it for coverage of the compiler set we might be supporting.

.travis.yml
oschwaldp-oci: Will have to check.

.travis.yml
oschwaldp-oci: Will have to check.

.travis.yml
jwillemsen: Did some work on ACE/TAO today, there is a clang 3.9 you could use, see the travis.yml in the ACE/TAO repo

.travis.yml
jwillemsen: Did some work on ACE/TAO today, there is a g++-6 you could use, see the travis.yml in the ACE/TAO repo

PR #425, opened 2017-01-20
ReadCondition test: compatibility with safety profile and avoid looping forever with unexpected return code from take_next_instance_w_condition

mitza-oci:


tests/DCPS/ReadCondition/ReadConditionTest.cpp
```diff
@@ -106,6 +106,7 @@ int run_test_instance (int argc, ACE_TCHAR *argv[])
                  << ret << endl;
             passed = false;
             done = true;
+            break;
```
jwillemsen: Same fix could be made to line 215, a potential loop is also there

mitza-oci: Yes, but this one is one the I observed failing.  I think we'll need further changes, but I wanted to get this much in first.

PR #458, opened 2017-02-13
New base project and example of its use, for placing generated code in a subdirectory.

mitza-oci:


jwillemsen:
Much smaller change, no possibility to do this without a new base project?

mitza-oci:
You don't need the base project if you want to assign to commandflags (aka dcps_ts_flags) in the project.  I'd rather have a base project since that lets the user opt-in to this behavior using a name that shows the intent and can be grepped.

I tried a version of this where the base project also took care of idlflags but it didn't work because the setting didn't apply to FooTypeSupport.idl.  That could possibly be fixed with an MPC change.

Issue #489, opened 2017-03-28
Segv when trying to trigger a sample_lost error


jwillemsen:
When trying to trigger a sample_lost callback we see one of our dds4ccm unit tests crash with the following callstack

```
(28695|140737032713984) ConditionImpl::signal_all()
(28695|140737032713984) StatusConditionImpl::get_trigger_value() - mask==0xffffffff, changes==0x5100.
(28695|140737032713984) ConditionImpl::signal_all() - number of sets: 0, locally: 0.
(28695|140737032713984) DataReaderImpl::data_received: reader 010318db.f219c891.70170000.00000007(13aa0bbc) writer 010318db.f219c891.70210000.00000002(a8736f5b) instance 0 is_new_instance 0 filtered 0 

Thread 7 "dancex11_deploy" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe4d77700 (LWP 28721)]
0x00007ffff4985d4c in ACE_6_4_2::ACE_Time_Value::sec (this=0x120) at /home/johnny/ACE/idl2c++11/ACE/ACE/ace/Time_Value.inl:123
123       return this->tv_.tv_sec;
(gdb) back
#0  0x00007ffff4985d4c in ACE_6_4_2::ACE_Time_Value::sec (this=0x120) at /home/johnny/ACE/idl2c++11/ACE/ACE/ace/Time_Value.inl:123
#1  0x00007ffff42c4294 in ACE_6_4_2::ACE_Time_Value::operator= (this=0x108, tv=...) at /home/johnny/ACE/idl2c++11/ACE/ACE/ace/Time_Value.inl:350
#2  0x00007fffec9a1639 in DDS_3_10_0::OpenDDS::DCPS::DataReaderImpl::accept_sample_processing (this=0x838c20, instance=..., header=..., is_new_instance=false) at DCPS/DataReaderImpl.cpp:3390
#3  0x00007fffec998af8 in DDS_3_10_0::OpenDDS::DCPS::DataReaderImpl::data_received (this=0x838c20, sample=...) at DCPS/DataReaderImpl.cpp:1522
[...]

PR #490, opened 2017-03-28
Fixed crashes as reported with issue #489

jwillemsen:
Need to check instance before using it, during testing we had some runs where instance was nil causing a crash


mitza-oci:
Please add a regression test.

jwillemsen:
@mitza-oci Hopefully soon we can use dds4ccm as unit test, see the issue we made for the callstack of the crash we got.

jwillemsen:
This is also AXCIOMA DDS4CCM, comparable with CIAO/connectors/dds4ccm/tests/QueryCondition/Different/descriptors/run_test.pl

jwillemsen:
Added full logs to the issue

jwillemsen:
Have a plain OpenDDS reproducer at https://github.com/jwillemsen/OpenDDS/tree/jwi-samplelost

Issue #495, opened 2017-03-29
Crash in filtering


jwillemsen:
We see a crash in handling of filters due to a nil pointer read

```
Thread 7 "dancex11_deploy" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffd3fff700 (LWP 1613)]
0x00007fffee1fa058 in DDS_3_10_0::OpenDDS::DCPS::MetaStructImpl<DDS_Native::CommonTestMessage>::getValue (this=0x7fffee4986c8 <DDS_3_10_0::OpenDDS::DCPS::MetaStruct const& DDS_3_10_0::OpenDDS::DCPS::getMetaStruct<DDS_Native::CommonTestMessage>()::msi>, stru=0x0, 
    field=0x862100 "iteration") at generated_code/common_ddsTypeSupportImpl.cpp:181
181           return typed.iteration;
(gdb) back
#0  0x00007fffee1fa058 in DDS_3_10_0::OpenDDS::DCPS::MetaStructImpl<DDS_Native::CommonTestMessage>::getValue (this=0x7fffee4986c8 <DDS_3_10_0::OpenDDS::DCPS::MetaStruct const& DDS_3_10_0::OpenDDS::DCPS::getMetaStruct<DDS_Native::CommonTestMessage>()::msi>, stru=0x0, 
    field=0x862100 "iteration") at generated_code/common_ddsTypeSupportImpl.cpp:181
#1  0x00007fffed050e90 in DDS_3_10_0::OpenDDS::DCPS::FilterEvaluator::DeserializedForEval::lookup (this=0x7fffd3ffc680, field=0x862100 "iteration") at DCPS/FilterEvaluator.cpp:105
#2  0x00007fffed0513a8 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::FieldLookup::eval (this=0x8620d0, data=...) at DCPS/FilterEvaluator.cpp:150
#3  0x00007fffed051e04 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::Comparison::eval (this=0x741560, data=...) at DCPS/FilterEvaluator.cpp:265
#4  0x00007fffed0528c2 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::Logical::eval (this=0x96d8e0, data=...) at DCPS/FilterEvaluator.cpp:406
[...]

PR #496, opened 2017-03-29
Check if item_registered_data_ is not null before calling into the…

jwillemsen:
…aluator, see issue #495

    * dds/DCPS/DataReaderImpl_T.h:

### DIFF
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -753,7 +753,7 @@ namespace OpenDDS {
 #ifndef OPENDDS_NO_OBJECT_MODEL_PROFILE
               && !item->coherent_change_
 #endif
-              ) {
+              && item->registered_data_) {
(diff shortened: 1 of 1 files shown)


mitza-oci:
Please add a regression test.

jwillemsen:
@mitza-oci Hopefully soon we can use dds4ccm as unit test, see the issue we made for the callstack of the crash we got.

mitza-oci:
Is there some test in CIAO we could use to reproduce the problem?  The callstack isn't enough to learn what application code puts OpenDDS in this failure state.  (I assume this is the same for PR #490 so I'm not going to replicate my comment there.)

jwillemsen:
This is related to issue #495, which is triggered by an AXCIOMA DDS4CCM unit test. Looks CIAO/connectors/dds4ccm/tests/QueryCondition/EventListener is a little bit comparable with the AXCIOMA version but that seems to run

jwillemsen:
Looks related to filter and unregistering keyed samples

Issue #337, opened 2016-11-18
Generated gen_skip_over functions don't check result of length demarshaling


jwillemsen:
I was looking at the generated code and noticed that for example 'BooleanSeqTypeSupportImpl.cpp' the function 'void gen_skip_over(Serializer& ser, CORBA::BooleanSeq*)' is generated as below. The demarshaling of length is not checked, this is done in the 'bool operator>>(Serializer& strm, CORBA::BooleanSeq& seq)' function which is just above. At the moment we get incorrect data just skipping length independent of demarshaling succeeded or not looks a risk

```
void gen_skip_over(Serializer& ser, CORBA::BooleanSeq*)
{
  ACE_UNUSED_ARG(ser);
  ACE_CDR::ULong length;
  ser >> length;
  ser.skip(length, 1);
}
```

mitza-oci:
This isn't data received from the network, it's used on the writer side to manage cached serialized samples.

jwillemsen:
Ok, but it doesn't seem a good thing to ignore all return values which could indicate an error situation

PR #513, opened 2017-04-12
Changed gen_skip_over to return a bool and check the return values of gen_skip_over/skip

jwillemsen:
See issue #337


jwillemsen:
Changes are complete

PR #515, opened 2017-04-12
Add some documentation and log after we have acquire the mutex

jwillemsen:


dds/DCPS/StatusConditionImpl.cpp
```diff
@@ -16,6 +16,8 @@ namespace DCPS {
 
 CORBA::Boolean StatusConditionImpl::get_trigger_value()
 {
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, g, lock_, false);
```
mitza-oci: Increasing scope of the lock could have significant performance implications with debug logging on (maybe even enough to change the system behavior with logging vs. non-logging).

jwillemsen: The mutex is already there and I only moved the logging to within the mutex. I do realize that the logging takes a little bit of time but at the moment we log the value outside of the mutex it could be that what we log is something else than what we execute a little bit later which could mean that the logs don't tell the truth

PR #537, opened 2017-04-30
Improved README.md

jwillemsen:
Link to support page in some places and removed references to TAO event services, not interesting for users that look for a DDS implementation


mitza-oci:
These updates were included (in some form) in the 3.11 release.

PR #547, opened 2017-05-03
Coverity Scan defect: checking result of dynamic; ACE logging fixes; etc.

oschwaldp-oci:


tests/DCPS/CompatibilityTest/publisher.cpp
```diff
@@ -230,6 +230,11 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
       ex._tao_print_exception ("Exception caught in main.cpp:");
       return 1;
     }
+  catch (std::runtime_error& err)
+  {
+    ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT("ERROR: main() - %s\n"),
```
jwillemsen: Should be%C

tests/DCPS/LargeSample/Writer.cpp
```diff
@@ -72,7 +72,7 @@ Writer::extend_sample(Messenger::Message& message)
   // Message ID is 0 to 9
   // Lengths will vary from 15k to 165k
   // Over 65K will require multiple fragments for udp/mcast
-  message.data.length(calc_sample_length(message.writer_id, message.sample_id));
+  message.data.length(calc_sample_length(message.sample_id, message.writer_id));
```
mitza-oci: I don't understand why this change is being made.  It doesn't have anything to do with a dynamic_cast.  The original order of the arguments appears to be correct based on the function declaration in Writer.h.

oschwaldp-oci: In Writer.h I see:

  static int calc_sample_length(long sample_id, long writer_id) {
    // Writer ID is 1 or 2
    // Sample ID is 0 to 9
    // Lengths will vary from 10k to 155k
    return int((sample_id * 1.5) + writer_id) * 10 * 1024;
  }

mitza-oci: OK, I must have missed that.

oschwaldp-oci: Thanks for double checking.

PR #549, opened 2017-05-04
Extend news file

jwillemsen:


mitza-oci:
These news items were included in the news for 3.11.

Issue #550, opened 2017-05-04
Crash in query filter


jwillemsen:
We see a crash in the OpenDDS at the moment we receive data and use a query filter

```
0x00007fffee2190f2 in DDS_3_10_0::OpenDDS::DCPS::MetaStructImpl<DDS_Native::CommonTestMessage>::getValue(void const*, char const*) const () from ../lib/libdata_ddsx11_opendds_types_impl.so.3.10.0
(gdb) back
#0  0x00007fffee2190f2 in DDS_3_10_0::OpenDDS::DCPS::MetaStructImpl<DDS_Native::CommonTestMessage>::getValue(void const*, char const*) const () from ../lib/libdata_ddsx11_opendds_types_impl.so.3.10.0
#1  0x00007fffece6665c in DDS_3_10_0::OpenDDS::DCPS::FilterEvaluator::DeserializedForEval::lookup (this=0x7fffffffc070, field=0x846c20 "iteration") at /home/johnny/ACE/idl2c++11/ACE/OpenDDS/dds/DCPS/FilterEvaluator.cpp:106
#2  0x00007fffece66b74 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::FieldLookup::eval (this=0x846bf0, data=...) at /home/johnny/ACE/idl2c++11/ACE/OpenDDS/dds/DCPS/FilterEvaluator.cpp:151
#3  0x00007fffece675e4 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::Comparison::eval (this=0x7f0260, data=...) at /home/johnny/ACE/idl2c++11/ACE/OpenDDS/dds/DCPS/FilterEvaluator.cpp:268
#4  0x00007fffece680a2 in DDS_3_10_0::OpenDDS::DCPS::(anonymous namespace)::Logical::eval (this=0x869d40, data=...) at /home/johnny/ACE/idl2c++11/ACE/OpenDDS/dds/DCPS/FilterEvaluator.cpp:409
#5  0x00007fffece68c2c in DDS_3_10_0::OpenDDS::DCPS::FilterEvaluator::eval_i (this=0x859040, data=...) at /home/johnny/ACE/idl2c++11/ACE/OpenDDS/dds/DCPS/FilterEvaluator.cpp:520
[...]

PR #551, opened 2017-05-04
Check typed_sample before we pass it to the filter, it is check for n…

jwillemsen:
…il further down in the operation

    * dds/DCPS/RakeResults_T.cpp:


jwillemsen:
See #550

PR #552, opened 2017-05-05
Fixed compiler warnings

mitza-oci:


jwillemsen:
Did see some of them today when compiling with msvc but didn't got time to fix them, thanks for taking care of them

Issue #553, opened 2017-05-05
Marked default QoS defines result in implicit activation of the domain participant factory


jwillemsen:
We are using PARTICIPANT_QOS_DEFAULT (and the other default QoS strings) to initialize some C++11 constants using uniform initialization:

```
const DomainParticipantQos PARTICIPANT_QOS_DEFAULT { DDSX11::VendorUtils::participant_qos_default () };
```

The `DDSX11::VendorUtils::participant_qos_default ()` uses internally the PARTICIPANT_QOS_DEFAULTS. This code is in a shared library which we load with dlopen and at this moment we see a lot of initialization to take place which at the end causes the system not to start.

Is there a way to implement PARTICIPANT_QOS_DEFAULT without that the full domain participant factory is tried to construct?

```
 	OpenDDS_Dcpsd.dll!DDS_3_10_0::OpenDDS::DCPS::Service_Participant::ReactorTask::wait_for_startup() Line 439	C++
>	OpenDDS_Dcpsd.dll!DDS_3_10_0::OpenDDS::DCPS::Service_Participant::get_domain_participant_factory(int & argc, char * * argv) Line 415	C++
 	OpenDDS_Dcpsd.dll!DDS_3_10_0::OpenDDS::DCPS::Marked_Default_Qos::marked_default_DomainParticipantQos() Line 23	C++
```

jwillemsen:
@mitza-oci Do you have some ideas, would like to use the *_QOS_DEFAULT magic defines without that they cause the full initialization of the dpf or other dds entities

mitza-oci:
I will take a look at it.  It's not an area of the code we've had to do anything with for a long time.

jwillemsen:
I was thinking about moving the default/initial QoS values out of the service participant so that they can be used without it.

mitza-oci:
Service Participant is distinct from Domain Participant Factory, though they are related.  I don't really have a problem with something like PARTICIPANT_QOS_DEFAULT depending on Service Participant as long is it doesn't also force a dependency on Domain Participant Factory.

jwillemsen:
Yes, the problem seems to be the usage of the get_domain_participant_factory which causes a lot of initialization to happen, including the creation of the reactor task which seems to hang at that moment on the wait_for_startup

mitza-oci:
[...]

PR #554, opened 2017-05-06
Change the implementation of the marked participant QoS

jwillemsen:
Changed the implementation of the marked participant QoS to match the other marked QoS, see issue #553.


dds/DCPS/Marked_Default_Qos.cpp
```diff
@@ -19,9 +19,13 @@ int const INVALID_ENUM_VALUE = 999;
 DDS::DomainParticipantQos
 Marked_Default_Qos::marked_default_DomainParticipantQos()
 {
-  DDS::DomainParticipantQos qos;
-  DDS::DomainParticipantFactory_var factory = TheParticipantFactory;
-  factory->get_default_participant_qos(qos);
+  DDS::DomainParticipantQos qos = TheServiceParticipant->initial_DomainParticipantQos();
+  qos.user_data.value.length(5);
```
mitza-oci: The "marked" value needs to be a value that's not possible for a user to select.  For DomainParticipant that would have to be autoenable_
created_entities not being true or false (i.e. treating the bool as a char).

tests/DCPS/WriteDataContainer/WriteDataContainerTest.cpp
```diff
@@ -350,6 +350,8 @@ int run_test(int argc, ACE_TCHAR *argv[])
 
   try
     {
+    DDS::DomainParticipantFactory_var dpf =
```
mitza-oci: Is this needed?  Not necessarily wrong but this is more of a component-level test that doesn't need the core DDS running.

jwillemsen: Yes, this is needed else we don't initialize the service participant and all kinds of mandatory members are not set causing crashes

dds/DCPS/Marked_Default_Qos.cpp
jwillemsen: couldn't get that working, that is why I went for a magic value, any suggestions to try?

tests/DCPS/WriteDataContainer/WriteDataContainerTest.cpp
jwillemsen: The monitor member is for example zero at that point, not set to the default monitor impl

mitza-oci: OK, I see where it was happening implicitly before.  It would be nice to have a mock DomainParticipantImpl there, but that's for another PR.

PR #564, opened 2017-05-16
Fixed typo and removed trailing spaces

jwillemsen:


mitza-oci:
I merged the typo fix.  The HTML documents are generated by TiddlyWiki so I'm not going to modify a lot of lines just for whitespace, I have no idea how that tool manages whitespace (for instance, if it will come back after an edit).

jwillemsen:
Where can I find the sources of those html documents so that we can make updates in the future?

mitza-oci:
They use TiddlyWiki which includes within the document a JavaScript program for editing the document (so they are also the sources).
https://en.wikipedia.org/wiki/TiddlyWiki

PR #578, opened 2017-05-31
Simplify topic implementation, store the number of entities used in t…

jwillemsen:
…he topic description, removed not necessary dynamic casts, use _var

    * dds/DCPS/TopicDescriptionImpl.cpp:
    * dds/DCPS/TopicDescriptionImpl.h:
    * dds/DCPS/TopicImpl.cpp:
    * dds/DCPS/TopicImpl.h:

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -502,7 +502,7 @@ DomainParticipantImpl::delete_topic_i(
       dynamic_cast<DomainParticipantImpl*>(dp.in());
 
     if (the_dp_servant != this ||
-        (!remove_objref && the_topic_servant->entity_refs())) {
+        (!remove_objref && the_topic_servant->has_entity_refs())) {
       // If entity_refs is true (nonzero), then some reader or writer is using
       // this topic and the spec requires delete_topic() to fail with the error:
       return DDS::RETCODE_PRECONDITION_NOT_MET;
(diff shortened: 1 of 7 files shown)


jwillemsen:
http://scoreboard.ociweb.com/OpenDDS/PR578/2017_05_31_18_34_Brief.html#section_6 seems unrelated

dds/DCPS/TopicImpl.cpp
```diff
@@ -68,19 +67,17 @@ TopicImpl::set_qos(const DDS::TopicQos & qos)
 
     } else {
       qos_ = qos;
-      DomainParticipantImpl* part =
-        dynamic_cast<DomainParticipantImpl*>(this->participant_);
 
       Discovery_rch disco =
-        TheServiceParticipant->get_discovery(part->get_domain_id());
+        TheServiceParticipant->get_discovery(this->participant_->get_domain_id());
```
mitza-oci: Please keep in mind the coding guideline "this-> is not used unless required for disambiguation or to access members of a template-dependent base class."  I don't think you need to go back and change these files just for this issue, but if you end up changing them anyway it would be good to bring them up to date with the guideline.

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -765,17 +764,17 @@ DDS::ReturnCode_t DomainParticipantImpl::delete_contentfilteredtopic(
     if (DCPS_debug_level > 3) {
       ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
         ACE_TEXT("DomainParticipantImpl::delete_contentfilteredtopic, ")
-        ACE_TEXT("can't delete a content-filtered topic \"%C\" ")
+        ACE_TEXT("can't delete contentfiltered topic \"%C\" ")
```
mitza-oci: I don't understand why this is changing.  The existing log messages are fine, including below when it says "used by a reader".

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -838,12 +837,24 @@ DDS::ReturnCode_t DomainParticipantImpl::delete_multitopic(
   ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, topics_protector_,
                    DDS::RETCODE_OUT_OF_RESOURCES);
   DDS::MultiTopic_var mt = DDS::MultiTopic::_duplicate(a_multitopic);
-  CORBA::String_var mt_name = mt->get_name();
-  TopicDescriptionMap::iterator iter = topic_descrs_.find(mt_name.in());
+  CORBA::String_var name = mt->get_name();
```
mitza-oci: Why is this changing?

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -765,17 +764,17 @@ DDS::ReturnCode_t DomainParticipantImpl::delete_contentfilteredtopic(
     if (DCPS_debug_level > 3) {
       ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
         ACE_TEXT("DomainParticipantImpl::delete_contentfilteredtopic, ")
-        ACE_TEXT("can't delete a content-filtered topic \"%C\" ")
+        ACE_TEXT("can't delete a contentfiltered topic \"%C\" ")
```
mitza-oci: Please check the changes that are being made to log message text on this line and in the block below.  My previous comment was that we should just use the log message text as it existed before this PR.  Also the log message that starts on 853 would then need to change for consistency.

jwillemsen: Every DDS entity that uses the topic now calls add_entity_ref, so it could be also that a datawriter is still using the topic, not purely a reader, so we can't say anymore that a reader uses it

mitza-oci: It is not possible for a writer to use Content-Filtered Topic.

jwillemsen: Yes, you are right, will fix tomorrow, laptop is offline

PR #580, opened 2017-05-31
Removed dr/dw get_topic_name, just use the member directly

jwillemsen:
    * dds/DCPS/DataReaderImpl.h:
    * dds/DCPS/DataWriterImpl.h:
    * dds/DCPS/PublisherImpl.cpp:

### DIFF
--- a/dds/DCPS/DataWriterImpl.cpp
+++ b/dds/DCPS/DataWriterImpl.cpp
@@ -1345,7 +1345,7 @@ DataWriterImpl::enable()
                                            qos_.reliability.max_blocking_time,
                                            n_chunks_,
                                            domain_id_,
-                                           get_topic_name(),
+                                           topic_name_.in(),
                                            get_type_name(),
 #ifndef OPENDDS_NO_PERSISTENCE_PROFILE
                                            durability_cache,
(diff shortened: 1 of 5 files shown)


dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1345,7 +1345,7 @@ DataWriterImpl::enable()
                                            qos_.reliability.max_blocking_time,
                                            n_chunks_,
                                            domain_id_,
-                                           get_topic_name(),
+                                           topic_name_.in(),
```
mitza-oci: Why have .in() here (and line 1476)?

jwillemsen: topic_name_ is a CORBA::String_var, the constructor of WriteDataContainer expects a const char*, so we have should use .in(). Could be that it compiles without .in() with gcc but we had problems with that with TAO with some compilers so I always use .in().

mitza-oci: The code is clear and simpler without .in() so that's what we prefer in OpenDDS.  I've never seen a problem with a compiler not allowing this.  Primarily we want users and developers to read and understand the code without getting bogged down in details of the IDL to C++ mapping.  We have some notes on this on an internal wiki page, I'll try moving them to the GH wiki.

jwillemsen: Removed, maybe add the guidelines to the repository itself under docs?

jwillemsen: Ok, found it, but that file needs some major updates related to svn, bug reporting, etc, maybe make an issue?

jwillemsen: Made issue #618

PR #586, opened 2017-05-31
Test extension for reproducing issue with query condition

jwillemsen:
Extended test with new dispose_qc commandline option that uses a query condition to read the data in the listener. This should be similar to the dispose test case where we get one dispose but in case of a query condition we don't get a dispose but a no data when we try to read the dispose data. Not fixed yet, this is just the test extension for the moment

### DIFF
--- a/tests/DCPS/FooTest3_0/SubDriver.cpp
+++ b/tests/DCPS/FooTest3_0/SubDriver.cpp
@@ -24,6 +24,7 @@ using namespace ::OpenDDS::DCPS;
 
 SubDriver::SubDriver()
   : num_writes_ (0),
+    num_disposed_ (0),
     shutdown_pub_ (1),
     add_new_subscription_ (0),
     shutdown_delay_secs_ (10),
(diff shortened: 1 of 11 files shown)


jwillemsen:
This is just the test extension, PR #590 has the same test extensions and a proposed fix

tests/DCPS/FooTest3_0/SubDriver.cpp
```diff
@@ -173,6 +197,25 @@ SubDriver::init(int& argc, ACE_TCHAR* argv[])
...
+#ifndef OPENDDS_NO_QUERY_CONDITION
+  if (qc_usage_)
+  {
+    DDS::StringSeq params(1);
+    params.length (1);
+    params[0] = "101010";
+
+    DDS::QueryCondition_var qc =
+      datareader_->create_querycondition (DDS::NOT_READ_SAMPLE_STATE,
+                                          DDS::NEW_VIEW_STATE | DDS::NOT_NEW_VIEW_STATE,
```
mitza-oci: Isn't this just ANY_VIEW_STATE?

tests/DCPS/FooTest3_0/SubDriver.h
```diff
@@ -12,6 +12,7 @@
 #include <vector>
 
 class DataReaderListenerImpl;
+class DataReaderQCListenerImpl;
```
mitza-oci: The forward declaration seems to be unused.

jwillemsen: Removed

tests/DCPS/FooTest3_0/SubDriver.cpp
jwillemsen: Yes

jwillemsen:
Closing, got this into #590

PR #589, opened 2017-05-31
Updated locking for RtpsUdpDataLink::remove_sample

mitza-oci:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -126,6 +126,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink {
 private:
   virtual void stop_i();
   virtual void send_i(TransportQueueElement* element, bool relink = true);
+  RemoveResult remove_sample(const DataSampleElement* sample);
```
jwillemsen: Virtual here?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -459,6 +459,14 @@ RtpsUdpDataLink::send_i(TransportQueueElement* element, bool relink)
   DataLink::send_i(element, relink);
 }
 
+RemoveResult
+RtpsUdpDataLink::remove_sample(const DataSampleElement* sample)
+{
+  // see comment in RtpsUdpDataLink::send_i() for lock order
+  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
```
jwillemsen: Shouldn't this be a ACE_GUARD_RETURN?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
mitza-oci: Yes, indeed it is.  No need to repeat the keyword.

Issue #581, opened 2017-05-31
No test with querycondition, keyed data and dispose 


jwillemsen:
We lack an OpenDDS unit test which combines query condition and keyed data. At the dds4ccm level we have one and we see that the listeners are triggered at the moment an instance is disposed but when we read the data in the lister we get no data back but we would expect instances with instance state == NOT_ALIVE_DISPOSED_INSTANCE_STATE

jwillemsen:
Add log of the dds4ccm unit test, at line 95139 the sender starts to call dispose

[querycondition.txt](https://github.com/objectcomputing/OpenDDS/files/1041660/querycondition.txt)

jwillemsen:
Created a new branch jwi-qcextension with a test extension as reproducer, needs some more cleanup before I can make a PR but on first sight it looks to reproduce our dds4ccm issue

jwillemsen:
See #586 for a test extension

jwillemsen:
Problem is that ->registered_data_ for the dispose sample is nil which means we can't apply filtering

```
>	DcpsFooType3d.dll!DDS_3_11_0::OpenDDS::DCPS::RakeResults<TAO_2_4_3::TAO::DCPS::ZeroCopyDataSeq<Xyz::Foo,20> >::insert_sample(DDS_3_11_0::OpenDDS::DCPS::ReceivedDataElement * sample, DDS_3_11_0::OpenDDS::DCPS::RcHandle<DDS_3_11_0::OpenDDS::DCPS::SubscriptionInstance> instance, unsigned __int64 index_in_instance) Line 89	C++
 	DcpsFooType3d.dll!DDS_3_11_0::OpenDDS::DCPS::DataReaderImpl_T<Xyz::Foo>::take_i(TAO_2_4_3::TAO::DCPS::ZeroCopyDataSeq<Xyz::Foo,20> & received_data, DDS_3_11_0::DDS::SampleInfoSeq & info_seq, int max_samples, unsigned int sample_states, unsigned int view_states, unsigned int instance_states, DDS_3_11_0::DDS::QueryCondition * a_condition) Line 1315	C++
[...]

PR #590, opened 2017-06-01
Test extension and fix for using query conditions on keyed data

jwillemsen:
Test extension and fix for using query conditions together with keyed_data and using dispose, see issue #550 and #581


jwillemsen:
Some tests are now failing because with NOT_ALIVE_DISPOSED_INSTANCE_STATE we could get valid_data at the moment there are keys used. Before updating these few tests we first have to decide whether the idea of the change (to keep the received data with a dispose so that we can apply a filter) is the direction to go

mitza-oci:
I cancelled the Jenkins build on this one.  Need time on the server for other PRs which are nearly ready to merge.

mitza-oci:
We do want the ability to query the disposed data.  However, to maintain some safety we should check if the query refers to any non-key fields.  Such queries would not mach any disposed data samples (current behavior).

mitza-oci:
See PR #900

PR #594, opened 2017-06-02
Do not merge: PR for testing merge of other PRs

mitza-oci:


jwillemsen:
586 and 590 will result in some new failing tests, all other PRs should be fine. These two PRs fix the querycondition with dispose but changes that we can get valid data with a dispose which require some more test changes when you agree

jwillemsen:
If there are merge problems due to conflicting changes, just assign a PR back to me and I will remerge with master and fix the conflict.

jwillemsen:
Maybe merge the PRs which are ok first to master, then I can fix any other comments tomorrow and after that look at 586 and 590?

PR #601, opened 2017-06-06
Changed num_zero_copies to has_zero_copies, we are only interested wh…

jwillemsen:
…ether there are zero copies, no need to calculate how many exactly

    * dds/DCPS/DataReaderImpl.h:
    * dds/DCPS/SubscriberImpl.cpp:

### DIFF
--- a/dds/DCPS/DataReaderImpl.h
+++ b/dds/DCPS/DataReaderImpl.h
@@ -423,7 +423,7 @@ public:
    * @retval true We have zero-copy samples loaned out
    * @retval false We have no zero-copy samples loaned out
    */
-  virtual bool has_zero_copies();
+  bool has_zero_copies();
 
   /// Release the instance with the handle.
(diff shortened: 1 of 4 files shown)


dds/DCPS/DataReaderImpl.h
```diff
@@ -417,11 +417,13 @@ class OpenDDS_Dcps_Export DataReaderImpl
...
-  /** This method is used for a precondition check of delete_datareader.
+  /**
+   * This method is used for a precondition check of delete_datareader.
    *
-   * @returns the number of outstanding zero-copy samples loaned out.
+   * @retval true We have zero-copy samples loaned out
+   * @retval false We have no zero-copy samples loaned out
    */
-  virtual int num_zero_copies();
+  virtual bool has_zero_copies();
```
mitza-oci: I know you didn't change this, but as I reviewed it I noticed that there's no reason for this to be virtual.

PR #605, opened 2017-06-06
Initialise local variables to resolve gcc warnings

jwillemsen:
* dds/DCPS/RTPS/RtpsDiscovery.cpp:


mitza-oci:
There are no warnings on the current version of GCC.  We are attempting to keep the code warning-free on the latest versions of each major compiler.  The older versions seem to be too prone to issue warnings on perfectly reasonable code.

jwillemsen:
The warnings are reported in one of the gcc builds sponsored by OCI on the scoreboard, wouldn't harm to initialize them to zero?

mitza-oci:
Just to follow-up on the discussion here.  In general I'd rather have code that's uncluttered by workarounds to support old compilers that issue warnings that they should not.  There will always be scoreboard builds on these older compilers since that's what users have.

PR #608, opened 2017-06-06
Sample lost test extension

jwillemsen:
Extend the test to log the number of sample lost, sample rejected, and budget exceeded callbacks


jwillemsen:
This test now expects that #607 has been merged, else it will report an error that it got budget exceeded callbacks

PR #616, opened 2017-06-08
Added --doc_group option that will download ACE+TAO from the DOC_Group

jwillemsen:


configure
```diff
@@ -21,7 +21,9 @@ use strict;
 use warnings;
 
 ## Version of OCI's TAO to download
-my $tao_version = '2.2a';
+my $oci_tao_version = '2.2a';
+## Version of DOC Group ACE/TAO to download, uses the ACE version number
+my $doc_tao_version = '6.4.3';
```
mitza-oci: Is this file going to need to change after each DOC release?  Would it be better to have a copy or symlink on the Vanderbilt server that doesn't have the version number?

configure
```diff
@@ -528,11 +531,19 @@ if ($opts{'safety-profile'}) {
...
+    $file = 'ACE+TAO-src-' . $doc_tao_version . '.'
+      . ($slash eq '/' ? 'tar.gz' : 'zip');
+  } else {
+    $urlbase = 'http://download.ociweb.com/TAO-' . $oci_tao_version . '/';
+    $file = 'ACE+TAO-' . $oci_tao_version . '_with_latest_patches_NO_makefiles.'
+      . ($slash eq '/' ? 'tar.gz' : 'zip');
+  }
   $tao_maj = 2 if $opts{'dry-run'};
-  my $download_message = "Downloading ACE+TAO $tao_version with latest patches";
+  my $download_message = "Downloading ACE+TAO $file";
```
mitza-oci: Please update this to maintain the current message when --doc_group is not used.  The reason for this is that "OCITAO 2.2a" and "OCITAO 2.2a with latest patches" are different versions.

configure
jwillemsen: Yes, it should be changed to my idea, else if someone has for example OpenDDS 3.12 and runs configure two years in the future he will get the latest ACE/TAO from that moment, that is probably not tested and will not work

PR #617, opened 2017-06-08
Fix issue where setting DCPSBitTransportPort without DCPSBitTransport…

PaulCalabrese:
…IPAddress has no effect


dds/DCPS/InfoRepoDiscovery/InfoRepoDiscovery.cpp
```diff
@@ -291,8 +291,12 @@ InfoRepoDiscovery::bit_config()
     TcpInst_rch tcp_inst = static_rchandle_cast<TcpInst>(inst);
 
     tcp_inst->datalink_release_delay_ = 0;
-    tcp_inst->local_address(bit_transport_port_,
-                            bit_transport_ip_.c_str());
+    if (bit_transport_ip_ != "") {
```
mitza-oci: Use .empty() instead of comparison with "" for std::string.

PR #621, opened 2017-06-09
Fix FooTest5_0 test failures

jwillemsen:


jwillemsen:
Adding some more logging and fixed the failure of the FooTest5_0 test on the scoreboard, the test didn't cleanup the DDS entities in case it tests for a incompatible QoS

dds/DCPS/PublisherImpl.cpp
```diff
@@ -61,7 +61,8 @@ PublisherImpl::~PublisherImpl()
     ACE_ERROR((LM_ERROR,
         ACE_TEXT("(%P|%t) ERROR: ")
         ACE_TEXT("PublisherImpl::~PublisherImpl, ")
-        ACE_TEXT("some datawriters still exist.\n")));
+        ACE_TEXT("%d datawriters and %d publications still exist.\n"),
```
mitza-oci: size_t needs a different logging formatter to avoid 32-bit/64-bit mismatch (also in SubscriberImpl)

PR #632, opened 2017-06-12
Added support to run DPFactoryQos and Lifespan using rtps discovery

jwillemsen:
* tests/DCPS/DPFactoryQos/rtps_disc.ini:
      Added.

    * bin/dcps_tests.lst:
    * tests/DCPS/DPFactoryQos/run_test.pl:


tests/DCPS/DPFactoryQos/rtps_disc.ini
```diff
@@ -0,0 +1,13 @@
+[common]
+DCPSGlobalTransportConfig=$file
+
+[domain/411]
```
mitza-oci: Domain IDs above 231 are not supported by RTPS Discovery (see docs/design/RTPS and devguide).

tests/DCPS/DPFactoryQos/run_test.pl
```diff
@@ -12,11 +12,19 @@
 use strict;
 
 my $test = new PerlDDS::TestFramework();
+my $is_rtps_disc = 0;
```
mitza-oci: These changes in run_test.pl shouldn't be needed.  The Perl module for TestFramework knows about rtps_disc and make sure setup_discovery() does the right thing.

jwillemsen:
Will fix these later today

tests/DCPS/DPFactoryQos/rtps_disc.ini
jwillemsen: There are a lot of tests using domain id 411, should we update all?

tests/DCPS/DPFactoryQos/run_test.pl
jwillemsen: Need to test this, are a few tests that all do this

mitza-oci: tests/DCPS/Messenger is an example that uses PerlDDS::TestFramework (though it's fairly complex since it has a lot of different possible configurations), $is_rtps_disc was left in there but it doesn't do anything.  There are probably better tests to use as an example, though not sure which ones off the top of my head.  You could look at those that have an "rtps_disc" entry in test list and find the ones with minimal number of lines in run_test.pl

tests/DCPS/DPFactoryQos/rtps_disc.ini
mitza-oci: If those tests are using RTPS Discovery, the port number will be wrong (the formula in the spec results in a number > 65535 and the actual port will wrap around).  It's not a big deal just for tests but if a user sees it and tries it in real code, there may be some problems with interop or if we end up checking the range and issuing a warning or error.

jwillemsen: Updated all 411 to 111

tests/DCPS/DPFactoryQos/run_test.pl
jwillemsen: Made a separate issue for checking is_rtps_disc

tests/DCPS/DPFactoryQos/run_test.pl
```diff
@@ -12,19 +12,17 @@
 use strict;
 
 my $test = new PerlDDS::TestFramework();
-my $is_rtps_disc = 0;
 my $DCPScfg = "";
 
 if ($ARGV[0] eq "rtps_disc") {
   $DCPScfg = "-DCPSConfigFile " . $ARGV[0] . ".ini ";
```
mitza-oci: Now $DCPScfg is unused so this block can go away (or simply revert the file to its state before the PR started)

tests/DCPS/DPFactoryQos/run_test.pl
```diff
@@ -14,11 +14,6 @@
 my $test = new PerlDDS::TestFramework();
 my $DCPScfg = "";
```
mitza-oci: Unused variable is still here.

mitza-oci:
There are some scoreboard failures for LivelinessTimeout, Lifespan, and DPFactoryQos with rtps_disc.  Are you already working on those?

jwillemsen:
Was working at the ddsx11 layer today and didn't see them early this morning. Will be tomorrow before I can hava a loon

PR #641, opened 2017-06-13
Fixed transport client leakage

jwillemsen:
DPFactoryQoS now doesn't report any leaked datareaders anymore, making a PR to get this to compile on CI


jwillemsen:
With these changes and the changes from #642 combined I was able to run the DPFactoryQos test with RTPS discovery memory leak free.

jwillemsen:
Reconnect test seems to fail because on_connection_deleted isn't called, but that is called from the destructor of TcpConnection which is also refcounted.

jwillemsen:
Have been looking at the Reconnect test and I get the impression that we as DataWriter don't get the on_connection_deleted because we now cleanly detach from the transport before the event happens. Previously the transport kept the datawriter (with associated listener) alive. At the moment I remove the cleanup_transport from the DataWriter::cleanup() the test does run. Could it that we get the event now after the DataWriter cleanup has been called?

mitza-oci:
retest this please
(starting Jenkins build)

mitza-oci:
retest this please
(starting Jenkins build)

mitza-oci:
retest this please
(starting other Jenkins build)

jrw972:
retest this please
(starting other{2} Jenkins build)

jrw972:
retest this please
(starting Jenkins build)

dds/DCPS/DiscoveryBase.h
```diff
@@ -1224,7 +1221,10 @@ namespace OpenDDS {
         if (part == domain->second.end()) {
           return false;
         }
-        participant = part->second;
+        // Use reference counting to ensure participant
```
jwillemsen: Removed the comment

mitza-oci:
When working on PR #710 we found that an explicit method call such as TransportClient::cleanup_transport() was not necessary to fix leaks.  If there are still scenarios where this does help, please let us know details of which tests to run.

jwillemsen:
Let us wait on the valgrind build, when #710 would fix the same issue we should see a decrease in test failures

jwillemsen:
Did run some tests locally and the results are much better, good to see a different and more easier fix. Restarted the valgrind build to get fresh results

PR #643, opened 2017-06-13
Don't put the listeners on the stack, they are reference counted

jwillemsen:
* tests/DCPS/StaticDiscoveryReconnect/StaticDiscoveryTest.cpp:


mitza-oci:
There is a utility class LocalObject_NoRefCount in LocalObject.h.  If you can guarantee that the lifetime of the object exceeds any use of it by the middleware, this is just as safe and may be a bit cleaner.  I'm not sure that would work in this case since I haven't read the rest of the test.

Please avoid unnecessary changes like struct -> class and using .in().

PR #648, opened 2017-06-13
Fix gcc7.1 warnings

jwillemsen:
* dds/DCPS/Serializer.cpp:

### DIFF
--- a/dds/DCPS/Serializer.cpp
+++ b/dds/DCPS/Serializer.cpp
@@ -62,36 +62,52 @@ Serializer::swapcpy(char* to, const char* from, size_t n)
   switch (n) {               // 2   4   8   16
   case 16:
     to[ 15] = from[ n - 16]; // x   x   x    0
+    // fallthrough
   case 15:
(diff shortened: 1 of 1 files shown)


dds/DCPS/Serializer.cpp
```diff
@@ -62,36 +62,52 @@ Serializer::swapcpy(char* to, const char* from, size_t n)
   switch (n) {               // 2   4   8   16
   case 16:
     to[ 15] = from[ n - 16]; // x   x   x    0
+    // fallthrough
```
mitza-oci: GCC is changing its output of warnings based on a comment?  I guess that's good??

mitza-oci: https://gcc.gnu.org/onlinedocs/gcc-7.1.0/gcc/Warning-Options.html

It would be good to have ACE_CASE_FALLTHROUGH so it can expand to [[fallthrough]]; or one of its alternatives and it would be less likely to be removed by someone thinking it's just a comment.

PR #650, opened 2017-06-14
No need to unregister instances which have been unregistered already,…

jwillemsen:
… this could happen when the user has already unregistered his instances explicitly and after that deletes the datawriter


dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1704,10 +1704,14 @@ DataWriterImpl::unregister_instances(const DDS::Time_t& source_timestamp)
       this->data_container_->instances_.begin();
 
     while (it != this->data_container_->instances_.end()) {
-      DDS::InstanceHandle_t handle = it->first;
-      ++it; // avoid mangling the iterator
-
-      this->unregister_instance_i(handle, source_timestamp);
+      // Only unregister instance that haven't been unregistered yet
```
mitza-oci: This comment isn't adding any new info.

jwillemsen: Will remove asap

PR #651, opened 2017-06-14
In delete_publisher also call delete_contained_entities on the publis…

jwillemsen:
…her so that we can the same callchain for subscriber and publisher


mitza-oci:
The spec says that DomainParticipant::delete_publisher may not be called to delete Publisher that has any contained entities (DataWriters).  This is already checked so delete_contained_entities() is not needed.

jwillemsen:
Ok, subsciber does call this

Issue #654, opened 2017-06-15
DataWriter/DataReader cleanup should set its listener to 0


jwillemsen:
At the moment the delete_datawriter/delete_datareader is called we call cleanup in the DR/DW. At that moment it is not guaranteed that the dr/dw destructor will run due to ref counting. We should set the listener in the dw/dr to 0 in the cleanup so that we don't invoke the listener anymore after the dr/dw has been deleted from the viewpoint of the user. This could explain the last remaining test failures #641, listeners are called after the dw/dr has been deleted from the user perspective.

jwillemsen:
Fixes merged

PR #656, opened 2017-06-16
At the moment the users calls delete_datawriter/delete_datareader it …

jwillemsen:
…can happen the DDS entity is deleted for the user but the underlying object still exists due to reference counting. As first step in the cleanup now set the listener to nil so that we don't callback onto the listener after the user has called delete on the DR/DW, see issue #654


jwillemsen:
@mitza-oci In the OCI build the test `tests/DCPS/Reconnect/run_test.pl bp_timeout` fails. As far as I can determine this test incorrectly assumes that a listener is invoked after the DW has been deleted. Adam, any suggestions?

jwillemsen:
On master I see in our build also a failure for `tests/DCPS/Reconnect/run_test.pl restart_sub`

mitza-oci:
Let's focus on just `tests/DCPS/Reconnect/run_test.pl bp_timeout` for now (not restart_sub), since it seems to fail consistently on this branch.  I don't know that it is assuming a listener is called after DW is deleted.  It does want the listener to be invoked at some point.  If it is deleting the DW too early, could that be fixed by changing when the subscribing program declares itself finished via subscriber_finished.txt?

jwillemsen:
The subscriber program does use that file

mitza-oci:
retest this please
(starting Jenkins build)

PR #658, opened 2017-06-16
Change dp::is_clean to return a bool instead of int

jwillemsen:
See issue #571

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -163,7 +163,7 @@ DomainParticipantImpl::delete_publisher(
     return DDS::RETCODE_ERROR;
   }
 
-  if (the_servant->is_clean() == 0) {
+  if (!the_servant->is_clean()) {
     ACE_ERROR((LM_ERROR,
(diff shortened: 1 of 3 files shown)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1747,22 +1747,22 @@ DomainParticipantImpl::create_new_topic(
...
+  bool sub_is_clean = subscribers_.empty();
+  bool topics_is_clean = topics_.size() == 0;
 
   if (!TheTransientKludge->is_enabled()) {
     // There are four topics and builtin topic subscribers
     // left.
 
-    sub_is_clean = sub_is_clean == 0 ? subscribers_.size() == 1 : 1;
-    topics_is_clean = topics_is_clean == 0 ? topics_.size() == 4 : 1;
+    sub_is_clean = !sub_is_clean ? subscribers_.size() == 1 : true;
```
mitza-oci: Not part of your change, but shouldn't this be checking if Built-In Topics are enabled before assuming one subscriber and four topics?

jwillemsen: Maybe make a new issue to check later?

PR #659, opened 2017-06-16
Attempt to fix a problem in versioned namespace build

mitza-oci:


jwillemsen:
This looks risky because another product (fe ACE or TAO) could also have a Version.h and we could pickup that version. Using dds/Version.h makes sure we get the OpenDDS version.h

mitza-oci:
`#include "Version.h"` finds the file Version.h either in the directory of the file doing the including (which is found first), or if it's not found there, in one of the system or -I directories.

https://gcc.gnu.org/onlinedocs/gcc-7.1.0/cpp/Search-Path.html#Search-Path

jwillemsen:
ok

PR #661, opened 2017-06-17
Corrected check for when to use find_POA

jwillemsen:
* dds/DCPS/InfoRepoDiscovery/InfoRepoDiscovery.cpp:


mitza-oci:
I see that the current code is wrong however I don't think we can based the check on CORBA_E_MICRO.  If someone actually tries to build with a Micro configuration it will fail due to create_POA() a few lines down.

jwillemsen:
Is it supported to compile OpenDDS with CORBA/e micro, looks create_POA is also disabled in that case as you pointed out. If it is not supported, we could always use find_POA

mitza-oci:
Building with CORBA/e micro is not supported (unless you consider OpenDDS safety profile which is really more than that, it doesn't even have libTAO).  We do support building with CORBA/e compact as there is a scoreboard build for it.  So the macros should be in terms of compact and not micro.

I'm not sure how the code got to the current state but I agree that the alternate path shouldn't be needed, just find_POA().

PR #664, opened 2017-06-19
DDS Vendor Interoperability Table

matthewpang:
Added a graphic showing DDS vendor interoperability


ChuckRenner:

mitza-oci:
I don't see this as being a candidate for in-repo documentation since it's also dependent on specific versions of 3rd-party products.  How about keeping it as a GitHub issue (or two, one for each interop problem)?  That way we can track progress and test results.  I have some test results with OpenSplice that I can add to it.

matthewpang:
Sounds good lets do that!

PR #666, opened 2017-06-21
Add osx to travis-ci

jwillemsen:
* .travis.yml:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ compiler:
 env:
   matrix:
     - FUZZ=1
-    - NO_INLINE=1
-    - NO_BIT=1
-    - STATIC=1 NO_DEBUG=1
-    - SAFETY_PROFILE=1
-    - USES_WCHAR=1
+    - NO_INLINE=1 XERCES3=1
+    - NO_BIT=1 XERCES3=1
+    - STATIC=1 NO_DEBUG=1 XERCES3=1
+    - SAFETY_PROFILE=1 XERCES3=1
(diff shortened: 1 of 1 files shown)


jwillemsen:
Looks the warnings for the usage of register are causing travis-ci to fail for osx. travis is using ACE+TAO latest micro, fixes for that are on master

mitza-oci:
The Travis config would need to be updated with -Wno-deprecated-register until after the DOC Group release.

jwillemsen:
Should we do this in the travis-ci file or within configure? Probably configure is a better place given the possibility to also use OCI 2.2a which does use register

mitza-oci:
It's only a problem for Travis, a normal build doesn't error-out due to this:

> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated

jwillemsen:
I disabled inlining, that should also reduce the size of the log file

.travis.yml
```diff
@@ -41,7 +41,9 @@ matrix:
           packages:
             - libxerces-c-dev
             - clang++-3.9
-      env: COMPILER=clang++-3.9
+      env: COMPILER=clang++
+    - os: osx
+      env: COMPILER=clang++ XERCES3=0 INLINE=0
```
mitza-oci: There is already NO_INLINE=1 on line 10, let's not have both variables.

jwillemsen: Missed that, updated the file

PR #668, opened 2017-06-22
Introduce a new Time_Helper.{h,cpp,inl} which only provides the helpe…

jwillemsen:
…r for time. This way the QoS_Helper.h has only to be included in some implementation files and is not visible anymore from the generated TypeSupportImpl

    * dds/DCPS/Time_Helper.cpp:
    * dds/DCPS/Time_Helper.h:
    * dds/DCPS/Time_Helper.inl:
      Added.

    * dds/DCPS/DataDurabilityCache.cpp:
    * dds/DCPS/DataWriterImpl.h:
    * dds/DCPS/InstanceState.cpp:
    * dds/DCPS/OfferedDeadlineWatchdog.cpp:
    * dds/DCPS/Qos_Helper.h:
    * dds/DCPS/Qos_Helper.inl:
    * dds/DCPS/ReceivedDataElementList.h:
    * dds/DCPS/ReceivedDataStrategy.cpp:
    * dds/DCPS/ReplayerImpl.h:
    * dds/DCPS/RequestedDeadlineWatchdog.cpp:
    * dds/DCPS/WaitSet.cpp:
    * dds/DCPS/WriteDataContainer.cpp:
    * dds/DCPS/WriterInfo.cpp:
    * dds/DdsDcpsCore.idl:
    * dds/DdsDcpsInfrastructure.idl:


dds/DCPS/Time_Helper.inl
```diff
@@ -0,0 +1,245 @@
...
+ */
+
+#include "ace/OS_NS_string.h"
+#include "ace/Truncate.h"
+
+#include <cstring>
+
+namespace {
+
+const ACE_UINT32 NSECS_IN_SEC = 1000000000;
```
mitza-oci: I realize this was the same in the old file, but I'd rather not have anonymous-namespace objects in .inl files.  Since the constant is only used in one function we could have it as a static const local to that function.

dds/idl/be_produce.cpp
```diff
@@ -206,6 +206,7 @@ void postprocess(const char* fn, ostringstream& content,
     }
     if (which == BE_GlobalData::STREAM_H) {
       out << "#include \"dds/DCPS/Definitions.h\"\n";
+      out << "#include \"dds/DdsDcpsC.h\"\n";
```
mitza-oci: Could this be an include of a more specific header?

jwillemsen: This file includes again DdsDcpsDomainC.h which defines for example the DomainParticipant, we need to expose that to the user

jwillemsen: We need DdsDcpsC.h else we don't get all IDL defined DDS types visible for the user. Tried some more specific but that lead to compile issues in tests

PR #670, opened 2017-06-22
Fixed typo in repoctl console output

jwillemsen:
* tools/repoctl/repoctl.cpp:


mitza-oci:
Updated title of PR to describe what it is

PR #671, opened 2017-06-22
Removed _var_ptr/_ptr_type from the generated type support implementa…

jwillemsen:
…tion, we can get these from the base, also remove _narrow, also usable from the IDL generated stub

    * dds/idl/ts_generator.cpp:

### DIFF
--- a/dds/idl/ts_generator.cpp
+++ b/dds/idl/ts_generator.cpp
@@ -162,8 +162,6 @@ bool ts_generator::gen_struct(AST_Structure*, UTL_ScopedName* name,
       "public:\n"
       "  typedef OpenDDS::DCPS::DDSTraits<" << short_name << "> TraitsType;\n"
       "  typedef " << short_name << "TypeSupport TypeSupportType;\n"
-      "  typedef " << short_name << "TypeSupport::_var_type _var_type;\n"
-      "  typedef " << short_name << "TypeSupport::_ptr_type _ptr_type;\n"
       "\n"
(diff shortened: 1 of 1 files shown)


mitza-oci:
The code you removed is there for a reason.
```
error C2385: ambiguous access of '_var_type'
 note: could be the '_var_type' in base 'OpenDDS::DCPS::LocalObject<Messenger::MessageTypeSupport>'
 note: or could be the '_var_type' in base 'OpenDDS::DCPS::LocalObject<OpenDDS::DCPS::TypeSupport>'
```

But this illustrates that we need some test "coverage" of this compile-time ability.

jwillemsen:
According to the comment in LocalObject.h that template is there to resolve ambiguity, doesn't that work? Isn't it very risky to allow to use objects on the stack, they could be gone at the moment we get delayed callbacks.

mitza-oci:
re: Comment in LocalObject, I believe that works for single inheritance but in this case we have TypeSupportImpl which has it's own definition of these members.

re: on the stack, I don't see that being relevant here.  I took a test that used listeners normally and had it use the _var_type typedef instead (as one might do when writing code in a template), with your change it failed (see earlier comment).

jwillemsen:
I can add the _var_type/_ptr_type back again, what about narrow, also required?

mitza-oci:
I believe those all work the same way, it's just class member lookup with multiple inheritance.  I think we should just close this PR, I don't see anything changing here.

PR #675, opened 2017-06-23
Use a scope to release the guard, much safer than manually releasing …

jwillemsen:
…the guard. Also reduce the number of exit points

    * dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp:

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -240,13 +240,12 @@ RtpsUdpDataLink::get_locators(const RepoId& local_id,
   }
 
   const GUIDSeq_var peers = peer_ids(local_id);
-  if (!peers.ptr()) {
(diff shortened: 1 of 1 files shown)


mitza-oci:
I think the limited use of ``guard.release()`` is OK when it's just for the last action in the method.  It avoids unnecessarily complicating the method.  Same for single-exit-point style, it's unnecessarily complicating things.

jwillemsen:
Also the guard is safer in case exceptions occur

jwillemsen:
What about the risk of exceptions? Isn't the scope making sure that we always release the lock?

mitza-oci:
The general structure of these methods is:
```
void foo()
{
  Guard g(lock);
  action1();
  g.release();
  action2();
}
```
Where action1 requires the lock and action2 must be called without the lock.  What is the "risk of exceptions" here?  If action1 throws, the lock is released due to g's destructor.  If action2 throws, the lock is already released by design.

PR #683, opened 2017-06-26
Also set -std=... for the host tree when the user has specified a -st…

jwillemsen:
…d= flag


mitza-oci:
We don't know that the host and target are the same compiler, so --std=foo could be an error for the host compiler.  Is it actually needed for anything?  The host is just building self-contained compile-time tools so the version of C++ std shouldn't matter.

jwillemsen:
Yes, you are right, closing

PR #688, opened 2017-06-27
Guard several generated methods with OPENDDS_NO_MULTI_TOPIC, they are…

jwillemsen:
… not necessary when the user has disabled multitopics. Extend travis-ci build with a configuration to also build with multi topic disabled

    * .travis.yml:
    * dds/DCPS/FilterEvaluator.h:
    * dds/idl/metaclass_generator.cpp:


.travis.yml
```diff
@@ -12,6 +12,7 @@ env:
     - STATIC=1 NO_DEBUG=1
     - SAFETY_PROFILE=1
     - USES_WCHAR=1
+    - MULTITOPIC=1
```
mitza-oci: Should this be "NO_MULTITOPIC"?

Issue #506, opened 2017-04-10
Issues in dp::delete_contained_entities and related operations


jwillemsen:
The domain participant `delete_contained_entities` is going through the reactor to delete all contained entities and at the end it returns `shutdown_result_`. When I now check the DomainParticipantImpl::handle_exception implementation I notice that in several error conditions a return is done which returns the DDS retcode to the reactor instead of `shutdown_result_` is set to the error value. As far as I can tell this could lead to a problem that an error code is not returned to the caller which doesn't see the error. For debugging purposes it would also help if some logging is added in the calls to all delete_contained_entities to log what is deleted (especially when it fails)

jwillemsen:
In `SubscriberImpl::delete_contained_entities` not all members maps are cleared, something to check in this class but also all others that implement `delete_contained_entities`

jwillemsen:
This method and others use the `set_deleted` flag, but that flag is only reset at the moment there is no failure, shouldn't a guard be used so that the flag is also reset at the moment there is a failure?

mitza-oci:
@jwillemsen since #692 was merged, is this issue resolved?

jwillemsen:
Very likely, will close

PR #692, opened 2017-06-29
Instead of returning an error we have to set the shutdown_result_, th…

jwillemsen:
…is method is called from the ACE reactor which is not aware of the DDS return codes

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -2221,13 +2221,13 @@ DomainParticipantImpl::handle_exception(ACE_HANDLE /*fd*/)
       DDS::ReturnCode_t result = subPtr->delete_contained_entities();
 
       if (result != DDS::RETCODE_OK) {
-        ret = result;;
+        ret = result;
       }
 
       result = delete_subscriber(subPtr);
 
       if (result != DDS::RETCODE_OK) {
-        ret = result;;
+        ret = result;
       }
 
       --subsize;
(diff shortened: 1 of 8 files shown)


jwillemsen:
Part of #506

mitza-oci:
retest this please
(starting Jenkins build)

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -2187,16 +2189,16 @@ DomainParticipantImpl::handle_exception(ACE_HANDLE /*fd*/)
       = pubPtr->delete_contained_entities();
 
       if (result != DDS::RETCODE_OK) {
-        return result;
+        ret = result;
```
mitza-oci: Before this change, a non-OK result stopped further processing.  Now it's proceeding with the next steps.  Is that what we want?

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -2219,20 +2221,19 @@ DomainParticipantImpl::handle_exception(ACE_HANDLE /*fd*/)
       DDS::ReturnCode_t result = subPtr->delete_contained_entities();
 
       if (result != DDS::RETCODE_OK) {
-        return result;
+        ret = result;;
```
mitza-oci: This line has two ;s (also below on 2230)

dds/DCPS/DomainParticipantImpl.cpp
jwillemsen: Not 100% sure, maybe the next steps will result in other errors but this gives the easiest control flow

jrw972:
retest this please

jrw972:
retest this please

mitza-oci:
@huangminghuang please look at this change to DomainParticipantImpl (not necessarily other files) and see if it can be combined with what you're working on

PR #693, opened 2017-06-30
Disable std::(w)string insertion/extraction when using safety profile

jwillemsen:
* dds/DCPS/Serializer.h:
    * dds/DCPS/Serializer.inl:


mitza-oci:
Safety profile travis error:
`SerializerTest.cpp:74:14: error: no match for ‘operator<<’`

Issue #682, opened 2017-06-26
Configure script should update config.h/platform_macros.GNU when they exist


jwillemsen:
At the moment the configure script finds an existing ACE/TAO tree it only recreates default.features. It should also recreate config.h/platform_macros.GNU and safe the old version as backup. I did wanted to build with safety-profile and this requires modified files which where not updated by configure because I have an existing tree

mitza-oci:
It may be a bug that default.features is being overwritten.  The configure script is not designed to re-configure.  You can start from a clean tree using `git clean` (or for an end-user, just removing the working directory and re-expanding the archive).

There are essentially two modes supported by configure:
- configure and build ACE+TAO along with OpenDDS
- use an existing build of ACE+TAO and just configure/build OpenDDS

If the latter mode is being used, we may not overwrite ACE's configuration.  OpenDDS will append to default.features (should be features that ACE+TAO doesn't user).

jwillemsen:
Haven't checked, but could be it will append to default.features, but by appending there and not to platform_macros.GNU I think we get inconsistent configurations, especially when special flags are passed to configure

mitza-oci:
user_macros.GNU was created for that reason.  It is used by configure.

jwillemsen:
When using special configure flags and an existing build we could check if the necessary flags are available in the config.h/platform_macros.GNU/default.features, not that the user thinks he has enabled some special configuration but that it doesn't work due to an existing ACE tree

PR #708, opened 2017-08-14
Configure improvements

jrw972:
Fix to not overwrite/append to default.features.  See #682.


configure
```diff
@@ -1148,6 +1157,84 @@ EOT
...
+        my ($key, $value) = split(/=/, $f);
+        $value = "1" unless defined $value;
+        $requested_macros{$key} = $value;
+      }
+      for my $f (@{$opts{'macros'}}) {
+        my ($key, $value) = split(/=/, $f);
+        $value = "1" unless defined $value;
+        $requested_macros{$key} = $value;
+      }
+      for my $key ('debug', 'optimize', 'inline', 'static', 'ipv6') {
```
mitza-oci: This looks like the @ace_macros list defined above

configure
```diff
@@ -1148,6 +1157,84 @@ EOT
...
+      }
+        # Convert the requested macros.
+      my %requested_macros;
+      for my $f (@platformmacros) {
+        my ($key, $value) = split(/=/, $f);
+        $value = "1" unless defined $value;
+        $requested_macros{$key} = $value;
+      }
+      for my $f (@{$opts{'macros'}}) {
+        my ($key, $value) = split(/=/, $f);
```
mitza-oci: Since this loop is the same as the one above, they could be combined.

configure
```diff
@@ -1148,6 +1157,84 @@ EOT
       }
       dump_and_unlink($DF) if $opts{'dry-run'};
     }
+
+    if (!$buildtao && !defined $opts{'safety-profile'}) {
+      # Read the existing features.
```
mitza-oci: This comment and those below in this chunk can probably be removed (or revised?)

Issue #694, opened 2017-07-07
Dependencies when building Java bindings (generated file not present when needed)


dmorilha:
I am getting an error while trying to build OpenDDS from the GitHub source revision: `23c3c5f5b05846d85cf9853de4946200f8a64b5e`

Here's the error:

```
GNUmakefile: /Users/daniel/src/OpenDDS/java/dds/GNUmakefile.dcps_java MAKEFLAGS=

perl ../../java/build_scripts/javac_wrapper.pl -sourcepath . -d classes -classpath . -implicit:none -classpath ../../lib/i2jrt.jar DDS/DataWriterListener.java
./DDS/TopicDescriptionOperations.java:5: error: cannot find symbol
  DDS.DomainParticipant get_participant();
     ^
  symbol:   class DomainParticipant
  location: package DDS
./DDS/PublisherOperations.java:16: error: cannot find symbol
  DDS.DomainParticipant get_participant();
     ^
  symbol:   class DomainParticipant
  location: package DDS
2 errors
make[1]: *** [classes/DDS/DataWriterListener.class] Error 1
make: *** [dcps_java] Error 2
Daniels-MBP:OpenDDS daniel$
```

mitza-oci:
I believe you are seeing a case where the generated makefiles don't accurately capture all of the dependencies among generated files.  Please try the following as a workaround:

```
cd /Users/daniel/src/OpenDDS
. setenv.sh
cd java/dds
make -f GNUmakefile.dcps_java DDS/DomainParticipant.java
make
# if there are errors, repeat the make -f step to generate required files
# once there are no errors proceed to
cd $DDS_ROOT
make
```

dmorilha:
that was the only thing... after following your instruction everything worked fine. the only extra thing is that `make install` does nothing...

mitza-oci:
OK, as for `make install`, did your configure command include `--prefix=`?  If not it won't attempt to install.  If that still doesn't work please open a new issue.  It is possible that Mac OS X / macOS `make install` isn't working as we don't have CI for that.  It's an ACE feature but we'll help get it working.

dmorilha:
I did append `--prefix` in the `configure` line. I will open a new issue for that. Shall we close this issue here or does the Makefile deserves a fix before closing it?

mitza-oci:
Let's leave it open and see if we can get the dependencies sorted out.

jwillemsen:
[...]

PR #735, opened 2017-09-05
Use gcc7 on travis-ci

jwillemsen:


mitza-oci:
Now we seem to be running in to issue #694 on the java-enabled build.  Is that the only one using gcc7 or are all of them (besides clang) moved up to 7?

jwillemsen:
With java I reverted back to gcc6

mitza-oci:
The problem with issue #694 doesn't seem to be unique to this branch.  Something else on travis must have changed causing this to crop up on other builds as well.  I'm working on a actual fix for #694.

PR #745, opened 2017-09-19
Mention Docker tag scheme

jrw972:


mitza-oci:
Shouldn't this PR target the gh-pages branch instead of website-next-release?

PR #749, opened 2017-10-03
Fix Monitor test

PaulCalabrese:
Add new DCPSMonitor config flag to fix Monitor test and make initialization easier.
Enhance Monitor test to really check the functionality is operational.

### DIFF
--- a/dds/monitor/MonitorFactoryImpl.cpp
+++ b/dds/monitor/MonitorFactoryImpl.cpp
@@ -177,6 +177,7 @@ MonitorFactoryImpl::initialize()
   DDS::DataWriterQos dw_qos;
   publisher->get_default_datawriter_qos(dw_qos);
   dw_qos.durability.kind = DDS::TRANSIENT_LOCAL_DURABILITY_QOS;
+  dw_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
 
   OpenDDS::DCPS::ServiceParticipantReportTypeSupport_var sp_ts =
     new OpenDDS::DCPS::ServiceParticipantReportTypeSupportImpl();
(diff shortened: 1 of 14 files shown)


tests/DCPS/Monitor/run_test.pl
```diff
@@ -80,6 +84,13 @@
 
 unlink $dcpsrepo_ior;
 
+my $mon_count = `grep -c Report mon.out`;
```
mitza-oci: This should be able to run on systems without 'grep'.  Equivalent functionality takes just a few lines of perl.

dds/monitor/MonitorFactoryImpl.cpp
```diff
@@ -177,6 +177,7 @@ MonitorFactoryImpl::initialize()
   DDS::DataWriterQos dw_qos;
   publisher->get_default_datawriter_qos(dw_qos);
   dw_qos.durability.kind = DDS::TRANSIENT_LOCAL_DURABILITY_QOS;
+  dw_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
```
mitza-oci: This may increase resources needed by the monitor library.  Is it needed?

PaulCalabrese:
I'll fix the grep usage. I didn't mean to leave that in this form.

The monitor test was seeing a high variability in the number of monitoring
samples received. Things got much more consistent after I added the
KEEP_ALL on both sides. I am open to suggestions for what would be an
appropriate combination of settings (maybe set max_samples_per_instance?).



Paul

mitza-oci:
Does the test need to count the number of samples?  Would the number of instances be more useful?

PR #758, opened 2017-10-24
Trying to get Travis CI to complete within the time limit

mitza-oci:


jwillemsen:
Disabling optimization will also help to reduce time on travis-ci

mitza-oci:
That's true but not relevant to the current problem https://travis-ci.org/objectcomputing/OpenDDS/builds/290166603.  There are 10 total build configurations.  2 have optimization enabled and neither of those is a problem because I've already adjusted them to finish within the time limit.

PR #770, opened 2017-12-08
Use DOC Group TAO 2.4.6

jwillemsen:
* configure:


mitza-oci:
Can't set status; build succeeded.

PR #773, opened 2017-12-11
Made Windows header file name case compatible with MinGW headers

mitza-oci:


mitza-oci:

PR #774, opened 2017-12-11
Added an #include needed for libc++ on FreeBSD

mitza-oci:


mitza-oci:

PR #777, opened 2017-12-13
Wireshark2

iguessthislldo:
- Dissector Compatibility with Wireshark 2.x while keeping compatibility with at least 1.12.
- Filtering by sample payload content.
- Small change to the IDL compiler for ITL files to help in sample dissection.
- Using OpenDDS::DCPS::Serializer in sample dissection.
- Using Wireshark's "expert" functionality to mark some problems with packets in the GUI as well as printing the problem like before.


iguessthislldo:
I figured those gitignores would conflict when I did this.

dds/idl/itl_generator.cpp
```diff
@@ -285,6 +286,12 @@ bool itl_generator::gen_struct(AST_Structure*, UTL_ScopedName* /* name */,
                   << Open(this)
                   << Indent(this) << "\"kind\" : \"alias\",\n"
                   << Indent(this) << "\"name\" : \"" << repoid << "\",\n"
+
+  // Check if this is defined as a primary data type
+                  << Indent(this) << "\"note\" : { \"is_dcps_data_type\" : "
+                  << ((idl_global->is_dcps_type(name) == 0) ? "false" : "true")
```
mitza-oci: This could be simplified by swapping true/false so you don't need == 0 and extra parens

tools/dissector/packet-opendds.cpp
```diff
@@ -462,56 +501,141 @@ namespace OpenDDS
...
+        );
+        params.tvb = tvb_;
+        params.info = pinfo_;
+        params.tree = contents_tree;
+        params.offset = offset;
+
+        /* 
+         * Handle Seg Faults that might be caused by incorrect ITL file
+         *
+         * Check for enviromental variable $OPENDDS_DISSECTORS_SIGSEGV,
```
mitza-oci: no "al" here, it's an "environment variable"

tools/dissector/packet-opendds.cpp
```diff
@@ -573,56 +708,32 @@ namespace OpenDDS
...
-            }
-
-          dissect_dds (tvb_, pinfo_, tree_);
-        }
-      else
-        {
-          this->dissect ();
-        }
+      // This was simplified from WS 1.x code,
+      // And modified to avoid adding multiple DCPC subtrees to the packet tree
```
mitza-oci: Is DCPC a typo for DCPS or something else?

tools/dissector/sample_manager.cpp
```diff
@@ -347,6 +436,51 @@ namespace OpenDDS
...
     }
+
+    void
+    Sample_Manager::add_protocol_field(
+      int * hf_index,
+      const std::string & full_name, const std::string & short_name,
+      enum ftenum ft, field_display_e fd
+    ) {
+      // Get copies of the names
+      char * _full_name = strdup(full_name.c_str());
```
mitza-oci: Does strdup belong to std:: or ACE_OS:: or something else?  Also the _leading_underscore looks odd here.

tools/dissector/packet-opendds.cpp
iguessthislldo: No, its a typo

tools/dissector/packet-opendds.cpp
iguessthislldo: Hmm, you're right. I have always called them that in my head.

tools/dissector/sample_manager.cpp
iguessthislldo: Actually it's from [POSIX string.h](http://pubs.opengroup.org/onlinepubs/9699919799/functions/strdup.html)... and I'm pretty sure I knew that when I wrote it, so I don't know why I used it. It allocates and returns a copy of a string, which I need to keep for the rest of execution. I might change it to keep `std::string`s instead and use c_str() but I don't know how much to count on the pointer remaining valid. I can also just use strlen, new and strcpy, which does the same thing as strdup.

I guess the underscore was an attempt to signify it was a copy of the names but yea I can change them to something like `full_name_copy` if I keep the variable.

tools/dissector/packet-opendds.h
```diff
@@ -14,13 +14,16 @@ extern "C" {
...
+
 #include <glib.h>
 #include <gmodule.h>
 
 #include <epan/value_string.h>
 #include <epan/ipproto.h>
 #include <epan/packet.h>
 #include <epan/dissectors/packet-tcp.h>
+#include <epan/expert.h>
+#include <wsutil/report_message.h>
```
RomainNaour: Using wsutil/report_message.h break the build for Wireshark 2.2 (used in ubuntu 16.04)
The header report_err.h has been renamed to report_message.h with Wireshark 2.4 release.
https://github.com/wireshark/wireshark/commit/2c44a7f930967335b442e899c17e1eb46831e3c1

tools/dissector/README
```diff
@@ -9,13 +9,17 @@ supported.
 * See also
```
RomainNaour: Also change the line "The OpenDDS DCPS dissector provides a Wireshark 1.2.x/1.3.x compatible" on the top of this file. The dissector is no longer compatible with Wireshark 1.x.

tools/dissector/sample_manager.h
```diff
@@ -34,6 +34,7 @@ extern "C" {
 #include <string>
 #include <map>
 #include <vector>
+#include <forward_list>
```
RomainNaour: This is a C++11 feature [1].

This break the build here:
In file included from /usr/include/c++/5/forward_list:35:0,
                 from ../../tools/dissector/sample_manager.h:37,
                 from packet-opendds.cpp:10:
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.


[1] http://www.cplusplus.com/reference/forward_list/forward_list

tools/dissector/README
iguessthislldo: The dissector does still support Wireshark 1 or at least that is the intention, it just drops sample payload dissection support if it is compiled against 1. I have tested it with the last 1.x release, 1.12, although testing with older versions might be necessary.

tools/dissector/sample_manager.h
iguessthislldo: Thank you for catching that, I would have assumed the build system was defaulting to C++03 on my end and I missed that was 11 in the documentation.

tools/dissector/packet-opendds.h
iguessthislldo: Alright, the signature of the function I was using (`report_failure`) didn't change so I set it to include report_message if using 2.4 or greater, else report_err.

tools/dissector/README
iguessthislldo: Actually, I realized I don't know why I assumed payload dissection had to be dropped other than I had to rewrite most of code for it. I just removed the preprocessor condition and it compiled and loaded the of the pcaps I've been using. So the whole dissector should work in at least Wireshark 1.12. I would have to see about ones before that because I'm not familiar with the API changes. The only thing I know that doesn't compile with Wireshark 1.12 is the FT_CHAR field display type value.

tools/dissector/plugin.cpp
```diff
@@ -24,8 +24,13 @@ extern "C" {
...
 
 #ifndef ACE_AS_STATIC_LIBS
-extern "C"
-dissector_Export const gchar version[] = DDS_VERSION;
+
+#if WIRESHARK_VERSION >= WIRESHARK_VERSION_NUMBER(2, 5, 0)
+  extern "C" dissector_Export const gchar plugin_version[] = DDS_VERSION;
+  extern "C" dissector_Export const gchar plugin_release[] = VERSION_RELEASE;
+#else
+  extern "C"dissector_Export const gchar version[] = DDS_VERSION;
```
mitza-oci: There should be a space here between "C" and dissector_Export

tools/dissector/ws_common.h
```diff
@@ -11,27 +11,110 @@
...
+{
+  return tvb_get_string_enc(alloc, tvb, offset, length, ENC_ASCII);
+}
+#else
+# define ws_tvb_get_string ::tvb_get_string
+#endif
+
+#if WIRESHARK_VERSION >= WIRESHARK_VERSION_NUMBER(2, 2, 0)
+# define WS_DISSECTOR_RETURN_TYPE int
+# define WS_DISSECTOR_RETURN_VALUE return 0; //TODO
```
mitza-oci: What needs to be done to resolve the //TODO comment?

tools/dissector/ws_common.h
```diff
@@ -11,27 +11,110 @@
...
+# define ws_dissector_add_handle dissector_add_for_decode_as
+# define WS_CONV_IDX conv_index
+// create_dissector_handle wants the new dissector_t here
+#define WS_DISSECTOR_EXTRA_PARAM , void*
```
mitza-oci: This line isn't indented (between the # and "define") like the others in its block (following the #if / #endif structure).  There are a few more like this below, I won't leave comments for each one.

iguessthislldo:
I got the dissector compiled and running on Windows. As discussed the .lib files needed to be changed, but there were a few other changes that need to be made as well:

- An additional parameter for MPC for the Wireshark Build destination is needed for when that is different than the source directory. It is needed to include config.h which is generated when wireshark is built.

- There is type name collision as well between ACE and Wireshark at least on Windows amd64. `ssize_t` is defined in both `ACE_wrappers\ace\os_include\sys\os_types.h` and the config file mentioned above, at least for Windows x86_64. I wasn't sure how to resolve this properly and for now I macro define `ssize_t` in `ws_config.h` which is included instead of config.h.

tools/dissector/ws_common.h
iguessthislldo: That was there before me, I tried to figure out early on because it was a TODO but moved onto the sample dissection because I could understand that better. It needs to return the number of bytes it dissected. There is a related marco for the tcp dissection that returns how many more bytes the dissector needs. I don't know how applicable these are to this dissector though, given that it has worked as is, but I will look into it this week.

tools/dissector/packet-opendds.cpp
```diff
@@ -430,15 +439,24 @@ namespace OpenDDS
                    i < sample.content_filter_entries_.length();
                    i++)
                 {
+                  // Get Entry Value
                   const GUID_t &filter = sample.content_filter_entries_[i];
                   DCPS::GuidConverter converter(filter);
                   std::stringstream strm;
-                  strm << "filter [" << i << "] = " << converter << std::ends;
+                  std::string guid = strm.str();
```
mitza-oci: This doesn't look right -- `guid` will always be an empty string and the stringstream is unsued

tools/dissector/packet-opendds.cpp
```diff
@@ -462,56 +501,147 @@ namespace OpenDDS
...
-                      "no topic for %s\n",
-                      std::string(converter).c_str()));
-          offset += header.message_length_; // skip marshaled data
-          return;
-        }
 
+      if (data_name == 0) {
+        ACE_DEBUG ((LM_DEBUG,
+                    "DDS_Dissector::dissect_sample_payload: "
+                    "no topic for %s\n",
```
mitza-oci: ACE_DEBUG log messages use %C for const char* parameters, not %s

tools/dissector/packet-opendds.cpp
```diff
@@ -462,56 +501,147 @@ namespace OpenDDS
...
+
+        /*
+         * Handle Seg Faults that might be caused by incorrect ITL file
+         *
+         * Check for environment variable $OPENDDS_DISSECTORS_SIGSEGV,
+         * If it exists don't handle seg faults.
+         * Either way Wireshark will crash if the ITL file is the wrong type,
+         * or technically inconsistent size.
+         */
+        bool handle = ACE_OS::getenv(ACE_TEXT("OPENDDS_DISSECTOR_SIGSEGV")) == 0;
```
mitza-oci: Add this to README?

tools/dissector/packet-opendds.cpp
```diff
@@ -828,41 +940,86 @@ namespace OpenDDS
                 }
         },
         { &hf_sample_content_filt,
+            { "Content Filters",
```
mitza-oci: Is this the boolean flag for content filter?  Maybe change the name the "Content Filters" to something else since the one below also uses that.

tools/dissector/packet-opendds.cpp
iguessthislldo: I discovered that this morning when I ran the content filter test. I have fixed it in the commits I'm about to push.

tools/dissector/sample_manager.cpp
```diff
@@ -36,211 +39,218 @@ namespace OpenDDS
...
-          if (pos != limit) {
-            return new Sample_Array(*pos, do_array(pos + 1, limit, element_type));
-          }
-          else {
-            return get_dissector(map, element_type);
-          }
+      void visit (itl::Fixed&) {
+        dissector = new Sample_Dissector();
+        dissector->add_field(new Sample_Field(Sample_Field::Undefined, ""));
+        ACE_DEBUG ((LM_WARNING, ACE_TEXT ("Fixed-point types are not supported\n")));
```
mitza-oci: OpenDDS does support the `fixed` data type now (probably not when this code was first written) for some use cases.  If you have time it would be good to add it.

tools/dissector/sample_manager.cpp
```diff
@@ -251,6 +261,19 @@ namespace OpenDDS
...
     //--------------------------------------------------------------------
+
+    Sample_Manager::~Sample_Manager()
+    {
+      if (hf_array_ != NULL) {
+        delete [] hf_array_;
+      }
+      // Free Dynamically Generated Field Names
+      while (!field_names_.empty()) {
+        free(field_names_.front());
```
mitza-oci: I think these were allocated with ACE_OS::strdup, so this should be ACE_OS::free to match.

tools/dissector/sample_manager.cpp
```diff
@@ -266,21 +289,86 @@ namespace OpenDDS
...
+        Sample_Base::clear_ns();
+        std::string name(i->first);
+
+        // If name begins with "IDL:", remove it
+        const std::string idl_prefix("IDL:");
+        if (name.substr(0, idl_prefix.size()) == idl_prefix) {
+            name.erase(0, idl_prefix.size());
+        }
+
+        // If name contains ':', remove everything after it
```
mitza-oci: s/after it/after the last :/

tools/dissector/sample_dissector.h
```diff
@@ -182,31 +259,41 @@ namespace OpenDDS
...
       /// empty.
-      gint ett_payload_;
+      gint ett_;
       std::string subtree_label_;
 
-    private:
-      /// The list of fields defining the sample.
-      Sample_Field *field_;
-
+      bool is_struct_ = false;
```
mitza-oci: in-class initialization is not supported before C++11

tools/dissector/sample_dissector.cpp
```diff
@@ -31,13 +32,171 @@ namespace OpenDDS
...
+
+#if ACE_SIZEOF_WCHAR == 4
+      /*
+       * We have to trim the extra 16 bits of zeros between the characters
+       * before we can pass it to glib's g_convert. Otherwise it will
+       * interpret them as NULL characters and stop after the first two bytes.
+       * g_convert doesn't respsect the size passed to it if it thinks it can
+       * keep going.
+       */
+      uint16_t* trimmed = new uint16_t[length + 1];
```
mitza-oci: Could this be std::vector instead?  That would reduce complexity of dealing with dynamic allocation here.

tools/dissector/sample_dissector.cpp
```diff
@@ -31,13 +32,171 @@ namespace OpenDDS
...
     }
 
+    Sample_Base::~Sample_Base() {
+      for (
+        Field_Contexts::iterator i = field_contexts_.begin();
+        i != field_contexts_.end();
+        ++i
+      ) {
+        delete i->second;
+        i->second = NULL;
```
mitza-oci: Is there a need for this line?

tools/dissector/sample_dissector.cpp
```diff
@@ -91,261 +250,523 @@ namespace OpenDDS
...
+        p.serializer >> longdouble_value;
+        s << longdouble_value;
+        break;
+
+      case String:
+        ACE_CDR::Char * string_value;
+        string_value = NULL;
+        p.serializer.read_string(string_value);
+        s << string_value;
+        delete [] string_value;
```
mitza-oci: I don't think there's any guarantee that the string is allocated with new[].  This block can probably be simplified by using CORBA::String_var or TAO::String_Manager.  Same for WString below.

tools/dissector/sample_dissector.cpp
```diff
@@ -91,261 +250,523 @@ namespace OpenDDS
...
+            get_ns()  + " is not a registered wireshark field."
+          );
+        } else {
+
+          size_t location = params.buffer_pos();
+
+          // Set Field
+          switch (this->type_id_) {
+
+          case Sample_Field::Boolean:
```
mitza-oci: This code is in the class Sample_Field, so the Sample_Field:: scope shouldn't be needed here.

tools/dissector/sample_dissector.cpp
```diff
@@ -91,261 +250,523 @@ namespace OpenDDS
...
+          add_protocol_field(FT_STRINGZ);
+          break;
+
+        case Sample_Field::WString:
+          add_protocol_field(FT_STRINGZ);
+          break;
+
+        default:
+          throw Sample_Dissector_Error(
+            get_ns()  + " is not a valid Type Field."
```
mitza-oci: s/Type Field/Field Type/

tools/dissector/sample_dissector.cpp
```diff
@@ -388,88 +809,79 @@ namespace OpenDDS
...
+            );
+          } else {
+            proto_item* item = proto_tree_add_none_format(
+              ADD_FIELD_PARAMS,
+              outstream.str().c_str()
+            );
+            params.tree = proto_item_add_subtree(item, ett_);
+          }
+        } else {
+          if (is_root_) is_root_ = false;
```
mitza-oci: Simplify to `is_root_ = false;`

tools/dissector/sample_dissector.cpp
```diff
@@ -505,54 +938,87 @@ namespace OpenDDS
...
-      size_t len = 4;
-      guint32 count = *(reinterpret_cast< guint32 * >(data));
-      data += 4;
-      while (count > 0)
-        {
-          size_t elen = element_->compute_length(data);
-          data += elen;
-          len += elen;
-          --count;
+      // Add Item
```
mitza-oci: There should be a lot of common code between Sequence and Array.  After sequence has read the length from the buffer, can this delegate to a method shared by Sequence and Array (possibly in a base class)?

tools/dissector/sample_dissector.cpp
iguessthislldo: No, I guess I was just following the rule to NULL unused pointers, but it is a moot point here.

tools/dissector/packet-opendds.cpp
```diff
@@ -546,7 +546,7 @@ namespace OpenDDS
           tvb_,
           offset,
           (gint) header.message_length_,
-          "No Topic Found for %s \n",
+          "No Topic Found for %C \n",
```
mitza-oci: This was changed along with the ACE_DEBUG statements, but it's not an ACE_DEBUG (also on line 583)

iguessthislldo:
@mitza-oci I think I resolved all your comments, except for the Fixed type support, which I've added to my TODO list.

PR #781, opened 2017-12-19
configure: default compiler for macOS is clang

mitza-oci:


mitza-oci:
See Issue #779

mitza-oci:

mitza-oci:

PR #786, opened 2018-01-15
Fix uninitialized values in configure script

jrw972:


configure
```diff
@@ -1197,8 +1197,10 @@ EOT
         open(my $fh, '<', $file) or die "ERROR: Could not open $file: $!";
         while (my $row = <$fh>) {
           chomp $row;
-          unless ($row =~ /^include/) {
-            my ($key, $value) = split(/=/, $row);
+          next if ($row =~ /^\s*#/);
+          $row =~ s/\s*#.*$//;
+          if ($row =~ /^\s*(\w+)\s*[:?+]=\s*(.*)$/) {
```
okellogg: Oops, my bad: The characters [:?+] shall be optional, therefore the line should be
```
if ($row =~ /^\s*(\w+)\s*[:?+]?=\s*(.*)$/) {
```
Notice the added "?" after the character group.

mitza-oci:

mitza-oci:

okellogg:
> Can't set status; build failed.

 The AppVeyor build errors out with
```
      (ClCompile target) -> 
         c:\projects\opendds\tools\dissector\ws_config.h(7): fatal error C1083: Cannot open include file: 'config.h': No such file or directory [C:\projects\opendds\tools\dissector\dissector.vcxproj]
```
and the OpenDDS_Pull_Requests build shows test failures in
```
Autobuild_Tests.tests/DCPS/Deadline/run_test.pl rtps_disc
Autobuild_Tests.tests/DCPS/WaitForAck/run_test.pl rtps
```

Are you saying that these failures are caused by the change to `configure` ?
(I'm having a hard time seeing the correlation.)
Thanks.

mitza-oci:
"Can't set status; build failed." was an automated post from the Jenkins github PR plugin, which should be fixed now

Wireshark dissector is currently in flux, when the next build starts those errors should go away.

Deadline and WaitForAck failures are not due to a change in configure.

okellogg:
@mitza-oci Thanks for clarifying.
IIUC there is no point then in making a new pull request? Should be possible eventually to proceed with this one?

configure
```diff
@@ -1197,8 +1197,10 @@ EOT
         open(my $fh, '<', $file) or die "ERROR: Could not open $file: $!";
         while (my $row = <$fh>) {
           chomp $row;
-          unless ($row =~ /^include/) {
-            my ($key, $value) = split(/=/, $row);
+          next if ($row =~ /^\s*#/);
```
mitza-oci: This line appears to be unnecessary since the comments are already removed by the next line, I don't think we're gaining much by determining that it's a comment-only line.

okellogg: Agreed.

PR #787, opened 2018-01-15
Wireshark 2 Dissector: Fixes for Windows and Scoreboard

iguessthislldo:
- Changes to the configure script to work with [MPC Wireshark 2 Windows Support](https://github.com/DOCGroup/MPC/pull/47).

- Various Changes to meet OCI Jenkins Scoreboard errors (mostly accidental C++11 I wrote).

- Fixed bug I introduced that made random fields a struct because is_struct was uninitialized.

- Dissector with Sample Dissection prints every ITL file it finds and the types they contain.


jwillemsen:
See also http://buildlogs.remedy.nl/fc_opendds_versioned/

mitza-oci:
@jwillemsen are you using Fedora's wireshark-devel package?
If so, set the environment variable WIRESHARK_SRC to /usr/include/wireshark

jwillemsen:
Yes, using that, wouldn't it be easier to default WIRESHARK_SRC to that path on linux and only require the user to set it when it is installed on another location?

mitza-oci:
That is something we can look at as we finalize the Wireshark2 changes.  I expect most end-users will get the setting via the configure script.

iguessthislldo:
This could be helpful to be built in, but are there any other distros provide packages like this that might put them in other locations?

mitza-oci:
Ubuntu matches it: https://packages.ubuntu.com/xenial/amd64/libwireshark-dev/filelist
The configure script has a mechanism for "use_system_pkg" that can be adapted for this.  It will probably need a special case because the variable refers to the include directory itself (/usr/include/wireshark) instead of a *_ROOT that gets /include concatenated later.

tools/dissector/sample_dissector.h
```diff
@@ -51,7 +51,9 @@ namespace OpenDDS
       {
       }
 
-      const char* what() const throw (){
+      ~Sample_Dissector_Error() {}
+
+      const char* what() const throw() override {
```
mitza-oci: `override` is from C++11.  If you can get away with leaving it off, that's probably the way to go.  If not we'll need an #ifdef to conditionally add `override`.

jwillemsen:
Added WIRESHARK_SRC to our build scripts

configure
```diff
@@ -154,9 +154,12 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'java:s', 'Java development kit (use JAVA_HOME)',
     'jboss:s', 'JBoss application server (use JBOSS_HOME)',
     'ant:s', 'Ant for JBoss (use ANT_HOME or system pkg)',
-    'wireshark:s', 'Wireshark (use WIRESHARK_SRC)',
-    'glib:s', 'GLib for wireshark (use GLIB_ROOT or system pkg)',
-    'rapidjson', 'RapidJSON for wireshark itl (optional)',
+    'wireshark:s', 'Wireshark 1.x Source (use WIRESHARK_SRC)',
+	'wireshark2:s', 'Wireshark 2.x Source (use WIRESHARK_SRC)',
```
mitza-oci: Replace tabs with spaces on the 3 lines starting here

iguessthislldo: I thought I got all those but I think I was just greping `tools/dissector`.

mitza-oci:
@jwillemsen looks like some combination of GLIB_VERSION, GLIB_LIB_DIR, GLIB_ROOT is required for Fedora's glib package.

jwillemsen:
Updated our build to set GLIB_LIB_DIR to lib64

tools/dissector/sample_dissector.h
iguessthislldo: I thought one of the builds complained about it, but I looked through them yesterday and I couldn't find it, they only complained about the destructor. I think I mixed them up and tried to fixed something that didn't need fixing.

PR #788, opened 2018-01-15
Filter Eval: improved error message

mitza-oci:
and fixed an unrelated unused variable warning


mitza-oci:

PR #793, opened 2018-01-19
Dissector macOS support, configure script, fixed throw spec problem

iguessthislldo:
Dissector macOS support as well as changing the configure script to reflect [MPC#48](https://github.com/DOCGroup/MPC/pull/48) and the dissector/README to reflect both of these.

Also:
 - Some bits of the `Fixed` type support for the sample payload dissector are in here, but nowhere near done.
 - The loose throw error that I failed to fix in the scoreboards.
 - Default to `/usr/include/wireshark` if it exists and `--wireshark` was left blank to support wireshark development packages installed there.

### DIFF
--- a/tools/dissector/macos_postbuild.sh
+++ b/tools/dissector/macos_postbuild.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 # This script takes care of two problems with the dissector file on macOS:
 #
 #   $DYLD_LIBRARY_PATH is not persevered between processes after
--- a/tools/dissector/macos_postbuild.sh
+++ b/tools/dissector/macos_postbuild.sh
@@ -1,0 +1,25 @@
+# This script takes care of two problems with the dissector file on macOS:
+#
+#   $DYLD_LIBRARY_PATH is not persevered between processes after
+#   v10.11 "El Capitan" for security reasons and the rpath for the OpenDDS
+#   libraries is relative. To remedy this we add a absolute rpath to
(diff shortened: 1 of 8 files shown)


tools/dissector/macos_postbuild.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# This script takes care of two problems with the dissector file on macOS:
+#
+#   $DYLD_LIBRARY_PATH is not persevered between processes after
+#   v10.11 "El Capitan" for security reasons and the rpath for the OpenDDS
+#   libraries is relative. To remedy this we add a absolute rpath to
+#   $DDS_ROOT/lib.
+#
+#   It produces dylib files which Wireshark will except but only if they have
```
mitza-oci: "will accept"

tools/dissector/sample_manager.cpp
```diff
@@ -154,12 +154,13 @@ namespace OpenDDS
         }
       }
 
-      void visit (itl::Fixed&) {
-        // TODO Fixed types are supported in OpenDDS now, and the dissector
-        // should handle them.
+      void visit (itl::Fixed& fixed) {
+#ifndef ACE_HAS_CDR_FIXED
```
mitza-oci: The preprocessor logic looks backwards to me: If ACE_HAS_CDR_FIXED is not defined, attempt to use Fixed, otherwise warn.

PR #794, opened 2018-01-24
dissector: Fixed Type Support and expert changes

iguessthislldo:
Added support for dissection of OpenDDS fixed-point numbers which are
converted to doubles for numerical comparisons within Wireshark.
If ACE_CDR::Fixed is missing, a message will be displayed that this is so.

Also started marking some packets with a warning instead of marking it
as malformed. For example, when an individual field fails for some
reason or non essential information is unavailable, like the dissector
ITL file or Fixed type is missing like above.

### DIFF
--- a/tools/IntermediateTypeLang/cpp/itl/itl.hpp
+++ b/tools/IntermediateTypeLang/cpp/itl/itl.hpp
@@ -606,13 +606,15 @@ namespace itl {
       std::map<std::string, std::string> values;
       bool isConstrained = false;
 
-      if (!extractUint("base", value, base)) throw std::runtime_error("No base for fixed");
+      // if (!extractUint("base", value, base)) throw std::runtime_error("No base for fixed");
       if (!extractUint("digits", value, digits)) throw std::runtime_error("No digits for fixed");
       if (!extractUint("scale", value, scale)) throw std::runtime_error("No scale for fixed");
 
+      /*
       if (base == 0) {
         throw std::runtime_error("Illegal base for fixed");
       }
+      */
 
       if (scale > digits) {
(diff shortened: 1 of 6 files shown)


tools/IntermediateTypeLang/cpp/itl/itl.hpp
```diff
@@ -606,13 +606,15 @@ namespace itl {
       std::map<std::string, std::string> values;
       bool isConstrained = false;
 
-      if (!extractUint("base", value, base)) throw std::runtime_error("No base for fixed");
+      // if (!extractUint("base", value, base)) throw std::runtime_error("No base for fixed");
```
mitza-oci: It's expecting a 'base' so we could just generate base 10 for all of ours.

tools/dissector/sample_dissector.cpp
```diff
@@ -1256,6 +1286,82 @@ namespace OpenDDS
...
+        if (hf == -1) {
+          throw Sample_Dissector_Error(
+            get_ns()  + " is not a registered wireshark field."
+          );
+        }
+
+#ifdef ACE_HAS_CDR_FIXED
+        // Extract a ACE_CDR::Fixed, give it to WS as a double, and display
+        // it using Fixed.to_string().
+        FACE::Octet raw[len];
```
mitza-oci: The bounds of a C Array need to be known at compile time.  You can use DDS::OctetSeq or std::vector as runtime-sized continuous byte containers.

iguessthislldo: Yea, I copied it from the Fixed_T template and even changed the name without thinking. I'm going to set it to the max size of 32.

Issue #808, opened 2018-02-12
IDL Union conversion corner case


pcoscia:
Hi:

I have come across an issue converting IDL Union's to C++ using the OpenDDS IDL Generator.
The union is written to optionally define a variable if the given boolean is true, and do nothing otherwise. The IDL from the codebase I am using has been simplified below.

`
union Type switch (boolean)
{
        case TRUE : My_Type value;
};
 `

When the code is converted to C++, the result is the expected 'if' statement for the True case but an unfinished 'else' (without semicolon or braces) where the generator expected a False case.

`if (uni._d()) 
    {
      if ((size + padding) % 8) {
        padding += 8 - ((size + padding) % 8);
      }
      size += gen_max_marshaled_size(uni.value());
    }
  else                 //compile error due to this 'else'
}`

I have tracked down what I believe is the cause of the issue I am having in the file $(DDS_ROOT_DIRECTORY)/dds/idl/dds_generator.h and the function generateSwitchForUnion(...) on line 589 in OpenDDS Version 3.12.

In this function, the 'if' and 'else' are written before checking for true and false, which means that when the IDL does not specify a False case, the else is written but no statement is written.

The issue I am having would be fixed if the idl_generator code did not assume that the True and False statements would be given in every case.

I hope that is enough info but happy to clarify anything that doesn't make sense.

mitza-oci:
Yes, I think that's clear.  I know we have special cases for boolean-discriminated unions because newer compilers were issuing warnings with switch(bool).  Please add a test case somewhere in tests/DCPS/Compiler.

pcoscia:
@mitza-oci I have created a fork and added a project that should make the issue I am having clear.
Link: https://github.com/pcoscia/OpenDDS/tree/Issue-%23808-IDL-Union-conversion
If there are any issues, please let me know.

mitza-oci:
The branch seems to be missing odds_pubsub.mpb (or we can replace the reference to it in idl_union_conversion.mpc with ___?).

pcoscia:
Apologies, oddspubsub.mpb that is a helper specific to the project I am working on. It looks like you have solved the issue anyway.
Thank you for your assistance, your changes have solved the problem I was having.

pcoscia:
@mitza-oci A colleague and I have been setting up ODDS with the latest release (3.12.2) and coming up with an issue that the changes made to fix this bug have not been merged into the releases. 
[...]

PR #811, opened 2018-02-14
IDL compiler: support all valid bool-discriminated unions

mitza-oci:
Test cases for issue #808


mitza-oci:
After seeing that the tests failed to compile as expected, this PR now has the fix for the issue.

PR #815, opened 2018-02-22
Updated DOC Group TAO version

jwillemsen:


mitza-oci:
This is listing the version we tested with at the time 3.12 was released.

Issue #789, opened 2018-01-17
Intra-Process Communication with FACE


KyleStevenson3:
I’m trying to use OpenDDS’s FACE TS implementation to communicate across threads and I’m getting an error on the console output and I don't get the Hello World. 

If OpenDDS is supposed to support intra process communication here are the details explaining what happened and my setup.

> creating OpenddsFaceSubscriber
> Subscriber: about to Receive_Message()
> creating OpenddsFacePublisher
> Publisher: about to Send_Message()
> (11000|10900) WARNING: DataReaderImpl::get_handle_instance: lookup for 0x4 failed
> (11000|10900) populate_header_received: failed to read BIT publication data.

The program I'm testing with is basically the DevGuideExamples/FACE/Simple, but on separate threads instead of separate processes as it might be used inside a FACE UoP. The publisher and subscriber were moved into classes with FACE::TS::Initialize() calls removed.

I’m using the face_config.ini as it is in the Simple example. Building on Windows 7 VS2013.

Main looks like this:

```
#include "ace/config.h"
#include "openddsfacesubscriber.h"
#include "openddsfacepublisher.h"
#include <thread>
#include <chrono>

void startSubThread();

int main(int argc, char *argv[])
{
    OpenDDS::DCPS::set_DCPS_debug_level(4);

    FACE::RETURN_CODE_TYPE status;
    FACE::TS::Initialize("face_config.ini", status);

    if (status != FACE::RC_NO_ERROR) {
      return static_cast<int>(status);
    }

    std::thread subscriber_thread( startSubThread );
    subscriber_thread.detach();
    std::this_thread::sleep_for(std::chrono::seconds(5));

    OpenddsFacePublisher publisher;
    publisher.run();
    return EXIT_SUCCESS;
}

void startSubThread()
{
    OpenddsFaceSubscriber subscriber;
    subscriber.run();
}

int OpenddsFaceSubscriber::run()
{
  // Initialize the TS interface
  FACE::RETURN_CODE_TYPE status;
//  FACE::TS::Initialize("face_config.ini", status);

//  if (status != FACE::RC_NO_ERROR) {
//    return static_cast<int>(status);
//  }

  // Create the sub connection
  FACE::CONNECTION_ID_TYPE connId;
  FACE::CONNECTION_DIRECTION_TYPE dir;
  FACE::MESSAGE_SIZE_TYPE max_msg_size;
  FACE::TS::Create_Connection(
[...]

PR #819, opened 2018-03-11
Fix for in-process messaging in FACE TSS

kpansky:
First pass at a solution to #789 

Not sure if the approach taken is what you were describing in the issue discussion.

Added a test case to the Messenger tests


mitza-oci:
ok to test on Jenkins

mitza-oci:
Since tests/FACE/SingleProcessMessenger/run_test.pl is a new script, please add it to the list in bin/dcps_tests.lst.

Issue #803, opened 2018-02-02
Including typedef in IDL


kaoao:
When I try to include a typedef from some base IDL, the opendds_idl compiler does not generate the include statement for the *C.h file.

Consider the following 2 IDL files:

included_file.idl:
```
typedef unsigned long UniqueId;
```
and main.idl:

```
#include "included_file.idl"

struct MainMessage {
	UniqueId myId;
};
```

Code generation works when I wrap `UniqueId` in a `struct` but not when the `typedef` is standalone. Also, I tried using the `import` keyword but to no avail.

The commands I use to generate the code are below:
```
opendds_idl -Sa -St -GfaceTS -Lface -Wb,export_include=main_Export.h -Wb,export_macro=main_Export -o gen main.idl
opendds_idl -Sa -St -GfaceTS -Lface -Wb,export_include=included_file_Export.h -Wb,export_macro=included_file -o gen included_file.idl
```

Any help is appreciated.

mitza-oci:
Please try this version https://github.com/objectcomputing/OpenDDS/pull/804 and let us know if it fixes this issue.

kaoao:
Works now. Thanks

mitza-oci:
That's good, thanks for trying it.

In order to facilitate merging this to master, would you please create a test case for this?  It can be a compile-only test in the tests/FACE/Compiler directory.

mitza-oci:
Fixed in #804, this issue remains open until we get test coverage

PR #820, opened 2018-03-11
Test for #803

kpansky:
Test case for #803 

Succeeds with #804 in, fails with it out.


mitza-oci:
Thanks for your contribution.  Please review the errors on the Travis CI build (you can ignore the ones where it times out at 50 min, we'll attempt to restart those).

PR #826, opened 2018-03-21
Regenerated and disabled writer generation

jwillemsen:
* dds/DCPS/QOS_XML_Handler/dds_qos.cpp:
    * dds/DCPS/QOS_XML_Handler/dds_qos.hpp:
    * dds/DCPS/QOS_XML_Handler/qos_xml_handler.mpc:


jwillemsen:
The writer code could be used to take an IDL presentations and write it as XML on disk, but that would require a lot more additional conversion code between IDL->XML. Because this is all not used we don't generate the writer, saves 11k SLOC

PR #828, opened 2018-03-26
Built In Topic Test Fixes

iguessthislldo:
This has fixes for `tests/DCPS/BuiltInTopicTest/run_test.pl` and I tried to but couldn't fix anything with `tests/DCPS/BuiltInTopicTest/prst_repo_run_test.pl`. There are a few questions that I have put in the source.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -2248,8 +2248,9 @@ DataReaderImpl::instances_liveliness_update(WriterInfo& info,
     const ACE_Time_Value& when)
 {
   ACE_GUARD(ACE_Recursive_Thread_Mutex, instance_guard, this->instances_lock_);
-  for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
-       iter != instances_.end(); ++iter) {
+  for (SubscriptionInstanceMapType::iterator iter = instances_.begin(),
+      next = iter; iter != instances_.end(); iter = next) {
+    ++next;
     iter->second->instance_state_.writer_became_dead(
         info.writer_id_, liveliness_changed_status_.alive_count, when);
   }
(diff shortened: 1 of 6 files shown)


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -2248,9 +2248,8 @@ DataReaderImpl::instances_liveliness_update(WriterInfo& info,
     const ACE_Time_Value& when)
 {
   ACE_GUARD(ACE_Recursive_Thread_Mutex, instance_guard, this->instances_lock_);
-  for (SubscriptionInstanceMapType::iterator iter = instances_.begin(),
-      next = iter; iter != instances_.end(); iter = next) {
-    ++next;
+  for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+       iter != instances_.end(); ++iter) {
```
iguessthislldo: I made this change but I should've have asked why it has the `next` iterator, because it looks redundant to me.

dds/DCPS/transport/framework/DataLink.inl
```diff
@@ -107,11 +107,14 @@ DataLink::send(TransportQueueElement* element)
 
   element = this->customize_queue_element(element);
   if (!element) {
+    element->data_dropped(true);
```
iguessthislldo: This series of `element->data_dropped(true)` seems to have fixed the memory leak I was looking at. I just wanted to confirm that this is correct and doesn't mess up the element in other places.

dds/DCPS/DataReaderImpl.cpp
mitza-oci: Looks like it would be used if the call to writer_become_dead invalidated the iterator (even indirectly).

dds/DCPS/transport/framework/DataLink.inl
mitza-oci: If element is null (the `if` on 109 succeeds), how can this line invoke a method on it?

iguessthislldo: ...I guess it wouldn't be able to. I think the idea I had in mind should have been written like:
```
TransportQueueElement * test_element = this->customize_queue_element(element);
if (test_element)  {
  element = test_element;
} else {
  element->data_dropped(true);
  return;
}
```

PR #831, opened 2018-03-30
Fix Reconnect Test and DataLink Leaks

iguessthislldo:
- Re-added the `element->data_dropped()`s in `DataLink`, except for the one after `customize_element()`. `customize_element()` calls `data_dropped` or passes it to `TransportCustomizedElement` which drops it when it's no longer needed.

- The Reconnect test hides the program output. Huangming had changed it in 373f41c
from collecting individual program output in a single file to having one
for each program, except the InfoRepo. I added an InfoRepo log file and did
some cleanup/small improvements.

- For the Reconnect sub_init_crash test the publisher has always (or at least in the commit
before the WeakRcHandle PR) waited for subscribers until the test script
timedout, killed it and reported it as an error. The fact that no more
subscribers are spawned seems to be intentional, as it just tries to see
if the publisher will crash but it was also hidden before the PR.
Now it is killed and it shouldn't report an error unless
the publisher returns a non zero status.

### DIFF
--- a/tests/DCPS/Reconnect/.gitignore
+++ b/tests/DCPS/Reconnect/.gitignore
@@ -24,3 +24,4 @@
 /subscriber1.log
 /subscriber2.log
 /repo.log
+/core*
--- a/tests/DCPS/Reconnect/.gitignore
+++ b/tests/DCPS/Reconnect/.gitignore
@@ -19,9 +19,8 @@
 /subscriber_ready.txt
 /publisher_finished.txt
 /publisher_ready.txt
-/test.log
 /publisher1.log
 /publisher2.log
 /subscriber1.log
 /subscriber2.log
-
+/repo.log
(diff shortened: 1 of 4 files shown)


tests/DCPS/Reconnect/.gitignore
```diff
@@ -19,9 +19,9 @@
 /subscriber_ready.txt
 /publisher_finished.txt
 /publisher_ready.txt
-/test.log
 /publisher1.log
 /publisher2.log
 /subscriber1.log
 /subscriber2.log
-
+/repo.log
```
mitza-oci: Not a big deal, but this should be covered by the wildcard ignore in the root .gitignore.  You can use the script tools/scripts/update_gitignore.pl to maintain these.  core could be covered by your individual git config.

PR #836, opened 2018-04-04
Dissector: Various Improvements 

iguessthislldo:
- Gave dissector proper return value for WS 2.2 and later.
- Some formatting fixes and more comments
- Fixed a malformed packet error where multiple samples caused a bounds check error inside Wireshark's ethernet dissector.
- Updated README to mention 2.6 where applicable
- Renamed for loop variable that clashed with it's parent for loop's variable (using `i` is very enticing).
- Updated Field Registration to allow for "canceling" of fields to be registered with Wireshark.

### DIFF
--- a/tools/dissector/packet-opendds.cpp
+++ b/tools/dissector/packet-opendds.cpp
@@ -567,7 +567,7 @@ namespace OpenDDS
           payload_item, ett_sample_payload);
 
       // Try to Dissect Payload
-      Sample_Dissector_rch data_dissector =
+      Sample_Dissector* data_dissector =
         Sample_Manager::instance().find(data_name);
       if (!data_dissector) {
         ACE_DEBUG ((LM_DEBUG,
@@ -653,7 +653,7 @@ namespace OpenDDS
 #ifndef NO_EXPERT
           proto_tree_add_expert_format(
             contents_tree, pinfo_, &ei_sample_payload_error,
-            tvb_, offset, -1, e.what()
+            tvb_, offset, -1, "%s", e.what()
           );
 #endif
           ACE_DEBUG ((LM_DEBUG,
--- a/tools/dissector/packet-opendds.cpp
+++ b/tools/dissector/packet-opendds.cpp
@@ -567,9 +567,9 @@ namespace OpenDDS
           payload_item, ett_sample_payload);
 
       // Try to Dissect Payload
-      Sample_Dissector *data_dissector =
-        Sample_Manager::instance().find (data_name);
-      if (data_dissector == 0) {
+      Sample_Dissector_rch data_dissector =
+        Sample_Manager::instance().find(data_name);
+      if (!data_dissector) {
         ACE_DEBUG ((LM_DEBUG,
                     "DDS_Dissector::dissect_sample_payload: "
                     "couldn't dissect payload: "
--- a/tools/dissector/packet-opendds.cpp
+++ b/tools/dissector/packet-opendds.cpp
@@ -536,9 +536,8 @@ namespace OpenDDS
       if (data_name == 0) {
         ACE_DEBUG ((LM_DEBUG,
                     "DDS_Dissector::dissect_sample_payload: "
-                    "no topic for %C\n",
+                    "couldn't dissect payload: no topic for %C\n",
                     std::string(converter).c_str()));
-        offset += header.message_length_; // skip marshaled data
 
         // Mark Packet
 #ifndef NO_EXPERT
@@ -549,10 +548,12 @@ namespace OpenDDS
           tvb_,
           offset,
           (gint) header.message_length_,
-          "No Topic Found for %s \n",
+          "Couldn't Dissect Payload: No Topic Found for %s \n",
           std::string(converter).c_str()
         );
 #endif
+
+        offset += header.message_length_; // skip marshaled data
 
         return;
       }
@@ -571,6 +572,7 @@ namespace OpenDDS
       if (data_dissector == 0) {
         ACE_DEBUG ((LM_DEBUG,
                     "DDS_Dissector::dissect_sample_payload: "
+                    "couldn't dissect payload: "
                     "no dissector found for %C \n",
                     data_name));
 
@@ -583,6 +585,7 @@ namespace OpenDDS
           tvb_,
           offset,
           (gint) header.message_length_,
+          "Couldn't Dissect Payload: "
           "No Dissector Found for %s \n",
           data_name
         );
@@ -592,6 +595,7 @@ namespace OpenDDS
       } else {
         // Push DCPS Primary Data Type Namespace
         // Convert, for example, "1::2::3" to "1.2.3"
+        Sample_Base::clear_ns();
         std::string ns;
         char c = 'a';
         bool delimiter = false;
@@ -643,7 +647,20 @@ namespace OpenDDS
         ACE_Sig_Action sig_act(sample_dissect_handle_sigsegv);
         if (handle) sig_act.register_action(SIGSEGV); // Handle Segfaults
 
-        offset = data_dissector->dissect(params); // Dissect Sample Payload
+        try {
+          offset = data_dissector->dissect(params); // Dissect Sample Payload
+        } catch (Sample_Dissector_Error& e) {
+#ifndef NO_EXPERT
+          proto_tree_add_expert_format(
+            contents_tree, pinfo_, &ei_sample_payload_error,
+            tvb_, offset, -1, e.what()
+          );
+#endif
+          ACE_DEBUG ((LM_DEBUG,
+            "DDS_Dissector::dissect_sample_payload: %C\n", e.what()
+          ));
+          offset = params.offset + header.message_length_;
+        }
 
         // Clean Up
         if (handle) sig_act.handler(SIG_DFL); // Restore default segfault behavior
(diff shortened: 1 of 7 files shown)


iguessthislldo:
I think these are all the small Improvements I have at the moment.

The following is just as much for me as anyone else:

The thing that I have been trying to work through is the ability to recover if there are problems with the ITL files (Namespace clashes mostly) and still be able to dissect using ITL that are not bad or linked to the bad ones. The dissection data structures would have to be deleted though, but that's not immediately possible for the most part because the they refer to each other (why I asked about the WeakRcHandle before). 

I also have to "cancel" the field registration metadata that's going to be passed Wireshark. I have that partially figured that out by having the Dissector/Field objects take ownership of their Wireshark field data. Then instead of the objects immediately putting it in the global `Sample_Manager` with `add_protocol_field`, they are collected from the objects with the bad dissector objects already been removed.

tools/dissector/sample_dissector.cpp
```diff
@@ -763,10 +764,12 @@ namespace OpenDDS
           add_protocol_field(FT_STRINGZ);
           break;
 
+        /*
```
mitza-oci: Why is this part commented-out?

tools/dissector/packet-opendds.cpp
```diff
@@ -643,7 +647,20 @@ namespace OpenDDS
         ACE_Sig_Action sig_act(sample_dissect_handle_sigsegv);
         if (handle) sig_act.register_action(SIGSEGV); // Handle Segfaults
 
-        offset = data_dissector->dissect(params); // Dissect Sample Payload
+        try {
+          offset = data_dissector->dissect(params); // Dissect Sample Payload
+        } catch (Sample_Dissector_Error& e) {
```
jwillemsen: Catch as const&

tools/dissector/sample_dissector.cpp
```diff
@@ -196,11 +195,67 @@ namespace OpenDDS
...
-    void Sample_Base::add_protocol_field(enum ftenum ft, field_display_e fd) {
-      Field_Context * fc = get_context();
-      Sample_Manager::instance().add_protocol_field(
-        &fc->hf_, get_ns(), fc->label_, ft, fd
-      );
+    void Sample_Base::push_idl_name(std::string name) {
+      Sample_Base::clear_ns();
+
+      // If name begins with "IDL:", remove it
+      const std::string idl_prefix("IDL:");
```
jwillemsen: Why not use find?

tools/dissector/sample_dissector.cpp
```diff
@@ -196,11 +195,67 @@ namespace OpenDDS
...
+          } else {
+              l++;
+          }
+      }
+    }
+
+    void Sample_Base::add_protocol_field() {
+      Field_Context* fc = get_context();
+      if (fc) {
+        printf("Field: %s\n", get_ns().c_str());
```
jwillemsen: ACE logging?

tools/dissector/sample_dissector.h
```diff
@@ -63,39 +66,64 @@ namespace OpenDDS
...
+        char* data, size_t size, bool swap_bytes, Serializer::Alignment align
       );
       Wireshark_Bundle(const Wireshark_Bundle & other);
 
       size_t buffer_pos();
       guint8* get_remainder();
     };
 
-    /// Holds Wireshark field index in relationship to a namespace
+    /// Holds Wireshark field infomation in relationship to a namespace
```
jwillemsen: Typo

tools/dissector/sample_manager.cpp
```diff
@@ -328,51 +333,49 @@ namespace OpenDDS
...
+          Sample_Base::clear_ns();
         }
 
-        // Create WS Fields off this namespace location
-        Sample_Dissector & dissector = *i->second;
-        dissector.init_ws_proto_tree();
-        Sample_Base::clear_ns();
+      } catch (const Sample_Dissector_Error & e) {
+        ACE_DEBUG((LM_ERROR,
+          ACE_TEXT("Error during dissector initization of %s: %s\n"),
```
jwillemsen: %C is for char

tools/dissector/sample_dissector.cpp
iguessthislldo: I didn't write this recently, just moved it to it's own function, but I think it's because it's just looking for "IDL:" at the beginning of the string only.

tools/dissector/sample_dissector.cpp
iguessthislldo: This was a quick debugging print I meant to take out.

tools/dissector/sample_manager.cpp
iguessthislldo: Adam has told that before, but it's a hard habit to break.

tools/dissector/sample_dissector.cpp
```diff
@@ -136,19 +136,17 @@ namespace OpenDDS
         i != field_contexts_.end();
         ++i
       ) {
-        delete i->second;
```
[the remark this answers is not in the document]
iguessthislldo: I didn't know delete and free ignored null values, so this means I have some personal code that is doing the same thing that I can simplify.

iguessthislldo:
I have decided to give up on trying to make the parts of the dissector tree deletable. @mitza-oci assuming the code I just pushed is alright, I think this is ready to be merged.

tools/dissector/sample_dissector.cpp
mitza-oci: I agree that it would be cleaner to use find() (comparing result to 0).  Might as well clean it up now that we're looking at it.

tools/dissector/sample_dissector.cpp
```diff
@@ -196,37 +193,94 @@ namespace OpenDDS
...
+      if (l != std::string::npos) {
+          name.erase(l, name.size() - l);
+      }
+
+      // Push namespace when we find a '/'
+      name.push_back('/'); // For last namespace
+      l = 0;
+      for (size_t index = 0; index != name.size(); index++) {
+          if (name[index] == '/') {
+              Sample_Base::push_ns(name.substr(index - l, l));
```
mitza-oci: This code is already within the Sample_Base scope, so mentioning that scope here shouldn't be necessary.

tools/dissector/sample_dissector.cpp
```diff
@@ -196,37 +193,94 @@ namespace OpenDDS
...
+          std::string("Could not create context becuase it already exists for: ") +
+          get_ns()
+        );
+      }
+
+      Field_Context* fc = new Field_Context;
+      fc->long_name_ = get_ns();
+      fc->short_name_ = get_label();
+      fc->hf_ = -1;
+      fc->hf_info_ = { &fc->hf_, {
```
mitza-oci: The syntax of `x = {a, b};` is only in C++11 (unless I'm missing something here?), we should use something that's backwards compatible OR document the requirements on compilers and flags (for example, GCC 5.x and earlier requires -std=c++11).

tools/dissector/sample_dissector.cpp
```diff
@@ -196,37 +193,94 @@ namespace OpenDDS
...
+      // Note: Wireshark calls this second struct member "abbrev" (What we
```
mitza-oci: See if this comment can be reworded to make it more clear ("Wireshark calls" and "is the string" is a bit clunky).
You can remove the line "This was confusing to me...".

tools/dissector/sample_dissector.cpp
```diff
@@ -196,37 +193,94 @@ namespace OpenDDS
...
+        nested_(n),
         next_(0)
     {
     }
 
-    Sample_Field::~Sample_Field ()
+    Sample_Field::~Sample_Field()
     {
-      delete nested_;
+      /* This doesn't work since a Dissector can be referenced by multiple
```
mitza-oci: Does that mean this has a leak?  Or is there some other mechanism already managing it?

tools/dissector/sample_dissector.cpp
iguessthislldo: [cppreference.com says that this works for pre C++11 if the class/struct has no constructor](https://en.cppreference.com/w/cpp/language/aggregate_initialization) and I just tested it and the syntax works with `g++ -std=c++03`.

mitza-oci: That's for initialization, not for assignment.  GCC should at least be warning for this.

iguessthislldo: You are right, I was thinking it would work for everywhere if it worked for initialization. Would it be alright if I put it inside of constructor member initialization list? I could also create a default value variable and copy it.

mitza-oci: Yes, Field_Context can have whatever constructor(s) you find useful.

tools/dissector/sample_dissector.cpp
iguessthislldo: The Wireshark dissector API doesn't have a cleanup callback for dissectors, at least one that I've seen, so this might have never been called until I started messing with it. The dissector tree lives as long as Wireshark is alive and is not wasted space because it can be used multiple times, so I'm not quite sure if that qualifies as a leak or not...

tools/dissector/sample_dissector.cpp
iguessthislldo: So GCC gives the same warning for struct initializer list in a member initializer list as it does for trying to assign it (the Extended Initializer Warning). It's not really a problem, but it wasn't what I was expecting would happen since I would have thought using member initializer lists would be equivalent to assigning a new variable.

PR #843, opened 2018-04-12
WIP: Update Article-Intro.html with info about where the "StockQuoterTypeSupportC.*" files are generated.

krhohio:
Please add an entry telling where the #include "StockQuoterTypeSupportC.h" file shown on line 1121 in this file can be found.

### DIFF
--- a/about/articles/Article-Intro.html
+++ b/about/articles/Article-Intro.html
@@ -1117,6 +1117,9 @@ method, which is called when a new Quote data sample is available. The
 other six methods have empty implementations. We'll also use a default
 constructor and destructor.
 </p>
+<p>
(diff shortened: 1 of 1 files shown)


krhohio:
Just found that this file is included in a couple of .gitignore files....
so I guess this Merge Request should say "...how file can be generated." instead!

krhohio:
Here's some hits on "StockQuoter" but didn't notice any StockQuoterTypeSupportC.h related:<br>
https://github.com/objectcomputing/OpenDDS/search?p=4&q=StockQuoter&type=&utf8=%E2%9C%93

mitza-oci:
I'm going to mark this closed since there's no proposed change yet.  If you'd like to propose a change we'll review it and merge.

Issue #849, opened 2018-04-15
ERROR: Can't extract ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip...


krhohio:
After cloning and switching to the DDS-3.12.2 tag I ran configure from the VS2017 Command Prompt and received the following output:
```
C:\...\3.12.x-OpenDDS>configure
Downloading ACE+TAO 2.2a with latest patches
Extracting archive ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip
ERROR: Can't extract ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip, extract it to C:/.../Working/3.12.x-OpenDDS
and run this script again.
Stopped at configure line 615.
```
I used 7zip to successfully extract the Zip contents but I'm wondering if  perhaps the non-existence of the (literal) Windows Zip program caused the above ERROR?  But then I think that Windows 10 has a build-in Zip capability?  Maybe I don't have *.zip files configured to open with my 7zip program?

krhohio:
In addition I received the following output after extracting the Zip using 7zip:
(i.e. same error as above even though the files have been extracted)

` Directory of C:\Users\Kent\Documents\Working\3.12.x-OpenDDS

04/14/2018  10:37 PM        58,647,557 ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip
               1 File(s)     58,647,557 bytes
               0 Dir(s)  177,205,264,384 bytes free

C:\Users\Kent\Documents\Working\3.12.x-OpenDDS>
C:\Users\Kent\Documents\Working\3.12.x-OpenDDS>configure
Extracting archive ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip
ERROR: Can't extract ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip, extract it to C:/Users/Kent/Documents/Working/3.12.x-OpenDDS
and run this script again.
Stopped at configure line 615.`

mitza-oci:
The configure script is working as designed.  If you'd like to contribute changes to it, of course we're open to that.

Issue #783, opened 2018-01-06
Configure Script can't locate Extracted ACE Dir on Windows 10


waheedejaz:
OpenDDS VERSION: 3.12
TAO VERSION: 2.2a
Windows 10

The configure script can't extract the downloaded .zip file of ACE+TAO.

```
Extracting archive ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip
ERROR: Can't extract ACE+TAO-2.2a_with_latest_patches_NO_makefiles.zip, extract it to E:/OpenDDS
and run this script again.
Stopped at configure line 615.
```
I manually extracted it, placed it in the same folder and re-run but still it gives same message.

waheedejaz:
One workaround to this is to manually define ACE_ROOT and TAO_ROOT in the environment variables and run configure script with --ace=%ACE_ROOT% --tao=%TAO_ROOT%

In my case: 
```
ACE_ROOT = OpenDDS/ACE+TAO-2.2a_with_latest_patches_NO_makefiles/ace_wrappers/
TAO_ROOT=%ACE_ROOT%/tao/
```
then I ran:
`configure --ace=%ACE_ROOT% --tao=%TAO_ROOT%`

mitza-oci:
You can install the Perl module Archive::Extract to give it the ability to unzip from within the Perl script.

If you unzip it yourself make sure that ACE_wrappers is the directory directly under the OpenDDS root.  This is what the configure script message that you saw before ("extract it to...") indicates.

PR #850, opened 2018-04-16
configure: Added alt unzip method: Archive::Zip

iguessthislldo:
In response to #849 and #783 I looked up if ActiveState Perl came with any zip modules on Windows, and it does, `Archive::Zip`, at least with 5.24.2.203. I added it as an alternative if `Archive:Extract` isn't installed, and `configure` will fall back to the manual extraction error message if `Archive::Zip` isn't installed or if it fails.

### DIFF
--- a/configure
+++ b/configure
@@ -613,8 +613,7 @@ if (!$ace_src || !$tao_src) {
       else {
         $err = system(($^O eq 'solaris' ? 'g' : '') . "tar xzf $ddsroot/$file");
       }
-    }
-    if ($err) {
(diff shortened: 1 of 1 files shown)


iguessthislldo:
I guess it would also try to use Archive::Zip if tar fails...

iguessthislldo:
Alright, now that I fixed my Perl `use` vs `require` misconceptions and it didn't break the tests this time, I think this should work automatically on Windows, assuming ActiveState Perl is installed and they are consistent with what modules they have installed by default.

PR #853, opened 2018-04-16
Use RcEventHandler in TransportReceiveStrategy subclasses

iguessthislldo:
This is an attempt to correct a Access Violation on Windows when the ACE Reactor Implementation `WFMO_Reactor` for the Multicast Transport tries tries to access the Transport's `MulticastReceivedStrategy` as part of it's cleanup. The problem is that it has already been deleted earlier in the destruction of the transport. This enables `ACE_Event_Handler`'s reference counting policy on Windows and connects it to `MulticastReceiverStrategy`'s `RcObject` count, keeping the `MulticastReceivedStrategy` alive until `WFMO_Reactor` is done with it and calls `remove_reference()` in `make_changes_in_current_info()`.

This is based off `RcEventHandler`, which might be preferred to this.

### DIFF
--- a/dds/DCPS/transport/multicast/MulticastReceiveStrategy.cpp
+++ b/dds/DCPS/transport/multicast/MulticastReceiveStrategy.cpp
@@ -18,9 +18,6 @@ namespace DCPS {
 MulticastReceiveStrategy::MulticastReceiveStrategy(MulticastDataLink* link)
   : link_(link)
 {
-#if WIN32
-	this->reference_counting_policy().value(ACE_Event_Handler::Reference_Counting_Policy::ENABLED);
-#endif
 }
 
 ACE_HANDLE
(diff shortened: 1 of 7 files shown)


iguessthislldo:
I will have to fix the indentation. I have my vim set up to look up indentation but I don't know how to do that with VS...

dds/DCPS/transport/multicast/MulticastReceiveStrategy.cpp
```diff
@@ -18,6 +18,9 @@ namespace DCPS {
 MulticastReceiveStrategy::MulticastReceiveStrategy(MulticastDataLink* link)
   : link_(link)
 {
+#if WIN32
```
mitza-oci: Could this be made unconditional?  ACE Reactor reference counting seems like a good idea in general.

If it remains conditional, see the preprocessor conditions in MulticastInst.cpp:129

iguessthislldo: I don't see why not, I did it because it's a Windows problem. Would you think it would be a good idea to apply it to other Transport Strategies as well, since it seems at least some of them are also `ACE_Event_Handler`s?

mitza-oci: Yes, I think that makes sense.  It would apply to all the network-based transports (not shmem).

jwillemsen: Sounds also to me a good thing to always use reference counting

PR #858, opened 2018-04-24
Use xercesc with wchar enabled

jwillemsen:


jwillemsen:
Closing, doesn't work yet

PR #863, opened 2018-05-07
BIT Test

iguessthislldo:
- Bit Test:
    - ~Rewrote `BuiltInTopicTest/prst_repo_run_test.pl` to make it more like the reconnect test. It will dump the logs of the apps only if one of them return a non-zero status.~ It will print to `stdout`/`stderr` again since hiding the logs is no longer necessary.
    - Replaced C++ streams with ACE Logging for control using `-ORBLogFile`.
    - Moved BIT read instance to end of subscriber run as a short term solution to how Info Repo handles BITs now.
    - Adjusted timing and message count of `prst_repo_run_test.pl`.
- Added functions for `DDS::InstanceStateKind/Mask` and `TcpConnection::reconnect_state_` to convert them into strings.
- Added a few new debug logging points along the way.
- For InfoRepo persistence:
    - Persist the last participant id explicitly instead of implicitly through whatever the greatest participant id happens to be.
    - Do not Autodispose BIT Readers Data
    - BIT Entities are now created first if restoring from persistence, giving them the same GUIDs as before.
- Bit Publisher Participant on the Info Repo uses the RTPS GUID Builtin Entity Kinds for Built-in-Topic Writers, Readers, and Built-In-Topics themselves.


dds/DCPS/InstanceState.h
```diff
@@ -131,6 +131,12 @@ class OpenDDS_Dcps_Export InstanceState : public ACE_Event_Handler {
 
   DDS::InstanceHandle_t instance_handle() const { return handle_; }
 
+  /// Return string of the name of the current instance state
+  std::string instance_state_string() { return instance_state_string(instance_state_); }
```
mitza-oci: This can be a const member function.
See PoolAllocator.h for a safety-profile-compatible way of using std::string.

dds/DCPS/InstanceState.cpp
```diff
@@ -5,6 +5,8 @@
  * See: http://www.opendds.org/license.html
  */
 
+#include <sstream>
```
mitza-oci: Safety profile excludes the C++ standard library streams.  See SafetyProfileStreams.h for helpful utility functions.

dds/DCPS/DomainParticipantFactoryImpl.cpp
```diff
@@ -90,6 +90,15 @@ DomainParticipantFactoryImpl::create_participant(
     this, domainId, value.id, par_qos, a_listener, mask, value.federated
   );
 
+  if (DCPS_debug_level > 0) {
```
mitza-oci: This should probably be a higher debug level.  See section 7.5.1 of the Developer's Guide.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1355,7 +1360,36 @@ int ignored)
               results.insert_sample(item, inst, ++i);
             }
         }
-    }
+      }
+    else
+      {
+        if (OpenDDS::DCPS::DCPS_debug_level > 0) {
```
mitza-oci: See the comment I wrote for InstanceState.cpp about streams.

tests/DCPS/BuiltInTopicTest/DataReaderListener.cpp
```diff
@@ -32,7 +31,8 @@ void DataReaderListenerImpl::on_data_available(DDS::DataReader_ptr reader)
   try {
     ::Messenger::MessageDataReader_var message_dr = ::Messenger::MessageDataReader::_narrow(reader);
     if (CORBA::is_nil (message_dr.in ())) {
-      cerr << "DataReaderListener: read: _narrow failed." << endl;
+      ACE_ERROR((LM_ERROR, ACE_TEXT(
```
mitza-oci: With this change, will the run_test.pl or prst_repo_run_test.pl stdout/stderr be useful?  Maybe the -ORBLogFile args need to go away?  In general, different tests have different approaches to this.  We can discuss offline too.

iguessthislldo: The reason I made the test this way what that when I began working on this test in earnest I was thinking it will inherently would print error messages that needed to be ignored. This was the case in the past but that doesn't seem to happen now.

PR #867, opened 2018-05-15
Use gcc8 as default and add clang6 as new compiler, gcc5/6 and clang5…

jwillemsen:
… as different configurations

    * .travis.yml:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
           packages:
             - libxerces-c-dev
             - clang++-5.0
-      env: COMPILER=clang++-5.0
+      env: COMPILER=clang++-5.0 NO_INLINE=1
     - compiler: clang
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


mitza-oci:
I see what it's complaining about with safety profile + GCC 8.  That's a combo we haven't tried yet.

In PoolAllocator.h (with safety profile enabled), there's a assumption that lexically putting the keyword `const` next to the key type produces the correct value type.  That's not correct when `K` is a pointer.

I think the right way to handle it is to write the equivalent of C++11's `std::add_const` metafunction (1-liner) and use it here.

mitza-oci:
Safety profile build will work once #869 is merged.

mitza-oci:
@jwillemsen please merge master to your branch, seems like there's no simple way to get travis to build the new merge commit on its own

jwillemsen:
@mitza-oci Master has been merged

mitza-oci:
The builds on Travis's matrix that are just configured as COMPILER=X are taking a long time.  We may need to add other options to those lines to keep them within the time limit.

PR #868, opened 2018-05-15
tests/security/certs/README.md

iguessthislldo:
Renamed file so Github will render as markdown.
Also some minor changes to the file itself.


mitza-oci:
OK, next time you can push directly to the security branch without making a PR.

PR #873, opened 2018-06-12
Use DOC Group TAO 2.5.0

jwillemsen:


configure
```diff
@@ -52,7 +52,9 @@ my %platforminfo =
                 'libpath' => 'DYLD_LIBRARY_PATH',
                 'versions' => {5 => 'leopard', 6 => 'snowleopard', 7 => 'lion',
                                8 => 'mountainlion', 9 => 'mavericks',
-                               10 => 'yosemite', 11 => 'elcapitan'},
+                               10 => 'yosemite', 11 => 'elcapitan',
```
mitza-oci: I don't think this is needed now that there's a common config-macosx.h.  Have you tried configuring on any of those versions?

jwillemsen: No, haven't tried that, just extended the list with the versions we currently have in ACE. Maybe merge this PR and make an issue to use the common MacOSX files in a separate PR?

mitza-oci: Since it's been working fine for Sierra and High Sierra without this change, I'm not inclined to merge it.  Instead we need to remove this part of the configure script (now that we can assume the user has a new enough ACE).

PR #900, opened 2018-08-14
Query Condition Fixes

iguessthislldo:
Contains the changes from #590, as well changes to omit dispose messages from queries if the query refers to fields other than "key".

### DIFF
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -8,6 +8,7 @@
 #include "dds/DCPS/TypeSupportImpl.h"
 #include "dds/DCPS/Watchdog.h"
 #include "dcps_export.h"
+#include "dds/DCPS/GuidConverter.h"
 
 #include "ace/Bound_Ptr.h"
 #include "ace/Time_Value.h"
@@ -1339,8 +1340,10 @@ int ignored)
   OpenDDS::DCPS::SubscriptionInstance_rch inst = get_handle_instance(a_handle);
   if (!inst) return DDS::RETCODE_BAD_PARAMETER;
 
-  if ((inst->instance_state_.view_state() & view_states) &&
-      (inst->instance_state_.instance_state() & instance_states))
+  InstanceState& state_obj = inst->instance_state_;
+  bool valid_view_state = state_obj.view_state() & view_states;
+  bool valid_instance_state = state_obj.instance_state() & instance_states;
+  if (valid_view_state && valid_instance_state)
     {
       size_t i(0);
       for (OpenDDS::DCPS::ReceivedDataElement* item = inst->rcvd_samples_.head_;
@@ -1355,7 +1358,34 @@ int ignored)
               results.insert_sample(item, inst, ++i);
             }
         }
-    }
+      }
+    else
+      {
+        if (OpenDDS::DCPS::DCPS_debug_level >= 8) {
+          OPENDDS_STRING msg;
+          if (!valid_view_state) {
+            msg += "view state is not valid";
+            if (!valid_instance_state) {
+              msg += " and ";
+            }
+          }
+          if (!valid_instance_state) {
+            msg = msg
+              + "instance state is "
+              + state_obj.instance_state_string()
+              + " while the validity mask is "
+              + InstanceState::instance_state_string(instance_states);
+          }
+          GuidConverter conv(get_subscription_id());
+          ACE_DEBUG((LM_DEBUG,
+            ACE_TEXT(
+              "(%P|%t) DataReaderImpl_T::read_instance_i: "
+              "will return no data reading sub %C because:\n  %C\n"
+            ),
+            OPENDDS_STRING(conv).c_str(), msg.c_str()
+          ));
+        }
+      }
 
   results.copy_to_user();
 
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -36,8 +36,8 @@ namespace OpenDDS {
     typedef DDSTraits<MessageType> TraitsType;
     typedef typename TraitsType::MessageSequenceType MessageSequenceType;
 
-    typedef OPENDDS_MAP_CMP(MessageType, DDS::InstanceHandle_t,
-                            typename TraitsType::LessThanType) InstanceMap;
+    typedef OPENDDS_MAP_CMP_T(MessageType, DDS::InstanceHandle_t,
+                              typename TraitsType::LessThanType) InstanceMap;
 
     class SharedInstanceMap
       : public RcObject
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -1628,9 +1628,8 @@ void store_instance_data(
 
   if ((is_dispose_msg || is_unregister_msg) && it == instance_map_.end())
   {
-     return;
+    return;
   }
-
 
   if (it == instance_map_.end())
   {
@@ -1922,11 +1921,6 @@ void finish_store_instance_data(unique_ptr<MessageTypeWithAllocator> instance_da
         item->dec_ref();
       }
     }
-  }
-
-  if (is_dispose_msg || is_unregister_msg)
-  {
-    instance_data.reset();
   }
 
   bool event_notify = false;
--- a/dds/DCPS/FilterEvaluator.cpp
+++ b/dds/DCPS/FilterEvaluator.cpp
@@ -22,6 +22,7 @@
 #include <stdexcept>
 #include <cstring>
 #include <algorithm>
+#include <cctype>
 
 namespace {
   const char MOD[] = "MOD";
@@ -44,6 +45,7 @@ FilterEvaluator::FilterEvaluator(const char* filter, bool allowOrderBy)
   : extended_grammar_(false)
   , filter_root_(0)
   , number_parameters_(0)
+  , has_unsafe_fieldnames_(false)
 {
   const char* out = filter + std::strlen(filter);
   yard::SimpleTextParser parser(filter, out);
@@ -62,6 +64,12 @@ FilterEvaluator::FilterEvaluator(const char* filter, bool allowOrderBy)
     } else {
       filter_root_ = walkAst(iter);
     }
+  }
+
+  // Check for Unsafe Field References
+  for (AstNode* iter = parser.GetAstRoot()->GetFirstChild(); iter;
+      iter = iter->GetSibling()) {
+    walkForUnsafeFieldNames(AstNodeWrapper(iter));
   }
 }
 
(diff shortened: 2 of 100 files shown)


dds/DCPS/FilterEvaluator.h
```diff
@@ -92,6 +92,11 @@ class OpenDDS_Dcps_Export FilterEvaluator : public RcObject {
 
   size_t number_parameters() const { return number_parameters_; }
 
+  bool has_unsafe_fields() const
```
jwillemsen: What is an "unsafe" field, wouldn't it be better to make it "has_key_fields"

iguessthislldo: Probably, I wasn't quite sure what to call it.

dds/DCPS/ReceivedDataElementList.h
```diff
@@ -20,10 +20,6 @@
 
 #include "dds/DdsDcpsInfrastructureC.h"
 
-#if !defined (ACE_LACKS_PRAGMA_ONCE)
```
mitza-oci: Why was this removed?  Seems like we have it in many places, see PR #897 for example.

dds/idl/metaclass_generator.cpp
```diff
@@ -341,7 +341,7 @@ namespace {
 
   void gen_isDcpsKey(IDL_GlobalData::DCPS_Data_Type_Info* info)
   {
-    if (info) {
+    if (info && info->key_list_.size()) {
```
jwillemsen: !empty instead of size?

dds/DCPS/QueryConditionImpl.h
```diff
@@ -51,10 +51,19 @@ class OpenDDS_Dcps_Export QueryConditionImpl
...
+   * Returns true if the sample matches the query.
+   */
   template<typename Sample>
-  bool filter(const Sample& s) const
+  bool filter(const Sample& s, bool has_invalid_data) const
   {
     ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, lock_, false);
+    const MetaStruct& meta = getMetaStruct<Sample>();
+    // Omit the sample from results if there are non-key fields in the query
+    // and the query only has key fields.
```
simpsont-oci: Shouldn't this say something like "and the received sample only has key fields."?

dds/DCPS/QueryConditionImpl.h
```diff
@@ -51,10 +51,19 @@ class OpenDDS_Dcps_Export QueryConditionImpl
...
+    if (evaluator_.has_non_key_fields(meta) && has_invalid_data) {
```
simpsont-oci: It seems like the call to "evaluator_.has_non_key_fields(meta)" is going to be constant for every pairwise combination of query string and sample type. Would it make sense to cache the result from this call somewhere? Possibly in a map inside FilterEvaluator?

dds/DCPS/DataSampleHeader.inl
```diff
@@ -161,3 +161,10 @@ OpenDDS::DCPS::DataSampleHeader::test_flag(DataSampleHeaderFlag flag,
   // Test flag bit.
   return base[FLAGS_OFFSET] & mask_flag(flag);
 }
+
+ACE_INLINE bool OpenDDS::DCPS::DataSampleHeader::valid_data() const {
```
simpsont-oci: This function name is a little misleading to me. Something like "is_unregister_or_dispose()" seems more appropriate.

dds/DCPS/QueryConditionImpl.h
mitza-oci: Once a QueryCondition is created, its expression can't change.  The QueryConditionImpl object can hold any additional info you need.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1053,8 +1058,10 @@ namespace OpenDDS {
     if (!sample.header_.content_filter_) { // if this is true, the writer has already filtered
       using OpenDDS::DCPS::ContentFilteredTopicImpl;
       if (content_filtered_topic_) {
-        if (sample.header_.message_id_ == OpenDDS::DCPS::SAMPLE_DATA
-            && !content_filtered_topic_->filter(static_cast<MessageType&>(*data))) {
+        if (
+          !content_filtered_topic_->filter(static_cast<MessageType&>(*data),
```
simpsont-oci: What's going on with the parentheses here?

dds/DCPS/DataSampleHeader.inl
iguessthislldo: It is a little but at the moment, but this function decides what samples are marked as having valid data, which is implementation defined. We just changed it, so I image it could change again. At least that was my thought process behind this function.

dds/DCPS/QueryConditionImpl.h
iguessthislldo: Yes, it was a typo and I will push a correction.

tests/DCPS/ContentFilteredTopic/ContentFilteredTopicTest.cpp
```diff
@@ -348,6 +344,163 @@ bool run_unsignedlonglong_test(const DomainParticipant_var& dp,
...
+  Topic_var topic = dp->create_topic(
+    "Messenger Topic", type_name,
+    TOPIC_QOS_DEFAULT, 0, DEFAULT_STATUS_MASK);
+
+  // Create ContentFilteredTopic
+  ContentFilteredTopic_var cft = dp->create_contentfilteredtopic(
+    expect_dispose ? "Safe Filtered Messenger Topic" : "Unsafe Filtered Messenger Topic",
+    topic, filter, StringSeq());
+  if (!cft) {
+    cerr << "ERROR: run_dispose_filter_test: creating ContentFilteredTopic failed" << endl;
```
simpsont-oci: Should these ERROR statements match the function name, run_single_dispose_filter_test?

dds/DCPS/FilterEvaluator.cpp
```diff
@@ -129,19 +143,45 @@ FilterEvaluator::~FilterEvaluator()
   delete filter_root_;
 }
 
+bool FilterEvaluator::has_non_key_fields(const MetaStruct& meta) const
+{
+  // Check Order
```
mitza-oci: Is this line (and 158) needed?

dds/DCPS/FilterEvaluator.h
```diff
@@ -175,6 +184,8 @@ class OpenDDS_Dcps_Export MetaStruct {
   virtual bool compare(const void* lhs, const void* rhs,
                        const char* fieldSpec) const = 0;
 
+  virtual bool isDcpsKey(const char*) const = 0;
```
mitza-oci: The parameter should have a name

dds/DCPS/QueryConditionImpl.h
```diff
@@ -51,10 +51,21 @@ class OpenDDS_Dcps_Export QueryConditionImpl
 
   bool hasFilter() const;
 
+  /**
+   * Returns true if the sample matches the query.
+   */
   template<typename Sample>
-  bool filter(const Sample& s) const
+  bool filter(const Sample& s, bool has_invalid_data) const
```
mitza-oci: My comment from ContentFilteredTopic applies here too.

dds/DCPS/ReceivedDataElementList.h
```diff
@@ -51,38 +52,48 @@ class OpenDDS_Dcps_Export ReceivedDataElement {
       ref_count_(1),
       mx_(mx)
   {
-
     this->destination_timestamp_ = time_value_to_time(ACE_OS::gettimeofday());
 
     this->source_timestamp_.sec = header.source_timestamp_sec_;
     this->source_timestamp_.nanosec = header.source_timestamp_nanosec_;
+
+    // When we have a dispose or unregister instance we shouldn't let the
```
mitza-oci: This could be reworded in terms of SampleInfo from the spec (which is where `valid_data` comes from)

dds/DCPS/ReceivedDataElementList.h
```diff
@@ -51,38 +52,48 @@ class OpenDDS_Dcps_Export ReceivedDataElement {
...
+  long ref_count()
+  {
     return this->ref_count_.value();
   }
 
   PublicationId pub_;
 
-  /// Data sample received
-  void * const registered_data_;  // ugly, but works....
+  /// Data sample received, could only be the keyed fields
```
mitza-oci: "key fields" instead of "keyed fields"

dds/idl/metaclass_generator.cpp
```diff
@@ -338,6 +338,26 @@ namespace {
     }
     be_global->impl_ << "    }\n";
   }
+
+  void gen_isDcpsKey(IDL_GlobalData::DCPS_Data_Type_Info* info)
+  {
+    if (info && info->key_list_.size()) {
+      for (
```
mitza-oci: The use of the key_list_ in keys_generator.cpp:74 looks simpler, see if that style would work here too.

dds/idl/metaclass_generator.cpp
```diff
@@ -338,6 +338,26 @@ namespace {
...
+        IDL_GlobalData::DCPS_Key_List::CONST_ITERATOR i(info->key_list_);
+        !i.done(); i.advance()
+      ) {
+        ACE_TString* key;
+        i.next(key);
+        be_global->impl_ <<
+          "    if (!ACE_OS::strcmp(field, \"" << *key << "\")) {\n"
```
mitza-oci: This is a narrow-character stream so the string in key is going to need conversion in wchar builds.

dds/DCPS/FilterEvaluator.cpp
iguessthislldo: Probably not, I just try to comment on what think of as "paragraphs" of code and these are definitely short and simple paragraphs so I would be alright with taking them out.

dds/DCPS/ContentFilteredTopicImpl.h
```diff
@@ -42,10 +42,21 @@ class OpenDDS_Dcps_Export ContentFilteredTopicImpl
...
-  bool filter(const Sample& s) const
+  bool filter(const Sample& s, bool sample_only_has_key_fields) const
   {
     ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, lock_, false);
+    const MetaStruct& meta = getMetaStruct<Sample>();
+    /*
+     * Omit the sample from results if the filter references non-key fields
+     * and the sample only has key fields.
+     */
+    if (filter_eval_.has_non_key_fields(meta) && sample_only_has_key_fields) {
```
jwillemsen: Maybe swap order, if sample_only_has_key_fields is false no need to execute has_non_key_fields

dds/DCPS/QueryConditionImpl.h
```diff
@@ -51,10 +51,21 @@ class OpenDDS_Dcps_Export QueryConditionImpl
...
-  bool filter(const Sample& s) const
+  bool filter(const Sample& s, bool sample_only_has_key_fields) const
   {
     ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, lock_, false);
+    const MetaStruct& meta = getMetaStruct<Sample>();
+    /*
+     * Omit the sample from results if the query references non-key fields
+     * and the sample only has key fields.
+     */
+    if (evaluator_.has_non_key_fields(meta) && sample_only_has_key_fields) {
```
jwillemsen: switch order?

PR #904, opened 2018-08-15
Deferred URL parsing logic to LocalAccessCredentialData members

mmcparlane:


dds/DCPS/security/AccessControl/LocalAccessCredentialData.cpp
```diff
@@ -6,125 +6,57 @@
...
-#include <fstream>
-
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace Security {
 
 LocalAccessCredentialData::LocalAccessCredentialData()
 {
+
```
mitza-oci: Not sure why this line (and 21) are here, but OK...

dds/DCPS/security/AccessControl/LocalAccessCredentialData.cpp
```diff
@@ -6,125 +6,57 @@
...
 }
 
 bool LocalAccessCredentialData::load(const DDS::PropertySeq& props,
                                      DDS::Security::SecurityException& ex)
 {
-  const std::string file("file:");
-  bool permission = false, governance = false, ca = false;
+  std::string name, value;
```
mitza-oci: The previous change reduced the scope of `name` and `value` which also enables const.  The change proposed here undoes this.

dds/DCPS/security/AccessControl/LocalAccessCredentialData.cpp
```diff
@@ -6,125 +6,57 @@
...
 
   for (size_t i = 0; i < props.length(); ++i) {
-    const std::string name = props[i].name.in();
-    const std::string value = props[i].value.in();
+    name = props[i].name;
```
mitza-oci: Change the type of `i` to `unsigned int` to match what sequences use -- this will avoid a truncation warning on some compilers/platforms where size_t is 64-bit.

PR #905, opened 2018-08-16
We are using now DOC Group TAO 2.5.1 for OpenDDS

jwillemsen:
* README.md:


mitza-oci:
The README gets updated as part of the release process.

PR #911, opened 2018-08-17
CMake on PATH now overrides auto-detection routine

mmcparlane:


configure
```diff
@@ -920,23 +921,39 @@ sub path_parent_dir {
...
 
-    die "ERROR: CMake not found on PATH"
-      unless defined $opts{'cmake'};
+      unless ($opts{'cmake'}) {
+        print "CMake not found in path; searching default location...\n"
+          if $opts{'verbose'};
+
+        my $cmake_default =
+          (is_windows() ?
+            (-f 'c:\\Program Files\\CMake\\bin\\cmake.exe' ?
```
jwillemsen: There is a windows env for the program files path

mitza-oci: Yes, we need to use the environment for this:
```
ProgramFiles=C:\Program Files
ProgramFiles(x86)=C:\Program Files (x86)
```

mmcparlane: Alas! Thank you. I shall use that instead :)

PR #913, opened 2018-08-20
dissector: Check if Deserialization Failed

iguessthislldo:
I decided put checks on all the deserializations, instead of just the WChar type like discussed.

### DIFF
--- a/tools/dissector/sample_dissector.cpp
+++ b/tools/dissector/sample_dissector.cpp
@@ -322,106 +322,127 @@ namespace OpenDDS
       TAO::WString_Manager wstring_value;
       std::string converted;
 
-      switch (this->type_id_) {
+      switch (type_id_) {
       case Char:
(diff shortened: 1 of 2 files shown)


tools/dissector/sample_dissector.cpp
```diff
@@ -322,106 +322,127 @@ namespace OpenDDS
...
 
       case Boolean:
         ACE_CDR::Boolean boolean_value;
-        p.serializer >> ACE_InputCDR::to_boolean(boolean_value);
-        if (boolean_value)
-          s << "true";
-        else
-          s << "false";
+        if (p.serializer >> ACE_InputCDR::to_boolean(boolean_value)) {
+          s << boolean_value ? "true" : "false";
```
mitza-oci: this needs parens around the right-hand operand of <<

PR #914, opened 2018-08-20
Fixing unreachable code warning. Fix multiple permissions grant support.

simpsont-oci:
Fixing unreachable code warning for check_XXX_topic by adding subject name check. Fixing subject name and domain id checking for multiple grants per permissions file.


simpsont-oci:
Ah, yes it does. I did run the test but forgot to add to the list.

PR #916, opened 2018-08-21
Flexible git submodule support

mmcparlane:


MPC/config/googlemock.mpb
```diff
@@ -1,5 +1,8 @@
 project : googletest {
-  includes += $(DDS_ROOT)/tests/googletest/googlemock/include
-  libpaths += $(DDS_ROOT)/tests/googletest/build/googlemock/$(Configuration)
+
+  // For backwards-compatibility.
```
mitza-oci: This isn't going to make sense to an external audience.
It could be "for compatibility with tests/gtest_setup.txt instructions" or "for compatibility with build tree (not installed) gtest".
Same thing in googletest.mpb.

configure
```diff
@@ -156,14 +156,15 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'wireshark_build:s', 'Wireshark CMake Build Location',
     'wireshark_lib:s', 'Wireshark CMake libraries location, relative to wireshark_build',
     'glib:s', 'GLib for Wireshark (use GLIB_ROOT or system pkg)',
-    'rapidjson', 'RapidJSON for Wireshark Sample Dissection(optional)',
+    'rapidjson:s', 'RapidJSON for Wireshark Sample Dissection(optional)',
```
mitza-oci: I realize it was there before, but I don't see why the space is missing before (optional).

configure
```diff
@@ -156,14 +156,15 @@ my @specs = # Array of array-refs, each inner array is an option group which
...
     'qt:s', 'Qt (use QTDIR or system pkg)',
     'qt-include:s', 'Qt include dir (use QT4_INCDIR or QTDIR/include)',
     'boost:s', 'Boost (use BOOST_ROOT or system pkg)',
     'boost-version:s', 'Boost version (only if needed to find headers)',
     'xerces3:s', 'Xerces-C++ 3 for QoS XML handling, DDS Security (optional)',
     'ssl:s', 'OpenSSL for DDS Security (optional)',
     'cmake:s', 'Path to cmake binary for compiling Google Test Framework (optional)',
+    'gtest:s', 'Path to Google Test Framework root (optional)',
```
mitza-oci: Remove "root"

configure
```diff
@@ -156,14 +156,15 @@ my @specs = # Array of array-refs, each inner array is an option group which
...
     'wireshark_lib:s', 'Wireshark CMake libraries location, relative to wireshark_build',
     'glib:s', 'GLib for Wireshark (use GLIB_ROOT or system pkg)',
-    'rapidjson', 'RapidJSON for Wireshark Sample Dissection(optional)',
+    'rapidjson:s', 'RapidJSON for Wireshark Sample Dissection(optional)',
     'qt:s', 'Qt (use QTDIR or system pkg)',
     'qt-include:s', 'Qt include dir (use QT4_INCDIR or QTDIR/include)',
     'boost:s', 'Boost (use BOOST_ROOT or system pkg)',
     'boost-version:s', 'Boost version (only if needed to find headers)',
     'xerces3:s', 'Xerces-C++ 3 for QoS XML handling, DDS Security (optional)',
     'ssl:s', 'OpenSSL for DDS Security (optional)',
```
mitza-oci: Rename --ssl to --openssl

configure
```diff
@@ -954,18 +960,73 @@ sub targets_win64 {
...
 }
 
+sub system_default_install_dir {
+  if (is_windows()) {
+    return targets_win64() ? $ENV{'ProgramFiles'} : $ENV{'ProgramFiles(x86)'};
+  }
+  return '/usr';
+}
+
+## Detect Google Test
```
mitza-oci: This is completely clear without the comment (same with "detect RapidJSON" and "Build google test")

configure
```diff
@@ -1121,25 +1182,34 @@ sub win32_cmake_generator {
...
   my $cwd = Cwd::getcwd();
-  my $build_dir = 'tests' . $slash . 'googletest' . $slash . 'build';
+  my $build_dir = $opts{'gtest'} . $slash . 'build';
+  my $install_dir = $build_dir . $slash . 'install';
   my $cmake = $opts{'cmake'};
   my @cmake_cmds = ((is_windows() ?
                       ["$cmake",
                        "-D", "gtest_force_shared_crt=ON",
                        "-D", "CMAKE_CXX_FLAGS=/D_SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING=1",
+                       "-D", "CMAKE_INSTALL_PREFIX=$install_dir",
```
mitza-oci: Perhaps this is redundant with the earlier discussion, but what do we gain by doing another step here (adding complexity)?

MPC/config/googlemock.mpb
mmcparlane: Agreed. I'll get the change in!

configure
mmcparlane: Along the lines of our earlier discussion, instead of depending upon the Google Test source tree, sub-modules will depend upon the include/lib directory of the compiled sources (which is Google Test is used in the system packages).

configure
```diff
@@ -954,18 +960,73 @@ sub targets_win64 {
   return $arch eq 'x64';
 }
 
+sub system_default_install_dir {
+  if (is_windows()) {
+    return targets_win64() ? $ENV{'ProgramFiles'} : $ENV{'ProgramFiles(x86)'};
```
simpsont-oci: Isn't there usually a space between "ProgramFiles" and "(x86)"? Although maybe that's just the environment variable and not the path. More importantly, we need to make sure we're supporting win32 arch, which is still windows, but is not 64 bit, and won't have the (x86) stuff in it.

configure
mitza-oci: Within this conditional block we know we're not using a pre-built or system-installed Google Test, since we are actually building it.

mmcparlane: True, however, within the googletest.mpb having one standard include/lib directory based on the GTEST_ROOT is much cleaner, instead of having dependencies on the googletest sources. Suppose Google Test decided to move their files around in the sources (which they are free to do) then we would have to make a change; by using this install directory we guard against such changes in the future.

mitza-oci: The submodule setup guarantees we get version 1.8.  If a user doesn't use submodules, our documentation states we're compatible with 1.8.

mmcparlane: The benefit of being correct/future-proof outweighs the small (and temporary) complexity added. Especially given that the code change is already in! :)

PR #917, opened 2018-08-23
Fix googletest compatibility path

mmcparlane:


tests/gtest_setup.txt
```diff
@@ -1,16 +1,80 @@
...
+
+  2. Follow the steps in "Using a Git Submodule" with the following differences:
+      a. The install location can be anywhere you desire, or it can be completely
+         omitted (in which case the default system location will be used).
+
+      b. An elevated terminal/cmd shell will probably be required if the install
+         prefix is omitted.
+
+      c. When invoking the OpenDDS configure script, if the default system location
+         is _not_ used, then the gtest switch must be used to inform MPC where the
```
jwillemsen: Shouldn't it say "inform the configure script", that this is passed internally to MPC is not important for the user at this point

mmcparlane: Agreed. I am also going to add other instructions about setting the environment variable and invoking mwc directly (bypassing configure altogether) as well.

PR #919, opened 2018-08-28
Dockerfile for 32-bit shapes demo

jrw972:


Dockerfile.shapes32
```diff
@@ -0,0 +1,30 @@
...
+RUN dpkg --add-architecture i386 && apt-get update && apt-get install -y \
+    curl \
+    g++ \
+    make \
+    perl-base \
+    perl-modules \
+    lib32gcc-5-dev \
+    lib32stdc++-5-dev \
+    libboost1.58-dev \
+    libc6-dev-i386 \
```
mitza-oci: Looks like this shouldn't be needed.  libc6-dev-i386 provides libraries in /usr/lib32, which shouldn't be used.  libc6-dev:i386, on the other hand, provides libraries in /usr/lib/i386-linux-gnu, which should be used.

PR #920, opened 2018-08-29
Simplified some code and added some const

jwillemsen:
* dds/DCPS/RecorderImpl.cpp:


mitza-oci:
We are following the core guidelines style for const placement:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-const

PR #923, opened 2018-08-30
Fix detection of MSVC compiler x.x.x.x

ajkerzner:
A Microsoft C/C++ compiler version with 3 or more periods (e.g. "Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x64") will now also be detected.


ajkerzner:
Fixes issue #923

ajkerzner:
I do not believe that the one build error was something I introduced.

PR #924, opened 2018-08-30
Use Qt5 in monitor and ishapes

iguessthislldo:
This relies on [Qt5 Support in MPC](https://github.com/DOCGroup/MPC/pull/69).


tools/excelRTD/ExcelRTD.mpc
```diff
@@ -9,7 +9,7 @@ project(ExcelRTD): taolib, dcpslib, dcps_inforepodiscovery, dcps_tcp, dcps_monit
     compile_flags += /bigobj
   }
 
-  specific (vc10, vc11, vc12, vc14) {
+  specific (vc10, vc11, vc12, vc14, vs2017) {
```
jwillemsen: Could this not be `specific(prop:microsoft)`

iguessthislldo: Well it's a Excel plugin so the whole thing is Windows specific, but I got a different idea:

The README says we support MSVC 9, which I'm guessing would be the current reason for the else case but I'm thinking maybe flip the expression:

```
specific (vc9) {
  macros += RTD_DLL_NAME=<%equote%>$(TargetFileName)<%equote%>
} else {
  macros += RTD_DLL_NAME=<%quote%>$(TargetFileName)<%quote%>
}
```

mitza-oci: We can also drop vc9 support for this feature.

iguessthislldo:
Converted the ishapes README to Markdown and merged the README and docs for Monitor. Documentation for ExcelRTD still needs to be rewritten/converted to Markdown.

configure
```diff
@@ -1323,6 +1306,74 @@ sub push_libpath {
...
+  my $qt_lib;
+
+  if ($try_to_use_qt_system_pkg) {
+    # Try to use pkg-config to get Qt locations
+    my $mesg = ", stopped\nPlease either install Qt or provide the Qt " .
+      "locations (see docs/qt.md for details).\n";
+    `pkg-config --print-variables Qt5Core`;
+    if ($!) {
+      die "ERROR: Trying to use system Qt package but could not confirm " .
+        "Qt and/or pkg-config exists on the system$mesg" .
```
mitza-oci: This is being reported as an error on at least some versions of Perl.  The statement seems to end in `.` instead of `;`.

PR #930, opened 2018-09-13
Starter azure devops pipeline file

jwillemsen:


jwillemsen:
@mitza-oci Can you setup an azure org and merge this so that we can further enhance this azure setup

mitza-oci:
See #946

PR #932, opened 2018-09-18
Better support for Qt5 in the Configure Script

iguessthislldo:
- Automatic Support for when Qt Build tools have a suffix
- Automatic Support for when Qt is an installed vcpkg package.

I will still need to update the documentation (and fix a few thing the qt documentation).

### DIFF
--- a/configure
+++ b/configure
@@ -265,9 +265,6 @@ my $cross_compile = 0;
 my %opts = %{parseArgs()};
 
 my $debug = exists $opts{'debug'} ? $opts{'debug'} : 1;
-if ($opts{'verbose'}) {
-  print "Debug is " . ($debug ? 'enabled' : 'disabled') . '\n';
-}
 
 $opts{'host'} = perlOS_to_host() unless $opts{'host'};
 
--- a/configure
+++ b/configure
@@ -264,9 +264,16 @@ sub parseArgs {
 my $cross_compile = 0;
 my %opts = %{parseArgs()};
 
+my $debug = exists $opts{'debug'} ? $opts{'debug'} : 1;
+if ($opts{'verbose'}) {
+  print "Debug is " . ($debug ? 'enabled' : 'disabled') . '\n';
+}
+
 $opts{'host'} = perlOS_to_host() unless $opts{'host'};
 
-my ($slash, $exeext) = ($opts{'host'} eq 'win32') ? ('\\', '.exe') : ('/', '');
+my $is_windows = $opts{'host'} eq 'win32';
+
+my ($slash, $exeext) = is_windows ? ('\\', '.exe') : ('/', '');
 
 my %specific =
   ($slash eq '/' ?
@@ -382,7 +389,7 @@ else {
   }
   if (!defined $opts{'compiler'}) {
     die "ERROR: Can't find a compiler, set PATH or run this script with the ".
-      "--compiler option.\n" . ($slash eq '\\' ? "  For Microsoft Visual C++, ".
+      "--compiler option.\n" . (is_windows ? "  For Microsoft Visual C++, ".
                                 "run this script from the Visual Studio ".
                                 "Command Prompt.\n" : '') . "Stopped";
   }
@@ -463,7 +470,7 @@ sub setSomeEnv {
   if ($opts{'dry-run'} && !$notdir && looksRelative($val)) {
     $val = getcwd . $slash . $val;
   }
-  $val =~ s!/!\\!g if $slash eq '\\';
+  $val =~ s!/!\\!g if is_windows;
   $hashref->{$name} = $val;
 }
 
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -264,9 +264,16 @@ sub parseArgs {
 my $cross_compile = 0;
 my %opts = %{parseArgs()};
 
+my $debug = exists $opts{'debug'} ? $opts{'debug'} : 1;
+if ($opts{'verbose'}) {
```
mitza-oci: Is this level of detail needed in the verbose log output?  It already dumped out the `%opts`.

docs/qt.md
```diff
@@ -12,16 +12,55 @@ Qt5 support must be enabled in the build system.
 
 ## Qt5
 
-Qt5 is a framework primarily used for cross-platform GUI based applications
-and is freely available from [qt.io](https://qt.io) or through your
-system's package manager:
+Qt5 is a framework primarily used for cross-platform GUI based applications.
+The Qt official website as of writing is [qt.io](https://qt.io).
+Qt can be downloaded directly from there but might require a account to be
```
mitza-oci: It hasn't been my experience that an account (not `a account`) is required, but I had to create one to file a bug.  I didn't attempt to download prebuilt binaries or QtCreator.  I think we can skip the discussion of accounts here.

docs/qt.md
```diff
@@ -12,16 +12,55 @@ Qt5 support must be enabled in the build system.
...
-  - `qtbase5-dev` on Ubuntu
-  - `qt5-qtbase-devel` on Fedora
-  - Windows and macOS would probably be best served by downloading it from [qt.io](https://qt.io)
+Also remember the options listed here are not the only ones.
+Any functional instance of Qt5 would work as long as it is built and has the
+headers and the basic Qt development tools (QtCreator is not required).
 
-Any instance of Qt5 would work as long as it is built and has the headers
-and basic Qt development tools.
+### Getting Qt Using Package Mangers
```
mitza-oci: s/Mangers/Managers/

docs/qt.md
```diff
@@ -12,16 +12,55 @@ Qt5 support must be enabled in the build system.
...
+
+Many package managers have Qt development packages available:
+
+#### Linux
+
+  - `qtbase5-dev` on Debian based systems
```
mitza-oci: `Debian-based` (same with Fedora below)

docs/qt.md
```diff
@@ -12,16 +12,55 @@ Qt5 support must be enabled in the build system.
...
+#### macOS
+
+  - `qt5` using [brew](https://brew.sh/)
+  - `qt5-qtbase` using [MacPorts](https://www.macports.org/)
+
+#### Windows
+
+The only package manager on Windows that appears to have a reliable Qt5
+development package is [vcpkg](https://github.com/Microsoft/vcpkg) as
+`qt5-base`. There is a `qt5` package that installs a complete Qt environment,
```
mitza-oci: Please clarify here that the `qt5` package is not required.

docs/qt.md
```diff
@@ -49,9 +88,13 @@ or the install prefix. If headers were places some where other than
...
 also specify their location using `--qt-include`.
 
+Other options can be used listed in "Without the Configure Script" section
+below.
+
 #### Windows Example
 
-```configure --qt C:\Qt\5.11.1\msvc2017_64
+```
+configure --qt C:\Qt\5.11.1\msvc2017_64
```
mitza-oci: Use `--qt=` for consistency with configure's help message.

docs/qt.md
```diff
@@ -64,23 +107,35 @@ It is possible to build OpenDDS without using the configure script, although
 this is not recommended unless one is familiar with OpenDDS and it's build
```
mitza-oci: s/it's/its/

docs/qt.md
```diff
@@ -64,23 +107,35 @@ It is possible to build OpenDDS without using the configure script, although
...
 system, MPC. This list explains how the configure script tells MPC how to
 build Qt5 applications, but not how to build OpenDDS without the configure
-script in general.
+script in general. These environment variables can also be used with the
+configure script to override their values.
 
-  - The `qt5` feature must be enabled (e.g. `-features qt5=1`).
+  - The `qt5` feature must be enabled (e.g. `perl mpc.pl ... -features qt5=1 ...`).
```
mitza-oci: I see why this change is here but it seems like it could be more confusing than helpful.  When I run MPC the script I typically use is `mwc.pl` and I don't have a `perl` prefix since that's not necessary with the `#!` behavior on Linux and the file association on Windows.

docs/qt.md
```diff
@@ -64,23 +107,35 @@ It is possible to build OpenDDS without using the configure script, although
...
   - `QTDIR` environment variable points to the root of the Qt instance.
     - This is supplied to the configure script using the `--qt` option and
```
mitza-oci: This entire section is about configuring the build manually (without the configure script) so the references to the configure script here could be moved elsewhere.

docs/qt.md
iguessthislldo: They definitely seem to want you to create an account, but I just tried their installer and it does has an option to skip signing in, so I will remove it.

configure
iguessthislldo: No, I think I was using it to test the debug variable above.

PR #936, opened 2018-09-23
 Wrong entity_refs_ counter when deleting data reader in the TopicDescriptionImpl

### LINKED ISSUE #935: Entity reference count is not decreased in the TopicDescriptionImpl when deleting a data reader
The entity_refs_ is not drecreased when calling `delete_datareader`, therefore when trying to call `delete_topic` I receive `RETCODE_PRECONDITION_NOT_MET`.

Similar in the `delete_contentfilteredtopic`, in this case is drecreased for the related topic but never for the filtered topic.

On the other hand, the `delete_multitopic` never increase the reference count for the multitopic itself (It does for the related topics), therefore I never receive `RETCODE_PRECONDITION_NOT_MET` even if I didn't delete the data reader.

Attached a fix proposal.
[Patches.zip](https://github.com/objectcomputing/OpenDDS/files/2407919/Patches.zip)

jmmorato:
Fixes #935

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -125,14 +125,6 @@ DataReaderImpl::DataReaderImpl()
 DataReaderImpl::~DataReaderImpl()
 {
   DBG_ENTRY_LVL("DataReaderImpl","~DataReaderImpl",6);
-
-#ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
-  OwnershipManagerPtr owner_manager = this->ownership_manager();
-  if (owner_manager) {
-    owner_manager->unregister_reader(topic_servant_->type_name(), this);
-  }
-#endif
-
 }
 
 // this method is called when delete_datareader is called.
@@ -143,6 +135,14 @@ DataReaderImpl::cleanup()
   // back onto the listener at the moment the related DDS entity has been
   // deleted
   set_listener(0, NO_STATUS_MASK);
+
+#ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
+  OwnershipManagerPtr owner_manager = this->ownership_manager();
+  if (owner_manager) {
+    owner_manager->unregister_reader(topic_servant_->type_name(), this);
+  }
+#endif
+
   topic_servant_ = 0;
 
 #ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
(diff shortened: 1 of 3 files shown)


jmmorato:
In fact, the test already exists, at least for the content filtered topic `tests\DCPS\ContentFilteredTopic ` (I got the idea for my unit test from there), but I guess it is not called during CI process.
For the MultiTopic could be easily extended in `tests\DCPS\MultiTopic`. For the Topic itself, I cannot see something similar on the test folder but perhaps you could give me an idea where is the best place to code it.

BTW, I having problems to configure  --security --features=versioned_namespace=1 --no-inline (the configuration failing in the pull request) because a not found qt5_witges dependency. If someone could take a look on it would be great.

iguessthislldo:
> BTW, I having problems to configure --security --features=versioned_namespace=1 --no-inline (the configuration failing in the pull request) because a not found qt5_witges dependency. If someone could take a look on it would be great.

Actually there two seperate known issues going on here:

1. Currently we do not pay Travis, so we get to use Travis for free because OpenDDS is open source, but our time is limited to 50 minutes, so this fails in at least one of the cases all the time. It would actually be surprising if everything finished up in 50 minutes and passed.

2. There has been a recent switch from Qt4 to Qt5 and that change had to be made across MPC, ACE/TAO, and OpenDDS. Unfortunately these changes have not been incorporated into the versions of ACE/TAO/MPC that OpenDDS configure script downloads by default yet. For now It fails unless you  download the latest MPC separately. It's not hard thing to fix though, you download MPC from https://github.com/DOCGroup/MPC and set the `--mpc` option to there when you run the configure script.

mitza-oci:
ok to test

This should start the Jenkins build on our server.  When I have some more time available I can give it a full review and take a look at what broke and how to address it with test coverage.

mitza-oci:
Now that #938 is merged, the required test coverage has been added.
In the Jenkins build for this PR,  the Ownership test now fails with:
```
ERROR: <subscriber> exited with coredump from signal 11 : SEGV
ERROR: subscriber returned 255 
test FAILED.
Error: tests/DCPS/Ownership/run_test.pl returned with status 256
```

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -143,8 +143,15 @@ DataReaderImpl::cleanup()
   // back onto the listener at the moment the related DDS entity has been
   // deleted
   set_listener(0, NO_STATUS_MASK);
+  topic_servant_ = 0;
```
iguessthislldo: This is referenced in `~DataReaderImpl()`, which is causing the segfault in the ownership test.

PR #949, opened 2018-10-10
Fixing build when no-builtin-topic and security flags are both configured

simpsont-oci:
Found a build error when trying different configure option combos for footprint testing.

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -1869,10 +1869,15 @@ namespace {
 
   bool
   is_bit(const char* topic_name) {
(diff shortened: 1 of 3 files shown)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1869,10 +1869,15 @@ namespace {
 
   bool
   is_bit(const char* topic_name) {
+#if !defined (DDS_HAS_MINIMUM_BIT)
```
mitza-oci: Could this be refactored to use the existing function topicIsBIT in dds/DCPS/BuiltInTopicUtils.h?

PR #950, opened 2018-10-11
Added doc_group_master switch to download ACE/TAO/MPC from DOC Group …

mmcparlane:
…Github


mmcparlane:
The diff on this is a bit confusing; basically, mostly everything that existed within the changed if-block was wrapped in an else {...} and indented accordingly.

jwillemsen:
Wouldn't it be easier to add a flag to specify the ACE/TAO branch and a flag to specify the MPC branch? That gives the flexibility to also use the configure script when using a custom ACE/TAO branch

mitza-oci:
@jwillemsen We thought about adding more extensive options, but the "advanced" user can just run `git clone` before `configure` and do any number of different things (shallow clone, use a fork, checkout a branch/tag).  If the resulting repos are in the expected locations, `configure` will find them automatically.  If not, their locations can be specified via environment or command-line options.

The motivation for doing this is that maintainers (not end users) may need the latest versions of the dependencies.  When I first wrote it, the configure script was not intended for maintainers to use.  But developers found it useful, so we thought this option would be convenient.

PR #959, opened 2018-10-23
 Add clang7 to travis-ci configuration 

jwillemsen:


mitza-oci:
Should we remove one or more of the older clang versions?  If we only add build configurations we're adding to the overall build latency.

jwillemsen:
clang5 is still heavily used, would like to propose to keep it there for some more time.

PR #971, opened 2018-11-02
Add vcpkg with xercesc/openssl to the Visual Studio 2017 jobs

jwillemsen:
* azure-pipelines.yml:


jwillemsen:
@mitza-oci any feedback on this PR?

mitza-oci:
I started to work on it, then had to go do other things.  One simple thing would be to enable DDS security which I don't think is in there yet.  I'd like to see if we can somehow cache the results of building the 3rd-party libs from source so it doesn't add to the build times.  I saw that vcpkg has some limited support for debug-only or release-only builds but it seemed to be buggy (at least for xerces).

jwillemsen:
Standard azure pipelines doesn’t have a cache between jobs as far as I could find. One alternative is to compile and zip the packages and download them from somewhere but that adds maintenance effort.

mitza-oci:
Looks like this could be useful? https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=vsts&tabs=yaml

ras0219-msft:
We (vcpkg) are actively working on a recommendation for this scenario; currently we're thinking that the simplest option will be to:
1. Try to download a pipeline artifact named "deps" from the most recent successful build. If successful, unpack it as `installed/`
2. `vcpkg remove --outdated`
3. `vcpkg install <all your things>`
4. Pack up the `installed/` folder as "deps" and upload it as an artifact

This should give you equivalent "caching" behavior as appveyor.

jwillemsen:
@ras0219-msft Yes, that is equivalent to caching and as far as I can tell it will work. Probably different job configurations can result in different invocations of vcpkg, would be nice when this works on a per job basis (which we can have multiple ones in one azure-pipelines.yml)

@ras0219-msft We also want to use clang5/6/7 on azure-devops on Linux as we currently do on travis-ci, is there any documentation how to add that?

ras0219-msft:
>would be nice when this works on a per job basis 

I think you could just name the artifact with a tag per job -- `deps_x86`, `deps_x64`, etc.

>how to add [clang]?

_Right now_, it is as easy as just `CXX=clang++ CC=clang vcpkg install xyz`.

However, this is arguably a bug -- your current external environment should not be affecting the build as much as it currently does. In the long term, it would be preferable to have a dedicated triplet[1] and cmake toolchain file like `x64-linux-clang`.

[1] https://github.com/Microsoft/vcpkg/blob/master/docs/users/triplets.md

jwillemsen:
@mitza-oci @ras0219-msft Found a way to get clang to work, testing right now on a smaller project, when that works I will extend the azure-pipelines for OpenDDS

jwillemsen:
See #998 for the PR to add clang as compiler for Ubuntu Linux, created #999 to add Visual Studio 2015 as next step. @ras0219-msft Any schedule for the caching behavior availability?

ras0219-msft:
No, you'd need to implement it yourself via the publish/download build artifacts tasks:

1. https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-build-artifacts?view=vsts
2. https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-build-artifacts?view=vsts

jwillemsen:
@ras0219-msft Do you maybe have an azure pipelines yml with vcpkg which from which we can copy the exact steps? Or maybe a reference to another project that already does use this?

ras0219-msft:
No, sorry :(

jwillemsen:
> No, sorry :(

@ras0219-msft Hopefully the azure devops team can make an example soon. Btw, any idea why on https://github.com/DOCGroup/ACE_TAO/pull/759 azure-devops isn't shown on the checks tab?

mitza-oci:
Unfortunately http://strawberryperl.com/ is currently offline, hampering our Windows builds on Azure Pipelines.  Hopefully it will be back soon, or else we'll need to change the Perl download location.

PR #974, opened 2018-11-05
Don't use sudo anymore and use travis_latest as group

jwillemsen:
as described in https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-group: travis_latest
 language: cpp
 
 compiler:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+group: travis_latest
 language: cpp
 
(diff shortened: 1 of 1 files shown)


jwillemsen:
Looks travis is taking longer and longer, maybe reduce what we build there and increase what we build on Azure DevOps, there is no limit there

mitza-oci:
I agree that with Azure Pipelines offering multi-platform support without time limits we can de-emphasize Travis as we keep migrating build configurations to Azure.

For this PR in particular, why do we want `travis_latest` instead of LTS?

jwillemsen:
Changed it just to travis_latest because we do it in more projects, just use the latest version for CI.

jwillemsen:
On azure multiple gcc versions works, only haven't had time to try to add clang

PR #976, opened 2018-11-08
Also check ACE_HAS_POSIX_SEM

jwillemsen:
* dds/DCPS/transport/shmem/ShmemTransport.cpp:

### DIFF
--- a/dds/DCPS/transport/shmem/ShmemTransport.cpp
+++ b/dds/DCPS/transport/shmem/ShmemTransport.cpp
@@ -144,12 +144,12 @@ ShmemTransport::configure_i(ShmemInst& config)
                              0 /*no name*/);
   ACE_sema_t ace_sema = *pSem;
   ok = (*pSem != 0);
-# elif defined (ACE_HAS_POSIX_SEM) && !defined (ACE_LACKS_UNNAMED_SEMAPHORE)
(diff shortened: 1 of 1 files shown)


jwillemsen:
See ace/os_include/os_semaphore.h, the ACE_sema_t struct is guarded by ACE_HAS_POSIX_SEM

mitza-oci:
The earlier `ifdef` (at the start of `ShmemTransport::configure_i`) was intended to provide the user a nicer error message.  We could also extend that to cover the impacted API level of Android either directly or via the `ACE_HAS_POSIX_SEM` macro.
The convention in OpenDDS is not to add extra `()` around a macro name after the `defined`, `ifdef`, `ifndef` operators, I can't think of a case where having `()` helps with anything.

mitza-oci:
ACE doesn't have a common abstraction, we need to pick from ACE_Pagefile_Memory_Pool and ACE_Shared_Memory_Pool (see ShmemDataLink.h).  Once we are dealing with ACE_Pagefile_Memory_Pool we are on Win32 and can just use its stable API.

It's been a long time but as far as I can remember the other ACE utilities weren't desirable because of interactions with the filesystem, assumptions on mapping addresses, or other unwanted OS interactions.

Issue #982, opened 2018-11-23
Recommendation on update std::string to OPENDDS_STRING


niuxu18:
- OpenDDS-3.13: 

Our team works on enhance logging practices by learning form historical log revisions.
We find four patches in OpenDDS evolution which updated _std::string_ to _OPENDDS_
And we find two missed spots in OpenDDS-3.13:

- The missed spots are:

1. File: dds/DCPS/InfoRepoDiscovery/DataWriterRemoteImpl.cpp, Line: 41, Code:

```
void
DataWriterRemoteImpl::add_association(const RepoId& yourId,
         const ReaderAssociation& reader,
         bool active)
{
   if (DCPS_debug_level) {
      GuidConverter writer_converter(yourId);
      GuidConverter reader_converter(reader.readerId);
      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) DataWriterRemoteImpl::add_association - ")
            ACE_TEXT("local %C remote %C\n"),
            std::string(writer_converter).c_str(),
            std::string(reader_converter).c_str()));
   }
```

2. File: dds/DCPS/InfoRepoDiscovery/DataReaderRemoteImpl.cpp, Line: 41, Code:

```
void
DataReaderRemoteImpl::add_association(const RepoId& yourId,
         const WriterAssociation& writer,
         bool active)
{
   if (DCPS_debug_level) {
      GuidConverter writer_converter(yourId);
      GuidConverter reader_converter(writer.writerId);
      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) DataReaderRemoteImpl::add_association - ")
            ACE_TEXT("local %C remote %C\n"),
           std::string(writer_converter).c_str(),
           std::string(reader_converter).c_str()));
   }
```

- Corresponding patches that support our opinion are:

1. Version: OpenDDS-3.6 vs OpenDDS-3.7, File: dds/DCPS/DataReaderImpl.cpp, Line: 1338, Patch:
```
} else if (DCPS_debug_level > 4) {
   // This may not be an error since it could happen that the sample
   // is delivered to the datareader after the write is dis-associated
   // with this datareader.
   GuidConverter reader_converter(subscription_id_);
   GuidConverter writer_converter(header.publication_id_);
   ACE_DEBUG((LM_DEBUG,
      ACE_TEXT("(%P|%t) DataReaderImpl::writer_activity: ")
[...]

PR #983, opened 2018-11-23
replace std::string with OPENDDS_STRING in some case

niuxu18:
By learning from historical log revisions, we find several log revisions all **update std::string to OPENDDS_STRING**. With this knowledge, we find some missed log revision in the OPENDDS-3.13 version. Following is our patches.

If more information of the historical log revisions we have found is needed, please leave a comment ^^


mitza-oci:
> If more information of the historical log revisions we have found is needed, please leave a comment

Read the code that defines OPENDDS_STRING in `dds/DCPS/PoolAllocator.h`.  It is only distinct from `std::string` when safety profile is in use.  Safety profile only covers a subset of OpenDDS features, see the developer's guide for more details.

See Issue #982

niuxu18:
> > If more information of the historical log revisions we have found is needed, please leave a comment
> 
> Read the code that defines OPENDDS_STRING in `dds/DCPS/PoolAllocator.h`. It is only distinct from `std::string` when safety profile is in use. Safety profile only covers a subset of OpenDDS features, see the developer's guide for more details.
> 
> See Issue #982

Yeah, i got it! Thanks for your reply ^^

PR #1009, opened 2018-12-11
Add OpenDDS to the cmake project name to make clear these are OpenDDS…

jwillemsen:
… projects

    * tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt:
    * tests/cmake_integration/Messenger/Messenger_2/CMakeLists.txt:
    * tests/cmake_integration/README.md:

### DIFF
--- a/tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
+++ b/tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
@@ -2,7 +2,7 @@
 # file or http://www.opendds.org/license.html for details.
 #
 
-project(Messenger_cmake CXX)
+project(OpenDDS_Messenger_cmake CXX)
 cmake_minimum_required(VERSION 3.8.2)
 
 find_package(OpenDDS)
--- a/tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
+++ b/tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
@@ -2,7 +2,7 @@
 # file or http://www.opendds.org/license.html for details.
 #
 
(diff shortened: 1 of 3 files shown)


jwillemsen:
Prevents duplicate project names within a ROS2 tree

mitza-oci:
I'm not opposed to the change but I can't say I understand why it would be needed.  These are just examples of using the CMake support, they aren't intended to be part of any other workspace.

jwillemsen:
When checking out opendds as part of ros2 the ros2 tooling complained about the duplicate project name

mitza-oci:
So it finds and uses all `CMakeLists.txt` recursively without a way to exclude anything?  If using submodules you'll also have the rapidjson and gtest ones.  Where can we file an issue on "ros2 tooling"?

jwillemsen:
I took the easy approach, still learning ros2

mitza-oci:
Please add a link to the GitHub repo of the ros2 tool that's doing this.

jwillemsen:
Probably the colcon ros2 tooling has this limitation

mitza-oci:
According to https://index.ros.org/doc/ros2/Linux-Development-Setup/#build-the-code-in-the-workspace, AMENT_IGNORE may be used for this.

jwillemsen:
Yes, I am using that for testing but would it cause any harm to just let OpenDDS provide unique project names within its own tree for simplicity?

mitza-oci:
Like I was saying before, I don't think it's simpler to change upstream code in unexpected ways to satisfy some odd downstream requirement.  There will be other CMakeLists.txt files in the source tree if submodules are used, and it's possible that more of this kind of thing will be added in the future.

PR #1014, opened 2018-12-13
Remove old file, contents is already in README.md

jwillemsen:
* examples/DCPS/ishapes/SHAPESDEMO_README:
      Deleted.


iguessthislldo:
From when I worked on the Qt5 transition, I think there was some script that created a separate archive for this that included the ishapes executable, maybe some other files, and this file but not the README(.md). I think I meant to update this but forgot about it.

iguessthislldo:
Looking at the blames it was PR #454

jwillemsen:
Why not package the README.md with the separate archive?

mitza-oci:
README.md is a typical readme file for maintainers or those building from source.  SHAPESDEMO_README is included with the [binary archives](http://opendds.org/quickstart/GettingStartedShapesDemo.html) for those users.  One or both of these files probably need updates.

mitza-oci:
We will refresh these docs as part of #1008

PR #1023, opened 2018-12-20
Multicast repeater tool

jrw972:
We are moving away from non-standard discovery (InfoRepo) and
transports (TCP, UDP, multicast, etc.) and moving toward RTPS
discovery and transport.  RTPS relies on multicast for discovery.
However, many environments, like the cloud and the public internet do
not support multicast.

Solution: Write a multicast to unicast repeater for discovery.  The
repeater runs on each host with a participant and subscribes to the
discovery multicast group.  When it receives a message from the
participant, it forwards it via unicast to other hosts running the
repeater.  When the repeater receives a message via its unicast
socket, it repeats it on the multicast group so the local participants
can receive the message.  The repeater only needs to be used for
participant discovery since SEDP and data transport can use unicast.

The list of hosts to be contacted by the repeater can be static or
dynamic.  Currently, the repeater supports gathering a list of hosts
in a Google Cloud Platform (GCP) instance group.

### DIFF
--- a/bin/dcps_tests.lst
+++ b/bin/dcps_tests.lst
@@ -374,3 +374,5 @@ tests/FACE/Compiler/idl_test_fixed/run_test.pl: !DCPS_MIN !WCHAR !NO_BUILT_IN_TO
 
 # Test for unregister_type method added as a type support extension (assuming no requirement when performing minimal test)
 tests/DCPS/UnregisterType/run_test.pl: !DCPS_MIN
+
+tools/repeater/tests/basic_forward.pl: NODEJS
--- a/tools/repeater/README.rst
+++ b/tools/repeater/README.rst
@@ -1,0 +1,65 @@
+===========================
+Multicast Datagram Repeater
+===========================
+
+The multicast datagram repeater allows existing multicast-dependent programs
+to be used (without configuration or code changes) on networks that don't route
+multicast traffic among all hosts.
+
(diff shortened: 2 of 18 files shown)


bin/dcps_tests.lst
```diff
@@ -374,3 +374,5 @@ tests/FACE/Compiler/idl_test_fixed/run_test.pl: !DCPS_MIN !WCHAR !NO_BUILT_IN_TO
 
 # Test for unregister_type method added as a type support extension (assuming no requirement when performing minimal test)
 tests/DCPS/UnregisterType/run_test.pl: !DCPS_MIN
+
+tools/repeaters/tests/basic_forward.pl: NODEJS
```
mitza-oci: s/repeaters/repeater/

mitza-oci:
Please move files client.js, sender.js, receiver.js to the tests directory.

tools/repeater/README.rst
```diff
@@ -0,0 +1,64 @@
+===========================
+Multicast Datagram Repeater
+===========================
+
+The multicast datagram repeater allows existing multicast-dependent programs
+to be used (without configuration or code changes) on networks that don't route
+multicast traffic among all hosts.
+
+The multicast datagram repeater takes datagrams that are recevied on a
```
jwillemsen: Received

Issue #994, opened 2018-11-27
C++11 mapping: DCPS_DATA_KEY for complex expressions


mitza-oci:
When using the new IDL-to-C++11 mapping, the code that's generated by DCPS_DATA_KEY is not yet aware of complex key expressions (those containing a `.` or `[]`).

mmcparlane:
This issue is being worked in the following branch:
https://github.com/objectcomputing/OpenDDS/tree/issue-994-cpp11-mapping

PR #1030, opened 2019-01-10
Complex keys in the C++11 mapping (Issue 994)

mmcparlane:


dds/idl/gen_util.h
```diff
@@ -0,0 +1,33 @@
+
+#ifndef OPENDDS_IDL_GEN_UTIL_H
+#define OPENDDS_IDL_GEN_UTIL_H
+
+#include "be_extern.h"
+
+namespace {
```
mitza-oci: Remove the `namespace` here.  This is for two reasons:
- in general, anonymous namespaces shouldn't be in headers
- since `opendds_idl` is not a library, there is no chance of name collision with end-uesr code

dds/idl/gen_util.h
```diff
@@ -0,0 +1,33 @@
+
```
mitza-oci: Does this need its own file?  Other helpers are in dds_generator.h.

dds/idl/gen_util.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef OPENDDS_IDL_GEN_UTIL_H
+#define OPENDDS_IDL_GEN_UTIL_H
+
+#include "be_extern.h"
+
+namespace {
+  inline
+  std::string insert_cxx11_accessor_parens(
+                std::string full_var_name, bool is_union_member) {
```
mitza-oci: Convention is to pass string by const ref.

dds/idl/gen_util.h
```diff
@@ -0,0 +1,33 @@
...
+    const bool use_cxx11 = be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+    if (!use_cxx11 || is_union_member) return full_var_name;
+
+    std::string::size_type n = 0;
+    while ((n = full_var_name.find(".", n)) != std::string::npos) {
```
mitza-oci: There is an overload of `find` that takes a single character, seems like we should prefer that (here and in the while loop below).

dds/idl/gen_util.h
```diff
@@ -0,0 +1,33 @@
...
+      } else {
+        ++n;
+      }
+    }
+    n = 0;
+    while ((n = full_var_name.find("[", n)) != std::string::npos) {
+      full_var_name.insert(n, "()");
+      n += 3;
+    }
+    return full_var_name[full_var_name.length()-1] == ']'
```
mitza-oci: Prefer using `size()` instead of `length()` (works on all standard containers).  Also a minor nitpick, but the guidelines are to have space before and after the `-` binary operator.

dds/idl/gen_util.h
mmcparlane: Since full_var_name gets modified within the function a copy makes sense here, does it not?

mmcparlane: I suppose it gives a performance improvement when c++11 is not being used or if its a union member at the expense of another variable :)

mitza-oci:
Link to Issue #994

PR #1035, opened 2019-01-15
Test case for durable data taking too long to sync up

mitza-oci:
(using rtps_udp transport)

regression test for fix on branch rtps-durable


mitza-oci:
Integrating into PR #1001 which fixes the issue

PR #1039, opened 2019-01-17
Adding CMake tests to CI environments

mmcparlane:


mmcparlane:
Since the CMake tests are a bit more complicated I think it would be best to create a script like tests/cmake_integration/run_ci_tests.sh as in [complex-build-commands](https://docs.travis-ci.com/user/job-lifecycle/#complex-build-commands)

PR #1045, opened 2019-01-24
RTPS Discovery Performance Improvements

simpsont-oci:
Allow association to immediately reschedule heartbeat timer
Delay heartbeat messages for durable writers until durable data is ready
Immediately ack/nack first 'real' heartbeat data from writers


dds/DCPS/SequenceNumber.h
```diff
@@ -39,7 +39,7 @@ class OpenDDS_Dcps_Export SequenceNumber {
         // this code is here, despite the RTPS spec statement:
         // "sequence numbers never wrap"
         this->high_ = 0;
-        this->low_ = 1;
+        this->low_ = 0;
```
mitza-oci: Why is 0 considered part of the normal range of sequence numbers?  Could this be left at 1?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -317,8 +316,8 @@ RtpsUdpDataLink::associated(const RepoId& local_id, const RepoId& remote_id,
     RtpsWriter& w = writers_[local_id];
     w.remote_readers_[remote_id].durable_ = remote_durable;
     w.durable_ = local_durable;
-    enable_heartbeat = true;
-
+    w.ready_to_hb_ = !(local_durable && remote_durable);
```
mitza-oci: Simplify this expression: there is no way for a matched remote reader to be durable when the local writer isn't

dds/DCPS/SequenceNumber.h
simpsont-oci: I'm not sure why this code is here at all if the spec says it's not supposed to wrap. Maybe just have ++ do nothing when sequence number is at max value?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Sure. I originally had this is as "w.ready_to_hb_ = !local_durable" which was insufficient in the case when the remote reader wasn't durable so I just added remote_durable to the condition. Since the matched remote reader can't be durable when the local writer isn't durable, then I guess you can just use "!remote_durable" to cover all cases when the durable data doesn't need to be added before the first heartbeat.

PR #1053, opened 2019-02-05
Add support for Visual Studio 2019

jwillemsen:


jwillemsen:
Looks the cmake version is < 3.14 on azure vs2019, need to wait on that to continue this

azure-pipelines.yml
```diff
@@ -23,6 +23,8 @@ jobs:
   - script: |
       git clone -q --depth 1 git://github.com/Microsoft/vcpkg.git %VCPKG_ROOT%
       %VCPKG_ROOT%\bootstrap-vcpkg.bat
+    displayName: Bootstrap vcpkg
```
mitza-oci: This is what the `call` keyword is used for.

jwillemsen: Yes, you are right, fixed!

mitza-oci:
Should we split this into two PRs?  The first can update configure & mpb/mpc files without extending Azure CI.

jwillemsen:
> Should we split this into two PRs? The first can update configure & mpb/mpc files without extending Azure CI.

I have removed OpenSSL and the cmake test build at the end for vs2019, for both we need a newer cmake

jwillemsen:
Had to make a small change because it seems the release of VS2019 is now on azure, without OpenSSL and cmake test at the end the azure devops build with vs2019 is now green

mitza-oci:
In the Azure CI script, why did vcpkg switch back to using our own clone instead of the one preinstalled in the image?

jwillemsen:
> In the Azure CI script, why did vcpkg switch back to using our own clone instead of the one preinstalled in the image?

vs2019 lacks preinstalled and the vs2017 is not the most recent, for adding wchar as next step we need the most recent version

PR #1055, opened 2019-02-05
update idl v8 generation to use decimal strings for signed and unsign…

simpsont-oci:
…ed 64-bit integers to avoid javascript 'number' truncation for numbers above 2^53

### DIFF
--- a/dds/idl/v8_generator.cpp
+++ b/dds/idl/v8_generator.cpp
@@ -121,13 +121,14 @@ namespace {
           "  }\n";
         return;
 
-      } else if (v8Type == "Number") {
-        prefix = "static_cast<double>(";
-        suffix = ")";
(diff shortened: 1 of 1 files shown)


dds/idl/v8_generator.cpp
```diff
@@ -122,6 +124,10 @@ namespace {
       } else if (v8Type == "Number") {
         prefix = "static_cast<double>(";
         suffix = ")";
+      } else if (v8Type == "String" && (pt == AST_PredefinedType::PT_longlong || pt == AST_PredefinedType::PT_ulonglong)) {
```
mitza-oci: Is the check for v8Type here redundant?  Isn't it enough to check pt?

simpsont-oci: Hmm, yeah, redundant. I just wanted the change applied as narrowly as possible and saw checks being made against both type variables above and so I just added both. I'll fix it.

PR #1065, opened 2019-02-19
V8 publishing support

simpsont-oci:
Type support for writing capabilities for node.js / v8 code generation. Seems to work for all the node-opendds idl test cases (look for subsequent pull request there as well).


dds/idl/v8_generator.cpp
```diff
@@ -206,6 +414,32 @@ bool v8_generator::gen_struct(AST_Structure*, UTL_ScopedName* name,
       "    return OpenDDS::DCPS::copyToV8(*static_cast<const " << lname
                << "*>(source));\n"
       "  }\n\n"
+      "  void* fromV8(const v8::Local<v8::Object>& source) const\n"
+      "  {\n"
+      "    " << lname << "* result = new " << lname << "();\n"
```
mitza-oci: The heap allocation here doesn't seem to have a corresponding deallocation.  It makes sense to allocate here since we need to have a `void*` interface.  I think we need an additional virtual method to handle deallocation, to be called by node-opendds.cpp.  See MetaStructImpl (metaclass_generator.cpp) for another example where we need to manipulate data samples indirectly via a `void*`.

dds/idl/v8_generator.cpp
```diff
@@ -168,6 +353,19 @@ namespace {
...
   };
+
+  struct gen_field_copyfrom {
+    gen_field_copyfrom(const char* tgt, const char* src) : tgt_(tgt), src_(src) {}
+    const std::string tgt_, src_;
+    void operator()(AST_Field* field) const
+    {
+      const std::string fieldName = field->local_name()->get_string(),
+        source = src_,
+        prop = fieldName;
```
mitza-oci: The `prop` variable seems to be unnecessary, just use `fieldName` directly.  Also the use of `source` can just be `src_`.

dds/idl/v8_generator.cpp
```diff
@@ -243,6 +477,24 @@ bool v8_generator::gen_typedef(AST_Typedef*, UTL_ScopedName* name,
...
+        "  if (src->IsArray()) {\n"
+        "    length = src->Get(Nan::New(\"length\").ToLocalChecked())->ToObject()->Uint32Value();\n"
+        "  }\n"
+        "  out.length(length);\n"
+        "  for (uint32_t i = 0; i < length; ++i) {\n"
+        "  ";
+      gen_copyfrom("out", "src", elem, "i", true);
+      be_global->impl_ <<
+        "  }\n"
+        "  return;\n";
```
mitza-oci: Why `return;` here? (Same applies to gen_union and gen_struct.)

dds/idl/v8_generator.cpp
```diff
@@ -156,6 +156,191 @@ namespace {
...
+                    const char* prop, bool prop_index = false, bool fun_assign = false, std::ostream& strm = be_global->impl_)
+  {
+    const Classification cls = classify(type);
+    AST_PredefinedType::PredefinedType pt = AST_PredefinedType::PT_void;
+    const std::string propName = prop_index ? std::string(tgt) + "[" + prop + "]" : std::string(tgt) + "." + prop,
+      assign_prefix = fun_assign ? "(" : " = ",
+      assign_suffix = fun_assign ? ")" : "";
+
+    if (cls & CL_SCALAR) {
+      std::string clines;
```
mitza-oci: Why does `clines` need to be built first instead of output in-order?  In other words, I think it would be cleaner to have the prefix part of the line 263 block (conditional on `prop_index`) happen at this point.  That would let you replace `clines` with direct output.

dds/idl/v8_generator.cpp
```diff
@@ -156,6 +156,191 @@ namespace {
...
+        const std::string underscores =
+          dds_generator::scoped_helper(type->name(), "_"),
+          array = "gen_" + underscores + "_names";
+        clines += "      if (lv->IsNumber()) {\n";
+        clines += "        v8::Local<v8::Integer> li = Nan::To<v8::Integer>(lv).ToLocalChecked();\n";
+        clines += "        " + propName + assign_prefix + "static_cast<" + scoped(type->name()) + ">(li->Value())" + assign_suffix + ";\n";
+        clines += "      }\n";
+        clines += "      if (lv->IsString()) {\n";
+        clines += "        v8::Local<v8::String> ls = Nan::To<v8::String>(lv).ToLocalChecked();\n";
+        clines += "        std::string ss(ls->Length(), ' ');\n";
```
mitza-oci: Looks like this should use `Utf8Length` to match `WriteUf8`.  (I think this occurs in a few other places too.)

dds/idl/v8_generator.cpp
```diff
@@ -156,6 +156,191 @@ namespace {
...
+        clines += "        ls->WriteUtf8(&ss[0]);\n";
+        clines += "        for (uint32_t i = 0; i < sizeof(" + array + ")/sizeof(" + array + "[0]); ++i) {\n";
+        clines += "          if (ss == " + array + "[i]) {\n";
+        clines += "            " + propName + assign_prefix + "static_cast<" + scoped(type->name()) + ">(i)" + assign_suffix + ";\n";
+        clines += "            continue;\n";
```
mitza-oci: Should this be a `break`?

dds/idl/v8_generator.cpp
```diff
@@ -156,6 +156,191 @@ namespace {
...
+        clines += "        std::istringstream iss(ss);";
+        clines += "        " + scoped(type->name()) + " temp_ll;\n";
+        clines += "        if (ss.find(\"0x\") != std::string::npos) {\n";
+        clines += "          iss >> std::hex >> temp_ll;\n";
+        clines += "        } else {\n";
+        clines += "          iss >> temp_ll;\n";
+        clines += "        }\n";
+        clines += "        " + propName + assign_prefix + "temp_ll" + assign_suffix + ";\n";
+        clines += "      }\n";
+        clines += "      if (lv->IsNumber()) {\n";
```
mitza-oci: It might be good to refactor this part (if, Nan::To, assign) since it's very similar for longlong, the other ints, the floats, and boolean.

dds/idl/v8_generator.cpp
simpsont-oci: clines is a holdover from early on when I thought it would just be a couple of lines in the middle that held a conditional that was variable on the type. But sometimes the conditional was bigger than expected, and sometimes there was more than one conditional, so it grew into just "everything in the middle, after we've gotten the field's v8::Value". I'll try to clean this up a bit.

dds/idl/v8_generator.cpp
simpsont-oci: Good catch, I'll add a deletion method and update the node-opendds PR to use it.

dds/idl/v8_generator.cpp
simpsont-oci: Just blindly following the pattern of the other method. Turns out it doesn't make sense here, so I'll update it.

dds/idl/v8_generator.cpp
simpsont-oci: Again, just following the pattern from what was there before. It maybe helps the reader (of this file) tell when the generated method is complete, but pretty pointless in the output code. I'll get rid of them.

dds/idl/v8_generator.cpp
simpsont-oci: Doh. Yep. Fixing...

dds/idl/v8_generator.cpp
simpsont-oci: I combined all int-types into a single section, so now all int types have decimal and hex string parsing support. Looking back that this, though, I'm not sure if that's what you meant? I suppose I should add string parsing for float types and bool as well?

dds/idl/v8_generator.cpp
simpsont-oci: the last two examples you're looking at are actually for wide strings, so Length() is correct.

dds/idl/v8_generator.cpp
```diff
@@ -206,6 +428,37 @@ bool v8_generator::gen_struct(AST_Structure*, UTL_ScopedName* name,
...
+      "    return result;\n"
+      "  }\n\n"
+      "  void deleteFromV8Result(void* val) const\n"
+      "  {\n"
+      "    " << lname << "* delete_me = static_cast< " << lname << "*>(val);\n"
+      "    delete delete_me;\n"
+      "  }\n\n"
+      "  DDS::InstanceHandle_t register_instance_helper(DDS::DataWriter* dw, const void* data) const\n"
+      "  {\n"
+      "    " << lname << "DataWriter* dw_t = dynamic_cast<" << lname << "DataWriter*>(dw);\n"
```
jwillemsen: What when the dynamic cast fails?

simpsont-oci: I'd hope the correct type would always get passed since this is generated code, but OK. It doesn't hurt to add a check and return nil / bad parameter.

PR #1066, opened 2019-02-21
Improve Android Support and Documentation

iguessthislldo:
- Fix `ClassNotFoundException` caused by `TAOObject.finalize()` when using the Java API on Android.
- ~Prevent the possibility of packaged OpenSSL conflicting with the system SSL library.~ [Moved to ACE](https://github.com/DOCGroup/ACE_TAO/pull/858)
- Fix build error when building Compiler tests, because `DDS_HAS_WCHAR` was not defined.
- `android.md` 
  - Fix mistakes
  - Expand based on experience with making an Android app that uses OpenDDS.


jmmorato:
It would be nice to have extended documentation on "Cross-Compiling IDL Libraries" section ;)

iguessthislldo:
It would but that's why this is a draft pull request. Also I would probably have that section done already but there are bugs in the Makefile support for IDL to Java that makes generating Java IDL libraries a pain, at least for the app I'm using as a test bed for these changes. I need to figure out how to word the workaround.

jmmorato:
Sure, not pressure, your effort is already appreciated. 

I never worked with OpenDDS and Java, but I guess you are compiling with NDK the C++ code and after creating a JNI wrapper over that code, don't you? If so, my interest is mostly in the NDK compilation; if you can give me some clues on that will be welcome.

We can continue the discussion in private if you prefer.

iguessthislldo:
You're correct.

Most of the work has already been done (a little work by me, but mostly by others), this is just seeing how OpenDDS (also how it  doesn't) work with Android and documentation that and fixing what I can. 

The build support for Android lives in ACE/TAO and is described in the ACE's install html file. The actual support is defined in [`config-android.h`](https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/config-android.h) and [`platform_android.GNU`](https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/ace/include/makeinclude/platform_android.GNU). The trickiest parts of these files that I have dealt with are defining what compiler to look for on the `PATH` if it's not definitely clang (Older NDKs only) and dealing with the combinations of NDK and API versions. A good part of my Android OpenDDS efforts have been put into [OpenDDS-Android](https://github.com/iguessthislldo/OpenDDS-Android) which can build various combinations of API and NDK versions.

For the Java/JNI side, OpenDDS's existing Java API seems to work with Android, with the exception of the bug that I'm pushing a fix for here.

docs/android.md
```diff
@@ -43,23 +43,26 @@ are mostly for shorthand.
...
+   writing. The OpenDDS `configure` script can use DOC Group ACE/TAO
+   automatically by passing `--doc_group`.
+ - [Android Native Development Kit (NDK)](https://developer.android.com/ndk/)
+   r18 or higher. You can download it separately from android.com or using the
+   SDK Manager that comes with Android Studio. If you download the NDK using the
+   SDK Manager, this is located in `$SDK/ndk-bundle`.
+ - Some knowledge about OpenDDS and Android development will be assumed, but
+   more OpenDDS knowledge will be assumed than Android knowledge.
+
+In addionion to those, building OpenDDS with optional dependencies also
```
jwillemsen: addition

Issue #486, opened 2017-03-27
No support for top-level union types


jwillemsen:
OpenDDS lacks the support for unions as top-level union types. Where the DCPS specification is silent about union (not mentioning it at all) the DDS X-Types specification does mention unions as types for top-level, the discriminator may even be key. Below a small IDL example that is rejected by opendds_idl

```
module DDS_Native
{
  module Uni
  {
#pragma DCPS_DATA_TYPE "DDS_Native::Uni::TopUnionMessage"
    union TopUnionMessage switch (long) {
      case 4:
        short x;
      case 5:
      case 6:
        long y;
    };
  }; // module Uni
}; // module DDS_Native
```

iguessthislldo:
Closing because #1067 was merged.

Issue #366, opened 2016-12-21
No support for @key annotation


jwillemsen:
DDS-XTypes/IDL4 now specify @key for annotating the key fields of an IDL type. Seems OpenDDS doesn't support this yet.

mitza-oci:
There are lots of things in lots of other specs that OpenDDS doesn't support.  OpenDDS supports the specs that it is documented as supporting.  I'm not sure the usefulness of a GitHub issue for this sort of thing.

I'm sure we'll be adding more support for parts of XTypes and IDL4 in the future.  Does TAO_IDL_FE parse annotations?

jwillemsen:
As far as I know tao_idl doesn't parse annotations, ridlc does

mitza-oci:
I'm not sure how ridlc is relevant to this issue.  OpenDDS itself needs to know what the key fields are.  OpenDDS's IDL processing is based on the TAO_IDL_FE parser.

jwillemsen:
I mentioned ridlc as alternative, no idea how much effort it would be to extend TAO_IDL_FE to handle the IDL4 annotations

iguessthislldo:
Closing because #1067 was merged.

PR #1067, opened 2019-02-21
XTypes topic and key annotations, Unions as top-level types

iguessthislldo:
Building on [DOCGroup/ACE_TAO#723](https://github.com/DOCGroup/ACE_TAO/pull/723), this aligns `opendds_idl` closer to the XTypes standard by providing topic annotations such as `@topic` and `@key` as an alternative to `DCPS_DATA` pragma statements (addressing #366) and allowing unions to be used as topic types (addressing #486). This also makes IDL4 support the default for OpenDDS.

# TODO

- [x] Update for [DOCGroup/ACE_TAO#822](https://github.com/DOCGroup/ACE_TAO/pull/822) annotation API Changes
- [x] Fix `tests/FACE/Compiler` tests
- [x] Implement `@nested` and changes agreed on in [DDSXTY13-69](https://issues.omg.org/browse/DDSXTY13-69)
  - For each valid type (`struct` or `union`) whether or not to generate type support is determined by these rules:
    - If `@topic` is applied
      - Always generate
      - If `@nested` is also applied and `@nested.value` is true
        - Warn about conflicting annotations
    - else if `@nested` is applied
      - Generate if `@nested.value` is false.
    - else inherit the parent module's default.
  - For each submodule the type support generation default is:
    - If `@default_nested` is applied, generate by default if `@default_nested.value` is false.
    - If `@default_nested` isn't applied, inherit the parent module's default.
  - For the root module, the ultimate parent of all modules, do not generate type support by default, unless a command line argument is passed to change that.
- [x] Warn (once per file, command line option can silence) for any use of `#pragma DCPS_DATA_TYPE`
- [x] Add tests somewhere in `tests/DCPS/Compiler/annotations` for `@nested` and `@default_nested`.
- [x] UnionTopic Test
- [x] Update CMake Module to pass `--idl-version 4 --unknown-annotations ignore` to `tao_idl`.
- [x] Reference Developer's Guide in warning about `#pragma DCPS_DATA_TYPE`.
- [x] Update CMake Module to optionally pass `--default-nested`
- [x] Resolve CMake Issues
- [x] Support Multidimensional Array Keys
- [x] Refractor `marshal_generator.cpp`
- ~Have Java API work with Annotations~ Will be addressed in another PR


iguessthislldo:
Thought the java failures were just part of the new java changes, but I realized the errors were about type support for types that shouldn't have type support -- Java for the core OpenDDS library wasn't getting the new `--default-nested` option from MPC. I will push the fix after it builds on my machine.

cmake/dds_idl_sources.cmake
```diff
@@ -135,7 +135,7 @@ function(opendds_target_idl_sources target)
     unset(_ddsidl_cmd_arg_-o)
     unset(_ddsidl_cmd_arg_-Wb,java)
 
-    cmake_parse_arguments(_ddsidl_cmd_arg "-SI;-GfaceTS;-Wb,java" "-o" "" ${_ddsidl_flags})
+    cmake_parse_arguments(_ddsidl_cmd_arg "-SI;-GfaceTS;-Wb,java --default-nested" "-o" "" ${_ddsidl_flags})
```
iguessthislldo: @ceneblock Looking at this again, I'm not sure what this line is actually doing. I know it doesn't work though, because in Azure the call to opendds_idl in the cmake test doesn't include `--default-nested`.

dds/DCPS/Serializer.cpp
```diff
@@ -46,13 +48,29 @@ Serializer::reset_alignment()
...
+ * @brief copies data from one pointer to another
+ * @input to the destination string
+ * @input from the source string
+ * @input n how much of the source string to throw
+ * @throws std::out_of_range in the event that the from string is null
+ */
 void
 Serializer::smemcpy(char* to, const char* from, size_t n)
 {
+  if(from)
```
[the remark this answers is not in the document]
simpsont-oci: Actually, do we even need this? Just make this an assert or something so the calling code fixes its issues, don't slow down our memcpy's with an additional check and exception handling.

dds/idl/be_global.cpp
```diff
@@ -319,6 +325,9 @@ BE_GlobalData::parse_args(long& i, char** av)
   static const char EXPORT_FLAG[] = "--export=";
   static const size_t EXPORT_FLAG_SIZE = sizeof(EXPORT_FLAG) - 1;
 
+  static const char DEFAULT_TS_MACRO[] = "--default-nested";
+  static const size_t SZ_DEFAULT_TS_MACRO = sizeof(DEFAULT_TS_MACRO) - 1;
```
simpsont-oci: Any chance we can be consistent (either direction) for how we name our parameters and their sizes? (see above)

dds/idl/be_global.cpp
```diff
@@ -602,3 +613,147 @@ BE_GlobalData::get_include_block(BE_GlobalData::stream_enum_t which)
...
+ * by the module level of @default_nesed or through the --default_nested=
+ * @param node the node to be inspected
+ * @return if the node is default nested or not
+ * @note false means it should be considered nested.
+ * @author ceneblock
+ */
+bool
+BE_GlobalData::is_default_nested(AST_Decl* node)
+{
+  if(node->defined_in()){
```
simpsont-oci: space after keyword (if)

simpsont-oci: I won't harp on these anymore, but there are several more, so maybe do a global search / replace in touched files?

dds/idl/be_global.cpp
```diff
@@ -602,3 +613,147 @@ BE_GlobalData::get_include_block(BE_GlobalData::stream_enum_t which)
...
+
+    ///recurse to the top
+    bool default_nested_apply_value = is_default_nested(dynamic_cast<AST_Decl *>(node -> defined_in()));
```
simpsont-oci: What is going on with the spaces around "->" ? Also, do these need to be dynamic casts? If so, should we be checking for failure?

dds/idl/be_global.cpp
```diff
@@ -602,3 +613,147 @@ BE_GlobalData::get_include_block(BE_GlobalData::stream_enum_t which)
...
+
+    ///check if we have a default value
+    AST_Annotation_Appl *default_nested_apply = dynamic_cast<AST_Decl *>(node -> defined_in()) -> annotations().find("::@default_nested");
+
+    ///if we have a default value, then overwrite the parent's.
+    if(default_nested_apply){
+      default_nested_apply_value = AST_Annotation_Member::narrow_from_decl ((*default_nested_apply)["value"]) -> value ()->ev ()->u.bval;
```
simpsont-oci: Use a C++ cast here

dds/idl/be_global.cpp
```diff
@@ -602,3 +613,147 @@ BE_GlobalData::get_include_block(BE_GlobalData::stream_enum_t which)
...
+    }
+  }
+  return false;
+}
+
+bool
+BE_GlobalData::is_key(AST_Field* node)
+{
+  // Check for @key
+  return node && node->annotations().find(key_annotation_);
```
simpsont-oci: Does find return NULL on failure or something?

tests/DCPS/Messenger_Annotation/Args.h
```diff
@@ -0,0 +1,103 @@
+/*
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef MESSENGER_TEST_ARGS_H
```
simpsont-oci: Do these guards need to match file name / paths? It seems a shame to duplicate all this code (I'm assuming it's pretty similar) just for the sake of testing some different IDL compiler options. It would have been nice if  we have figured out a way to optionally change how the normal messenger test pulled in its IDL & type support.

dds/idl/be_global.cpp
iguessthislldo: I have some changes I've not pushed yet where I fixed this.

dds/idl/be_global.cpp
iguessthislldo: It returns a pointer to the specific annotation object or null if the node isn't annotationed with that annotation or the annotation pointer argument is null.

dds/DCPS/Serializer.cpp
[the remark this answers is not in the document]
iguessthislldo: It was something Eric added. I'm reassessing code in the PR in general. I was probably a little hasty in asking for your in Adam reviews.

dds/idl/be_global.cpp
iguessthislldo: I will be fixing the style.

dds/DCPS/Serializer.cpp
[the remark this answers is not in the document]
simpsont-oci: Yeah, Adam mentioned it wasn't really ready yet. I just realized my input had been requested a while back and so I was taking a look. Let me know when you think it's ready to look again. :)

dds/idl/be_global.cpp
iguessthislldo: > What is going on with the spaces around "->" ?

The style for this one is already fixed in changes I haven't pushed.

> Also, do these need to be dynamic casts?

`defined_in` returns an `UTL_Scope`, which is not an `AST_Decl`, but a lot of `AST_Decl`s are `UTL_Scope`s, so yes, the dynamic cast is necessary.

> If so, should we be checking for failure?

If it cast to null it's because the parent node wasn't an `AST_Decl`, which can only mean we hit the root module, which is a case handled in `is_default_nested`.

As for error handling in general, these functions really shouldn't fail; if they do, something is wrong with the fundamental structure of the AST. In addition to that tao_idl is pointer ridden, basically every value that's not a primitive is a pointer. I might I have become numb to it, but I don't want to check every single pointer in a case like this:

```c++
AST_Annotation_Member::narrow_from_decl((*default_nested_apply)["value"])->value ()->ev()->u.bval;
```

If I can be highly certain that these pointers are not going to be null.

cmake/dds_idl_sources.cmake
iguessthislldo: Yeah, still not sure and I don't what to get off track on this so I'm just reverting back to what Mason wrote. I'm setting `--default-nested` in `OpenDDSConfig.cmake` where it can be overridden by an option.

iguessthislldo:
Okay I've done a final once over of the PR and I believe it's ready assuming that last commit got UnionTopic in order.

@simpsont-oci The main question I have for you is how the RapidJSON and V8 generators would handle forcing the use of Union topic types as I've done here. Most of the rest of the generators didn't really have a problem the switch as long as I didn't mess it up like I did in marshalling. I can do testing if you're not sure, I just didn't want this to take up anymore time than it already is to figure out a Union Topic example for both of them.

iguessthislldo:
@mmcparlane If you have time to look over my CMake changes, I made some additions to the cmake module for an new argument, `--default-nested` that opendds_idl should be given by default, but should be controllable globally and on a library by library basis . The idea is sometime in the feature we stop giving it by default I think. I'm not too confident about the changes though because I had to look at CMake documentation alot and even then had to experiment to get what I thought seemed right.

The other major change is removing the Messenger files in the cmake tests to reduce duplicating the code and have it reference the regular Messenger test. If there was a good reason for the files to be duplicated, I can restore them.

tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
```diff
@@ -36,19 +36,17 @@ foreach(t ${all_targets})
 endforeach()
 
 # Copy configs/scripts into build-output directory
-if (EXISTS ${DDS_ROOT}/tests/DCPS/Messenger)
```
mmcparlane: This if-block exists to enable these tests to work outside of the source tree (e.g. with a make-installed OpenDDS)—these should not be removed.

tests/cmake_integration/Messenger/Messenger_2/CMakeLists.txt
```diff
@@ -38,19 +39,17 @@ foreach(t ${all_targets})
 endforeach()
 
 # Copy configs/scripts into build-output directory
-if (EXISTS ${DDS_ROOT}/tests/DCPS/Messenger)
```
mmcparlane: This if-block exists to enable these tests to work outside of the source tree (e.g. with a make-installed OpenDDS)—these should not be removed.

tests/cmake_integration/Messenger/Args.h
```diff
@@ -1,98 +0,0 @@
-/*
```
mmcparlane: I recall attempting to remove the duplication in the CMake messenger tests. However, it was causing link issues to share the same code directory. I think that it was something to do with the export header being generated by Messenger_1 and consequently being included in Messenger_2 (thereby defining export macros when it should not). Feel free to test this theory further but I recommend reverting these changes and for now living with the duplication.

There is seemingly related link errors showing up on the [scoreboard](https://dev.azure.com/opendds/OpenDDS/_build/results?buildId=734&view=logs&jobId=567bf023-8cc0-5d40-5ba2-7f319774a15e&taskId=bdfa9db1-a1d0-5626-30bf-6f3cf19b26a7&lineStart=80&lineEnd=81&colStart=1&colEnd=1). 

It appears as though scoreboard giving a false-positive for the CMake tests—the link failures are not being flagged!

tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
iguessthislldo: If we're using this file which is in the tests directory, then why wouldn't we be able to access Messenger, even if OpenDDS itself is installed somewhere else?

iguessthislldo: To elaborate on that, I changed it so it's copying from a location relative to this `CMakeLists.txt`. So unless this file is being copied somewhere else, I don't see how this won't work.

mmcparlane: For example, in the [Example Using Installed OpenDDS](https://github.com/objectcomputing/OpenDDS/blob/master/docs/cmake.md#example-using-installed-opendds-unix-only) for using CMake outside of the source tree, a make-installed version does not have anything under ${DDS_ROOT}/tests, which is what the target created within that if-block depends upon.

iguessthislldo: Alright, yes, I would be breaking that particular example. However, why copy `tests/cmake_integration/Messenger` somewhere else and not just do it there?

mmcparlane: I meant to say that the target created within the if-block depends upon ${DDS_ROOT}/tests/DCPS/Messenger. So, in order to make these run outside of the source tree the following would have to be copied:

* ${DDS_ROOT}/tests/DCPS/Messenger
* ${DDS_ROOT}/tests/cmake_integration/Messenger/Messenger_1
* Other perl dependencies to run things

Given that, the primary purpose is to test that the CMake build functions are working as expected (outside the source tree), not whether or not the perl scripts run them—for that they are being tested inside the source tree. It's not perfect, but it works to to the manual test!

mmcparlane: I would recommend putting a comment above the if-block just to note what I had to explain, but other than that I'm not sure it's worth the time spent to remove these; otherwise the make-install scenario needs to be tested manually and and the documentation will have to be changed/tested for not much benefit.

iguessthislldo: Okay, but I don't think you answered my question: why does it have to copied outside the source tree at all? We're not doing that on the jenkins install tests, but is there something that the cmake module does differently inside the source tree vs outside it?

mmcparlane: Sorry Fred I didn't answer your previous question! I suppose it wouldn't have to be copied out of the source tree—although early on in development when the CMake package was a CMake module I think there was a clash of the DDS_ROOTs going on.

However, I think it still keeps with the separation logically—suppose I downloaded a pre-built OpenDDS but I grabbed the cmake_integration/Messenger/* tests to run the simple example, I would have to pull in all the other stuff too.

mmcparlane: > We're not doing that on the jenkins install tests, but is there something that the cmake module does differently inside the source tree vs outside it?

The CMake package doesn't do anything differently—it all depends on the CMAKE_PREFIX_PATH. But, as I stated before, copying this outside of the source root makes for the example in the documentation—which is a manual exercise to build outside of the source tree. This is separate from automated tests.

mmcparlane: > But, as I stated before, copying this outside of the source root makes for the example in the documentation—which is a manual exercise to build outside of the source tree. This is separate from automated tests.

This is a bit strange, building something that doesn't run. Perhaps a new example can be concocted? It would be nice not to have a dependency on the Perl testing libraries, so something more standalone would be ideal! Because really we're using a unit test in a way that it shouldn't really.

tests/cmake_integration/Messenger/Args.h
mmcparlane: > Feel free to test this theory further but I recommend reverting these changes and for now living with the duplication.

I just tested the branch locally on Windows and my theory is wrong! So, not sure what the problem is—but you are probably good to go on your changes.

mmcparlane: > I just tested the branch locally on Windows and my theory is wrong! So, not sure what the problem is—but you are probably good to go on your changes.

Sorry, I mistakenly built this using the master branch. I'm also seeing the same issues locally that the build server is reporting.

mmcparlane: Alright I found the issue—it's not actually the Messenger_1/2 projects which collide as I thought, but the problem is when `$DDS_ROOT/tests/DCPS/Messenger` is built in the standard MPC build, the generated artifacts (TypeSupport, etc) are clashing with the CMake-Build-Generated artifacts.

So, I think we're out of luck on this one for the moment. There may be a way to fix it, but it would require ensuring that the first directory searched for includes is the ones generated in (for example) `%DDS_ROOT%\tests\cmake_integration\Messenger\Messenger_1\build` prior to `%DDS_ROOT%\tests\DCPS\Messenger`—but it doesn't seem practical to try and enforce that.

docs/cmake.md
```diff
@@ -62,9 +62,11 @@ within a CMakeLists.txt file. In addition, the following cache variables/options
...
 
 ### Cache Variables/Options Understood by OpenDDS
-|Cache Variable           | Description                                                |
-|------------------------ | ---------------------------------------------------------- |
-|`OPENDDS_CMAKE_VERBOSE`  | Print detailed status information at CMake-Generation time |
+
+| Cache Variable              | Description                                                | Default |
+| --------------------------- | ---------------------------------------------------------- | ------- |
+| `OPENDDS_CMAKE_VERBOSE`     | Print detailed status information at CMake-Generation time | `OFF`   |
+| `OPENDDS_NESTED_DEFAULT`    | Require topic types to be declared explicitly.             | `ON`    |
```
mitza-oci: This is listed as `OPENDDS_DEFAULT_NESTED` below, which one is correct?

docs/cmake.md
```diff
@@ -193,6 +196,17 @@ target_link_libraries(publisher messenger OpenDDS::OpenDDS)
 *Note:* This may issue a warning in earlier version of CMake due to the messenger library
 not having any sources added with it in the call to `add_library`.
 
+### `DEFAULT_NESTED`
+
+`DEFAULT_NESTED` overrides the global default, `OPENDDS_DEFAULT_NESTED`, which
+is `ON` by default. If this value is `ON`, then topic types for a given project
+must be declared using annotations. If it's `OFF`, then every valid type is
+assumed to be needed for use as a topic type by default, which might needlessly
```
mitza-oci:
```suggestion
assumed to be needed for use as a topic type by default, which might
```
(redundant)

docs/cmake.md
```diff
@@ -174,7 +177,7 @@ OPENDDS_TARGET_SOURCES(publisher
 ```
 
 Another snippet, based upon [Messenger with auxiliary IDL library], showcases how an
-auxiliary IDL library can be created for inclusion by other executables:
+auxiliary (or dynamic) IDL library can be created for inclusion by other executables:
```
mitza-oci: I'm not sure I get what the intent is here.  The only distinction we're making is that IDL-generated code could be directly incorporated into a target that has user code, or it could be compiled into a separate library.  Whether or not that library is dynamic shouldn't be relevant.

examples/DCPS/DistributedContent/FileInfo.idl
```diff
@@ -1,10 +1,9 @@
 module DistributedContent {
-
-#pragma DCPS_DATA_TYPE "DistributedContent::FileDiff"
-#pragma DCPS_DATA_KEY "DistributedContent::FileDiff file_id"
-
   typedef sequence<octet> OctetSeq;
+
+  @topic
   struct FileDiff {
+    @key
```
mitza-oci: I think these `@key` annotations would be more readable if they were on the same line, since each line defines its own field.

tests/DCPS/Compiler/annotations/is_topic_type_test/run_test.pl
```diff
@@ -0,0 +1,84 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
+     & eval 'exec perl -S $0 $argv:q'
+     if 0;
+
+use strict;
+use JSON::PP;
```
mitza-oci: This doesn't seem like it will work with older Perl versions.  It would at least need a check to not cause an error if this module isn't found.

tests/DCPS/Compiler/annotations/classic_langmap/topic_annotations_test_lib/.gitignore
```diff
@@ -0,0 +1,2 @@
+/topic_annotations_test*
```
mitza-oci: This could be an issue when someone tries to add a file that happens to match the pattern.  I would prefer to have the .gitignore files generated by the script.  If there are reasons not to do that (or if the script needs to be changed), we can discuss other approaches.

tests/DCPS/Compiler/idl_test1_lib/FooDef.idl
```diff
@@ -53,11 +53,18 @@ module Xyz {
...
   // +12
   typedef sequence<short> ShortSeq;
+#ifdef USE_TOPIC_ANNOTATIONS
+  @topic
+#else
+#  pragma DCPS_DATA_TYPE "Xyz::AStruct"
+#  pragma DCPS_DATA_KEY "Xyz::AStruct v1"
+#endif
   struct AStruct {
+#ifdef USE_TOPIC_ANNOTATIONS
```
mitza-oci: An alternative here would be to define a preprocessor macro that expands to either `@key` or to nothing.

tests/DCPS/UnionTopic/UnionTopic.cpp
```diff
@@ -0,0 +1,460 @@
...
+  }
+
+  bool
+  setup_test(DDS::Topic_var& topic)
+  {
+    DDS::ReturnCode_t rc;
+
+    // Create Subscriber
+    DDS::SubscriberQos subscriber_qos;
+    rc = participant_->get_default_subscriber_qos(subscriber_qos);
```
mitza-oci: Why not use SUBSCRIBER_QOS_DEFAULT?

tests/DCPS/UnionTopic/UnionTopic.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    // Create DataReader
+    DDS::DataReaderQos reader_qos;
+    rc = subscriber_->get_default_datareader_qos(reader_qos);
+    if (rc != DDS::RETCODE_OK) {
+      ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT("%N:%l setup_test() ERROR: ")
+        ACE_TEXT("get_default_datareader_qos failed!\n")), false);
+    }
+    reader_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
```
mitza-oci: Should this reader be reliable?

tests/DCPS/UnionTopic/UnionTopic.cpp
```diff
@@ -0,0 +1,460 @@
...
+    writer_ = publisher_->create_datawriter(
+      topic.in(), DATAWRITER_QOS_DEFAULT,
+      DDS::DataWriterListener::_nil(),
+      OpenDDS::DCPS::DEFAULT_STATUS_MASK);
+    if (!writer_) {
+      ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT("%N:%l setup_test() ERROR: ")
+        ACE_TEXT("create_datawriter failed!\n")), true);
+    }
+
+    // Block until Subscriber is associated
```
mitza-oci:
```suggestion
```

dds/idl/be_produce.cpp
```diff
@@ -289,11 +289,22 @@ BE_produce()
...
+  bool warned_dcps_data_type =
+    be_global->no_dcps_data_type_warnings_ ||
+    !idl_global->print_warnings();
   while (idl) {
     idl.getline(buffer, buffer_sz);
-
-    if (0 == strncmp("#include", buffer, 8)) { //FUTURE: account for comments?
+    lineno++;
+
+    if (!(warned_dcps_data_type || strncmp("#pragma DCPS_DATA_TYPE", buffer, 22))) {
```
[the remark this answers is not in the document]
iguessthislldo: Yes, absolutely this could be improved, I just didn't think they would have to go into this release. I have some ideas about having tao_idl help with this that are going to be part of the annotation migrator. If you want me to make this more flexible for this release I can.

tests/DCPS/Compiler/annotations/is_topic_type_test/run_test.pl
iguessthislldo: Okay, do you want it to do a false pass if the module isn't available?

tests/DCPS/Compiler/annotations/classic_langmap/topic_annotations_test_lib/.gitignore
iguessthislldo: The thing I don't like about the script is that it relies getting the correct files indexed before running it, so it's not really very different from manually creating the gitignore. Would it be possible to have MPC generate .gitignore files?

tests/DCPS/UnionTopic/UnionTopic.cpp
iguessthislldo: This test is based on Presentation, so that's an oversight on my part.

dds/idl/topic_keys.cpp
```diff
@@ -0,0 +1,447 @@
...
+      current_value_ = root_;
+      return *this;
+    }
+
+  // If we are an array, use the base type and repeat for every element
+  } else if (root_type_ == ArrayType) {
+    AST_Array* array_node = dynamic_cast<AST_Array*>(root_);
+    ACE_CDR::ULong array_dimension_count = array_node->n_dims();
+    if (array_dimension_count > 1) {
+      throw Error(root_, "using multidimensional arrays as keys is unsupported.");
```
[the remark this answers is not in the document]
iguessthislldo: I don't think the spec says anything about this. This is something I asked you about at one point and you said to let it be an error.

dds/idl/metaclass_generator.cpp
```diff
@@ -368,143 +370,210 @@ namespace {
...
+      "  typedef " << clazz << " T;\n\n"
+      "#ifndef OPENDDS_NO_MULTI_TOPIC\n"
+      "  void* allocate() const { return new T; }\n\n"
+      "  void deallocate(void* stru) const { delete static_cast<T*>(stru); }\n\n"
+      "  size_t numDcpsKeys() const { return " << key_count << "; }\n\n"
+      "#endif /* OPENDDS_NO_MULTI_TOPIC */\n\n"
+      "  bool isDcpsKey(const char* field) const\n"
+      "  {\n";
+    {
+      if (struct_node && key_count) { // TODO: How to handle union disc?
```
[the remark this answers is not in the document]
iguessthislldo: I don't know how to have it address union fields. This was something I wanted to leave until we did union queries.

dds/idl/topic_keys.h
```diff
@@ -0,0 +1,191 @@
...
+  class Iterator {
+  public:
+    /**
+     * Standard Iterator Type Declarations
+     */
+    ///{
+    typedef AST_Decl* value_type;
+    typedef AST_Decl** pointer;
+    typedef AST_Decl*& reference;
+    typedef std::output_iterator_tag iterator_category;
```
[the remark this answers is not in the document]
iguessthislldo: My mistake, I thought output meant output in respect to the container, but it looks like it's the opposite.

docs/cmake.md
iguessthislldo: `OPENDDS_DEFAULT_NESTED` is correct, this is a typo.

examples/DCPS/DistributedContent/FileInfo.idl
iguessthislldo: I don't necessarily disagree, I'm just using the style given by the spec and by usage of decorators in Python.

dds/idl/be_global.cpp
```diff
@@ -602,3 +619,93 @@ BE_GlobalData::get_include_block(BE_GlobalData::stream_enum_t which)
...
+  return node->annotations().find(topic_annotation_) || !is_nested(node);
+}
+
+bool
+BE_GlobalData::is_nested(AST_Decl* node)
+{
+  // If annotated with @nested, then @nested.value is the type's nested value.
+  AST_Annotation_Appl* nested = node->annotations().find(nested_annotation_);
+  if (nested) {
+    return dynamic_cast<AST_Annotation_Member*>((*nested)["value"])->
```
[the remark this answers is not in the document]
iguessthislldo: This is related to [the conversation I already had with Tim](https://github.com/objectcomputing/OpenDDS/pull/1067#discussion_r296289905). The issue is that TAO_IDL uses pointers heavily, but I don't think any of these things would fail unless someone messed up the AST somehow. However covariant return types is new to me, but makes sense and I agree would be perfect for this situation.

dds/idl/topic_keys.h
```diff
@@ -0,0 +1,191 @@
+#ifndef SAMPLE_KEYS_HEADER
```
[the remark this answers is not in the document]
iguessthislldo: A small leftover from when I was still using `@sample_type` instead of `@topic`.

dds/idl/keys_generator.cpp
```diff
@@ -6,84 +6,142 @@
...
-  }
-
-  be_global->header_ << be_global->versioning_begin() << "\n";
+struct KeyLessThanWrapper {
+  size_t n_;
+  const string cxx_name_;
 
+  explicit KeyLessThanWrapper(UTL_ScopedName* name)
+    : n_(0),
+      cxx_name_(scoped(name))
```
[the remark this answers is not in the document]
iguessthislldo: I shouldn't have blindly apply this change...

tests/cmake_integration/Messenger/Args.h
iguessthislldo: How about we have cmake copy everything into `cmake_integration/Messenger_X`? Also for Perl dependency we can make a script that runs perl script that runs messenger without PerlDDS.

dds/DCPS/Serializer.cpp
```diff
@@ -49,10 +49,8 @@ Serializer::reset_alignment()
 void
 Serializer::smemcpy(char* to, const char* from, size_t n)
 {
-    (void) ACE_OS::memcpy(
-      reinterpret_cast<void*>(to),
-      reinterpret_cast<const void*>(from),
-      n);
+  OPENDDS_ASSERT(from);
```
mitza-oci: If !n, should this assert?  Does having an assert here really help?

dds/idl/topic_keys.cpp
```diff
@@ -0,0 +1,500 @@
...
+    AST_Field* field = *field_ptrptr;
+    ss << (level_ ? "." : "") << field->local_name()->get_string();
+  } else if (root_type_ == UnionType) {
+    // Nothing
+  } else if (root_type_ == ArrayType) {
+    // Figure out what the vector version of the scalar pos_ is
+    std::vector<size_t>::reverse_iterator di, dfinished = dimensions_.rend();
+    size_t acc = pos_;
+    size_t div = 1;
+    std::list<size_t> results;
```
mitza-oci: Does this need a `std::list`?  Looks like it always adds to one side and then iterates last-in first-out.  The equivalent with vector is push_back and reverse iterators.  Not that we're so performance-sensitive inside the compiler...

dds/idl/topic_keys.cpp
```diff
@@ -0,0 +1,500 @@
...
+{
+  std::stringstream ss;
+  path_i(ss);
+  return ss.str();
+}
+
+void
+TopicKeys::Iterator::path_i(std::stringstream& ss)
+{
+  if (root_type_ == StructureType) {;
```
```suggestion
  if (root_type_ == StructureType) {
```

dds/DCPS/Serializer.cpp
iguessthislldo: Eric ran into a problem with `from` being null from using a sequence inside a union. He changed it to throw an exception. Tim suggested I change it to an assert. I don't really care one way or another.

cmake/tao_idl_sources.cmake
```diff
@@ -182,7 +182,7 @@ function(tao_idl_command name)
       MAIN_DEPENDENCY ${idl_file_path}
       COMMAND ${CMAKE_COMMAND} -E env "DDS_ROOT=${DDS_ROOT}"  "TAO_ROOT=${TAO_INCLUDE_DIR}"
         "${_tao_extra_lib_dirs}"
-        $<TARGET_FILE:tao_idl> -g ${GPERF_LOCATION} ${TAO_CORBA_IDL_FLAGS} -Sg -Wb,pre_include=ace/pre.h -Wb,post_include=ace/post.h -I${TAO_INCLUDE_DIR} -I${_working_source_dir} ${_converted_flags} ${idl_file_path}
+        $<TARGET_FILE:tao_idl> -g ${GPERF_LOCATION} ${TAO_CORBA_IDL_FLAGS} -Sg -Wb,pre_include=ace/pre.h -Wb,post_include=ace/post.h --idl-version 4 --unknown-annotations ignore -I${TAO_INCLUDE_DIR} -I${_working_source_dir} ${_converted_flags} ${idl_file_path}
```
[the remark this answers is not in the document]
iguessthislldo: There are not any that I can conceive of. Removing `--unknown-annotations` would just produce harmless warnings, but `--idl-version` is critical.

cmake/init.cmake
```diff
@@ -8,7 +8,8 @@ file(TO_CMAKE_PATH "${OPENDDS_ACE}" OPENDDS_ACE)
 file(TO_CMAKE_PATH "${OPENDDS_MPC}" OPENDDS_MPC)
 file(TO_CMAKE_PATH "${OPENDDS_TAO}" OPENDDS_TAO)
 
-option(OPENDDS_CMAKE_VERBOSE "Print verbose output when loading the OpenDDS Config Package")
+option(OPENDDS_CMAKE_VERBOSE "Print verbose output when loading the OpenDDS Config Package" OFF)
+option(OPENDDS_DEFAULT_NESTED "Require topic types to be declared explicitly" ON)
```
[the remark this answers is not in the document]
iguessthislldo: Yes, it's intended. I didn't think about the command line, I was thinking of changing in the CMakeLists file, but yes, default nested should be able to be changed on a project by project basis.

cmake/api_macros.cmake
```diff
@@ -155,7 +163,8 @@ macro(OPENDDS_TARGET_SOURCES target)
         TAO_IDL_FLAGS ${_tao_options} ${OPENDDS_TAO_BASE_IDL_FLAGS}
         DDS_IDL_FLAGS ${_opendds_options} ${OPENDDS_DDS_BASE_IDL_FLAGS}
         IDL_FILES ${_idl_sources_${scope}}
-        SCOPE ${scope})
+        SCOPE ${scope}
+        DEFAULT_NESTED ${_default_nested})
```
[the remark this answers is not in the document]
iguessthislldo: Here I wanted to make a finer control for default nested: It uses `OPENDDS_DEFAULT_NESTED` only if `OPENDDS_TARGET_SOURCES` wasn't passed `DEFAULT_NESTED`. Also it appears `_tao_options` is for `tao_idl`, while `--default-nested` is for `opendds_idl` only, so I'm assuming you mean `_opendds_options`?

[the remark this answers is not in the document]
iguessthislldo: Here I wanted to make a finer control for default nested: It uses `OPENDDS_DEFAULT_NESTED` only if `OPENDDS_TARGET_SOURCES` wasn't passed `DEFAULT_NESTED`. Also it appears `_tao_options` is for `tao_idl`, while `--default-nested` is for `opendds_idl` only, so I'm assuming you mean `_opendds_options`?

dds/idl/topic_keys.cpp
iguessthislldo: Yes I can change that. I guess I automatically thought list when I had to prepend elements.

cmake/tao_idl_sources.cmake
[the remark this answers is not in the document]
iguessthislldo: Thinking about it more, it might be possible in the future a user might want to change the version, but for now there's no reason to use 3 over 4.

cmake/init.cmake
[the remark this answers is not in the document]
mmcparlane: The general design centers around the idea that if a configuration option can be generated by the configure script then it should probably be handled like the `config.cmake`, but if it is something CMake specific (for instance like the verbose output for generating a cmake project) then it should be put in a cache variable.

So, let me rephrase my question. Is the DEFAULT_NESTED intended to be set at all in the configure script? For instance in MPC builds? If so, then I don't think it should be a CMake cache variable.

cmake/api_macros.cmake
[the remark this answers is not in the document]
mmcparlane: > so I'm assuming you mean _opendds_options?

Yes, _opendds_options I mean 😊

tests/cmake_integration/Messenger/Args.h
mmcparlane: Ah! Good idea.

cmake/init.cmake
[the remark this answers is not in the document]
iguessthislldo: No, it's not meant to be set by the configure script, it's meant to be set on a project by project basis or left to the default value.

iguessthislldo:
@mitza-oci It looks like the windows builds on Azure are working now, but Travis is being pushed over 50 minutes...

iguessthislldo:
If we get rid of the duplicate compiler tests for annotations, which is going to happen at some point anyway and combine this with #1211, it might bring Travis back down to under 50 minutes.

iguessthislldo:
@mitza-oci  Do you know how the Java API generator knows that a type needs typesupport? I thought just enabling IDL4 in `idl2jni_codegen` would make the java tests work with annotations, but they're missing Typesupport, so it's still relying on `DCPS_DATA_TYPE` somehow. I looked for checks for `DCPS_DATA_TYPE`, but that only exists in `ts_generator.cpp` java generator, which is also checking for annotations, so I'm at a loss.

mitza-oci:
> @mitza-oci Do you know how the Java API generator knows that a type needs typesupport? I thought just enabling IDL4 in `idl2jni_codegen` would make the java tests work with annotations, but they're missing Typesupport, so it's still relying on `DCPS_DATA_TYPE` somehow. I looked for checks for `DCPS_DATA_TYPE`, but that only exists in `ts_generator.cpp` java generator, which is also checking for annotations, so I'm at a loss.

First `opendds_idl` takes the user's IDL and generates `FooTypeSupport.idl`.  Then `idl2jni` takes `FooTypeSupport.idl` and generates Java bindings.  It doesn't need to know about which types are pragma'd/annotated.

iguessthislldo:
That's unfortunate, especially since it looked like at least one of them was almost done.

PR #1069, opened 2019-02-25
ICE for RTPS

jrw972:
Problem: Participants cannot communicate directly when behing NATs

Participants that are behind NATs cannot communicate directly.  The
RtpsRelay solves the problem of enabling communication.  However, all
of the data transferred between the participants must pass through the
relay.  This solution provides an optimization where participants can
communicate directly after bootstrapping discovery with the RtpsRelay.

Solution: Implement ICE for RTPS.  ICE involves exchanging candidate
information and then performing connectivity checks to determine a
pair of candidates that can be used to exchange data.  Candidate
information is exchange via SPDP; presumably by using the RtpsRelay.
ICE is then attempted for SEDP.  Similarly, SEDP is used to exchange
candidate information for the data transport.  ICE is then attempted
for the data transports.

### DIFF
--- a/dds/DCPS/ICE/Ice.cpp
+++ b/dds/DCPS/ICE/Ice.cpp
@@ -1,0 +1,90 @@
+/*
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include "DCPS/DdsDcps_pch.h" //Only the _pch include should start with DCPS/
+#include "Ice.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace ICE {
+
+#if OPENDDS_SECURITY
+
+bool
+candidates_sorted(const Candidate& x, const Candidate& y)
+{
+  if (x.address != y.address) {
+    return x.address < y.address;
+  }
+
+  if (x.base != y.base) {
+    return x.base < y.base;
+  }
+
+  return x.priority > y.priority;
+}
+
+bool candidates_equal(const Candidate& x, const Candidate& y)
+{
+  return x.address == y.address && x.base == y.base;
+}
+
+Candidate make_host_candidate(const ACE_INET_Addr& address)
+{
+  Candidate candidate;
+  candidate.address = address;
+  candidate.foundation = std::string("H") + std::string(address.get_host_addr()) + "U";
+  // See https://tools.ietf.org/html/rfc8445#section-5.1.2.1 for an explanation of the formula below.
+  candidate.priority = (126 << 24) + (65535 << 8) + ((256 - 1) << 0); // No local preference, component 1.
+  candidate.type = HOST;
+  candidate.base = address;
+  return candidate;
+}
+
+Candidate
+make_server_reflexive_candidate(const ACE_INET_Addr& address, const ACE_INET_Addr& base, const ACE_INET_Addr& server_address)
+{
+  Candidate candidate;
+  candidate.address = address;
+  candidate.foundation = std::string("S") + std::string(base.get_host_addr()) + "_" + std::string(server_address.get_host_addr()) + "U";
+  // See https://tools.ietf.org/html/rfc8445#section-5.1.2.1 for an explanation of the formula below.
+  candidate.priority = (100 << 24) + (65535 << 8) + ((256 - 1) << 0); // No local preference, component 1.
+  candidate.type = SERVER_REFLEXIVE;
+  candidate.base = base;
+  return candidate;
+}
+
+Candidate
+make_peer_reflexive_candidate(const ACE_INET_Addr& address, const ACE_INET_Addr& base, const ACE_INET_Addr& server_address, uint32_t priority)
+{
+  Candidate candidate;
+  candidate.address = address;
+  candidate.foundation = std::string("P") + std::string(base.get_host_addr()) + "_" + std::string(server_address.get_host_addr()) + "U";
+  candidate.priority = priority;
+  candidate.type = PEER_REFLEXIVE;
+  candidate.base = base;
+  return candidate;
+}
+
+Candidate
+make_peer_reflexive_candidate(const ACE_INET_Addr& address, uint32_t priority, size_t q)
+{
+  Candidate candidate;
+  candidate.address = address;
+  candidate.foundation = std::string("Q") + stringify(q) + "U";
+  candidate.priority = priority;
+  candidate.type = PEER_REFLEXIVE;
+  return candidate;
+}
+
+#endif /* OPENDDS_SECURITY */
+
+} // namespace ICE
+} // namespace OpenDDS
+
+OPENDDS_END_VERSIONED_NAMESPACE_DECL
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -260,6 +260,10 @@ Sedp::Sedp(const RepoId& participant_id, Spdp& owner, ACE_Thread_Mutex& lock) :
 
   automatic_liveliness_seq_ (DCPS::SequenceNumber::SEQUENCENUMBER_UNKNOWN()),
   manual_liveliness_seq_ (DCPS::SequenceNumber::SEQUENCENUMBER_UNKNOWN())
+#ifdef OPENDDS_SECURITY
+  , publication_agent_info_listener_(*this)
+  , subscription_agent_info_listener_(*this)
+#endif // OPENDDS_SECURITY
 {
   pub_bit_key_.value[0] = pub_bit_key_.value[1] = pub_bit_key_.value[2] = 0;
   sub_bit_key_.value[0] = sub_bit_key_.value[1] = sub_bit_key_.value[2] = 0;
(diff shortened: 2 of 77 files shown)


MPC/config/dcps_ice.mpb
```diff
@@ -0,0 +1,4 @@
+project: dcps, openssl {
```
iguessthislldo: This is going to have to be `ace_openssl` because of [ACE_TAO#858](https://github.com/DOCGroup/ACE_TAO/pull/858) if this is going to work on Android properly.

jrw972: We will pull out the openssl dependency and put it in the participant.

dds/DCPS/RTPS/Spdp.h
```diff
@@ -1,3 +1,5 @@
+
+
```
jrw972: Whitespace.

tests/DCPS/Messenger/rtps_disc.ini
```diff
@@ -7,7 +7,9 @@ DiscoveryConfig=uni_rtps
 [rtps_discovery/uni_rtps]
 SedpMulticast=0
 ResendPeriod=2
+SedpStunServerAddress=10.201.200.156:3478
```
jrw972: Remove.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2492,6 +2548,13 @@ Sedp::signal_liveliness_secure(DDS::LivelinessQosPolicyKind kind)
 }
 #endif
 
+ICE::Endpoint* Sedp::get_ice_endpoint() {
+  DCPS::RtpsUdpInst_rch rtps_inst =
+    DCPS::static_rchandle_cast<DCPS::RtpsUdpInst>(transport_inst_);
```
mitza-oci: Is this cast still needed?  get_ice_endpoint is defined in `TransportImpl`.

jrw972: The impl() method is protected in the base but public in the derived class.

mitza-oci: Instead of depending on the general-purpose `impl()`, how about adding a method to `TransportInst` itself to support this use?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1884,6 +1978,278 @@ Spdp::lookup_participant_auth_state(const DCPS::RepoId& id) const
...
+    r.entityId = ENTITYID_SEDP_BUILTIN_SUBSCRIPTIONS_WRITER;
+    ICE::Agent::instance()->start_ice(endpoint, l, r, agent_info);
+  }
+  if (avail & BUILTIN_ENDPOINT_PARTICIPANT_MESSAGE_DATA_WRITER) {
+    l.entityId = ENTITYID_P2P_BUILTIN_PARTICIPANT_MESSAGE_READER;
+    r.entityId = ENTITYID_P2P_BUILTIN_PARTICIPANT_MESSAGE_WRITER;
+    ICE::Agent::instance()->start_ice(endpoint, l, r, agent_info);
+  }
+
+#ifdef OPENDDS_SECURITY
```
mitza-oci: This is nested within the other preprocessor conditional, probably not what was intended.  We should also use if vs. ifdef consistently.

dds/DCPS/security/UtilityImpl.cpp
```diff
@@ -0,0 +1,40 @@
...
+  if (rc != 1) {
+    unsigned long err = ERR_get_error();
+    char msg[256] = { 0 };
+    ERR_error_string_n(err, msg, sizeof(msg));
+    ACE_ERROR((LM_ERROR,
+               ACE_TEXT("(%P|%t) Service_Participant::generate_random_bytes: ERROR '%C' returned by RAND_bytes(...)\n"),
+               msg));
+  }
+
+  // TODO:  Error reporting and return.
```
mitza-oci: TODO (here and line 34)

dds/DCPS/transport/framework/TransportClient.cpp
```diff
@@ -688,6 +688,21 @@ TransportClient::unregister_for_writer(const RepoId& participant,
...
+{
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, lock_, 0);
+  for (ImplsType::iterator pos = impls_.begin(), limit = impls_.end();
+       pos != limit;
+       ++pos) {
+    ICE::Endpoint* endpoint = (*pos)->get_ice_endpoint();
+    if (endpoint) { return endpoint; }
+  }
+
+  // TODO(jrw972):  This should return a set.  How would the semantics change?
```
mitza-oci: TODO here

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2948,6 +2917,78 @@ OpenDDS::DCPS::RtpsUdpDataLink::receive_strategy()
...
+    for (CORBA::ULong i = 0; i < peers->length(); ++i) {
+      get_addresses_i(local, peers[i], retval);
+    }
+  }
+
+  return retval;
+}
+
+void
+OpenDDS::DCPS::RtpsUdpDataLink::get_addresses_i(const RepoId& local, const RepoId& remote,
```
mitza-oci: Should this be called `accumulate_addresses`?

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -290,6 +314,17 @@ RtpsUdpTransport::configure_i(RtpsUdpInst& config)
                       false);
   }
 
+#if defined (ACE_RECVPKTINFO)
+  int sockopt = 1;
+  if (unicast_socket_.set_option(IPPROTO_IP, ACE_RECVPKTINFO, &sockopt, sizeof sockopt) == -1) {
+    ACE_ERROR_RETURN((LM_ERROR,
+                      ACE_TEXT("(%P|%t) ERROR: ")
+                      ACE_TEXT("RtpsUdpTransport::configure_i: setcokopt:")
```
mitza-oci: Typo here, perhaps should just be `set_option`

dds/DCPS/ICE/Ice.cpp
```diff
@@ -0,0 +1,88 @@
...
+bool candidates_equal(const Candidate& x, const Candidate& y)
+{
+  return x.address == y.address && x.base == y.base;
+}
+
+Candidate make_host_candidate(const ACE_INET_Addr& address)
+{
+  Candidate candidate;
+  candidate.address = address;
+  candidate.foundation += "H" + std::string(address.get_host_addr()) + "U";
```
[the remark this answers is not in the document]
jrw972: The += causes the type of the RHS to start as a string instead of a const char* which generates an error with +.

dds/DCPS/ICE/Ice.cpp
```diff
@@ -0,0 +1,88 @@
...
+  candidate.priority = (126 << 24) + (65535 << 8) + ((256 - 1) << 0); // No local preference, component 1.
```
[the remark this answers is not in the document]
jrw972: The expressions come from the ICE RFC.  I would prefer to leave them as is so they can be identified by the next maintainer.

dds/DCPS/RTPS/Spdp.cpp
jrw972: Fixed.  We should document the preferred if/ifdef/if defined style in the guidelines.

dds/DCPS/security/UtilityImpl.cpp
jrw972: Done.

dds/DCPS/transport/framework/TransportClient.cpp
jrw972: Fix.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: Done.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
jrw972: Fixed.

dds/DCPS/ICE/Ice.cpp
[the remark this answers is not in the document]
mitza-oci: Interesting... in this case there is `operator+(const char*, std::string)` which shouldn't be a problem.  Is it worth the head-scratching from others reading the code?

dds/DCPS/ICE/Ice.cpp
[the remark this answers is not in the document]
mitza-oci: OK, please add comment(s).

dds/DCPS/RTPS/Spdp.cpp
mitza-oci: I was thinking that the existing usage was the de facto guideline.

As usual my style preference is for brevity, unless that brevity can cause confusion.  The language allows one to write:
```
#if defined(MACRO)
```
but that's exactly equivalent to:
```
#ifdef MACRO
```

So the next step is if we should just use `#if MACRO`, which does the same thing unless something really strange occurs like `-DMACRO=false`.  In this case I think we're following precedent in other ACE/TAO/OpenDDS code and I think the `def/defined` keyword is a good logical link to the form of `-D` without an `=` "define this macro, but I don't care what value it has (1)".

dds/DCPS/ICE/Ice.cpp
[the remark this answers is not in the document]
jrw972: I think I'll just construct the string explicitly.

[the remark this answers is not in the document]
jrw972: Done.

dds/DCPS/ICE/Ice.cpp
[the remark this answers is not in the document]
jrw972: Done.

dds/DCPS/RTPS/Spdp.cpp
jrw972: I'm good with #ifdef.

PR #1073, opened 2019-03-08
According to the azure pipelines github repo perl should be in the pa…

jwillemsen:
…th on vs2017

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,12 +34,6 @@ jobs:
       filename: '"C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat"'
       arguments: $(VcVarsArch)
       modifyEnvironment: true
-  - task: BatchScript@1
-    displayName: Put perl on PATH
-    inputs:
-      filename: tools\scripts\prepend_path
-      arguments: 'C:\Strawberry\perl\bin'
-      modifyEnvironment: true
   - script: |
       if %BuildConfiguration%==Release set OPT=--no-tests --no-debug --optimize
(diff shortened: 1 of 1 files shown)


jwillemsen:
See https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Validate-Perl.ps1

jwillemsen:
Ok, looks there is a perl but it seems one the configure script doesn't support yet
```
##[command]"C:\windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\aa02a820-cb01-4190-9c92-7d918ab219c5.cmd""
ERROR: unknown host msys, stopped at configure line 348.
Options:
```

mitza-oci:
There is some Perl but not a Windows Perl (ActiveState, Strawberry, etc.).  I don't think we're going to support MSys since that's a very specialized use case.

jwillemsen:
Looking at https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/scripts/Installers/Install-Perl.ps1 I would expect it to be strawberry perl but it seems it is something different

mitza-oci:
I tried this when I was setting up azure-pipelines.yml.  I even looked into supporting msys but it uses nonstandard paths.  Seems like a bug in azure's image or their docs.

jwillemsen:
Let us retry when they ship a new image to azure

mitza-oci:
Should this be an issue in the Azure CI repo?

jwillemsen:
Yes, we should report it there also

mitza-oci:
Is there an issue we can link to?

jwillemsen:
Reported this upstream as https://github.com/Microsoft/azure-pipelines-image-generation/issues/822

PR #1075, opened 2019-03-13
Removed obsolete id tags and let the dds_fuzz script for these in the future

jwillemsen:


mitza-oci:
The developer may chose to run generate_export_file.pl from any supported version of ACE.  Although I'd rather not have it in the source code, there is (virtually) no cost to `$Id$` in a file that nobody will ever edit.

Issue #1005, opened 2018-12-07
After restarting OpenDDS, calling TransportRegistry::create_inst fails 


kmarcey:
I am working on integrating OpenDDS into an existing library as an additional protocol to interface with our product. Because of the way this library is used, a user may have several instances of it instantiated at once and the instances may be created and destroyed at any time.

I maintain a reference count of current instances using DDS and wait until the last one is destroyed before calling things like Service_Participant::shutdown. However if at some later time a user creates a new instance, I restart OpenDDS and attempt to redo the transport configuration, recreate the DomainParticipant, etc. Currently, using the 3.13 release this fails when I call TransportRegistry::create_inst. The call returns null and I see a message "TransportRegistry::create_inst: transport_type=rtps_udp is not registered." in the log. 

If I run my test client with the ACE_DEBUG=1 environment variable set, I see OpenDDS_Rtps_Udpd.dll being opened, seemingly successfully and I don't see any other obvious errors being reported.

I have attached the output when ACE_DEBUG=1 in case it is of any use. There are a few comments describing when I initialize OpenDDS for the first time, when I call Service_Participant::shutdown, and when I try to restart OpenDDS.

It seems that this might be the same issue described in #991 but I wasn't able to find any existing bug reports on this.

Is there anything special I need to do when restarting OpenDDS versus when I start it for the first time?

[ACE_DEBUG.txt](https://github.com/objectcomputing/OpenDDS/files/2658148/ACE_DEBUG.txt)

mitza-oci:
Fixed in #1081

PR #991, opened 2018-11-23
Proposal test to use CORBA to create/destroy participant

dczanella:
This test uses a CORBA interface start() to configure the OpenDDS in RTPS mode and create a participant. Following it uses a stop() procedure to destroy the participant and shutdown domain participant factory. Running a pair start()/stop() at first everything is OK, but at second time it crashes with a segmentation fault.

Iteration 0
DDS Start
(3640|140604325148448) NOTICE: not using file configuration - no configuration file specified.
(3640|140604325148448) NOTICE: Service_Participant::intializeScheduling() - no scheduling policy specified, not setting policy.
(3640|140604325148448) Service_Participant::set_repo_domain: Domain[ 4] = Repo[ DEFAULT_RTPS].
(3640|140604325148448) Service_Participant::get_discovery: returning repository for domain 4, repo DEFAULT_RTPS.
(3640|140604325148448) Spdp::SpdpTransport::open_unicast_socket() - opened unicast socket on port 8410
(3640|140604325148448) Spdp::SpdpTransport::SpdpTransport joining group  239.255.0.1:8400
(3640|140604325148448) DirectPriorityMapper:thread_priority() - mapped TRANSPORT_PRIORITY value 0 to thread priority 0.
(3640|140604325148448) TransportReceiveStrategy-mb Cached_Allocator_With_Overflow 266e418 with 1000 chunks
(3640|140604325148448) TransportReceiveStrategy-db Cached_Allocator_With_Overflow 266e4b8 with 100 chunks
(3640|140604325148448) TransportReceiveStrategy-data Cached_Allocator_With_Overflow 266e558 with 100 chunks
(3640|140604325148448) TransportImpl::open()
   transport_type:               rtps_udp
   name:                         _OPENDDS__SEDPTransportInst_0103080027cda7380e3800004
   queue_messages_per_pool:      10
   queue_initial_pools:          5
   max_packet_size:              2147481599
   max_samples_per_packet:       10
   optimum_packet_size:          4096
   thread_per_connection:        false
   datalink_release_delay:       10000
   datalink_control_chunks:      32
   local_address:                0.0.0.0:42668
   use_multicast:                true
   multicast_group_address:      239.255.0.1:8402
   multicast_interface:          
   nak_depth:                    32
   nak_response_delay:           200
   heartbeat_period:             1000
   heartbeat_response_delay:     500
   handshake_timeout:            10000
[...]

PR #1081, opened 2019-03-27
Service Participant, Transport Registry, RTPS libs updated to support certain Restart scenarios

mitza-oci:
Related to #1005 and possibly #991 

- Support restarting the same Service_Participant object after shutdown() 
- Updated RtpsUdpLoader to have an easy way to reload the transport (from application)
- Sedp: remove transport when we're done 
- TransportRegistry::remove_inst() calls shutdown on the transport

### DIFF
--- a/tests/DCPS/Restart/RestartTest.cpp
+++ b/tests/DCPS/Restart/RestartTest.cpp
@@ -1,0 +1,101 @@
+#define ACE_DOESNT_DEFINE_MAIN
+
+#include "MessengerTypeSupportImpl.h"
+
+#include "dds/DCPS/Marked_Default_Qos.h"
+#include "dds/DCPS/Service_Participant.h"
+#include "dds/DCPS/transport/framework/TransportRegistry.h"
+#include "dds/DCPS/transport/rtps_udp/RtpsUdpLoader.h"
+#include "dds/DCPS/RTPS/RtpsDiscovery.h"
+
+#include "ace/Init_ACE.h"
+
+#include <cstdlib>
+#include <sstream>
+#include <string>
+
+struct Application {
+  Application()
+  {
+    ACE::init();
+    ++instances_;
+    OpenDDS::DCPS::RtpsUdpLoader::load();
+    OpenDDS::RTPS::RtpsDiscovery::StaticInitializer initialize_rtps;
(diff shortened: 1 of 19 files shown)


tests/DCPS/Restart/RestartTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+    participant_ = dpf->create_participant(0, PARTICIPANT_QOS_DEFAULT, 0, 0);
+    TheTransportRegistry->bind_config(cfg, participant_);
+    participant_->enable();
+
+    DDS::TypeSupport_var ts = new Messenger::MessageTypeSupportImpl;
+    CORBA::String_var tn = ts->get_type_name();
+    ts->register_type(participant_, tn);
+    DDS::Topic_var topic = participant_->create_topic("Topic", tn, TOPIC_QOS_DEFAULT, 0, 0);
+
+    DDS::Publisher_var pub = participant_->create_publisher(PUBLISHER_QOS_DEFAULT, 0, 0);
```
jrw972: Would it be valuable to have the publishers send messages as part of the test?
Should the test check for association?

tests/DCPS/Restart/RestartTest.cpp
```diff
@@ -0,0 +1,101 @@
...
+  {
+    Application a1;
+    {
+      Application a2;
+    }
+    {
+      Application a3;
+    }
+  }
+  Application a4;
```
jrw972: Would it be valuable to insert some sleeps to allow timers and other events to fire before exercising the shutdown code?

tests/DCPS/Restart/RestartTest.cpp
mitza-oci: That was not necessary to reproduce the problem

tests/DCPS/Restart/RestartTest.cpp
mitza-oci: In this case we're exercising the less-common path of getting to shutdown very quickly.

dczanella:
This merge fixed the issue:  [Proposal test to use CORBA to create/destroy participant #991](https://github.com/objectcomputing/OpenDDS/pull/991).
How can I close it?

PR #1087, opened 2019-04-05
configure: s/_/-/g in options

iguessthislldo:
Replace underscore with dashes/minuses in `--doc_group` and
`--wireshark_*` options in the configure script. Updated any mention of
these arguments I could find that. The old options are still valid via
an alias map that processes the arguments before argument parsing.

Also fixed some typos in the Wireshark configure logic.

### DIFF
--- a/configure
+++ b/configure
@@ -126,6 +126,18 @@ sub targetUsage {
 
 ## arg processing and usage
 
+my %aliases = (
+  '--doc_group' => '--doc-group',
+  '--wireshark_cmake' => '--wireshark-cmake',
+  '--wireshark_build' => '--wireshark-build',
+  '--wireshark_lib' => '--wireshark-lib',
+);
+for my $arg (@ARGV) {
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -126,6 +126,18 @@ sub targetUsage {
...
 ## arg processing and usage
 
+my %aliases = (
+  '--doc_group' => '--doc-group',
+  '--wireshark_cmake' => '--wireshark-cmake',
+  '--wireshark_build' => '--wireshark-build',
+  '--wireshark_lib' => '--wireshark-lib',
+);
+for my $arg (@ARGV) {
+  if (exists $aliases{$arg}) {
```
mitza-oci: Doesn't look like this will work when an element of `@ARGV` is `--wireshark_build=/foo/bar`?

iguessthislldo: Good point, I don't usually use that style so I didn't think of that case. A simple regex should cover that though.

PR #1091, opened 2019-04-09
Security Fixes

iguessthislldo:
- Fixes the following problems related to Key Exchange ("Builtin Participant Volatile Message Secure"):
  - It always happens in the clear
  - It has unreliable QOS
  - During participant association, if one participant is sufficiently slower than the one taking the lead in authentication, it will fail to receive the keys because it was not ready to. The solution to resend them later is nonstandard but works.
- Provide the correct error message when using different domains in the governance file and in the code.
- Fix Mix up of Builtin Crypto Handles
- When updating local Participant QOS after creation ("SPDP Reliable Builtin"/"DCPS" - "Participant Secure"):
  - Fix SPDP deadlocking on itself
  - Fix typo that caused transport to always not send it
- Added `-DCPSSecurityFakeEncryption` "Fake" Encryption which does everything in security up to encryption and decryption for debugging purposes.
- Added  new logging points  and make all the Security logging controllable via a debug level  (`-DCPSSecurityDebugLevel`) or categories (`-DCPSSecurityDebug`).

TODO:
- [x] Undo Submessage info check
- [x] `acknowledged_by_reader()` -> `first_acknowledged_by_reader()`
- [x] Purge keys from `RepeatOnceWriter` in `Sedp::disassociate()`


dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -318,16 +327,25 @@ RtpsUdpReceiveStrategy::deliver_from_secure(const RTPS::Submessage& submessage)
...
+               ACE_TEXT("preprocess_secure_submsg failed RPCH %d, [%d.%d]: %C\n"),
                peer_pch, ex.code, ex.minor_code, ex.message.in()));
     return;
   }
 
   if (!ok) {
-    ACE_DEBUG((LM_WARNING, "(%P|%t) RtpsUdpReceiveStrategy: "
-               "decode_datawriter/reader_submessage failed [%d.%d]: %C\n",
-               ex.code, ex.minor_code, ex.message.in()));
+    bool dw = category == DATAWRITER_SUBMESSAGE;
```
mitza-oci: Since the call to the plugin failed, this method shouldn't assume that `category` is initialized to something useful (8.5.1.9.6).

dds/DCPS/GuidUtils.cpp
```diff
@@ -37,12 +37,27 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
 namespace OpenDDS { namespace DCPS {
 
+OPENDDS_STRING
+to_string(const EntityId_t& entityId)
+{
+  OPENDDS_STRING ret;
+
+  std::size_t len = sizeof(entityId.entityKey);
+  for (std::size_t i = 0; i < len; ++i) {
+    ret += to_dds_string(unsigned(entityId.entityKey[i]), true);
```
simpsont-oci: I appreciate that is is how the code was beforehand, but do we really need 16 calls to sprintf (complete with parsing of format string) (which is what to_dds_string is doing here) for converting a GUID (or any kind of byte sequence) to a hex string? I'd rather see something like:

char hex_char(unsigned char in)
{
  return (in < 10) ? ('0' + static_cast<uint16_t>(in)) : ('A' + static_cast<uint16_t>(in - 10u));
}

std::string hex_string(const char* buffer, size_t size)
{
  std::string result(size * 2, ' ');
  while (size != 0)
  {
    --size;

    result[size * 2] = hex_char((static_cast<unsigned char>(buffer[size]) & 0xF0u) >> 4u);
    result[size * 2 + 1] = hex_char(static_cast<unsigned char>(buffer[size]) & 0x0Fu);
  }
  return result;
}

(taken from one of my own projects)

simpsont-oci: Not sure why it chose to format it that way, but you get the idea.

dds/DCPS/RTPS/SecurityHelpers.h
```diff
@@ -9,16 +9,20 @@
 #define OPENDDS_SECURITY_HELPERS_H
 
 #include "dds/DdsSecurityCoreC.h"
+#include "dds/DdsDcpsGuidC.h"
```
simpsont-oci: Was this required for something?

dds/DCPS/security/CommonUtilities.cpp
```diff
@@ -79,6 +79,78 @@ void set_security_error(DDS::Security::SecurityException& ex,
...
+OPENDDS_STRING to_dds_string(const KeyOctetSeq& keyData)
+{
+  size_t l = keyData.length() * 2;
+  // For Spaces
+  if (l > 2) {
+    l += (l / 2) - 1;
+  }
+  OPENDDS_STRING rv;
+  rv.reserve(l);
+  for (size_t i = 0; i < keyData.length(); i++) {
```
simpsont-oci: Several places in here where a single good generic "bytes_to_hex_string" function would serve us well (see comment above).

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -306,8 +306,6 @@ Spdp::write_secure_updates()
 {
   if (shutdown_flag_.value()) { return; }
 
-  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
```
simpsont-oci: Was this causing issues, or did it just get moved elsewhere?

dds/DCPS/RTPS/SecurityHelpers.h
iguessthislldo: I was doing something with this header in a more isolated context and it didn't have `EntityId_t`. I ended up scrapping what I was doing, but kept this because it's correct and someone might have the same problem in the future.

dds/DCPS/RTPS/Spdp.cpp
iguessthislldo: The same guard exists like one or two function calls up from this one. It was a guaranteed deadlock.

simpsont-oci: But most of the time it the code wasn't deadlocking before, so there must have been something unique to this code path that was causing the issue, right? Did we normally just never test this code path? Do other methods that call into this function that call this code still need the lock here? If so, this exposes them to race conditions. Perhaps the code "one or two function calls up" needs to unlock before getting here, or we need to reassess our entry points?

dds/DCPS/RTPS/SecurityHelpers.h
simpsont-oci: If by "correct" you mean "an unnecessary burden on compilation," then I agree. :)

dds/DCPS/RTPS/Spdp.cpp
mitza-oci: Changing QoS of an existing participant (with Secure Discovery) wasn't used before.

simpsont-oci: Gotcha. This was the answer I was looking for!

dds/DCPS/RTPS/SecurityHelpers.h
iguessthislldo: I mean we ARE using C++, the king of burdens when it comes to compilation

dds/DCPS/RTPS/Sedp.h
```diff
@@ -232,6 +236,9 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
     Msg(MsgType mt, DCPS::MessageId id, const DDS::Security::ParticipantGenericMessage* data)
       : type_(mt), id_(id), pgmdata_(data) {}
 #endif
+
+    static OPENDDS_STRING msgTypeToString(const MsgType type);
```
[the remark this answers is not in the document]
iguessthislldo: I don't think I understand, are you saying I shouldn't put the const in the function declaration?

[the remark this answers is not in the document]
mitza-oci: Yes, `const` doesn't belong there.  You can use it in the definition if you'd like but it doesn't impact the caller.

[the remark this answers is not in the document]
iguessthislldo: I was not aware of that, I assumed it would be part of the function signature, but I guess as long as it's not a reference it doesn't matter since it gets copied regardless.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
mitza-oci: I think my comment from last week here was wrong.  I thought it was dealing with the preprocess step, not the decode step.

dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1409,6 +1505,14 @@ bool CryptoBuiltInImpl::decrypt(const KeyMaterial& master, Session& sess,
     return false;
   }
 
+  if (security_debug.fake_encryption) {
+    out.length(n);
+    for (unsigned i = 0; i < n; i++) {
```
mitza-oci: Use the sequence's `get_buffer()` method to change it all at once.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2872,6 +2933,49 @@ Sedp::Writer::set_header_fields(DCPS::DataSampleHeader& dsh,
...
+}
+
+void
+Sedp::RepeatOnceWriter::acknowledged_by_reader(const RepoId& rdr)
+{
+  RepoId remote_participant;
+  RTPS::assign(remote_participant.guidPrefix, rdr.guidPrefix);
+  remote_participant.entityId = ENTITYID_PARTICIPANT;
+
+  if (!std::memcmp(repo_id_.guidPrefix, rdr.guidPrefix, sizeof(DCPS::GuidPrefix_t))) {
```
simpsont-oci: Should you move the ACE_Guard statement up here to protect repo_id_?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2872,6 +2933,49 @@ Sedp::Writer::set_header_fields(DCPS::DataSampleHeader& dsh,
...
+  }
+
+  if (DCPS::DCPS_debug_level > 3) {
+    ACE_DEBUG((LM_DEBUG, "(%P|%t) Sedp::RepeatOnceWriter::acknowledged_by_reader(): %C\n",
+      OPENDDS_STRING(DCPS::GuidConverter(rdr)).c_str()));
+  }
+
+  ACE_Guard<ACE_Thread_Mutex> g(lock_);
+  if (remote_readers_.count(remote_participant)) {
+    RemoteReaderVector& remote_readers = remote_readers_[remote_participant];
```
simpsont-oci: This does the lookup twice (count + []). Use find() to grab the iterator, check to see it's not equal to .end(), and then dereference it get the value. Same applies for line 2969.

dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1638,9 +1754,20 @@ bool CryptoBuiltInImpl::decode_serialized_payload(
...
+      sending_datawriter_crypto, receiving_datareader_crypto));
+  }
+
   ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
-  if (!writer_options_[sending_datawriter_crypto].payload_) {
+  if (!encrypt_options_[sending_datawriter_crypto].payload_) {
     plain_buffer = encoded_buffer;
+    if (security_debug.encdec) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) {encdec} CryptoBuiltInImpl::decode_serialized_payload ")
+        ACE_TEXT("Sending datawriter should not be encrypting, returning input as plaintext\n"),
```
mitza-oci: The phrase "should not be encrypting" could be read as "it did encrypt even though it shouldn't have".  Could you reword this?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -936,18 +950,61 @@ void Sedp::associate_secure_readers_to_writers(const Security::SPDPdiscoveredPar
...
+  RepeatOnceWriter::RemoteReader info;
+  info.local_writer = local_writer;
+  info.remote_reader = remote_reader;
+  DDS::Security::DatawriterCryptoTokenSeq& dwcts = info.writer_tokens;
+  create_datawriter_crypto_tokens(dwch, drch, dwcts);
+
+  send_datawriter_crypto_tokens(local_writer, remote_reader, dwcts);
+  ACE_Guard<ACE_Thread_Mutex> g(participant_volatile_message_secure_writer_.lock_);
+  participant_volatile_message_secure_writer_.
+    remote_readers_[remote_participant].push_back(info);
```
simpsont-oci: So if the ACK never shows up, do these entries just stick around forever? I'm guessing we'll need to clear these out on participant unregister / dispose or we might wind up with collisions on reconnect. Are we clearing these out somewhere?

dds/DCPS/RTPS/Sedp.cpp
simpsont-oci: Oh nevermind, it looks like repo_id_ is coming from TransportClient?

dds/DCPS/RTPS/Sedp.cpp
simpsont-oci: 3 times, if you count erase!

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: Looks like this check could move up so the previous block doesn't have to run if we're just going to return here.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -379,8 +386,55 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
...
+    };
+    typedef OPENDDS_VECTOR(RemoteReader) RemoteReaderVector;
+    typedef OPENDDS_MAP_CMP(
+      DCPS::RepoId, RemoteReaderVector, DCPS::GUID_tKeyLessThan) RemoteReaderVectors;
+    RemoteReaderVectors remote_readers_;
+
+    /// Lock for remote_readers_ and remote_writers_
+    ACE_Thread_Mutex lock_;
+
+  } participant_volatile_message_secure_writer_;
```
mitza-oci: This changes the declaration order of `participant_volatile_message_secure_writer_` relative to `dcps_participant_secure_writer_`, do constructor initializer lists need to change as well?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1697,6 +1697,8 @@ RtpsUdpDataLink::received(const RTPS::AckNackSubmessage& acknack,
     callbacks[i]->reader_exists(remote, local);
   }
 
+  acknowledged_by_reader(local, remote);
```
simpsont-oci: If this gets called every time we get an ACK, then are we guaranteed to always send the keys twice? Should we be storing the sequence numbers used when we sent the keys the first time and comparing against the sequence numbers sent in the ACK so we only resend when necessary?

dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -313,22 +316,42 @@ DatawriterCryptoHandle CryptoBuiltInImpl::register_local_datawriter(
     // (requirements for which key appears first, etc.)
     bool used_h = false;
     if (security_attributes.is_submessage_protected) {
-      push_back(keys,
-                make_key(h, plugin_attribs & FLAG_IS_SUBMESSAGE_ENCRYPTED));
+      KeyMaterial_AES_GCM_GMAC key = make_key(h, plugin_attribs & FLAG_IS_SUBMESSAGE_ENCRYPTED);
```
mitza-oci: Constify?

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: Some combination of TransportClient and Sedp::Endpoint.  Not sure if we really need both.  In any case it's logically const (although not actually const).

dds/DCPS/RTPS/Sedp.cpp
iguessthislldo: > So if the ACK never shows up, do these entries just stick around forever?

Yes

> I'm guessing we'll need to clear these out on participant unregister / dispose or we might wind up with collisions on reconnect. Are we clearing these out somewhere?

No. We don't do any clean up for keys on participant dispose right now. When we do, clearing any pending key resends would probably be a good idea.

dds/DCPS/RTPS/Sedp.h
iguessthislldo: Yes, I have already changed it.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
iguessthislldo: So this is starting to get a bit hazy on why exactly this is set up how it is, but I think that the key sender has no way to know if the key exchange was successful, because the volatile reader accepts the sequence number as is when it first encounters it. Resending the keys is a waste of resources, but it's the only way to ensure the keys are being exchanged short of some sort of fundamental change in the key exchange process works.

simpsont-oci: Ah, right, sorry, I'd forgotten the underlying issue. If we successfully got all the keys we'd ACK the final key, and if we'd never gotten any of them but saw that the latest heartbeat was past the final key, we'd ACK the final key in that situation too since that's the first data point we'd know about. I suppose if we wanted to avoid the wastefulness we could only ever "resend" the keys, but then there's additional delay waiting for an ACKNACK, right? Maybe for OpenDDS we can send additional heartbeats on the volatile reader? Could we even send additional "initial" heartbeats (high SN is set to zero?) when we want to send keys to a reader to prompt the faster ACKNACK reply?

iguessthislldo:
I want to run this in the original P2P scenario on the Pi before this is merged to make sure this still works as expected.

iguessthislldo:
I accidentally removed the `virtual` keyword on `TransportSendListener::first_acknowledged_by_reader` and couldn't figure out why it wasn't working Friday.  Now that's fixed and I successfully tested with the Pi.

PR #1094, opened 2019-04-16
Fixed listener invocation for Presentation QoS policy coherent_access

c4m3lc4s3:
Updated test case for coherent data transfer using an asynchronous reader to verify potential bug.
Please refer to earlier correspondence: https://sourceforge.net/p/opendds/mailman/message/36633810/


tests/DCPS/Presentation/main.cpp
```diff
@@ -223,7 +320,8 @@ ACE_TMAIN(int argc, ACE_TCHAR** argv)
 
       for (size_t i = 0; i < SAMPLES_PER_TEST; ++i)
       {
-        const Foo foo = {0, 0, 0, 0};
+        Foo foo {0, 0, 0, 0};
```
mitza-oci: This line is failing to compile on the CI builds that don't have C++11 enabled.

c4m3lc4s3: Ok, thanks. It was unintentional to leave the equal sign out. I corrected it, and pushed the update. The build now fails on some timeout:
_"The job exceeded the maximum time limit for jobs, and has been terminated."_

mitza-oci:
Travis often hits the 50 min limit, I restarted
Not sure what's wrong with Azure CI, will try to look into it
The text below should kick off a Jenkins build on our server, if not I'll try to do that manually

retest this please

iguessthislldo:
I'm reviewing the test right now and seeing if there's a fix. Currently I'm getting an error that "Unexpected number of samples taken: 101. Expected: 100.", is this consistent with what you're seeing @c4m3lc4s3?

iguessthislldo:
Actually I'm getting that error for the ordered access subtest, the coherent access subtest succeeds.

c4m3lc4s3:
Hi. Yes, I get the same results. However, I think there might be a slight misunderstanding. The test I added is indeed for the coherent data transfer, but since I added an asynchronous reader (in parallel with the synchronous reader as in the original test case) I can imagine that the "main" function finishes before the coherent data is received in the asynchronous reader.
And hence you see the printout: "Testing ordered_access..." before the asynchronous reader for the coherent transfer receives the messages, which eventually fails the test case since it receives 101 packets while it should only receive 100 packets. I think the test case can be improved to be less confusing so the printouts are easier to understand.

I hope it clarifies. Thanks for looking into it.

iguessthislldo:
Yeah, me and @mitza-oci figured out what it's doing, should have posted an update here, but thanks for confirming that's what it does on your end.

> I can imagine that the "main" function finishes before the coherent data is received in the asynchronous reader.

It doesn't quite do that, at least as far as I'm seeing on my machine. As is it's calling the listener (asynchronous reader) but only after the coherent test is done and it moved on to the ordered part of the test, which is where the extra sample is coming from. If I modify it to try to wait for the listener to be called before moving on to ordered part of the test, it never does.

> I think the test case can be improved to be less confusing so the printouts are easier to understand.

I already have some improvements in mind. Overall the test shouldn't be trying to test two features right after the other with the same reader and writer.

Thanks for giving us the test case.

c4m3lc4s3:
Oh, I see. I did a test, and you're right, the 101st element in the sequence contains the first sample from the ordered test. But it's still a bit strange since it's always exactly one sample from the ordered test present in the received data. I tried sending 100, 50 and 10 samples. So is this still intended? And why don't I receive all samples from the ordered test as well in the same sequence?
I kind of expected that coherent data would always be presented alone to the reader, i.e. without any samples outside the coherent set, but this is maybe not what I should expect?
I agree that separating the test cases into separate tests, or at least using different readers would be a good idea.
Anyway, thanks for digging deeper into this.

iguessthislldo:
Alright so it appears that when a topic scope coherent set is completed, data listeners aren't called. This was what already appeared to be happening, but I think it's possible this never worked. I changed `ReceivedDataStrategy.cpp` so that it calls listeners when topic scope coherent set is completed. I also rewrote a lot of the presentation test for style, to the isolate coherent and ordered tests more (they now only have participant and topic objects in common), and to more clearly report this bug.

Misc changes not really related to this PR:
 - Added `retcode_to_dds_string`
- Print why a `delete_datareader` failed in an error message.

I'm going to go ahead and push these changes to this PR.

c4m3lc4s3:
Ok, thanks for the update. I can actually confirm that I saw a more strange behavior when I tested this with my own test application. At that point I tested it on both Windows and Linux, and I recall I got quite different results from these two platforms. So what you're saying that this might never have worked might be true.
Thanks for the help!

tests/DCPS/Presentation/main.cpp
```diff
@@ -8,350 +10,584 @@
...
       }
+    }
 
-      if (writer->get_publication_matched_status(matches) != ::DDS::RETCODE_OK)
-      {
-        ACE_ERROR_RETURN((LM_ERROR,
-                          ACE_TEXT("%N:%l main()")
-                          ACE_TEXT(" ERROR: Failed to get publication match status!\n")), -1);
+    if (!CORBA::is_nil(reader2_)) {
+      reader2_->set_listener(0, 0);
```
mitza-oci: Does it need to remove a listener if the object is being deleted on the next line?

tests/DCPS/Presentation/main.cpp
```diff
@@ -8,350 +10,584 @@
...
+
+  ACE_DEBUG((LM_DEBUG, ACE_TEXT("Testing coherent_access...\n")));
+
+  DDS::Publisher_var publisher;
+  DDS::Subscriber_var subscriber;
+  DDS::DataReader_var reader1;
+  DDS::DataReader_var reader2;
+  DDS::DataWriter_var writer;
+  Test t(participant, publisher, subscriber, reader1, reader2, writer);
+  if (t.setup_test(topic, true)) {
```
mitza-oci: Does `setup_test` return `true` if it fails?  Should it?
(Look like this applies to a number of functions in the test)

tests/DCPS/Presentation/main.cpp
```diff
@@ -8,350 +10,584 @@
...
 
-        if (foo.length() != SAMPLES_PER_TEST)
-        {
-          ACE_ERROR_RETURN((LM_ERROR,
-                            ACE_TEXT("%N:%l main()")
-                            ACE_TEXT(" ERROR: Unexpected number of samples taken: %d\n"), foo.length()), -1);
-        }
-      }
+  FooDataReader_var reader1_i = FooDataReader::_narrow(reader1);
+  if (CORBA::is_nil(reader1_i)) {
```
mitza-oci:
```suggestion
  if (!reader1_i) {
```
As long as you're overhauling this test I'd change all the `is_nil`s to the simpler form.

tests/DCPS/Presentation/main.cpp
```diff
@@ -8,350 +10,584 @@
...
+ACE_TMAIN(int argc, ACE_TCHAR** argv)
+{
+  int test_failures = 0;
+  try {
+    DDS::ReturnCode_t rc;
+
+    DDS::DomainParticipantFactory_var dpf =
+      TheParticipantFactoryWithArgs(argc, argv);
+
+    // Create Participant
```
mitza-oci:
```suggestion
```

tests/DCPS/Presentation/main.cpp
iguessthislldo: I suppose not

dds/DCPS/ReceivedDataStrategy.cpp
```diff
@@ -103,7 +103,7 @@ void
 ReceivedDataStrategy::accept_coherent(PublicationId& writer,
                                       RepoId& publisher)
 {
-  CoherentFilter    filter = CoherentFilter(writer, publisher);
+  CoherentFilter filter = CoherentFilter(writer, publisher);
```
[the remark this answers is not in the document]
iguessthislldo: I didn't even notice that, I just saw the offending whitespace! :laughing:

dds/DCPS/SafetyProfileStreams.h
```diff
@@ -27,6 +28,8 @@ OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(long long to_convert);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned long long to_convert, bool as_hex = false);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned long to_convert, bool as_hex = false);
 
+OpenDDS_Dcps_Export OPENDDS_STRING retcode_to_dds_string(DDS::ReturnCode_t value);
```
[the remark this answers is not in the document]
iguessthislldo: I don't might the slightly shorter name on it's own, but I was keeping it in line with the rest of the functions.

tests/DCPS/Presentation/main.cpp
iguessthislldo: Yes, I think it should be an check for "did this produce an error", since errors are the exceptional case in this test. If it's a matter of returning true vs. false, using false for error means using `!` everywhere to check for an error case.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -2913,39 +2913,38 @@ DataReaderImpl::update_ownership_strength (const PublicationId& pub_id,
 #ifndef OPENDDS_NO_OBJECT_MODEL_PROFILE
 bool DataReaderImpl::verify_coherent_changes_completion (WriterInfo* writer)
 {
-  if (this->subqos_.presentation.access_scope == ::DDS::INSTANCE_PRESENTATION_QOS
-      || ! this->subqos_.presentation.coherent_access) {
-    this->accept_coherent (writer->writer_id_, writer->publisher_id_);
-    this->coherent_changes_completed (this);
-    return true;
-  }
+  bool completed = false;
```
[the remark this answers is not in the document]
iguessthislldo: Sure

tests/DCPS/Presentation/main.cpp
iguessthislldo: I was wondering if that worked, but I couldn't remember so I just went with what the test already had.

dds/DCPS/SafetyProfileStreams.h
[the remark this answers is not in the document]
iguessthislldo: *mind

tests/DCPS/Presentation/main.cpp
mitza-oci: I don't think we've ever used this pattern (but correct me if I'm wrong).  I'm pretty sure we've used the opposite.  As is, I think this is too confusing for someone reading the code.  Using `!` seems like the right thing to do, just like checking for a null pointer.

iguessthislldo: Could've sworn that's what the convention was in QueryCondition test, but I can see it's not. I will concede that's what the pattern is and will change it.

I still think the pattern is wrong though. Most of the time we're not expecting an error, so the pattern should be simpler than the opposite case.

dds/DCPS/SafetyProfileStreams.h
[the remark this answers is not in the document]
iguessthislldo: Should I change the other `*to_dds_string` functions in `SafetyProfileStreams.h` too?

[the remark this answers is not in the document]
mitza-oci: I don't think anything else needs to be changed.  I don't want to just have `to_string` and conflict with namespace `std`.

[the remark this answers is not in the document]
iguessthislldo: Huh, I wasn't aware of `std::to_strng`.

Issue #962, opened 2018-10-30
Visual Studio /clr leads to linking errors


floatingbit:
Hi,

I am not clear whether .NET is supported by OpenDDS. According to the article https://objectcomputing.com/resources/publications/mnb/using-tao-and-opendds-with-net-part-1, we should be able to set the mpc attribute "managed = 1" which gives me multiple linking errors with the following message:

unresolved token (0A000B4E) "private: void __cdecl OpenDDS::DCPS::DataReaderImpl::LivelinessTimer::check_liveliness_i(bool,class ACE_Time_Value const &)" (?check_liveliness_i@LivelinessTimer@DataReaderImpl@DCPS@OpenDDS@@$$FAEAAX_NAEBVACE_Time_Value@@@Z) referenced in function "public: virtual void __cdecl OpenDDS::DCPS::DataReaderImpl::LivelinessTimer::CheckLivelinessCommand::execute(void)" (?execute@CheckLivelinessCommand@LivelinessTimer@DataReaderImpl@DCPS@OpenDDS@@$$FUEAAXXZ)

Please advise as I am trying to access .NET /clr projects from OpenDDS VC++ projects.

mitza-oci:
The articles is from a long time ago, I think if you'd try an OpenDDS version from that time frame it would probably work.  Not that I'm recommending using an old OpenDDS.

Please take a look at https://github.com/jmmorato/openddsharp and see how this issue is handled there.

floatingbit:
I'll take a look, thank you for your feedback!

jmmorato:
Take a look to the Patches folder. 

Better if you use the tag v0.2.1 because in the current version also includes some fixes that has been already intengrated in OpenDDS. From tag v0.2.1, you will need the Patches DataReaderImpl.*.patch and unique_ptr.h.patch.

However, it would be nice if you give a try to the openddsharp project ;)

floatingbit:
@jmmorato Testing it out already :)

I am having trouble getting your project up and running. Any chance you can share a README with a basic outline of steps on how to use your library?

jmmorato:
This is the ticket I have to document the build project:
https://github.com/jmmorato/openddsharp/issues/13

There you can found basic steps to build the project.

I plan to correctly document it and test it in a clean machine, but right now better to use the Nuget packages, especially for the IDL code generation part. 

You can read a "getting started" tutorial here:
http://openddsharp.com/Documentation/GetStarted

[...]

PR #1096, opened 2019-04-18
Visual Studio /clr leads to linking errors

jmmorato:
Changes needed to avoid linking errors when compiling with C++/CLI.

Related Issue: #962

### DIFF
--- a/dds/DCPS/unique_ptr.h
+++ b/dds/DCPS/unique_ptr.h
@@ -132,7 +132,7 @@ public:
   }
 
 private:
-  unique_ptr(const unique_ptr&);
+  unique_ptr(const unique_ptr&) {};
   unique_ptr& operator=(const unique_ptr&);
 
   T* ptr_;
--- a/dds/DCPS/unique_ptr.h
+++ b/dds/DCPS/unique_ptr.h
@@ -132,7 +132,7 @@ public:
   }
 
 private:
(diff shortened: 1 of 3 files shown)


mitza-oci:
Is there a change you can make to azure-pipelines.yml to test this?

If we had that set up we could detect this problem before a release.  Though it would be much better to have all supported compilers+options understand the syntax of standard C++ (98, not even new stuff), I guess there's not much we can do in this case.

jmmorato:
We could implement in a separated repository the example explained on https://objectcomputing.com/resources/publications/mnb/using-tao-and-opendds-with-net-part-1 and try to build it at the end of the Visual Studio pipelines.

dds/DCPS/unique_ptr.h
```diff
@@ -132,7 +132,7 @@ class unique_ptr {
   }
 
 private:
-  unique_ptr(const unique_ptr&);
+  unique_ptr(const unique_ptr&) {};
```
mitza-oci: `unique_ptr` is not copyable, so the copy constructor is deliberately declared but not defined.

jmmorato: Seems that the modification is not needed anymore to link C++/CLI projects. I would say it was fixed on this commit 8b02ca2

PR #1097, opened 2019-04-22
IDL Generator Improvements for JSON (V8, RapidJSON)

simpsont-oci:
Includes fixes for v8 generation, including support for arrays. Also adds a rapidjson generator and adds support to opendds_idl for disabling default generators and for adding arbitrary includes to cpp files. This is to support additional typesupport generation for types that are already included in core library (for example adding rapidjson generation support for existing QoS objects).


iguessthislldo:
What is going on with `MPC/config/dcps_rapidjson.mpb`?

simpsont-oci:
Good catch. Should be fixed now. Even if I can't spell overwrite on the first try...

tests/idl/rapidjson_generator/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+RapidJsonTestS.cpp
+RapidJsonTestS.h
+RapidJsonTestTypeSupport.idl
+RapidJsonTestTypeSupportC.cpp
+RapidJsonTestTypeSupportC.h
+RapidJsonTestTypeSupportC.inl
+RapidJsonTestTypeSupportImpl.cpp
+RapidJsonTestTypeSupportImpl.h
+RapidJsonTestTypeSupportS.cpp
+RapidJsonTestTypeSupportS.h
```
iguessthislldo: You can simplify these with wildcards, [See here for an example](https://github.com/objectcomputing/OpenDDS/pull/1067/files#diff-6b7d8a364cd6f0d9c12a10c7d6b2cc70R1)

mitza-oci: If you use wildcards then you're writing them by hand.  Please use the script tools/scripts/update_gitignore.pl

dds/idl/metaclass_generator.cpp
```diff
@@ -40,14 +40,17 @@ bool metaclass_generator::gen_enum(AST_Enum*, UTL_ScopedName* name,
 {
   ContentSubscriptionGuard csg(!be_global->v8());
   NamespaceGuard ng;
-  std::string decl = "const char* gen_" + scoped_helper(name, "_") + "_names[]";
-  be_global->header_ << "extern " << decl << ";\n";
-  be_global->impl_ << decl << " = {\n";
+  std::string array_decl = "const char* gen_" + scoped_helper(name, "_") + "_names[]";
+  std::string size_decl = "const CORBA::ULong gen_" + scoped_helper(name, "_") + "_names_size";
```
mitza-oci: Is there a reason to use `CORBA::ULong` here?  It's not really tied to the IDL mapping, so could it just be `std::size_t` to match C++ `sizeof`?

dds/idl/be_util.cpp
```diff
@@ -64,6 +70,9 @@ be_util::prep_be_arg (char *arg)
   } else if (0 == ACE_OS::strncasecmp(arg, WB_PCH_INCLUDE, SZ_WB_PCH_INCLUDE)) {
     be_global->pch_include(arg + SZ_WB_PCH_INCLUDE);
 
+  } else if (0 == ACE_OS::strncasecmp(arg, WB_CPP_INCLUDE, SZ_WB_CPP_INCLUDE)) {
+    be_global->add_cpp_include(arg + SZ_WB_CPP_INCLUDE);
```
mitza-oci: I think we should add this to the usage message even if it's an "internal" option.  There is some precedent for that (see the text "special cases" in usage).

dds/idl/v8_generator.cpp
```diff
@@ -343,7 +343,7 @@ namespace {
           "  {\n"
           "    v8::Local<v8::Value> lv = " << src << "->Get(" << prop << ");\n"
           "    v8::Local<v8::Object> lo = Nan::To<v8::Object>(lv).ToLocalChecked();\n"
-          "    copyFromV8(lo, " << propName << ");\n"
+          "    copyFromV8(lo, " << (fun_assign ? propName + "()" : propName) << ");\n"
```
mitza-oci: `propName` is on both sides of the conditional, can it be moved out? (also on line 355)

tests/idl/rapidjson_generator/Main.cpp
```diff
@@ -0,0 +1,7 @@
+#include "gtest/gtest.h"
```
mitza-oci: GoogleTest is only available conditionally.  We will have to discuss if we want to change that.

dds/DCPS/RapidJsonTypeConverter.h
```diff
@@ -0,0 +1,32 @@
...
+namespace DCPS {
+  class RapidJsonTypeConverter {
+  public:
+    virtual ~RapidJsonTypeConverter() {}
+
+    virtual void toRapidJson(const void* source, rapidjson::Value& dst, rapidjson::Value::AllocatorType& alloc) const = 0;
+    virtual void* fromRapidJson(const rapidjson::Value& source) const = 0;
+    virtual void deleteFromRapidJsonResult(void* obj) const = 0;
+
+    virtual DDS::InstanceHandle_t register_instance_helper(DDS::DataWriter* dw, const void* data) const = 0;
```
mitza-oci: Is the RapidJson support (in this iteration) going to support read/write operations like V8 does, or only the convert to/from operations?

tests/idl/rapidjson_generator/Main.cpp
simpsont-oci: I don't really know that we need this test to run all the time, since we're not interested in testing rapidJSON functionality so much as testing our interaction with it, so conditionally running it seems OK? I'm guessing there's something I can add to the MPC file to make it only get built / run when gtest is available...

dds/idl/v8_generator.cpp
simpsont-oci: I was trying to avoid the empty string, but I can move the final ) in, in order to remove the propName redundancy.

dds/idl/be_util.cpp
simpsont-oci: OK, I'll add it.

dds/idl/metaclass_generator.cpp
simpsont-oci: OK

dds/DCPS/RapidJsonTypeConverter.h
simpsont-oci: I'm not sure how it would support read / write, as it's not a JS engine like V8, just a JSON parser.

mitza-oci: I agree, so based on that fact can we remove the writer-related methods from this class?

tests/idl/rapidjson_generator/RapidJsonTest.cpp
```diff
@@ -0,0 +1,205 @@
+#include "gtest/gtest.h"
```
mitza-oci: This changes creates a new "category" of tests in the source tree.  Is that needed or can it be part of `tests/DCPS/Compiler`?

dds/idl/be_util.cpp
```diff
@@ -35,12 +35,18 @@ be_util::prep_be_arg (char *arg)
   static const size_t SZ_WB_VERSIONING_END = sizeof(WB_VERSIONING_END) - 1;
   static const char WB_PCH_INCLUDE[] = "pch_include=";
   static const size_t SZ_WB_PCH_INCLUDE = sizeof(WB_PCH_INCLUDE) - 1;
+  static const char WB_CPP_INCLUDE[] = "cpp_include=";
+  static const size_t SZ_WB_CPP_INCLUDE = sizeof(WB_CPP_INCLUDE) - 1;
   static const char WB_JAVA[] = "java";
   static const size_t SZ_WB_JAVA = sizeof(WB_JAVA) - 1;
   static const char WB_TAO_INC_PRE[] = "tao_include_prefix=";
   static const size_t SZ_WB_TAO_INC_PRE = sizeof(WB_TAO_INC_PRE) - 1;
+  static const char WB_NO_DEFAULT_GEN[] = "no_default_gen";
```
mitza-oci: There are two different kinds of command-line arguments in `opendds_idl`.  The ones that are "top-level" like -L and -G options are handled separately (not in this file).  The others (in this file) are the -Wb, options.
As part of this change I think we should reconcile this and come up with a more systematic way of handling it.  Since we have -G options already, it would be good for rapidjson to be there, probably v8 too for consistency.  We can discuss more next week.

iguessthislldo:
Tim, If you're making Google Test required for tests in this PR (or whoever does in whatever PR), can you update [`docs/dependencies.md`](https://github.com/objectcomputing/OpenDDS/blob/master/docs/dependencies.md#google-test)?

dds/idl/be_util.cpp
iguessthislldo: Currently Tim's Options Are:
```
-Wb,cpp_include=<include path>
-Wb,no_default_gen
-Wb,rapidjson
```

So I'm thinking maybe
```
-Wb,cpp_include=<include path> (Stays the same)
-Sdefault-gen
-Grapidjson
```

In addition to those, changing these:
```
-Wb,java[=<output_file>]
-Wb,v8
-ZC <include>
```

To be consistent with the rest:
```
-Gjava[=<output_file>]
-Gv8
-Wb,ts_impl_cpp_include=<include path>
```

mitza-oci: - `-Sdefault-gen` could be just `-Sdefault`
- `-Grapidjson` OK
- `-Wb,java` I'd rather keep this way (or some other option) than use `-Gjava` which would be for actually generating the Java mapping (the current one is really a modifier to the type-support generator, though I'm not sure end users care)
- `-Gv8` OK
- `-Wb,ts_impl_cpp_include` reduce to `-Wb,ts_cpp_include` ?

For V8 we would probably keep compatibility with the old option.  Not sure we need the others, -ZC is an internal option for safety profile.

mitza-oci:
Travis build with safety profile:
```
/home/travis/build/objectcomputing/OpenDDS/build/target/dds/DdsDcpsConditionSeqTypeSupportImpl.cpp:5:10: fatal error: =DdsDcpsInfrastructureC.h: No such file or directory
 #include "=DdsDcpsInfrastructureC.h"
```

MPC/config/dcps_rapidjson.mpb
```diff
@@ -0,0 +1,10 @@
+project: dcps_ts_defaults {
+
+  dcps_ts_flags += -Grapidjson
+
+  expand(RAPIDJSON_ROOT) {
```
mitza-oci: This duplicates part of `MPC/config/itl.mpb`, can that one be changed to inherit from this one?

tests/DCPS/Compiler/rapidjson_generator/Main.cpp
```diff
@@ -0,0 +1,7 @@
+#include "gtest/gtest.h"
+
+int main(int argc, char* argv[])
```
mitza-oci: This test needs to have a `run_test.pl` and an entry in `dcps_tests.lst`

iguessthislldo: Yes, I have a `run_test.pl` ready, I just need to figure out if it needs any flags for `dcps_tests.lst`

MPC/config/dcps_rapidjson.mpb
iguessthislldo: I went ahead and created a generic `rapidjson.mpd` which both inherit from.

iguessthislldo:
@simpsont-oci For Azure there are compile issues with RapidJsonTest for both macOS builds and "LinuxSecurityWithoutFeatures" that I think you might have a better time understanding than me.

mitza-oci:
@simpsont-oci Please check this rapidjson issue on macOS:
```
rapidjson/include/rapidjson/document.h:906:184: error: ambiguous conversion for functional-style cast from 'const long' to 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >'
    template <typename T> RAPIDJSON_DISABLEIF_RETURN((internal::OrExpr<internal::IsPointer<T>,internal::IsGenericValue<T> >), (bool)) operator==(const T& rhs) const { return *this == GenericValue(rhs); }
                                                                                                                                                                                       ^~~~~~~~~~~~~~~~
../../../../tests/googletest/build/install/include/gtest/gtest.h:1392:11: note: in instantiation of function template specialization 'rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >::operator==<long>' requested here
  if (lhs == rhs) {
          ^
../../../../tests/googletest/build/install/include/gtest/gtest.h:1421:12: note: in instantiation of function template specialization 'testing::internal::CmpHelperEQ<rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >, long>' requested here
    return CmpHelperEQ(lhs_expression, rhs_expression, lhs, rhs);
           ^
/Users/vsts/agent/2.150.3/work/1/s/tests/DCPS/Compiler/rapidjson_generator/RapidJsonTest.cpp:153:3: note: in instantiation of function template specialization 'testing::internal::EqHelper<false>::Compare<rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> >, long>' requested here
  ASSERT_EQ(val["bt"]["ul"], 2147483690);
```

PR #1099, opened 2019-04-22
configure: updated Qt5 vcpkg support

mitza-oci:
also cleaned up style for consistency


configure
```diff
@@ -590,8 +590,8 @@ if (!$ace_src || !$tao_src) {
 
     $ace_src = normalizePath('ACE_TAO/ACE');
     $tao_src = normalizePath('ACE_TAO/TAO');
-
-  } else {
+  }
+  else {
```
iguessthislldo: I wouldn't have guessed this was a recommend style, is this a Perl convention?

mitza-oci: Yes, it matches MPC and probably other Perl projects.

PR #1101, opened 2019-04-25
Security Registry and cleanup

simpsont-oci:
Fixing leak in SEDP's Task MQ allocation (caused by previous pull request to increase MQ size to help alleviate deadlock issues), and also a singleton "leak" from SecurityRegistry.


mitza-oci:
The commit that updates SEDP's use of ACE message queue was merged already as #1106

PR #1102, opened 2019-04-25
Add cmake test back to vs2019 now that cmake is upgraded on azure

jwillemsen:
* azure-pipelines.yml:


mitza-oci:
Looks like CMake has changed how they handle 64-bit builds:
```
Generators
* Visual Studio 16 2019        = Generates Visual Studio 2019 project files.
                                 Use -A option to specify architecture.
```

jwillemsen:
We also should add openssl back

Issue #999, opened 2018-11-29
No Visual Studio 2015 azure devops build


jwillemsen:
Looks we don't have a Visual Studio 2015 azure devops configuration

mitza-oci:
Is there enough added value in spending the build time on VS2015 in addition to 2017 + 2019?  It's all the same major version of the underlying VC++ compiler.

jwillemsen:
Maybe add just one configuration?

PR #1123, opened 2019-05-13
Add Visual Studio 2015 job to azure, see issue #999

jwillemsen:
* azure-pipelines.yml:


jwillemsen:
Implements issue #999

PR #1127, opened 2019-05-23
Always send beacons to the RtpsRelay

jrw972:
Problem: rtps_udp transport not known by RtpsRelay in all cases

The previous implementation sends beacons to the relay for readers.
However, a writer can timeout of the relay if it does not send
anything to the relay.

Solution: Send beacons to the relay if enabled.  This handles any
combination of writers and readers at the cost of efficiency.


mitza-oci:
The Jenkins build was stuck, not sure if there's an actual runtime issue with this change but let's try some more local testing too.

Issue #1100, opened 2019-04-23
Domain Participant interaction with Monitoring library


XZLeo:
Hi, I'm working on a gateway program based on OpenDDS recently. According to the userguide.html, the graphical monitor application in $DDSroot/tools/monitor may help. First I tried the monitor application in Ubuntu terminal which is in $DDSroot/ test/DCPS/Monitor. It worked and displayed topic, domainparticipant, and other entities' information in the terminal. Then I tried the graphical one in the same way. When I started the subscriber process in another terminal, the graphical monitor shut down and returned an error that indicated segment fault.  
![2019-04-23 17-20-50屏幕截图](https://user-images.githubusercontent.com/39155356/56569747-2c370000-65ec-11e9-984e-a538d8997c49.png)

mitza-oci:
Please run `ldd $DDS_ROOT/bin/monitor` and post the results here as plain text.
Also if you run the monitor GUI in the gdb debugger you should get a stack trace for the segfault.  If you post that here (again, as plain text) we can recommend some next steps.

XZLeo:
Thanks for your response. Here's the result.

leo@leo-Inspiron-5548:~/OpenDDS-3.13.2/bin$ ./monitor
Gtk-Message: 15:26:31.389: Failed to load module "canberra-gtk-module"

Then the monitor GUI opened but had the same problem when I opened subscriber as I described before.
-------------------------------------------------------------------------------------------------------------------
gdb debugger:

leo@leo-Inspiron-5548:~/OpenDDS-3.13.2/bin$ gdb monitor
GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from monitor...done.
(gdb) r
[...]

PR #1132, opened 2019-05-28
Issue 1100: Domain Participant interaction with Monitoring library

lewisdo:
When using the monitor GUI a segmentation fault was occurring.  It was determined that the monitor's report functions were being called before a GUID was assigned in the DomainParticipantImpl::enable() function.  The calls were moved so that they did not execute until after a value GUID was obtained.


mitza-oci:
Link to Issue #1100

PR #1137, opened 2019-06-04
Support newer versions of the JDK

mitza-oci:
Starting with JDK9, JPMS (Java Platform Module System) is in effect.
The impact for OpenDDS is mostly what we used to have for the "compact" JRE,
therefore the MPC feature full_jre (the opposite of compact) was renamed to
java_pre_jpms.  Java versions starting with 9 need to set that feature to 0.

Aside from building a JAR with the missing org.omg.CORBA classes that we need,
the build process for JNI code no longer uses javah as a separate step.
To minimize impact on normal (non-JNI) java projects, support for javah was
moved to its own base project, javah.mpb.

Applications using these newer JDK versions will depend on i2jrt_corba.jar in
addition to the other JAR files that have always been used.


iguessthislldo:
I have a condition in configure (line 1577 on master) in `generate_workspace()` that disables `full_jre` for cross-compiled targets.

iguessthislldo:
Exact Link: https://github.com/objectcomputing/OpenDDS/blob/7b98a762dcfc29965bce8bf758a2059c8a4deda9/configure#L1577

mitza-oci:
@iguessthislldo `configure` script support was not attempted, once the other PRs is merged I'll merge master here so we can sync it up

For `full_jre`, not all embedded targets will use a compact JRE.  It would be up to the user to supply `--features`.  However, that will soon be a moot point.  We will probably need to do JDK version detection in `configure`.

iguessthislldo:
From what I remember from making Android work, using java and cross-compiling didn't work at all, setting java would just do java for the host, so I didn't know we supported other cross compile targets with java. Regardless if they don't require `i2jrt_corba.jar`, then maybe we can set `java_pre_jpms=0` based on if the target is Android, since we know when we are building for Android and it always requires `i2jrt_corba.jar`.

mitza-oci:
> ... setting java would just do java for the host

I didn't see that in my test:
```
$ ./configure --target=linux-cross --target-compiler=/opt/cross-pi-gcc/bin/arm-linux-gnueabihf-g++ --java --no-tests
$ find -name default.features | xargs egrep 'java|jre'
./build/target/ACE_TAO/ACE/bin/MakeProjectCreator/config/default.features:java=1
./build/target/ACE_TAO/ACE/bin/MakeProjectCreator/config/default.features:full_jre=0
./build/host/ACE_TAO/ACE/bin/MakeProjectCreator/config/default.features:java=1
```

iguessthislldo:
Now I'm not sure about what was happening, because I went back to the configure script before #964 and it looks like it's working alright. Whatever I'm thinking of, I think it has to do with this like [these lines](https://github.com/objectcomputing/OpenDDS/pull/964/files#diff-e2d5a00791bce9a01f99bc6fd613a39dR1553), which, looking at now, should be taken out if java is already being set.

mitza-oci:
The host side gets java=1 in order to build idl2jni, but that may already be happening elsewhere.

PR #1138, opened 2019-06-04
RtpsDataLink Partial Restructuring & Bundled AckNacks

simpsont-oci:
RtpsReaders and RtpsWriters now use RCH and have own locking with weak RCH back to datalink. datareader_dispatch() unlocks before calls to individual reader methods. Locator access is limited to accessor methods and locked when called from outside the datalink. User writes no longer need to lock datalink mutex prior to call to send strategy write. AckNacks are now "gathered" from all readers before being "bundled" and sent out by locator address & (info) destination.


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1452,19 +1452,24 @@ RtpsUdpDataLink::send_bundled_responses(ResponseVec& responses)
 
   const RepoId no_dst = { { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }, { { 0, 0, 0 }, 0 } };
```
iguessthislldo: In `dds/DCPS/GuidUtils.h` there is `OpenDDS::DCPS::GUID_UNKNOWN`.

simpsont-oci: I saw that, but how do the single {0}s expand there? Do they turn into multiple {0}s or is it just the first one?

simpsont-oci: "If the number of initializer clauses is less than the number of members and bases (since C++17) or initializer list is completely empty, the remaining members and bases (since C++17) are initialized by their default member initializers, if provided in the class definition, and otherwise (since C++14) by empty lists, in accordance with the usual list-initialization rules (which performs value-initialization for non-class types and non-aggregate classes with default constructors, and aggregate initialization for aggregates). If a member of a reference type is one of these remaining members, the program is ill-formed."

So I think the first values get 0's and then the rest are default initialized?

mitza-oci: > which performs value-initialization for non-class types

which means guaranteed to be 0

simpsont-oci: Oh, the rest are initialized by empty lists, which performs value initialization for non-class types. Sure, I can switch over to GUID_UNKNOWN.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -798,14 +840,121 @@ RtpsUdpDataLink::customize_queue_element(TransportQueueElement* element)
...
+    GuardType guard(this->strategy_lock_);
+    if (this->send_strategy_) {
+      send_strategy()->encode_payload(pub_id, data, subm);
+    }
+  }
+#endif
+
+  Message_Block_Ptr hdr(submsgs_to_msgblock(subm));
+  hdr->cont(data.release());
+  RtpsCustomizedElement* rtps =
```
mitza-oci: Doesn't seem like we need a named local variable for this, it can just be returned

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -957,66 +1103,67 @@ void
 RtpsUdpDataLink::received(const RTPS::DataSubmessage& data,
                           const GuidPrefix_t& src_prefix)
 {
-  datareader_dispatch(data, src_prefix, &RtpsUdpDataLink::process_data_i);
+  datareader_dispatch(data, src_prefix, &RtpsUdpDataLink::RtpsReader::process_data_i);
```
```suggestion
  datareader_dispatch(data, src_prefix, &RtpsReader::process_data_i);
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1238,44 +1390,128 @@ RtpsUdpDataLink::WriterInfo::should_nack() const
 }
 
 bool
-RtpsUdpDataLink::RtpsReader::nack_durable(const WriterInfo& info)
+RtpsUdpDataLink::RtpsWriter::add_reader(const RepoId& id, const ReaderInfo& info)
 {
-  return durable_ && (info.recvd_.empty() ||
-                      info.recvd_.low() > info.hb_range_.first);
+  bool result = false;
```
```suggestion
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1238,44 +1390,128 @@ RtpsUdpDataLink::WriterInfo::should_nack() const
...
 }
 
-void
-RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
+bool
+RtpsUdpDataLink::RtpsWriter::has_reader(const RepoId& id) const
 {
-  using namespace OpenDDS::RTPS;
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, mutex_, false);
+  return remote_readers_.find(id) != remote_readers_.end();
```
```suggestion
  return remote_readers_.count(id);
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1341,85 +1577,233 @@ RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
...
+    for (ResponseIterVec::const_iterator it = response_bundles[i].begin(); it != response_bundles[i].end(); ++it) {
+      Response& res = **it;
+      RepoId dst = res.dst_guid_;
+      dst.entityId = ENTITYID_UNKNOWN;
+      if (dst != GUID_UNKNOWN && dst != prev_dst) {
+        memcpy(&idst.guidPrefix, dst.guidPrefix, sizeof(idst.guidPrefix));
+        ser << idst;
+        prev_dst = dst;
+      }
+      switch (res.sm_._d()) {
```
mitza-oci: This is essentially the generated code `operator<<(Serializer&, const Submessage&)`

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1444,54 +1828,44 @@ RtpsUdpDataLink::send_heartbeat_replies() // from DR to DW
...
-    };
-    gen_find_size(info_dst, size, padding);
-
-    ACE_Message_Block mb_acknack(size + padding); //FUTURE: allocators?
-    // byte swapping is handled in the operator<<() implementation
-    Serializer ser(&mb_acknack, false, Serializer::ALIGN_CDR);
-    std::memcpy(info_dst.guidPrefix, pos->writerid.guidPrefix,
-                sizeof(GuidPrefix_t));
-    ser << info_dst;
-    // Interoperability note: we used to insert INFO_REPLY submessage here, but
```
mitza-oci: We shouldn't lose this.  It could move to `docs/design/RTPS`.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1584,16 +1952,18 @@ void
...
+  datareader_dispatch(hb_frag, src_prefix, &RtpsUdpDataLink::RtpsReader::process_hb_frag_i);
 }
 
 bool
-RtpsUdpDataLink::process_hb_frag_i(const RTPS::HeartBeatFragSubmessage& hb_frag,
-                                   const RepoId& src,
-                                   RtpsReaderMap::value_type& rr)
+RtpsUdpDataLink::RtpsReader::process_hb_frag_i(const RTPS::HeartBeatFragSubmessage& hb_frag,
+                                               const RepoId& src,
+                                               ResponseVec& /*responses*/)
```
```suggestion
                                               ResponseVec&)
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2369,129 +2725,58 @@ RtpsUdpDataLink::send_heartbeats()
...
+      {HEARTBEAT, FLAG_E, HEARTBEAT_SZ},
+      ENTITYID_UNKNOWN, // any matched reader may be interested in this
+      pos->first.entityId,
+      {SN.getHigh(), SN.getLow()},
+      {lastSN.getHigh(), lastSN.getLow()},
+      {++heartbeat_counts_[pos->first]}
+    };
 
-    send_directed_heartbeats(subm);
+    Response response;
```
mitza-oci: Should Response have a constructor that handles some of this init?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2501,58 +2786,115 @@ RtpsUdpDataLink::send_heartbeats()
...
+  if (std::memcmp(&id_.entityId, &volatile_writer, sizeof(EntityId_t)) == 0) {
+    RepoIdSet guids = response.to_guids_;
+    response.dst_guid_ = GUID_UNKNOWN;
+    for (RepoIdSet::const_iterator it = guids.begin(); it != guids.end(); ++it) {
+      response.to_guids_.clear();
+      response.dst_guid_ = (*it);
+      response.sm_.heartbeat_sm().readerId = it->entityId;
+      responses.push_back(response);
+    }
+  } else
```
mitza-oci: May be easier to `return` here instead of splitting the if/else across the conditional block

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2922,18 +3287,30 @@ OpenDDS::DCPS::RtpsUdpDataLink::receive_strategy()
   return static_cast<OpenDDS::DCPS::RtpsUdpReceiveStrategy*>(receive_strategy_.in());
 }
 
-OPENDDS_SET(ACE_INET_Addr)
+OpenDDS::DCPS::RtpsUdpDataLink::AddrSet
```
```suggestion
RtpsUdpDataLink::AddrSet
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Gotcha. Is there a gen_find_size equivalent? I'm guessing yes... do we want to go ahead and switch this over to try to handle all submessages?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: I'm not really a fan of implicit integer / pointer -> boolean conversions. Having the explicit comparison keeps the types as you'd expect and causes helpful compiler warnings if/when the surrounding code changes, in addition to (usually) being optimized away by the compiler. I get that something like this might make the code-as-written a little more concise & readable, but I'm not sure the resilience / maintainability tradeoff is worth it. In addition, if remote_readers_ ever needed to turn into a multiset (for whatever reason), this would be doing more work than is strictly required by finding all the matches and not just _any_ match.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Which part? The code itself or the comment about INFO_REPLY?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: You can add the `!= 0` if you want, the more important part is that a `count` operation is different than a `find` operation.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: I think we might as well use those.  We may not actually support all submessage types in this method but it's fine to use the generated code.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: Just the comment about interoperability

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1341,85 +1577,233 @@ RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
...
+      }
+      for (RepoIdSet::iterator it2 = it->to_guids_.begin(); it2 != it->to_guids_.end(); ++it2) {
+        accumulate_addresses(it->from_guid_, *it2, addrs);
+      }
+      if (addrs.empty()) {
+        continue;
+      }
+      if (std::memcmp(&(it->dst_guid_.guidPrefix), &GUIDPREFIX_UNKNOWN, sizeof(GuidPrefix_t)) != 0) {
+        RepoId dst;
+        memcpy(dst.guidPrefix, it->dst_guid_.guidPrefix, sizeof(dst.guidPrefix));
```
```suggestion
        std::memcpy(dst.guidPrefix, it->dst_guid_.guidPrefix, sizeof(dst.guidPrefix));
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1341,85 +1577,233 @@ RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
...
+    {INFO_DST, FLAG_E, INFO_DST_SZ},
+    {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+  };
+
+  // Build bundles
+  typedef OPENDDS_VECTOR(ResponseIterVec) ResponseIterVecVec;
+  ResponseIterVecVec response_bundles; // a vector of vectors of iterators pointing to responses
+  OPENDDS_VECTOR(AddrSet) response_bundle_addrs; // for a bundle's address set
+  OPENDDS_VECTOR(size_t) response_bundles_sizes; // for allocating the bundle's buffer
+  size_t size = 0, padding = 0, prev_size = 0, prev_padding = 0; // keep track of previous values in case we overshoot
```
mitza-oci: These are managed together as a group.  What would it looks like if it were its own struct with small inline methods for updating the `prev` and "resetting"?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: I didn't consider find_size for this because the code above also manages the contents of the submessageHeader within the `switch` scope.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1341,85 +1577,233 @@ RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
...
+  if (responses.empty()) {
+    return;
+  }
+
+  typedef OPENDDS_VECTOR(ResponseVec::iterator) ResponseIterVec;
+  typedef OPENDDS_MAP_CMP(RepoId, ResponseIterVec, GUID_tKeyLessThan) DestResponseMap;
+  typedef OPENDDS_MAP(AddrSet, DestResponseMap) AddrDestResponseMap;
+  AddrDestResponseMap adr_map;
+
+  {
```
mitza-oci: Refactor this block to its own function?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1341,85 +1577,233 @@ RtpsUdpDataLink::send_ack_nacks(RtpsReaderMap::value_type& rr, bool finalFlag)
...
+          response_bundles_sizes.push_back(prev_size + prev_padding);
+          size -= prev_size; padding -= prev_padding; prev_size = 0; prev_padding = 0;
+          prev_dst = GUID_UNKNOWN;
+        } else {
+          prev_size = size; prev_padding = padding;
+        }
+        // Attempt to add the submessage response to the bundle
+        offset = size;
+        Response& res = **resp_it;
+        switch (res.sm_._d()) {
```
mitza-oci: Refactor this to its own function?  (This, with the other changes, may get the resulting `send_bundled_responses` within range of the arbitrary "fits on one page" metric.)

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1665,26 +2030,116 @@ RtpsUdpDataLink::received(const RTPS::AckNackSubmessage& acknack,
...
 
-  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
-  const RtpsWriterMap::iterator rw = writers_.find(local);
-  if (rw == writers_.end()) {
+  datawriter_dispatch(acknack, src_prefix, &RtpsUdpDataLink::RtpsWriter::process_acknack);
+}
+
+void
+RtpsUdpDataLink::RtpsWriter::gather_gaps_i(const RepoId& reader,
+                                           DisjointSequence gaps,
```
```suggestion
                                           const DisjointSequence& gaps,
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: I think it's OK to just duplicate this one line in the #else portion for code clarity.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -235,30 +251,80 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink {
 
   typedef OPENDDS_MAP_CMP(RepoId, ReaderInfo, GUID_tKeyLessThan) ReaderInfoMap;
 
-  struct RtpsWriter {
+  class  RtpsWriter : public RcObject {
+  protected:
```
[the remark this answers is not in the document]
simpsont-oci: I didn't want RtpsUdpDataLink's methods tinkering around with the internals of RtpsWriter, which this seemed to prevent. If you'd rather this be private, I'm OK with that for now, although someday it may make sense to split RtpsWriter into reliable and non-reliable subclasses so we can move the handling of non-reliable stuff out of RtpsUdpDataLink's methods. With that in mind, I don't see any harm in leaving them protected.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -210,6 +219,13 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink {
 
   } multi_buff_;
 
+  struct Response {
```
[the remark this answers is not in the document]
simpsont-oci: AnnotatedSubmessage? MetaMessage? DirectedSubmessage? (I like this one less, since we could extend the metadata to include more than locators / INFO_DST)

Issue #1110, opened 2019-04-30
configure failed on Windows7 Chinese version


jhenchen:
I'm using My VS2015 command line tool with the configure file.But It failed.The complete command line output is :
![output](https://user-images.githubusercontent.com/18344904/56998656-a3692700-6bde-11e9-9975-0723ce1030e1.png)
I have read the issue #762, but It maybe a new issue.

mitza-oci:
At least for the short term, you can hard-code the version in your local copy of the configure script.  

Perhaps the best way to support it would be to accept a command line argument for configure like `--vs-version=2015`.

jhenchen:
Thanks.

tiedun001:
change IDE to VS2017 and download the newest opendds version and others,then ./configure to make.

iguessthislldo:
I'm wondering if [vswhere](https://github.com/Microsoft/vswhere) could help with this for newer versions of VS that have it.

I took of the batch examples and modified it slightly:
```
@echo off
setlocal enabledelayedexpansion

set PATH=%PATH%;%ProgramFiles(x86)%\Microsoft Visual Studio\Installer

for /f "usebackq tokens=*" %%i in (`vswhere -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath`) do (
  set InstallDir=%%i
)

if exist "%InstallDir%\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt" (
  set /p Version=<"%InstallDir%\VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt"

  rem Trim
  set Version=!Version: =!
)

echo %Version%
pause
```
On my work laptop with VS2019 this results in `14.20.27508`.

It's not the exact same as the version of `cl` but MSBuild's version is always 5 major versions less than it's `cl`. Of course maybe if we're already in the vs dev environment, we can bypass vswhere and use the file with the version directly.

mitza-oci:
@iguessthislldo I don't want to have to maintain a complex script with dependencies on the locations of things on the filesystem and the compiler target architecture.  It would be easier to run `cl` and ask it to dump `_MSC_VER`.

mitza-oci:
> change IDE to VS2017 and download the newest opendds version and others,then ./configure to make.

I don't understand what this comment means.  Does the upgrade from VS 2015 to 2017 resolve the issue?

mitza-oci:
[...]

Issue #922, opened 2018-08-30
Configure script doesn't support MSVC Compiler versions like x.x.x.x


ajkerzner:
The regular expression on line 399 of the OpenDDS configure script does not cover cases where the Microsoft C/C++ compiler has a four-component version (e.g. x.x.x.x).

This regular expression  `(\d+)\.(\d+)\.[\d]+ for (x\d+)/i` should change to something like ` (\d+)\.(\d+)\.[\d\.]+ for (x\d+)/i`; or perhaps something safer, like a non-capturing group to optionally catch (but not capture) a 3rd dot followed by a number.

Seen on: "Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x64"
Also on this same system, this compiler is detected because its version is only three components: "Microsoft (R) C/C++ Optimizing Compiler Version 19.14.26433 for x86"

See: https://github.com/objectcomputing/OpenDDS/blob/e39d373cc78cc9bac4a936035611ced466e42743/configure#L399

mitza-oci:
Good observation, please open a Pull Request to change the regex.

ajkerzner:
Apologies for the delay. Will mmcparlane be fixing this, or I?

mitza-oci:
@ajkerzner you can go ahead

ajkerzner:
Okay, thanks.

ajkerzner:
Should be ready to merge in now. A build failure, but not sure why it failed. As all of the tests are for gcc, and my change only affects compilation with cl or cl.exe (re: MSVC), I think it's fine.
I tested it on a machine I use, and my one use/build case worked.

iguessthislldo:
Travis jobs are limited to 50 minutes if you don't pay them, it's not because of your PR.

PR #1141, opened 2019-06-07
configure: Detect MSVC Version Using C Macro

iguessthislldo:
Fix for #1110.

Instead of relying on using regex on the version message, which might fail in at least one major localization (#1110) or future versions (#922).

Also fixes the final message for VS users to tell them how they can open the sln file right then and there.

### DIFF
--- a/configure
+++ b/configure
@@ -50,7 +50,7 @@ my %platforminfo =
                                  16 => 'vc10', 17 => 'vc11', 18 => 'vc12',
                                  19 => 'vc14', 19.1 => 'vs2017',
                                  19.2 => 'vs2019'},
-               'cl_archs' = {'x64' => 1, 'Win32' => 1, 'ARM64' => 0, 'ARM' => 0}, # 1 if supported
+               'cl_archs' => {'x64' => 1, 'Win32' => 1, 'ARM64' => 0, 'ARM' => 0}, # 1 if supported
               },
    'macosx' => {
                 'compilers' => ['clang++'],
--- a/configure
+++ b/configure
@@ -50,6 +50,7 @@ my %platforminfo =
                                  16 => 'vc10', 17 => 'vc11', 18 => 'vc12',
                                  19 => 'vc14', 19.1 => 'vs2017',
                                  19.2 => 'vs2019'},
+               'cl_archs' = {'x64' => 1, 'Win32' => 1, 'ARM64' => 0, 'ARM' => 0}, # 1 if supported
               },
    'macosx' => {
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -470,32 +470,57 @@ if ($opts{'compiler'} =~ /cl(\.exe)?$/i) {
...
+      "script from the Visual Studio Command Prompt.\nStopped";
+  my $ver = 0;
+  my $arch = '';
+  while (my $line = <$cl_out_fd>) {
+    chomp ($line);
+    print "CL says: $line\n" if $opts{'verbose'};
+    # Convert _MSC_VER to the form \d+\.\d
+    if ($line =~ /^CL is (\d+)\d (\w+)$/) {
+      $ver = int($1) / 10;
+      $arch = $2;
```
mitza-oci: This `$arch` is passed down to cmake, so I don't think `x32` is valid.
https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html

iguessthislldo: Alright, didn't think of what it was being used for, but I can fix that. Should we worry about the ARM targets at all?

mitza-oci: I wouldn't worry too much about the ARM targets at this point, but if we can just pass it through that should be fine.  Looks like the function `win32_cmake_arch` may need to change for this.

iguessthislldo: > I wouldn't worry too much about the ARM targets at this point, but if we can just pass it through that should be fine.

Okay

> This $arch is passed down to cmake, so I don't think x32 is valid.

`win32_cmake_arch()` and `target_win64` both only look for `x64`, so actually `x32` works. Although, I'm not quite sure why I changed it from `x86` in your original comment to `x32`, so I think I will change that even if it works.

mitza-oci: The ones in my original notes were from running the compiler and looking at its "logo" output:
```
Microsoft (R) C/C++ Optimizing Compiler Version 19.21.27702.2 for x86
```
VS uses `Win32` as the platform name and it looks like `cmake -A` follows this.  I was thinking that our `win32_cmake_arch` could be more future-proof.

mitza-oci:
Doesn't look like it's valid Perl?
```
Can't modify constant item in scalar assignment at configure line 53, near "},"
```

iguessthislldo:
Yea it's not, pushing a fix, I should've run it before commiting.

iguessthislldo:
[Looks a qt server vcpkg was using went down](https://dev.azure.com/opendds/OpenDDS/_build/results?buildId=566&view=logs), Restarting the windows builds.

PR #1145, opened 2019-06-11
ICE code cleanup

jrw972:
Problem: ICE has duplicate code for candidates and stringification

Solution: Remove duplicate code and use safety streams for
stringification.


mitza-oci:
I'm merging this branch into #1144 since there's so much overlap.

PR #1146, opened 2019-06-12
Added dependencies to generated .java files based on forward declarations.

ocielliottc:
These files are required to compile the .java files, but, previously, may not have existed at that time.


mitza-oci:
The build on AppVeyor errored out with:
```
Can't use an undefined value as an ARRAY reference at C:\projects\opendds\ext\MPC/modules/ProjectCreator.pm line 1286, <GEN3904> line 26.
```

PR #1157, opened 2019-06-17
Provide a more accurate representation of dependencies

ocielliottc:
Provide a more accurate representation of dependencies if the version of MPC is 4.1.42 or higher.  We need the ProjectCreator in order to create .class file names from the .java files that we have based on forward declarations.


ocielliottc:
This leads to indirect circular dependencies that are displayed as warnings in Visual Studio 2019.  GNU Make does not detect these circular dependencies (that I have seen).

ocielliottc:
This last commit clears up the circular dependency warnings from Visual Studio.

Issue #995, opened 2018-11-27
InconsistentTopic test scoreboard failures


mitza-oci:
```
2018-11-27 13:28:10.973@LM_ERROR@(1856|3044) ERROR: Service_Participant::load_discovery_configuration Unable to load libraries for rtps_discovery
```
Either add support for static builds or exclude at runtime in the dcps_tests.lst.

mitza-oci:
The other problem with this test seems to be its dependence on timing.  At least that's the only explanation I can see for its failures.

mitza-oci:
With #996 it now works on static builds.  However I do see one failure on the scoreboard, not sure what's causing this.
```
.\.\\publisher.EXE -DCPSDebugLevel 1 -ORBVerboseLogging 1 -DCPSTransportDebugLevel 1 -ORBLogFile pub.log -DCPSConfigFile rtps_disc.ini -DCPSPendingTimeout 3
pub PID: 4984
.\.\\subscriber.EXE -DCPSDebugLevel 1 -ORBVerboseLogging 1 -DCPSTransportDebugLevel 1 -ORBLogFile sub.log -DCPSConfigFile rtps_disc.ini -DCPSPendingTimeout 3
sub PID: 1232
subscriber total count: 1
should have an inconsistent topic total count not equal zero.
2018-11-29 00:05:12: ERROR: pub returned 1 (started at 2018-11-29 00:05:12)
<<<<< sub.log >>>>>
2018-11-29 00:05:07.423@LM_NOTICE@(1232|4092) NOTICE: using DCPSDebugLevel value from command option (overrides value if it's in config file).
2018-11-29 00:05:07.425@LM_NOTICE@(1232|4092) NOTICE: "DCPSInfoRepo" is not defined in config file - using code default.
2018-11-29 00:05:07.425@LM_NOTICE@(1232|4092) NOTICE: "DCPSRTISerialization" is not defined in config file - using code default.
2018-11-29 00:05:07.425@LM_NOTICE@(1232|4092) NOTICE: "DCPSChunks" is not defined in config file - using code default.
2018-11-29 00:05:07.425@LM_NOTICE@(1232|4092) NOTICE: "DCPSChunkAssociationMutltiplier" is not defined in config file - using code default.
2018-11-29 00:05:07.425@LM_NOTICE@(1232|4092) NOTICE: "DCPSBitTransportPort" is not defined in config file - using code default.
[...]

PR #1164, opened 2019-06-19
Fix inconsistent topic test

jrw972:
Problem: Inconsistent topic only discovered with local reader/writer

A participant can only get inconconsistent topic when it has a local
reader or writer on that topic.  While usual, a participant should be
able to create a topic and then get inconsistent topic on it without
creating readers and writers.  This also lead to a race condition in
earlier versions of the InconsistentTopic test which made it fail
sporadically.  This is a fix for #995.

Solution: Track local and remote topics and link the local topic to
its impl through callbacks so that inconsistent topic can be reported
without a reader or writer.

### DIFF
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -204,6 +204,8 @@ namespace OpenDDS {
                        const DDS::TopicQos& qos,
                        bool has_dcps_key,
                        TopicCallbacks* topic_callbacks) {
+          assert(topic_callbacks != 0);
+
           local_data_type_name_ = data_type_name;
           local_qos_ = qos;
           has_dcps_key_ = has_dcps_key;
@@ -233,7 +235,9 @@ namespace OpenDDS {
             }
           }
 
-          topic_callbacks_->inconsistent_topic(inconsistent_topic_count_);
+          if (inconsistent_topic_count_ != 0) {
+            topic_callbacks_->inconsistent_topic(inconsistent_topic_count_);
+          }
         }
 
         void unset_local() {
@@ -350,6 +354,7 @@ namespace OpenDDS {
         const DDS::TopicQos local_qos() const { return local_qos_; }
         DCPS::RepoId topic_id() const { return topic_id_; }
         bool has_dcps_key() const { return has_dcps_key_; }
+        bool local_is_set() const { return topic_callbacks_; }
         const RepoIdSet& endpoints() const { return endpoints_; }
 
         bool is_dead() const {
@@ -485,7 +490,7 @@ namespace OpenDDS {
         typename OPENDDS_MAP(OPENDDS_STRING, TopicDetails)::iterator iter =
           topics_.find(topicName);
         if (iter != topics_.end()) {
-          if (iter->second.local_data_type_name() != dataTypeName) {
+          if (iter->second.local_is_set() && iter->second.local_data_type_name() != dataTypeName) {
             return DCPS::CONFLICTING_TYPENAME;
           }
           topicId = iter->second.topic_id();
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -6,6 +6,7 @@
 #ifndef OPENDDS_DDS_DCPS_DISCOVERYBASE_H
 #define OPENDDS_DDS_DCPS_DISCOVERYBASE_H
 
+#include "dds/DCPS/TopicDetails.h"
 #include "dds/DCPS/BuiltInTopicUtils.h"
 #include "dds/DCPS/DataReaderImpl_T.h"
 #include "dds/DCPS/DCPS_Utils.h"
@@ -18,7 +19,6 @@
 #include "dds/DCPS/SubscriberImpl.h"
 
 #include "dds/DdsDcpsCoreTypeSupportImpl.h"
-#include "dds/DCPS/TopicCallbacks.h"
 
 #ifdef OPENDDS_SECURITY
 #include "dds/DdsSecurityCoreC.h"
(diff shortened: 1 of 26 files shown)


tests/DCPS/InconsistentTopic/pubsub.cpp
```diff
@@ -140,8 +140,12 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[]) {
         cerr << "ERROR: not able to retrieve topic status." << endl;
         exit(1);
       }
+      if (status.total_count != 0) {
+        cerr << "ERROR: (alpha) participant 1 total_count should be 0 but is " << status.total_count << endl;
+        test_error = true;
+      }
       if (status.total_count_change != 0) {
-        cerr << "ERROR: participant 1 saw an inconsistent topic when it should not (first)" << endl;
+        cerr << "ERROR: (alpha) participant 1 total_count_chnage should be 0 but is " << status.total_count_change << endl;
```
jwillemsen: typo

dds/DCPS/TopicCallbacks.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_TOPICCALLBACKS_H
+#define OPENDDS_DCPS_TOPICCALLBACKS_H
+
+#include "dds/DCPS/Definitions.h"
+#include "dds/DCPS/DiscoveryListener.h"
```
mitza-oci: This include appears to be unused

dds/DCPS/TopicImpl.cpp
```diff
@@ -146,7 +146,8 @@ TopicImpl::enable()
                                              topic_name_.c_str(),
                                              type_name_.c_str(),
                                              qos_,
-                                             type_support_->has_dcps_key());
+                                             type_support_ ? type_support_->has_dcps_key() : false,
```
mitza-oci: When can type_support_ be null?  Could this be checked earlier (and we bail out) or is it valid?

dds/DCPS/DiscoveryBase.h
```diff
@@ -191,11 +192,193 @@ namespace OpenDDS {
...
+        OPENDDS_STRING name_;
+        OPENDDS_STRING local_data_type_name_;
+        DDS::TopicQos local_qos_;
+        DCPS::RepoId topic_id_;
         bool has_dcps_key_;
+        TopicCallbacks* topic_callbacks_;
         RepoIdSet endpoints_;
+
+        struct RemoteTopic {
+          OPENDDS_STRING data_type_name_;
```
mitza-oci: In normal cases, this is the same as local_data_type_name_.  We could make this an empty string if it matches local.

dds/DCPS/TopicCallbacks.h
mitza-oci: Appears to be an unused #include.

dds/DCPS/TopicImpl.cpp
mitza-oci: Can type_support_ be null?  Should we be checking this in an earlier step and bailing out?

dds/DCPS/TopicImpl.cpp
```diff
@@ -35,9 +34,10 @@ TopicImpl::TopicImpl(const RepoId                   topic_id,
     qos_(qos),
     listener_mask_(mask),
     listener_(DDS::TopicListener::_duplicate(a_listener)),
-    id_(topic_id),
+    id_(GUID_UNKNOWN),
     monitor_(0)
 {
+  last_inconsistent_topic_count_ = 0;
```
mitza-oci: Tracking the "last" count as member data doesn't seem to be necessary based on how this pattern works in other parts of the code.  I also don't think it's computing total_count_change correctly for cases that don't have a listener.  Please see other `total_count_change`s and interactions with the `get_*_status` method, like the ones in DataWriterImpl.

dds/DCPS/DiscoveryBase.h
```diff
@@ -272,15 +461,15 @@ namespace OpenDDS {
...
-              for (ep = td.endpoints_.begin(); ep!= td.endpoints_.end(); ++ep) {
-                match_endpoints(*ep, td, true /*remove*/);
-                if (shutting_down()) { return; }
-              }
+            TopicDetails& td = topics_[iter->second];
+            RepoIdSet ids = td.endpoints();
+            for (RepoIdSet::iterator ep = ids.begin(); ep!= ids.end(); ++ep) {
+              match_endpoints(*ep, td, true /*remove*/);
+              td.remove_pub_sub(*ep);
+              if (shutting_down()) { return; }
```
simpsont-oci: What prompts these checks for shutdown? Is something releasing & reacquiring a lock in here? Can the state change every time we call match_endpoints? If that's the case, maybe we need to be checking for shutdown immediately after match_endpoints, including the two calls above? Otherwise, why are we doing it here at all?

dds/DCPS/DiscoveryBase.h
```diff
@@ -191,11 +192,193 @@ namespace OpenDDS {
...
+        OPENDDS_STRING local_data_type_name_;
+        DDS::TopicQos local_qos_;
+        DCPS::RepoId topic_id_;
         bool has_dcps_key_;
+        TopicCallbacks* topic_callbacks_;
         RepoIdSet endpoints_;
+
+        struct RemoteTopic {
+          OPENDDS_STRING data_type_name_;
+          DDS::TopicQos qos_;
```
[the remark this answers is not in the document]
jrw972: I think we can just remove the qos.

dds/DCPS/DiscoveryBase.h
```diff
@@ -253,15 +442,15 @@ namespace OpenDDS {
...
             // break associations
-            typename OPENDDS_MAP(OPENDDS_STRING, TopicDetails)::iterator top_it =
-              topics_.find(topic_name);
-            if (top_it != topics_.end()) {
-              match_endpoints(to_ignore, top_it->second, true /*remove*/);
+            match_endpoints(to_ignore, td, true /*remove*/);
+            if (td.is_dead()) {
+              purge_dead_topic(topic_name);
             }
             return;
```
simpsont-oci: The extra {}'s here feel unnecessary and, in addition to the extra returns everywhere, make this code seem more confusing than it really is.

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
mitza-oci: The Topic Data QoS policy becomes part of the BIT data sample for publication/subscription.  That could be tracked elsewhere though.

dds/DCPS/TopicImpl.cpp
jrw972: I believe it can be null for recorder/replayer.

dds/DCPS/DiscoveryBase.h
```diff
@@ -191,11 +192,193 @@ namespace OpenDDS {
...
+        DCPS::RepoId topic_id_;
         bool has_dcps_key_;
+        TopicCallbacks* topic_callbacks_;
         RepoIdSet endpoints_;
+
+        struct RemoteTopic {
+          OPENDDS_STRING data_type_name_;
+          DDS::TopicQos qos_;
+          bool inconsistent_;
+          RepoIdSet endpoints_;
```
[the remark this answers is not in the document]
jrw972: It needs the full set.   This is to handle the case where the remote topics are discovered before the local topic is created.  In this case, it uses the endpoints for each consistent remote to build the set of endpoints that is used for matching.

tests/DCPS/DCPSInfoRepo/pubsub.cpp
```diff
@@ -118,14 +118,14 @@ bool pubsub(OpenDDS::DCPS::Discovery_rch disc, CORBA::ORB_var orb)
...
-                         longname.c_str(), "shortname", topicQos, key);
+                         longname.c_str(), "shortname", topicQos, key, 0);
     if (ts != OpenDDS::DCPS::PRECONDITION_NOT_MET) {
       failed = true;
       ACE_ERROR((LM_ERROR,
                  ACE_TEXT("ERROR: expected long topic name to be rejected\n")));
     }
     ts = disc->assert_topic(topicId, domain, pubPartId,
-                            "shortname", longname.c_str(), topicQos, key);
+                            "shortname", longname.c_str(), topicQos, key, 0);
```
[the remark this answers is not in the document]
jrw972: From the code
// check that topic/type name string bounds are enforced

dds/DCPS/DiscoveryBase.h
jrw972: The style guide says I should use {}s.

dds/DCPS/DiscoveryBase.h
jrw972: That was pre-existing code.  We should follow up with @mitza-oci

mitza-oci: `match_endpoints(a, b, true)` calls `remove_assoc` which calls back to DCPS.  Please take a look at that code and let me know if there are specific questions I can answer.

mitza-oci:
Errors/warnings from Travis build:
```
/home/travis/build/objectcomputing/OpenDDS/dds/DCPS/DomainParticipantImpl.cpp:464:10: warning: variable ‘has_keys’ set but not used [-Wunused-but-set-variable]
     bool has_keys = (topic_mask & TOPIC_TYPE_HAS_KEYS);
          ^
In file included from /home/travis/build/objectcomputing/OpenDDS/dds/DCPS/StaticDiscovery.h:12:0,
                 from /home/travis/build/objectcomputing/OpenDDS/dds/DCPS/Service_Participant.cpp:20:
../dds/DCPS/DiscoveryBase.h:193:28: error: declaration of ‘typedef struct OpenDDS::DCPS::TopicDetails OpenDDS::DCPS::EndpointManager<DiscoveredParticipantData_>::TopicDetails’ [-fpermissive]
       typedef TopicDetails TopicDetails;
                            ^
In file included from ../dds/DCPS/DiscoveryBase.h:9:0,
                 from /home/travis/build/objectcomputing/OpenDDS/dds/DCPS/StaticDiscovery.h:12,
                 from /home/travis/build/objectcomputing/OpenDDS/dds/DCPS/Service_Participant.cpp:20:
../dds/DCPS/TopicDetails.h:22:12: error: changes meaning of ‘TopicDetails’ from ‘struct OpenDDS::DCPS::TopicDetails’ [-fpermissive]
     struct TopicDetails {
```

PR #1169, opened 2019-06-21
Modifications to resolve build warnings from Visual Studio

ocielliottc:


dds/idl/rapidjson_generator.cpp
```diff
@@ -264,17 +267,22 @@ namespace {
...
-          ip << "  " << (pt == AST_PredefinedType::PT_octet ? "uint16_t" : temp_type.c_str()) << " " << temp_name << ";\n" <<
+          ip << "  " << (octet_type ? "uint16_t" : temp_type.c_str()) << " " << temp_name << ";\n" <<
           ip << "  if (ss.find(\"0x\") != std::string::npos) {\n" <<
           ip << "    iss >> std::hex >> " << temp_name << ";\n" <<
           ip << "  } else {\n" <<
           ip << "    iss >> " << temp_name << ";\n" <<
           ip << "  }\n" <<
-          ip << "  " << propName << assign_prefix << temp_name << assign_suffix << ";\n" <<
+          ip << "  " << propName << assign_prefix;
+        if (octet_type) {
```
simpsont-oci: I'm guessing it was complaining about truncated values here going from uint16_t to CORBA::Octet. We're only using uint16_t to avoid issues with istringstream and unsigned char (wants to read a character, not a number), but theoretically someone could pass in a value over 255, in which case we just want to cap it at 255. I don't know that the code / tests enforce that though. I'm fine with this change, but in the future we may want an assert or forced cap here or something.

dds/idl/rapidjson_generator.cpp
```diff
@@ -55,6 +55,8 @@ namespace {
         return "Int64";
       case AST_PredefinedType::PT_ulonglong:
         return "Uint64";
+      case AST_PredefinedType::PT_float:
+        return "Float";
```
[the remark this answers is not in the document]
ocielliottc: Yes, this change was made based on the latest rapidjson code.

dds/idl/rapidjson_generator.cpp
ocielliottc: True.  Casting isn't the best...

PR #1170, opened 2019-06-21
Use a macro to switch between std::auto_ptr and std::unique_ptr based…

ocielliottc:
… on the ACE_HAS_CPP11 macro

### DIFF
--- a/dds/DCPS/QOS_XML_Handler/dds_qos.hpp
+++ b/dds/DCPS/QOS_XML_Handler/dds_qos.hpp
@@ -65,6 +65,12 @@ namespace dds
 #include "ace/Null_Mutex.h"
 #include "ace/ace_wchar.h"
 
+#if defined(ACE_HAS_CPP11)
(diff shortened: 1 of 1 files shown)


mitza-oci:
"This code was generated by the XML Schema Compiler" https://github.com/DOCGroup/XSC

ocielliottc:
Ah, I guess I could have read the header comment.  Thanks.

jwillemsen:
XSC has a C++11 generation mode but that generates code which will only work with C++11. Maybe just extend XSC with a similar macro for the regular generation which works with C++03

mitza-oci:
> XSC has a C++11 generation mode but that generates code which will only work with C++11. Maybe just extend XSC with a similar macro for the regular generation which works with C++03

Or, since qos_xml is an optional feature, we could have it only build when C++11 is available.

mitza-oci:
Should this be closed because of #1181?

ocielliottc:
Yes, it should.

PR #1173, opened 2019-06-22
Check the debug level before logging the error

jwillemsen:
    * dds/DCPS/TopicImpl.cpp:

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -507,9 +507,11 @@ DomainParticipantImpl::create_topic_i(
       return new_topic;
 
     } else {
-      ACE_ERROR((LM_ERROR,
-                 ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_topic, ")
-                 ACE_TEXT("assert_topic failed with return value %d.\n"), status));
+      if (DCPS_debug_level >= 1) {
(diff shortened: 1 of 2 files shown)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -507,9 +507,11 @@ DomainParticipantImpl::create_topic_i(
       return new_topic;
 
     } else {
-      ACE_ERROR((LM_ERROR,
-                 ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_topic, ")
-                 ACE_TEXT("assert_topic failed with return value %d.\n"), status));
+      if (DCPS_debug_level >= 1) {
```
mitza-oci: There are a number of places where the error was considered severe enough to log unconditionally.  Looks like it has been this way for a while.  I'm not opposed to changing it, was there a particular test that prompted this?

jwillemsen: I noticed it when testing AXCIOMA with OpenDDS integration but also the CIAO with OpenDDS integration shows the same, see http://scoreboard.ociweb.com/doc_master_tick_linux_gcc_d1o0/index.html

mitza-oci: Right, I agree.  This is similar to a discussion I had with @iguessthislldo on #1094.

PR #1176, opened 2019-06-24
Increase size of the RTPS relay beacon message

brattonj-oci:
The beacon message size was increased to be 4 bytes.  The content of the message
is just zeriozed data.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -2587,6 +2587,9 @@ RtpsUdpDataLink::check_heartbeats()
 void
 RtpsUdpDataLink::send_relay_beacon()
 {
+  static const char beacon_msg[4] = { '\0', '\0', '\0', '\0' };
+  static const size_t beacon_msg_size = sizeof(beacon_msg);
+  
   const bool no_relay = config().rtps_relay_address() == ACE_INET_Addr();
   {
     ACE_GUARD(ACE_Thread_Mutex, g, lock_);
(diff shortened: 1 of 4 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2587,6 +2587,9 @@ RtpsUdpDataLink::check_heartbeats()
 void
 RtpsUdpDataLink::send_relay_beacon()
 {
+  static const char beacon_msg[4] = { '\0', '\0', '\0', '\0' };
```
mitza-oci: This is not a valid RTPS Submessage.  The 4-byte array `{1, 0, 0, 0}` is.  See `struct SubmessageHeader` in RtpsCore.idl.

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -116,7 +117,9 @@ int RelayHandler::handle_input(ACE_HANDLE)
     ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: RelayHandler::handle_input failed to deserialize RTPS header\n"));
     return 0;
   }
-  const auto empty_message = buffer->length() == 0;
+
+  // Beacon messages will be considered empty
```
mitza-oci: True, but the code doesn't do this.  The code below says that every message that happens to be 4 bytes long must be considered to be "empty" / "beacon".  We can afford a 4-byte memcmp to check.

mitza-oci:
The travis build reported:
```
Error: ./dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp:2593: found trailing whitespace
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2587,6 +2587,10 @@ RtpsUdpDataLink::check_heartbeats()
 void
 RtpsUdpDataLink::send_relay_beacon()
 {
+  // Use a pad submessage with for the beacon
+  static const ::CORBA::Octet beacon_msg[] = { 1, 0, 0, 0 };
```
mitza-oci: This could use the `PAD` constant and also if we move it to a common header like `BaseMessageTypes.h` it can be reused by the relay code instead of being repeated there.

PR #1177, opened 2019-06-24
Backwards-Compatible Removal of "Associated Writers" from SEDP Datareader Announcements

simpsont-oci:
This patch will safely remove the poorly-scaling vendor-specific extension for datareader announcements by adding new vendor-specific extension (OpenDDS Participant Flags) to participant announcements. The first (and only, so far) flag will indicate that incoming datareader annoucements are not expected to contain associated datawriters, allowing instances of OpenDDS that recognize this flag to refrain from sending these associations for/to this participant and its datawriters.

### DIFF
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -1161,7 +1161,7 @@ namespace OpenDDS {
           }
 
           // change this if 'writer_active' (above) changes
-          if (call_writer && !call_reader) {
+          if (call_writer && !call_reader && !is_expectant_opendds(reader)) {
             if (DCPS::DCPS_debug_level > 3) {
               ACE_DEBUG((LM_DEBUG,
                          ACE_TEXT("(%P|%t) EndpointManager::match - ")
@@ -1217,6 +1217,8 @@ namespace OpenDDS {
           }
         }
       }
+
+      virtual bool is_expectant_opendds(const GUID_t& endpoint) const = 0;
 
       virtual bool shutting_down() const = 0;
 
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -633,7 +633,7 @@ namespace OpenDDS {
         DCPS::TransportLocatorSeq trans_info_;
         RepoIdSet matched_endpoints_;
         DCPS::SequenceNumber sequence_;
-        RepoIdSet remote_opendds_associations_;
+        RepoIdSet remote_expectant_opendds_associations_;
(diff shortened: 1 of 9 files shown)


dds/DCPS/RTPS/RtpsCore.idl
```diff
@@ -287,6 +287,14 @@ module OpenDDS {
     const ParameterId_t PIDMASK_VENDOR_SPECIFIC = 0x8000;
     const ParameterId_t PIDMASK_INCOMPATIBLE = 0x4000;
 
+    typedef unsigned long ParticipantFlagsBits_t;
```
mitza-oci:
```suggestion
    typedef unsigned long OpenDDSParticipantFlags_t;
```
Possible future naming collision if spec changes... we could use modules but it doesn't seem necessary here.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1207,14 +1207,15 @@ Spdp::wait_for_acks()
...
 bool
-Spdp::is_opendds(const GUID_t& participant) const
+Spdp::is_expectant_opendds(const GUID_t& participant) const
 {
   const DiscoveredParticipantConstIter iter = participants_.find(participant);
   if (iter == participants_.end()) {
     return false;
   }
-  return 0 == std::memcmp(&iter->second.pdata_.participantProxy.vendorId,
+  bool is_opendds = 0 == std::memcmp(&iter->second.pdata_.participantProxy.vendorId,
```
```suggestion
  const bool is_opendds = 0 == std::memcmp(&iter->second.pdata_.participantProxy.vendorId,
```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1161,7 +1161,7 @@ namespace OpenDDS {
           }
 
           // change this if 'writer_active' (above) changes
-          if (call_writer && !call_reader && !is_opendds(reader)) {
+          if (call_writer && !call_reader) {
```
mitza-oci: The call to `association_complete` below could result in two calls to association_complete for the same peer, which the writer is not expecting.

dds/DCPS/DiscoveryBase.h
```diff
@@ -779,7 +779,7 @@ namespace OpenDDS {
             DCPS::WriterIdSeq writer_seq(1);
             writer_seq.length(1);
             writer_seq[0] = removing;
-            lsi->second.remote_opendds_associations_.erase(removing);
+            lsi->second.remote_expectant_opendds_associations_.erase(removing);
```
mitza-oci: If this data isn't actually changing, the call to `write_subscription_data` below should be skipped.

PR #1179, opened 2019-06-26
Log the return code when register instance fails

jwillemsen:


iguessthislldo:
Could you use `OpenDDS::DCPS::retcode_to_string`?

jwillemsen:
> Could you use `OpenDDS::DCPS::retcode_to_string`?

Will do, but that seems to be used in only one place at this moment, looks a more bigger replacement task needs to be done

iguessthislldo:
Yeah, I just added it. In the long term I would like it to replace all the instances of logging the return code integer, but for now I'm content with new instances and replacing instances I happen across.

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1520,7 +1520,8 @@ DataWriterImpl::register_instance_i(DDS::InstanceHandle_t& handle,
   if (ret != DDS::RETCODE_OK) {
     ACE_ERROR_RETURN((LM_ERROR,
                       ACE_TEXT("(%P|%t) ERROR: DataWriterImpl::register_instance_i: ")
-                      ACE_TEXT("register instance with container failed.\n")),
+                      ACE_TEXT("register instance with container failed, returned <%C>.\n"),
+                      retcode_to_string(ret)),
```
mitza-oci:
```suggestion
                      retcode_to_string(ret).c_str()),
```

jwillemsen: Wouldn’t it be easier to let it return a const char*?

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1573,12 +1574,13 @@ DataWriterImpl::register_instance_from_durable_data(DDS::InstanceHandle_t& handl
...
 
-  DDS::ReturnCode_t ret = register_instance_i(handle, move(data), source_timestamp);
+  const DDS::ReturnCode_t ret = register_instance_i(handle, move(data), source_timestamp);
   if (ret != DDS::RETCODE_OK) {
     ACE_ERROR_RETURN((LM_ERROR,
                       ACE_TEXT("(%P|%t) ERROR: DataWriterImpl::register_instance_from_durable_data: ")
-                      ACE_TEXT("register instance with container failed.\n")),
-                      ret);
+                      ACE_TEXT("register instance with container failed, returned <%C>.\n"),
+                      retcode_to_string(ret)),
```
mitza-oci:
```suggestion
                      retcode_to_string(ret).c_str()),
```

dds/DCPS/DataWriterImpl.cpp
mitza-oci: They're all static strings so it seems like that would be good.  Fred?

jwillemsen: See https://github.com/DOCGroup/CIAO/blob/master/connectors/dds4ccm/impl/Utils.h for how we do this in DDS4CCM

PR #1187, opened 2019-06-27
Fix compiler error caused by typename keyword

brattonj-oci:
Removed the typename keyword that was causing compiler errors on some of the
build platforms.  The inconsistent_topic_count_ variable was also converted to
an int to correct some size_t to int conversion warnings.


simpsont-oci:
Pulled size_t / int change into 1185, so closing this as duplicate.

PR #1193, opened 2019-07-01
Check that a topic is enabled when enabling data writer/reader

jrw972:
Problem: Data(Reader|Writer)Impl don't check that topic is enabled

Solution: Return PRECONDITION_NOT_MET if the topic is not enabled.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -1159,6 +1159,10 @@ DataReaderImpl::enable()
     return DDS::RETCODE_PRECONDITION_NOT_MET;
   }
 
+  if (!this->topic_servant_->is_enabled()) {
+    return DDS::RETCODE_PRECONDITION_NOT_MET;
+  }
+
   RcHandle<DomainParticipantImpl> participant = participant_servant_.lock();
   if (participant) {
     dp_id_ = participant->get_id();
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -1159,7 +1159,7 @@ DataReaderImpl::enable()
     return DDS::RETCODE_PRECONDITION_NOT_MET;
   }
 
-  if (!this->topic_servant_->is_enabled()) {
+  if (!topic_servant_->is_enabled()) {
     return DDS::RETCODE_PRECONDITION_NOT_MET;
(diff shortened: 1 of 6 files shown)


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1159,6 +1159,10 @@ DataReaderImpl::enable()
     return DDS::RETCODE_PRECONDITION_NOT_MET;
   }
 
+  if (!this->topic_servant_->is_enabled()) {
```
mitza-oci:
```suggestion
  if (!topic_servant_->is_enabled()) {
```

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1269,6 +1269,10 @@ DataWriterImpl::enable()
     return DDS::RETCODE_PRECONDITION_NOT_MET;
   }
 
+  if (!this->topic_servant_->is_enabled()) {
```
mitza-oci:
```suggestion
  if (!topic_servant_->is_enabled()) {
```

tools/modeling/tests/MessengerDpQos/publisher.cpp
```diff
@@ -51,13 +52,16 @@ int ACE_TMAIN(int argc, ACE_TCHAR** argv)
     } else {
       std::cout << "enabling publisher" << std::endl;
       if (participant->enable() != DDS::RETCODE_OK) {
-        std::cout << "bad return code enabling participant" << std::endl;
+        std::cout << "ERROR; bad return code enabling participant" << std::endl;
```
mitza-oci:
```suggestion
        std::cout << "ERROR: bad return code enabling participant" << std::endl;
```

tools/modeling/tests/SubscriberQos/subscriber.cpp
```diff
@@ -77,6 +77,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR** argv)
 
     DDS::DataReader_var reader = model.reader( Elements::DataReaders::reader);
     DDS::Subscriber_var subscriber = reader->get_subscriber();
+    DDS::Topic_var topic = DDS::Topic::_narrow(reader->get_topicdescription());
```
mitza-oci: TopicDescription reference is leaked here

PR #1201, opened 2019-07-05
publication_matched_status waits for first datareader ack

simpsont-oci:
for rtps reliable associations

### DIFF
--- a/dds/DCPS/DataWriterCallbacks.h
+++ b/dds/DCPS/DataWriterCallbacks.h
@@ -59,6 +59,11 @@ public:
                                    const TransportLocatorSeq& /*locators*/,
                                    DiscoveryListener* /*listener*/) { }
 
+  virtual void register_for_reader_exists(const RepoId& /*participant*/,
+                                          const RepoId& /*writerid*/,
+                                          const RepoId& /*readerid*/,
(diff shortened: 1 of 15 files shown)


iguessthislldo:
This sounds a lot like the `TransportSendListener::first_acknowledged_by_reader` I added for the security fixes. Is there a possibility that these callbacks can be merged?

simpsont-oci:
> This sounds a lot like the `TransportSendListener::first_acknowledged_by_reader` I added for the security fixes. Is there a possibility that these callbacks can be merged?

That was my first approach, but unfortunately first_acknowledged_by_reader is implemented by the TransportClients (DataWriterImpl & Sepd's Writers), but I needed to be able to get the acknack detection somewhere I could call association_complete() somehow, so I opted for extending DiscoveryListener instead, which is what StaticDiscovery uses (and it uses the same return callback (reader_exists)). It took several iterations to find a solution that worked for all situations and that didn't rewrite everything... at some point we really need to refactor all the base class inherited nonsense out of Discovery and the Transport framework and just provide solid interfaces that cover all our cases that each implementation can use however it wants.

simpsont-oci:
And sadly, this won't work for RTPS Discovery + non-RTPS transports. Back to the drawing board...

PR #1202, opened 2019-07-08
Rtps udp uses per remote invoke on start callbacks

simpsont-oci:


simpsont-oci:
I think this is ready.

The (single?) failing java test succeeds multiple runs on my machine with both java 8 and java 11, though both have a ton of run-time warnings from JNI. I'm attempting to run it on a VM to see if I can reproduce the issue.

I will glady add the INFO_DST optimization to pre-association Heartbeats as an optimization after this gets merged and I have a chance to resolve conflicts with my other PR. Doing it now seems like it will just cause more headache than it's worth when it comes to resolving conflicts.

I also think we can now probably remove all of the first_acknowledged_by_reader stuff, since reliable connections don't fully associate until the reader has ACKed, and so we can simply send the volatile keys once in Sedp::association_complete for the volatile writer/reader pair.

Also also, I'm not 100% sure the secure discovery is doing what we want (there may be unencrypted announcements and/or announcements via the unsecure writer) but I tried to preserve existing behavior so we can address it in a separate PR if necessary. I need to double check the spec to make sure the behavior for scenario SC5 is what we'd expect.

mitza-oci:
> The (single?) failing java test succeeds multiple runs on my machine with both java 8 and java 11, though both have a ton of run-time warnings from JNI. I'm attempting to run it on a VM to see if I can reproduce the issue.

The warnings are not relevant to this issue.  I think they have been there for a while based on a JDK version update, and they don't seem to impact anything.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1138,20 +1132,6 @@ Sedp::Task::svc_i(const ParticipantData_t* ppdata)
   }
 #endif
 
-  // Write durable data
```
mitza-oci: Is there any problem with the `defer_match_endpoints_` happening before `write_durable_*_data`?

simpsont-oci: Initial testing seems to indicate we can remove both defer_match_endpoints and first_acknowledged_by_reader. Pushing the removals as individual commits for each so hopefully it will be easy to revert if we discover issues later.

simpsont-oci:
Looks like using docgroup ACE changes the SIGPIPE to a hang of some kind. I'll keep investigating on Monday.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -392,51 +393,7 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
   Writer participant_stateless_message_writer_;
   Writer dcps_participant_secure_writer_;
 
-  /**
-   * Special Case for the Participant Volatile Message Secure Writer, which
-   * performs key exchange. This sends the keys after the first ACKNACK from a
-   * Reader to prevent a loss of keys in the transition between authentication
-   * and key exchange. See docs/design/security.md section titled "Slow
-   * Follower Key Exchange Issue" for details.
```
iguessthislldo: Can you remove or modify this section if it's no longer relevant?

simpsont-oci: This is all deleted, including the RepeatOnceWriter class. Is that what you meant?

simpsont-oci:
Looks like the revert for the TCP changes took us back down to just the failing rtps_disc_tcp test, which we can try to address in a different PR.

dds/DCPS/RTPS/Sedp.h
iguessthislldo: > See docs/design/security.md section titled "Slow Follower Key Exchange Issue" for details.

I put the details in `docs/design/security.md`, so that file needs to be revised.

PR #1206, opened 2019-07-12
Shutdown fails when Service Participant thread is interrupted

jrw972:
An interrupt delievered to the Service Participant thread causes the
run_reactor_event_loop function to return with an "interrupted system
call" error.  An attempt to shutdown will notify the now stopped
reactor.  Since the reactor isn't running,
DomainParticipantImpl::handle_exception is never called.  Since
handle_exception is never called, the condition variable is not
released and shutdown hangs.

The solution is to block signals in the Service Participant thread.


jrw972:
For #1207.

mitza-oci:
The title is confusing, it's not a "Shutdown fails" problem (though that's a symptom).
Is there ACE documentation that this is the right thing to do?
Should this impact RtpsRelay.cpp and DiscoveryBase.h as well?  Relay may have a problem since it may block the last thread that could handle it.
Seems like there should be a small class to encapsulate reusable functionality.

dds/DCPS/ReactorTask.h
```diff
@@ -26,12 +26,12 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
 namespace DCPS {
 
-class OpenDDS_Dcps_Export TransportReactorTask : public virtual ACE_Task_Base,
+class OpenDDS_Dcps_Export ReactorTask : public virtual ACE_Task_Base,
 public virtual RcObject {
 public:
 
-  TransportReactorTask(bool useAsyncSend);
-  virtual ~TransportReactorTask();
+  ReactorTask(bool useAsyncSend);
```
mitza-oci:
```suggestion
  explicit ReactorTask(bool useAsyncSend);
```

dds/DCPS/ReactorTask.inl
```diff
@@ -0,0 +1,63 @@
...
+
+#include "ace/Reactor.h"
+#include "debug.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+ACE_INLINE ACE_Reactor*
+OpenDDS::DCPS::ReactorTask::get_reactor()
+{
+  if (DCPS_debug_level >= 6) {
```
mitza-oci: I realize these are here because the previous version had a similar thing for transport logging, but the "levels" aren't 1-to-1.  I'd rather just not have not, I can't see the benefit.  There are debuggers and dynamic trace tools if someone wants to see every function that gets called, that's not what logging is for.

dds/DCPS/Service_Participant.cpp
```diff
@@ -224,33 +224,38 @@ Service_Participant::instance()
...
-Service_Participant::ReactorTask::svc()
-{
-  Service_Participant* sp = instance();
-  sp->reactor_->owner(ACE_Thread_Manager::instance()->thr_self());
-  sp->reactor_owner_ = ACE_Thread_Manager::instance()->thr_self();
-  this->wait_for_startup();
-  sp->reactor_->run_reactor_event_loop();
-  return 0;
-}
+// int
```
mitza-oci: Remove this

mitza-oci:
Azure CI build with VS2015 failed due to what looks like an actual bug in vcpkg: https://github.com/microsoft/vcpkg/issues/7405

PR #1211, opened 2019-07-21
Consolidate Messenger.idl Instances

simpsont-oci:
Consolidate several Messenger.idl instances in DCPS test applications for code reuse and compile times


simpsont-oci:
I'm not exactly sure what's up with appveyor. It's saying it can't find the Messenger type support include files, which have moved from tests/DCPS/Messenger to tests/DCPS/ConsolidatedMessengerIdl , but dcps_cm should set the mpc libs, after, libpath, and include fields correctly to pick these up, as is evidenced by the fact that all the other builds succeed for all the tests that have been moved over to do the same. I can even see the -I for the include for ConsolidatedMessenger include path, so I'm not entirely sure why it can't find the file. Has the ConsolidatedMessengerIdl not been built for some reason with appveyor's build? Or maybe in a different path than normal?

Aside from this one issue, I think this PR is probably ready to go (barring any glaring issues from the PR builder (non-rtps-related failures). There are probably more IDL files that can get consolidated at some point (the java messenger idl, the security attributes messenger idl, etc) but they're special case enough / have enough differences that I didn't want to pull them into this PR.

mitza-oci:
Update `appveyor.yml` to include the new code.

mitza-oci:
Will be merged in #1226

PR #1226, opened 2019-08-01
Consolidated IDL files from multiple tests

mitza-oci:
Variation of #1211 that leaves `tests/DCPS/Messenger` as its own standalone directory.


iguessthislldo:
I though messenger was being left alone, now I see it isn't. Why does #1211 have a separate library?

mitza-oci:
> 
> 
> I though messenger was being left alone, now I see it isn't. Why does #1211 have a separate library?

Not sure what you're asking about.  There is now common code instead of individual files in `tests/DCPS/{BuiltInTopicTest,Deadline,DPFactoryQos,...}`.

iguessthislldo:
I'm asking why have the Consolidated Messenger Library at all and not just link to DDS_Messenger_Idl?

mitza-oci:
> 
> 
> I'm asking why have the Consolidated Messenger Library at all and not just link to DDS_Messenger_Idl?

I suppose only @simpsont-oci can answer that directly, but I think it makes sense to have it be its own directory so that if a maintainer needs to change it, they know that the change will impact the whole group of tests that share the IDL.

PR #1233, opened 2019-08-06
Fixes for 1067

iguessthislldo:
Mostly Fixes for #1067

Most Significant Changes:
 - Migrated all the IDL in `dds` and `java`, and `tools/rtpsrelay/Relay.idl` to use annotations
- Check for `DCPS_DATA_TYPE` now allows for different whitespace.
- Annotation support in the Java API
- Renamed `dcps_test` MPC base project to `dcps_test_framework`. `dcps_test` was put all the places found that were part of the OpenDDS test suite and used to disable `DCPS_DATA_TYPE` warnings.
- Created common `msvc_bigobj` MPC base project for projects that need the `\bigobj` msvc argument.
- Renamed `dcps_test_lib` to `dcps_idl_only_lib` to better reflect what it is.

# TODO

- [x] Use a special mbp for tests to disable `DCPS_DATA_TYPE` warnings instead of using paths.
- [x] Fix `topic_type_annotation_test` which is a problem with arrays.
- [x] Fix Monitor and BIT Tests which might just be the previous issue.
- [x] Tries to compile nonexistent `QosPolicyCountTypeSupport.java` on Java build.
- [x] Simplify CMake default nested code


dds/idl/be_produce.cpp
```diff
@@ -275,12 +275,49 @@ void postprocess(const char* fn, ostringstream& content,
 
 } // namespace
 
+inline bool
+contains_dcps_data_type(const char* line)
```
mitza-oci: I'd like to avoid having to parse this.  Could we change it to warn based on finding that there was a `#pragma` during the normal codegen passes (a visitor or generator class)?

dds/idl/topic_keys.cpp
```diff
@@ -316,8 +322,12 @@ TopicKeys::Iterator::path()
 void
 TopicKeys::Iterator::path_i(std::stringstream& ss)
 {
+  const std::string& error_msg = "Can't get path for invalid topic key iterator!";
```
```suggestion
  const std::string error_msg = "Can't get path for invalid topic key iterator!";
```
Or just `const char*` and let it construct a `std::string` on demand

PR #1238, opened 2019-08-09
Fixes for RTPS discovery with TCP transport

simpsont-oci:


dds/DCPS/ReactorInterceptor.h
```diff
@@ -68,6 +69,8 @@ class OpenDDS_Dcps_Export ReactorInterceptor : public RcEventHandler {
   virtual ~ReactorInterceptor();
   int handle_exception(ACE_HANDLE /*fd*/);
   int handle_exception_i(ACE_Guard<ACE_Thread_Mutex>& guard);
+  void process_command_queue_i(ACE_Guard<ACE_Thread_Mutex>& guard);
```
mitza-oci: Passing around `ACE_Guard` doesn't seem to be necessary.  The `process_command_queue` function (whatever its name ends up being) should handle it internally.
Pseudocode:
```
while (true) {
  CommandPtr p;
  {
    LOCK
    if (empty) return;
    p = front; pop;
  }
  if (p) run_it;
}
```

dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -154,27 +198,31 @@ DataLink::invoke_on_start_callbacks(const RepoId& local, const RepoId& remote, b
...
-      TransportClient_rch client = it->first.lock();
-      if (client && local == client->get_repo_id() && remote == it->second) {
-        to_call.push_back(*it);
+    OnStartCallbackMap::iterator it = on_start_callbacks_.find(remote);
+    if (it != on_start_callbacks_.end()) {
+      RepoToClientMap::iterator it2 = it->second.find(local);
+      if (it2 != it->second.end()) {
+        client = it2->second;
+        it->second.erase(it2);
+        if (!it->second.size()) {
```
mitza-oci:
```suggestion
        if (it->second.empty()) {
```

dds/DCPS/ReactorInterceptor.h
simpsont-oci: I was trying to follow the convention of handle_exception_i, which is called by more than just handle_exception and presumably uses the guard to signify that the method should have the lock locked when it's called (I don't think it was using the guard either). I think it would also get optimized out by most compilers... but if you want to change it I'm OK with that.

dds/DCPS/transport/tcp/TcpDataLink.cpp
```diff
@@ -425,6 +446,75 @@ OpenDDS::DCPS::TcpDataLink::request_ack_received(const ReceivedDataSample& sampl
...
+
+  for (PairVec::const_iterator it = to_send.begin(); it != to_send.end(); ++it) {
+    send_association_msg(it->first, it->second);
+  }
+}
+
+void
+OpenDDS::DCPS::TcpDataLink::send_association_msg(const RepoId& local, const RepoId& remote)
+{
+  DCPS::GuidConverter cl(local);
```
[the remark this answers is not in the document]
simpsont-oci: Nope, these were there for some debugging printouts. Good catch.

dds/DCPS/ReactorInterceptor.h
simpsont-oci: Also, process_command_queue seems to always assume that the lock will be locked at the beginning and end of the method (see wait(), which calls it via handle_exception_i), so the reverse lock makes sense. I agree there's no real purpose to passing the guard (unless perhaps we made both public and private versions, and even then it's just a reminder to developers they should have a guard in place before they call the internal version).

dds/DCPS/ReactorInterceptor.h
```diff
@@ -35,14 +35,15 @@ class OpenDDS_Dcps_Export ReactorInterceptor : public RcEventHandler {
...
   bool should_execute_immediately();
-  void process_command_queue();
 
   template<typename T>
   void execute_or_enqueue(T& t)
   {
     if (should_execute_immediately()) {
-      ACE_GUARD(ACE_Thread_Mutex, guard, this->mutex_);
-      process_command_queue();
+      {
```
jrw972: From this code, I'm inferring that you want strict FIFO ordering so that anything that was deferred is executed first.  This makes sense.

simpsont-oci:
Also worth noting that all the tests were passing on my laptop and on my VM when I submitted this PR. I don't know where the 14 test failures on the PR builder are coming from, but if it's not from recent things in master that are getting merged in, duplicating the issues is probably going to be a pain.

dds/DCPS/ReactorInterceptor.h
simpsont-oci: I just needed the ReactorIntercepter's mutex to be unlocked during the call to Command::execute, since some commands may need to call back into the ReactorInterceptor. I think the FIFO behavior remains unchanged here, and I agree that it makes sense.

dds/DCPS/transport/tcp/TcpDataLink.cpp
```diff
@@ -447,6 +537,25 @@ OpenDDS::DCPS::TcpDataLink::receive_strategy()
 {
   return static_rchandle_cast<OpenDDS::DCPS::TcpReceiveStrategy>(receive_strategy_);
 }
+int
+OpenDDS::DCPS::TcpDataLink::make_reservation(const RepoId& remote_subscription_id,
+                                             const RepoId& local_publication_id,
+                                             const TransportSendListener_wrch& send_listener)
+{
+  int result = OpenDDS::DCPS::DataLink::make_reservation(remote_subscription_id, local_publication_id, send_listener);
```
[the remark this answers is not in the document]
mitza-oci: Suggestion was to not use redundant prefix `OpenDDS::DCPS::` as it adds to complexity of reading the code without actually doing anything different. (Also on line 553)

PR #1240, opened 2019-08-09
Fixes for CI CMake Builds

iguessthislldo:
- `configure`: Backing up Configuration Files:
  - Add `--no-backup` to `configure` to overwrite files
  - Generate timestamp just once
- Fix `run_ci_tests.pl` not failing when cmake or running the tests failed.
- Set `CMAKE_CXX_COMPILER` to `OPENDDS_COMPILER` in CMakeLists

@mitza-oci, @mmcparlane I'm wondering if we should just set the compiler on Unixes if we are already setting the standard.


iguessthislldo:
Looks like SecurityWithoutFeatures was always failing before. I will require make the CMakeLists require `OPENDDS_OWNDERSHIP_PROFILE` to be true and skip the CMake tests for SecurityWithoutFeatures.

PR #1247, opened 2019-08-16
Fix opendds_idl generate error when  #include <tao/ULongSeq.pidl> in the file.

qiup-Q:
will generate code like `#include <tao/ULongSeqTypeSupportImpl>` and it will cause compiled error.

### DIFF
--- a/dds/idl/be_global.cpp
+++ b/dds/idl/be_global.cpp
@@ -514,7 +514,7 @@ namespace {
         0 == ACE_OS::strcasecmp(idl.c_str() + len - 5, ".pidl")) {
       base_name.assign(idl.c_str(), len - 5);
       size_t slash = base_name.find_last_of("/\\");
-      if (slash != std::string::npos && slash > 3 && base_name.size() > 3
+      if (slash != std::string::npos && slash > =3 && base_name.size() > 3
           && base_name.substr(slash - 3, 3) == "tao"
           && base_name.substr(base_name.size() - 3) == "Seq") {
         base_name = "dds/CorbaSeq/" + base_name.substr(slash + 1);


mitza-oci:
Please update tests to demonstrate the problem.  In particular, the IDL files in `tests/DCPS/CorbaSeq` seem to be covering this already.

qiup-Q:
Here is the code logic mistake.
```c++
std::string transform_referenced(const std::string& idl, const char* suffix)
  {
    const size_t len = idl.size();
    string base_name;
    if (len >= 5 && 0 == ACE_OS::strcasecmp(idl.c_str() + len - 4, ".idl")) {
      base_name.assign(idl.c_str(), len - 4);

      // THE BUG IS HERE: when idl == "tao/ULongSeq.pidl"; all the conditions is TRUE but "slash > 3"
      // in actually , slash == 3 , so when this condition covering ?
    } else if (len >= 6 &&
        0 == ACE_OS::strcasecmp(idl.c_str() + len - 5, ".pidl")) {
      base_name.assign(idl.c_str(), len - 5);
      size_t slash = base_name.find_last_of("/\\");
      if (slash != std::string::npos && slash > 3 && base_name.size() > 3
          && base_name.substr(slash - 3, 3) == "tao"
          && base_name.substr(base_name.size() - 3) == "Seq") {
        base_name = "dds/CorbaSeq/" + base_name.substr(slash + 1);
      }
    }

    return base_name + suffix;
  }
```
In actually, I don't use MPC to generate Makefile or process idl files,I do some hard work  transform .mpc and mpb to QMAKE files , and When I process the monitor.idl , the error cause , but it's alright when i use MPC.

mitza-oci:
Thanks for the additional explanation, however that didn't address what I was asking for.  What can be changed in the tests in order to see the failure on master, which would then pass on your branch with a modified `be_global.cpp`?

PR #1256, opened 2019-08-26
Use ACE Monotonic Time

iguessthislldo:
- Have OpenDDS use monotonic timing where possible to protect OpenDDS from changes to the system clock.
- Use new types inspired by the `std::chrono` library to replace `ACE_Time_Value` and enforce time logic wherever possible. These are `TimeDuration`, `MonotonicTimePoint`, and `SystemTimePoint`.
- Add missing library linking to targets in CMake package.

**TODO:**

- [x] Fix LivelinessKeepAlive Test
- [x] Fix TimeBasedFilter Test
- [x] Fix whatever the CMake Test is doing on MacOS
- [x] Go through and check that initialization of `MonotonicTimePoint`s that are members is being done properly now. This was not being done in `SubscriptionInstance` and was causing Time Based Filter Test to fail.


jwillemsen:
Would prefer to have the fuzz check everywhere so that it can be ran locally, maybe add it as check to the base ACE fuzzer so that all users can use it

dds/DCPS/RTPS/ICE/Task.h
```diff
@@ -13,24 +13,27 @@
...
 #include "dds/Versioned_Namespace.h"
+#include "dds/DCPS/TimeTypes.h"
 #include <ace/Time_Value.h>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace ICE {
 
+using OpenDDS::DCPS::MonotonicTimePoint;
```
mitza-oci: See comment in Spdp.h (this applies to a few other places, I'm going to stop here)

dds/DCPS/RcEventHandler.h
```diff
@@ -30,13 +30,13 @@ class RcEventHandler
 
   }
 
-  ACE_Event_Handler::Reference_Count add_reference()
+  virtual ACE_Event_Handler::Reference_Count add_reference()
```
mitza-oci: It is a virtual member function with or without the keyword `virtual`.  Is there any reason to add it?  Unlike `override` in C++11, it doesn't help find errors.

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -2216,27 +2208,22 @@ DomainParticipantImpl::LivelinessTimer::~LivelinessTimer()
 void
 DomainParticipantImpl::LivelinessTimer::add_adjust(OpenDDS::DCPS::DataWriterImpl* writer)
 {
-  ACE_GUARD(ACE_Thread_Mutex,
-            guard,
-            this->lock_);
+  ACE_GUARD(ACE_Thread_Mutex, guard, this->lock_);
```
mitza-oci:
```suggestion
  ACE_GUARD(ACE_Thread_Mutex, guard, lock_);
```

dds/DCPS/RcEventHandler.h
iguessthislldo: I was trying to figure out what was going on with macOS and I found that it doesn't have any effect after trying it out, but forgot to remove it.

dds/DCPS/WaitSet.h
```diff
@@ -81,7 +82,7 @@ class OpenDDS_Dcps_Export WaitSet
 
   ACE_Recursive_Thread_Mutex lock_;
   ACE_Condition_Recursive_Thread_Mutex cond_;
-  ACE_Atomic_Op<ACE_Thread_Mutex, long> waiting_;
+  ACE_Atomic_Op<ACE_Thread_Mutex, bool> waiting_;
```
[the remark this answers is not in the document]
iguessthislldo: Alright, I will change it but will document that it should be treated as a boolean value.

dds/DCPS/security/AccessControlBuiltInImpl.cpp
```diff
@@ -1990,23 +1984,25 @@ bool AccessControlBuiltInImpl::RevokePermissionsTimer::start_timer(const ACE_Tim
...
 {
-  ACE_UNUSED_ARG(tv);
   ACE_GUARD_RETURN(ACE_Thread_Mutex,
       guard,
       this->lock_,
       -1);
 
-  ::DDS::Security::PermissionsHandle *pm_handle = (::DDS::Security::PermissionsHandle *)arg;
+  const ::DDS::Security::PermissionsHandle* pm_handle =
+    reinterpret_cast<const ::DDS::Security::PermissionsHandle*>(arg);
```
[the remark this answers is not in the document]
iguessthislldo: Alright. I feel like it shouldn't though, that since `void*` is a circumvention of the type system like `reinterpret_cast`.

PR #1262, opened 2019-08-29
ReactorInterceptor wait is broken

jrw972:
To support a straight enqueue while processing commands in the
ReactorInterceptor, it was necessary to release the queue lock before
each command was executed.  This broke the existing semantics of wait
because the following sequence was possible:

Non-reactor thread: execute_or_enqueue (command is placed on the
queue)
Reactor thread: command is take off the queue but not executed yet
Non-reactor thread: wait (returns immediately since the queue is
empty)
... wait returned without the command being executed ...

Solution: enqueue and execute_or_enqueue return a Command with a wait
method.  A thread can use this wait on a specify command to complete.
The ReactorInterceptor signals the Command to unblock any waiting thread.

### DIFF
--- a/dds/DCPS/ReactorInterceptor.h
+++ b/dds/DCPS/ReactorInterceptor.h
@@ -16,6 +16,7 @@
 #include "RcEventHandler.h"
 #include "dcps_export.h"
 #include "unique_ptr.h"
+#include "RcHandle_T.h"
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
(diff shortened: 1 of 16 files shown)


dds/DCPS/ReactorInterceptor.h
```diff
@@ -62,15 +65,26 @@ class OpenDDS_Dcps_Export ReactorInterceptor : public RcEventHandler {
...
+    if (immediate) {
       process_command_queue_i();
     }
     return command;
   }
 
-  CommandPtr enqueue(Command* c, bool immediate = false)
+  CommandPtr enqueue(Command* c)
+  {
+    assert(c);
```
mitza-oci: `OPENDDS_ASSERT` to comply with safety profile requirements

PR #1274, opened 2019-09-09
SPDP participant sequence check

wilsongOCI:
GuidGenerator Code was modified to add a random generated number to the counter value.  The counter value will increment from the initial random generated number.     
                                                                                                                                                                                                           Added RTPSDiscovery configuration value  MaxSpdpSequenceMsgResetChecks.  This value is the number of sequence  numbers that must be received that are less than the initial last                                                       sequence number value before a discovered participant is removed.    
                                                                                                                                                                           Added Spdp code changes to check for a sequence number reset. A reset occurs when a defined number of sequence numbers are received that are  less than than an initial last sequence value.                                                                          On a reset the discovered participant will be removed and rediscovered. 

tests\transport\spdp contains the test code for the participant sequence check


tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+  seq = first_seq;
+
+  for (;;) {
+    if (!part1.send_data(test_part_guid.entityId, seq, plist, send_addr)) {
+      return false;
+    }
+
+    reactor_wait();
+
+    for (int i = 0; i < 3; i++)
```
iguessthislldo:
```suggestion
    for (int i = 0; i < 3; i++) {
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -480,35 +485,73 @@ Spdp::handle_participant_data(DCPS::MessageId id, const ParticipantData_t& cpdat
...
   }
 }
+bool
+Spdp::validateSequenceNumber(const DCPS::SequenceNumber& seq, DiscoveredParticipantIter& iter)
+{
+  if ((iter->second.last_seq_.getHigh() == ACE_INT32_MAX) &&
+     (iter->second.last_seq_.getLow() == ACE_UINT32_MAX)) {
+    return true;
+  } else {
+    if (seq.getValue() < iter->second.last_seq_.getValue()) {
```
iguessthislldo:
```suggestion
    if (seq < iter->second.last_seq_) {
```

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+    }
+  };
+
+  int irtn = OpenDDS::RTPS::ParameterListConverter::to_param_list(pdata, plist);
+
+  if (irtn < 0) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+      ACE_TEXT("spdp_transport - run_test - ")
+      ACE_TEXT("failed to convert from SPDPdiscoveredParticipantData ")
+     ACE_TEXT("to ParameterList\n")));
```
iguessthislldo:
```suggestion
      ACE_TEXT("to ParameterList\n")));
```

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+     return false;
```
iguessthislldo:
```suggestion
    return false;
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -480,35 +485,73 @@ Spdp::handle_participant_data(DCPS::MessageId id, const ParticipantData_t& cpdat
...
 
-    // Must unlock when calling into part_bit() as it may call back into us
-    ACE_Reverse_Lock<ACE_Thread_Mutex> rev_lock(lock_);
+    // Checks if sequence numbers are increasing
+
+    bool ValidSeq;
+
+    // Check if subroutne call is from Sedp Security
+
+    if (currentSeq.getValue() != 0) {
```
iguessthislldo: I think you should get rid of `ValidSeq`, make this check part of `validateSequenceNumber()`, and put `validateSequenceNumber(currentSeq, iter)` in the `if` evaluation.

tests/transport/spdp/rtps_disc.ini
```diff
@@ -0,0 +1,18 @@
+[common]
+DCPSDebugLevel=10
+DCPSTransportDebugLevel=5
+DCPSGlobalTransportConfig=$file
+DCPSMonitor=1
```
iguessthislldo: I don't think you need monitor enabled.

tests/transport/spdp/rtps_disc.ini
```diff
@@ -0,0 +1,18 @@
...
+DCPSInfoRepo=127.0.0.1:12345
```
iguessthislldo: And this is a SPDP test, you can get rid of this.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+  const DDS::Subscriber_var sVar;
+
+  spdp->init_bit(sVar);
+
+  SequenceNumber_t first_seq = {0, 1}, seq = first_seq;
+  bool bfirst = true;
+
+  // Test for performing sequence reset.
+
+  for (;;)
```
iguessthislldo:
```suggestion
  for (;;) {
```

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -392,24 +393,35 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
...
+            } else {
+              ACE_ERROR_RETURN((LM_ERROR,
+                                ACE_TEXT("(%P|%t) RtpsDiscovery::Config::discovery_config(): ")
+                                ACE_TEXT("Invalid entry (%C) for MaxSpdpTimerPeriod in ")
+                                ACE_TEXT("[rtps_discovery/%C] section.\n"),
+                                string_value.c_str(), rtps_name.c_str()), -1);
+            }
+        } else if (name == "MaxSpdpSequenceMsgResetChecks") {
+            const OPENDDS_STRING& string_value = it->second;
+            u_short value;
```
iguessthislldo: @mitza-oci This definitely not something that needs to be changed right now, since we're already using `u_short` and `u_long` but this is the first time I've really noticed them. I tried googling them and they aren't standard C++ primitive names and we don't seem to be handling them in ACE so I don't feel comfortable with them from a portability perspective. However they probably won't ever be a problem though since it seem so many platforms seem to define them, but I just want to complain I guess...

mitza-oci: They are defined in ACE (ace/os_include/os_stdint.h), so no problem using them here.  Wouldn't necessarily want to see them in public interfaces.

iguessthislldo: Indeed they are. Looks like I miss typed `typedef` when greping my grep for `u_long` in ACE. I'm officially withdrawing my complaint.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+  ACE_SOCK_Dgram test_part_sock;
+  ACE_INET_Addr test_part_addr;
+  if (!open_appropriate_socket_type(test_part_sock, test_part_addr)) {
+    std::cerr << "ERROR: run_test() unable to open test_part_sock" << std::endl;
+    exit(1);
+  }
+  test_part_sock.get_local_addr(test_part_addr);
+#ifdef OPENDDS_SAFETY_PROFILE
+  part1_addr.set(test_part_addr.get_port_number(), "127.0.0.1");
```
iguessthislldo:
```suggestion
  test_part_addr.set(test_part_addr.get_port_number(), "127.0.0.1");
```

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+#include "dds/DCPS/transport/rtps_udp/RtpsUdpInst.h"
+#ifdef ACE_AS_STATIC_LIBS
+#include "dds/DCPS/transport/rtps_udp/RtpsUdp.h"
+#endif
+
+#include "dds/DCPS/transport/framework/NetworkAddress.h"
+
+#include "dds/DCPS/RTPS/BaseMessageTypes.h"
+#include "dds/DCPS/RTPS/GuidGenerator.h"
```
iguessthislldo: `GuidGenerator.h` apparently doesn't have header guards and it looks like pragma once doesn't look like it works on safety profile, so it conflicts from being included in `RtpsDiscovery.h`. I would suggest adding header guards to `GuidGenerator.h`.

dds/DCPS/RTPS/Spdp.h
```diff
@@ -98,7 +98,9 @@ class OpenDDS_Rtps_Export Spdp : public DCPS::LocalParticipant<Sedp> {
   void handle_participant_crypto_tokens(const DDS::Security::ParticipantVolatileMessageSecure& msg);
 #endif
 
-  void handle_participant_data(DCPS::MessageId id, const ParticipantData_t& pdata);
+  void handle_participant_data(DCPS::MessageId id, const ParticipantData_t& pdata, const SequenceNumber_t seq);
```
mitza-oci:
```suggestion
  void handle_participant_data(DCPS::MessageId id, const ParticipantData_t& pdata, const SequenceNumber_t& seq);
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -480,35 +486,68 @@ Spdp::handle_participant_data(DCPS::MessageId id, const ParticipantData_t& cpdat
...
+    return true;
+  }
+
+  if ((iter->second.last_seq_.getHigh() == ACE_INT32_MAX) &&
+     (iter->second.last_seq_.getLow() == ACE_UINT32_MAX)) {
+    return true;
+  } else {
+    if (seq.getValue() < iter->second.last_seq_.getValue()) {
+      ++iter->second.SeqResetChkCount_;
+      //iter->second.seqResetCandidate_.setValue(seq.getValue());
```
```suggestion
```

tests/transport/spdp/rtps_disc.ini
```diff
@@ -0,0 +1,14 @@
...
+DiscoveryConfig=uni_rtps
+
+[rtps_discovery/uni_rtps]
+SedpMulticast=0
+ResendPeriod=2
+MaxSpdpSequenceMsgResetChecks=5
+
+[transport/the_rtps_transport]
+transport_type=rtps_udp
+use_multicast=0
```
mitza-oci:
```suggestion
use_multicast=0
```
(current file doesn't end in a newline)

tests/transport/spdp/run_test.pl
```diff
@@ -0,0 +1,19 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
```
mitza-oci: Once the test validation code is ready, make sure to add this `run_test.pl` to the overall list in `dcps_tests.lst`.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1749,7 +1752,9 @@ void Sedp::process_discovered_writer_data(DCPS::MessageId message_id,
       }
 
     } else if (qosChanged(iter->second.writer_data_.ddsPublicationData,
-                          wdata.ddsPublicationData)) { // update existing
+                          wdata.ddsPublicationData))
```
mitza-oci: I can't tell why this is changing.  Was there some automated reformatting run on it?  If so we can use git to restore it.

iguessthislldo:
Also it doesn't look like you saw the review comments in the middle of my review. If you click on the box that says "5 hidden conversations" and "Load more...". It should show you my entire review.

dds/DCPS/RTPS/Sedp.cpp
wilsongOCI: Is this just an issue with the line 1755?

mitza-oci: I think there were a few of them, but this was more visible since the comment and the `{` moved.  I'd suggest reading through the overall diff (in the web interface or `git diff master`) and see if there are changes you didn't intend to have.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+  DDS_TEST spdp_friend(spdp, test_part_guid);
+
+  const DDS::Subscriber_var sVar;
+  spdp->init_bit(sVar);
+
+  // Create a Parameter List
+  OpenDDS::RTPS::ParameterList plist;
+
+  BuiltinEndpointSet_t availableBuiltinEndpoints =
```
mitza-oci:
```suggestion
  const BuiltinEndpointSet_t availableBuiltinEndpoints =
```

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,485 @@
...
+  test_part_sock.get_local_addr(test_part_addr);
+  test_part_addr.set(test_part_addr.get_port_number(),
+#ifdef OPENDDS_SAFETY_PROFILE
+    "127.0.0.1"
+#else
+    "localhost"
+#endif
+    );
+
+  ACE_INET_Addr send_addr("239.255.0.1:7400");
```
mitza-oci: Need to determine if we should be changing port numbers (see OPENDDS_RTPS_DEFAULT_D0) or binding to loopback only, etc., to make sure two concurrent tests don't conflict.

dds/DCPS/RTPS/GuidGenerator.h
```diff
@@ -63,10 +66,13 @@ class OpenDDS_Rtps_Export GuidGenerator {
   pid_t pid_;
   ACE_Thread_Mutex counter_lock_;
   ACE_UINT16 counter_;
+  ACE_UINT16 init_counter_;
```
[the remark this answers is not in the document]
iguessthislldo: Yes, it can be local now that you mention it.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,485 @@
...
+  , guid_(make_guid(guid.guidPrefix, ENTITYID_PARTICIPANT))
+  {
+  }
+
+  bool check_for_participant(bool expect_participant)
+  {
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, g, spdp_->lock_, true);
+    const bool error = spdp_->has_discovered_participant(guid_) != expect_participant;
+    if (error) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("Error: expected to %Cfind participant, but did%C\n"),
```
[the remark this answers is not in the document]
iguessthislldo: I had to google what this meant and it looks like the rule against it is considered obsolete by most guidelines. Regardless, would you prefer "Didn't expect to find the participant, but did" and "Expected to find the participant, but didn't"?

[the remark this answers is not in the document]
mitza-oci: I was thinking "expected not to find" would read better than "expected to not find", but perhaps this is obsolete.  Although the succinctness and use of trinary are both good as far as I'm concerned, in terms of readability/maintainability it may be better to just write out both strings, something like `"Error: %C\n", expected ? "message one" : "message two"`

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,485 @@
...
+  void remove_participant()
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, spdp_->lock_);
+    Spdp::DiscoveredParticipantIter iter = spdp_->participants_.find(guid_);
+    if (iter != spdp_->participants_.end()) {
+      spdp_->remove_discovered_participant(iter);
+    }
+  }
+
+  const RcHandle<Spdp>(spdp_);
```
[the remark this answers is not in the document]
iguessthislldo: ... I'm not sure either, but I'm kinda impressed with myself.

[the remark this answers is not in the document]
iguessthislldo: It looks like you can put parentheses around a fair number of names in gcc.

```C++
#include <stdio.h>

class X {
public:
    (X)(int (value)) : value_(value) {
    }

    void (print)() {
        (printf)("%d\n", value_);
    }
    int((value_));
};

int (main)() {
    X (x)(3);
    (x).print();
}
```

You can't do it to class name while declaring the name and to methods.

tests/transport/spdp/spdp_transport.cpp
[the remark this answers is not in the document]
iguessthislldo: I'm actually not picky about the phrasing as along it conveys the meaning, so is this good with you?:

```c++
ACE_ERROR((LM_ERROR, ACE_TEXT("Error: %C\n"), expect_participant ?
  "Expected to find the participant, but didn't" : 
  "Expected not to find the participant, but did");
```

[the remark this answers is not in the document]
mitza-oci: Looks good

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,485 @@
...
+    const DataSubmessage ds = {
+      {DATA, FLAG_E | FLAG_D, 0},
+      0, DATA_OCTETS_TO_IQOS, ENTITYID_UNKNOWN, writer, seq, ParameterList()
+    };
+
+    size_t size = 0, padding = 0;
+    gen_find_size(hdr_, size, padding);
+    gen_find_size(ds, size, padding);
+
+    size += sizeof(plist);
```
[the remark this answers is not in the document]
iguessthislldo: I missed that's what was going on. I'm guessing since it's an IDL type it would use `gen_find_size` like everything else?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1144,7 +1144,7 @@ Sedp::Task::svc_secure_i(DCPS::MessageId id,
                          const Security::SPDPdiscoveredParticipantData* ppdata)
 {
   DCPS::unique_ptr<const Security::SPDPdiscoveredParticipantData> pdata(ppdata);
-  spdp_->handle_participant_data(id, *pdata);
+  spdp_->handle_participant_data(id, *pdata, DCPS::SequenceNumber(0));
```
mitza-oci:
```suggestion
  spdp_->handle_participant_data(id, *pdata, DCPS::SequenceNumber::ZERO());
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -534,9 +549,10 @@ Spdp::data_received(const DataSubmessage& data, const ParameterList& plist)
     return;
   }
 
-  DCPS::MessageId msg_id = (data.inlineQos.length() && disposed(data.inlineQos)) ? DCPS::DISPOSE_INSTANCE : DCPS::SAMPLE_DATA;
-
-  handle_participant_data(msg_id, pdata);
+  handle_participant_data(
+    (data.inlineQos.length() && disposed(data.inlineQos)) ?  DCPS::DISPOSE_INSTANCE : DCPS::SAMPLE_DATA,
+    pdata,
+    DCPS::SequenceNumber(data.writerSN.high, data.writerSN.low));
```
mitza-oci: It is slightly more verbose, but I'd rather have this conversion from the RTPS high/low sequence numbers to `SeqeunceNumber` as it appears in `RtpsUdpDataLink::RtpsReader::process_data_i`.  I think this change (along with the note above) would allow `SequenceNumber.h` to remain unchanged in this PR.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -0,0 +1,501 @@
...
+      spdp_->remove_discovered_participant(iter);
+    }
+  }
+
+  bool get_multicast_address(ACE_INET_Addr& addr) {
+    if (!spdp_->tport_) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("ERROR: Null SPDP Transport\n")));
+      return false;
+    }
+    OPENDDS_SET(ACE_INET_Addr)& send_addrs = spdp_->tport_->send_addrs_;
```
mitza-oci: I'm not sure about this assumption, especially with concurrent work ongoing with RtpsRelay.  Instead of checking `send_addrs_`, I suggest having the SpdpTransport constructor store the multicast address directly as a data member.

PR #1278, opened 2019-09-16
Add an application participant to RtpsRelay

jrw972:


tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -281,21 +281,65 @@ void HorizontalHandler::process_message(const ACE_INET_Addr&,
...
+  expiration_addr_map_.insert(std::make_pair(expiration, remote_str));
+
   if (is_beacon_message) {
     return;
   }
 
+  if (a_src_guid == application_participant_guid_) {
+    if (application_participant_addr_.empty()) {
+      application_participant_addr_ = remote_str;
+      ACE_DEBUG((LM_INFO, "(%P|%t) %N:%l SpdpHandler::handle_input application participant %s has address %s\n", application_participant_guid_.c_str(), application_participant_addr_.c_str()));
```
mitza-oci:
```suggestion
      ACE_DEBUG((LM_INFO, "(%P|%t) %N:%l SpdpHandler::handle_input application participant %C has address %C\n", application_participant_guid_.c_str(), application_participant_addr_.c_str()));
```

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -281,21 +281,65 @@ void HorizontalHandler::process_message(const ACE_INET_Addr&,
...
+    for (const auto p : addr_expiration_map_) {
+      if (p.first != application_participant_addr_) {
+        enqueue_message(p.first, a_buffer);
+      }
+    }
+
+    return;
+  }
+
+  // Always send to the application participant.
```
mitza-oci: Does this `Always` include the cases where line 335 returns?

tools/rtpsrelay/RelayHandler.h
```diff
@@ -78,15 +78,31 @@ class HorizontalHandler : public RelayHandler {
...
+              GroupTable& a_group_table,
+              RoutingTable& a_routing_table,
+              const OpenDDS::DCPS::RepoId& application_participant_guid,
+              const ACE_Time_Value& lifespan,
+              const ACE_Time_Value& purge_period);
 
 private:
   GroupTable& mutable_group_table_;
+  std::string application_participant_addr_;
+  std::string const application_participant_guid_;
```
mitza-oci:
```suggestion
  const std::string application_participant_guid_;
```

mitza-oci: Yes, we are using west const.

tools/rtpsrelay/RelayHandler.cpp
jrw972: No.  If the message was from the application participant it does not need to be reflected.

mitza-oci: I guess I just found this comment a little confusing.  It may be better without the comment, or with a different comment.  Or maybe it's just me.

These types of comments (line 328 and this one) look like they were placeholders that didn't get removed.

PR #1280, opened 2019-09-16
Security: RTPS-level protection

mitza-oci:
Along with other updates to DDS Security plugins, key exchange for participant-level symmetric keys, and transport support for parsing the right messages and dropping the wrong ones


dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1608,10 +1652,156 @@ bool CryptoBuiltInImpl::verify(const KeyMaterial& master, Session& sess,
...
+      }
+    }
+  }
+
+  if (!haveCryptoHeader || !haveCryptoFooter) {
+    return CommonUtilities::set_security_error(ex, -9, 0, "Failed to find SRTPS_PREFIX/POSTFIX wrapper");
+  }
+
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  const KeySeq& keyseq = keys_[sending_participant_crypto];
```
simpsont-oci: Do we need a check here first to see if the key exists in keys_ to avoid adding a default initialized one?

dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1774,36 +1958,41 @@ bool CryptoBuiltInImpl::decode_serialized_payload(
   mb_in.wr_ptr(encoded_buffer.length());
   Serializer de_ser(&mb_in, Serializer::SWAP_BE, Serializer::ALIGN_CDR);
   CryptoHeader ch;
-  de_ser >> ch;
+  if (!(de_ser >> ch)) {
+    return CommonUtilities::set_security_error(ex, -3, 4, "Failed to deserialize CryptoHeader");
+  }
 
   const KeySeq& keyseq = keys_[sending_datawriter_crypto];
```
simpsont-oci: See question above. Should we check to see if the key exists in key_ here before accidentally adding a default one?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -659,10 +663,15 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink {
...
-
   typedef OPENDDS_MAP_CMP(RepoId, DDS::Security::NativeCryptoHandle,
                           GUID_tKeyLessThan)::const_iterator PeerHandlesCIter;
+  PeerHandlesMap peer_crypto_handles_;
+
+  typedef OPENDDS_MAP_CMP(RepoId, DDS::Security::EndpointSecurityAttributesMask,
+                          GUID_tKeyLessThan) EndpointSecurityAttributesMap;
+  EndpointSecurityAttributesMap endpoint_security_attributes_;
+
+  const ACE_INET_Addr placeholder_address_;
```
simpsont-oci: This could be static and/or anonymous in the .cpp

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -132,9 +151,135 @@ RtpsUdpReceiveStrategy::receive_bytes(iovec iov[],
...
+    }
+
+    DDS::OctetSeq plain;
+    SecurityException ex = {"", 0, 0};
+    if (!crypto->decode_rtps_message(plain, encoded, receiver, sender, ex)) {
+      if (security_debug.warn) {
+        ACE_DEBUG((LM_WARNING, "decode_rtps_message SecurityException [%d.%d]: %C\n",
+                   ex.code, ex.minor_code, ex.message.in()));
+      }
+      if (ex.code == -10 && ex.minor_code == 1) {
```
simpsont-oci: If we're checking these security exception codes, maybe they need to be enums / constants for clarity?

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -319,6 +343,52 @@ RtpsUdpSendStrategy::encode_payload(const RepoId& pub_id,
...
+  int idx = 0; // unused
+  SecurityException ex = {"", 0, 0};
+  if (crypto->encode_rtps_message(encoded_rtps_message, plain_rtps_message,
+                                  send_handle, recv_handles, idx, ex)) {
+    Message_Block_Ptr out(new ACE_Message_Block(encoded_rtps_message.length()));
+    const char* raw = reinterpret_cast<const char*>(encoded_rtps_message.get_buffer());
+    out->copy(raw, encoded_rtps_message.length());
+    return out.release();
+  }
+  if (ex.code == 0 && ex.minor_code == 0) {
```
simpsont-oci: Should we make a shorthand method for checking if the security exception hasn't been set? And does SecurityException initialize to undefined or a non-error state?

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -355,6 +425,13 @@ namespace {
                  ex.message.in()));
     }
   }
+
+  void check_stateless_volatile(EntityId_t writerId, bool& stateless_or_volatile)
```
simpsont-oci: I see this being called with both sender and receiver EntityIds below. Should we be checking against the STATELESS_READER and _VOLATILE_SECURE_READER EntityIds here as well?

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -567,8 +638,7 @@ RtpsUdpSendStrategy::pre_send_packet(const ACE_Message_Block* plain)
       break;
     }
 
-    if (!ok || (octetsToNextHeader == 0 && msgId != RTPS::PAD
-                && msgId != RTPS::INFO_TS)) {
+    if (!ok || (octetsToNextHeader == 0 && msgId != RTPS::PAD && msgId != RTPS::INFO_TS)) {
```
simpsont-oci: This is the third loop I've seen for walking through RTPS Submessages in this PR alone. Does it make sense to try to make a generic function / object to capture this logic in a generic way?

tests/security/CryptoTransformTest.cpp
```diff
@@ -303,60 +301,62 @@ TEST_F(CryptoTransformTest, encode_rtps_message_NegativeIndex)
...
-      EXPECT_EQ((n + 1), index);
-    } else {
-      // Out of bounds test
-      EXPECT_FALSE(get_inst().encode_rtps_message(
-        output, get_buffer(), handle, get_participants(), index, ex));
-      EXPECT_EQ(n, index);
-    }
-  }
-}
+// Multiple receiver handles is not currently supported in RTPS Message encoding
```
simpsont-oci: Does this test need to be rewritten, or is this OK pending future work?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4013,36 +4014,46 @@ Sedp::add_security_info(const DCPS::TransportLocatorSeq& locators,
...
-  DDS::Security::DatawriterCryptoHandle dw_handle = DDS::HANDLE_NIL;
-  DDS::Security::DatareaderCryptoHandle dr_handle = DDS::HANDLE_NIL;
+  DCPS::RepoId remote_part;
+  const DCPS::RepoId local_part = spdp_.guid();
+  if (std::memcmp(writer.guidPrefix, local_part.guidPrefix, sizeof writer.guidPrefix) == 0) {
+    remote_part = reader;
+  } else if (std::memcmp(reader.guidPrefix, local_part.guidPrefix, sizeof reader.guidPrefix) == 0) {
+    remote_part = writer;
+  } else {
+    return locators;
```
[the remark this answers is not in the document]
mitza-oci: That should be caught much earlier.  By the time it gets here we are handling an association which has to have at least one side local.  If both sides are local it should still work since the `lookup_participant_crypto_info` a few lines below this will return nil.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -935,18 +944,20 @@ Spdp::match_authenticated(const DCPS::RepoId& guid, DiscoveredParticipant& dp)
   dp.crypto_handle_ = key_factory->register_matched_remote_participant(crypto_handle_, dp.identity_handle_, dp.permissions_handle_, dp.shared_secret_handle_, se);
```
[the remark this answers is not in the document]
mitza-oci: Even in the cases where the local `crypto_handle_` is nil (no full-msg protection), the remote participants still need to get registered.  I tried without it but `preprocess_secure_submsg` requires the remote participant handle to work.

dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1077,54 +1127,40 @@ bool CryptoBuiltInImpl::encode_submessage(
...
   if (!authOnly) {
-    size += 8; // body submessage header + seq len
+    size += RTPS::SMHDR_SZ + SEQLEN_SZ;
   }
 
   size += pOut->length(); // submessage inside wrapper
-  if ((size + padding) % 4) {
-    padding += 4 - ((size + padding) % 4);
-  }
+  align(4, size, padding);
```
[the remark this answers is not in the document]
mitza-oci: Yes, I added one but forgot to use it here.

tests/security/CryptoTransformTest.cpp
mitza-oci: The latter, it's for origin authentication.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
mitza-oci: The spec says:
```
This operation may optionally not perform any transformation of the input RTPS message. In this case,
the operation shall return false but not set the exception object. In this situation the DDS
implementation shall send the original RTPS message.
```

If this happened in multiple places I'd be for creating a more systematic way of handing it, but it's only this one case.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
mitza-oci: I agree it's odd but I'm not sure what the best way to fix it is.  In theory, another plugin could be in use here, and the spec is silent about error codes.

Because key exchange happens asynchronously, we don't know if the key is available at any given time.  If we don't have the key yet, we drop the message without logging an error.  The plugin implementation doesn't have a way to (standardly) report this state.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
mitza-oci: The "sender" call is for the writer messages like Data and Heartbeat, while the "receiver" call is for AckNack (etc.) so in either case they are writers.

PR #1287, opened 2019-09-19
Make "--no-tests" the Default in the Configure Script

iguessthislldo:
Also added bits to make it easier to run `DevGuideExamples/DCPS/Messenger` in RTPS mode so the user can run RTPS with no tests and added an example `CMakeLists.txt`.

### DIFF
--- a/DevGuideExamples/DCPS/Messenger/Messenger.mpc
+++ b/DevGuideExamples/DCPS/Messenger/Messenger.mpc
@@ -7,7 +7,7 @@ project(*idl): dcps {
   custom_only = 1
 }
 
-project(*publisher) : dcpsexe, dcps_tcp {
+project(*publisher) : dcpsexe, dcps_tcp, dcps_rtps, dcps_rtps_udp {
   requires += no_opendds_safety_profile
   exename   = publisher
   after    += *idl
@@ -21,7 +21,7 @@ project(*publisher) : dcpsexe, dcps_tcp {
   }
 }
 
-project(*subscriber) : dcpsexe, dcps_tcp {
+project(*subscriber) : dcpsexe, dcps_tcp, dcps_rtps, dcps_rtps_udp {
   requires += no_opendds_safety_profile
   exename   = subscriber
   after    += *publisher
--- a/DevGuideExamples/DCPS/Messenger/Messenger.mpc
+++ b/DevGuideExamples/DCPS/Messenger/Messenger.mpc
@@ -7,7 +7,7 @@ project(*idl): dcps {
   custom_only = 1
 }
 
-project(*publisher) : dcpsexe, dcps_tcp, dcps_rtps, dcps_rtps_udp {
+project(*publisher) : dcpsexe, dcps_tcp, dcps_rtps_udp {
   requires += no_opendds_safety_profile
   exename   = publisher
   after    += *idl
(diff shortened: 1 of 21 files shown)


DevGuideExamples/DCPS/Messenger/Messenger.mpc
```diff
@@ -7,7 +7,7 @@ project(*idl): dcps {
   custom_only = 1
 }
 
-project(*publisher) : dcpsexe, dcps_tcp {
+project(*publisher) : dcpsexe, dcps_tcp, dcps_rtps, dcps_rtps_udp {
```
mitza-oci:
```suggestion
project(*publisher) : dcpsexe, dcps_tcp, dcps_rtps_udp {
```
doesn't need both...

DevGuideExamples/DCPS/Messenger/Publisher.cpp
```diff
@@ -78,21 +82,26 @@ ACE_TMAIN(int argc, ACE_TCHAR *argv[])
...
                         ACE_TEXT("ERROR: %N:%l: main() -")
                         ACE_TEXT(" create_publisher failed!\n")),
-                       -1);
+                       1);
     }
 
     // Create DataWriter
+    DDS::DataWriterQos writer_qos;
+    publisher->get_default_datawriter_qos(writer_qos);
+    writer_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
```
mitza-oci:
```suggestion
```
Already reliable.

DevGuideExamples/DCPS/Messenger/Publisher.cpp
```diff
@@ -78,21 +82,26 @@ ACE_TMAIN(int argc, ACE_TCHAR *argv[])
...
+    writer_qos.history.kind = DDS::KEEP_ALL_HISTORY_QOS;
```
mitza-oci: This is an issue with the "minimum profile".  We discussed a bit, but just recording it here.

INSTALL.md
```diff
@@ -84,7 +86,7 @@ bin/auto_run_tests.pl
 **For Windows:**
 
 ```
-bin\auto_run_tests.pl
+perl bin\auto_run_tests.pl
```
mitza-oci: Not sure why this is added?

DevGuideExamples/DCPS/Messenger/README.md
```diff
@@ -0,0 +1,27 @@
+# Messenger Example
+
+A very basic OpenDDS example, covered in the OpenDDS Developer's Guide.
+
+If you just want to build and run this, it will be built along with OpenDDS and
+can be run using `perl run_test.pl` assuming the `setenv` file has been sourced
+or OpenDDS has been installed. To have it use the DDS standard RTPS protocol
```
mitza-oci: This seems to state that `run_test.pl` will work without environment variables as long as OpenDDS has been installed, which I don't think is the case.

INSTALL.md
iguessthislldo: I think I added this because this didn't work on Windows when I tried it. Is it supposed work without using `perl`?

DevGuideExamples/DCPS/Messenger/README.md
iguessthislldo: That's true, my mistake.

INSTALL.md
mitza-oci: Yes, assuming Perl is correctly installed.  I've used this with both ActiveState and Strawberry.

PR #1296, opened 2019-10-02
Delete news item for trainings that have already been given in the past

jwillemsen:


mitza-oci:
It is meant to work like a mailing list archive.  Each "News" item has a date when it was published.  See https://gcc.gnu.org/ for an example that inspired this (with an event that happened last month).

I agree that in this case there's not a lot of value in having the old news item.

tottens:
I don't think there's much value in keeping the news feed clean of past events.  By that logic, we'd also have to delete the DDS Foundation Inaugural event and the recent DDS Foundation Meet & Greet.  The old events serve as an indication to anyone paying attention that there's a continuous history of events related to DDS and the OpenDDS open source project.  I do, however, think that going forward we should try to avoid words like "Upcoming" in the news headlines announcing events.

PR #1301, opened 2019-10-04
Fix Intermittent SharedTransport Shmem Test Failures

simpsont-oci:
Primary Issue: We can't be sure of publisher write speed vs reader read speed and the shmem transport doesn't have backpressure, so the shared memory segment needs to be able to hold the entire 100 messages or there's always the potential that the writer will fill the segment. A better solution would be to implement real checks for backpressure, but is beyond the scope of this change.

Solution: Increase test memory segment to accommodate 100 messages.

Secondary Issue: Occasional SEGV due to reads during shutdown. Introduce locking around shutdown and allocator acquisition, and introduce additional checks in send and receive strategies after they've grabbed a pointer to the allocator. We will eventually want to migrate this to use RCH rather than straight pointers in case it gets deleted while the strategy is interacting with it, but again that's beyond the scope of this change and will probably only (rarely) cause issues during shutdown.

### DIFF
--- a/dds/DCPS/transport/shmem/ShmemSendStrategy.cpp
+++ b/dds/DCPS/transport/shmem/ShmemSendStrategy.cpp
@@ -41,16 +41,19 @@ ShmemSendStrategy::start_i()
   const size_t n_elems = datalink_control_size_ / sizeof(ShmemData),
     extra = datalink_control_size_ % sizeof(ShmemData);
 
-  void* mem = alloc->calloc(datalink_control_size_);
-  if (mem == 0) {
+  void* mem = 0;
+  if (alloc == 0 || (mem = alloc->calloc(datalink_control_size_)) == 0) {
     VDBG_LVL((LM_ERROR, "(%P|%t) ERROR: ShmemSendStrategy for link %@ failed "
               "to allocate %B bytes for control\n", link_, datalink_control_size_), 0);
     return false;
   }
 
   ShmemData* data = reinterpret_cast<ShmemData*>(mem);
-  data[(extra >= sizeof(int)) ? n_elems : (n_elems - 1)].status_ =
-    SHMEM_DATA_END_OF_ALLOC;
+  const size_t limit = (extra >= sizeof(int)) ? n_elems : (n_elems - 1);
+  for (size_t i = 0; i < limit; ++i) {
+    data[i].status_ = SHMEM_DATA_FREE;
+  }
+  data[limit].status_ = SHMEM_DATA_END_OF_ALLOC;
   alloc->bind(bound_name_.c_str(), mem);
 
   ShmemAllocator* peer = link_->peer_allocator();
@@ -101,8 +104,8 @@ ShmemSendStrategy::send_bytes_i(const iovec iov[], int n)
   }
 
   ShmemAllocator* alloc = link_->local_allocator();
-  void* from_pool = alloc->malloc(pool_alloc_size);
-  if (from_pool == 0) {
+  void* from_pool = 0;
+  if (alloc == 0 || (from_pool = alloc->malloc(pool_alloc_size)) == 0) {
     VDBG_LVL((LM_ERROR, "(%P|%t) ERROR: ShmemSendStrategy for link %@ failed "
               "to allocate %B bytes for data\n", link_, pool_alloc_size), 0);
     errno = ENOMEM;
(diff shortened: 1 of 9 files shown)


dds/DCPS/transport/shmem/ShmemTransport.cpp
```diff
@@ -247,22 +250,34 @@ ShmemTransport::ReadTask::ReadTask(ShmemTransport* outer, ACE_sema_t semaphore)
 int
 ShmemTransport::ReadTask::svc()
 {
-  while (true) {
+  while (!stopped_) {
     ACE_OS::sema_wait(&semaphore_);
     if (stopped_) {
       return 0;
     }
```
iguessthislldo: Does this have to stay here if you changed to loop to use `stopped_`?

iguessthislldo: Actually nevermind, we checking it because we're coming out of `sema_wait`, right?

simpsont-oci: Yeah, I didn't want it to do the other stuff after a long wait for the semaphore if it had stopped in between.

dds/DCPS/transport/shmem/ShmemTransport.cpp
```diff
@@ -67,12 +67,12 @@ ShmemTransport::connect_datalink(const RemoteTransport& remote,
   ShmemDataLinkMap::iterator iter = links_.find(key.second);
   if (iter != links_.end()) {
     ShmemDataLink_rch link = iter->second;
-    VDBG_LVL((LM_DEBUG, ACE_TEXT("(%P|%t) ShmemTransport::connect_datalink ")
-              ACE_TEXT("link found.\n")), 2);
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ShmemTransport::connect_datalink ")
```
mitza-oci: Do we want this to be an unconditional log message?  (same on line 74)

dds/DCPS/transport/shmem/ShmemSendStrategy.cpp
```diff
@@ -41,16 +41,19 @@ ShmemSendStrategy::start_i()
...
               "to allocate %B bytes for control\n", link_, datalink_control_size_), 0);
     return false;
   }
 
   ShmemData* data = reinterpret_cast<ShmemData*>(mem);
-  data[(extra >= sizeof(int)) ? n_elems : (n_elems - 1)].status_ =
-    SHMEM_DATA_END_OF_ALLOC;
+  const size_t limit = (extra >= sizeof(int)) ? n_elems : (n_elems - 1);
+  for (size_t i = 0; i < limit; ++i) {
+    data[i].status_ = SHMEM_DATA_FREE;
```
mitza-oci: Is this needed?  I thought the `calloc` above was taking care of it.

dds/DCPS/transport/shmem/ShmemTransport.cpp
```diff
@@ -247,22 +250,34 @@ ShmemTransport::ReadTask::ReadTask(ShmemTransport* outer, ACE_sema_t semaphore)
...
 void
 ShmemTransport::ReadTask::stop()
 {
+  if (stopped_) {
+    return;
+  }
   stopped_ = true;
   ACE_OS::sema_post(&semaphore_);
-  wait();
+  ACE_Task_Base::wait();
```
mitza-oci: Why does it have the `ACE_Task_Base::` scope?

dds/DCPS/transport/shmem/ShmemTransport.cpp
simpsont-oci: Whoops, missed changing this one back from debugging.

dds/DCPS/transport/shmem/ShmemTransport.cpp
simpsont-oci: There was a period where it didn't seem like wait was getting called so I made it explicit in case I was missing something. I'll change it back.

dds/DCPS/transport/shmem/ShmemSendStrategy.cpp
simpsont-oci: Ah, yeah, I see now that calloc has a default initial_value parameter set to '0', which should match FREE.

PR #1309, opened 2019-10-07
java12 idl2jni corba runtime support

lij-oci:
added java classes for java12 idl2jni corba runtime support.


java/idl2jni/corba/org/omg/CORBA/ACTIVITY_REQUIRED copy.java
```diff
@@ -0,0 +1,21 @@
+package org.omg.CORBA;
+
+public final class ACTIVITY_COMPLETED extends SystemException {
```
mitza-oci: The name of this source file doesn't match the Java class name.

Issue #1297, opened 2019-10-02
opendds_idl is inserting path to include file incorrectly - using cmake


tmaslach:
I have attached an example that shows the problem.  Simply run cmake and try to build.  You will get a failure when it attempts to build bTypeSupportImpl.cpp.

[OpenDDSIdlBug.zip](https://github.com/objectcomputing/OpenDDS/files/3683382/OpenDDSIdlBug.zip)

Some explanation:
This project includes an a.idl and b.idl.  b.idl includes a.idl.  
Both idl files are in a subfolder of the project named dds.
I'm using the cmake interface, specifically opendds_target_sources to get the idl files compiled.  Because the idl files are in a subfolder, I have added `OPENDDS_IDL_OPTIONS -I${PROJECT_SOURCE_DIR}/dds` as an argument to opendds_target_sources.   
Cmake generates the project fine, but when opendds_idl runs, it generates the following include in bTypeSupportImpl.cpp:

```C++
#include "dds\aTypeSupportImpl.h"
```

It should be:
```C++
#include "aTypeSupportImpl.h"
```
since all the files are output in the same folder (the build folder).

I'm not sure if the problem is in cmake or opendds_idl.  I suspect it is opendds_idl.  I found this is what the build process calls:
```
opendds_idl.exe -IC:/repos/OpenDDSIdlBug -IC:/repos/OpenDDSIdlBug/dds --default-nested C:/repos/OpenDDSIdlBug/dds/b.idl"
```

This does generate the bad include.  However, if I remove the first -I argument (`-IC:/repos/OpenDDSIdlBug`), everything is generated correctly.  The first -I argument is added automatically by opendds_target_sources, I'm manually adding the second one. 

My workaround right now is to not put these files into a subfolder.  But, that's not a good solution, since the idl files will be in a git submodule.

Thanks!

tmaslach:
I forgot to add: I was able to successfully recreate this on Ubuntu.  I did not try any other platform.

iguessthislldo:
Yes, this looks like a bug in how `opendds_idl` selects the relative path to include in the `TypeSupportImpl`. It looks like the cmake module is adding the equivalent of `-I${PROJECT_SOURCE_DIR}` for out of source builds before your `-I${PROJECT_SOURCE_DIR}/dds`, but `opendds_idl` is just picking the first include path that matches the include from the IDL file. I will put a PR up shortly to have it pick the longest path that matches, which fixes your example.

[...]

PR #1315, opened 2019-10-08
Use Shortest Relative Include Candidate in opendds_idl

iguessthislldo:
Fix for #1297

### DIFF
--- a/dds/idl/be_global.cpp
+++ b/dds/idl/be_global.cpp
@@ -526,20 +526,20 @@ namespace {
 
   std::string make_relative(const std::string& absolute)
   {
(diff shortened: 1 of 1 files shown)


iguessthislldo:
Coming back to this again, I'm thinking this is the wrong solution. I actually think the whole function has the wrong premise. I don't think you can just pick out the right path *just* given the include path and the absolute. I think I will have to work on this some more.

mitza-oci:
@iguessthislldo and/or @jrw972 do you want to close this one?

iguessthislldo:
Assuming Justin doesn't need it, closing.

PR #1319, opened 2019-10-11
release script: use default download URL

mitza-oci:


iguessthislldo:
I looked at #1049 and it looks like I added `--download-url` to replace all the hardcoded urls but I never  set a default.

PR #1322, opened 2019-10-14
RtpsRelay: Message validation

jrw972:


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -259,9 +260,12 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
   ACE_INET_Addr spdp(rtps_discovery->get_spdp_port(application_domain, application_participant_id), "127.0.0.1");
   ACE_INET_Addr sedp(rtps_discovery->get_sedp_port(application_domain, application_participant_id), "127.0.0.1");
 
-  SpdpHandler spdp_vertical_handler(reactor, association_table, lifespan, application_participant_id, spdp);
-  SedpHandler sedp_vertical_handler(reactor, association_table, lifespan, application_participant_id, sedp);
-  DataHandler data_vertical_handler(reactor, association_table, lifespan, application_participant_id);
+  OpenDDS::Security::SecurityConfig_rch conf = TheSecurityRegistry->default_config();
```
mitza-oci: Need to cover builds that are have security disabled at compile time using `OPENDDS_SECURITY` preprocessor macro

tools/rtpsrelay/RelayHandler.h
```diff
@@ -55,7 +55,13 @@ class VerticalHandler : public RelayHandler {
   VerticalHandler(ACE_Reactor* a_reactor,
                   const AssociationTable& a_association_table,
                   const ACE_Time_Value& lifespan,
-                  const OpenDDS::DCPS::RepoId& application_participant_guid);
+                  const OpenDDS::RTPS::RtpsDiscovery_rch& rtps_discovery,
+                  DDS::DomainId_t application_domain,
+                  const OpenDDS::DCPS::RepoId& application_participant_guid
+#ifdef OPENDDS_SECURITY
```
mitza-oci: This is getting a bit hard to manage.  I wonder if it would be better to have a consistent number of parameters and use the macro replacement just for its type (it could be `int` without security).  I think that would help cleaning up the calling code as well.

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -84,22 +86,19 @@ int RelayHandler::handle_input(ACE_HANDLE)
...
-    static_cast<CORBA::Octet>(buffer->rd_ptr()[0]) == OpenDDS::RTPS::BEACON_MSG_ID;
-
-  OpenDDS::DCPS::RepoId src_guid;
+  const auto& header = mp.header();
   std::memcpy(src_guid.guidPrefix, header.guidPrefix, sizeof(OpenDDS::DCPS::GuidPrefix_t));
   src_guid.entityId = OpenDDS::DCPS::ENTITYID_PARTICIPANT;
 
   buffer->rd_ptr(rd_ptr);
-  process_message(remote, ACE_Time_Value().now(), src_guid, buffer.get(), is_beacon_message);
+  process_message(remote, ACE_Time_Value().now(), src_guid, buffer.get());
```
mitza-oci: @iguessthislldo should this use new time utility?  Time_Point?

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -288,6 +315,104 @@ bool SpdpHandler::do_normal_processing(const ACE_INET_Addr& a_remote,
...
+    ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: RelayHandler::parse_message failed to deserialize RTPS header\n"));
+    return false;
+  }
+
+  const auto& header = message_parser.header();
+  OpenDDS::DCPS::RepoId src_guid;
+  std::memcpy(src_guid.guidPrefix, header.guidPrefix, sizeof(OpenDDS::DCPS::GuidPrefix_t));
+  src_guid.entityId = OpenDDS::DCPS::ENTITYID_PARTICIPANT;
+
+  do {
```
mitza-oci: It seems to be safe to do it this way, but I'm not sure there's any value in having `do`/`while` here instead of just plain `while` (with the same condition).  A message without submessges is not legal, but this should be robust to it anyway.

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -288,6 +315,104 @@ bool SpdpHandler::do_normal_processing(const ACE_INET_Addr& a_remote,
...
+          DDS::Security::ParticipantCryptoHandle remote_crypto_handle = rtps_discovery_->get_crypto_handle(application_domain_, application_participant_guid_, src_guid);
+          if (remote_crypto_handle == DDS::HANDLE_NIL) {
+            ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: RelayHandler::parse_message no crypto handle for client %C\n", guid_to_string(src_guid).c_str()));
+            return false;
+          }
+
+          DDS::OctetSeq encoded_buffer, plain_buffer;
+          DDS::Security::SecurityException ex;
+
+          encoded_buffer.length(msg->length());
```
mitza-oci: This doesn't support `ACE_Message_Block` chaining, which I think is fine for how the Relay is set up (AFAICT).  However it may be good to add an explicit to check for the `cont()` pointer being null.

tools/rtpsrelay/RelayHandler.cpp
jrw972: I'll use MonotonicTimePoint.

tools/rtpsrelay/RelayHandler.cpp
jrw972: From the code for the MessageParser, it looked like the submessage was not initialized until I called parseSubMessage.  Thus, I didn't want to call hasNextSubmessage until I called parseSubMessage.  Am I missing something?

mitza-oci: Looks like the previous uses of it rely on `remaining`:
https://github.com/objectcomputing/OpenDDS/blob/acb816282aee2477efd0e65418ff87d09ddf003f/dds/DCPS/security/CryptoBuiltInImpl.cpp#L1028

tools/rtpsrelay/RelayHandler.cpp
iguessthislldo: I'm not as concerned about using the new time types outside the main libraries, but it would be nice. As a warning, `ACE_Time_Value`s require explicit conversions to interact with the new time types, so if just you change one type, most if not all the things the former `ACE_Time_Value` touched will become invalid.

PR #1325, opened 2019-10-15
Export RcObject's Symbols

simpsont-oci:
Export RcObject symbol in order to facilitate WeakRcHandle::lock()'s dynamic_cast (needs typeinfo).


iguessthislldo:
@mitza-oci @simpsont-oci I tested messenger with NDK r20 on my phone. Without this change messenger acts like smart lock app, starts to work via debug log, but doesn't exchange messages. With this change it worked. :+1:

PR #1327, opened 2019-10-16
RtpsRelay: Use INFO_DST when routing

jrw972:


tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -431,6 +366,132 @@ bool VerticalHandler::parse_message(OpenDDS::RTPS::MessageParser& message_parser
...
+                                        ACE_Message_Block* msg)
+{
+  // Determine how many guids we can pack into a UDP message.
+  const auto max_guids = (65536 - msg->length()) / sizeof (OpenDDS::DCPS::RepoId) - 1;
+
+  auto pos = guids.begin();
+  const auto limit = guids.end();
+
+  while (pos != limit) {
+    ACE_Message_Block* header_block = new ACE_Message_Block(256);
```
mitza-oci: Why 256?

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -431,6 +366,132 @@ bool VerticalHandler::parse_message(OpenDDS::RTPS::MessageParser& message_parser
...
+    OpenDDS::DCPS::Serializer ser(header_block);
+
+    RelayHeader relay_header;
+    auto& to = relay_header.to();
+    while (pos != limit && to.size() < max_guids) {
+      to.push_back(guid_to_guid(*pos));
+      ++pos;
```
mitza-oci: Looks like a `for` loop

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -431,6 +366,132 @@ bool VerticalHandler::parse_message(OpenDDS::RTPS::MessageParser& message_parser
...
+
+  const size_t size_after_header = mp.remaining();
+
+  msg->rd_ptr(size_before_header - size_after_header);
+
+  for (const auto& guid : relay_header.to()) {
+    const auto p = vertical_handler_->find(guid_to_guid(guid));
+    if (p != vertical_handler_->end()) {
+      for (const auto& addr : p->second) {
+        vertical_handler_->enqueue_message(addr, msg);
```
mitza-oci: This pattern (for-each loop around `enqueue_message`) happens a few times.  Would it be cleaner to provide an `enqueue_message` overload that takes an AddressSet and does the loop internally?

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -431,6 +366,132 @@ bool VerticalHandler::parse_message(OpenDDS::RTPS::MessageParser& message_parser
...
+  if (src_guid == application_participant_guid_) {
+    if (remote != application_participant_addr_) {
+      // Something is impersonating our application participant.
+      return false;
+    }
+
+    // SPDP message is from the application participant.
+    // Cache the SPDP announcement from the application participant.
+    if (spdp_message_) {
+      spdp_message_->release();
```
mitza-oci: `Message_Block_Ptr` could be useful here

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -442,47 +503,54 @@ void SpdpHandler::purge(const GuidAddr& ga)
...
-                         const RelayAddressesSet& relay_addresses)
+                         const RelayAddressesMap& relay_addresses_map)
 {
-  if (local_guids.empty() && relay_addresses.empty()) {
+  if (relay_addresses_map.empty()) {
     return;
   }
 
-  const OpenDDS::DCPS::RepoId a_src_guid = to_participant_guid(x);
+  const OpenDDS::DCPS::RepoId src_guid = to_participant_guid(x);
```
mitza-oci:
```suggestion
  const auto src_guid = to_participant_guid(x);
```

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -214,116 +216,38 @@ void VerticalHandler::process_message(const ACE_INET_Addr& a_remote,
...
+  if (do_normal_processing(remote, src_guid, to, msg)) {
+    RelayAddressesMap relay_addresses_map;
+    association_table_.populate_relay_addresses_map(relay_addresses_map, src_guid, to);
+
+    size_t fan_out = 0;
+    for (const auto& p : relay_addresses_map) {
+      const auto& addrs = p.first;
+      const auto& guids = p.second;
+      fan_out += guids.size();
+      if (addrs != relay_addresses_) {
```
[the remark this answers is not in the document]
jrw972: They are not sets.  They are triples containing the horizontal SPDP/SEDP/DATA addresses of a relay instance which for its identity.  The condition is used to determine if the message needs to be sent to a different relay.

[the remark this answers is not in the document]
mitza-oci: I must have confused it with a different type.

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -431,6 +345,154 @@ bool VerticalHandler::parse_message(OpenDDS::RTPS::MessageParser& message_parser
...
+      to.push_back(repoid_to_guid(*pos));
+    }
+
+    size_t size = 0, padding = 0;
+    OpenDDS::DCPS::gen_find_size(relay_header, size, padding);
+    ACE_Message_Block* header_block = new ACE_Message_Block(size + padding);
+    OpenDDS::DCPS::Serializer ser(header_block);
+    ser << relay_header;
+    header_block->cont(msg.get()->duplicate());
+    RelayHandler::enqueue_message(addr, OpenDDS::DCPS::Message_Block_Shared_Ptr(header_block));
```
mitza-oci: Since `header_block`  will be managed by Shared_Ptr, it would be cleaner to start it out that way when it's allocated on line 402.

dds/DCPS/Message_Block_Ptr.h
```diff
@@ -22,6 +22,42 @@ struct Message_Block_Deleter
...
+
+  ACE_Message_Block& operator*() const { return *msg_; }
+
+  ACE_Message_Block* operator->() const { return msg_.get(); }
+
+  ACE_Message_Block* get() const { return msg_.get(); }
+
+  static void swap(Message_Block_Shared_Ptr& a, Message_Block_Shared_Ptr& b)
+  {
+    std::swap(a.msg_, b.msg_);
```
mitza-oci: This needs be an unqualified swap for ADL.  Looks to be the cause of some of the CI build problems.  On some platforms without unique_ptr, it is emulated in the OpenDDS header.
https://en.cppreference.com/w/cpp/named_req/Swappable

PR #1328, opened 2019-10-16
Support setting a default Configuration File

iguessthislldo:
One of two changes I'm bringing over from bench.

It lets someone set a default for `DCPSConfigFile` that will be used if `-DCPSConfigFile` was not passed to `TheParticipantFactoryWithArgs` or if `TheParticipantFactory` was called. Of course this means it must be called before `TheParticipantFactory(WithArgs)`. If this is not used at all, then OpenDDS will initialize normally.

Here is an example of how I'm using it in the bench test controller application:

```C++
  TheServiceParticipant->default_configuration_file(
    ACE_TEXT_CHAR_TO_TCHAR(join_path(bench_root, "control_opendds_config.ini").c_str()));
```

### DIFF
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -2000,25 +2000,23 @@ Service_Participant::delete_replayer(Replayer_ptr replayer)
   return ret;
 }
 
-DDS::Topic_ptr
-Service_Participant::create_typeless_topic(
+DDS::Topic_ptr Service_Participant::create_typeless_topic(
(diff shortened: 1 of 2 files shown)


mitza-oci:
> It lets someone set a default for `DCPSConfigFile` that will be used if `-DCPSDebugFile` was not [...]

Should be `-DCPSConfigFile`?

iguessthislldo:
> Should be -DCPSConfigFile?

Yes, I think I've typed `DCPSDebugLevel` one too many times.

PR #1329, opened 2019-10-16
Using Typedefs of Structs as Keys

iguessthislldo:
This was the other change I made in bench that I wanted to bring over. It allows typedefs of structs to be keys. The problem is that it doesn't seem to be a problem with annotations, only pragmas, but I added a annotation test anyway.

### DIFF
--- a/dds/idl/marshal_generator.cpp
+++ b/dds/idl/marshal_generator.cpp
@@ -870,11 +870,7 @@ namespace {
     for (size_t i = 0; i < fields.size(); ++i) {
       string field_name = fields[i]->local_name()->get_string();
       if (field_name == key_base) {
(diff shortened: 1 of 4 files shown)


dds/idl/marshal_generator.cpp
```diff
@@ -871,6 +871,10 @@ namespace {
       string field_name = fields[i]->local_name()->get_string();
       if (field_name == key_base) {
         AST_Type* field_type = fields[i]->field_type();
+        AST_Typedef* typedef_node = dynamic_cast<AST_Typedef*>(field_type);
```
mitza-oci: This could be `resolveActualType` from dds_generator.h.  That also handles typedefs of forward-declared (struct|union).

PR #1331, opened 2019-10-17
added equals() and hashCode() to TAOObject

lij-oci:


java/idl2jni/runtime/i2jrt/TAOObject.java
```diff
@@ -23,6 +23,19 @@ protected void finalize() {
     _jni_fini();
   }
 
+  @Override
+  public boolean equals(java.lang.Object o) {
+    if ((o != null) && (o instanceof TAOObject)) {
```
mitza-oci:
```suggestion
    if (o instanceof TAOObject) {
```

java/idl2jni/runtime/i2jrt/TAOObject.java
```diff
@@ -23,6 +23,19 @@ protected void finalize() {
...
+      return _jni_ptr == ((TAOObject)o)._jni_ptr;
+    }
+    return this == o;
```
mitza-oci:
```suggestion
    return false;
```

PR #1332, opened 2019-10-18
Message_Block_Ptr.h pre-C++11 compatibility

mitza-oci:


dds/DCPS/Message_Block_Ptr.h
```diff
@@ -41,7 +41,7 @@ class OpenDDS_Dcps_Export Message_Block_Shared_Ptr {
     return *this;
   }
 
-  explicit operator bool() const { return msg_.get(); }
+  operator bool() const { return msg_.get(); }
```
simpsont-oci: For things like this (and override, etc), does it make sense to introduce some preprocessor macros so we can preserve the additional compile time checks on systems that have C++11 compilers?

mitza-oci: We have those in some cases.  In this case, the existing usage in `unique_ptr.h` had a non-explicit operator bool, so it looked to be OK.

PR #1335, opened 2019-10-21
Add option for disabling builtin SEDP writers

jrw972:
This allows the application participant in the RtpsRelay to not send heartbeats and other writer messages.

### DIFF
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -381,13 +381,17 @@ Sedp::init(const RepoId& guid,
   subscriptions_secure_reader_->enable_transport_using_config(reliable, durable, transport_cfg_);
 #endif
 
-  participant_message_writer_.enable_transport_using_config(reliable, durable, transport_cfg_);
+  if (spdp_.available_builtin_endpoints() & DISC_BUILTIN_ENDPOINT_PARTICIPANT_ANNOUNCER) {
+    participant_message_writer_.enable_transport_using_config(reliable, durable, transport_cfg_);
+  }
   participant_message_reader_->enable_transport_using_config(reliable, durable, transport_cfg_);
 
 #ifdef OPENDDS_SECURITY
   participant_message_secure_writer_.set_crypto_handles(spdp_.crypto_handle());
   participant_message_secure_reader_->set_crypto_handles(spdp_.crypto_handle());
-  participant_message_secure_writer_.enable_transport_using_config(reliable, durable, transport_cfg_);
+  if (spdp_.available_builtin_endpoints() & DDS::Security::BUILTIN_PARTICIPANT_MESSAGE_SECURE_WRITER) {
+    participant_message_secure_writer_.enable_transport_using_config(reliable, durable, transport_cfg_);
+  }
   participant_message_secure_reader_->enable_transport_using_config(reliable, durable, transport_cfg_);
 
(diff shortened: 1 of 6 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -106,9 +106,56 @@ namespace {
...
-                       const DDS::DomainParticipantQos& /*qos*/,
+                       const DDS::DomainParticipantQos& qos,
                        RtpsDiscovery* disco)
 {
+  bool disable_writers = false;
+
+  const DDS::PropertySeq& properties = qos.property.value;
+  for (unsigned int idx = 0; idx != properties.length(); ++idx) {
+    const char* name = properties[idx].name.in();
+    if (std::strcmp("OpenDDS.Rtps.Discovery.DisableWriters", name) == 0) {
```
jrw972: I'm looking for suggestions as to what header should define the constant "OpenDDS.Rtps.Discovery.DisableWriters", what the actual key should be, and what the value should be and how we should handle it.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1426,6 +1448,14 @@ Sedp::update_topic_qos(const RepoId& topicId, const DDS::TopicQos& qos)
 DDS::ReturnCode_t
 Sedp::remove_publication_i(const RepoId& publicationId, LocalPublication& pub)
 {
+  if (!(spdp_.available_builtin_endpoints() & (DISC_BUILTIN_ENDPOINT_PUBLICATION_ANNOUNCER
```
mitza-oci: I think we don't need this block (and the corresponding one in `remove_subscription_i`) because the entity being removed couldn't have been added in the first place.

dds/DCPS/RTPS/Spdp.cpp
mitza-oci: a. RtpsDiscovery.h
b. "OpenDDS.RtpsDiscovery.EndpointAnnouncements"
c. "0" or "false" (case insensitive) to disable

simpsont-oci: I think there are normally 3 heartbeats without security, right? Publications, Subscriptions, and Participant Messages? And more for security... does "EndpointAnnouncements" cover the rest of them? Also, does this need to be a qos setting? Can it just be be a config entry?

mitza-oci: We are leaving Participant Message (liveliness) for a future iteration, but I agree it may be an issue.  We can rename the constant when we get there, if needed.

It is QoS because it impacts a specific DomainParticipant.  The config file only works at the domain level.

mitza-oci: Looking at the spec again, I think we could go ahead and treat `participant_message_writer_` and `participant_message_secure_writer_` the same as the endpoint discovery writers.  There's nothing that could be written on "participant messages" (liveliness updates) that would be relevant in this scenario.

jrw972: > Also, does this need to be a qos setting? Can it just be be a config entry?

We wanted it in the qos so that you can selectively apply it to specific participants.

@mitza-oci Are you still okay with "OpenDDS.RtpsDiscovery.EndpointAnnouncements"?

mitza-oci: Yes, unless anyone has other suggestions?

Issue #1336, opened 2019-10-22
Omission of scope when linking target against Threads::Threads in cmake macro OPENDDS_TARGET_SOURCES causes cmake error 


JuneZA:
It was necessary to modify line 160 of cmake/api_macros.cmake (in current master) to the following:
`target_link_libraries(${target} PUBLIC Threads::Threads)`
to overcome the error (CMP0023):  

> CMake Error at /opt/OpenDDS/cmake/api_macros.cmake:160 (target_link_libraries): The keyword signature for target_link_libraries has already been used with the target "myTarget".  All uses of target_link_libraries with a target must be either all-keyword or all-plain.

jwillemsen:
Please open a pull request with the necessary change

mitza-oci:
Does `${target}` represent the user's target or something internal to OpenDDS?  If it's the user's target, wouldn't it be better to set up the dependency on an internal target?

iguessthislldo:
> Does `${target}` represent the user's target or something internal to OpenDDS?

Yes, `${target}` is what the user is passing to `OPENDDS_TARGET_SOURCES`.

> If it's the user's target, wouldn't it be better to set up the dependency on an internal target?

Maybe that target could be the OpenDDS::* libraries, which I assume have to be linked anyway. It looks like we can tell it to do that using [`INTERFACE_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html#prop_tgt:INTERFACE_LINK_LIBRARIES) when we're defining them.

iguessthislldo:
Okay so looking at `OpenDDSConfig.cmake`, it looks like it already has some of the machinery for this. I'm thinking if we add the non ACE/TAO/OpenDDS dependencies to these lists and make sure the IDL target is linked too, it should fix the concerns I have and leave the decision of the external linking up to the existing option we have on `OPENDDS_TARGET_SOURCES` and if the user links to the OpenDDS::* libraries.

iguessthislldo:
Alright, so this issue is my fault. In #1256 I needed to add `Threads::Threads` to linker code for inlined code, but I did that by calling `target_link_libraries` directly on the user target instead of declaring it as a interface dependency of ACE like I did with the other dependencies. I don't think I was completely aware of how dependencies were working. I will create a new PR to fix this.

iguessthislldo:
[...]

PR #1337, opened 2019-10-22
 Included scope for linking target when using cmake

JuneZA:
Included PUBLIC scope for linking target against Threads::Threads in cmake macro OPENDDS_TARGET_SOURCES.

### DIFF
--- a/cmake/api_macros.cmake
+++ b/cmake/api_macros.cmake
@@ -157,7 +157,7 @@ macro(OPENDDS_TARGET_SOURCES target)
     target_link_libraries(${target} ${OPENDDS_DCPS_LINK_DEPS})
   endif()
 
-  target_link_libraries(${target} Threads::Threads)
+  target_link_libraries(${target} PUBLIC Threads::Threads)
 
   foreach(scope PUBLIC PRIVATE INTERFACE)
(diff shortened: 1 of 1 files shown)


iguessthislldo:
Link to #1336 

Also why did you close this? Also do you we need `PUBLIC` for all uses of `target_link_libraries`?

JuneZA:
I closed the pull request when I realised that simply putting the scope with `target_link_libraries` inside the macro was not a general solution, as the Messenger examples failed to build, and I needed to look into it more. 

CMake specifies that plain and keyword signatures for `target_link_libraries` cannot be mixed for a target - there's an example at [https://cmake.org/cmake/help/latest/policy/CMP0023.html](url) - so if the use of `target_link_libraries` is without keywords inside `OPENDDS_TARGET_SOURCES`, it prohibits the user from specifying the scope when linking elsewhere in the project. It happened to be that the project I was working on had the scope specified for the target and I came across this issue.

I was not suggesting you need 'PUBLIC' scope for all uses of `target_link_libraries`, as this may be in some cases the wrong relationship. Perhaps using a plain signature is a simpler solution, as the target may be an executable or a library, but I could not easily get an answer of whether it would be best practice.

PR #1338, opened 2019-10-22
Add Thrasher Test for rtps_udp transport, Fix WaitForAck & Durability Issues It Exposed

simpsont-oci:
On the DDS Build Scoreboard I noticed a stack trace pointing to a corrupted free in DomainParticipantFactoryImpl::delete_participant's use of Util::remove() and it looks like we're not correctly using the participants_protector_ mutex in this instance. Fixed this, added locking / copying around another accessor of the same data structure (internals are RcHandles), and migrated to a non-recursive mutex because it didn't look like we needed one.

### DIFF
--- a/dds/DCPS/DomainParticipantFactoryImpl.h
+++ b/dds/DCPS/DomainParticipantFactoryImpl.h
@@ -75,7 +75,7 @@ public:
     DDS::DomainParticipantFactoryQos & qos);
 
   /// Expose the participants for reading.
-  const DPMap& participants() const;
+  DPMap participants() const;
 
   void cleanup();
 
@@ -92,7 +92,7 @@ private:
   /// Protect the participant collection.
   /// Use recursive mutex to allow nested acquisition and
   /// release of a mutex that occurs in the same thread.
-  ACE_Recursive_Thread_Mutex  participants_protector_;
+  mutable ACE_Recursive_Thread_Mutex  participants_protector_;
 };
 
 } // namespace DCPS
--- a/dds/DCPS/DomainParticipantFactoryImpl.h
+++ b/dds/DCPS/DomainParticipantFactoryImpl.h
@@ -74,8 +74,8 @@ public:
   virtual DDS::ReturnCode_t get_qos(
     DDS::DomainParticipantFactoryQos & qos);
 
-  /// Expose the participants for reading.
-  DPMap participants() const;
+  /// Make a copy of the participants map for reading.
+  const DPMap participants() const;
 
   void cleanup();
 
@@ -84,15 +84,13 @@ private:
   DDS::DomainParticipantFactoryQos qos_;
 
   /// The default qos value of DomainParticipant.
-  DDS::DomainParticipantQos   default_participant_qos_;
+  DDS::DomainParticipantQos default_participant_qos_;
 
   /// The collection of domain participants.
-  DPMap                       participants_;
+  DPMap participants_;
 
   /// Protect the participant collection.
-  /// Use recursive mutex to allow nested acquisition and
-  /// release of a mutex that occurs in the same thread.
-  mutable ACE_Recursive_Thread_Mutex  participants_protector_;
+  mutable ACE_Thread_Mutex participants_protector_;
 };
 
 } // namespace DCPS
--- a/dds/DCPS/WriteDataContainer.cpp
+++ b/dds/DCPS/WriteDataContainer.cpp
@@ -1426,6 +1426,12 @@ WriteDataContainer::wait_ack_of_seq(const MonotonicTimePoint& abs_deadline, cons
   DDS::ReturnCode_t ret = DDS::RETCODE_OK;
   ACE_GUARD_RETURN(ACE_SYNCH_MUTEX, guard, this->wfa_lock_, DDS::RETCODE_ERROR);
 
+  SequenceNumber last_acked = acked_sequences_.last_ack();
+  SequenceNumber acked = acked_sequences_.cumulative_ack();
+  if (sequence == last_acked && sequence == acked && sending_data_.size() != 0) {
+    acked_sequences_.insert(sending_data_.head()->get_header().sequence_.previous());
+  }
+
   while (MonotonicTimePoint::now() < deadline) {
 
     if (!sequence_acknowledged(sequence)) {
--- a/dds/DCPS/WriteDataContainer.cpp
+++ b/dds/DCPS/WriteDataContainer.cpp
@@ -120,6 +120,10 @@ WriteDataContainer::WriteDataContainer(
 
 WriteDataContainer::~WriteDataContainer()
 {
+  ACE_GUARD(ACE_Recursive_Thread_Mutex,
+            guard,
+            this->lock_);
+
   if (this->unsent_data_.size() > 0) {
     ACE_DEBUG((LM_WARNING,
                ACE_TEXT("(%P|%t) WARNING: WriteDataContainer::~WriteDataContainer() - ")
@@ -461,6 +465,10 @@ WriteDataContainer::num_all_samples()
 ACE_UINT64
 WriteDataContainer::get_unsent_data(SendStateDataSampleList& list)
 {
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
+                   guard,
+                   this->lock_, 0);
+
   DBG_ENTRY_LVL("WriteDataContainer","get_unsent_data",6);
   //
   // The samples in unsent_data are added to the local datawriter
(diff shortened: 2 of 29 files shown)


dds/DCPS/DomainParticipantFactoryImpl.h
```diff
@@ -74,8 +74,8 @@ class OpenDDS_Dcps_Export DomainParticipantFactoryImpl
   virtual DDS::ReturnCode_t get_qos(
     DDS::DomainParticipantFactoryQos & qos);
 
-  /// Expose the participants for reading.
-  const DPMap& participants() const;
+  /// Make a copy of the participants map for reading.
+  const DPMap participants() const;
```
mitza-oci:
```suggestion
  DPMap participants() const;
```

tests/DCPS/Thrasher/ParticipantTask.cpp
```diff
@@ -44,160 +84,152 @@ ParticipantTask::svc()
     DDS::Publisher_var publisher;
     DDS::DataWriter_var writer;
     FooDataWriter_var writer_i;
-    DDS::StatusCondition_var cond;
-    DDS::WaitSet_var ws = new DDS::WaitSet;
+    //DDS::StatusCondition_var cond;
```
mitza-oci: remove these lines?

tests/DCPS/Thrasher/Subscriber.cpp
```diff
@@ -135,34 +174,22 @@ ACE_TMAIN(int argc, ACE_TCHAR** argv)
...
-          ACE_ERROR_RETURN((LM_ERROR,
-                            ACE_TEXT("%N:%l: main()")
-                            ACE_TEXT(" wait failed!\n")), 8);
-
-        if (reader->get_subscription_matched_status(matches) != ::DDS::RETCODE_OK)
-        {
-          ACE_ERROR ((LM_ERROR,
-            "ERROR: failed to get subscription matched status\n"));
-          return 1;
+      for (size_t x = 0; x < n_publishers; ++x) {
```
mitza-oci: This is showing up as a signed/unsigned comparison in CI builds

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -461,6 +465,10 @@ WriteDataContainer::num_all_samples()
 ACE_UINT64
 WriteDataContainer::get_unsent_data(SendStateDataSampleList& list)
 {
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
```
mitza-oci: There are a few of these where the (recursive) lock is already locked on all paths into the function.

simpsont-oci: Right, revisiting this. I didn't realize that DataWriterImpl was using the WriteDataContainer's lock instead of its own. Hooray:
1) pending_data() was always covered and never called externally, so I removed that lock and made it private. I reverted the name change since the lock is public and we can't have nice things.
2) remove_oldest_sample was also only ever called internally by dispose() (which has its own lock) and obtain_buffer() which doesn't lock but is only ever called by DataWriterImpl::write() which (ugh) locks WriteDataContainer's (public!) lock manually before calling obtain_buffer(). It's gross, but it's covered, so I removed both calls to the recursive lock in there (there was an additional one protecting a call to pending_data() in there for good measure?).
3) get_unsent_data()'s lock was secretly being made thirty levels up the stack as well, probably multiple times because why not. Removing this one as well.
4) wait_ack_of_seq(), which now touches sending_data_, needs the (non-wfa) lock and was unprotected when called by DataWriterImpl... which, by the way, looks like it's actually unlocked between the time it calls send_ack_request() and when it calls into wait_ack_of_seq() leaving room for someone else to bump the sequence number, but maybe that's not a huge deal if we're just potentially waiting on a later seq than strictly necessary? I'll defer to the existing code on this one unless I hear otherwise.

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -120,6 +120,10 @@ WriteDataContainer::WriteDataContainer(
 
 WriteDataContainer::~WriteDataContainer()
 {
+  ACE_GUARD(ACE_Recursive_Thread_Mutex,
+            guard,
+            this->lock_);
```
jrw972: this->

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -1424,7 +1421,16 @@ WriteDataContainer::wait_ack_of_seq(const MonotonicTimePoint& abs_deadline, cons
 {
   const MonotonicTimePoint deadline(abs_deadline);
   DDS::ReturnCode_t ret = DDS::RETCODE_OK;
-  ACE_GUARD_RETURN(ACE_SYNCH_MUTEX, guard, this->wfa_lock_, DDS::RETCODE_ERROR);
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, this->lock_, DDS::RETCODE_ERROR);
```
jrw972: Look at Tim's comments on this.

tests/DCPS/Thrasher/DataReaderListenerImpl.h
```diff
@@ -47,10 +49,17 @@ class DataReaderListenerImpl
       DDS::DataReader_ptr reader,
       const DDS::SampleLostStatus& status);
 
+  bool
+  wait_received(const OpenDDS::DCPS::TimeDuration& duration, size_t target);
+
+  OPENDDS_MAP(size_t, OPENDDS_SET(size_t)) x_and_y_map;
```
jrw972: Better name.

tests/DCPS/Thrasher/ParticipantTask.cpp
```diff
@@ -17,21 +17,61 @@
...
+{
+  DDS::StatusCondition_var condition = writer->get_statuscondition();
+  condition->set_enabled_statuses(DDS::PUBLICATION_MATCHED_STATUS);
+  DDS::WaitSet_var ws = new DDS::WaitSet;
+  ws->attach_condition(condition);
+  DDS::ConditionSeq conditions;
+  DDS::PublicationMatchedStatus ms = { 0, 0, 0, 0, 0 };
+  DDS::Duration_t timeout = { 3, 0 };
+  DDS::ReturnCode_t stat;
+  do {
```
mitza-oci:
```suggestion
  while (true) {
```
Also in Subscriber.cpp -- should these be combined?

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -120,6 +120,10 @@ WriteDataContainer::WriteDataContainer(
 
 WriteDataContainer::~WriteDataContainer()
 {
+  ACE_GUARD(ACE_Recursive_Thread_Mutex,
```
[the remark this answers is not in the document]
simpsont-oci: I'm not so much worried about concurrent access to the destructor as I am having the destructor be called while another thread is modifying one of the protected members (especially maps / sets which may change in shape & size). I recognize this doesn't fully resolve the issue (if this is the case, we'll still get read access violations), but I wasn't sure if our usage pattern prevented this scenario (When does the transport reactor stop in relation to destructor being called? If it's after, have we fully disconnected all reactor task code paths from the WriteDataContainer before the destructor is called?) so I added this as a measure of protection in the worst case.

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -966,15 +970,8 @@ WriteDataContainer::remove_oldest_sample(
...
-  // Signal if there is no pending data.
-  {
-    ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
-                     guard,
-                     this->lock_,
-                     DDS::RETCODE_ERROR);
-
-    if (!pending_data())
-      empty_condition_.broadcast();
+  if (!pending_data()) {
```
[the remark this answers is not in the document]
simpsont-oci: Yeah, see my other response to your comment from 2 weeks ago. I misunderstood where the locks were being called (DataWriterImpl uses WriteDataContainer's lock) and so only the addition of the lock to wait_ack_of_seq was necessary.

tests/DCPS/Thrasher/ParticipantTask.cpp
simpsont-oci: Yes, and probably for more than these two tests.

dds/DCPS/WriteDataContainer.cpp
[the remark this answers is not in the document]
mitza-oci: Once the destructor is reached, no other thread can be interacting with the object.  (That's the concurrent access I was talking about yesterday.)  If that's not true, it should be fixed elsewhere -- probably in DataWriterImpl.

mitza-oci:
@simpsont-oci please rename this PR

PR #1339, opened 2019-10-23
Support Implied Keys

iguessthislldo:
- Support Implied Keys:
  - If a nested struct has no specified keys and a field of the nested struct type is marked with `@key`, all the fields in the nested struct are implied to be marked with `@key`.
- `opendds_idl` will now error if the set IDL version is less than 4.
- Added topic annotation migration guide and updated deprecation warning and news file to point to it.


.gitignore
```diff
@@ -1,5 +1,6 @@
 .depend.*
 GNUmakefile*
+makefile_name.local
```
mitza-oci: Something is wrong if `make` is writing this file.  It should be a "phony" target (a makefile goal that doesn't represent a file).

dds/idl/be_init.cpp
```diff
@@ -7,13 +7,16 @@
 
 #include "../Version.h"
 
-#include "tao/Version.h"
-#include "global_extern.h"
+#include <ace/OS_NS_stdlib.h>
```
mitza-oci: Does it need to be first?  Usual order would be dds, tao, ace, stdlib.

dds/idl/topic_keys.cpp
```diff
@@ -75,7 +75,8 @@ TopicKeys::Iterator::Iterator()
     root_type_(InvalidType),
     level_(0),
     recursive_(false),
-    element_count_(0)
+    element_count_(0),
```
mitza-oci: This is why the recommended style puts commas on the left...

dds/idl/topic_keys.cpp
```diff
@@ -190,17 +195,37 @@ TopicKeys::Iterator::operator++()
...
+      // If a nested struct marked as a key has no keys, all the fields are
+      // implied to be keys.
+      if (pos_ == 0) {
+        if (level_ > 0) {
+          implied_keys_ = true;
+          for (; pos_ < field_count && implied_keys_; ++pos_) {
+            AST_Field** field_ptrptr;
+            struct_root->field(field_ptrptr, pos_);
+            AST_Field* field = *field_ptrptr;
+            if (be_global->is_key(field)) {
```
mitza-oci: Does this need to be recursive?  Struct A (`@topic`) has a field of type struct B, which has a field of type struct C which has some `@key` annotated field.

.gitignore
iguessthislldo: I was using `make --touch` to avoid having to partial rebuild of OpenDDS after changing `opendds_idl`. It worked, but it created a bunch of these files. I can see now that you're right, it's not marked as `PHONY:` in `rules.local.GNU`, so it's getting created when using `make --touch`. I will submit a ACE/TAO PR for this.

dds/idl/topic_keys.cpp
iguessthislldo: No, per my understanding, in that situation you're describing `C`'s key is not a key of `A`; the topic type `A` has to have a `@key` on the field to get the key search started in that branch of the data type.

```
struct C {
  @key long key;
  long not_key;
};

struct B {
  C c_value;
};

@topic
struct A {
  B b_value;
};
```

Sorta unrelated: I'm wondering what should happen in this situation:

```
struct C {
  @key long key;
  long not_key;
};

struct B {
  @key(FALSE) C c_value;
};

@topic
struct A {
  @key B b_value;
};
```

As written in this PR, `@key(FALSE)` is equivalent to no annotation at all, so the key on `c_value` would be implied to be a key in this case. I'm wondering if there should be an exception though for `@key(FALSE)` since it's so explicitly saying it's not a key and this should also be part of the spec.

dds/idl/be_init.cpp
iguessthislldo: I wasn't aware there was a usual order. My preference, although I didn't really follow it here, to do companion header (if applicable source file), then stdlib, 3rd party libs (ace then tao), and finally local headers. I'm willing to follow that order though if it's the convention. Should the order you're talking about be in the guidelines?

iguessthislldo: I wasn't aware there was a usual order. My preference, although I didn't really follow it here, to do companion header (if applicable source file), then stdlib, 3rd party libs (ace then tao), and finally local headers. I'm willing to follow that order though if it's the convention. Should the order you're talking about be in the guidelines?

.gitignore
iguessthislldo: Done: https://github.com/DOCGroup/ACE_TAO/pull/975

iguessthislldo: Done: https://github.com/DOCGroup/ACE_TAO/pull/975

dds/idl/topic_keys.cpp
iguessthislldo: Actually I'm going to go ahead and implement this. That is the only scenario I can think of where being able to say `@key(FALSE)` makes sense. Can you add this to the implied key spec issue?

iguessthislldo:
Another question: do implied keys apply to unions? If we're being consistent I think these rules should apply to union discriminators as well:

If a union field or array field in a struct is marked as a key or implied to be a key, then the discriminator is assumed to be the key unless marked with `@key(FALSE)`.

iguessthislldo:
I wonder if the Github user key has blocked me by now.

iguessthislldo:
Another thought following my previous ones: If a user has marked a field in a struct of another type (union or struct) and all the possible keys have been marked with @key(FALSE), what does that mean? Is it an error or a possible chance to warn the user? Right now I think it would error with structs, which is the behavior before this PR.

mitza-oci:
> I wonder if the Github user key has blocked me by now.

Right... I'm trying to use `@key` formatting but I may forget sometimes.

mitza-oci:
> Another question: do implied keys apply to unions? If we're being consistent I think these rules should apply to union discriminators as well:
> 
> If a union field or array field in a struct is marked as a key or implied to be a key, then the discriminator is assumed to be the key unless marked with `@key(FALSE)`.

I could see it going either way.  The other argument would be that the semantics of `@key` in the outer struct seem to be saying "use this whole thing as a key," and that doesn't apply to the union example.

iguessthislldo:
So I'm kinda confused. I changed the PR Jenkins configure step like 2 days ago to use `--ace-github-latest`, because it couldn't compile opendds_idl, but it's still acting like it doesn't have [ACE/TAO #967](https://github.com/DOCGroup/ACE_TAO/pull/967) and giving the same error...

mitza-oci:
> > I could see it going either way. The other argument would be that the semantics of `@key` in the outer struct seem to be saying "use this whole thing as a key," and that doesn't apply to the union example.
> 
> I wasn't thinking of it applying it to a struct necessarily, I was thinking of applying it directly to the union field like so:
> 
> ```
> union UnionType switch (int) {
> // ...
> };
> 
> @topic
> struct StructType {
>   @key UnionType union_value;
> };
> ```

This is exactly the one I was referring to.  Does it make sense to say that "union_value is the key of StructType"?  It's not really, only the discriminator can be a key so the union value as a whole isn't.

dds/idl/be_global.h
```diff
@@ -173,19 +174,27 @@ class BE_GlobalData {
   bool root_default_nested() const;
 
   /**
-   * Check if a struct field has been declared a key.
+   * If node has the key annotation, this sets value to the key annotation
+   * value and returns true.
+   * If node does not have the key annotation, this sets value false and
```
mitza-oci: "sets value false" or "sets value to the default"?

dds/idl/be_global.cpp
```diff
@@ -650,32 +647,46 @@ BE_GlobalData::is_default_nested(UTL_Scope* scope)
   return root_default_nested_; // True if --default-nested was passed
 }
 
-bool
-BE_GlobalData::is_key(AST_Field* node)
+bool BE_GlobalData::check_key(AST_Field* node, bool& value)
 {
   KeyAnnotation* key = dynamic_cast<KeyAnnotation*>(builtin_annotations_["::@key"]);
-  return key->node_value(dynamic_cast<AST_Decl*>(node));
+  return key->node_value_exists(dynamic_cast<AST_Decl*>(node), value);
```
mitza-oci: Why is this casting?  A field is a decl.

dds/idl/be_init.cpp
```diff
@@ -56,7 +58,10 @@ BE_post_init(char*[], long)
   ACE_CString included;
   DRV_add_include_path(included, include_dds.c_str(), 0, true);
 
-  if (idl_global->idl_version_ >= IDL_VERSION_4) {
+  if (idl_global->idl_version_ < IDL_VERSION_4) {
+    idl_global->ignore_files_ = true; // Exit without parsing files
+    be_global->error("opendds_idl requires IDL version to be 4 or greater");
```
mitza-oci: When would this error occur?  Is the user explicitly setting a version?  Should that be allowed, or should the BE just set the version it wants to use?

dds/idl/topic_keys.cpp
```diff
@@ -190,17 +191,42 @@ TopicKeys::Iterator::operator++()
...
         throw Error(root_, "Invalid Key Iterator");
       }
-      ACE_CDR::ULong field_count = struct_root->nfields();
+      const ACE_CDR::ULong field_count = struct_root->nfields();
+      // If a nested struct marked as a key has no keys, all the fields are
+      // implied to be keys (expect those marked with @key(FALSE)).
+      if (pos_ == 0) {
+        if (level_ > 0) {
+          implied_keys_ = true;
+          for (; pos_ < field_count && implied_keys_; ++pos_) {
```
```suggestion
          for (int field_idx = 0; field_idx < field_count && implied_keys_; ++field_idx) {
```
This pass doesn't need to touch `pos_`

dds/idl/topic_keys.cpp
```diff
@@ -190,17 +191,42 @@ TopicKeys::Iterator::operator++()
...
+            AST_Field** field_ptrptr;
+            struct_root->field(field_ptrptr, pos_);
+            AST_Field* field = *field_ptrptr;
```
mitza-oci: Consider write a helper for this multi-step process of iterating over fields in a struct (since it occurs below as well).  There is already support for something similar in dds_visitor::visit_structure.

dds/idl/topic_keys.cpp
```diff
@@ -190,17 +191,42 @@ TopicKeys::Iterator::operator++()
...
+            if (has_key_annotation && key_annotation_value) {
+              implied_keys_ = false;
+            }
+          }
+          pos_ = 0;
+        } else {
+          implied_keys_ = false;
+        }
+      }
+      // Iterate over the fields
```
mitza-oci: This is already clearly what the loop is doing, no need to add a comment.

tests/DCPS/Compiler/key_annotation/main.cpp
```diff
@@ -84,6 +86,9 @@ int ACE_TMAIN(int, ACE_TCHAR**)
   failed |= assert_key_count<KeyedUnion>(1);
   failed |= assert_key_count<KeyedUnionStruct>(2);
   failed |= assert_key_count<MultidimensionalArrayStruct>(2 * 3 + 2 * 3 * 4);
+  failed |= assert_key_count<ImpliedKeys::StructA>(5);
```
mitza-oci: Could this check for the specific keys in addition to the number of keys?

iguessthislldo: It could and, in fact, I would like it to. I just haven't gotten around to it.

dds/idl/be_init.cpp
iguessthislldo: > When would this error occur? Is the user explicitly setting a version?

Yes.

> Should that be allowed, or should the BE just set the version it wants to use?

That is a good question. My goal in the FE was that the BE and the user would to be able to have some sort of control over IDL version, which is why I have a "default default" version for TAO_IDL_FE to set, a default version that the BE can set, and the version that the user can set. However it doesn't currently cover a minimum version and I came to conclusion that `opendds_idl` should probably ultimately restrict itself to >=IDL4 given we would like people to use annotations. However this isn't strictly necessary to enforce that at the moment, so I'm willing to change it back.

dds/idl/be_global.h
iguessthislldo: It should say "sets value to false".

dds/idl/topic_keys.cpp
iguessthislldo: I'm not sure I understand what you're suggesting here. `dds_visitor::visit_structure` is rewrapping the fields in a vector. It makes sense to avoid using this awkward code in the all generators, but here I would have to either store the vector in the iterator or recreate the vector on each call to `++`. Both of those seem more awkward than direct access, unless I'm not seeing something.

dds/idl/be_init.cpp
mitza-oci: As to what the usual order is (and why), there's some good background info here:
https://stackoverflow.com/a/14243688

dds/idl/be_init.cpp
mitza-oci: I guess what I was missing when I wrote my comment was that the FE is already parsing `--idl-version` and we can't turn it off from the BE.  So that we don't expand the scope of this change to the FE, I'm OK leaving that for now.  We could change it later -- it would be good to have more modular functionality that the BE (really the driver) can decide to invoke from the FE.  There are some other tao_idl issues here that are out of scope for this PR.

Since parsing version 4 is backwards-compatible, I can't think of a reason it wouldn't be set that way (aside from accidental misconfiguration).

dds/idl/topic_keys.cpp
mitza-oci: I started out thinking that it would be good to have a real iterator type (not necessarily a vector, an iterator that just wraps an integer and has a pointer to the ast node).  In topic_keys, that iterator could be stored in your iterator.  In dds_visitor, the vector constructor that takes a pair of iterators could be used.

But if that's not practical, we could have single helper function like `AST_Field* struct_field(AST_Struct*, int)` could be used by all 3.

dds/idl/be_init.cpp
iguessthislldo: That makes sense. This might have prevented some of the trouble I had with https://github.com/objectcomputing/OpenDDS/pull/1339/commits/7424d65e550153435296fdcee9e3b85912659fba. I've included this rule in #1398 and will definitely try to follow it in the future.

dds/idl/topic_keys.cpp
iguessthislldo: Okay, I was too transfixed by the idea of using a vector. I will look into creating a simple iterator.

iguessthislldo: Also I forgot about this, but I added an `operator[size_t]` to `AST_Structure` in the original annotation work. It does have the minor problem of returning `AST_Decl*` instead of `AST_Field*` though, which was an oversight on my part.

PR #1341, opened 2019-10-24
SPDP improvements for the RtpsRelay

jrw972:
* Add INFO_DST to SPDP messages that are sent in response to
  discovered a new participant.  This allows the RtpsRelay to only
  forward the SPDP message to the intended destination.
* Both SPDP and RTPS transports send a beacon to the RtpsRelay.  The
  period of the beacon is configurable.  The beacon is sent
  unconditionally.
* Polling was converted to events in SPDP to handle authentication
  timeouts and resends.
* Three helper classes:
  * JobQueue - allows a thread to post a job on a Reactor.
  * PeriodicTask - a wrapper for periodic tasks.
  * SporadicTask - a wrapper for sporadic (one-shot) tasks.
* ReactorTask has an implicit ReactorInterceptor.

As an aside, the JobQueue was motivated by a deadlock where a
ReactorInterceptor command acquired a lock so that a subsequenct
execute_or_enqueue was not safe because the same lock was already
held.  This suggests that there are probably other situations where
ReactorInterceptor commands are acquiring locks which may result in
the same problem.  I recommend the following:
  1. Don't use the ReactorInterceptor as a general-purpose job queue.
     Use JobQueue.
  2. Limit ReactorInterceptor commands to dealing with the reactor and
     perhaps a small number of control variables.  Furthermore, all state
     updates should be performed by ReactorInterceptor commands.  This
     implicitly provides atomicity.  PeriodicTask and
     SporadicTask are examples of this.
  3. Avoid waiting for ReactorInterceptor commands.  Specifically, it
     is common to wait for a timer id so that it can be cancelled
     later.  An alternative approach would be to writer a helper and
     use multiple instances so the timer id is not necessary.
     Alternatively, the helper can maintain a list of timer ids.  In
     either case, writing a helper is good because it separates out
     the complexity of deadling with the reactor.

### DIFF
--- a/dds/DCPS/JobQueue.h
+++ b/dds/DCPS/JobQueue.h
@@ -1,0 +1,69 @@
+/*
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_JOB_QUEUE_H
+#define OPENDDS_DCPS_JOB_QUEUE_H
+
+#include "RcEventHandler.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+class JobQueue : public RcEventHandler {
+public:
+  class Job : public RcObject {
+  public:
+    virtual ~Job() { }
+    virtual void execute() = 0;
+  };
+  typedef RcHandle<Job> JobPtr;
+
+  explicit JobQueue(ACE_Reactor* reactor)
+  {
+    this->reactor(reactor);
+  }
+
+  void enqueue(JobPtr job)
+  {
+    ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
+    const bool empty = job_queue_.empty();
+    job_queue_.push(job);
+    if (empty) {
+      reactor()->notify(this);
+    }
+  }
+
+private:
+  RcHandle<ReactorInterceptor> interceptor_;
+  ACE_Thread_Mutex mutex_;
+  OPENDDS_QUEUE(JobPtr) job_queue_;
+
+  int handle_exception(ACE_HANDLE /*fd*/)
+  {
+    ACE_Reverse_Lock<ACE_Thread_Mutex> rev_lock(mutex_);
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, mutex_, -1);
+    size_t count = job_queue_.size();
+    while (count--) {
+      JobPtr job = job_queue_.front();
+      job_queue_.pop();
+      {
+        ACE_GUARD_RETURN(ACE_Reverse_Lock<ACE_Thread_Mutex>, rev_guard, rev_lock, -1);
+        job->execute();
+      }
+    }
+    return 0;
+  }
+};
+
+} // namespace DCPS
+} // namespace OpenDDS
+
+OPENDDS_END_VERSIONED_NAMESPACE_DECL
+
+#endif /* OPENDDS_DCPS_JOB_QUEUE_H  */
--- a/dds/DCPS/JobQueue.h
+++ b/dds/DCPS/JobQueue.h
@@ -57,6 +57,11 @@ private:
         job->execute();
       }
     }
+
+    if (!job_queue_.empty()) {
+      reactor()->notify(this);
+    }
+
     return 0;
   }
 };
--- a/dds/DCPS/PeriodicTask.h
+++ b/dds/DCPS/PeriodicTask.h
@@ -1,0 +1,140 @@
+/*
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_PERIODIC_TASK_H
+#define OPENDDS_DCPS_PERIODIC_TASK_H
+
+#include "RcEventHandler.h"
+#include "ReactorInterceptor.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+class PeriodicTask : public RcEventHandler {
+public:
+  explicit PeriodicTask(RcHandle<ReactorInterceptor> interceptor)
+    : interceptor_(interceptor)
+    , enabled_(false)
+  {
+    reactor(interceptor->reactor());
+  }
+
+  virtual ~PeriodicTask() {}
+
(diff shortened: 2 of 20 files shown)


dds/DCPS/PeriodicTask.h
```diff
@@ -0,0 +1,140 @@
...
+  disable_i()
+  {
+    if (enabled_) {
+      reactor()->cancel_timer(this);
+      enabled_ = false;
+    }
+  }
+};
+
+template <typename Outer>
```
mitza-oci: Is there a better name we can use in place of `Outer`?

dds/DCPS/PeriodicTask.h
```diff
@@ -0,0 +1,140 @@
...
+class PmfPeriodicTask : public PeriodicTask {
+public:
+  typedef void (Outer::*PMF)(const MonotonicTimePoint&);
+
+  PmfPeriodicTask(RcHandle<ReactorInterceptor> interceptor, Outer* outer, PMF function)
```
mitza-oci: Should this support reference counting on `outer`?  Could be opt-in using RcObject?

dds/DCPS/ReactorTask.h
```diff
@@ -67,6 +70,18 @@ public virtual RcObject {
 
   enum State { STATE_NOT_RUNNING, STATE_OPENING, STATE_RUNNING };
 
+  class Interceptor : public DCPS::ReactorInterceptor {
+  public:
+    Interceptor(DCPS::ReactorTask* task) : ReactorInterceptor(task->get_reactor(), task->get_reactor_owner()), task_(task) {}
```
mitza-oci:
```suggestion
    explicit Interceptor(DCPS::ReactorTask* task)
     : ReactorInterceptor(task->get_reactor(), task->get_reactor_owner())
     , task_(task)
     {}
```

dds/DCPS/ReactorTask.h
```diff
@@ -67,6 +70,18 @@ public virtual RcObject {
...
+    bool reactor_is_shut_down() const
+    {
+      return task_->is_shut_down();
+    }
+
+  private:
+    DCPS::ReactorTask* task_;
```
mitza-oci:
```suggestion
    DCPS::ReactorTask* const task_;
```

dds/DCPS/JobQueue.h
```diff
@@ -0,0 +1,69 @@
...
+  int handle_exception(ACE_HANDLE /*fd*/)
+  {
+    ACE_Reverse_Lock<ACE_Thread_Mutex> rev_lock(mutex_);
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, mutex_, -1);
+    size_t count = job_queue_.size();
+    while (count--) {
+      JobPtr job = job_queue_.front();
+      job_queue_.pop();
+      {
+        ACE_GUARD_RETURN(ACE_Reverse_Lock<ACE_Thread_Mutex>, rev_guard, rev_lock, -1);
```
mitza-oci: Is `count` from before the while loop still accurate after unlocking?

dds/DCPS/PeriodicTask.h
```diff
@@ -0,0 +1,140 @@
...
+      periodic_task_->enable_i(reenable_, period_);
+    }
+
+    PeriodicTask* const periodic_task_;
+    const bool reenable_;
+    const TimeDuration period_;
+  };
+
+  struct ScheduleDisableCommand : public ReactorInterceptor::Command {
+    ScheduleDisableCommand(PeriodicTask* hb)
```
mitza-oci:
```suggestion
    explicit ScheduleDisableCommand(PeriodicTask* hb)
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -332,7 +332,19 @@ Sedp::init(const RepoId& guid,
   }
 
   {
-    ACE_GUARD_RETURN(ACE_Thread_Mutex, g, rtps_inst->lock_, -1);
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, g, rtps_inst->rtps_relay_address_lock_, -1);
+    rtps_inst->rtps_relay_address_ = disco.sedp_rtps_relay_address();
+  }
+
+  {
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, g, rtps_inst->rtps_relay_address_lock_, -1);
```
mitza-oci: Should this object be renamed to `relay_configuration_lock_` or similar?

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -109,6 +110,8 @@ RtpsUdpInst::load(ACE_Configuration_Heap& cf,
     ACE_INET_Addr addr(rtps_relay_address_s.c_str());
     rtps_relay_address(addr);
   }
+  GET_CONFIG_TIME_VALUE(cf, sect, ACE_TEXT("DataRtpsRelayBeaconPeriod"),
```
```suggestion
  GET_CONFIG_TIME_VALUE(cf, sect, ACE_TEXT("RtpsRelayBeaconPeriod"),
```

dds/DCPS/transport/rtps_udp/RtpsUdpInst.h
```diff
@@ -98,6 +100,8 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpInst : public TransportInst {
   ACE_INET_Addr local_address_;
   OPENDDS_STRING local_address_config_str_;
   ACE_INET_Addr rtps_relay_address_;
+  mutable ACE_SYNCH_MUTEX rtps_relay_address_lock_;
```
mitza-oci:
```suggestion
  mutable ACE_SYNCH_MUTEX rtps_relay_config_lock_;
```

dds/DCPS/PeriodicTask.h
jrw972: Yes.  How about `Delegate`?

dds/DCPS/PeriodicTask.h
jrw972: That is a great question.  If you look at how it is used in Spdp, then it would create a cycle with SpdpTransport.  How about a WeakRcHandle?

mitza-oci: Yes, WeakRcHandle is good for this.

dds/DCPS/JobQueue.h
jrw972: Intentionally no.  I wanted to allow the callback to add new events to the queue.  This may create livelock if not bounded so I sample the queue size before and then only process that many items.

PR #1342, opened 2019-10-25
Add test for a late joining relay

jrw972:


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -332,6 +332,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     publisher->get_default_datawriter_qos(writer_qos);
 
     writer_qos.durability.kind = DDS::TRANSIENT_LOCAL_DURABILITY_QOS;
+    writer_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
```
mitza-oci: Does this need to be here?  Writers default to reliable.  What is the reader set to?

jrw972: No.  I just wanted it to be explicit.

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -412,6 +413,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     publisher->get_default_datawriter_qos(writer_qos);
 
     writer_qos.durability.kind = DDS::TRANSIENT_LOCAL_DURABILITY_QOS;
+    writer_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;;
```
jrw972:
```suggestion
    writer_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
```

PR #1347, opened 2019-10-31
Fix CMake Library Scope Issue #1336

iguessthislldo:
Fix for #1336 

As stated in the issue:

> Alright, so this issue is my fault. In #1256 I needed to link `Threads::Threads` for inlined code, but I did that by calling `target_link_libraries` directly on the user target instead of declaring it as a interface dependency of ACE like I did with the other dependencies. I don't think I was completely aware of how dependencies were working. I will create a new PR to fix this.

This adds `Threads::Threads` to ACE's `INTERFACE_LINK_LIBRARIES` using the existing machinery for that. I'm also fixing a similar issue with Xerces that would come up with trying to compile with security enabled.

Also:
- Allow test messenger to find key files in `tests/security` for security using `DDS_ROOT` instead of assuming it's relative to the normal `tests/DCPS/Messenger` location. This allows CMake messenger tests to run with security as long as `DDS_ROOT` is set.
- Add checks for trailing whitespace and tabs in CMake files to `dds_fuzz.pl`.


tests/DCPS/Messenger/publisher.cpp
```diff
@@ -86,12 +86,26 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
...
       append(props, "OpenDDS.RtpsRelay.Groups", "Messenger", true);
 
-#if defined(OPENDDS_SECURITY)
+#ifdef OPENDDS_SECURITY
+      // Determine the path to the keys
+      OPENDDS_STRING path_to_tests;
+      const char* dds_root = ACE_OS::getenv("DDS_ROOT");
+      if (dds_root && dds_root[0]) {
+        // Use DDS_ROOT in case we are one of the CMake tests
+        path_to_tests = OPENDDS_STRING("file:") + dds_root + "/tests/";
```
mitza-oci: This doesn't appear to work with an installed OpenDDS.  We should be supporting that and testing it with the CMake tests.  Is it in one of the CI builds?

cmake/OpenDDSConfig.cmake
```diff
@@ -222,10 +257,14 @@ set(OPENDDS_RTPS_UDP_DEPS
...
 
-set(OPENDDS_SECURITY_DEPS
-  OpenDDS::Rtps
-  ACE::XML_Utils
-)
+if(OPENDDS_SECURITY)
+  set(OPENDDS_SECURITY_DEPS
+    OpenDDS::Rtps
+    ACE::XML_Utils
+    OpenSSL::SSL
```
mitza-oci: SSL shouldn't actually be a dependency, only Crypto.  Not sure if we'd need to change something in the OpenDDS build itself to reflect this.  At least ldd only shows libcrypto.

tests/DCPS/Messenger/publisher.cpp
iguessthislldo: > This doesn't appear to work with an installed OpenDDS.

No it wouldn't, we would have to have it know where the security files were or maybe copy the files like we copy the rest of the files.

> We should be supporting that and testing it with the CMake tests. Is it in one of the CI builds?

It should get built when the build is doing security, but `run_ci_tests just does` just runs `run_test.pl` without any arguments, so it would only be running inforepo/tcp.

For tick it looks like it builds with installed secure OpenDDS but it doesn't look like I set any of the scoreboard builds to run the cmake tests at all.

cmake/OpenDDSConfig.cmake
iguessthislldo: I'm just trying to match what MPC is doing.

iguessthislldo:
@mitza-oci I didn't mean to request you review on this again.

mitza-oci:
There's a merge conflict on this one

tests/cmake_integration/Messenger/Messenger_1/CMakeLists.txt
```diff
@@ -44,7 +44,9 @@ OPENDDS_TARGET_SOURCES(subscriber
 )
 
 foreach(t ${all_targets})
-  target_link_libraries(${t} OpenDDS::OpenDDS)
+  # Use of PRIVATE library scope here is a regression test for
+  # https://github.com/objectcomputing/OpenDDS/issues/1336
+  target_link_libraries(${t} PRIVATE OpenDDS::OpenDDS)
```
mitza-oci: This is probably due to my lack of in-depth CMake knowledge.  What does it mean for an executable to have a private library dependency?

Also the `foreach` here looks very strange, but I'm used to declarative instead of procedural build definitions.

iguessthislldo: I think that questions is a valid one, because as far as I know it doesn't mean anything. It should just test that CMP0023 is happy.

iguessthislldo: I think that questions is a valid one, because as far as I know it doesn't mean anything. It should just test that CMP0023 is happy.

mitza-oci: Let's move that elsewhere, since this file is essentially here as documentation of how the CMake support should be used.

iguessthislldo: I'm not sure where else this would go. Also I'm not so sure this is a good CMake example given we're copying the messenger test files here. That's part of the reason I added a CMakeLists.txt to the devguide messenger in #1287.

mitza-oci: Could we just remove it then?

iguessthislldo: Made it PUBLIC scope as discussed.

PR #1349, opened 2019-10-31
RtpsRelay:  Relay addresses not updated when client changes servers

jrw972:
Previously, the relay addresses were bundled with the reader and
writer entries.  Suppose there are two relays A and B and a client
connects to A, switches to B, and then switches back to A before the
participant lease duration.  This causes A to write entries for the
participant and then B will write entries for the participant.
However, the client is associated with A but the topic says it should
be associated with B.

To solve this problem, remove the relay addresses from the reader and
writer entries and create a new topic for participant to relay
pairings (Responsible Relay).  The relays maintain the Responsible
Relay topic based on the traffic that they are observing.  The Reader
and Writer entries are just for matching.

For this change, we had to remove the local/remote distinction in the
QosIndex.  This means that every relay must maintain a complete
association table.

### DIFF
--- a/tools/rtpsrelay/AssociationTable.h
+++ b/tools/rtpsrelay/AssociationTable.h
@@ -11,23 +11,17 @@ namespace RtpsRelay {
 
 class AssociationTable {
 public:
-  explicit AssociationTable(const RelayAddresses& relay_addresses) :
-    relay_addresses_(relay_addresses)
-  {}
   void insert(const WriterEntry& entry,
-              RelayAddressesMap& relay_addresses_map);
+              GuidSet& guids);
   void remove(const WriterEntry& entry);
   void insert(const ReaderEntry& entry,
-              RelayAddressesMap& relay_addresses_map);
+              GuidSet& guids);
   void remove(const ReaderEntry& entry);
 
-  void populate_relay_addresses_map(RelayAddressesMap& relay_addresses_map,
-                                    const OpenDDS::DCPS::RepoId& from,
-                                    const GuidSet& to) const;
+  void lookup_destinations(GuidSet& to,
+                           const OpenDDS::DCPS::RepoId& from) const;
 
 private:
-  const RelayAddresses& relay_addresses_;
-
   typedef std::map<OpenDDS::DCPS::RepoId, WriterPtr, OpenDDS::DCPS::GUID_tKeyLessThan> WritersMap;
   WritersMap writers_;
   typedef std::map<OpenDDS::DCPS::RepoId, ReaderPtr, OpenDDS::DCPS::GUID_tKeyLessThan> ReadersMap;
--- a/tools/rtpsrelay/RelayHandler.cpp
+++ b/tools/rtpsrelay/RelayHandler.cpp
@@ -147,12 +147,16 @@ void RelayHandler::enqueue_message(const std::string& addr, const OpenDDS::DCPS:
 VerticalHandler::VerticalHandler(ACE_Reactor* reactor,
                                  const RelayAddresses& relay_addresses,
                                  const AssociationTable& association_table,
+                                 GuidRelayAddressesDataWriter_ptr responsible_relay_writer,
+                                 GuidRelayAddressesDataReader_ptr responsible_relay_reader,
                                  const OpenDDS::DCPS::TimeDuration& lifespan,
                                  const OpenDDS::RTPS::RtpsDiscovery_rch& rtps_discovery,
                                  DDS::DomainId_t application_domain,
                                  const OpenDDS::DCPS::RepoId& application_participant_guid,
                                  const CRYPTO_TYPE& crypto)
   : RelayHandler(reactor, association_table)
+  , responsible_relay_writer_(responsible_relay_writer)
+  , responsible_relay_reader_(responsible_relay_reader)
   , relay_addresses_(relay_addresses)
   , horizontal_handler_(nullptr)
   , lifespan_(lifespan)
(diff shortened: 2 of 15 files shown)


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -266,193 +266,207 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
 
-  SpdpHandler spdp_vertical_handler(reactor, relay_addresses, association_table, lifespan, rtps_discovery, application_domain, application_participant_id, crypto, spdp);
-  SedpHandler sedp_vertical_handler(reactor, relay_addresses, association_table, lifespan, rtps_discovery, application_domain, application_participant_id, crypto, sedp);
-  DataHandler data_vertical_handler(reactor, relay_addresses, association_table, lifespan, rtps_discovery, application_domain, application_participant_id, crypto);
+  DDS::Subscriber_var bit_subscriber = application_participant->get_builtin_subscriber();
 
-  spdp_horizontal_handler.vertical_handler(&spdp_vertical_handler);
-  sedp_horizontal_handler.vertical_handler(&sedp_vertical_handler);
-  data_horizontal_handler.vertical_handler(&data_vertical_handler);
+  GuidRelayAddressesTypeSupportImpl::_var_type guid_relay_addresses_ts =
```
mitza-oci:
```suggestion
  GuidRelayAddressesTypeSupport_var guid_relay_addresses_ts =
```
Not a big deal but I'd rather stick to the standard mapping in this case (also below with WriterEntry, just search for _var_type).

tools/rtpsrelay/RelayHandler.h
```diff
@@ -108,9 +110,13 @@ class VerticalHandler : public RelayHandler {
   void process_message(const ACE_INET_Addr& remote,
                        const OpenDDS::DCPS::MonotonicTimePoint& now,
                        const OpenDDS::DCPS::Message_Block_Shared_Ptr& msg) override;
-  void send(const RelayAddressesMap& relay_addresses_map,
+  void send(const GuidSet& to,
             const OpenDDS::DCPS::Message_Block_Shared_Ptr& msg);
+  void populate_relay_addresses_map(RelayAddressesMap& relay_addresses_map,
```
mitza-oci: Would this be cleaner to just have the `RelayAddressesMap` as the return type?  We can now assume move semantics and probably some level of copy elision.

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -371,6 +385,67 @@ void VerticalHandler::send(const RelayAddressesMap& relay_addresses_map,
...
+  GuidRelayAddressesSeq received_data;
+  DDS::SampleInfoSeq info_seq;
+  const auto ret = responsible_relay_reader_->read_instance(received_data, info_seq, 1, handle,
+                                                            DDS::ANY_SAMPLE_STATE, DDS::ANY_VIEW_STATE, DDS::ANY_INSTANCE_STATE);
+  if (ret != DDS::RETCODE_OK) {
+    ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: VerticalHandler::read_relay_addresses failed to read\n"));
+    return relay_addresses;
+  }
+
+  return received_data[0].relay_addresses();
```
mitza-oci: This uses the result of the "unconstrained" (ANY_INSTANCE_STATE) read without checking valid_data or other state flags in SampleInfo.  Are there similar cases in the relay that need to be checked?

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -266,193 +266,207 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+    return EXIT_FAILURE;
+  }
 
-    if (!reader) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: failed to create Reader data reader\n"));
-      return EXIT_FAILURE;
-    }
+  DDS::DataReader_var responsible_relay_reader_var =
+    relay_subscriber->create_datareader(responsible_relay_topic, reader_qos, nullptr,
+                                        DDS::DATA_AVAILABLE_STATUS);
```
mitza-oci: Why does this specify `DATA_AVAILABLE_STATUS` when there's no listener?

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -266,193 +266,207 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+    ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: failed to narrow Reader data writer\n"));
+    return EXIT_FAILURE;
+  }
 
-    if (!writer) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) %N:%l ERROR: failed to create Writer data writer\n"));
-      return EXIT_FAILURE;
-    }
+  DDS::DataReader_var reader_reader_var = relay_subscriber->create_datareader(readers_topic, reader_qos,
+                                                                              new ReaderListener(association_table, spdp_vertical_handler),
```
mitza-oci: Looks like it will leak this listener object.

tools/rtpsrelay/AssociationTable.h
```diff
@@ -11,23 +11,17 @@ namespace RtpsRelay {
 
 class AssociationTable {
 public:
-  explicit AssociationTable(const RelayAddresses& relay_addresses) :
-    relay_addresses_(relay_addresses)
-  {}
   void insert(const WriterEntry& entry,
-              RelayAddressesMap& relay_addresses_map);
+              GuidSet& guids);
```
mitza-oci: see comment in `RelayHandler.h` about using the return type instead of out-parameters.

jrw972: I would like to use an out parameter here because buried inside the qos index code is a loop that populates the map.  If I adopt this strategy everywhere, then I'm back to copying.

PR #1354, opened 2019-11-01
Spdp uses destructed lock on shutdown

jrw972:
The SPDP transports sends unregister/dispose messages as part of
shutdown.  If it is using the RtpsRelay, then it must acquire the lock
for the address.  However, the lock has already been destroyed.

The solution was to move all of the configuration for RtpsDiscovery to
a separate configuration object that will not be destroyed until the
Spdp object is destroyed.

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscovery.cpp
+++ b/dds/DCPS/RTPS/RtpsDiscovery.cpp
@@ -41,9 +41,8 @@ namespace RTPS {
 
 using DCPS::TimeDuration;
 
-RtpsDiscovery::RtpsDiscovery(const RepoKey& key)
-  : DCPS::PeerDiscovery<Spdp>(key)
-  , resend_period_(30 /*seconds*/) // see RTPS v2.1 9.6.1.4.2
+RtpsDiscoveryConfig::RtpsDiscoveryConfig()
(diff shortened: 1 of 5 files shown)


simpsont-oci:
Should the RtpsDiscoveryConfig move to it's own .h/.cpp?

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -505,8 +510,9 @@ OpenDDS::DCPS::RepoId
 RtpsDiscovery::generate_participant_guid() {
   OpenDDS::DCPS::RepoId id = GUID_UNKNOWN;
   ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_, id);
-  if (!guid_interface_.empty()) {
-    if (guid_gen_.interfaceName(guid_interface_.c_str()) != 0) {
+  OPENDDS_STRING guid_interface = config_->guid_interface();
```
iguessthislldo: It looks like this and the next `guide_interface` strings can be const.

iguessthislldo:
> Should the RtpsDiscoveryConfig move to it's own .h/.cpp?

I think it's okay when classes are this tightly coupled.

PR #1358, opened 2019-11-03
fixing wchar support for xerces and security builds

simpsont-oci:


dds/DCPS/QOS_XML_Handler/QOS_XML_Loader.cpp
```diff
@@ -86,72 +86,72 @@ namespace DCPS {
...
     if (!profile_name)
       {
         if (DCPS_debug_level > 5)
           {
-            ACE_ERROR ((LM_ERROR,
-              ACE_TEXT ("QOS_XML_Loader::get_datawriter_qos - ")
-              ACE_TEXT ("Error parsing profile string <%C>, returning.\n"),
+            ACE_ERROR((LM_ERROR,
+              ACE_TEXT("QOS_XML_Loader::get_datawriter_qos - ")
+              ACE_TEXT("Error parsing profile string <%C>, returning.\n"),
```
jwillemsen: Loggin tchar schould use %s instead of %C

simpsont-oci: Thanks, fixed a bunch of those.

jwillemsen:
Also check the ACE_ERROR usage, they should use LM_ERROR, seems some are using LM_DEBUG

iguessthislldo:
> Also check the ACE_ERROR usage, they should use LM_ERROR, seems some are using LM_DEBUG

Yes, also:

- The minimum debug levels for errors should either zero or 1 depending on the situation, but definitely not 7.
- The messages should say they're an error and have `(%P|%t)` like:
https://github.com/objectcomputing/OpenDDS/blob/c27a8738708867ae326f61e9609d66ce9fe79fc1/dds/DCPS/BitPubListenerImpl.cpp#L41-L43

dds/DCPS/QOS_XML_Handler/QOS_XML_Loader.cpp
```diff
@@ -8,76 +8,76 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
+                             qos_profile));
               }
             return 0;
           }
 
-        char * ret = (char*)ACE_OS::malloc (ACE_OS::strlen (file_name) + 5);
-        ret = ACE_OS::strcpy (ret, file_name);
-        ret = ACE_OS::strcat (ret, ".xml");
-        ACE_OS::free (buf);
+        ACE_TCHAR * ret = (ACE_TCHAR*) ACE_OS::malloc(ACE_OS::strlen(file_name) + 5);
```
mitza-oci: This doesn't look right, strlen() measures characters while malloc() wants bytes.  In general there's no reason to avoid using a class like TString or even std::basic_string that does all of this automatically.  (I realize it was like this before.)  Please see the WCHAR notes (in docs/design) for more info on using switchable character types.

dds/DCPS/QOS_XML_Handler/QOS_XML_Loader.cpp
```diff
@@ -8,76 +8,75 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
 
-        char * ret = (char*)ACE_OS::malloc (ACE_OS::strlen (file_name) + 5);
-        ret = ACE_OS::strcpy (ret, file_name);
-        ret = ACE_OS::strcat (ret, ".xml");
-        ACE_OS::free (buf);
-        return ret;
+        ACE_TString ret(file_name);
+        ret += ACE_TEXT(".xml");
+        ACE_OS::free(buf);
+        return ret.rep();
```
mitza-oci: Looks like it's returning a pointer that will be freed before it can be used.

dds/DCPS/QOS_XML_Handler/QOS_PubSub_T.cpp
```diff
@@ -50,84 +50,84 @@ QOS_PubSub_T<XML_QOS_TYPE, DDS_QOS_TYPE>::read_qos (DDS_QOS_TYPE& dds_qos, const
...
           dds_qos.partition.name.length(static_cast<CORBA::ULong>(xml_qos->partition().name().count_element()));
           CORBA::ULong pos = 0;
-          for (::dds::stringSeq::element_const_iterator it = xml_qos->partition ().name ().begin_element ();
-               it != xml_qos->partition ().name ().end_element ();
+          for (::dds::stringSeq::element_const_iterator it = xml_qos->partition().name().begin_element();
+               it != xml_qos->partition().name().end_element();
                ++it, ++pos)
             {
-              dds_qos.partition.name[pos] = ::CORBA::string_dup (it->get()->c_str());
+              dds_qos.partition.name[pos] = ACE_NS_WString(it->get()->c_str()).char_rep();
```
mitza-oci: Why is this `ACE_NS_WString`?  It's not clear to me why that would be used over one of the other converters.

dds/DCPS/QOS_XML_Handler/QOS_XML_Loader.cpp
simpsont-oci: http://www.dre.vanderbilt.edu/Doxygen/5.7.9/html/ace/a00651.html#a2bca515bf58fa5e32c67561f3a141a83

"This method allocates memory for a copy of the string and returns a pointer to the new area. The caller is responsible for freeing the memory when finished"

dds/DCPS/QOS_XML_Handler/QOS_PubSub_T.cpp
simpsont-oci: I think this was early in my efforts, didn't realize ACE_TEXT_ALWAYS_* macros existed.

dds/DCPS/QOS_XML_Handler/QOS_XML_Loader.cpp
mitza-oci: OK, I thought that may have been the `c_str` equivalent.  Looking at this one had me thinking we should update the interface to return objects instead of pointers, but I'm not sure if that breaks anything for DDS4CCM.

jwillemsen: Yes, that very likely will need some dds4ccm changes, feel free to update dds4ccm when required

simpsont-oci: Where are we on this? It's doing what it's always done, but now it compiles with wchar support. Am I doing more here?

mitza-oci: ACE's doxygen:
```
   * This method allocates memory for a copy of the string and returns
   * a pointer to the new area. The caller is responsible for freeing
   * the memory when finished; use delete []
```
so the caller of this is now tasked with remembering to use `delete[]` instead of some other function/operator.  I don't think that's what we want.

Now that I'm in front of a full development environment looking at how this is used, I think we should change these two private methods.  This is what `std::basic_string` (with its various template arguments) is for.

PR #1360, opened 2019-11-06
RTPS 2.4 - Add & Apply Security Flag to DATA & DATAFRAG Submessages

kuznetsovmoci:
Also modified Attributes Test app for variable message size to allow testing with Fragments.


tests/security/attributes/SecurityAttributesMessage.idl
```diff
@@ -5,7 +5,9 @@
  * See: http://www.opendds.org/license.html
  */
 
-module SecurityAttributes {
+#include <dds/DdsDcpsCore.idl>
+
+module SecurityAttributes{
```
simpsont-oci:
```suggestion
module SecurityAttributes {
```

tests/security/attributes/Writer.cpp
```diff
@@ -102,6 +102,10 @@ Writer::svc()
     message.subject      = "Review";
     message.text         = "Worst. Movie. Ever.";
     message.count        = 0;
+    message.extra_space.length(args_.extra_space);
+    for (int i = 0; i < args_.extra_space; i++) {
+      message.extra_space[i] = 'a';
```
simpsont-oci: Why don't we go ahead and make this a memset if we're going to set the values at all.

kuznetsovmoci: Done.

PR #1362, opened 2019-11-07
Use Standard C Exit Statuses in Messenger Tests

iguessthislldo:
My understanding is on Windows has a wider range for exit statuses, but POSIX requires applications to return in the range of 0 - 255. If it does not fit in this range it gets truncated, at least on linux (Fun fact: If 256 were returned, it actually will be interpreted as a success because it gets truncated to 0). I already have fixed this for DevGuide Messenger in #1287.


mitza-oci:
I'm not sure there's anything actually wrong with `return -1;`, but I'm not opposed to changing it.  Was there a problem caused by this?

https://pubs.opengroup.org/onlinepubs/9699919799/functions/exit.html

iguessthislldo:
No, not a real problem. I would consider bad practice in general however, because it's not possible to test for that value using `$?`, at least using bash.

mitza-oci:
> 
> 
> No, not a real problem. I would consider bad practice in general however, because it's not possible to test for that value using `$?`, at least using bash.

You can test that `$?` is nonzero, or test for 255 knowing the `waitpid` truncation logic.

iguessthislldo:
> You can test that $? is nonzero, or test for 255 knowing the waitpid truncation logic.

Yes, you could work around this, but I think it would be much simpler if programs were normal and returned 1 or some other low number when they have an error.

mitza-oci:
If we're going to change it we could use `EXIT_SUCCESS` and `EXIT_FAILURE` since "magic numbers" should be avoided.

iguessthislldo:
> If we're going to change it we could use `EXIT_SUCCESS` and `EXIT_FAILURE` since "magic numbers" should be avoided.

I'm fine with that.

PR #1363, opened 2019-11-08
Use ACE Monotonic Time Support Status Macros

iguessthislldo:
Make use of `ACE_HAS_MONOTONIC_TIME_POLICY` and `ACE_HAS_MONOTONIC_CONDITIONS` preprocessor macros from DOCGroup/ACE_TAO#974.


dds/DCPS/DataReaderImpl.h
```diff
@@ -42,7 +42,7 @@
 #include "RcEventHandler.h"
 #include "TopicImpl.h"
 #include "DomainParticipantImpl.h"
-#include "TimeTypes.h"
+#include "time_types.h"
```
mitza-oci: Why did this change?

iguessthislldo: From the Commit:

> I think headers that don't have a "main" class (Ex. WaitSet in
WaitSet.h) shouldn't have a capitalized camel case name like a class.
I meant to do this in the original work, but lost track of it trying to
get the monotonic stuff working.

iguessthislldo: Looking around I guess nothing else is following that logic besides debug.h/cpp.

dds/DCPS/time_types.h
```diff
@@ -38,15 +38,34 @@ typedef TimePoint_T<SystemClock> SystemTimePoint;
...
- * macOS. Use SystemClock instead, because ACE_Monotonic_Time_Policy falls back
- * to returning ACE_Time_Value::zero for some reason.
- */
-typedef SystemClock MonotonicClock;
-#else
+#ifdef ACE_OCI_PATCHLEVEL
+#  if ACE_MAJOR_VERSION > 6 || (ACE_MAJOR_VERSION == 6 && \
+        (ACE_MINOR_VERSION > 2 || (ACE_MINOR_VERSION == 2 && \
+          (ACE_BETA_VERSION > 0 || (ACE_BETA_VERSION == 0 && ACE_OCI_PATCHLEVEL > 16)))))
+#    define OPENDDS_ACE_HAS_MONOTONIC_SUPPORT_MACROS
```
iguessthislldo: Maybe it would be better for ACE to have POSIX style support status macros for this? Ex:
ACE always defines `ACE_HAS_MONOTONIC_TIME_POLICY`, but sets it to -1 if it's not supported. That way I don't have to do this version check. Or alternatively have a macro set for when ACE supports the support macros. I guess that's getting weird at that level, but I think that's just an example of how better the POSIX way is.

dds/DCPS/DataReaderImpl.h
mitza-oci: Yeah, I'd like to avoid this churn unless there's a more compelling reason to have it.

dds/DCPS/time_types.h
mitza-oci: If you do that, consider renaming the ACE macro.  I'm not sure "has monotonic policy" makes sense with that usage.  If ACE is going to change that will need to happen soon.

dds/DCPS/DataReaderImpl.h
iguessthislldo: Alright, then I'll revert it.

dds/DCPS/time_types.h
iguessthislldo: > If you do that, consider renaming the ACE macro. I'm not sure "has monotonic policy" makes sense with that usage.

In what way? Would `ACE_SUPPORTS_MONOTONIC_TIME_POLICY` and `ACE_SUPPORTS_MONOTONIC_CONDITIONS` make more sense?

mitza-oci: Yes, that's along the lines of what I was thinking.  Perhaps even removing `SUPPORTS`?

iguessthislldo: Oh, okay, yeah that makes sense because it's not really a boolean statement any more. The POSIX macros are like that: `_POSIX_MONOTONIC_CLOCK`. Also I guess ACE still "has" `ACE_Monotonic_Time_Policy`, it just wouldn't work.

mitza-oci: There's a longstanding problem with these in that "ACE has something or other" looks like a sentence where ACE is the subject.  There are some macros where this is true.  In other cases what it really means is that "this is an ACE macro so it must start with ACE, and the platform has something or other".  See `ACE/ace/README` for some details.

PR #1365, opened 2019-11-08
RTPS discovery and transport cannot handle network changes

jrw972:
The locators determined by RTPS discovery and used by the RTPS
transport are set at start time.  This prevents OpenDDS from be used
in situations where the network configuration is expected to change
like mobile phones, tablets, etc.

To solve this problem, code was added to detect network changes and
update the locators announced by SPDP and SEDP.  Support was also
added for checking for changed locators and updating the RTPS
transport to use the new locators.  ICE was also updated so that the
published AgentInfo is updated when the network configuration changes.

Only Linux is supported at this time.

Diagnostic logging for these features can be turned on with
DCPSDebugLevel > 4.


dds/DCPS/Service_Participant.cpp
```diff
@@ -435,6 +442,13 @@ Service_Participant::get_domain_participant_factory(int &argc,
       }
 
       this->monitor_.reset(this->monitor_factory_->create_sp_monitor(this));
+
+#ifdef ACE_LINUX
+      network_config_publisher_ = make_rch<LinuxNetworkConfigPublisher>(reactor_task_.interceptor());
```
simpsont-oci: I'd rather see this tied to the choice of Discovery / Transport. Shmem applications aren't going to need to monitor network changes.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -33,6 +34,7 @@
 #include "dds/DCPS/DiscoveryListener.h"
 #include "dds/DCPS/ReactorInterceptor.h"
 #include "dds/DCPS/RcEventHandler.h"
+#include "dds/DCPS/JobQueue.h"
```
simpsont-oci: Have we come full circle? TaskQueue -> ReactorInterceptor -> JobQueue?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -58,7 +60,7 @@ class ReceivedDataSample;
 typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
 typedef RcHandle<RtpsUdpTransport> RtpsUdpTransport_rch;
 
-class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink {
+  class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual NetworkConfigListener {
```
simpsont-oci: I don't think this is the right place for this inheritance. The transport's datalink, as I understand it, is a connection between two transport instances. I think these listener callbacks would impact the transport at a higher level than the datalink, maybe even on the (RtpsUdp)TransportImpl itself. This feels like it needs some design discussion.

jrw972:
> This PR is gigantic. Here are a handful of comments I came up with before giving up and going back to other things.

I'm going to take that as a compliment.

dds/DCPS/DataReaderImpl.h
```diff
@@ -360,6 +360,8 @@ class OpenDDS_Dcps_Export DataReaderImpl
   /// process a message that has been received - could be control or a data sample.
   virtual void data_received(const ReceivedDataSample& sample);
 
+  void network_change();
```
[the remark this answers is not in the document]
jrw972: I agree that `network_change` is too specific.  The intent is that something has changed in the transport that needs to be communicated through discovery...so I think I'll name it `transport_discovery_change`.

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -790,6 +790,19 @@ DataWriterImpl::unregister_for_reader(const RepoId& participant,
   TransportClient::unregister_for_reader(participant, writerid, readerid);
 }
 
+void
+DataWriterImpl::update_locators(const RepoId& readerId,
+                                const TransportLocatorSeq& locators)
+{
+  ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, this->lock_);
```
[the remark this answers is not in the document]
jrw972: Does anyone really know what locks _should_ be held?  I don't think I need the global lock.

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -2712,6 +2725,16 @@ LivenessTimer::handle_timeout(const ACE_Time_Value &tv,
   return 0;
 }
 
+void DataWriterImpl::network_change()
+{
+  this->populate_connection_info();
```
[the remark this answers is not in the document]
jrw972: Agreed.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3245,6 +3245,18 @@ DataReaderImpl::unregister_for_writer(const RepoId& participant,
...
 
+void
+DataReaderImpl::update_locators(const RepoId& writerId,
+                                const TransportLocatorSeq& locators)
+{
+  ACE_READ_GUARD(ACE_RW_Thread_Mutex, read_guard, writers_lock_);
+  WriterMapType::const_iterator iter = writers_.find(writerId);
+
+  if (iter != writers_.end()) {
+    TransportClient::update_locators(writerId, locators);
```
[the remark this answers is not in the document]
jrw972: I got rid of the global lock in DataWriterImpl.  By holding the locks, we ensure that the transport stays in sync with the DataReader/DataWriter.  Otherwise, we can leak locators in the transport.

As to the locking...well...you know.

dds/DCPS/NetworkConfigPublisher.h
```diff
@@ -0,0 +1,94 @@
...
+                   bool can_multicast)
+    : index_(index)
+    , name_(name)
+    , can_multicast_(can_multicast)
+  {}
+  int index() const { return index_; }
+  const OPENDDS_STRING& name() const { return name_; }
+  bool can_multicast() const { return can_multicast_; }
+
+  AddressSet addresses;
```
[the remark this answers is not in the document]
jrw972: Yes.  The `NetworkInterface` objects are by value and const when necessary.

dds/DCPS/Service_Participant.cpp
jrw972: The scenario I'm targeting is RTPS Discovery and RTPS Transport.  After we add ICE, that means we would need to run three copies of the same service to detect network changes.  Transports and discoveries that don't use it won't.  I can see an argument for lazily creating it.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
jrw972: So this is interesting and I've been thinking about it.  If you take the TaskQueue/ActiveObject idea but instead replace it with a thread running a Reactor event loop, then you get a ReactiveObject.  Atomicity is enforced through the event handlers.  Consequently, you should only need one or two locks.  One lock for putting things on a queue for deferred processing (which is in the JobQueue) and another lock to protect data structures.  Essentially, we could use this to implement a message passing scheme between Discovery, Transports, ICE, NetworkConfigPublisher, etc.  If we can ever decouple the DataReaders and DataWriters from the transport framework, this would clean up a lot of the reverse locking madness.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
jrw972: In principle, I agree with you.  It should be a Transport level thing.  If we ever updated say the tcp transport to respond to network changes, then the Transport would be `NetworkConfigListener`.  With that being said however, each Transport is free to implement this however they choose.

For RTPS, the datalink has the information necessary to process network changes.  If the RtpsUdpTransport were the NetworkConfigListener, then it would just be dispatching down to the RtpsUdpDataLink.

Also, if you trace the call sequence then a network change goes from RtpsUdpDatalink -> Datalink -> TransportSendListeners and TransportReceiverListeners -> DataReaderImpl and DataWriterImpl -> Discovery.  I did not see anything in the framework where something higher than the datalink was keeping tracking of all the TransportClients.  If the TransportImpl maintained a list of TransportClients, then it would be a different story.  We could go from TransportImpl -> TransportClient -> Discovery.

So, my current plan is to leave it as is hoping that we can retire the transport framework before we need to make the aforementioned improvements.

dds/DCPS/NetworkConfigPublisher.h
```diff
@@ -0,0 +1,94 @@
...
+
+  bool operator==(int index) const
```
mitza-oci: Non-commutative == may be confusing

dds/DCPS/NetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,130 @@
...
+    listeners = listeners_;
+    NetworkInterfaces::iterator pos = std::find(network_interfaces_.begin(), network_interfaces_.end(), index);
+    if (pos != network_interfaces_.end()) {
+      nic = *pos;
+      network_interfaces_.erase(pos);
+      publish = true;
+    }
+  }
+  if (publish) {
+    while (!nic.addresses.empty()) {
```
mitza-oci: Is this part of the NCP -> Listener interaction needed?  Wouldn't a Listener know from `remote_interface` that all addresses belonging to that interface are effectively removed?

dds/DCPS/LinuxNetworkConfigPublisher.h
```diff
@@ -0,0 +1,47 @@
...
+#include <ace/SOCK_Netlink.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+class OpenDDS_Dcps_Export LinuxNetworkConfigPublisher : public RcEventHandler, public NetworkConfigPublisher {
+public:
+  LinuxNetworkConfigPublisher(ReactorInterceptor_rch interceptor);
```
mitza-oci:
```suggestion
  explicit LinuxNetworkConfigPublisher(ReactorInterceptor_rch interceptor);
```

dds/DCPS/LinuxNetworkConfigPublisher.h
```diff
@@ -0,0 +1,47 @@
...
+  bool open();
+  bool close();
+
+private:
+  ACE_HANDLE get_handle() const;
+  int handle_input(ACE_HANDLE);
+  void read_messages();
+  void process_message(const struct nlmsghdr* header);
```
mitza-oci:
```suggestion
  void process_message(const nlmsghdr* header);
```

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+  reactor(interceptor->reactor());
+}
+
+bool LinuxNetworkConfigPublisher::open()
+{
+  // Listener to changes in links and IPV4 addresses.
+  ACE_Netlink_Addr addr;
+  addr.set(getpid(), RTMGRP_NOTIFY | RTMGRP_LINK | RTMGRP_IPV4_IFADDR);
+  if (socket_.open(addr, AF_NETLINK, NETLINK_ROUTE) != 0) {
+    ACE_ERROR((LM_ERROR, "LinuxNetworkConfigPublisher::open: could not open socket: %m\n"));
```
mitza-oci: Prefer to have some form of "error" in the message itself so that if logging is not set to verbose the user still realizes that it's an error.

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+    return false;
+  }
+
+  read_messages();
+
+  // Request a dump of the addresses.
+  request.header.nlmsg_len = NLMSG_LENGTH(sizeof(request.msg));
+  request.header.nlmsg_type = RTM_GETADDR;
+  request.header.nlmsg_flags = NLM_F_REQUEST | NLM_F_DUMP;
+  request.header.nlmsg_pid = getpid();
```
mitza-oci: `getpid()`'s value probably didn't change since line 55

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+int LinuxNetworkConfigPublisher::handle_input(ACE_HANDLE)
+{
+  read_messages();
+  return 0;
+}
+
+void LinuxNetworkConfigPublisher::read_messages()
+{
+  char buffer[4096];
```
mitza-oci:
```suggestion
  char buffer[SOME_CONSTANT];
```

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+  for (;;) {
+    ssize_t buffer_length = socket_.recv(buffer, 4096, 0);
+    if (buffer_length <= 0) {
```
mitza-oci: If `buffer_length` is 0, errno shouldn't be checked.

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+      const struct nlmsgerr* msg = reinterpret_cast<struct nlmsgerr*>(NLMSG_DATA(header));
+      ACE_ERROR((LM_ERROR, "LinuxNetworkConfigPublisher::process_message: NETLINK error: %C\n", strerror(-msg->error)));
+    }
+    break;
+  case RTM_NEWADDR:
+    {
+      const struct ifaddrmsg* msg = reinterpret_cast<struct ifaddrmsg*>(NLMSG_DATA(header));
+      size_t address_length = 0;
+      switch(msg->ifa_family) {
+      case AF_INET:
```
mitza-oci: IPv6 support?

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+      const struct nlmsgerr* msg = reinterpret_cast<struct nlmsgerr*>(NLMSG_DATA(header));
+      ACE_ERROR((LM_ERROR, "LinuxNetworkConfigPublisher::process_message: NETLINK error: %C\n", strerror(-msg->error)));
+    }
+    break;
+  case RTM_NEWADDR:
+    {
+      const struct ifaddrmsg* msg = reinterpret_cast<struct ifaddrmsg*>(NLMSG_DATA(header));
+      size_t address_length = 0;
+      switch(msg->ifa_family) {
```
mitza-oci:
```suggestion
      switch (msg->ifa_family) {
```

dds/DCPS/transport/framework/TransportClient.cpp
```diff
@@ -714,6 +730,21 @@ TransportClient::unregister_for_writer(const RepoId& participant,
...
 
+void
+TransportClient::update_locators(const RepoId& remote,
+                                 const TransportLocatorSeq& locators)
+{
+  ACE_GUARD(ACE_Thread_Mutex, guard, lock_);
+  for (ImplsType::iterator pos = impls_.begin(), limit = impls_.end();
+       pos != limit;
+       ++pos) {
+    RcHandle<TransportImpl> impl = (*pos).lock();
```
mitza-oci:
```suggestion
    RcHandle<TransportImpl> impl = pos->lock();
```

dds/DCPS/transport/framework/TransportClient.cpp
```diff
@@ -136,6 +137,8 @@ TransportClient::enable_transport_using_config(bool reliable, bool durable,
   }
   passive_connect_duration_ = TimeDuration::from_msec(duration);
 
+  populate_connection_info();
```
mitza-oci: This is significantly different from what was there before.  It's going to get all possible impl objects, not using `check_transport_qos`.

dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -1148,6 +1148,26 @@ DataLink::ImmediateStart::execute() {
 }
 
 
+void
+DataLink::network_change()
```
mitza-oci:
```suggestion
DataLink::network_change() const
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+                             const ACE_INET_Addr&)
+{
+  struct JoinMulticastGroup : public JobQueue::Job {
+    JoinMulticastGroup(DCPS::RcHandle<RtpsUdpDataLink> tport,
+                       const DCPS::NetworkInterface& nic)
+      : tport_(tport)
+      , nic_(nic)
+    {}
+
+    void execute() override
```
```suggestion
    void execute()
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+                                const ACE_INET_Addr&)
+{
+  struct LeaveMulticastGroup : public DCPS::JobQueue::Job {
+    LeaveMulticastGroup(DCPS::RcHandle<RtpsUdpDataLink> tport,
+                       const DCPS::NetworkInterface& nic)
+      : tport_(tport)
+      , nic_(nic)
+    {}
+
+    void execute() override
```
```suggestion
    void execute()
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -296,6 +407,13 @@ RtpsUdpDataLink::add_locator(const RepoId& remote_id,
 {
   ACE_GUARD(ACE_Thread_Mutex, g, lock_);
   locators_[remote_id] = RemoteInfo(address, requires_inline_qos);
+
+  if (DCPS::DCPS_debug_level > 3) {
+    const DCPS::GuidConverter conv(remote_id);
+    ACE_TCHAR addr_buff[256] = {};
+    address.addr_to_string(addr_buff, 256);
+    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locator %C is now at %C\n"), OPENDDS_STRING(conv).c_str(), addr_buff));
```
```suggestion
    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locator %C is now at %s\n"), OPENDDS_STRING(conv).c_str(), addr_buff));
```

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -255,6 +259,17 @@ RtpsUdpInst::host_addresses() const {
   return addresses;
 }
 
+void
+RtpsUdpInst::update_locators(const RepoId& remote_id,
+                             const TransportLocatorSeq& locators)
+{
+  TransportImpl_rch impl = this->impl();
```
```suggestion
  TransportImpl_rch impl = impl();
```

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -255,6 +259,17 @@ RtpsUdpInst::host_addresses() const {
...
+  if (impl) {
+    DCPS::RtpsUdpTransport_rch rtps_impl = DCPS::static_rchandle_cast<DCPS::RtpsUdpTransport>(impl);
```
```suggestion
    RtpsUdpTransport_rch rtps_impl = static_rchandle_cast<RtpsUdpTransport>(impl);
```

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -115,7 +115,7 @@ namespace {
   {
     if (security_debug.warn) {
       ACE_TCHAR addr_buff[256] = {};
-      remote.addr_to_string(addr_buff, 256, 0);
+      remote.addr_to_string(addr_buff, 256);
```
mitza-oci: Is this change what we want?  I think the concern is that multiple addresses can map to the same DNS name, so this may be discarding useful info.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -311,6 +311,30 @@ RtpsUdpTransport::unregister_for_writer(const RepoId& /*participant*/,
   }
 }
 
+void
+RtpsUdpTransport::update_locators(const RepoId& remote,
+                                  const TransportLocatorSeq& locators)
+{
+  const TransportBLOB* blob = config().get_blob(locators);
+  if (!blob) {
```
mitza-oci:
```suggestion
  if (!blob || is_shut_down_) {
```

dds/DCPS/RTPS/ICE/AgentImpl.cpp
```diff
@@ -87,7 +87,13 @@ int AgentImpl::handle_timeout(const ACE_Time_Value& a_now, const void* /*act*/)
 
 AgentImpl::AgentImpl() :
   ReactorInterceptor(TheServiceParticipant->reactor(), TheServiceParticipant->reactor_owner()),
-  remote_peer_reflexive_counter_(0) {}
+  remote_peer_reflexive_counter_(0)
+{
+  DCPS::NetworkConfigPublisher_rch ncp = TheServiceParticipant->network_config_publisher();
+  if (ncp) {
+    ncp->add_listener(rchandle_from(this));
```
mitza-oci: Does this need a corresponding `remove_listener`?

dds/DCPS/RTPS/ICE/AgentImpl.cpp
```diff
@@ -215,6 +221,21 @@ void AgentImpl::check_invariants() const
...
+  for (EndpointManagerMapType::const_iterator pos = endpoint_managers_.begin(),
+         limit = endpoint_managers_.end(); pos != limit; ++pos) {
+    pos->second->network_change();
+  }
+}
+
+void AgentImpl::remove_address(const DCPS::NetworkInterface& interface,
+                               const ACE_INET_Addr& address)
+{
+  add_address(interface, address);
```
mitza-oci: Why does remove do an add?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1796,27 +1966,44 @@ void Sedp::process_discovered_writer_data(DCPS::MessageId message_id,
...
+      }
+
+      if (locatorsChanged(iter->second.writer_data_.writerProxy, wdata.writerProxy)) {
+        topic_name = get_topic_name(iter->second);
+        OPENDDS_MAP(OPENDDS_STRING, TopicDetails)::const_iterator top_it = topics_.find(topic_name);
+        using DCPS::RepoIdSet;
+        const RepoIdSet& assoc =
+          (top_it == topics_.end()) ? RepoIdSet() : top_it->second.endpoints();
+        for (RepoIdSet::const_iterator i = assoc.begin(); i != assoc.end(); ++i) {
+          if ((i->entityId.entityKind & 4) == 0) continue; // publication
```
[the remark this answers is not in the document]
iguessthislldo: It looks like he would want `isWriter()`?

tests/DCPS/Utilities/.gitignore
```diff
@@ -0,0 +1,4 @@
+/ClockPublisher
+/ClockSubscriber
+/UtilitiesC*
+/UtilitiesTypeSupport*
```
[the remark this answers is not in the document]
jrw972: Where is the script?

tests/DCPS/Utilities/ClockPublisher.cpp
```diff
@@ -0,0 +1,152 @@
+/*
```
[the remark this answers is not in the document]
jrw972: Neither.  I needed a simple publisher/subscriber that I could leave running while doing manual testing.  If such a pair already exists, we should scrap this and provide an index of "useful" programs for developers.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -168,6 +168,90 @@ bool paramsChanged(OpenDDS::DCPS::ContentFilterProperty_t& dest,
   return false;
 }
 
+OpenDDS::DCPS::LocatorSeq transport_locator_to_locator_seq(const OpenDDS::DCPS::TransportLocator& info)
```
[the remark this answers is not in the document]
jrw972: Yes.

tests/DCPS/Utilities/.gitignore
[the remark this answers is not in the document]
iguessthislldo: `tools/scripts/update_gitignore.pl` All it does is put the untracked files in the gitignore though.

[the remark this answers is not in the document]
mitza-oci: @iguessthislldo what else should it do?  Please answer in a separate issue...

tests/DCPS/Utilities/ClockPublisher.cpp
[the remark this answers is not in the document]
mitza-oci: I think that counts as "tools".  Add a README with that explanation and you're good to go.

dds/DCPS/NetworkConfigPublisher.cpp
jrw972: Yes.  But I found that using it is slightly easier if we don't use composite events.  For most cases, listeners won't even care about the interfaces and only care about changes in addresses.

dds/DCPS/LinuxNetworkConfigPublisher.cpp
jrw972: Not in this iteration.

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
jrw972: Generates compiler errors.

dds/DCPS/RTPS/ICE/AgentImpl.cpp
jrw972: We have to poll for changes so the same code is executed in both.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
jrw972: I was testing in networks where DNS was not working.  In this case, any debug message that prints out an IP address blocks waiting for a timeout.  For the exact reason you stated it is more important to know the IP than it is to resolve the hostname.

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
mitza-oci: Well maybe the automatic variable needs a different name?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+    ACE_DEBUG((LM_INFO,
+               ACE_TEXT("(%P|%t) RtpsUdpDataLink::leave_multicast_group ")
+               ACE_TEXT("leaving group %C %C:%hu\n"),
+               nic.name().c_str (),
+               config().multicast_group_address_str_.c_str (),
+               config().multicast_group_address_.get_port_number()));
+  }
+
+  if (0 == multicast_socket_.leave(config().multicast_group_address_, ACE_TEXT_CHAR_TO_TCHAR(nic.name().c_str()))) {
+        joined_interfaces_.erase(nic.name());
```
iguessthislldo:
```suggestion
    joined_interfaces_.erase(nic.name());
```

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
mitza-oci: I must have been confused reading the ACE docs, I think we're agreeing on this one... are there other places that should be changed to make it consistent?

dds/DCPS/RTPS/ICE/AgentImpl.cpp
mitza-oci: Should the method name be changed to reflect this?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+  if (joined_interfaces_.count(nic.name()) == 0 || !nic.addresses.empty()) {
+    return;
+  }
+
+  if (DCPS::DCPS_debug_level > 3) {
+    ACE_DEBUG((LM_INFO,
+               ACE_TEXT("(%P|%t) RtpsUdpDataLink::leave_multicast_group ")
+               ACE_TEXT("leaving group %C %C:%hu\n"),
+               nic.name().c_str (),
+               config().multicast_group_address_str_.c_str (),
```
iguessthislldo:
```suggestion
               config().multicast_group_address_str_.c_str(),
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+{
+  if (joined_interfaces_.count(nic.name()) == 0 || !nic.addresses.empty()) {
+    return;
+  }
+
+  if (DCPS::DCPS_debug_level > 3) {
+    ACE_DEBUG((LM_INFO,
+               ACE_TEXT("(%P|%t) RtpsUdpDataLink::leave_multicast_group ")
+               ACE_TEXT("leaving group %C %C:%hu\n"),
+               nic.name().c_str (),
```
iguessthislldo:
```suggestion
               nic.name().c_str(),
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+
+  if (!config().multicast_interface_.empty() && nic.name() != config().multicast_interface_) {
+    return;
+  }
+
+  if (DCPS::DCPS_debug_level > 3) {
+    ACE_DEBUG((LM_INFO,
+               ACE_TEXT("(%P|%t) RtpsUdpDataLink::join_multicast_group ")
+               ACE_TEXT("joining group %C %C:%hu\n"),
+               nic.name().c_str (),
```
iguessthislldo:
```suggestion
               nic.name().c_str(),
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+               config().multicast_group_address_str_.c_str (),
```
iguessthislldo:
```suggestion
               config().multicast_group_address_str_.c_str(),
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -288,6 +295,110 @@ RtpsUdpDataLink::open(const ACE_SOCK_Dgram& unicast_socket)
...
+  job_queue_->enqueue(DCPS::make_rch<JoinMulticastGroup>(rchandle_from(this), nic));
+}
+
+void
+RtpsUdpDataLink::remove_address(const DCPS::NetworkInterface& nic,
+                                const ACE_INET_Addr&)
+{
+  struct LeaveMulticastGroup : public DCPS::JobQueue::Job {
+    LeaveMulticastGroup(DCPS::RcHandle<RtpsUdpDataLink> tport,
+                       const DCPS::NetworkInterface& nic)
```
iguessthislldo:
```suggestion
                        const DCPS::NetworkInterface& nic)
```

dds/DCPS/transport/framework/TransportDefs.h
```diff
@@ -235,6 +235,11 @@ enum RemoveResult {
 
 typedef ACE_CDR::Long Priority;
 
+typedef size_t ConnectionInfoFlags;
+static const ConnectionInfoFlags CONNINFO_UNICAST = (1 << 0);
+static const ConnectionInfoFlags CONNINFO_MULTICAST = (1 << 0);
```
iguessthislldo: Are these both supposed to be 1?

dds/DCPS/security/SSL/SignedDocument.cpp
```diff
@@ -292,8 +292,8 @@ PKCS7* SignedDocument::PKCS7_from_SMIME_file(const std::string& path)
   if (!in) {
     ACE_ERROR((LM_ERROR,
```
iguessthislldo:
```suggestion
    ACE_ERROR((LM_WARNING,
```

dds/DCPS/security/SSL/Certificate.cpp
```diff
@@ -415,8 +415,8 @@ void Certificate::load_cert_bytes(const std::string& path)
   if (!in) {
     ACE_ERROR((LM_ERROR,
```
iguessthislldo:
```suggestion
    ACE_ERROR((LM_WARNING,
```

dds/DCPS/Service_Participant.cpp
```diff
@@ -2021,6 +2028,23 @@ void Service_Participant::default_configuration_file(const ACE_TCHAR* path)
   default_configuration_file_ = path;
 }
 
+NetworkConfigPublisher_rch Service_Participant::network_config_publisher()
+{
+  if (!network_config_publisher_) {
+#ifdef ACE_LINUX
+    network_config_publisher_ = make_rch<LinuxNetworkConfigPublisher>(reactor_task_.interceptor());
+#endif
+  }
```
iguessthislldo: I don't think it would matter to the compiler, but I think it would look better if you put the ifdef around the the if block.

dds/DCPS/Service_Participant.cpp
```diff
@@ -2021,6 +2028,23 @@ void Service_Participant::default_configuration_file(const ACE_TCHAR* path)
...
+
+  if (network_config_publisher_ && !network_config_publisher_->open()) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("ERROR: Service_Participant::get_domain_participant_factory could not open network config publisher\n ")));
```
iguessthislldo:
```suggestion
    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Service_Participant::get_domain_participant_factory could not open network config publisher\n")));
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4433,7 +4607,7 @@ Sedp::handle_datareader_crypto_tokens(const DDS::Security::ParticipantVolatileMe
     ACE_DEBUG((LM_DEBUG,
       ACE_TEXT("(%P|%t) Sedp::handle_datareader_crypto_tokens() - ")
       ACE_TEXT("received tokens for unknown remote reader: (%C). Caching.\n"),
-      OPENDDS_STRING(DCPS::GuidConverter(msg.source_endpoint_guid)).c_str()));
+               DCPS::LogGuid(msg.source_endpoint_guid).c_str()));
```
iguessthislldo:
```suggestion
      DCPS::LogGuid(msg.source_endpoint_guid).c_str()));
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2141,14 +2300,30 @@ void Sedp::process_discovered_reader_data(DCPS::MessageId message_id,
...
+
+      if (checkAndAssignLocators(iter->second.reader_data_.readerProxy, rdata.readerProxy)) {
+        topic_name = get_topic_name(iter->second);
+        OPENDDS_MAP(OPENDDS_STRING, TopicDetails)::const_iterator top_it =
+          topics_.find(topic_name);
+        using DCPS::RepoIdSet;
+        const RepoIdSet& assoc =
+          (top_it == topics_.end()) ? RepoIdSet() : top_it->second.endpoints();
+        for (RepoIdSet::const_iterator i = assoc.begin(); i != assoc.end(); ++i) {
+          if (i->entityId.entityKind & 4) continue; // subscription
```
iguessthislldo:
```suggestion
          if (DCPS::GuidConverter(*i).isReader()) continue; // subscription
```

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+          break;
+        }
+      }
+
+    }
+    break;
+  case RTM_NEWLINK:
+    {
+      OPENDDS_STRING name;
+      const struct ifinfomsg* msg = reinterpret_cast<struct ifinfomsg*>(NLMSG_DATA(header));
```
iguessthislldo: Adam has already mentioned this, but `struct` keyword isn't needed in C++.

dds/DCPS/LinuxNetworkConfigPublisher.cpp
```diff
@@ -0,0 +1,243 @@
...
+      int rta_length = IFLA_PAYLOAD(header);
+      for (const struct rtattr* attr = reinterpret_cast<const struct rtattr*>(IFLA_RTA(msg));
+           RTA_OK(attr, rta_length);
+           attr = RTA_NEXT(attr, rta_length)) {
+        switch (attr->rta_type) {
```
iguessthislldo:
```suggestion
        if (attr->rta_type == IFLA_IFNAME) {
```

Unless maybe you're planning on supporting more types.

tests/DCPS/Utilities/.gitignore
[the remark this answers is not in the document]
iguessthislldo: I didn't really mean that it should necessarily be doing something else, I was just giving the gist of what the script does.

dds/DCPS/LinuxNetworkConfigPublisher.cpp
jrw972: I'll give it a try.

dds/DCPS/transport/framework/TransportClient.cpp
jrw972: I added a check to `check_transport_qos` in `populate_connection_info`.

dds/DCPS/transport/framework/TransportDefs.h
jrw972: Great catch.

dds/DCPS/Service_Participant.cpp
jrw972: Eventually there will be blocks for other platforms.  The `if` still needs to wrap all of it.

jrw972:
Please squash before merging.

mitza-oci:
Azure CI Linux Security:
```
dds/DCPS/LinuxNetworkConfigMonitor.cpp:235:1: error: expected ‘}’ at end of input
```

mitza-oci:
> 
> 
> Azure CI Linux Security:
> 
> ```
> dds/DCPS/LinuxNetworkConfigMonitor.cpp:235:1: error: expected ‘}’ at end of input
> ```

It is missing `OPENDDS_END_VERSIONED_NAMESPACE_DECL`

PR #1366, opened 2019-11-08
Rename dds_fuzz.pl to lint.pl and Have it do Generic Text File Checks

iguessthislldo:
- Picking up changes to `dds_fuzz.pl` where I'm leaving it in #1347 because it's out of scope there and requires more work.
- Rewrote to have a more declarative and generic structure.
- Take over generic text file checks from `ace fuzz.pl`. Instead of just checking a file based on its file name/extension, if perl considers it to be a text file, do the basic text file checks on it unless it matches a blacklisted file name pattern.
- Do other checks like empty files, redundant newlines, and binary files.
- Renamed to `lint.pl`


jwillemsen:
Maybe our fuzzr can also help, see https://github.com/RemedyIT/taox11/blob/master/.github/workflows/fuzzr.yml for a github action that uses it in combination with the fuzzers that we have in axcioma, see https://github.com/RemedyIT/axcioma/tree/master/bin/fuzzers.

jwillemsen:
PR https://github.com/objectcomputing/OpenDDS/pull/1568 changes the usage of the fuzzer script from travis to github actions

iguessthislldo:
> Maybe our fuzzr can also help, see https://github.com/RemedyIT/taox11/blob/master/.github/workflows/fuzzr.yml for a github action that uses it in combination with the fuzzers that we have in axcioma, see https://github.com/RemedyIT/axcioma/tree/master/bin/fuzzers.

It looks interesting and looks like you guys put a lot of thought into it, but I don't know Ruby at all, so it's of limited use to me. I do like that it registers checks and how modulal it all looks. I don't really know how to set up Perl modules, but I might look into that to split `dds_fuzz.pl` out into a `opendds_linting` module along similar lines.

> PR #1568 changes the usage of the fuzzer script from travis to github actions

As long as `dds_fuzz.pl` isn't changed, it won't impact these changes.

mitza-oci:
`Text file has redundent newlines at the end`
Is this getting too "picky"?  (If we keep it, redundant is spelled with an a at the end.  Or we could call them superfluous.

iguessthislldo:
> Is this getting too "picky"?

Probably. I'm fine with axing it.

iguessthislldo:
Here are the default checks I have so far.
```
  - trailing_whitespace
    - Text file has trailing whitespace at the end of lines
  - final_newline
    - Text file must end with one and only one newline
  - gettimeofday
    - ACE_OS::gettimeofday() and "ACE_Time_Value().now()" are forbidden in the core libraries.
      See the "Time" section in docs/guidelines.md.
  - is_binary
    - File looks like a binary file
  - is_empty
    - File is empty
  - tabs
    - Text file has tabs
```

These are two that will make default in a future PR, but will require more work to fix in OpenDDS that can't be easily fixed en masse like the whitespace checks. Also to make them practical they will probably require magic lines like fuzz.pl has to disable them for exceptions.

```
  - nonprefixed_public_macros
    - Public macros must be prefixed with OPENDDS or ACE
  - line_length
    - Text file either has extremely long lines or has invalid UTF-8
```

This is the non-default check that I don't think I will ever make default because it's really slow.
```
  - utf8
    - Text file has invalid UTF-8
```

mitza-oci (approved):

PR #1368, opened 2019-11-11
ICE Agent Info exchanged in the clear when security is enabled

jrw972:
When security is enabled for a participant, only include the ICE agent
info in the secure SPDP messages.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -1257,10 +1257,8 @@ Spdp::attempt_authentication(const DCPS::RepoId& guid, DiscoveredParticipant& dp
 
 void Spdp::update_agent_info(const DCPS::RepoId&, const ICE::AgentInfo&)
 {
-#ifdef OPENDDS_SECURITY
   if (is_security_enabled())
     write_secure_updates();
-#endif
 }
 #endif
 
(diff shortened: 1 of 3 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1241,6 +1254,14 @@ Spdp::attempt_authentication(const DCPS::RepoId& guid, DiscoveredParticipant& dp
 
   return;
 }
+
+void Spdp::update_agent_info(const DCPS::RepoId&, const ICE::AgentInfo&)
+{
+#ifdef OPENDDS_SECURITY
```
```suggestion
```
could be redundant

dds/DCPS/RTPS/Spdp.h
```diff
@@ -52,7 +52,7 @@ class RtpsDiscovery;
 
 /// Each instance of class Spdp represents the implementation of the RTPS
 /// Simple Participant Discovery Protocol for a single local DomainParticipant.
-class OpenDDS_Rtps_Export Spdp : public DCPS::LocalParticipant<Sedp> {
+class OpenDDS_Rtps_Export Spdp : public DCPS::LocalParticipant<Sedp>, public ICE::AgentInfoListener {
```
mitza-oci: needs to be conditional on OPENDDS_SECURITY

PR #1379, opened 2019-11-18
Multitopic with the C++11 Mapping

iguessthislldo:
- Generate correct metastruct string code for C++11.
- MultiTopic Test:
  - Convert to be able to use either classic or C++11 mapping
  - Add various types to test the C++11 mapping
  - Try to improve error reporting
- Try to fix multitopic error messages


dds/idl/metaclass_generator.cpp
```diff
@@ -117,6 +117,13 @@ namespace {
     }
   }
 
+  std::string string_type(Classification cls)
+  {
+    return be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11 ?
+      ((cls & CL_WIDE) ? "std::wstring" : "std::string") :
```
mitza-oci: I'm not all that happy about the Codacy (cppcheck) changes.... but OK.  I wonder if we can override some of these?  Operator precedence rules exist for a reason and compilers are pretty good about warning if we get it wrong.

dds/DCPS/MultiTopicDataReaderBase.cpp
```diff
@@ -50,10 +50,8 @@ namespace {
       try {
         outer_->data_available(reader);
       } catch (std::exception& e) {
-        if (OpenDDS::DCPS::DCPS_debug_level) {
-          ACE_ERROR((LM_ERROR, "(%P|%t) MultiTopicDataReaderBase::Listener::"
-                     "on_data_available(): %C", e.what()));
-        }
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: MultiTopicDataReaderBase::Listener::on_data_available: %C\n"),
```
mitza-oci: This changed debug level -- was that intended?

dds/DCPS/MultiTopicDataReaderBase.cpp
```diff
@@ -274,29 +264,23 @@ void MultiTopicDataReaderBase::data_available(DDS::DataReader_ptr reader)
       if (gen.info_[i].valid_data) {
         incoming_sample(gen.samples_[i], gen.info_[i], topic.c_str(), meta);
       } else if (gen.info_[i].instance_state != ALIVE_INSTANCE_STATE) {
-        DataReaderImpl* resulting_impl =
-          dynamic_cast<DataReaderImpl*>(resulting_reader_.in());
-
+        DataReaderImpl* resulting_impl = dynamic_cast<DataReaderImpl*>(resulting_reader_.in());
```
mitza-oci: I'd like to avoid having these kinds of changes in the revision history.  It's not actually changing anything, and it takes some effort on the part of the reader to figure that out.

dds/idl/metaclass_generator.cpp
iguessthislldo: Yeah, I thought this was overzealous as well. There's no way a compiler would mess up there.

tests/DCPS/MultiTopic/classic/MultiTopicTest_Classic.mpc
```diff
@@ -0,0 +1,16 @@
...
+  requires += multi_topic
+  idlflags += -I..
+  TypeSupport_Files {
+    ../MultiTopicTest.idl
+  }
+  IDL_Files {
+    ../MultiTopicTest.idl
+    MultiTopicTestTypeSupport.idl
+  }
+  includes += .
```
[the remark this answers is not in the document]
iguessthislldo: Yes, for the generated headers.

dds/idl/langmap_generator.cpp
```diff
@@ -1597,7 +1597,11 @@ struct Cxx11Generator : GeneratorBase
     const std::string lang_field_type = generator_->map_type(type);
     const Classification cls = classify(actual_field_type);
     if (!(cls & (CL_PRIMITIVE | CL_ENUM))) {
-      return "    new(&_" + name + ") " + lang_field_type + ";\n";
+      std::stringstream ss;
```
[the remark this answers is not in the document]
iguessthislldo: I guess it doesn't really... is there a reason not to use it if I was writing this from scratch?

dds/DCPS/MultiTopicDataReaderBase.cpp
iguessthislldo: For this particular one I'm not sure. I know the other ones were hiding errors in the SQL and I didn't think that was right.

dds/idl/langmap_generator.cpp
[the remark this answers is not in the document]
mitza-oci: I/O Streams are ugly, slow, and awkward to use.  They have one important feature though, they can easily convert numbers to strings (in all C++ std versions).  In `opendds_idl` I/O Streams tend to be only created when needed.  If just using `std::string` is sufficient, go with that.

tests/DCPS/Compiler/C++11/idl_test1_main/main.cpp
```diff
@@ -42,6 +42,12 @@ bool idl2cxx11_test()
   u = std::move(u2);
   swap(u3, u);
 
+  {
+    // Regression Test for Invalid Free
```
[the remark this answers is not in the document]
iguessthislldo: Gotcha, will remove it.

dds/DCPS/MultiTopicDataReaderBase.cpp
iguessthislldo: Alright, I will try to keep that in mind.

dds/idl/langmap_generator.cpp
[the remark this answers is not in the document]
iguessthislldo: Alright. I was under the impression that stringstream would be at least somewhat faster than string concatenation.

tests/DCPS/Compiler/C++11/idl_test1_main/main.cpp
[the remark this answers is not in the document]
mitza-oci: Since this was removed, is the change in `langmap_generator.cpp` still needed?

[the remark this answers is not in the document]
iguessthislldo: I don't see a reason not to include it, it's just a small safety feature.

mitza-oci:
- [ ] Azure CI VS2017 & 2019 compile errors

dds/DCPS/MultiTopicDataReaderBase.cpp
```diff
@@ -259,8 +250,7 @@ void MultiTopicDataReaderBase::data_available(DDS::DataReader_ptr reader)
       " could not be cast to DataReaderImpl.");
   }
   DataReaderImpl::GenericBundle gen;
-  ReturnCode_t rc = dri->read_generic(gen, NOT_READ_SAMPLE_STATE,
-                                      ANY_VIEW_STATE, ANY_INSTANCE_STATE,false);
+  ReturnCode_t rc = dri->read_generic(gen, NOT_READ_SAMPLE_STATE, ANY_VIEW_STATE, ANY_INSTANCE_STATE, false);
```
mitza-oci:
```suggestion
  const ReturnCode_t rc = dri->read_generic(gen, NOT_READ_SAMPLE_STATE, ANY_VIEW_STATE, ANY_INSTANCE_STATE, false);
```

dds/DCPS/MultiTopicDataReader_T.cpp
```diff
@@ -108,11 +105,11 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::join(
       ReturnCode_t ret = other_dri->read_instance_generic(other_data.ptr_,
         info, ih, READ_SAMPLE_STATE, ANY_VIEW_STATE, ALIVE_INSTANCE_STATE);
       if (ret != RETCODE_OK && ret != RETCODE_NO_DATA) {
-        std::ostringstream rc_ss;
-        rc_ss << ret;
-        throw std::runtime_error("In join(), incoming DataReader for " +
-          OPENDDS_STRING(other_topic) + " read_instance_generic, error #" +
-          rc_ss.str());
+        std::ostringstream ss;
```
mitza-oci: Now that int-to-string isn't used, this doesn't need ostringstream (also below).

tests/DCPS/MultiTopic/MultiTopicTest.idl
```diff
@@ -1,49 +1,75 @@
...
-#pragma DCPS_DATA_KEY "PlanInfo flight_id1"
-#pragma DCPS_DATA_KEY "PlanInfo flight_id2"
+enum FlightType {
+  visual,
+  instrument,
+  mixed
+};
+
+typedef string<3> Airport; // 3 Letter IATA Code
+typedef sequence<Airport> Airports;
```
mitza-oci: https://opendds.org/faq.html#idl-builtin
This is why the MSVC builds are failing on Azure CI.

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -7,25 +7,305 @@
...
+  return std::strcmp(a.in(), b.in());
+}
+
+bool sequences_are_not_equal(const Airports& a, const Airports& b)
+{
+  const unsigned len = a.length();
+  if (len != b.length()) {
+    return true;
+  }
+  for (unsigned i = 0; i < len; i++) {
```
mitza-oci:
```suggestion
  for (unsigned i = 0; i < len; ++i) {
```

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -7,25 +7,305 @@
...
+    os << "Invalid(" << value._d() << ")";
+    break;
+  }
+  return os;
+}
+
+template <typename T>
+bool arrays_are_not_equal(const T* a, const T* b, size_t length)
+{
+  for (size_t i = 0; i < length; i++) {
```
mitza-oci:
```suggestion
  for (size_t i = 0; i < length; ++i) {
```

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -7,25 +7,305 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+template <typename T>
+void print_array(std::ostream& os, const T* array, size_t length)
+{
+  for (size_t i = 0; i < length; i++) {
```
mitza-oci:
```suggestion
  for (size_t i = 0; i < length; ++i) {
```

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -7,25 +7,305 @@
...
+    if (i) {
+      os << ", ";
+    }
+    os << array[i];
+  }
+}
+
+template <typename T>
+void print_sequence(std::ostream& os, const T& seq) {
+  for (size_t i = 0; i < get_sequence_length(seq); i++) {
```
mitza-oci:
```suggestion
  for (unsigned int i = 0; i < get_sequence_length(seq); ++i) {
```
avoids warnings on some compilers due to narrowing from 64-bit (unsigned) size_t to 32-bit unsigned int used by sequence `operator[]`

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -7,25 +7,305 @@
...
 const int N_ITERATIONS = 5;
+const Duration_t max_wait = {10, 0};
 
-void check(ReturnCode_t ret)
+void check_rc(ReturnCode_t ret, const char* message, const char* extra_message = 0)
 {
   if (ret != RETCODE_OK) {
     std::ostringstream oss;
-    oss << "DDS API call failed with return code: " << ret;
+    oss << "Failed to " << message;
```
mitza-oci: ostringstream not needed here

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -116,131 +387,211 @@ bool run_multitopic_test(const Publisher_var& pub, const Subscriber_var& sub)
...
-    if (ret != RETCODE_OK) return false;
-    mi.flight_id1 = 100;
-    mi.more = "Extra info for all flights with id1 == 100";
-    ret = midw->write(mi, HANDLE_NIL);
-    if (ret != RETCODE_OK) return false;
 
-    // Write samples (Unrelated)
+    mi.departure_date() = 299;
+    mi.more() = "Shouldn't see this";
+    for (unsigned i = 0; i < even_more_length; i++) {
```
mitza-oci:
```suggestion
    for (unsigned i = 0; i < even_more_length; ++i) {
```

tests/DCPS/MultiTopic/MultiTopicTest.cpp
```diff
@@ -258,28 +609,22 @@ bool run_multitopic_test(const Publisher_var& pub, const Subscriber_var& sub)
...
   ConditionSeq active;
-  ret = ws->wait(active, infinite);
-  if (ret != RETCODE_OK) return false;
+  check_rc(ws->wait(active, max_wait), "wait for dispose");
   ws->detach_condition(rc);
   ResultingDataReader_var res_dr = ResultingDataReader::_narrow(dr);
   ResultingSeq data;
   SampleInfoSeq info;
-  ret = res_dr->read_w_condition(data, info, LENGTH_UNLIMITED, rc);
+  check_rc(res_dr->read_w_condition(data, info, LENGTH_UNLIMITED, rc), "");
```
mitza-oci: Looks like this needs a description string

tests/DCPS/MultiTopic/run_test.pl
```diff
@@ -10,16 +10,22 @@
...
+    $is_rtps_disc = 1;
+  }
+  if ($arg eq 'cpp11' || $arg eq 'classic') {
+    $dir = $arg;
+  }
 }
 
 my $DCPSREPO;
-my $args;
+my $args = '';
```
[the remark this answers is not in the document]
iguessthislldo: Yes, I was putting in logging arguments.

tests/DCPS/MultiTopic/MultiTopicTest.idl
iguessthislldo: Alright, but how does this work for a bounded string?

tests/DCPS/MultiTopic/MultiTopicTest.cpp
iguessthislldo: What is the benefit for doing this for integers?

tests/DCPS/MultiTopic/MultiTopicTest.cpp
iguessthislldo: Yes, can't imagine why I left this blank.

tests/DCPS/MultiTopic/MultiTopicTest.idl
mitza-oci: I'm not sure, we might have to do so more to get it working.  But I'd start with `#include`ing TAO's IDL file since we know that solves some form of this problem.

tests/DCPS/MultiTopic/MultiTopicTest.cpp
mitza-oci: It is in the guidelines.  I believe this is common with ACE as well.  Rationale is that if the developer is in the habit of always writing it prefix (when the expression result is unused), it will be right without having to think about it.  It is also a case of the general principle of using the least complex operation that meets the requirements.

tests/DCPS/MultiTopic/MultiTopicTest.idl
iguessthislldo: Huh, I don't think it should've worked if it's a bounded string, but I guess I can't really complain if it worked.

iguessthislldo:
The issue on the PR jenkins doesn't make any sense:

```
make[1]: *** No rule to make target '../MultiTopicTestC.h', needed by 'idl_stubs'.
```

I'm guessing it has to do with the idl being in the parent directory, but I don't know where it's actually getting it from. I tried looking at what `idl_stubs` is depending on on my machine and it looks normal to me. Is this some problem with parallel jobs?

dds/idl/langmap_generator.cpp
```diff
@@ -1597,7 +1597,8 @@ struct Cxx11Generator : GeneratorBase
     const std::string lang_field_type = generator_->map_type(type);
     const Classification cls = classify(actual_field_type);
     if (!(cls & (CL_PRIMITIVE | CL_ENUM))) {
-      return "    new(&_" + name + ") " + lang_field_type + ";\n";
+      return "    new(&_" + name + ") " + lang_field_type + ";\n"
```
mitza-oci: The change in this file was related to a change in a test which has since been reverted.  Should this also be reverted?  I don't see a need for it, but let me know if I'm missing something.

iguessthislldo: I would prefer it stay if we're not throwing an error for it yet. A user using OpenDDS incorrectly is still using OpenDDS and think we should be able to handle situations like this if it's not a burden on us.

mitza-oci: I agree with the overall goal.  I'm just not sure which kinds of misuse this protects against.

The version on master doesn't use `_set` for much at all.  It could probably do without it.

To protect against some misuse, instead of using `_disc` (along with `_set`) to determine the active member of the union, we could add a second data member of the same type.  This wouldn't be changed by `_d()`.

iguessthislldo: So maybe something like this:

```C++
void MiscUnion::_activate(int16_t d)
{
  if (d != _ctor_disc) {
    _reset();
  }
  switch (d) {
  case 0: {
    break;
  }
  case 1:
  case 2: {
    new(&_string_value) std::string;
    break;
  }
  case 3: {
    break;
  }
  default:
    break;
  }
  _disc = d;
  _ctor_disc = d;
}

void MiscUnion::_reset()
{
  switch (_ctor_disc) {
  case 0: {
    break;
  }
  case 1:
  case 2: {
    _string_value.~basic_string();
    break;
  }
  case 3: {
    break;
  }
  default:
    break;
  }
}
```

iguessthislldo: So maybe something like this?

```C++
void MiscUnion::_activate(int16_t d, bool init = false)
// Set init to true in constructor so reset won't act based on garbage.
{
  if (!init && d != _constructor_disc) {
    _reset();
  }
  _disc = d;
  _constructor_disc = d;
  switch (_constructor_disc) {
  case 0: {
    break;
  }
  case 1:
  case 2: {
    new(&_string_value) std::string;
    break;
  }
  case 3: {
    break;
  }
  default:
    break;
  }
}

void MiscUnion::_reset()
{
  switch (_constructor_disc) {
  case 0: {
    break;
  }
  case 1:
  case 2: {
    _string_value.~basic_string();
    break;
  }
  case 3: {
    break;
  }
  default:
    break;
  }
}
```

mitza-oci: Yes, this is pretty close to what I was thinking.  In looking at it again, maybe we name it `_active` or something similar.  Every switch generated (operator=, etc.) would use this.  It doesn't provide error handling in all cases (that could come later, depending on how we want to do it).

As long as were here, we could attempt to deal with potential name conflicts between user-provided names and generated names.  (While also not violating C++ name requirements.)

mitza-oci: Let's merge this one without the langmap_generator.cpp changes.  They are not related to multitopic / content subscription.

Issue #1252, opened 2019-08-20
Duplicate messages received by Subscriber


nikhilbagul:
Hello all,

While trying to work with the Open DDS example of Messenger.minimal at
DevGuideExamples/DCPS/Messenger.minimal
I was able to set up the Publisher and Subscriber successfully on a Mac with High Sierra. Everything works as expected when I am publishing 'n' number of messages from Publisher. I am receiving exactly 'n' number of messages on the Subscriber.
But, whenever I try to modify the Publisher to send messages on ad-hoc basis/waiting for user input to publish a message, the Subscriber receives duplicate messages i.e > 1 of the same message for the one Published message.
I am trying to get my head around the message QoS policies but haven't had any luck so far.

I am using RTPS as my communication service and OpenDDS version 3.13.
Any pointers are much appreciated.

Please find below the link to the repository that I have created as you mentioned. Please let know if you need any other info from my end.

[Github repo link](https://github.com/nikhilbagul/OpenDDS-PubSub-QtPrototype)

Best,
Nikhil

mitza-oci:
Please set the data reader's Reliability QoS to RELIABLE.

mitza-oci:
Is this problem fixed with the latest changes on master?

PR #1381, opened 2019-11-19
RTPS 2.4 directed write changes

lij-oci:
directed write changes


tests/transport/rtps_direct_write/Domain.cpp
```diff
@@ -45,11 +45,11 @@ Domain::Domain(int argc, ACE_TCHAR* argv[])
   } catch (const CORBA::BAD_PARAM& e) {
     e._tao_print_exception("ERROR: CORBA::BAD_PARAM");
     cleanup();
-    throw e;
+    throw;
   } catch (const OpenDDS::DCPS::Transport::NotConfigured& e) {
     ACE_ERROR((LM_ERROR, "ERROR: OpenDDS::DCPS::Transport::NotConfigured.\n"));
     cleanup();
-    throw e;
+    throw;
```
iguessthislldo: I didn't know you could rethrow an exception like that.

simpsont-oci:
Yeah, it can be required when you catch using an ellipsis (...) but don't
want to prevent the exception from escalating or don't want to
inadvertently change from a base class to an (unknown to current context)
derived exception type.

tests/transport/rtps_direct_write/Domain.cpp
simpsont-oci: Yeah, it can be required when you catch using an ellipsis (...) but don't
want to prevent the exception from escalating or don't want to
inadvertently change from a base class to an (unknown to current context)
derived exception type.

mitza-oci: You can *only* rethrow like that

lij-oci:
DirectedWrite still does not work when the subscriber is set to Best Effort.
[subscriber_BestEffort.log](https://github.com/objectcomputing/OpenDDS/files/3872025/subscriber_BestEffort.log)
[subscriber_Reliable.log](https://github.com/objectcomputing/OpenDDS/files/3872026/subscriber_Reliable.log)

lij-oci:
1. If the subscriber is not set to reliable, the current code does the following:

1) RtpsUdpDataLink::associated() will return without calling readers_of_writer_.insert().

2) RtpsUdpDataLink::datareader_dispatch() will not call, if readers_of_writer_ is empty, 
   RtpsUdpDataLink::RtpsReader::process_data_i(), which contains the logic to set 
   RtpsUdpDataLink receive_strategy - whether or not to withhold data, 
   including the newly added DirectedWrite matching logic.

3) RtpsUdpReceiveStrategy::deliver_sample_i() calls 
   RtpsUdpDataLink::received(), which in turn, calls 
   RtpsUdpDataLink::datareader_dispatch(),
   to set readers_selected_ and readers_withheld_. 

   Based on 1) and 2) above, readers_selected_ and readers_withheld_ will be empty
   when RtpsUdpDataLink::received() returns. Therefore, no data is withheld for unmatched readers.
 
2. This problem is not only for DirectedWrite. As long as process_data_i() is skipped, 
its strategies to withhold data will not be applied, such as duplicate seq#, disjoint,
and various combinations of empty, cumulative_ack, durable_ and seq. 

I did an experiment: sent the same seq #3 twice, and the subscriber got both, 
along with the unmatched DirectedWrite. See s3_subscriber.log.
[s3_subscriber.log](https://github.com/objectcomputing/OpenDDS/files/3888965/s3_subscriber.log)

Then I commented out the return at the beginning of RtpsUdpDataLink::associated():
  if (!local_reliable) {
    // return;
  }
Run the same test again. No unmatched DirectedWrite messages nor messages with the same seq# were received. See s3m_subscriber.log.
[s3m_subscriber.log](https://github.com/objectcomputing/OpenDDS/files/3888966/s3m_subscriber.log)

This seems to suggest that trying to move the DirectedWrite checking logic to another place may not be a good solution.

3. Questions:
3.1 What are the reasons for RtpsUdpDataLink::associated() to return if(!local_reliable)? I found no comments in the code.
3.2 Since writer is default to reliable and reader to best_effort, could it not return here at least for readers?

4. If this is not a DirectedWrite specific problem, should we close the DirectedWrite PR and start another PR to handle the strategies of withholding data when the subscriber is not set to reliable?

lij-oci:
Is this related to Issue  #1252 Duplicate messages received by Subscriber?

bin/dcps_tests.lst
```diff
@@ -332,6 +332,7 @@ tests/DCPS/SubscriberCycle/run_test.pl: !DCPS_MIN !DDS_NO_CONTENT_FILTERED_TOPIC
 tests/transport/rtps/run_test.pl: !DCPS_MIN RTPS
 tests/transport/rtps_reliability/run_test.pl: !DCPS_MIN RTPS
 tests/transport/spdp/run_test.pl: !DCPS_MIN RTPS
+tests/transport/rtps_direct_write/run_test.pl: !DCPS_MIN RTPS
```
mitza-oci:
```suggestion
tests/transport/rtps_directed_write/run_test.pl: !DCPS_MIN RTPS
```
Let's make it "directed write" for consistency with the spec.  That way someone can search the repository and find this test.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1124,6 +1124,25 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
     WriterInfo& info = wi->second;
     SequenceNumber seq;
     seq.setValue(data.writerSN.high, data.writerSN.low);
+
+    // determines if this is a direct write to this reader
```
mitza-oci: As we discussed, this will be cleaner to integrate in RtpsUdpReceiveStrategy.  RtpsUdpDataLink has additional functions that are only used for reliable readers.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1124,6 +1124,25 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
...
+    bool directWrite = false;
+    bool directWriteNoMatch = true;
```
```suggestion
    bool directWriteMatch = false;
```
In general, avoid using a negatively-named boolean variable since it's harder for a reader to follow.

tests/transport/rtps_direct_write/Domain.cpp
```diff
@@ -0,0 +1,94 @@
...
+      case 'h':
+        host = opts.opt_arg();
+        break;
+      case 'p':
+        port = static_cast<u_short>(ACE_OS::atoi(opts.opt_arg()));
+        break;
+      }
+    }
+#ifdef OPENDDS_SAFETY_PROFILE
+    if (host == ACE_TEXT("localhost"))
```
mitza-oci:
```suggestion
    if (host == ACE_TEXT("localhost")) {
```

tests/transport/rtps_direct_write/Domain.cpp
```diff
@@ -0,0 +1,94 @@
...
+    subRdrId[1] = createID(0x76543210, 0x222222, OpenDDS::DCPS::ENTITYKIND_USER_READER_WITH_KEY);
+  } catch (const OpenDDS::DCPS::Transport::Exception& e) {
+    ACE_ERROR((LM_ERROR, "EXCEPTION: %C\n", typeid(e).name()));
+    cleanup();
+    throw;
+  } catch (const CORBA::Exception& e) {
+    ACE_ERROR((LM_ERROR, "EXCEPTION: %C\n", e._info().c_str()));
+    cleanup();
+    throw;
+  } catch(...) {
```
mitza-oci:
```suggestion
  } catch (...) {
```

tests/transport/rtps_direct_write/Domain.cpp
```diff
@@ -0,0 +1,94 @@
...
+  OpenDDS::DCPS::RepoIdBuilder idBd;
+  idBd.federationId(0x01234567); // guidPrefix1
+  idBd.participantId(participantId); // guidPrefix2
+  idBd.entityKey(key);
+  idBd.entityKind(kind);
+  return idBd;
+}
+
+void Domain::cleanup() {
+  if(dpf != nullptr) {
```
mitza-oci:
```suggestion
  if (dpf) {
```

tests/transport/rtps_direct_write/Domain.cpp
```diff
@@ -0,0 +1,94 @@
...
+    TheServiceParticipant->shutdown();
+    dpf = nullptr;
```
mitza-oci:
```suggestion
    dpf = 0;
```

tests/transport/rtps_direct_write/TestMsg.h
```diff
@@ -0,0 +1,72 @@
...
+}
+
+bool operator>>(Serializer& strm, TestMsg& stru)
+{
+  return (strm >> stru.key) && (strm >> stru.value.out());
+}
+
+size_t gen_max_marshaled_size(const TestMsg& stru, bool align)
+{
+  ACE_UNUSED_ARG(stru);
```
mitza-oci: This macro is used for generated code and when different lines are skipped/included by the preprocessor.  It's not needed here.

tests/transport/rtps_direct_write/publisher.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+#include <iostream>
+#include <sstream>
+#include <cstring>
+#include <ctime>
+
+class SimpleDataWriter : public TransportSendListener, public TransportClient
+{
+public:
+  explicit SimpleDataWriter(const Domain& d, const LocatorSeq& locators)
```
mitza-oci:
```suggestion
  SimpleDataWriter(const Domain& d, const LocatorSeq& locators)
```

tests/transport/rtps_direct_write/subscriber.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <iostream>
+#include <sstream>
+
+class SimpleDataReader : public TransportReceiveListener, public TransportClient
+{
+public:
+  explicit SimpleDataReader(const Domain &d, const int readerIndex)
+    : domain(d), index(readerIndex), done_(false), seq_()
+  {
+    //enable_transport(false, false); //(reliable, durable)
```
mitza-oci:
```suggestion
```

tests/transport/rtps_direct_write/subscriber.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+class DDS_TEST {
+public:
+  DDS_TEST(int argc, ACE_TCHAR* argv[]) : domain(argc, argv) {
+    domain.setHostAddress();
+  }
+
+  int run(){
```
mitza-oci:
```suggestion
  int run() {
```

mitza-oci:
We can handle the best-effort reader in a separate PR.  If there is a separate bug, that's the right way to do it.  Start by adding a test (DCPS level, not transport), which could be on this branch.

tests/transport/rtps_direct_write/publisher.cpp
```diff
@@ -0,0 +1,305 @@
...
+  const Header hdr = { {'R', 'T', 'P', 'S'}, PROTOCOLVERSION, VENDORID_OPENDDS,
+    {local_prefix[0], local_prefix[1], local_prefix[2], local_prefix[3],
+     local_prefix[4], local_prefix[5], local_prefix[6], local_prefix[7],
+     local_prefix[8], local_prefix[9], local_prefix[10], local_prefix[11]} };
+
+  const ACE_Time_Value now = ACE_OS::gettimeofday();
+  log_time(now);
+  const double conv = 4294.967296; // NTP fractional (2^-32) sec per microsec
+  const InfoTimestampSubmessage it = { {INFO_TS, 1, 8},
+    {static_cast<ACE_CDR::Long>(now.sec()),
```
mitza-oci: This line was an error on Azure CI, probably due to the Time_t changes from another branch.  This may be a case where explicitly merging master into this branch is the best way forward.

mitza-oci: There are a few related errors on the Azure CI Linux/clang builds (even with the latest revision 4829273)

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: I forgot to write it here yesterday, but this was attempting to be an answer to your point #3 from the large post above.  The purpose of RtpsUdpDataLink's internal data structures like RtpsWriter and ReaderInfo is for reliability (see comments in RtpsUdpDataLink.h).  When reliability isn't needed, the transport framework should support the functionality needed.

lij-oci: Yeah, I understand a little better now. So the other checks like the duplicate seq# are not required by the specs for BestEffort. I factored out the DirectedWrite logic into RtpsUdpReceiveStrategy. The logic is comparing the inline qos reader id with the current reader id. How do we get the current reader id in RtpsUdpReceiveStrategy? To call it elsewhere, such as in RtpsUdpDataLink::RtpsReader, then we still need the readers to be inserted in RtpsUdpDataLink::associated(). I wish associated() were not tied to reliability settings.

mitza-oci: Duplicate sequence numbers probably need to be handled somewhere, but it's not in the scope of this effort.
The main point here is the division of responsibilities between the framework and the specific transport type.  The design goal is to have the framework handle things that are common across transports.  The framework already has support for DataLink::data_received and DataLink::data_received_include, so those should be used here.
So to try and answer your question: there is no "current reader id", that doesn't appear at this layer.  Instead the specific transport type (RtpsUdpReceiveStrategy) can make method calls on the DataLink to tell it to deliver this sample only to those associated readers that have a specific GUID (or are in given set of GUIDs).

tests/transport/rtps_directed_write/publisher.cpp
```diff
@@ -0,0 +1,308 @@
...
+private:
+  static const bool hostIsBigEndian = !ACE_CDR_BYTE_ORDER;
+  static const ACE_CDR::ULong encap = 0x00000100; // {CDR_LE, options} in BE format
+  static const CORBA::Octet DE  = OpenDDS::RTPS::FLAG_D | OpenDDS::RTPS::FLAG_E;
+  static const CORBA::Octet DEQ = OpenDDS::RTPS::FLAG_D | OpenDDS::RTPS::FLAG_E | OpenDDS::RTPS::FLAG_Q;
+
+  void force_inline_qos(bool val) {
+    OpenDDS::DCPS::RtpsUdpDataLink::force_inline_qos_ = val;
+  }
+  void log_time(const ACE_Time_Value& t) const {
```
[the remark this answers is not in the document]
iguessthislldo: > Do we have IO real functions for ACE_Time_Value, etc? If so, we should probably use them here.

There is a `std::ostream <<`, if that's what you mean. Other than that I think all we have is whatever the standard library and ACE are providing.

> Presumably we don't have to worry about calling gettimeofday manually with the monotonic time changes if we haven't enabled monotonic time, even if we're putting the results in an INFO_TS?

Monotonic time should only be used for internal timing and using `gettimeofday` is fine as long as it's a test and not one of the main libraries.

tests/transport/rtps_directed_write/Domain.cpp
```diff
@@ -0,0 +1,98 @@
...
+        break;
+      }
+    }
+#ifdef OPENDDS_SAFETY_PROFILE
+    if (host == ACE_TEXT("localhost")) {
+      host = "127.0.0.1";
+    }
+#endif
+    if (host.empty() || port == 0) {
+      ACE_DEBUG((LM_INFO, "ERROR: -h <host> and -p <port> are missing\n"));
```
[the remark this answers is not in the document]
mitza-oci: That is not required due to ACE's support for both character types:
https://github.com/DOCGroup/ACE_TAO/blob/3cb77e292612ea00bfc9b39d2d7103c50fa5e54c/ACE/ace/Log_Msg.h#L572

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -815,6 +814,38 @@ RtpsUdpReceiveStrategy::do_not_withhold_data_from(const RepoId& sub_id)
...
+  for (CORBA::ULong i = 0; i < ds.inlineQos.length() && !match; ++i) {
+    if (ds.inlineQos[i]._d() == RTPS::PID_DIRECTED_WRITE) {
+      directedWrite = true;
+      if (ds.inlineQos[i].guid() == readerID) {
+        match = true;
+      }
+    }
+  }
+  if (directedWrite && !match) {
+    readers_withheld_.insert(readerID);
```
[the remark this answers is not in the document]
mitza-oci: `readers_withheld_`'s definition is not changing in this PR, so while we might want to change it I'd like to keep that for a different PR (unless we really need it here).

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1218,6 +1218,13 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
     WriterInfo& info = wi->second;
     SequenceNumber seq;
     seq.setValue(data.writerSN.high, data.writerSN.low);
+
+    if (link->receive_strategy()->withholdDirectedWrite(data, id_)) {
```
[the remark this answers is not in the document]
mitza-oci: I think this is essentially what I was saying in https://github.com/objectcomputing/OpenDDS/pull/1381#discussion_r351527163, so hopefully we agree.

tests/transport/rtps_directed_write/Domain.cpp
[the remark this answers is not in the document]
simpsont-oci: lol... ACE_ANTI_TCHAR... never knew that existed.

tests/transport/rtps_directed_write/Domain.cpp
```diff
@@ -0,0 +1,98 @@
+#include "Domain.h"
+#include "dds/DCPS/transport/framework/TransportExceptions.h"
+#include <ace/Get_Opt.h>
+
+Domain::Domain(int argc, ACE_TCHAR* argv[]) :
```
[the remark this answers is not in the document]
lij-oci: Domain is replaced by AppConfig. I created Domain with the training exercises to manage domain id, domain participant, etc. Here, in the transport level, it doesn't fit. I felt the same way when I borrowed it here for lack of better names, my bad.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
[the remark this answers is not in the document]
lij-oci: The original code was like this without the 3-line if-statement:
    recvd_sample_ = &sample;
    readers_selected_.clear();
    readers_withheld_.clear();
    // If this sample should be withheld from some readers in order to maintain
    // in-order delivery, link_->received() will add it to readers_withheld_ otherwise
    // it will be added to readers_selected_
    link_->received(data, receiver_.source_guid_prefix_);
    if (readers_selected_.empty() && readers_withheld_.empty()) {
      listDirectedWriteReaders(data);
    }
    recvd_sample_ = 0;

I had the same feeling that not only the two lists readers_withheld_ and readers_selected_, but also recvd_sample_ were used more like local variables instead of instance variables. Notice that recvd_sample_ is set to &sample right before clearing the two lists and then set to 0 right after the call to link_->received(). I think this is because of the closely tied relationship between strategy and datalink objects. That is why link_->received() can fill the two lists without even taking them as mutable references (it could not do that if they were local variables).

I also agree with Adam that it's best to change this tangled relationship in a separate PR, because that will affect all the other delivery strategies, not just the DirectedWrite. 

The real reason that best effort readers did not get the appropriate strategies applied to them is that the current code does not fill the readers_of_writer_ list in RtpsUdpDataLink::associated(). Strategies like duplicate seq# are still not applied to best effort readers. There should be a better way than to tie association with receiving strategies, because this does not extend well. Adding a new strategy that needs to work for both, such DirectedWrite, will have problems.  

I had another version of directedWrite code:
    if (processDirectedWrite(sample, data)) {
      break;
    }
The processDirectedWrite() method did not use the two members (readers_withheld_ and readers_selected_). It retrieved the directedWrite guids into a local variable, and call link_->data_received_include(sample, readers);
This version worked for best effort readers, but hung on reliable readers.

Now we have a way to process DirectedWrite messages for both reliable and best effort readers, and a test that can successfully verify for both situations. I'd like to use a separate PR to separate the responsibilities between the strategy and the datalink objects.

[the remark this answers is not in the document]
mitza-oci: In this PR, please see what needs to happen to remove the call from `RtpsUdpDataLink::RtpsReader` to `RtpsUdpReceiveStrategy`.

The data link's responsibility is to implement the reliability protocol for rtps_udp.  It shouldn't need to be involved in directed write.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -416,10 +429,26 @@ RtpsUdpReceiveStrategy::deliver_sample_i(ReceivedDataSample& sample,
 
 #ifdef OPENDDS_SECURITY
         if (decode_payload(sample, data)) {
-          link_->data_received_include(sample, readers_selected_);
+          if (directedWriteReaders.empty()) {
```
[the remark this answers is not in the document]
mitza-oci: Keep in mind that directed write is both unused and (if/when used) necessarily small sets (so they can fit in a data submessage).  I'm not opposed to making something more efficient, but we can use it to our advantage that one of the two sets involved is small.

dds/DCPS/GuidUtils.cpp
```diff
@@ -51,6 +52,17 @@ to_string(const GUID_t& guid)
     '.' + to_string(guid.entityId);
 }
 
+bool operator<(const GUID_t& lhs, const GUID_t& rhs)
```
simpsont-oci: @mitza-oci I know historically we haven't overloaded operator< for our set & map types, even though the syntax gets a lot cleaner, but I can't remember if there was a great reason (e.g. does FACE require avoiding operator overload?). Think I would prefer we stick with one approach to minimize duplicate / unused code, which will unfortunately mean getting rid of this function and switching to specify Guid_tKeyLessThan below. I'm OK with leaving both of the template functions, since they won't get instantiated unless they're used.

mitza-oci: I don't know of any requirement to avoid overloading.
At this point in the release cycle we should be consistent with how it's already done.  We can revisit it and change it everywhere some other time.

dds/DCPS/Util.h
```diff
@@ -140,6 +140,35 @@ void push_back(Seq& seq, const typename Seq::value_type& val)
...
 }
 
+// Constructs a sorted intersect of the given two sorted ranges [a,aEnd) and [b,bEnd).
+// (for pre-c++17 code for the similar effect of std::set_intersection in c++17)
+template <typename InputIteratorA, typename InputIteratorB, typename OutputIterator>
+OutputIterator intersect_sorted_ranges(InputIteratorA a, InputIteratorA aEnd,
+                                       InputIteratorB b, InputIteratorB bEnd,
+                                       OutputIterator intersect)
+{
+  while (a != aEnd && b != bEnd) {
```
mitza-oci: This function could delegate to the other one using a LessThan which class < (like std::less but for two different types).
We could also require `A` and `B` to be the same type, but it's probably better to have the same interface as std.

dds/DCPS/Util.h
```diff
@@ -140,6 +140,35 @@ void push_back(Seq& seq, const typename Seq::value_type& val)
...
+    if      (*a < *b) { ++a; }
```
mitza-oci: guidelines:
> 
> - ... Do not add extra spaces to make syntax
>   elements (that span lines/statements) line up. This only causes unnecessary
>   changes in adjacent lines as the code evolves.

dds/DCPS/GuidUtils.cpp
```diff
@@ -51,6 +52,13 @@ to_string(const GUID_t& guid)
     '.' + to_string(guid.entityId);
 }
 
+void intersect(const RepoIdSet& a, const RepoIdSet& b, RepoIdSet& result)
+{
+  result.clear();
+  intersect_sorted_ranges(a.cbegin(), a.end(), b.cbegin(), b.end(),
```
mitza-oci: I'm not sure about the use of `cbegin` here as there's no matching `cend`, but probably more relevant is that `cbegin` is not supported pre-C++11.

dds/DCPS/Util.h
mitza-oci: Please update this code per guidelines (see 12/19 comment), or comment here as to why not.

PR #1398, opened 2019-12-02
Update Guidelines

iguessthislldo:
Update guidelines based on cases such as https://github.com/objectcomputing/OpenDDS/pull/1379#discussion_r351477282 and https://github.com/objectcomputing/OpenDDS/pull/1339#discussion_r338597776.

Removed bit in Doxygen, because it doesn't really work like I thought it did.

### DIFF
--- a/docs/guidelines.md
+++ b/docs/guidelines.md
@@ -305,8 +305,8 @@ For more information, see [the Doxygen manual](http://www.doxygen.nl/manual/).
 
 ### Preprocessor
 
-- Prefer `#ifdef ...` and `#ifndef ...` to `#if defined ...` and
-  `#if !defined ...` when testing if a single macro is defined.
+- Prefer `#ifdef` and `#ifndef` to `#if defined` and `#if !defined` when
+  testing if a single macro is defined.
 - As stated before, preprocessor macros visible to user code must begin with
   `OPENDDS_`.
 - Ignoring the header guard if there is one, preprocessor statements should be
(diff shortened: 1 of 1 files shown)


docs/guidelines.md
```diff
@@ -290,29 +297,55 @@ The extra `*` on the multiline comment and `/` on the single line comment are
...
+
+- Prefer `#ifdef` and `#ifndef` to `#if defined` and `#if !defined` when
+  testing if a single macro is defined.
+- As stated before, preprocessor macros visible to user code must begin with
+  `OPENDDS_`.
+- Ignoring the header guard if there is one, preprocessor statements should be
+  indented using two spaces starting at the pound symbol, like so:
+
+```C++
+#if defined(X) && defined(Y)
```
mitza-oci:
```suggestion
#if defined X && defined Y
```
Any reason not to?  `defined` is an operator and not a function.

docs/guidelines.md
```diff
@@ -206,7 +207,9 @@ The punctuation placement rules can be summarized as:
   desirable.
 - Use the constructor initializer list and make sure its order matches the
   declaration order.
-- Prefer pre-increment/decrement (`++x`) to post-increment/decrement (`x++`)
+- Prefer pre-increment/decrement (`++x`) to post-increment/decrement (`x++`).
+  For objects this avoids creating a unused copy of the original object. It's
+  also recommended to do this for primitives just to enforce the habit.
```
mitza-oci: We didn't necessarily write justifications for all of these (that might turn in to a much larger document).  In this case we could de-emphasize the performance angle, since compilers can figure that out anyway.  It's a logically simpler operation -- the similar rule is in C++ core guidelines and https://google.github.io/styleguide/cppguide.html

iguessthislldo: I think it would be nice, but I'm fine with leaving out a justification. My issue was I read this originally I thought it was just talking about iterators. How about:

```
- Prefer pre-increment/decrement (`++x`) to post-increment/decrement (`x++`) for both objects and non-objects.
```

Also I could find a rule in the [Google Guidelines](https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement), but not in the Core guidelines.

mitza-oci: I searched for "prefix" on http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines

docs/guidelines.md
iguessthislldo: No, just the way I'm used to writing it. Do you want me to add that as well since you're asking about it?

docs/guidelines.md
iguessthislldo: I missed [that](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#p9-dont-waste-time-or-space). I was looking for it as its own rule.

iguessthislldo:
I don't have any more changes at the moment, so I'm okay with merging it.

PR #1389, opened 2019-11-21
New built-in topic for participant network location/connection details

jonesc-oci:
These are changes for the new participant location built-in topic. They include a test for the new feature as well as a shutdown listener for the ICE agent which fixes a memory issue.

PR #1399, opened 2019-12-03
Fix tests that started failing with #1389

jonesc-oci:
Participant Location (connection endpoint) changes to fix failing tests

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -1934,14 +1934,23 @@ bool
 DomainParticipantImpl::is_clean() const
 {
   bool sub_is_clean = subscribers_.empty();
-  bool topics_is_clean = topics_.size() == 0;
+  bool topics_is_clean = true;
+  int non_bit_topic_count = 0;
+
+  // check that the only remaining topics are built-in topics
+  for (TopicMap::const_iterator it = topics_.begin(); it != topics_.end(); ++it) {
+    if (!topicIsBIT(it->second.pair_.svt_->get_name(),it->second.pair_.svt_->get_type_name())) {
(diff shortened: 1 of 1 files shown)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1934,14 +1934,22 @@ bool
 DomainParticipantImpl::is_clean() const
 {
   bool sub_is_clean = subscribers_.empty();
-  bool topics_is_clean = topics_.size() == 0;
+  bool topics_is_clean = true;
+  int non_bit_topic_count = 0;
+
+  // check that the only remaining topics are built-in topics
+  for (TopicMap::const_iterator it = topics_.begin(); it != topics_.end(); ++it) {
+    if (!topicIsBIT(it->second.pair_.svt_->get_name(),it->second.pair_.svt_->get_type_name())) {
```
mitza-oci: this leaks memory from the strings

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1934,14 +1934,22 @@ bool
...
+      ++non_bit_topic_count;
```
mitza-oci: does it need to track this count separately?

dds/DCPS/DomainParticipantImpl.cpp
mitza-oci: The change you made works.  The other way to deal with it is `TopicImpl::type_name` instead of `get_type_name`, then the similar method would have to be added for `name`, I don't think it's there yet.

jonesc-oci: I've added a TopicImpl::topic_name() which avoids extra string copies in is_clean().

PR #1410, opened 2019-12-13
ParticipantLocation built-in topic Java support and test.

jonesc-oci:


java/dds/OpenDDS/DCPS/BuiltinTopicUtils.java
```diff
@@ -1,4 +1,4 @@
-/*
+OpenDDSParticipantLocation/*
```
iguessthislldo: This looks like a typo

PR #1416, opened 2019-12-19
Replaced monitor ptr with smart ptr to fix memory leaks.

kuznetsovmoci:


mitza-oci:
Error: ./dds/DCPS/TopicImpl.cpp:41: found trailing whitespace

PR #1419, opened 2019-12-22
Docker updates

adamsj-oci:
Adds ability to switch FROM at build time using docker args. If no BASIS environment variable is provided to the hooks/build file then the build will use ubuntu:bionic by default. Otherwise, providing a BASIS environment variable to the hooks/build will switch the FROM line.


adamsj-oci:
The unsuccessful  checks don't seem to be relevant. The Codacy check is referring to a docker image tag which is there but managed in a different way that this tool doesn't seem to handle. 

I'm not sure why the test failed. It looks like it failed in a similar way on the previous PR build. It doesn't seem to be related to a docker change. In any case, the resulting docker image should be exactly the same.

iguessthislldo:
I agree with you about codacy, it isn't handling that correctly. About the OpenDDS PR, it's running all the tests, so that just happens to be broken.

hooks/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
mitza-oci: Add a comment at the top of this file, or a README in this directory, to explain what this is doing here.  Is the name of this file hard-coded in some docker thing or can it be moved somewhere else?

adamsj-oci: I added a README with a link to the supporting documentation since you can add other custom build phase hook files for Dockerhub.

hooks/README
```diff
@@ -0,0 +1,2 @@
+The hooks directory represents build phase hooks for Dockerhub builds. 
+https://docs.docker.com/docker-hub/builds/advanced/
```
mitza-oci: This is missing a newline (github shows an error marker).

adamsj-oci: Added newline.

hooks/build
iguessthislldo: I would rather not have something like this in the root directory if we don't need it. How are you passing this variable from dockerhub and is there another way you could do it?

PR #1420, opened 2019-12-30
Fixes for Monitor Test

kuznetsovmoci:


iguessthislldo:
This needs a more specific title. If it's just a fix for the monitor test, then "Fixes for Monitor Test" would be a decent name.

PR #1422, opened 2020-01-03
Improvements for Topic Type Annotations

iguessthislldo:
- Made root default nested behavior like the traditional behavior or `--default-nested` is now the default.
- Use the term "topic type annotations" instead of "topic annotations" to try to avoid giving the impression that only the `@topic` annotation is the only thing being talked about or it's the only annotation that should be used.
- Warn when structs that are declared to be topic types using both `DCPS_DATA_TYPE` and the new topic type system have inconsistent keys between the two systems.
- Suppress DCPS_DATA_TYPE warnings in generated modeling IDL.

### DIFF
--- a/MPC/modules/IDLBase.pm
+++ b/MPC/modules/IDLBase.pm
@@ -107,13 +107,12 @@ my $tsreg  = 'TypeSupport\.idl';
 # ************************************************************
 
 sub do_cached_parse {
-  my($self, $file, $flags, $called_from_base) = @_;
+  my($self, $file, $flags) = @_;
 
-  ## If we are being called from the base type (i.e., IDLBase::get_output)
-  ## we need to set the default_nested to true.  Otherwise, in certain
-  ## environments, we can get extra output files being added to the project
-  ## that are never going to exist.
-  $self->{'default_nested'} = $called_from_base;
+  # By default opendds_idl considers all types not to be topic types (aka
+  # "nested") unless --no-default-nested was passed or annotations say
+  # otherwise.
+  $self->{'default_nested'} = 1;
 
   ## Set up the macros and include paths supplied in the command flags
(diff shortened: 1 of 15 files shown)


dds/idl/dds_visitor.cpp
```diff
@@ -233,21 +236,79 @@ dds_visitor::visit_structure(AST_Structure* node)
...
+        bool header = false;
+        for (set<string>::iterator i = topic_type_keys.begin();
+            i != topic_type_keys.end(); ++i) {
+          if (dcps_data_type_keys.find(*i) == dcps_data_type_keys.end()) {
+            if (!header) {
+              message += "\n\n"
+                "  The following keys were declared using @key, but not DCPS_DATA_KEY:";
+              header = true;
+            }
+            message += string("\n    ") + *i;
```
mitza-oci:
```suggestion
            message += "\n    " + *i;
```
Not a big deal but `operator+` already handles this.

MPC/modules/IDLBase.pm
```diff
@@ -127,8 +127,8 @@ sub do_cached_parse {
       elsif ($arg =~ /^\-I(.+)/) {
         push(@include, $1);
       }
-      elsif ($arg eq '--default-nested') {
-        $self->{'default_nested'} = 1;
+      elsif ($arg eq '--no-default-nested') {
+        $self->{'default_nested'} = 0;
```
mitza-oci: Is this being set to 1 somewhere else when `--default-nested` or no arg is present?  Java builds are failing on CI.

iguessthislldo: Up on line 116 we are setting `default_nested` based on if we are being called from the base, which might be the problem. Also we should probably check need both `--default-nested` and `--no-default-nested` though because the current behavior you can use both and the last one is the one that takes effect.

PR #1424, opened 2020-01-03
Connection method Coverity fixes

jonesc-oci:


mitza-oci:
Java test on Jenkins build: `ERROR: take_next_sample`

PR #1426, opened 2020-01-03
Monitor test report count fixes

kuznetsovmoci:


tests/DCPS/Monitor/Writer.cpp
```diff
@@ -129,6 +129,9 @@ Writer::svc()
     e._tao_print_exception("Exception caught in svc():");
   }
 
+  // Patch to prevent premature shutdown
+  Sleep(1000);
```
mitza-oci: This isn't going to compile on all platforms

PR #1429, opened 2020-01-07
Fix Issues with Installed OpenDDS

iguessthislldo:
I'm using a prebuilt OpenDDS debian package for CI testing PyOpenDDS. I applied a workaround for https://github.com/DOCGroup/ACE_TAO/pull/1015 (just copying the file) , but [it still didn't work](https://github.com/iguessthislldo/pyopendds/commit/1845e136d1ab3f099b2dec1d60ce24d431237087/checks?check_suite_id=387454185).

I looked into it and it's looking into PREFIX/bin for it, but scripts are put into PREFIX/share/ace/bin. As that directory also has symlinks to the binaries PREFIX/bin, change the ACE_BIN_DIR variable to that.

### DIFF
--- a/DevGuideExamples/DCPS/Messenger/CMakeLists.txt
+++ b/DevGuideExamples/DCPS/Messenger/CMakeLists.txt
@@ -1,7 +1,7 @@
 project(OpenDDS_DevGuide_Messenger CXX)
 cmake_minimum_required(VERSION 3.8.2)
 
-find_package(OpenDDS)
+find_package(OpenDDS REQUIRED)
 
 set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -224,6 +224,7 @@ jobs:
         ConfigOpts: --no-inline --java=/usr/lib/jvm/zulu-12-azure-amd64
       Release:
         ConfigOpts: --no-debug --optimize --java=/usr/lib/jvm/zulu-12-azure-amd64
+        InstallPrefix: 'install'
       Safety:
         ConfigOpts: --safety-profile
       SafetyBaseNoBuiltinTopics:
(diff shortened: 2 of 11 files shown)


iguessthislldo:
@adamsj-oci are you still having this issue even after you copy the file?

mitza-oci:
Could we extend (at least one of) the Azure Pipelines CI Linux builds to do a `make install` and then build both an MPC and a CMake example app using the installed OpenDDS?

adamsj-oci:
> @adamsj-oci are you still having this issue even after you copy the file?

Once I add the file in the /usr/local/share/ace/bin, the missing script error goes away. I'm not sure if the script is actually working correctly once it's found.

iguessthislldo:
> Could we extend (at least one of) the Azure Pipelines CI Linux builds to do a make install and then build both an MPC and a CMake example app using the installed OpenDDS?

Doing that now. As part of that I'm trying to build Messenger using the installed OpenDDS and MPC on my computer, but it's broken too...

iguessthislldo:
There are issues with path references in PREFIX/share/dds that I'm going to be pusing fixes for and a problem with links to TAO_IDL libraries in PREFIX/share/ace/lib. Also it would be nice if MPC could be installed.

iguessthislldo:
This is mostly unrelated, but we install PerlDDS, but ACE doesn't install PerlACE, so it appears pointless to me. Since I'm going to create a PR with ACE/TAO, can I add PerlACE or remove PerlDDS here?

mitza-oci:
> 
> 
> This is mostly unrelated, but we install PerlDDS, but ACE doesn't install PerlACE, so it appears pointless to me. Since I'm going to create a PR with ACE/TAO, can I add PerlACE or remove PerlDDS here?

I don't think we need to install the support scripts for running tests, since the tests aren't installed.  They may be useful for someone doing application development, but in that case they can be copied separately.

mitza-oci:
In tests/cmake_integration/Messenger/Messenger_?/CMakeLists.txt should the `find_package(OpenDDS)` use `REQUIRED`?

iguessthislldo:
> In tests/cmake_integration/Messenger/Messenger_?/CMakeLists.txt should the `find_package(OpenDDS)` use `REQUIRED`?

Yes, good catch. I also needed to change devguide messenger to use `REQUIRED` as well.

iguessthislldo:
Should I also have it run the Messengers built with the installed OpenDDS?

mitza-oci:
> Should I also have it run the Messengers built with the installed OpenDDS?

We will try this as part of #1437

PR #1437, opened 2020-01-09
Integration Test cleanups

kuznetsovmoci:


azure-pipelines.yml
```diff
@@ -147,6 +147,16 @@ jobs:
   - script: |
       perl "%DDS_ROOT%\tests\cmake_integration\run_ci_tests.pl" --build-config=$(BuildConfiguration) --generator="Visual Studio 15 2017" --arch $(BuildPlatform)
     displayName: Compile and run CMake tests
+  - script: |
+      cd DevGuideExamples\DCPS\Messenger
+      perl appveyor_run_test.pl
```
mitza-oci: Consider renaming this perl script.
Should a similar section be added to Linux and macOS configurations?

azure-pipelines.yml
```diff
@@ -314,3 +312,11 @@ jobs:
...
       $DDS_ROOT/tests/cmake_integration/run_ci_tests.pl
     displayName: Compile and run CMake tests
+  - script: |
+      cd DevGuideExamples/DCPS/Messenger
+      perl integration_run_test.pl
+    displayName: Run Examples DCPS Messenger
+  - script: |
+      cd tests/DCPS/Messenger
+      perl run_test.pl rtps
+    displayName: Run DCPS Messenger Test
```
iguessthislldo: Github says the file is missing a final newline.

azure-pipelines.yml
kuznetsovmoci: I've added it for both, but the Test fails for MacOS (Example works fine).

dds/monitor/SPMonitorImpl.cpp
```diff
@@ -46,8 +46,15 @@ SPMonitorImpl::report()
     report.host = this->hostname_.c_str();
     report.pid  = this->pid_;
     DDS::DomainParticipantFactory_var pf = TheParticipantFactory;
-    const DomainParticipantFactoryImpl::DPMap& participants =
-      dynamic_cast<DomainParticipantFactoryImpl*>(pf.in())->participants();
+
+    OpenDDS::DCPS::DomainParticipantFactoryImpl* pi = dynamic_cast<DomainParticipantFactoryImpl*>(pf.in());
```
mitza-oci: Please make this a separate PR, it's not related to the other changes.

tests/DCPS/Messenger/run_test.pl
```diff
@@ -29,6 +29,19 @@
 my $is_rtps_disc = 0;
 my $DCPSREPO;
 
+my $filename = "subscriber";
```
mitza-oci: Could this be changed in the common modules ($test->process or $test->start_process)?  It should apply to any test, and the common modules already know about file names extensions, etc.

.travis.yml
```diff
@@ -89,6 +89,14 @@ script:
   - if [ "$SAFETY_PROFILE" != "1" ]; then
       $DDS_ROOT/tests/cmake_integration/run_ci_tests.pl;
     fi
+  - if [ "$NO_TESTS" == "1" ]; then
+      echo Building Messenger;
+      source setenv.sh;
```
mitza-oci: I thought it already had the environment set

.travis.yml
```diff
@@ -89,6 +89,14 @@ script:
...
+      cd tests/DCPS/Messenger;
+      mwc.pl -type gnuace;
```
mitza-oci: You can combine this line with the previous one since `mwc.pl` takes a directory name as a command line argument.  It avoid the need to manage the `cd` state in different steps.

mitza-oci:
I added a PR to vcpkg to fix the Xerces problem https://github.com/microsoft/vcpkg/pull/9702

.travis.yml
kuznetsovmoci: Without it it gives an error: "mwc.pl: command not found"

mitza-oci: Right, it has some environment variables but not exactly the same as setenv.sh.  There's no need for it to be on the PATH when you know where to find it (`$ACE_ROOT/bin/mwc.pl`).

tests/DCPS/Messenger/run_test.pl
kuznetsovmoci: I could be wrong, but it looks like common module functions like 'process' do not care about extension, as they can start the process without it. I can add this to the 'process' function:

  my $extension = ".exe";
  my $executable_ext = $executable + $extension;
  if (!(-e $executable) && !(-e $executable_ext)) {
    print STDERR "ERROR: executable \"$executable\" does not exist \n";
    $self->{status} = -1;
    return;
  }

Also, messenger script does not use 'process', or 'start_process', but 'PerlDDS::create_process'. I can either change that to use process / start_process or add similar check in PerlDDS::create_process

.travis.yml
kuznetsovmoci: This sequence was actually wrong. The correct one is:
      cd tests/DCPS/Messenger;
      mwc.pl -type gnuace .;
      make DDS_Messenger_Subscriber DDS_Messenger_Publisher;
      cd ../../..;

I am guessing I need to be in tests/DCPS/Messenger to correctly run nmake?

dds/monitor/SPMonitorImpl.cpp
kuznetsovmoci: Will do.

tests/DCPS/Messenger/run_test.pl
mitza-oci: Some older scripts don't use "PerlDDS::TestFramework", you can migrate them to use it.

.travis.yml
mitza-oci: I wouldn't do it that way, but maybe that's just me.  I don't like managing the current directory state separately, it's something else that can go wrong:
```
$ACE_ROOT/bin/mwc.pl -type gnuace tests/DCPS/Messenger;
make -C tests/DCPS/Messenger
```

kuznetsovmoci: I agree. This is done now.

.travis.yml
kuznetsovmoci: Fixed.

tests/DCPS/Messenger/run_test.pl
kuznetsovmoci: Question: If I add the code above to 'process' sub, how would it stop the run_test script that called it, if 'process' fails? It sets $self->{status} = -1;, but I am not sure it does anything. I can start returning $self->{status}, and then check it in run_test.pl. Hope it won't break any existing scripts...

kuznetsovmoci: I have added the code above to 'process' sub.

PR #1439, opened 2020-01-09
OpenDDS crashes when UseIce=1 but DCPSSecurity=0

jrw972:
Solution: Check for DCPSSecurity whenever UseIce is active.


mitza-oci:
Looks like it needs ifdefs
```
error C2039: 'get_security': is not a member of 'OpenDDS::DCPS::Service_Participant'
```

PR #1440, opened 2020-01-10
Make install with DESTDIR (for Java bindings)

iguessthislldo:
The last build of #1429 was actually a failure for the Linux Release build, but I have no idea why. I'm also adding `set -e` to the scripts in the Linux builds to make them fail on non-zero exit statuses.

### DIFF
--- a/dds/InfoRepo/DCPSInfoRepo.mpc
+++ b/dds/InfoRepo/DCPSInfoRepo.mpc
@@ -156,9 +156,4 @@ project(*Main): dcpsexe, iortable, bidir_giop, DCPSInfoRepo_bits, imr_client, sv
 
   TypeSupport_Files {
   }
-
-  verbatim(gnuace, postinstall) {
-"	@$(MKDIR) $(DESTDIR)$(INSTALL_PREFIX)/share/dds/bin"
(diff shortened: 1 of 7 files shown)


dds/InfoRepo/DCPSInfoRepo.mpc
```diff
@@ -156,4 +156,9 @@ project(*Main): dcpsexe, iortable, bidir_giop, DCPSInfoRepo_bits, imr_client, sv
 
   TypeSupport_Files {
   }
+
+  verbatim(gnuace, postinstall) {
+"	@$(MKDIR) $(DESTDIR)$(INSTALL_PREFIX)/share/dds/bin"
```
mitza-oci: Why is this a `share` file instead of just in `bin`?

iguessthislldo: It was a fix for PerlDDS not finding info repo unless I did this. I was going to look into it more, but it looks like I accidentally committed it. Do you want me to remove it?

mitza-oci: Yes, if we're going to start using PerlDDS with the installed tree, then PerlDDS will have to be adapted to that.  Once it is then it could be useful for more general testing than just CI.

PR #1441, opened 2020-01-11
Check for single participant with required location mask in ParticipantLocationTopic

jrw972:
The publisher and subscriber in the ParticipantLocationTopic test
should discover each other as well as the application participant in
the relay.  Furthermore, they may discover participants due to
cross-talk and interference.  The publisher and subscriber should
check that a single participant has all of the expected locations.

Solution: Record the accumulated mask for each participant and check
that at least one participant has the expected mask.


mitza-oci:
Jenkins build (for this test):
```
2020-01-14 02:08:03: ERROR: subscriber returned 1 (started at 2020-01-14 02:07:58)
```

Issue #1433, opened 2020-01-08
macro naming conflict with other libraries


xieshuaix:
The following macro in dds/DCPS/Definitions.h is not namespace guarded and causes naming conflicts when using your library with popular ones including boost, torch (C++ distributions of PyTorch), etc. Could you please fix this? Thanks.

```
#define DUP true
#define NO_DUP false
```

iguessthislldo:
By "namespace guarded", I assume you mean that the macros should be called something like `OPENDDS_DUP`. We have do have a rule that says that for the reasons you described and this is breaking that rule. Thanks for pointing it out.

As for fixing it I think we should probably just get rid of the macros all together and have the callers use true and false.

mitza-oci:
@xieshuaix please create a PR to remove it

xieshuaix:
> @xieshuaix please create a PR to remove it

Done. Please check the PR: [https://github.com/objectcomputing/OpenDDS/pull/1442](url)

iguessthislldo:
I'm closing this because of #1442 was merged. From #1366 I can see there are other macros that don't follow the `OPENDDS` rule, but this what the shortest I think.

PR #1442, opened 2020-01-12
Remove macros: DUP and NO_DUP to avoid naming conflicts.

xieshuaix:
This PR removes two macros to avoid naming conflicts with other libraries as explained in this issue: #1433


iguessthislldo:
The macros were being used, so you can't just take them out. You have to replace the usages of `DUP` and `NO_DUP` with `true` and `false`.

xieshuaix:
> The macros were being used, so you can't just take them out. You have to replace the usages of `DUP` and `NO_DUP` with `true` and `false`.

Done.

jwillemsen:
Windows seems to fail due to https://github.com/microsoft/vcpkg/issues/9654

iguessthislldo:
Thanks, I have added a check for this to my WIP linter in #1366.

PR #1443, opened 2020-01-13
docker option to pick ACE source

adamsj-oci:
Use `ACE_CONFIG_OPTION` and `ACE_SOURCE_DIR` options to pick difference ACE source. 

`ACE_CONFIG_OPTION` can be used to select `--ace-github-latest` instead of the default `--doc-group`.

`ACE_SOURCE_DIR` copies the correct files based on the ACE source to `/usr/local/share/ace/MPC`. Defaults to `/opt/OpenDDS/ACE_wrappers/MPC`. Use `/opt/OpenDDS/ACE_TAO/ACE/MPC` when using `--ace-github-latest` in `ACE_CONFIG_OPTION`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,8 @@ RUN apt-get update && apt-get install -y \
     libxerces-c-dev \
     libssl-dev \
     perl-base \
-    perl-modules
+    perl-modules \
+    git
 
 WORKDIR /usr/src/gtest
 RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
@@ -20,11 +21,13 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
 
 ADD . /opt/OpenDDS
 
+ARG ACE_CONFIG_OPTION="--doc-group"
+ARG MPC_ROOT="/opt/OpenDDS/ACE_wrappers/MPC"
 RUN cd /opt/OpenDDS && \
-    ./configure --prefix=/usr/local --security --doc-group --std=c++11 && \
+    ./configure --prefix=/usr/local --security --std=c++11 ${ACE_CONFIG_OPTION} && \
     make && \
     make install && \
-    cp -a /opt/OpenDDS/ACE_wrappers/MPC /usr/local/share/ace/MPC
+    cp -a ${MPC_ROOT} /usr/local/share/ace/MPC
 
 ENV ACE_ROOT=/usr/local/share/ace \
     TAO_ROOT=/usr/local/share/tao \
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -224,7 +224,7 @@ jobs:
         ConfigOpts: --no-inline --java=/usr/lib/jvm/zulu-12-azure-amd64
       Release:
         ConfigOpts: --no-debug --optimize --java=/usr/lib/jvm/zulu-12-azure-amd64
-        InstallPrefix: 'install'
+        DoMakeInstall: true
       Safety:
         ConfigOpts: --safety-profile
       SafetyBaseNoBuiltinTopics:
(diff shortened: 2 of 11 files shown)


iguessthislldo:
Why `ACE_wrappers/MPC`?

iguessthislldo:
Actually forget my last comment, I now see you're copying MPC. I would suggest calling it `MPC_SOURCE_DIR` or maybe `MPC_ROOT`, which is the convention, instead.

adamsj-oci:
> Why `ACE_wrappers/MPC`?

That much I didn't pick. That's the original path. I'm just adjusting it when the `--ace-github-latest` option is used in the docker build.

hooks/build
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 if [[ -z "${BASIS}" ]]; then
-    docker build -t $IMAGE_NAME .
+    docker build --build-arg ACE_CONFIG_OPTION=$ACE_CONFIG_OPTION --build-arg MPC_ROOT=$MPC_ROOT -t $IMAGE_NAME .
 else
-    docker build --build-arg BASIS=$BASIS -t $IMAGE_NAME .
+    docker build --build-arg BASIS=$BASIS --build-arg= ACE_CONFIG_OPTION=$ACE_CONFIG_OPTION --build-arg MPC_ROOT=$MPC_ROOT -t $IMAGE_NAME .
```
mitza-oci:
```suggestion
    docker build --build-arg BASIS=$BASIS --build-arg ACE_CONFIG_OPTION=$ACE_CONFIG_OPTION --build-arg MPC_ROOT=$MPC_ROOT -t $IMAGE_NAME .
```
just a guess...?

Dockerfile
```diff
@@ -20,11 +21,13 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
...
 
+ARG ACE_CONFIG_OPTION="--doc-group"
+ARG MPC_ROOT="/opt/OpenDDS/ACE_wrappers/MPC"
 RUN cd /opt/OpenDDS && \
-    ./configure --prefix=/usr/local --security --doc-group --std=c++11 && \
+    ./configure --prefix=/usr/local --security --std=c++11 ${ACE_CONFIG_OPTION} && \
     make && \
     make install && \
-    cp -a /opt/OpenDDS/ACE_wrappers/MPC /usr/local/share/ace/MPC
+    cp -a ${MPC_ROOT} /usr/local/share/ace/MPC
```
mitza-oci:
```suggestion
    cp -a ${MPC_ROOT} /usr/local/share/MPC
```
This shouldn't be in `ace`, but we may want to keep compatibility with old docker image? @jrw972

hooks/build
adamsj-oci: I just found this too. Fixed.

Dockerfile
jrw972: I don't think backward compatibility is an issue.  IMHO, it is highly unlikely that any use of the Docker image actually cares.

My only request is that after this PR is done, go through the "Compile the Messenger Example" of https://opendds.org/quickstart/GettingStartedDocker.html to make sure it still works.

azure-pipelines.yml
```diff
@@ -233,7 +233,10 @@ jobs:
         ConfigOpts: --security --features=versioned_namespace=1
         PackageDeps: libxerces-c-dev libssl-dev cmake
       SecurityWithoutFeatures:
-        ConfigOpts: --no-inline --no-debug --no-built-in-topics --no-content-subscription --no-ownership-profile --no-object-model-profile --no-persistence-profile --security
+        ConfigOpts: >
+          --security --no-inline --no-debug
+          --no-built-in-topics --no-content-subscription --no-ownership-profile
```
neeleym: I don't think that we should be using --no-built-in-topics as those are needed for dynamic discovery aspects. Has there been prior discussion to drop them out for some reason?

mitza-oci: That's not part of Jeremy's changes. Jeremy - this branch now contains rewritten commits, please avoid this.  Merge again and force-push to reset your branch.

neeleym: Understood ... ignore my question.

Dockerfile
mitza-oci: I don't know the answer to these, but I'm in favor of `/usr/local/share/MPC`.

mitza-oci:
I don't know what's going on with this PR, but your source branch looks OK.  Maybe we just need to close it and open a new PR for the same source.

azure-pipelines.yml
adamsj-oci: Gotcha. I didn't use a separate branch for this thinking it was a quick change.

PR #1445, opened 2020-01-14
Revert default rtps_udp nak_depth to zero (infinite) to prevent lost data for reliable connections

simpsont-oci:
Though we should probably limit this per RtpsWriters based on DataWriter resource limits QoS.


dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -35,7 +35,7 @@ RtpsUdpInst::RtpsUdpInst(const OPENDDS_STRING& name)
   , ttl_(1)
   , multicast_group_address_(7401, "239.255.0.2")
   , multicast_group_address_str_("239.255.0.2:7401")
-  , nak_depth_(32) // default nak_depth in OpenDDS_Multicast
+  , nak_depth_(0) // default nak_depth in OpenDDS_Multicast
```
```suggestion
  , nak_depth_(0)
```

PR #1446, opened 2020-01-15
Add Support for Building for Android on Windows

iguessthislldo:
Fix some issues in the configure script to allow building OpenDDS for Android on Windows and add instructions to `android.md`.

For the record, this PR has a sister: https://github.com/DOCGroup/ACE_TAO/pull/1022, although the changes aren't functionally necessary.

One issue that I realized when doing this is we might need some sort of mechanism to recognize when the cross compilers have C++11+ enabled by default (See second commit).


iguessthislldo:
I'm going to update and run OpenDDS-Android on these two PRs just to make sure I didn't break the normal cross-compile.

iguessthislldo:
Building on OpenDDS-Android [here](https://travis-ci.org/iguessthislldo/OpenDDS-Android/builds/637287510)

configure
```diff
@@ -1022,12 +1042,23 @@ sub write_platform_macros {
...
+        if ($is_windows) {
+          print $PMG "HOST_EXE_EXT = .exe\n";
+        }
         print $PMG 'TAO_IDL_DEP = $(TAO_IDL)$(HOST_EXE_EXT)', "\n";
-        print $PMG 'TAO_IDL_PREPROCESSOR = ', $opts{'compiler'}, "\n";
-        print $PMG 'TAO_IDLFLAGS += -g $(HOST_ACE)/bin/ace_gperf', "\n";
+        if ($has_host_compiler) {
+          print $PMG 'TAO_IDL_PREPROCESSOR = ', $opts{'compiler'}, "\n";
+        }
+        print $PMG 'TAO_IDLFLAGS += -g $(HOST_ACE)/bin/ace_gperf$(HOST_EXE_EXT)', "\n";
```
mitza-oci:
```suggestion
        print $PMG 'TAO_IDLFLAGS += -g $(HOST_ACE)/bin/ace_gperf', "\n";
```
Should be OK without it since this is a command line not a file name.

docs/android.md
```diff
@@ -57,12 +53,40 @@ To follow along this guide and build OpenDDS you will need:
...
+## Building on Windows
+
+If using you're using Windows Subsystem for Linux (WSL), Docker, or anything
+like that, then when you done with the guide you can copy the resulting
+libraries from the virtual environment to Windows and where they can be used in
+Android Studio as they would be used on Linux. Also you can skip the rest of
+this section and any **Windows Users:** notices.
+
+If you want to build OpenDDS for Android on Windows without any kind of
+vitalization you will have to follow the **Windows Users:** notices.
```
mitza-oci:
```suggestion
virtualization you will have to follow the **Windows Users:** notices.
```

docs/android.md
```diff
@@ -109,6 +142,37 @@ those sections before configuring and building OpenDDS.
 PATH=$PATH:$TOOLCHAIN/bin make # Pass -j/--jobs with an appropriate value or this'll take a while...
 ```
 
+**Windows Users:** The command is similar with these exceptions:
+
+- `--host-tools` with the location of the OpenDDS host tools that were built
+  using Visual Studio must be passed to `configure`.
+
+- In addition to the Android toolchain, you will also need MinGW utilities in
+  your `%PATH` and the command prompt environment for Visual Studio. The
```
mitza-oci:
```suggestion
  your `PATH` and the command prompt environment for Visual Studio. The
```

docs/android.md
```diff
@@ -109,6 +142,37 @@ those sections before configuring and building OpenDDS.
...
+  default installation location of the MinGW utilities is
+  `C:\MinGW\msys\1.0\bin`. The Visual Studio command prompt is the same as when
```
mitza-oci: Re-word this part (starting `The Visual Studio`), give the user explicit steps to run ("Start a new Visual Studio Command Prompt...").

docs/android.md
```diff
@@ -109,6 +142,37 @@ those sections before configuring and building OpenDDS.
...
+  running environment or else the `ACE_ROOT` and `TAO_ROOT` environment
+  variables will mess up the build for Android, so start a new one if you
+  haven't already.
+
+In the end the commands will look like this:
+
+```Batch
+configure --doc-group --target=android --macros=ANDROID_ABI=armeabi-v7a --host-tools=%HOST_DDS%
+call "C:\Program Files (x86)\Microsoft Visual Studio\Version\Offering\Common7\Tools\VsDevCmd.bat"
+set PATH=%TOOLCHAIN%/bin;%PATH%
```
mitza-oci:
```suggestion
set PATH=%TOOLCHAIN%\bin;%PATH%
```

docs/android.md
```diff
@@ -233,48 +297,59 @@ the [ABI/architecture table](#abi-table)).
...
- - Required to use InfoRepo Peer Discovery:
-   - `$DDS_ROOT/lib/libOpenDDS_InfoRepoDiscovery.so`
-   - `$ACE_ROOT/lib/libTAO_PortableServer.so`
+ - The following are the peer discovery libraries. Static discovery is built
+   into `libOpenDDS_Dcps.so`, but most likely you will want one of these:
+
+   - Required to use RTPS Peer Discovery:
+     - `$DDS_ROOT/lib/libOpenDDS_Rtps.so`
+
+   - Required to use the Information Repository Peer Discovery:
```
mitza-oci:
```suggestion
   - Required to use  DCPSInfoRepo Discovery:
```

docs/android.md
```diff
@@ -109,6 +142,37 @@ those sections before configuring and building OpenDDS.
...
+set PATH=%TOOLCHAIN%/bin;%PATH%
+set PATH=C:\MinGW\msys\1.0\bin;%PATH%
+make
+REM Pass -j/--jobs with an appropriate value or this'll take a while...
+```
+
+You can get rid of the second line if you're already in a Visual Studio
+"Developer Command Prompt" or adjust it if your Visual Studio is different
+version or configuration. See
+[here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/how-to-set-environment-variables-for-the-visual-studio-command-line)
```
mitza-oci: This is in the C# reference guide, is it relevant?  I've never used `VsDevCmd`.  The links that VS puts on the start menu use something like `...\VC\Auxiliary\Build\vcvarsx86_amd64.bat`

docs/android.md
```diff
@@ -233,48 +297,59 @@ the [ABI/architecture table](#abi-table)).
...
+   - Required to use RTPS Peer Discovery:
+     - `$DDS_ROOT/lib/libOpenDDS_Rtps.so`
+
+   - Required to use the Information Repository Peer Discovery:
+     - `$DDS_ROOT/lib/libOpenDDS_InfoRepoDiscovery.so`
+       - Depends on `$ACE_ROOT/lib/libTAO_PortableServer.so`
 
  - Required to use OpenDDS Security:
    - `$ACE_ROOT/lib/libACE_XML_Utils.so`
    - `$DDS_ROOT/lib/libOpenDDS_QOS_XML_XSC_Handler.so`
```
mitza-oci: QOS_XML_XSC is not required for OpenDDS Security

configure
```diff
@@ -154,8 +154,10 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'std=s', 'C++ standard version (for compilers that use -std)',
     'target=s', 'Cross-compile target (none): see --target-help',
     'target-compiler=s', 'Compiler for target (if req\'d): see --target-help',
-    'host-tools=s', 'DDS_ROOT of host tools for cross compile' .
-      $argIndent . '(build them)',
+    'host-tools=s', 'DDS_ROOT of host tools for cross compile (build' .
+      $argIndent . 'them)',
```
mitza-oci: probably don't need to split this line, it could just say '(build)'

configure
```diff
@@ -523,43 +526,45 @@ if (!exists $platforminfo{$opts{'target'}}->{'needs_i2jrt_corba'}) {
...
-  if (!defined $opts{'compiler'}) {
-    die "ERROR: Can't find a compiler, set PATH or run this script with the ".
-      "--compiler option.\n" . ($is_windows ? "  For Microsoft Visual C++, ".
-                                "run this script from the Visual Studio ".
-                                "Command Prompt.\n" : '') . "Stopped";
+  else {
+    print "Auto-detecting compiler\n" if $opts{'verbose'};
+    for my $stdcomp (@{$platforminfo{$opts{'host'}}->{'compilers'}}) {
+      my $path = which($stdcomp);
+      print "$stdcomp: $path\n";
```
mitza-oci: Verbose debug only?

configure
```diff
@@ -890,6 +897,19 @@ if (!$ace_src || !$tao_src) {
 print "Using ace_src: $ace_src\n" if $opts{'verbose'};
 print "Using tao_src: $tao_src\n" if $opts{'verbose'};
 
+if ($is_windows && $opts{'target'} eq 'android') {
+  # Rename the VERSION files in ACE, TAO, and OPENDDS. On Windows the cross
+  # compiler will pick them up instead of the version standard header the
+  # standard library is trying to include.
+  if (exists($opts{'dry-run'})) {
+    print("Would rename VERSION files to get around version header Windows issue.\n");
```
mitza-oci:
```suggestion
    print "Would rename VERSION files to get around version header Windows issue.\n";
```

docs/android.md
iguessthislldo: Yea, I had trouble wording this, making it explicit would be better.

configure
iguessthislldo: It was a debugging statement I added, I'll remove it.

configure
iguessthislldo: I got to say I really don't like this aspect of Perl.

iguessthislldo: I got to say I really don't like this aspect of Perl.

docs/android.md
iguessthislldo: I didn't even notice that. I think it's all the same for what we're doing because it worked for me.

configure
```diff
@@ -890,6 +897,19 @@ if (!$ace_src || !$tao_src) {
...
+  } else {
+    rename("$ace_src\\VERSION", "$ace_src\\VERSION.txt");
```
mitza-oci: Perl's stdlib is fine with `/` here, and we may end up using this on non-Windows platforms.  As `<version>` is in the process of becoming a C++ standard, and Windows isn't the only case-insensitive filesystem around, maybe we should go ahead and change these in the source repos.

iguessthislldo: Yes, I'm fine with that, as long as they don't add a `configure` standard header as well...

docs/android.md
iguessthislldo: > vitalization - the state of being vitalized and filled with life

No, I think that is correct (: 

I guess neovim's default spelling dictionary doesn't have "virtualization".

docs/android.md
mitza-oci: Let's avoid having additional details here compared to INSTALL and configure which just say "Visual Studio Command Prompt" should be used.  "Developer Command Prompt" is actually a different thing.  It defaults to 32-bit... probably a .NET thing?

iguessthislldo: I still don't really understand, but sure.

docs/android.md
iguessthislldo: Yes, that makes sense now. I don't think I knew what it was at the time, I probably just saw it was in the `lib` directory.

iguessthislldo:
I forgot to say that with the last set of commits and adding `export CROSS_COMPILE` to ACE in https://github.com/DOCGroup/ACE_TAO/pull/1022, I'm now able to build OpenDDS for Java.

mitza-oci:
> 
> 
> I forgot to say that with the last set of commits and adding `export CROSS_COMPILE` to ACE in [DOCGroup/ACE_TAO#1022](https://github.com/DOCGroup/ACE_TAO/pull/1022), I'm now able to build OpenDDS for Java.

Use of jni_check.pl is optional, so it may be easier for `configure` to just disable it using MPC feature `jni_check=0`.
The JNI functions aren't different across platforms, so as long as they're being "checked" somewhere there's little added benefit in checking them on cross-compiles.

mitza-oci:
Travis CI Java build: looks like JAVA_PLATFORM is not being set by `configure`

iguessthislldo:
> Use of jni_check.pl is optional, so it may be easier for configure to just disable it using MPC feature jni_check=0.

Alright, I will do that and take the export out of the ACE PR.

> Travis CI Java build: looks like JAVA_PLATFORM is not being set by configure

Yeah, I saw that. I think I messed up setting that variable.

iguessthislldo:
I will disable `jni_check`, but I think I will leave the export of the `CROSS_COMPILE` variable in the ACE/TAO PR, as it might be useful at some point to have the toolchain prefix.

PR #1448, opened 2020-01-15
Docker build: optional ACE source

adamsj-oci:
Use `ACE_CONFIG_OPTION` and `ACE_SOURCE_DIR` options to pick difference ACE source.

`ACE_CONFIG_OPTION` can be used to select `--ace-github-latest` instead of the default `--doc-group`.

`ACE_SOURCE_DIR` copies the correct files based on the ACE source to `/usr/local/share/ace/MPC`. Defaults to `/opt/OpenDDS/ACE_wrappers/MPC`. Use `/opt/OpenDDS/ACE_TAO/ACE/MPC` when using `--ace-github-latest` in `ACE_CONFIG_OPTION`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,9 +32,10 @@ RUN cd /opt/OpenDDS && \
 ENV ACE_ROOT=/usr/local/share/ace \
     TAO_ROOT=/usr/local/share/tao \
     DDS_ROOT=/usr/local/share/dds \
+    MPC_ROOT=/usr/local/share/MPC \
     PATH=".:/usr/local/share/ace/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
 WORKDIR /opt/OpenDDS/tests/DCPS/Messenger
-RUN . /opt/OpenDDS/setenv.sh && mwc.pl -type gnuace && make
+RUN mwc.pl -type gnuace && make
 
 WORKDIR /opt/workspace
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN cd /opt/OpenDDS && \
     ./configure --prefix=/usr/local --security --std=c++11 ${ACE_CONFIG_OPTION} && \
     make && \
     make install && \
-    cp -a ${MPC_ROOT} /usr/local/share/ace/MPC
+    cp -a ${MPC_ROOT} /usr/local/share/MPC
 
 ENV ACE_ROOT=/usr/local/share/ace \
     TAO_ROOT=/usr/local/share/tao \
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,11 +22,11 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
 ADD . /opt/OpenDDS
 
 ARG ACE_CONFIG_OPTION="--doc-group"
-ARG MPC_ROOT="/opt/OpenDDS/ACE_wrappers/MPC"
 RUN cd /opt/OpenDDS && \
     ./configure --prefix=/usr/local --security --std=c++11 ${ACE_CONFIG_OPTION} && \
     make && \
     make install && \
+    . /opt/OpenDDS/setenv.sh && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -20,11 +21,13 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
 
 ADD . /opt/OpenDDS
 
+ARG ACE_CONFIG_OPTION="--doc-group"
+ARG MPC_ROOT="/opt/OpenDDS/ACE_wrappers/MPC"
```
mitza-oci: This shouldn't be an `ARG` since the configure script knows it already.  After the configure script run, source the setenv.sh script that it generated.

Dockerfile
```diff
@@ -20,11 +21,13 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
...
 RUN cd /opt/OpenDDS && \
-    ./configure --prefix=/usr/local --security --doc-group --std=c++11 && \
+    ./configure --prefix=/usr/local --security --std=c++11 ${ACE_CONFIG_OPTION} && \
     make && \
     make install && \
-    cp -a /opt/OpenDDS/ACE_wrappers/MPC /usr/local/share/ace/MPC
+    cp -a ${MPC_ROOT} /usr/local/share/ace/MPC
```
mitza-oci:
```suggestion
    cp -a ${MPC_ROOT} /usr/local/share/MPC
```
See #1443

Dockerfile
adamsj-oci: It's an ARG so I can change it in the docker build for the OpenDDS/ROS2 version. As far as I know, there's no other way to influence the docker build externally.

mitza-oci: You don't need to do that for `MPC_ROOT`, which is what this comment is about.

adamsj-oci: The copy has a source directory of `/opt/OpenDDS/ACE_wrappers/MPC` for the `--doc-group option` and `/opt/OpenDDS/ACE_TAO/MPC` for the `--ace-github-latest` option. Are you saying sourceing `setenv.sh` with `--ace-github-latest` would remove the requirement for the copy?

mitza-oci: All it does is tell you where the MPC repository code is located.  It does this by setting the environment variable `MPC_ROOT`.  This is exactly the source directory you're looking for.

adamsj-oci: Got it. That makes more sense.

Dockerfile
```diff
@@ -20,18 +21,20 @@ RUN cmake CMakeLists.txt && make && cp ./*.a /usr/lib
...
+    cp -a ${MPC_ROOT} /usr/local/share/MPC
 
 ENV ACE_ROOT=/usr/local/share/ace \
     TAO_ROOT=/usr/local/share/tao \
     DDS_ROOT=/usr/local/share/dds \
     PATH=".:/usr/local/share/ace/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
 WORKDIR /opt/OpenDDS/tests/DCPS/Messenger
-RUN mwc.pl -type gnuace && make
+RUN . /opt/OpenDDS/setenv.sh && mwc.pl -type gnuace && make
```
mitza-oci: Does it work without this setenv?  Shouldn't be needed due to the docker `ENV` above.

adamsj-oci: I can't remember the specific error but it was looking for the MPC_ROOT so I used the setenv.sh in this run instruction.

adamsj-oci: Just reran. Here are the details:
Step 13/14 : RUN mwc.pl -type gnuace && make
 ---> Running in 168179d6a35e
ERROR: Unable to find the MPC modules in /usr/local/share/ace/MPC.
You can set the MPC_ROOT environment variable to the location of MPC.
The command '/bin/sh -c mwc.pl -type gnuace && make' returned a non-zero code: 255

mitza-oci: Yes, it needs MPC_ROOT, but you shouldn't use setenv.sh here since that's the source tree and not the installed tree.  You can update the ENV block above to have the correct MPC_ROOT

PR #1455, opened 2020-01-17
Adding Messenger test to installed DDS

kuznetsovmoci:


mitza-oci:
Error: ./bin/PerlDDS/Run_Test.pm:631: found trailing whitespace

bin/PerlDDS/Run_Test.pm
```diff
@@ -611,7 +618,19 @@ sub setup_discovery {
...
   $params = "" if !defined($params);
-  $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo" if !defined($executable);
+  if (!defined($executable)) {
+    $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo";
+    my $extension = ".exe";
+    my $executable_ext = $executable.$extension;
+    if (!(-e $executable) && !(-e $executable_ext)) {
+      if (!defined($ENV{OPENDDS_INSTALL_PREFIX})) {
+        print STDERR "ERROR: OPENDDS_INSTALL_PREFIX is not defined\n";
+        exit 1;
```
iguessthislldo: The only thing I should see that might really have to be improved is how to handle this error case. It should probably mention that it couldn't find `$DDS_ROOT/bin/DCPSInfoRepo`. Also I'm not entirely sure if we should be doing a `exit` here as libraries really shouldn't do that. A little further down on line 644 we are saying `$self->{status} = -1;`, but I don't know what that means exactly and I'm confused because it looks like we should be returning after that, be we don't.

bin/PerlDDS/Run_Test.pm
```diff
@@ -611,7 +618,19 @@ sub setup_discovery {
   my $params = shift;
   my $executable = shift;
   $params = "" if !defined($params);
-  $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo" if !defined($executable);
+  if (!defined($executable)) {
+    $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo";
+    my $extension = ".exe";
+    my $executable_ext = $executable.$extension;
+    if (!(-e $executable) && !(-e $executable_ext)) {
```
iguessthislldo:
```suggestion
    if (!(-e $executable) && !(-e "${executable}.exe")) {
```

A little simplification.

azure-pipelines.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       echo ##vso[task.setvariable variable=VcVarsArch]%VCVARS_ARCH%
     displayName: Set VcVarsArch
   - script: |
-      git clone -q --depth 1 git://github.com/Microsoft/vcpkg.git %VCPKG_ROOT%
+      git clone -q --depth 1 git://github.com/mitza-oci/vcpkg.git %VCPKG_ROOT% -b issue9654
```
mitza-oci: Upstream vcpkg has been fixed

bin/PerlDDS/Run_Test.pm
```diff
@@ -611,7 +618,18 @@ sub setup_discovery {
   my $params = shift;
   my $executable = shift;
   $params = "" if !defined($params);
-  $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo" if !defined($executable);
+  if (!defined($executable)) {
+    $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo";
+    if (!(-e $executable) && !(-e "${executable}.exe")) {
+      if (!defined($ENV{OPENDDS_INSTALL_PREFIX})) {
+        print STDERR "ERROR: OPENDDS_INSTALL_PREFIX is not defined. Couldn't find $DDS_ROOT/bin/DCPSInfoRepo\n";
```
iguessthislldo: `$` needs to be escaped. Also the wording could be clearer:

```suggestion
        print STDERR "ERROR: Couldn't find \$DDS_ROOT/bin/DCPSInfoRepo. It " .
          "needs to be built or \$OPENDDS_INSTALL_PREFIX needs to be defined " .
          "if OpenDDS is installed.\n";
```

azure-pipelines.yml
kuznetsovmoci: I reversed changes in the two scripts. This script used mitza-oci originally for some reason. I changed it to Microsoft too.

bin/PerlDDS/Run_Test.pm
```diff
@@ -611,7 +618,20 @@ sub setup_discovery {
...
-  $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo" if !defined($executable);
+  if (!defined($executable)) {
+    $executable = "$ENV{DDS_ROOT}/bin/DCPSInfoRepo";
+    if (!(-e $executable) && !(-e "${executable}.exe")) {
+      if (!defined($ENV{OPENDDS_INSTALL_PREFIX})) {
+        print STDERR "ERROR: Couldn't find \$DDS_ROOT/bin/DCPSInfoRepo. It " .
+          "needs to be built or \$OPENDDS_INSTALL_PREFIX needs to be defined " .
+          "if OpenDDS is installed.\n";
+        $self->{status} = -1;
+        return;
```
iguessthislldo: I know I referenced this in my other comment, but is this going to stop tests when this error is triggered?

bin/PerlDDS/Run_Test.pm
mitza-oci: I'm OK with an `exit` here since this is a library only built for running OpenDDS tests, and if we can't find OpenDDS then the overall Perl process has nothing more it can do.

I think the PR is general is just about ready to merge, but I want to make sure we have consensus on how `Run_Test.pm` should work first.

kuznetsovmoci: I actually have changed it yesterday, following Fred's suggestion, but can change it back of course. Where in the test process $self->{status} = -1; get processed, and what is the outcome?

PR #1458, opened 2020-01-17
Directed write test cleanup

lij-oci:
also added .gitignore


tests/transport/rtps_directed_write/.gitignore
```diff
@@ -1,2 +1,6 @@
-/publisher
-/subscriber
+publisher
```
mitza-oci: I don't think the .gitignore file should be changed by this PR, nothing changing in this PR impacts it.

PR #1460, opened 2020-01-17
Prevent early data pass-through for incomplete reliable rtps associations

simpsont-oci:


dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -1181,8 +1187,8 @@ operator<<(std::ostream& str, const DataLink& value)
        localId != value.assoc_by_local_.end();
        ++localId) {
     for (RepoIdSet::const_iterator
-         remoteId = localId->second.begin();
-         remoteId != localId->second.end();
+         remoteId = localId->second.associated_.begin();
```
mitza-oci: This isn't new to your commit, but this statement being split across 2 lines made it hard for me to understand at a glance

dds/DCPS/transport/framework/DataLink.h
```diff
@@ -332,6 +334,8 @@ class OpenDDS_Dcps_Export DataLink
 
   void network_change() const;
 
+  bool is_local_reliable(const RepoId& id) const;
```
mitza-oci: Is this called?  If not, can the `LocalAssociationInfo` be removed?

PR #1461, opened 2020-01-20
Fix deadlock causes by packet removal during reliable resend

simpsont-oci:
Primary issue was that packet removal was violating the datalink (e.g. writer) / send-strategy locking order requirement. Secondary issue included the fact that the code path for removal bounced back and forth between the strategy and datalink (including send buffer) several times.

This solution overrides datalink framework methods so they can be implemented in rtps writer directly while the lock is held, and manually handles retaining items in the send buffer to avoid the back & forth. Should also be slightly more efficient in terms of comparing match criteria / building TQE & SeqNum sets.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -148,10 +148,12 @@ RtpsUdpDataLink::add_delayed_notification(TransportQueueElement* element)
   return false;
 }
 
-void RtpsUdpDataLink::do_remove_sample(const RepoId& pub_id,
-  const TransportQueueElement::MatchCriteria& criteria)
+RemoveResult
+RtpsUdpDataLink::remove_sample(const DataSampleElement* sample)
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
+  RepoId pub_id = sample->get_pub_id();
+
+  ACE_Guard<ACE_Thread_Mutex> g(lock_);
   RtpsWriter_rch writer;
   RtpsWriterMap::iterator iter = writers_.find(pub_id);
   if (iter != writers_.end()) {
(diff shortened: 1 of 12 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -70,8 +70,8 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
 
   bool add_delayed_notification(TransportQueueElement* element);
 
-  void do_remove_sample(const RepoId& pub_id,
-                        const TransportQueueElement::MatchCriteria& criteria);
+  RemoveResult remove_sample(const DataSampleElement* sample) override;
```
mitza-oci:
```suggestion
  RemoveResult remove_sample(const DataSampleElement* sample);
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -70,8 +70,8 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
...
+  void remove_all_msgs(const RepoId& pub_id) override;
```
mitza-oci:
```suggestion
  void remove_all_msgs(const RepoId& pub_id);
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -227,7 +227,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
       , outer_(outer)
     {}
 
-    void retain_all(const RepoId& pub_id);
+    void retain_all(const RepoId&) {}
```
mitza-oci: Should the base class change to not be pure virtual?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -161,56 +163,107 @@ void RtpsUdpDataLink::do_remove_sample(const RepoId& pub_id,
...
+    tqe->data_dropped(true);
+    result = REMOVE_FOUND;
   }
+  return result;
+}
+
+void
+RtpsUdpDataLink::RtpsWriter::remove_all_msgs()
+{
+  SnToTqeMap sn_tqe_map;
```
mitza-oci: Is there a need for this to be default-constructed up front?  Could it just be copied from `elems_not_acked_` on 255?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -161,56 +163,107 @@ void RtpsUdpDataLink::do_remove_sample(const RepoId& pub_id,
...
-    OPENDDS_SET(SequenceNumber) sns_to_release;
     SnToTqeMap::iterator it = elems_not_acked_.begin();
-    while (it != elems_not_acked_.end()) {
-      if (criteria.matches(*it->second)) {
-        sn_tqe_map.insert(RtpsWriter::SnToTqeMap::value_type(it->first, it->second));
-        sns_to_release.insert(it->first);
-        SnToTqeMap::iterator last = it;
-        ++it;
-        elems_not_acked_.erase(last);
+    while (!found && it != elems_not_acked_.end()) {
```
mitza-oci: I see that `found` has to be kept for later, but I still wonder if this would be cleaner as a `for` loop with `break` avoiding the unwanted increment instead of needing an `else` block.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -161,56 +163,107 @@ void RtpsUdpDataLink::do_remove_sample(const RepoId& pub_id,
...
+
+  link->send_strategy()->remove_all_msgs(id_);
+
+  ACE_GUARD(ACE_Thread_Mutex, g2, elems_not_acked_mutex_);
+
+  sn_tqe_map.swap(elems_not_acked_);
+
+  g2.release();
+
+  SnToTqeMap::iterator release_iter = sn_tqe_map.begin();
```
mitza-oci: Another one that looks just like a for loop.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: The swap on 255 is intended to be a fast copy + clear of elems_not_acked_ and this is the recipient. I think we still need a valid (default constructed?) destination for the swap to work, unless there's a swap constructor or something I don't know about (move constructors?)

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: you're not wrong...

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: That's fine to keep a swap, those are useful.  I guess my comment is just to move the declaration down to where it's needed.

PR #1463, opened 2020-01-21
[3.13] Handling multiple transport instances

neeleym:
Removed break from active processing side to allow continuance of loop to process candidate links when the first connection link fails to succeed. The loop was kicking out too early. 

On the passive side of association, for loops that have multiple impls to look at for candidate association was kicking out after the first indication of a successful impl existance, but when the link was null, it did not proceed to the next one in the list to check further, but was simply returning out of the associate method.


mitza-oci:
Related PR for master is #1465, we'll merge this one after that one

PR #1467, opened 2020-01-22
Sedp: transport may still have a reference to Writers during shutdown

mitza-oci:


dds/DCPS/RTPS/Sedp.h
```diff
@@ -307,9 +308,12 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
     }
 #endif
 
+    void shutting_down() { shutting_down_ = true; }
+
   protected:
     DCPS::RepoId repo_id_;
     Sedp& sedp_;
+    ACE_Atomic_Op<ACE_SYNCH_MUTEX, bool> shutting_down_;
```
simpsont-oci: I recognize you probably copied this from below... but does ACE_SYNCH_MUTEX buy us anything here over ACE_Thread_Μutex? The ability to compile when mutexes aren't supported? The ability to make this a NOOP whenever we want to make our ACE_Atomic_Op unatomic?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -259,34 +259,34 @@ const bool Sedp::host_is_bigendian_(!ACE_CDR_BYTE_ORDER);
 Sedp::Sedp(const RepoId& participant_id, Spdp& owner, ACE_Thread_Mutex& lock) :
   DCPS::EndpointManager<ParticipantData_t>(participant_id, lock),
   spdp_(owner),
-  publications_writer_(
-    make_id(participant_id, ENTITYID_SEDP_BUILTIN_PUBLICATIONS_WRITER), *this),
+  publications_writer_(make_rch<Writer>(
```
simpsont-oci: Odd that the reader was already using RCH but not the writer, though perhaps for similar reasons?

dds/DCPS/RTPS/Sedp.h
mitza-oci: Yes, it was just a refactor from derived to base.  We can change it.

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: The reader is more fundamentally driven by callbacks from the transport.  The writer has a few too -- as we've seen here.  On the writer side, even data_delivered can be done without RCH as long as remove_all_msgs is used (see other PR).

mitza-oci: Those 4 security attributes tests passed on my local build.

PR #1468, opened 2020-01-22
Secure Participant Discovery modification: First draft for review

kuznetsovmoci:


dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -282,6 +282,20 @@ namespace {
 
     return OpenDDS::Security::DPDK_ORIGINAL;
   }
+
+  bool has_identity_status_token(const ParameterList& param_list)
+  {
+    CORBA::ULong length = param_list.length();
+    for (CORBA::ULong i = 0; i < length; ++i) {
+      const Parameter& param = param_list[i];
+      switch (param._d()) {
```
mitza-oci: Looks like it should be an `if` (also don't need a separately named `param` object)

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -613,13 +613,17 @@ Spdp::handle_participant_data(DCPS::MessageId id,
 #ifdef OPENDDS_SECURITY
     // Non-secure updates for authenticated participants are used for liveliness but
     // are otherwise ignored. Non-secure dispose messages are ignored completely.
-    if (iter->second.auth_state_ == DCPS::AS_AUTHENTICATED &&
-        pdata.dataKind != Security::DPDK_SECURE &&
-        id != DCPS::DISPOSE_INSTANCE &&
-        id != DCPS::DISPOSE_UNREGISTER_INSTANCE) {
+    if (!(iter->second.auth_state_ == DCPS::AS_AUTHENTICATED && pdata.hasIdentityStatusToken)) {
```
mitza-oci: The conditional can't be based on the presence of an parameter which is optional in the message.

mitza-oci:
Replaced by #1531

PR #1471, opened 2020-01-23
Call remove_all_msgs from SEDP transport client destructors to avoid leaks

simpsont-oci:


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2874,6 +2874,7 @@ Sedp::Writer::Writer(const RepoId& pub_id, Sedp& sedp)
 
 Sedp::Writer::~Writer()
 {
+  remove_all_msgs();
```
mitza-oci: Move to Endpoint's destructor?

PR #1474, opened 2020-01-24
Fix Bug in TransportSendBuffer's Fragment Resend Logic

simpsont-oci:


dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -377,24 +377,34 @@ SingleSendBuffer::resend_fragments_i(const SequenceNumber& seq,
...
-        sent = it->first;
-      }
-      if (it == it2) {
-        break;
+  const FragmentMap::const_iterator fm_it = fragments_.find(seq);
+  if (fm_it == fragments_.end()) {
+    return;
+  }
+  const BufferMap& buffers = fm_it->second;
+  const OPENDDS_VECTOR(SequenceRange)& psr = requested_frags.present_sequence_ranges();
```
mitza-oci: I'm not sure making this a reference actually helps anything, but it also won't hurt anything.

dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -377,24 +377,34 @@ SingleSendBuffer::resend_fragments_i(const SequenceNumber& seq,
...
+
+  SequenceNumber frag_min;
+  size_t i = 0;
```
mitza-oci: Prefer defining these two as close to their usage as possible.

dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -377,24 +377,34 @@ SingleSendBuffer::resend_fragments_i(const SequenceNumber& seq,
...
+  BufferMap::const_iterator it = buffers.lower_bound(psr.front().first);
+  BufferMap::const_iterator end = buffers.lower_bound(psr.back().second);
+  if (end != buffers.end()) {
+    ++end;
+  }
+
+  // Iterate over both containers simultaneously
+  while (i < psr.size() && it != end) {
+    if (psr[i].second < frag_min) {
+      // While the range is too low, increment range
```
mitza-oci: I think this is clear without the comment.  If you do want to keep the comment, consider revising it since it's not incrementing the range, but advancing to the next range in the collection

dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -377,24 +377,34 @@ SingleSendBuffer::resend_fragments_i(const SequenceNumber& seq,
...
+  size_t i = 0;
+  BufferMap::const_iterator it = buffers.lower_bound(psr.front().first);
+  BufferMap::const_iterator end = buffers.lower_bound(psr.back().second);
+  if (end != buffers.end()) {
+    ++end;
+  }
+
+  // Iterate over both containers simultaneously
+  while (i < psr.size() && it != end) {
+    if (psr[i].second < frag_min) {
```
mitza-oci: Does `frag_min` need to start at 0 or is it ok at 1?

simpsont-oci: Spec says minimum fragment number is 1, and we want this to be the minimum fragment number of the current fragment buffer (it's inserted into map by the maximum fragment number), so we want it to start at 1.

PR #1478, opened 2020-01-25
Participant location participant key

jrw972:


dds/DCPS/DiscoveryBase.h
```diff
@@ -1635,6 +1635,16 @@ namespace OpenDDS {
         }
 
         DiscoveredParticipantData pdata_;
+        struct LocationUpdate {
+          OpenDDS::DCPS::ParticipantLocation mask_;
```
mitza-oci:
```suggestion
          ParticipantLocation mask_;
```
It's already in the namespace.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -396,91 +396,102 @@ namespace {
...
     return;
   }
 
-  OpenDDS::DCPS::ParticipantLocationBuiltinTopicData& ld = iter->second.location_data_;
+  const RepoId guid = iter->first;
 
-  OPENDDS_STRING addr = "";
-  ACE_TCHAR buffer[256];
+  DiscoveredParticipant::LocationUpdateList location_updates = iter->second.location_updates_;
+  iter->second.location_updates_.clear();
```
mitza-oci: This looks like it could be a `swap` (with previous line), should be more efficient.

dds/DdsDcpsCore.idl
```diff
@@ -447,6 +447,7 @@ module OpenDDS {
     BUILT_IN_TOPIC_TYPE
     struct ParticipantLocationBuiltinTopicData {
             BUILT_IN_TOPIC_KEY OctetGuidArray16 guid;
+            DDS::BuiltinTopicKey_t participant_key;
```
mitza-oci: As we discussed offline, I think we can avoid adding this.  Let me know if you need more info.

PR #1485, opened 2020-01-27
Improve RTPS Performance By Reducing Lock Contention / Improving Efficiency for Removing Sample

simpsont-oci:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -427,7 +427,10 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
   /// lock_ protects data structures accessed by both the transport's thread
   /// (TransportReactorTask) and an external thread which is responsible
   /// for adding/removing associations from the DataLink.
-  mutable ACE_Thread_Mutex lock_;
+  mutable ACE_Thread_Mutex misc_lock_;
```
mitza-oci: Update comment block on lines directly above this.

simpsont-oci: On second thought, I'm removing misc_lock_ and reverting that lock order change I made... best_effort_heartbeat_count_ was the only thing under misc_lock_ and it just seems to make sense to move it to writers_lock_.

simpsont-oci: Also cleaned up some this-> stuff, had to rename a few variables.

PR #1486, opened 2020-01-28
Bug fix for executable existence check in tests

kuznetsovmoci:
Tested on OCI1444


mitza-oci:
I'm cancelling some of the CI builds since they're not testing this part anyway

PR #1488, opened 2020-01-29
Reduce InstanceState's Contention for ReactorInterceptor Mutex

simpsont-oci:


dds/DCPS/ReactorInterceptor.cpp
```diff
@@ -51,6 +51,7 @@ void ReactorInterceptor::process_command_queue_i()
     ACE_GUARD(ACE_Reverse_Lock<ACE_Thread_Mutex>, rev_guard, rev_lock);
     command->execute();
     command->executed();
+    command.reset();
```
simpsont-oci: This force the call to delete to be called when the mutex lock is unlocked (previously, scoping would relock the lock and _then_ delete the command pointer.

PR #1489, opened 2020-01-29
ReliableBestEffortReaders test

lij-oci:
ReliableBestEffortReaders tests whether messages are received correctly by both reliable and best-effort readers associated to the same writer.


tests/DCPS/ReliableBestEffortReaders/.gitignore
```diff
@@ -0,0 +1,9 @@
+publisher
+subscriber
+GNUmakefile*
```
mitza-oci: Remove the hand-written .gitignore and use the script to generate one.

tests/DCPS/ReliableBestEffortReaders/run_test.pl
```diff
@@ -0,0 +1,42 @@
...
+# -*- perl -*-
+
+use lib "$ENV{ACE_ROOT}/bin";
+use lib "$ENV{DDS_ROOT}/bin";
+use PerlDDS::Run_Test;
+use strict;
+
+sub do_test {
+  my $reliable = shift;
+  my $port = PerlACE::random_port();
```
mitza-oci: This appears to be unused.

tests/DCPS/ReliableBestEffortReaders/TestMsg.idl
```diff
@@ -0,0 +1,7 @@
+@topic
```
mitza-oci: Does this test need its own IDL library?  Or can it use an existing one?

bin/dcps_tests.lst
```diff
@@ -144,6 +144,7 @@ tests/DCPS/Reliability/run_test.pl: !DCPS_MIN !DDS_NO_OWNERSHIP_PROFILE !OPENDDS
 tests/DCPS/Reliability/run_test.pl rtps: !DCPS_MIN !DDS_NO_OWNERSHIP_PROFILE
 tests/DCPS/Reliability/run_test.pl keep-last-one: !DCPS_MIN !OPENDDS_SAFETY_PROFILE
 tests/DCPS/Reliability/run_test.pl rtps keep-last-one: !DCPS_MIN
+tests/DCPS/ReliableBestEffortReaders/run_test.pl: !DCPS_MIN
```
mitza-oci:
```suggestion
tests/DCPS/ReliableBestEffortReaders/run_test.pl: RTPS !DCPS_MIN
```

tests/DCPS/ReliableBestEffortReaders/TestMsg.idl
lij-oci: I searched the idl's in all the DCPS tests. Two close ones are Partition/Test.idl and MultiDiscovery/TestMsg.idl. They have a long and a string members, but I would like one with a seq# separated from the key field, in case we need to setup different testing scenarios and still be able to check seq# separately.

mitza-oci: How about `tests/DCPS/ConsolidatedMessengerIdl/Messenger.idl`?

lij-oci: Replaced TestMsg.idl with the shared ConsolidatedMessengerIdl/Messenger.idl. That's pretty cool. I'll check in the code shortly. I think we should use that shared idl in the transport level tests, too. Or create something like that.

lij-oci:
Now the only failed test is this test itself. That means it successfully identified the problem that when both reliable and best-effort readers are associated with the same writer, the best-effort readers will not get the messages.

PR #1492, opened 2020-01-29
[3.13] Adding TCP transport active connection timeout option

neeleym:
Added tcp connection timeout for active tcp connection attempts.
When unroutable locations were being found in the OS interfaces
the connection attempt was hanging for an indefinite amount of time.
However, this was not configureable on the active side. Configurable
option (active_conn_timeout_period) was added as a TCP-specific option
with a default of 5 seconds.


neeleym:
Replaced by #1495 ... closing.

PR #1495, opened 2020-01-29
[3.13] Adding TCP connection timeout for active tcp connection attempts

neeleym:
Added TCP connection timeout for active tcp connection attempts.
When unreachable locations were being found in the OS interfaces
the connection attempt was hanging for an indefinite amount of time.
However, this was not configurable on the active side. Configurable
option (active_conn_timeout_period) was added as a TCP-specific option
with a default of 5 seconds.


mitza-oci:
Link to PR for master #1493

PR #1496, opened 2020-01-29
Guard against void RCH dereferences in SPDP transport destructor

simpsont-oci:


jrw972:
I'm assuming you found a code path where close is called without open (or a partial open).  Can you give me some details on the code path?

simpsont-oci:
I had several core files for this from a single test run in the test environment. I don't have logs and have since deleted the core files, but this was coming from a "normal" shutdown that had been triggered right away, probably due to some system-related error condition that impacted multiple processes on startup. Ultimately, we'll want to fix / handle whatever the error condition was, but in the mean time this should hopefully prevent the segfault.

Issue #570, opened 2017-05-26
QoS_XML results in linking of xercesc with the user code


jwillemsen:
At the moment an application uses the QoS_XML support (as for example the test tests/DCPS/QoS_XML/dump test) there shouldn't be xercesc in the linking of the application. The usage of xercesc should be hidden within the QoS_XML library and shouldn't be exposed to any user code.

mitza-oci:
Wouldn't that break static builds?

jwillemsen:
Yes, if done incorrectly it could. The usage of xercesc should really an internal item, when using shared libraries our dds4ccm connectors shouldn't need to link with xercesc because opendds uses it internally.

mitza-oci:
Can you be more specific than "if done incorrectly"?  I guess the configuration would have to be conditional on static/shared.  Is there an example of this pattern somewhere else within ATCD?

ocielliottc:
There is a feature, static_libs_only, that is set when the -static option is used.  You can define a feature project based on this that can be used to specify different settings when the -static option is used.

ex.,

feature(static_libs_only) {
  libs += foo
}

jwillemsen:
Ok, something to try at some point, thanks @ocielliottc

mitza-oci:
If we use this, also make sure that the configure script uses `-static` with gnuace.  The current version doesn't since the generated makefiles can be used for static or shared.

PR #1503, opened 2020-01-30
Keep Linkage Private for OpenSSL on Android

iguessthislldo:
This is pulled from https://github.com/objectcomputing/OpenDDS/pull/1066. I thought it was fixed by https://github.com/DOCGroup/ACE_TAO/pull/858, but I didn't make sure OpenDDS was using `ace_openssl` instead of `openssl` directly. Now I see `ace_openssl` is broken in OpenDDS on top of the fact that the issue never was fixed. This applies the fix from #1066 to get around `ace_openssl`. Will have to create some sort of test for this at some point if we ever get Android testing.

### DIFF
--- a/MPC/config/dds_openssl.mpb
+++ b/MPC/config/dds_openssl.mpb
@@ -1,4 +1,16 @@
 feature(openssl): openssl {
+  // Android preloads the system SSL library (either OpenSSL or BoringSSL) for
+  // the Java Android API, so we must:
+  //   1. Statically link OpenSSL to this library
+  //   2. Keep our OpenSSL symbols internal
+  // Number 1 is described in android.md. We are doing number 2 here.
+  verbatim (gnuace, top) {
+    comma := ,
+    exclude_openssl_libs := -Wl$(comma)--exclude-libs$(comma)libcrypto.a -Wl$(comma)--exclude-libs$(comma)libssl.a
+  }
+  specific (gnuace) {
+    compile_flags += $(if $(ANDROID_ABI),$(exclude_openssl_libs))
+  }
 }
 
 feature(!no_vxworks_openssl) {
--- a/MPC/config/dds_openssl.mpb
+++ b/MPC/config/dds_openssl.mpb
@@ -4,7 +4,7 @@ feature(openssl): openssl {
   //   1. Statically link OpenSSL to this library
   //   2. Keep our OpenSSL symbols internal
   // Number 1 is described in android.md. We are doing number 2 here.
-  verbatim (gnuace, top) {
+  verbatim (gnuace, top, 1) {
     comma := ,
     exclude_openssl_libs := -Wl$(comma)--exclude-libs$(comma)libcrypto.a -Wl$(comma)--exclude-libs$(comma)libssl.a
   }


MPC/config/dds_openssl.mpb
```diff
@@ -1,4 +1,16 @@
 feature(openssl): openssl {
+  // Android preloads the system SSL library (either OpenSSL or BoringSSL) for
+  // the Java Android API, so we must:
+  //   1. Statically link OpenSSL to this library
+  //   2. Keep our OpenSSL symbols internal
+  // Number 1 is described in android.md. We are doing number 2 here.
+  verbatim (gnuace, top) {
+    comma := ,
+    exclude_openssl_libs := -Wl$(comma)--exclude-libs$(comma)libcrypto.a -Wl$(comma)--exclude-libs$(comma)libssl.a
```
mitza-oci: Does it really need `$(comma)`?  Would it be different if this was `=` instead of `:=`?

MPC/config/dds_openssl.mpb
```diff
@@ -1,4 +1,16 @@
 feature(openssl): openssl {
+  // Android preloads the system SSL library (either OpenSSL or BoringSSL) for
+  // the Java Android API, so we must:
+  //   1. Statically link OpenSSL to this library
+  //   2. Keep our OpenSSL symbols internal
+  // Number 1 is described in android.md. We are doing number 2 here.
+  verbatim (gnuace, top) {
```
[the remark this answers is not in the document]
iguessthislldo: What does this do?

MPC/config/dds_openssl.mpb
iguessthislldo: > Does it really need $(comma)?

I copied this as is from #1066. I remember I spent quite a bit of time getting this working which led to the comma variable, I don't remember why it's needed though.

>  Would it be different if this was = instead of :=?

I try to prefer `:=` to `=`, but maybe.

iguessthislldo:
If I'm going to spend more time on this I think I would be worth it to make a test for this, sorta like `jni_check.pl`.

MPC/config/dds_openssl.mpb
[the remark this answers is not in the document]
jwillemsen: See https://github.com/DOCGroup/MPC/blob/cb7bb9393707341beac8908db9c600080cc85fa3/docs/README#L289, it combines the verbatims of multiple projects, something that can happen when you integrate OpenDDS into AXCIOMA

MPC/config/dds_openssl.mpb
mitza-oci: It's fine as-is, just wondering what was going on there.  Seems like it has to "hide" the comma from the `$(if`.

MPC/config/dds_openssl.mpb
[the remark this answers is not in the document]
iguessthislldo: It could be expressed better, something like `append` instead of just `1`...

MPC/config/dds_openssl.mpb
iguessthislldo: Yeah that sounds familiar, so that's probably it.

jwillemsen:
Would me nice when MPC would help with this, with xercesc we have the same, we have to link with xercesc for the XML parsing library but user code doesn't need to link with xercesc at all, see https://github.com/objectcomputing/OpenDDS/issues/570

iguessthislldo:
I did a build with this and it's giving me this for compiling objects:

```
clang90++: warning: -Wl,--exclude-libs,libcrypto.a: 'linker' input unused [-Wunused-command-line-argument]
clang90++: warning: -Wl,--exclude-libs,libssl.a: 'linker' input unused [-Wunused-command-line-argument]
```

So maybe I should change `compile_flags` to linker flags.

Also is there a way to tell MPC to only link a particular library statically?

PR #1507, opened 2020-02-01
Testing [skip ci]

iguessthislldo:
Seeing what CI honors [skip ci]. Travis should, not sure about others.
I'm certain the PR test runner won't.


iguessthislldo:
I forgot we only have Travis and Azure now, so the "others" is just Azure. It doesn't look like Azure worked though. From a quick google it looked like it should, but doing some more reading it looks like it doesn't work with PRs for some reason.

iguessthislldo:
[This is the issue I'm looking at](https://github.com/Microsoft/azure-pipelines-agent/issues/858) if anyone's interested.

PR #1508, opened 2020-02-02
idl2jni: fixed relative path to Version.h

mitza-oci:


iguessthislldo:
One level too many? I'm surprised I didn't catch that earlier. I swear I did a java build on my machine for #1440.

PR #1510, opened 2020-02-03
reverted jni_check.pl change from PR #1446

mitza-oci:


iguessthislldo:
Is this causing problems?

mitza-oci:
> 
> 
> Is this causing problems?

Yes, see scoreboard

mitza-oci:
> 
> 
> I would have liked the script to be able to work with cross compiled Java, but I guess it's not that important.

It's not all that important since the same checks are done in a non-cross build.  But that doesn't mean we won't do it, just not right now.

PR #1512, opened 2020-02-03
Fix memory leaks in ParticipantLocationTopic test

jrw972:


tests/DCPS/ParticipantLocationTopic/ParticipantLocationListenerImpl.h
```diff
@@ -24,7 +24,7 @@ class ParticipantLocationListenerImpl
 {
 public:
   //Constructor
-  ParticipantLocationListenerImpl(const char* id);
+  ParticipantLocationListenerImpl(const std::string& id);
```
mitza-oci:
```suggestion
  explicit ParticipantLocationListenerImpl(const std::string& id);
```

PR #1518, opened 2020-02-04
rtps_udp 2.4 interop fixes

mitza-oci:
- allow heartbeats with the "first" number one larger than the "last"
- for volatile durability, first received data is delivered

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1325,20 +1325,13 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
     seq.setValue(data.writerSN.high, data.writerSN.low);
 
     info.frags_.erase(seq);
-    if (info.recvd_.empty()) {
-      // This is the first data / heartbeat we've seen
-      // Initialize the reader so we can potentially deliver data
-      if (durable_) {
-        info.hb_range_.first = 1;
-        info.hb_range_.second = seq;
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1393,7 +1383,8 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
       info.held_.insert(std::make_pair(seq, *sample));
       info.recvd_.insert(seq);
       link->deliver_held_data(id_, info, durable_);
-    } else if (info.recvd_.disjoint() || info.recvd_.cumulative_ack() != seq.previous()) {
+    } else if (info.recvd_.disjoint() ||
+               (!info.first_delivered_data_ && info.recvd_.cumulative_ack() != seq.previous())) {
```
simpsont-oci: I don't think you want the check for info.first_delivered_data_ to be down this far. I'd put it back up alongside the check for recvd_.empty(). The rest of this logic should work correctly for both durable and non-durable if we've configured recvd_ correctly by the time the code gets here.

mitza-oci: OK, I will push a revised version.  Keep in mind that I'm working on the case where `process_heartbeat_i` has already inserted into `info.recvd_`.  The failed tests on GHPRB are passing on my local build with this update.

mitza-oci:
The tests failing on GHPRB are not failing for my locally, not sure what's going on there.

mitza-oci:
> 
> 
> The tests failing on GHPRB are not failing for me locally, not sure what's going on there.

Tests passed this time.

PR #1523, opened 2020-02-05
Fix asynchronous connect error checking on Windows

huangminghuang:


dds/DCPS/transport/tcp/TcpConnection.cpp
```diff
@@ -584,7 +584,7 @@ OpenDDS::DCPS::TcpConnection::active_reconnect_i()
 
     TcpConnection* pconn = this;
     int ret = transport.connector_.connect(pconn, this->remote_address_,  ACE_Synch_Options::asynch);
-    if (ret == -1 && errno != EWOULDBLOCK) {
+    if (ret == -1 && errno != EWOULDBLOCK && errno != EINPROGRESS) {
```
mitza-oci: Does this also need to change in `TcpTransport::connect_datalink`?

huangminghuang: I did this change based on the error log from the scoreboard. I don't have windows machine to test it. However, I don't quite understand why the code in TcpTransport::connect_datalink() work as it is while I need to check EINPROGRESS for reconnect.  Maybe it's because I reused the TcpTransport::connector_ object to reconnect and the behavior of the ACE_Connect is not consistent when you use the same connect object to invoke asynchronous connect in the second time.  As this moment, I won't change TcpTransport the same way simply because there is no tests showing errors there.

mitza-oci:
Some of the CI build results may not be accurately reflecting what's being merged since they have master at 67464262b13aed41a7f60fea76b086dc53e70081, which was pushed by mistake

PR #1528, opened 2020-02-10
Upgrade to TAO 2.5.8

jwillemsen:


iguessthislldo:
Should we also update the DOC Group version referenced in the README?

iguessthislldo:
Another thing is that the README says you can download from either vanderbilt.edu or github, but it looks like vanderbilt.edu is pointing to github now, so should we just remove vanderbilt.edu?

jwillemsen:
Downloads are on both places, vanderbilt and github, only github has not all releases made in the past

PR #1530, opened 2020-02-11
RTPS 2.4 - Switch to smaller SN sets in AckNack / Gap

wilsongOCI:
Allows SN sets to be of zero size based on the changes in the RTPS 2.4 (2.3) specification.


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1828,9 +1828,9 @@ RtpsUdpDataLink::RtpsReader::gather_ack_nacks_i(MetaSubmessageVec& meta_submessa
       wi->second.ack_pending_ = false;
 
       SequenceNumber ack;
-      CORBA::ULong num_bits = 1;
+      CORBA::ULong num_bits = 0;
       LongSeq8 bitmap;
-      bitmap.length(1);
+      bitmap.length(0);
```
mitza-oci: This line and the line below should be removed.  The default-constructed sequence already has length 0, and indexing position 0 is undefined behavior.

wilsongOCI: Update completed

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2422,9 +2421,9 @@ RtpsUdpDataLink::RtpsWriter::gather_gaps_i(const RepoId& reader,
     bitmap.length(bitmap_num_longs(base, gaps.high()));
     (void) gaps.to_bitmap(bitmap.get_buffer(), bitmap.length(), num_bits);
   } else {
-    bitmap.length(1);
+    bitmap.length(0);
```
mitza-oci: Same issue with this one.

wilsongOCI: Also, I am going to remove the  num_bits = 0 since it is initialized to 0 when it is declared.

mitza-oci:
CI tests are failing on this branch.  We will need to fix (or at least understand why) before merging.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2422,9 +2421,7 @@ RtpsUdpDataLink::RtpsWriter::gather_gaps_i(const RepoId& reader,
     bitmap.length(bitmap_num_longs(base, gaps.high()));
     (void) gaps.to_bitmap(bitmap.get_buffer(), bitmap.length(), num_bits);
   } else {
-    bitmap.length(1);
     bitmap[0] = 0;
```
simpsont-oci: Now that the length is zero (default) we can't index the first position (zero) or we'll access invalid memory. Same for linke 1833.

PR #1532, opened 2020-02-12
Azure Pipelines: Adding Messenger tests for all Visual Studio builds

kuznetsovmoci:


mitza-oci:
Since vcpkg broke VS2015 (again), should we make that a non-security build?

jwillemsen:
Be aware that vs2015 support on azure will be removed coming March, see https://github.com/DOCGroup/ACE_TAO/issues/1033

mitza-oci:
Script step to generate sln files (make this conditional so it only runs for Release)
```
mwc.pl -type [vs2019|other] tests\DCPS\Messenger DevGuideExamples\DCPS\Messenger
```
Step to build: use VSBuild task pointing to the new sln files in the `solution:` argument

mitza-oci:
In order to make the VS2015 build pass, remove the vcpkg steps from that section.
Also change the VS2017 section to be the same as VS2019.

mitza-oci:
I'm cancelling this run on Azure Pipelines so that other PRs have time to build.  We will need further changes here anyway.  As I was saying before, I recommend "prototyping" these changes in your own forked repo until they're ready.

PR #1535, opened 2020-02-12
Remove delete / release from within SingleSendBuffer mutex locking / RtpsWriter mutex locking

simpsont-oci:
…


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -222,13 +223,18 @@ RtpsUdpDataLink::RtpsWriter::remove_sample(const DataSampleElement* sample)
...
 
   g.release();
 
   if (found) {
     tqe->data_dropped(true);
+    for (size_t i = 0; i < removed.size(); ++i) {
+      delete removed[i].first;
+      removed[i].second->release();
+    }
+    removed.clear();
```
mitza-oci: Do we need this line?

PR #1542, opened 2020-02-17
Add Support For Building Reliable Best Effort Readers Test Statically

lij-oci:


tests/DCPS/ReliableBestEffortReaders/publisher.cpp
```diff
@@ -5,6 +5,11 @@
 #include <iostream>
 #include <assert.h>
 
+#ifdef ACE_AS_STATIC_LIBS
+# include "dds/DCPS/RTPS/RtpsDiscovery.h"
+# include "dds/DCPS/transport/rtps_udp/RtpsUdp.h"
```
iguessthislldo: A couple of minor things from the guidelines that also apply to the subscriber:

- Indents should be two spaces.
- Use `<>` if the path isn't relative to the current file.
- OpenDDS includes should be grouped together and before standard library includes.

PR #1554, opened 2020-02-20
Unused argument handling in enqueue()

kuznetsovmoci:


iguessthislldo:
I think this is intended. If security isn't enabled it's going to warn about `bSecureParticipant` being unused.

mitza-oci:
> 
> 
> I think this is intended. If security isn't enabled it's going to warn about `bSecureParticipant` being unused.

Yes, that is why we have ACE_UNUSED_ARG.

iguessthislldo:
> Yes, that is why we have ACE_UNUSED_ARG

I read this completely wrong. For some reason I thought the line was being deleted instead of added...

PR #1555, opened 2020-02-20
Fix Mutex Contention Within ReactorInterceptor

simpsont-oci:
The primary motivator for this was mutrace's output. When running a mutrace'd worker alongside a bench scenario, the ReactorInterceptor show up in most of the top contention slots, which I'm guessing comes mostly from holding the lock while calling notify when the reactor thread is idle (and so it immediately tries to grab the lock). The addition of the state flag and the queue swapping was intended to prevent some of the back & forth.

### DIFF
--- a/dds/DCPS/ReactorInterceptor.cpp
+++ b/dds/DCPS/ReactorInterceptor.cpp
@@ -21,6 +21,7 @@ namespace DCPS {
 ReactorInterceptor::ReactorInterceptor(ACE_Reactor* reactor,
                                        ACE_thread_t owner)
   : owner_(owner)
+  , state_(NONE)
 {
   this->reactor(reactor);
 }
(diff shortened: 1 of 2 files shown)


dds/DCPS/ReactorInterceptor.cpp
```diff
@@ -43,16 +44,22 @@ int ReactorInterceptor::handle_exception(ACE_HANDLE /*fd*/)
...
+    ACE_Guard<ACE_Reverse_Lock<ACE_Thread_Mutex> > rev_guard(rev_lock);
+    while (!cq.empty()) {
+      CommandPtr command = cq.front();
+      cq.pop();
+      command->execute();
+      command->executed();
+    }
   }
+  state_ = NONE;
 }
```
jrw972: I believe this is fundamentally flawed.  At one point, there was logic that measure the size of the queue at the beginning and only executed that many commands.  This was to prevent potential live lock where a command repeatedly puts itself on the queue.  I would recommend doing the swap first, process the queue, and then check the queue again after the loop to see if it needs to notify itself again.

simpsont-oci: That's a good consideration, but I think it's beyond the scope of this PR. You're arguing for a design change of the current ReactorInterceptor (to avoid starvation of the non-ReactorInterceptor reactor tasks), but the change in this PR intentionally preserves the current command execution behavior (run until the member queue is empty) while preventing unnecessary calls to notify by using a state flag and using queue::swap (which is evidently C++11, so I guess we switch to dequeue) in order to lock / contend the mutex less frequently.

PR #1558, opened 2020-02-22
Use "-dev" version suffix to indicate development versions (master branch between releases)

iguessthislldo:
- New version scheme where between releases OpenDDS master will have a version consisting of the next planned major or minor version and `-dev` (`3.14.0-dev` in the current case).
- Release Script Changes:
  - Modify for new version scheme, allowing next version to be overridden.
  - Use `Getopt::Long`.
  - Turn `STEPS` argument into an option because of issues with the step expressions and `Getopt::Long`
  - Release date inserted into files is now shorter: Feb 22 2020, for example.
  - Regularize step names.
- Added `OPENDDS_IS_RELEASE` to `dds/Version.h` which is defined as 1 in releases, 0 all other times.
- Update `AUTHORS` file.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # OpenDDS Releases
 
-## Version 3.14 of OpenDDS
-Released %TIMESTAMP%
+## Version 3.14.0 of OpenDDS
+OpenDDS 3.14.0 is currently in development, so this list might change.
 
 ### Additions:
 - Union topic types and IDL annotations for topic types (#1067):
@@ -20,8 +20,8 @@ Released %TIMESTAMP%
 - Java API can now be used on Android
 
 ### Notes:
-- Removed outdated `-Gws` option of `opendds_idl`. This option produced sample
-  descriptions for the Wireshark dissector before OpenDDS v3.7.
+- Removed outdated `-Gws` option of `opendds_idl`. This option produced topic
+  type descriptions for the Wireshark dissector before OpenDDS v3.7.
 
 ## Version 3.13.1 of OpenDDS
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # OpenDDS Releases
 
-## Version 3.14.0 of OpenDDS
-OpenDDS 3.14.0 is currently in development, so this list might change.
+## Version 3.14 of OpenDDS
+OpenDDS 3.14 is currently in development, so this list might change.
 
 ### Additions:
 - Union topic types and IDL annotations for topic types (#1067):
@@ -20,8 +20,8 @@ OpenDDS 3.14.0 is currently in development, so this list might change.
 - Java API can now be used on Android
 
 ### Notes:
-- Removed outdated `-Gws` option of `opendds_idl`. This option produced topic
-  type descriptions for the Wireshark dissector before OpenDDS v3.7.
+- Removed outdated `-Gws` option of `opendds_idl`. This option produced sample
+  descriptions for the Wireshark dissector before OpenDDS v3.7.
 
 ## Version 3.13.1 of OpenDDS
 
--- a/PROBLEM-REPORT-FORM
+++ b/PROBLEM-REPORT-FORM
@@ -1,6 +1,6 @@
-This is OpenDDS version 3.13, released Wed Aug 22 14:15:40 UTC 2018
+This is OpenDDS version 3.14.0-dev (NOT A RELEASE)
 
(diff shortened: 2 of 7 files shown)


iguessthislldo:
A couple of notes about these changes:

- I've tested most of the release steps that I think are affected, but not all of them.
- I've almost certainly broken micro releases.
- I'm still not sure how to handle repeating the steps after doing all of them or stopping during the post release steps. Theoretically this was possible before. If you where do this the release steps as of now wouldn't validate properly because of the post release changes. I'm thinking the post release steps should either be a seperate set of steps, or some sort of flag file should be created in the workspace to say that we're done with the release now and to ignore those steps.
- I've tried to leave a future possibility of having the release script do pre-releases like we did for 3.14 in code for handling versions.

NEWS.md
```diff
@@ -1,7 +1,7 @@
 # OpenDDS Releases
 
-## Version 3.14 of OpenDDS
-Released %TIMESTAMP%
+## Version 3.14.0 of OpenDDS
```
mitza-oci:
```suggestion
## Version 3.14 of OpenDDS
```

PROBLEM-REPORT-FORM
```diff
@@ -1,6 +1,6 @@
-This is OpenDDS version 3.13, released Wed Aug 22 14:15:40 UTC 2018
+This is OpenDDS version 3.14.0-dev (NOT A RELEASE)
```
mitza-oci:
```suggestion
This is OpenDDS version 3.14-dev (NOT A RELEASE)
```

NEWS.md
```diff
@@ -20,8 +20,8 @@ Released %TIMESTAMP%
 - Java API can now be used on Android
 
 ### Notes:
-- Removed outdated `-Gws` option of `opendds_idl`. This option produced sample
-  descriptions for the Wireshark dissector before OpenDDS v3.7.
+- Removed outdated `-Gws` option of `opendds_idl`. This option produced topic
```
mitza-oci: I think this is already changed in the other PR, may be easiest to leave it alone here to prevent conflicts.

NEWS.md
iguessthislldo: I think this was a byproduct of me rewriting the code that processes the version to use. When do you want the full form of the version used?

mitza-oci: I'm just trying to keep it consistent with what we have now.  I think that is when the 3rd number is a 0, the trailing `.0` is omitted.

tools/scripts/gitrelease.pl
```diff
@@ -52,10 +53,28 @@
...
+}
+
 my $insert_news_template_after_line = 1;
 sub insert_news_template($$) {
-  my $version = shift;
-  my $timestamp = shift;
+  my $settings = shift();
+  my $post_release = shift();
+
+  print("pr: $post_release\n");
```
mitza-oci: Is this for debugging purposes only?  Should there be a `--verbose`?

tools/scripts/gitrelease.pl
```diff
@@ -1799,95 +1955,120 @@ sub message_news_template_file_section {
...
+  'next-version=s' => \$next_version,
+  'version-id=s' => \$version_id,
+  'skip-devguide!' => \$skip_devguide,
+  'skip-doxygen!' => \$skip_doxygen ,
+  'skip-website!' => \$skip_website ,
+  'skip-ftp!' => \$skip_ftp,
+) or die "See --help for options.\nStopped";
+
+if (!(scalar(@ARGV) == 0 || scalar(@ARGV) == 2)) {
+  die "Expecting at 0 or 2 positional arugments. See --help.\nStopped";
```
mitza-oci:
```suggestion
  die "Expecting 0 or 2 positional arguments. See --help.\nStopped";
```

tools/scripts/gitrelease.pl
iguessthislldo: It's a debug print, I'll take it out.

NEWS.md
iguessthislldo: I think that's kinda ambiguous, but I will change it back.

tools/scripts/gitrelease.pl
iguessthislldo: For some reason the spell checker doesn't work in perl strings in my vim. It's kinda annoying. For the help message I piped it into vim to spell check it.

NEWS.md
iguessthislldo: That's fine

PR #1559, opened 2020-02-22
fix ACE_Connector::close issue

huangminghuang:


dds/DCPS/transport/tcp/TcpTransport.h
```diff
@@ -139,6 +139,11 @@ class OpenDDS_Tcp_Export TcpTransport
   /// Used to accept passive connections on our local_address_.
   unique_ptr<TcpAcceptor> acceptor_;
 
+  class Connector : public ACE_Connector<TcpConnection, ACE_SOCK_Connector>
```
mitza-oci: Where is this used?  Should it be changed on line 148?

PR #1571, opened 2020-02-27
Disable network config modifier for safety profile.

jonesc-oci:


dds/DCPS/NetworkConfigModifier.h
```diff
@@ -10,7 +10,7 @@
 
 #include "ace/config.h"
 
-#ifdef ACE_HAS_GETIFADDRS
+#if defined(ACE_HAS_GETIFADDRS) && !defined(OPENDDS_SAFETY_PROFILE)
```
mitza-oci:
```suggestion
#if defined(ACE_HAS_GETIFADDRS) && !defined(OPENDDS_SAFETY_PROFILE)
```

PR #1572, opened 2020-02-27
Tweak Security Logging

iguessthislldo:
Tweak security logging categories and bring some more logging points under the control of them.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -585,7 +585,7 @@ Spdp::handle_participant_data(DCPS::MessageId id,
           if (DCPS::security_debug.auth_debug) {
             ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) {auth_debug} Spdp::handle_participant_data - ")
               ACE_TEXT("Incompatible security attributes in discovered participant: %C\n"),
-              std::string(DCPS::GuidConverter(guid)).c_str()));
+              OPENDDS_STRING(DCPS::GuidConverter(guid)).c_str()));
           }
           participants_.erase(guid);
         } else { // allow_unauthenticated_participants == true
@@ -604,7 +604,7 @@ Spdp::handle_participant_data(DCPS::MessageId id,
             if (DCPS::security_debug.auth_debug) {
               ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) {auth_debug} Spdp::handle_participant_data - ")
                 ACE_TEXT("Incompatible security attributes in discovered participant: %C\n"),
-                std::string(DCPS::GuidConverter(guid)).c_str()));
+                OPENDDS_STRING(DCPS::GuidConverter(guid)).c_str()));
             }
             participants_.erase(guid);
           } else { // allow_unauthenticated_participants == true
@@ -859,7 +859,7 @@ Spdp::handle_handshake_message(const DDS::Security::ParticipantStatelessMessage&
       ACE_DEBUG((LM_WARNING,
         ACE_TEXT("(%P|%t) {auth_warn} Spdp::handle_handshake_message() - ")
         ACE_TEXT("received handshake for undiscovered participant %C. Ignoring.\n"),
-                 std::string(DCPS::GuidConverter(src_participant)).c_str()));
+                 OPENDDS_STRING(DCPS::GuidConverter(src_participant)).c_str()));
     }
     return;
   }
@@ -994,7 +994,7 @@ Spdp::handle_handshake_message(const DDS::Security::ParticipantStatelessMessage&
           ACE_TEXT("expecting %C from %C. Security Exception[%d.%d]: %C\n"),
           iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REQUEST_SENT ?
             "reply" : "final message",
-          std::string(DCPS::GuidConverter(src_participant)).c_str(),
+          OPENDDS_STRING(DCPS::GuidConverter(src_participant)).c_str(),
           se.code, se.minor_code, se.message.in()));
       }
       return;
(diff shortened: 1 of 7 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1111,10 +1138,12 @@ Spdp::handle_participant_crypto_tokens(const DDS::Security::ParticipantVolatileM
   // If discovery hasn't initialized / validated this participant yet, ignore volatile message
   DiscoveredParticipantIter iter = participants_.find(src_participant);
   if (iter == participants_.end()) {
-    ACE_DEBUG((LM_WARNING,
-      ACE_TEXT("(%P|%t) Spdp::handle_participant_crypto_tokens() - ")
-      ACE_TEXT("received tokens for undiscovered participant %C. Ignoring.\n"),
-               std::string(DCPS::GuidConverter(src_participant)).c_str()));
+    if (OpenDDS::DCPS::security_debug.auth_warn) {
+      ACE_ERROR((LM_WARNING,
```
```suggestion
      ACE_DEBUG((LM_WARNING,
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1145,61 +1174,86 @@ Spdp::match_authenticated(const DCPS::RepoId& guid, DiscoveredParticipantIter& d
 
   dp->shared_secret_handle_ = auth->get_shared_secret(dp->handshake_handle_, se);
   if (dp->shared_secret_handle_ == 0) {
-    ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: ")
-      ACE_TEXT("Spdp::match_authenticated() - ")
-      ACE_TEXT("Unable to get shared secret handle. Security Exception[%d.%d]: %C\n"),
+    if (OpenDDS::DCPS::security_debug.auth_warn) {
+      ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) {auth_warn} ")
```
```suggestion
      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) {auth_warn} ")
```
?

dds/DCPS/security/AccessControl/Permissions.cpp
```diff
@@ -221,12 +221,14 @@ int Permissions::load(const SSL::SignedDocument& doc)
       }
     }
 
-    if (!valid_subject) {
-      ACE_DEBUG((LM_WARNING,
-        ACE_TEXT("(%P|%t) AccessControlBuiltInImpl::load_permissions_file: Unable to parse subject name, ignoring grant.\n")));
-    } else if (contains_subject_name(rule_holder_.subject)) {
-      ACE_DEBUG((LM_WARNING,
-        ACE_TEXT("(%P|%t) AccessControlBuiltInImpl::load_permissions_file: Ignoring grant with duplicate subject name.\n")));
+    if (!valid_subject && DCPS::security_debug.access_warn) {
```
mitza-oci: Combining these conditions means that when !valid_subject and not logging, the `contains_subject_name`  function on 228 will run.  That's a change from before.

dds/DCPS/RTPS/Spdp.cpp
iguessthislldo: Is it that this should be optionally compiled because of the possibility of it being run a lot?

mitza-oci: Is it really an `ACE_ERROR`?  If so, why is the log priority `LM_WARNING`?  Based on the comment it doesn't seem like "an error" - it's possible to get these messages in a different order, that's not a a problem.

iguessthislldo: I haven't been really trying to make the decision of what logging level a message should be in these changes. The reason I changed what log macro is being used is that the ACE book says `ACE_ERROR` is for "errors and warnings", although I see I haven't applied that change consistently.

dds/DCPS/security/AccessControl/Permissions.cpp
iguessthislldo: Yes you're right, I will fix this.

dds/DCPS/RTPS/Spdp.cpp
mitza-oci: I don't have the ACE book in front of me to get more context on that.  Based on Log_Msg.h/cpp it's just setting the 'operation status' which is data that can be logged.  In a simple model where you make an OS or stdlib call and it returns an error, I guess you can differentiate those easily.  This code is a few more layers up, so it's not as straightforward.
The other consideration (probably more important) is for the reader/maintainer of this code: Do we want to communicate ERROR here?

iguessthislldo: > I don't have the ACE book in front of me to get more context on that. Based on Log_Msg.h/cpp it's just setting the 'operation status' which is data that can be logged.

I knew it set the status, but the ACE book says if `ACE_NDEBUG` is defined it would leave out `ACE_DEBUG` statements, assumingly by using something like `#define ACE_DEBUG(X)`. In fact it states that's the entire reason for `ACE_ERROR` and `ACE_DEBUG` being macros. However you're right it doesn't do that, at least not anymore. I looked at the code and did a test to be sure and yea, `ACE_NDEBUG` doesn't have an effect on `ACE_DEBUG`. I feel betrayed...

iguessthislldo: > The other consideration (probably more important) is for the reader/maintainer of this code: Do we want to communicate ERROR here?

Regardless of what the logging macros are doing I think that should be the log category's job.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1111,10 +1138,12 @@ Spdp::handle_participant_crypto_tokens(const DDS::Security::ParticipantVolatileM
...
   if (iter == participants_.end()) {
-    ACE_DEBUG((LM_WARNING,
-      ACE_TEXT("(%P|%t) Spdp::handle_participant_crypto_tokens() - ")
-      ACE_TEXT("received tokens for undiscovered participant %C. Ignoring.\n"),
-               std::string(DCPS::GuidConverter(src_participant)).c_str()));
+    if (OpenDDS::DCPS::security_debug.auth_warn) {
+      ACE_ERROR((LM_WARNING,
+        ACE_TEXT("(%P|%t) {auth_warn} Spdp::handle_participant_crypto_tokens() - ")
+        ACE_TEXT("received tokens for undiscovered participant %C. Ignoring.\n"),
+        std::string(DCPS::GuidConverter(src_participant)).c_str()));
```
iguessthislldo: Should this be an `OPENDDS_STRING`?

mitza-oci: It doesn't need to be since there's no safety profile support here (yet), but it doesn't hurt to do that now

PR #1575, opened 2020-03-01
TransportClient: corrected map iterator type

mitza-oci:


simpsont-oci:
Good catch. Odd that it compiled.

mitza-oci:
> 
> 
> Good catch. Odd that it compiled.

It didn't compile everywhere, but it does seem like the newer compilers (the ones that we have CI checks for) are using the same -- or compatible/assignable -- iterator types across map and multimap

PR #1578, opened 2020-03-02
Control for Security Debug in PerlDDS

iguessthislldo:
Default to passing `-DCPSSecurityDebug 9` when `-Config OPENDDS_SECURITY` is passed, unless `-DCPSSecurityDebug` or `-DCPSSecurityDebugLevel` was specified in the test or `DCPSSecurityDebug` or `DCPSSecurityDebugLevel` environment variables were set.

Fixes secure messenger test failing on non-secure scoreboard builds caused by #1572. `-Config OPENDDS_SECURITY` will have to be passed on all security builds, but I'm ready to commit that.


mitza-oci:
Azure CI builds cancelled since they're not relevant for Perl changes.

PR #1591, opened 2020-03-12
Fixing bench worker listener registration for windows

simpsont-oci:


performance-tests/bench_2/builder/ListenerFactory.cpp
```diff
@@ -0,0 +1,73 @@
+#include "ListenerFactory.h"
+
+namespace Builder {
+
+Bench_Builder_Export
+void RegisterDomainParticipantListener(const std::string& name, const std::function<DDS::DomainParticipantListener::_var_type(const Builder::PropertySeq&)>& factory)
```
iguessthislldo: Why are these function names capitalized?

iguessthislldo:
Could you explain why this was needed for Windows?

performance-tests/bench_2/builder/ListenerFactory.cpp
simpsont-oci: Hmm, they started life as non-templated classes that inherited from the appropriate template specialization, but then I changed my mind and just made them functions. I'll rename them to use snake case.

simpsont-oci:
We needed some kind of workaround for windows because visual studio complained about the static member variables of an exported template class (For example, VS2015 says, "Builder::ListenerFactory<TopicListener>::s_factory_map: definition of dllimport static data member not allowed" when trying to build Bench_Worker). For the release, we made it compile, but unfortunately as it was it couldn't run on windows because both Bench_Worker and Builder would each define static listener factory maps... the worker would register the type but the builder library wouldn't be able to 'find' it in when trying to call the factory.

PR #1595, opened 2020-03-16
Fix ICE deadlock from multiple participants in the same process

jrw972:


dds/DCPS/RTPS/ICE/Checklist.cpp
```diff
@@ -189,7 +189,15 @@ void Checklist::unfreeze()
   for (CandidatePairsType::iterator pos = frozen_.begin(), limit = frozen_.end(); pos != limit;) {
     const CandidatePair& cp = *pos;
 
-    if (!endpoint_manager_->agent_impl->contains(cp.foundation)) {
+    // The second check allows the Checklist to start work on remote
+    // foundations that also below to its local agent meaning that the
```
mitza-oci:
```suggestion
    // foundations that also belong to its local agent meaning that the
```

PR #1596, opened 2020-03-17
Fix Filtering Instance Registration by Non-Key Fields

iguessthislldo:
Fix for bug introduced in #900 where instance registration samples are treated as full data samples for filtering.


dds/DCPS/DataSampleHeader.inl
```diff
@@ -169,5 +169,6 @@ OpenDDS::DCPS::DataSampleHeader::valid_data() const
   return !(message_id_ == DISPOSE_INSTANCE
     || message_id_ == DISPOSE_UNREGISTER_INSTANCE
     || message_id_ == UNREGISTER_INSTANCE
+    || message_id_ == INSTANCE_REGISTRATION
```
mitza-oci: Update doxygen comment for this method in DataSampleHeader.h

bjones-ametek:
I'd like to see a specific test added for this, since it was causing a segfault when using an enum field in the filter. That would be a significant problem if the bad behavior were re-introduced in the future.

mitza-oci:
> I'd like to see a specific test added for this, since it was causing a segfault when using an enum field in the filter. That would be a significant problem if the bad behavior were re-introduced in the future.

Right, we have some ideas on how test it.  We'll update this PR to include testing.

PR #1602, opened 2020-03-19
Fixing WCHAR compile issue in Bench Common lib

simpsont-oci:


mitza-oci:
Can we update one of the CI builds to cover this?

simpsont-oci:
> Can we update one of the CI builds to cover this?
Sure, we just need to make one of the windows builds use WCHAR. Any preference which one?

mitza-oci:
> 
> 
> > Can we update one of the CI builds to cover this?
> 
> Sure, we just need to make one of the windows builds use WCHAR. Any preference which one?

No preference.  We could even add one without it taking longer than it did yesterday, since VS2105 was just removed.

PR #1620, opened 2020-04-06
fixed leaks in bench_2

mitza-oci:


performance-tests/bench_2/node_controller/main.cpp
```diff
@@ -442,8 +444,9 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     std::cerr << "register_type failed for Report" << std::endl;
     return 1;
   }
+  type_name = report_ts->get_type_name();
   DDS::Topic_var report_topic = participant->create_topic(
-    report_topic_name, report_ts->get_type_name(), TOPIC_QOS_DEFAULT, nullptr,
+    report_topic_name,type_name, TOPIC_QOS_DEFAULT, nullptr,
```
mitza-oci, line 449:
```suggestion
    report_topic_name, type_name, TOPIC_QOS_DEFAULT, nullptr,
```

performance-tests/bench_2/builder/Participant.cpp
```diff
@@ -19,8 +19,9 @@ Participant::Participant(const ParticipantConfig& config, ParticipantReport& rep
   Log::log() << "Creating participant: '" << name_ << "' in domain " << domain_ << std::endl;
 
   // Customize QoS Object
+  DDS::DomainParticipantFactory_var dpf = TheParticipantFactory;
   DDS::DomainParticipantQos qos;
-  TheParticipantFactory->get_default_participant_qos(qos);
+  dpf->get_default_participant_qos(qos);
```
iguessthislldo, line 24: I wish this wasn't necessary or at least the type system helped enforce this by returning something like a `shared_ptr`.

mitza-oci: Eventually we'll do just that...

performance-tests/bench_2/node_controller/main.cpp
iguessthislldo: Why did you review yourself?

mitza-oci: It let me fix in with a few clicks in the GitHub web interface, instead of actually doing a commit in my repo.

performance-tests/bench_2/builder/Participant.cpp
simpsont-oci: I don't have the code in front of me... but presumably this is returning something allocated that I should have been trapping in a scoped wrapper?

Issue #1594, opened 2020-03-14
Can't use InfoRepo Discovery on Android


loveslikey:
Hi:
    I am trying to run opendds on android with DCPSInfoRepo. But I got error,when execute "DevGuideExamples\DCPS\Messenger" example. It seem not process the response data, because of topic not success matched. Here is the log file.
[subscriber.log](https://github.com/objectcomputing/OpenDDS/files/4332440/subscriber.log)
[publisher.log](https://github.com/objectcomputing/OpenDDS/files/4332441/publisher.log)
[DCPSInfoRepo.log](https://github.com/objectcomputing/OpenDDS/files/4332442/DCPSInfoRepo.log)
env:
opendds: 3.14
ACE+TAO:6.5.8
os:window 10 +google emulator
android: Android SDK built for x86_64 ,android version 10,
config:

```
[common]
DCPSInfoRepo=corbaloc::192.168.1.10:12345/DCPSInfoRepo
DCPSGlobalTransportConfig=subscriptiontransport
[config/subscriptiontransport]
transports=tcp

[config/publicationtransport]
transports=tcp

[transport/tcp]
transport_type=tcp
```

### **but it can work at underneath config,**

```
[common]
DCPSGlobalTransportConfig=$file

[domain/42]
DiscoveryConfig=fast_rtps42
DefaultTransportConfig=rtps_config42

[rtps_discovery/fast_rtps42]
ResendPeriod=2
SedpMulticast=0
TTL=20

[config/rtps_config42]
transports=the_rtps_transport42

[transport/the_rtps_transport42]
transport_type=tcp
local_address=10.0.0.14:
use_multicast=0
ttl=20
```

ClaytonCalabrese:
Hello, I have been trying to replicate your process in order to find the problem. 

What emulated device are you using that has an x86_64 architecture?

loveslikey:
Great! I used android studio  comes with the android emulator. In fact,it replicated not just on the simulator, but  also replicated on a real machine of android os( arm64  architecture). I hava tested all kinds of situation about this problem, it can be replicated.

iguessthislldo:
For OpenDDS team members in the future: also see #1325 which was the fix the last time we saw this problem (which was ultimately caused by https://github.com/DOCGroup/ACE_TAO/pull/858) and had the same workaround.

iguessthislldo:
@loveslikey, could you try out the changes in #1626 and confirm those changes work for what you're trying to do?

[...]

PR #1621, opened 2020-04-06
Changing visibility in android cross compiling

ClaytonCalabrese:
Android platform OpenDDS was failing due to shared library visibility.  This script change will detect if the build target is android and add the 'no_hidden_visibility=1' to the macros file.  

See: 
https://github.com/objectcomputing/OpenDDS/issues/1594


mitza-oci:
Should we close this due to #1626?

PR #1622, opened 2020-04-06
Updates from testing the PR #1616 changes with Thrasher

mitza-oci:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2106,7 +2106,7 @@ RtpsUdpDataLink::RtpsReader::gather_association_ack_nacks_i(MetaSubmessageVec& m
   }
 
   for (WriterInfoMap::iterator wi = remote_writers_.begin(); wi != remote_writers_.end(); ++wi) {
-    if (wi->second.first_valid_hb_ && wi->second.first_delivered_data_) {
+    if (wi->second.first_valid_hb_ || wi->second.first_delivered_data_) {
```
jrw972, line 2109: I think I told you wrong.  This should &&.  The expression reads "If we don't have a valid heartbeat and we don't have data then send an acknack."

PR #1626, opened 2020-04-08
Adding Missing Exports for Getting InfoRepo Disc. Working on Android

ClaytonCalabrese:
Fix for #1594

These exports are required in order for the messenger example ( and likely other opendds) to run on android when compiled with clang.  There was a dynamic cast failing during info repo discovery because these exports were missing.


iguessthislldo:
Excellent, Thanks Clayton!

There are almost certainly other missing exports, but I don't think we can do anything about that at the moment. As long as this works I think that's good for now.

PR #1643, opened 2020-04-24
Fix Fragmentation Issues

simpsont-oci:
Fixing a number of issues with fragmentation:
1) Not handling duplicate fragments correctly (combined ranges can exceed duplicate range).
2) Storing fragments for writers we've never associated with (potentially never cleaned)
3) For RTPS, where we can calculate total # of fragments, we only ever nack the 'next' fragment, wasting time.

Minor tweaks to node_controller & control domain

### DIFF
--- a/dds/DCPS/transport/framework/DataLink.cpp
+++ b/dds/DCPS/transport/framework/DataLink.cpp
@@ -963,10 +963,10 @@ DataLink::release_resources()
 }
 
 bool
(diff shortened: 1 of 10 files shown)


mitza-oci:
tests/transport/rtps_reliability/run_test.pl failed on GHPRB.  After that's fixed we are probably ready to merge.

PR #1646, opened 2020-04-27
Fix data races in SPDP

jrw972:


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2625,6 +2626,8 @@ void
 Spdp::SpdpTransport::join_multicast_group(const DCPS::NetworkInterface& nic,
                                           bool all_interfaces)
 {
+  ACE_GUARD(ACE_Thread_Mutex, g, outer_->lock_);
```
mitza-oci, line 2629: Is this safe with register_handler() on line 2652?

jrw972: Kind of.  I just rebased this branch and added a commit that forces the join_multicast_group to use the JobQueue so that it will be safe.

PR #1650, opened 2020-04-28
When no std version has been specified by the user for g++ detect the…

jwillemsen:
… version first so that the detected version results in a possible setting in platform_macros.GNU. Without this change with gcc >= 6 no c++14 is set as default

    * configure:


mitza-oci:
The logic of the existing configure script is that with a new enough g++, we don't pass a -std command-line argument since the one it defaults to is fine.

PR #1654, opened 2020-04-28
Remove reader_associator_ from RtpsUdpDataLink to prevent errant acknacks

simpsont-oci:
Fix for the Thrasher RTPS & Bench CI Disco Tests.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -109,7 +109,6 @@ RtpsUdpDataLink::RtpsUdpDataLink(RtpsUdpTransport& transport,
                      config.heartbeat_response_delay_)
   , heartbeat_(reactor_task->interceptor(), config.heartbeat_period_, *this, &RtpsUdpDataLink::send_heartbeats)
   , heartbeatchecker_(reactor_task->interceptor(), *this, &RtpsUdpDataLink::check_heartbeats)
-  , reader_associator_(reactor_task->interceptor(), *this, &RtpsUdpDataLink::send_association_ack_nacks)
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2375,7 +2317,7 @@ RtpsUdpDataLink::bundle_mapped_meta_submessages(AddrDestMetaSubmessageMap& adr_m
     for (DestMetaSubmessageMap::iterator dest_it = addr_it->second.begin(); dest_it != addr_it->second.end(); ++dest_it) {
       for (MetaSubmessageIterVec::iterator resp_it = dest_it->second.begin(); resp_it != dest_it->second.end(); ++resp_it) {
         // Check before every meta_submessage to see if we need to prefix a INFO_DST
-        if (dest_it->first != GUID_UNKNOWN && dest_it->first != prev_dst) {
+        if (dest_it->first != prev_dst) {
```
jrw972, line 2320: Why was this reverted?

simpsont-oci:
It wasn't, this is a different line of code from the one you changed in
send_bundled_submessages. The bundling logic ought to match the sending
logic, and since you updated the sending logic for this check, I updated
the bundling logic to match.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: It wasn't, this is a different line of code from the one you changed in
send_bundled_submessages. The bundling logic ought to match the sending
logic, and since you updated the sending logic for this check, I updated
the bundling logic to match.

jrw972: Got it.

PR #1655, opened 2020-04-29
Don't use unitialized addresses for discovery/ICE

jrw972:


dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -566,33 +566,37 @@ RtpsUdpTransport::IceEndpoint::host_addresses() const {
...
 
-  transport.unicast_socket_.get_local_addr(addr);
-  if (addr.is_any()) {
-    ICE::AddressListType addrs;
-    DCPS::get_interface_addrs(addrs);
-    for (ICE::AddressListType::iterator pos = addrs.begin(), limit = addrs.end(); pos != limit; ++pos) {
-      if (pos->get_type() == AF_INET) {
-        pos->set_port_number(addr.get_port_number());
-        addresses.push_back(*pos);
+  addr = transport.config().local_address();
```
mitza-oci, line 569: combine this with the statement above (declare and initialize in the same statement)

PR #1656, opened 2020-04-29
Fix missing include

jrw972:


dds/DCPS/RTPS/ICE/Checklist.h
```diff
@@ -17,6 +17,7 @@
 #include "Task.h"
 
 #include "dds/DCPS/TimeTypes.h"
+#include "dds/DCPS/GuidUtils.h"
```
mitza-oci, line 20: This doesn't appear to be sufficient to fix the error we saw on the scoreboard.
(Side note: may be good to look in to how to get a similar setup on Azure Pipelines.)
Since `operator<<` here is inline, it looks like it needs `<iostream>`, may be good to move the implementation to the .cpp

PR #1664, opened 2020-05-01
Make use of DOC Group ACE+TAO x.5.9

jwillemsen:


iguessthislldo:
@mitza-oci Do we also need to update the required DOC Group version and/or OCI patch number in the README?

mitza-oci:
> 
> 
> @mitza-oci Do we also need to update the required DOC Group version and/or OCI patch number in the README?

Typically these don't change until a release is made.  We could revisit that.

PR #1669, opened 2020-05-05
Configurable maximum message size for RTPS (rtps_udp, SEDP)

jrw972:
The RTPS messages needed for security can be large enough that they
trigger IP fragmentation.  Usages of RTPS that traverse the public
Internet will most likely use security and therefore be subject to IP
fragmentation.  Unfortunately, many routers and router-like systems
don't support IP fragmentation correctly.

Fortunately, RTPS supports fragmentation.  The goal of this change to
expose the limit at which fragmentation occurs to the user.


dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -508,10 +510,22 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
...
-              ACE_TEXT("(%P|%t) RtpsDiscovery::Config::discovery_config(): ")
-              ACE_TEXT("Invalid entry (%C) for MaxSpdpSequenceMsgResetChecks in ")
-              ACE_TEXT("[rtps_discovery/%C] section.\n"),
-              string_value.c_str(), rtps_name.c_str()), -1);
+                              ACE_TEXT("(%P|%t) RtpsDiscovery::Config::discovery_config(): ")
+                              ACE_TEXT("Invalid entry (%C) for MaxSpdpSequenceMsgResetChecks in ")
+                              ACE_TEXT("[rtps_discovery/%C] section.\n"),
+                              string_value.c_str(), rtps_name.c_str()), -1);
+          }
+        } else if (name == "UdpMaxMessageSize") {
```
mitza-oci, line 518:
```suggestion
        } else if (name == "SedpMaxMessageSize") {
```
making it clear that this doesn't impact SPDP.  The use of "UDP" for the constant in TransportSendStrategy was because it's UDP's inherent max size, which we can't change.

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -100,6 +101,8 @@ RtpsUdpInst::load(ACE_Configuration_Heap& cf,
   GET_CONFIG_STRING_VALUE(cf, sect, ACE_TEXT("multicast_interface"),
                           multicast_interface_);
 
+  GET_CONFIG_VALUE(cf, sect, ACE_TEXT("udp_max_message_size"), udp_max_message_size_, size_t);
```
mitza-oci, line 104:
```suggestion
  GET_CONFIG_VALUE(cf, sect, ACE_TEXT("max_message_size"), udp_max_message_size_, size_t);
```

mitza-oci: There are still some ini files using the old name

PR #1671, opened 2020-05-05
Fixed a corner case for IDL union code generation in Java

mitza-oci:
Unions with implicit defaults that don't have a branch selected need their discriminator values copied from C++ to Java.
There may be one more issue with unions that I'll add to this PR, but I wanted to get CI going on this part of it.

### DIFF
--- a/java/idl2jni/codegen/im_jni.cpp
+++ b/java/idl2jni/codegen/im_jni.cpp
@@ -1641,14 +1641,17 @@ ostream &operator<< (ostream &o, AST_Expression::AST_ExprValue *ev)
 }
 
 bool idl_mapping_jni::gen_union(UTL_ScopedName *name,
-                                const std::vector<AST_UnionBranch *> &branches, AST_Type *discriminator,
-                                AST_Expression::ExprType, const AST_Union::DefaultValue &, const char *)
+                                const std::vector<AST_UnionBranch *> &branches,
+                                AST_Type *discriminator,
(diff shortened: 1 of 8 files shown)


java/idl2jni/codegen/im_jni.cpp
```diff
@@ -1791,6 +1795,30 @@ bool idl_mapping_jni::gen_union(UTL_ScopedName *name,
       "  " + disc_ty + " jdisc = source._d ();\n";
   }
 
+  if (!hasDefault && default_value.computed_ != 0) {
```
jrw972, line 1798: Why is hasDefault negated?

mitza-oci: This is from the IDL-to-Java spec (IDL-to-C++ is very similar):
> Two default modifier methods, both named __default(), are generated if there is no explicit default case label, and the set of case labels does not completely cover the possible values of the discriminant.

jrw972: So I'm assuming that `default_value.computed_ != 0` is a way of detecting if the case labels don't completely cover the values of the discriminant?

mitza-oci: https://github.com/DOCGroup/ACE_TAO/blob/7c298ab69026c64c027d63530f7ca275f0bc9c73/TAO/TAO_IDL/include/ast_union.h#L122

PR #1674, opened 2020-05-07
RtpsRelay should not forward beacons

jrw972:
Note: INFO_SRC was added to the PAD messages.


jrw972:
Found it.  The pads were all from the application participant which were not subject to decryption which means they were not processed correctly.

PR #1675, opened 2020-05-08
XTypes: Assignability

sonndinh:


dds/DCPS/TypeAssignability.h
```diff
@@ -0,0 +1,133 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_TYPE_ASSIGNABILITY_H
+#define OPENDDS_DCPS_TYPE_ASSIGNABILITY_H
+
+#include "dds/DCPS/TypeObject.h"
```
iguessthislldo, line 9:
```suggestion
#include "TypeObject.h"
```

ClaytonCalabrese:
The test failures are irrelevant for this PR. The Travis CI PR will be fixed by my code so that also does not apply to this PR.  The build failures need sorting out before merge.

dds/DCPS/TypeAssignability.h
sonndinh: I saw some headers are included using the `dds/DCPS/` prefix, and some are not. Do we have a rule for when to use which?

dds/DCPS/TypeAssignability.h
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+  private:
+    MinimalTypeObject tmp_;
+    //    std::map<TypeIdentifier, MinimalTypeObject> table_;
+  };
+
+  // Set of pairs of members with each pair contains members from
+  // two structure types that have the same member ID and name
+  typedef std::vector<std::pair<const MinimalStructMember*,
```
mitza-oci, line 34: Use `OPENDDS_VECTOR` so we get allocator support in safety profile builds.

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+  }
+
+  /**
+   * @brief At least one input type object must be TK_ALIAS
+   */
+  bool TypeAssignability::assignable_alias(const MinimalTypeObject& ta,
+                                           const MinimalTypeObject& tb) const
+  {
+    if (TK_ALIAS == ta.kind && TK_ALIAS != tb.kind) {
+      const TypeIdentifier& tia = *ta.alias_type.body.common.related_type.in();
```
mitza-oci, line 115:
```suggestion
      const TypeIdentifier& tia = *ta.alias_type.body.common.related_type;
```
I don't think `.in()` would be needed.  Of course, keep it in if required to compile.

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+      for (size_t j = 0; j < tb.struct_type.member_seq.members.size(); ++j) {
+        MemberId id_b = tb.struct_type.member_seq.members[j].common.member_id;
+        const NameHash& h_b = tb.struct_type.member_seq.members[j].detail.name_hash;
+        ACE_CDR::ULong
+          name_b = (h_b[0] << 24) | (h_b[1] << 16) | (h_b[2] << 8) | (h_b[3]);
+        if ((name_a == name_b && id_a != id_b) ||
+            (id_a == id_b && name_a != name_b)) {
+          return false;
+        } else if (name_a == name_b && id_a == id_b) {
+          matched_members.push_back(std::pair<const MinimalStructMember*,
```
mitza-oci, line 251: Use `make_pair`?

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+    }
+
+    // Members for which both optional is false and must_understand is true in
+    // either T1 or T2 appear in both T1 and T2.
+    // Members marked as key in either T1 or T2 appear in both T1 and T2.
+    for (size_t i = 0; i < ta.struct_type.member_seq.members.size(); ++i) {
+      MemberFlag flags = ta.struct_type.member_seq.members[i].common.member_flags;
+      MemberId id = ta.struct_type.member_seq.members[i].common.member_id;
+      bool found = false;
+      if ((flags & IS_OPTIONAL) == 0 &&
```
mitza-oci, line 304: The two sides of the `&&` could be combined into one mask/compare operation.  But I'm OK letting the optimizer do that if it wants to.  I'm not sure one way would be easier to read and maintain than the other.

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+   */
+  bool TypeAssignability::assignable_struct(const MinimalTypeObject& ta,
+                                            const MinimalTypeObject& tb) const
+  {
+    if (TK_STRUCTURE != tb.kind) {
+      return false;
+    }
+
+    // Extensibility kind must match
+    TypeFlag extensibility_mask = IS_FINAL | IS_APPENDABLE | IS_MUTABLE;
```
mitza-oci, line 228:
```suggestion
    const TypeFlag extensibility_mask = IS_FINAL | IS_APPENDABLE | IS_MUTABLE;
```
In general, make these locals variables const if it would be an error for them to change after initialization.

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+    // For any member m2 of T2, if there is a member m1 of T1 with the same
+    // ID, then the type KeyErased(m1.type) is-assignable-from the type
+    // KeyErased(m2.type)
+    for (size_t i = 0; i < matched_members.size(); ++i) {
+      const CommonStructMember& member = matched_members[i].second->common;
+      const MinimalTypeObject* toa = 0;
+      const MinimalTypeObject* tob = 0;
+      bool type_matched = false;
+      // KeyErased(T) is defined only for struct and union
+      if (get_struct_member(&tob, member)) {
```
mitza-oci, line 275: Does the 1st parameter type really need to be a `**` or could it be `*&`?

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+      const Sequence<LBound>& bounds_b = tb.array_type.header.common.bound_seq;
+      if (TI_PLAIN_ARRAY_SMALL == ta.kind) {
+        const Sequence<SBound>& bounds_a = ta.array_sdefn.array_bound_seq;
+        if (bounds_a.members.size() != bounds_b.members.size()) {
+          return false;
+        }
+
+        for (size_t i = 0; i < bounds_a.members.size(); ++i) {
+          if (static_cast<LBound>(bounds_a.members[i]) != bounds_b.members[i]) {
+              return false;
```
mitza-oci, line 1365:
```suggestion
            return false;
```

dds/DCPS/TypeAssignability.cpp
```diff
@@ -0,0 +1,2083 @@
...
+        return false;
+      }
+    } else {
+      return false;
+    }
+
+    // All literals in tb must appear as literals in toa
+    for (size_t j = 0; j < literals_b.members.size(); ++j) {
+      const NameHash& h_b = literals_b.members[j].detail.name_hash;
+      ACE_CDR::ULong
```
mitza-oci, line 1888: Try to avoid splitting the variable's type and name on separate lines.  We are not using a strict 80-column limit, but it's also good to keep lines at a reasonable length -- so these two style guidelines are somewhat at odds, but it usually works out OK.  One thing that will save 2 characters is that the style guide says that the method itself starts at 0 indent.

dds/DCPS/TypeAssignability.h
iguessthislldo: For this specifically, no, but it probably should be, because I don't see a reason why to use a include like this for a file that's in the same directory. [These are the rules on includes we have right now](https://github.com/objectcomputing/OpenDDS/blob/master/docs/guidelines.md#includes).

mitza-oci: Right, I think it falls into the general rule: if you can express the same concept in fewer lines/characters and it's still clear and maintainable, then that shorter form is preferred.

iguessthislldo: I have went ahead and proposed this as a guideline: #1704

sonndinh: Great. I've removed the prefix. I think we may want to update other files to use this guideline for consistency.

dds/DCPS/TypeAssignability.h
sonndinh: I've updated this. In general, should I always use the macros defined for OpenDDS when working with containers?

dds/DCPS/TypeAssignability.cpp
sonndinh: Removing `.in()` compiles, so I've removed it.

dds/DCPS/TypeAssignability.cpp
sonndinh: I've updated to use `make_pair`.

dds/DCPS/TypeAssignability.cpp
sonndinh: I've changed to combine them into one comparison. It's a bit harder to read but I think it is ok with a comment.

dds/DCPS/TypeAssignability.cpp
sonndinh: Fixed this and other places.

dds/DCPS/TypeAssignability.cpp
sonndinh: Yes, `*&` works too. I've updated to use `*&`.

dds/DCPS/TypeAssignability.cpp
sonndinh: Fixed.

dds/DCPS/TypeAssignability.cpp
sonndinh: Updated to avoid line splitting.

sonndinh:
> Assuming Adam's suggestions are made I am satisfied. The logic looks good to me and I think the upcoming testing will figure out any problems. Additionally, this PR is completely separate from existing OpenDDS, so nothing existing will break.

Thanks. The previous compiling errors on Windows were fixed by including `DCPS/DdsDcps_pch.h`.

dds/DCPS/TypeAssignability.h
mitza-oci: Yes, it's needed when the code in question is part of the core libraries that will be built when Safety Profile is enabled.  It's not a problem to also use in other contexts.

PR #1677, opened 2020-05-08
Remove OfferedDeadlineWatchdog

jrw972:
The watchdog uses timer_ids and therefore must wait() when scheduling
timers.  This effectively introduces a lock which leads to deadlock
when the Deadline QoS is used.  The deadly embrace occurs between the
writing thread which has the DataWriterImpl lock and needs the reactor
lock and the Reactor thread servicing the watchdog which has the
reactor thread and needs the DataWriterImpl lock.

Solution: Rewrite the OfferedDeadlineWatchdog using SporadicTask.

In general, wait()ing on Command objects in the ReactorInterceptor is
unsafe and the wait functionality should be removed to reduce the
chance of similar deadlocks in the future.


dds/DCPS/WriteDataContainer.h
```diff
@@ -116,7 +119,7 @@ typedef OPENDDS_MAP(DDS::InstanceHandle_t, PublicationInstance_rch)
  *           we do not deadlock; and, 2) we incur the cost of
  *           obtaining the lock only once.
  */
-class OpenDDS_Dcps_Export WriteDataContainer : public PoolAllocationBase {
+class OpenDDS_Dcps_Export WriteDataContainer : public RcObject {
```
jrw972, line 122: @mitza-oci We need to discuss the ramifications of changing this to RcObject.

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -1500,6 +1488,159 @@ WriteDataContainer::log_send_state_lists (OPENDDS_STRING description)
...
+        deadline_task_->schedule(deadline_period);
+      }
+    } else if (deadline_period == TimeDuration::max_value) {
+      if (!deadline_map_.empty()) {
+        deadline_task_->cancel();
+      }
+
+      deadline_map_.clear();
+    } else {
+      // TODO: Which is correct?  The first can cause immediate deadline misses?  The second corresponds to the existing behavior.
```
jrw972, line 1520: @mitza-oci Another discussion point.

dds/DCPS/WriteDataContainer.h
```diff
@@ -528,6 +531,28 @@ class OpenDDS_Dcps_Export WriteDataContainer : public PoolAllocationBase {
   DDS::DurabilityServiceQosPolicy const & durability_service_;
 
 #endif
+
+  /// Timer responsible for reporting missed offered deadlines.
+  RcHandle<DCPS::PmfSporadicTask<WriteDataContainer> > deadline_task_;
+  TimeDuration deadline_period_; // TimeDuration::zero_value means no deadline.
+  typedef std::multimap<MonotonicTimePoint, PublicationInstance_rch> DeadlineMapType;
```
mitza-oci, line 538: needs allocator-enabled STL container

dds/DCPS/WriteDataContainer.cpp
mitza-oci: discussed

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -1500,6 +1488,159 @@ WriteDataContainer::log_send_state_lists (OPENDDS_STRING description)
...
+      //const TimeDuration adjustment = deadline_period - deadline_period_;
+      const TimeDuration adjustment = deadline_period;
+      DeadlineMapType new_map;
+      for (PublicationInstanceMapType::iterator iter = instances_.begin();
+           iter != instances_.end();
+           ++iter) {
+        iter->second->deadline_ += adjustment;
```
simpsont-oci, line 1527: So if the previous deadline_ was an hour in the future, but new deadline is 5 seconds, this looks like it's bumping the next deadline and next wait to 1hr + 5 seconds, which seems wrong?

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -1500,6 +1488,159 @@ WriteDataContainer::log_send_state_lists (OPENDDS_STRING description)
...
+        new_map.insert(std::make_pair(iter->second->deadline_, iter->second));
+      }
+      std::swap(new_map, deadline_map_);
+
+      if (!deadline_map_.empty()) {
+        deadline_task_->cancel();
+        deadline_task_->schedule(deadline_map_.begin()->first - MonotonicTimePoint::now());
```
simpsont-oci, line 1534: I would maybe update the logic for calculating the next scheduled deadline similar between this line and how it's calculated above for a "new" deadline... i.e. have a temporary variable named deadline, just to make it clear what's going on here. Also be aware that sometimes canceling timers can take a while, so calculating "now" as close to function entry as possible is probably the desired behavior.

dds/DCPS/WriteDataContainer.h
```diff
@@ -528,6 +531,28 @@ class OpenDDS_Dcps_Export WriteDataContainer : public PoolAllocationBase {
...
+  DeadlineMapType deadline_map_;
+
+  /// Lock for synchronization of @c status_ member.
+  ACE_Recursive_Thread_Mutex& deadline_status_lock_;
```
simpsont-oci, line 542: Yuck. Understandable, but yuck.

dds/DCPS/WriteDataContainer.h
```diff
@@ -528,6 +531,28 @@ class OpenDDS_Dcps_Export WriteDataContainer : public PoolAllocationBase {
...
+  /// Reverse lock used for releasing the @c status_lock_ listener upcall.
+  ACE_Reverse_Lock<ACE_Recursive_Thread_Mutex> deadline_reverse_status_lock_;
```
simpsont-oci, line 544: Do we need this as a member? Can we just create these inline as we need them?

dds/DCPS/DataSampleElement.h
```diff
@@ -31,6 +32,7 @@ typedef Cached_Allocator_With_Overflow<DataSampleElement, ACE_Null_Mutex>
 class TransportSendListener;
 struct PublicationInstance;
 typedef RcHandle<PublicationInstance> PublicationInstance_rch;
+typedef WeakRcHandle<PublicationInstance> PublicationInstance_wrch;
```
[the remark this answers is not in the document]
jrw972: I'll take it out see what happens.

dds/DCPS/WriteDataContainer.cpp
jrw972: I agree.  I will change it so it resets all instances to now() + deadline_period.

dds/DCPS/WriteDataContainer.cpp
jrw972: In this case, it is better to defer now() to as late as possible.  If I use now() upon function entry, then the relative deadline to schedule will be larger and the timer will fire after the deadline.

dds/DCPS/WriteDataContainer.h
jrw972: I'll inline it.

dds/DCPS/WriteDataContainer.h
jrw972: Yeah.  Passing around references to locks...what could possibly go wrong?

PR #1684, opened 2020-05-12
Mailing List Form

kroegerk:
This should allow you to add a functional subscription form to the website. Feel free to adjust the CSS to match the styles on the site; let me know if you have any questions. Once you've got it on a page, let me know, and we can test it to make sure it's working.

### DIFF
--- a/_includes/mailing-list-form
+++ b/_includes/mailing-list-form
@@ -1,0 +1,308 @@
+<?php
+
+?>
+
+<style type="text/css" scoped>
+
+	.mauticform_wrapper { 
+		max-width: 800px;
+		margin-left: auto;
+		margin-right:auto;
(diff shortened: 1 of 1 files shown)


iguessthislldo:
I'm very confused about what the purpose of this is. This doesn't look like it has anything to do with our mailing lists. Also why does it have a PHP wrapper at the top when we use Jekyll?

mitza-oci:
> 
> 
> I'm very confused about what the purpose of this is. This doesn't look like it has anything to do with our mailing lists. Also why does it have a PHP wrapper at the top when we use Jekyll?

You can ignore it.  We aren't ready to merge at this point.

mitza-oci:
Closing, we'll open a new PR when it's ready.

PR #1685, opened 2020-05-12
Add mailing list form to home page

kroegerk:
I may be doing this wrong. I failed to save the form file with a file extension, and I'm not sure where to assign attributes to the class I just made up. Pretty sure I need some assistance figuring all this out.


mitza-oci:
This is separate from #1684 so I'm going to close it, we'll need to get those changes integrated.  We will discuss more via email.

PR #1686, opened 2020-05-13
Replace clang9/gcc9 with clang10/gcc10 in azure pipelines builds

jwillemsen:


mitza-oci:
I think I looked this up last week and Azure Pipelines's documentation says there is no GCC 10 installed

https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md

jwillemsen:
It is installed through PackageDeps, it is now available as optional package in ubuntu

mitza-oci:
> 
> 
> It is installed through PackageDeps, it is now available as optional package in ubuntu

Right... I guess this should be better documented.  It is not in any of ubuntu's own packages for 18.04 (https://packages.ubuntu.com/).  Looks like as a side-effect of the installation of the earlier versions, the PPA repo "ubuntu-toolchain-r/test" is activated in apt.

https://github.com/actions/virtual-environments/blob/5abcc104b2d27650a716e8ec1b4e823c1fbae30a/images/linux/scripts/installers/gcc.sh#L29

My real question for this PR: a little while ago we went through and culled some build configurations across both Travis and Azure.  The idea was that having every version of 3 different compilers on 3 platforms increases the build time too much relative to the value that it provides.  So when adding these 2 builds, are there 2 we can take away?

jwillemsen:
Between 8/9/10 I don't see much differences in results, you could remove 8/9. On github actions you get 20 parallel runners for free (azure has 10), so maybe someone has time to migrate to github actions?

mitza-oci:
> Between 8/9/10 I don't see much differences in results, you could remove 8/9. 

Please do that in this PR.

> On github actions you get 20 parallel runners for free (azure has 10), so maybe someone has time to migrate to github actions?

Right, this makes sense to work on, but it will be a little while before our team has time to do so.

jwillemsen:
> > Between 8/9/10 I don't see much differences in results, you could remove 8/9.
> 
> Please do that in this PR.

Done

PR #1691, opened 2020-05-15
Only send auth replies for auth requests

jrw972:
One participant (the initiator) has discovered another
participant (the replier) and sent an auth request.  Due to SPDP
reset, it removes the participant but immediately rediscovers it and
sends another auth request.  If the replier receives the first auth
request, it will continue to respond to it but authentication will
fail as the initiator has changed the Diffie-Hellman parameters.

Solution:

1. The replier will not periodically resend replies.  It will only
send auth replies in response to an auth request.
2. If the replier detects that the auth request has changed, it will
recompute a new auth reply.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -915,7 +915,23 @@ Spdp::handle_handshake_message(const DDS::Security::ParticipantStatelessMessage&
   DCPS::RepoId reader = src_participant;
   reader.entityId = ENTITYID_P2P_BUILTIN_PARTICIPANT_STATELESS_READER;
 
-  if (iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY && msg.related_message_identity.source_guid == GUID_UNKNOWN) {
+  if ((iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY || iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY_SENT) && msg.related_message_identity.source_guid == GUID_UNKNOWN) {
+    // The initiator will retry using the same request.  If the
+    // initiator retries very quickly and we change the response,
+    // i.e., different diffie, then we will never authenticate.  Thus,
+    // if the initiator sends the same request, then we should send
+    // the same response.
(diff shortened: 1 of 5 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -915,7 +915,23 @@ Spdp::handle_handshake_message(const DDS::Security::ParticipantStatelessMessage&
...
 
-  if (iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY && msg.related_message_identity.source_guid == GUID_UNKNOWN) {
+  if ((iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY || iter->second.auth_state_ == DCPS::AS_HANDSHAKE_REPLY_SENT) && msg.related_message_identity.source_guid == GUID_UNKNOWN) {
+    // The initiator will retry using the same request.  If the
+    // initiator retries very quickly and we change the response,
+    // i.e., different diffie, then we will never authenticate.  Thus,
+    // if the initiator sends the same request, then we should send
+    // the same response.
+    if (iter->second.has_last_stateless_msg_ &&
+        iter->second.last_handshake_request_data_ == msg.message_data[0]) {
```
mitza-oci, line 925: Make sure that `message_data` is nonempty before using it.

dds/DCPS/security/AuthenticationBuiltInImpl.cpp
```diff
@@ -332,7 +332,7 @@ ::DDS::Security::ValidationResult_t AuthenticationBuiltInImpl::begin_handshake_r
     return DDS::Security::VALIDATION_FAILED;
   }
 
-  LocalParticipantData& local_data = *(handshake_data.first);
+  const LocalParticipantData& local_data = *(handshake_data.first);
```
mitza-oci, line 335:
```suggestion
  const LocalParticipantData& local_data = *handshake_data.first;
```
If you're going to change the line...

dds/DCPS/RTPS/Spdp.cpp
jrw972: Line 891 has a check that msg.message_data.length() != 0.

PR #1696, opened 2020-05-19
Migrate RtpsUdpDatalink's heartbeat replies to MultiTask

simpsont-oci:
For consistent sending of pre-association AckNacks

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -1833,7 +1833,6 @@ Spdp::SpdpTransport::open()
 #endif
 
   local_sender_ = DCPS::make_rch<SpdpMulti>(reactor_task_.interceptor(), outer_->config_->resend_period(), ref(*this), &SpdpTransport::send_local);
-  local_sender_->enable(TimeDuration());
 
 #ifdef OPENDDS_SECURITY
   auth_deadline_processor_ = DCPS::make_rch<SpdpSporadic>(reactor_task_.interceptor(), ref(*this), &SpdpTransport::process_auth_deadlines);
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -1833,6 +1833,7 @@ Spdp::SpdpTransport::open()
 #endif
 
   local_sender_ = DCPS::make_rch<SpdpMulti>(reactor_task_.interceptor(), outer_->config_->resend_period(), ref(*this), &SpdpTransport::send_local);
+  local_sender_->enable(outer_->config_->resend_period());
 
 #ifdef OPENDDS_SECURITY
   auth_deadline_processor_ = DCPS::make_rch<SpdpSporadic>(reactor_task_.interceptor(), ref(*this), &SpdpTransport::process_auth_deadlines);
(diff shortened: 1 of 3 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1833,7 +1833,6 @@ Spdp::SpdpTransport::open()
 #endif
 
   local_sender_ = DCPS::make_rch<SpdpMulti>(reactor_task_.interceptor(), outer_->config_->resend_period(), ref(*this), &SpdpTransport::send_local);
-  local_sender_->enable(TimeDuration());
```
jrw972, line 1836: Is this enabled by default?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -925,7 +930,7 @@ RtpsUdpDataLink::stop_i()
   }
 
   nack_reply_.cancel();
-  heartbeat_reply_.cancel();
+  heartbeat_reply_.disable_and_wait();
```
jrw972, line 933: This should be `disable()`.  MultiTask (like the others) has a weak pointer to the delegate so you can disable it without waiting.

dds/DCPS/RTPS/Spdp.cpp
simpsont-oci: Hmm, I was seeing the participant message twice so I assumed yes, but the code says otherwise. Is something else writing the local announcement? If so, this can be enabled with the actual delay instead of empty TimeDuration.

jrw972: For platforms with a network config monitor, it writes the announcement after it joins the multicast group.

simpsont-oci: What platforms won't have a network config monitor? Do I just need to deal with the fact that there will sometimes be a duplicate SPDP announcement?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Hmm, all the other ones in stop_i() are also disable_and_wait()... I would expect us to want to wait in case shutdown finishes and MultiTask (PmfMultiTask, in this case) gets erased before the cancel command runs?

dds/DCPS/RTPS/Spdp.cpp
jrw972: Linux.  Mac.  You should always be prepared to handle a duplicate SPDP announcement.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I my opinion, it would be better to dynamically allocate all of the *Tasks including heartbeat_reply_ so you can take advantage of the weak pointer and not wait during shutdown.    So, you are correct that you will need to `disable_and_wait()` unless you dynamically allocate `heartbeat_replay_`.

jrw972: I looked into this a little further and the *Tasks don't support what I was claiming.  They could be modified to support it.  I'll discuss that with you in a different venue.

PR #1699, opened 2020-05-20
Prevent Fragment Storage / Reassembly For Previously Completed Sequence Numbers

simpsont-oci:


dds/DCPS/transport/framework/TransportReassembly.cpp
```diff
@@ -168,7 +183,7 @@ TransportReassembly::get_gaps(const SequenceNumber& seq, const RepoId& pub_id,
   // length is number of (allocated) words in bitmap, max of 8
   // numBits is number of valid bits in the bitmap, <= length * 32, to account for partial words
   const FragInfoMap::const_iterator iter = fragments_.find(FragKey(pub_id, seq));
-  if (iter == fragments_.end() || length == 0) {
+  if (iter == fragments_.end() || iter->second.complete_ || length == 0) {
```
mitza-oci, line 186: I realize the previous version didn't do this, but should it check `length` first?

simpsont-oci: Sure, why not. I'll move it out and up before the call to find().

PR #1705, opened 2020-05-27
Improve check_invariants in ICE EndpointManager

jrw972:
The check was improved to check the consistency between the
EndpointManager data structures and the Checklists.

### DIFF
--- a/dds/DCPS/RTPS/ICE/Checklist.h
+++ b/dds/DCPS/RTPS/ICE/Checklist.h
@@ -161,6 +161,16 @@ struct Checklist : public Task {
 
   void check_invariants() const;
 
+  bool has_transaction_id(const STUN::TransactionId& a_tid) const
+  {
+    return std::find(connectivity_checks_.begin(), connectivity_checks_.end(), a_tid) != connectivity_checks_.end();
+  }
+
+  bool has_guid_pair(const GuidPair& a_guid_pair) const
+  {
+    return guids_.find(a_guid_pair) != guids_.end();
+  }
+
   void unfreeze();
 
   void unfreeze(const FoundationType& a_foundation);
--- a/dds/DCPS/RTPS/ICE/EndpointManager.cpp
+++ b/dds/DCPS/RTPS/ICE/EndpointManager.cpp
@@ -722,10 +722,34 @@ void EndpointManager::compute_active_foundations(ActiveFoundationSet& a_active_f
 
 void EndpointManager::check_invariants() const
 {
+  // Check for expired deferred checks.
+  size_t deferred = 0;
+  for (DeferredTriggeredChecksType::const_iterator pos = deferred_triggered_checks_.begin(),
+         limit = deferred_triggered_checks_.end(); pos != limit; ++pos) {
+    const DeferredTriggeredCheckListType& list = pos->second;
+
+    OPENDDS_ASSERT(!list.empty());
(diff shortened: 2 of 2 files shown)


dds/DCPS/RTPS/ICE/EndpointManager.cpp
```diff
@@ -722,11 +722,35 @@ void EndpointManager::compute_active_foundations(ActiveFoundationSet& a_active_f
...
-       limit = username_to_checklist_.end(); pos != limit; ++pos) {
+         limit = username_to_checklist_.end(); pos != limit; ++pos) {
     ChecklistPtr checklist = pos->second;
+    OPENDDS_ASSERT(checklist->original_remote_agent_info().username == pos->first);
     checklist->check_invariants();
   }
+
+  for (TransactionIdToChecklistType::const_iterator pos = transaction_id_to_checklist_.begin(),
+         limit = transaction_id_to_checklist_.end(); pos != limit; ++pos) {
+    ChecklistPtr checklist = pos->second;
```
mitza-oci, line 745: Is this going to be an "unused variable" warning when asserts are disabled?  (same on 751)

dds/DCPS/RTPS/ICE/Checklist.h
```diff
@@ -161,6 +161,16 @@ struct Checklist : public Task {
...
   void check_invariants() const;
 
+  bool has_transaction_id(const STUN::TransactionId& a_tid) const
+  {
+    return std::find(connectivity_checks_.begin(), connectivity_checks_.end(), a_tid) != connectivity_checks_.end();
+  }
+
+  bool has_guid_pair(const GuidPair& a_guid_pair) const
+  {
+    return guids_.find(a_guid_pair) != guids_.end();
```
mitza-oci, line 171: Use `count()` here?  I guess it's equivalent.

jrw972: When returning a boolean, I tend to prefer a boolean expression.  It's longer, but in my opinion, easier to read.  When using `count` I have to make the extra mental step that non-zero returns as true.

mitza-oci: Starting in C++20 you can use `std::set::contains`.  Until then, I tend to use `count` since "does it have this key" (without needing an iterator) seems like the most fundamental operation on sets.  We don't need to change this code.

dds/DCPS/RTPS/ICE/EndpointManager.cpp
```diff
@@ -723,14 +723,12 @@ void EndpointManager::compute_active_foundations(ActiveFoundationSet& a_active_f
...
 {
   // Check for expired deferred checks.
-  size_t deferred = 0;
   for (DeferredTriggeredChecksType::const_iterator pos = deferred_triggered_checks_.begin(),
          limit = deferred_triggered_checks_.end(); pos != limit; ++pos) {
     const DeferredTriggeredCheckListType& list = pos->second;
 
     OPENDDS_ASSERT(!list.empty());
     OPENDDS_ASSERT(list.front().expiration_date >= MonotonicTimePoint::now() - agent_impl->get_configuration().server_reflexive_address_period() - agent_impl->get_configuration().server_reflexive_address_period());
-    deferred += list.size();
```
mitza-oci: Now `list` is also an unused variable.  I wonder if/when compilers will start warning about loops with empty bodies.

PR #1712, opened 2020-06-02
XTypes: fixes to wireshark for xtypes

ClaytonCalabrese:
small function call changes to deal with updates that are part of xtypes

### DIFF
--- a/tools/dissector/packet-opendds.cpp
+++ b/tools/dissector/packet-opendds.cpp
@@ -194,7 +194,7 @@ namespace OpenDDS
       T t;
 
       guint len = std::min(ws_tvb_length(tvb) - offset,
-                           static_cast<guint>(t.max_marshaled_size()));
+                           static_cast<guint>(t.get_max_serialized_size()));
       const guint8* data = tvb_get_ptr(tvb, offset, len);
 
       ACE_Message_Block mb(reinterpret_cast<const char*>(data));
@@ -290,8 +290,9 @@ namespace OpenDDS
       offset += len;
 
       // hf_sequence
-      size_t size = 0, padding = 0;
-      gen_find_size(header.sequence_, size, padding);
+      size_t size = 0;
+      const Encoding encoding(Encoding::KIND_CDR_UNALIGNED);
+      serialized_size(encoding, size, header.sequence_);
       len = static_cast<gint>(size);
       proto_tree_add_uint64(ltree, hf_sequence, tvb_, offset, len,
                             gint64(header.sequence_.getValue()));
@@ -347,8 +348,9 @@ namespace OpenDDS
       offset += len;
 
       // hf_sample_sequence
-      size_t size = 0, padding = 0;
-      gen_find_size(sample.sequence_, size, padding);
+      size_t size = 0;
+      const Encoding encoding(Encoding::KIND_CDR_UNALIGNED);
+      serialized_size(encoding, size, sample.sequence_);
       len = static_cast<gint>(size);
       if (sample.message_id_ == SAMPLE_DATA) {
         proto_tree_add_uint64(ltree, hf_sample_sequence, tvb_, offset, len,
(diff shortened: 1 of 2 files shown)


tools/dissector/packet-opendds.cpp
```diff
@@ -290,8 +290,9 @@ namespace OpenDDS
       offset += len;
 
       // hf_sequence
-      size_t size = 0, padding = 0;
-      gen_find_size(header.sequence_, size, padding);
+      size_t size = 0;
+      const Encoding encoding(Encoding::KIND_CDR_UNALIGNED);
+      serialized_size(encoding, size, header.sequence_);
```
iguessthislldo, line 295:
```suggestion
      size_t size = serialized_size(encoding, header.sequence_);
```

You can use this when the size starts at zero.

tools/dissector/packet-opendds.cpp
```diff
@@ -347,8 +348,9 @@ namespace OpenDDS
       offset += len;
 
       // hf_sample_sequence
-      size_t size = 0, padding = 0;
-      gen_find_size(sample.sequence_, size, padding);
+      size_t size = 0;
+      const Encoding encoding(Encoding::KIND_CDR_UNALIGNED);
+      serialized_size(encoding, size, sample.sequence_);
```
iguessthislldo, line 353:
```suggestion
      size_t size = serialized_size(encoding, sample.sequence_);
```

tools/dissector/packet-opendds.cpp
```diff
@@ -451,7 +453,7 @@ namespace OpenDDS
 
                   // Get Entry Size
                   size = 0;
-                  gen_find_size(filter, size, padding);
+                  serialized_size(encoding, size, filter);
```
iguessthislldo, line 456:
```suggestion
                  size = serialized_size(encoding, filter);
```

tools/dissector/packet-opendds.cpp
```diff
@@ -290,8 +290,9 @@ namespace OpenDDS
       offset += len;
 
       // hf_sequence
-      size_t size = 0, padding = 0;
-      gen_find_size(header.sequence_, size, padding);
+      size_t size = 0;
+      const Encoding encoding(Encoding::KIND_CDR_UNALIGNED);
```
iguessthislldo, line 294: Could you factor the encoding objects out into a common one?

ClaytonCalabrese: Depends.  There are multiple functions.  I thought I did factor out to the minimum but I can take another look.

tools/dissector/packet-opendds.cpp
ClaytonCalabrese: If I don't reset size back to 0 will it not cause problems?

tools/dissector/packet-opendds.cpp
ClaytonCalabrese: Or are you saying to try and make it a single variable they all have access to within the namespace?

tools/dissector/packet-opendds.cpp
iguessthislldo: This is a template wrapper around the generated `serialized_size` that passes a zero to it and returns the result. It's equivalent, just a bit smaller.

tools/dissector/packet-opendds.cpp
iguessthislldo: > Or are you saying to try and make it a single variable they all have access to within the namespace?

Yes, I think you can just define a single encoding object for these cases and put it in the anonymous namespace higher up in the file.

ClaytonCalabrese:
> Also were any changes needed for the sample dissector?

There are some of the old serializer constructor in data members that I will need to factor out in my other PR.  I'm not sure we can be certain until they are both together.

ClaytonCalabrese:
> > Also were any changes needed for the sample dissector?
> 
> There are some of the old serializer constructor in data members that I will need to factor out in my other PR. I'm not sure we can be certain until they are both together.

After merging there are changes needed to the serializer endianess but nothing for this PR. I think this PR is ready to go.

PR #1715, opened 2020-06-03
XTypes: Unit tests for primitive types assignability

sonndinh:


sonndinh:
> Very clean. I noticed you have placeholders for the string test are they going to be part of this merge or are we going to merge piece by piece?

I'll make another PR for string test. The idea is to have one PR for each type, which has its own Trello card. It makes the process a bit linear, but we'll see if it works well as we do tests for other types.

PR #1668, opened 2020-05-05
XTypes: Encoding-Aware Serialized Size Functions

iguessthislldo:
Quite a bit of XCDR behavior changes the size of encoded data, so the functions for getting this should be aware of the encoding or at least have the encoding available to use in the future. At the same time the the names of theses size functions weren't consistent ("marshaled size" and "find size") and had an old "gen" prefix most of the time. This makes the function names more consistent by using the names `serialized_size` and `max_serialized_size`.

PR #1721, opened 2020-06-05
XTypes: More XCDR Serialization Support

iguessthislldo:
Continue Work on XCDR Serialization Support from #1668.


iguessthislldo:
Closed because I accidentally merged from the upstream master instead of upstream xtypes. Reopened as #1722.

PR #1723, opened 2020-06-05
Associations are removed after expire-associate

jrw972:
Suppose that a Participant A is suspended long enough that it expires
for Participant B.  Participant A is then resumed.  In RTPS, this will
most likely cause a number of timers to fire include the one that will
cause Participant A to 1) expire Participant B and 2) send SPDP.  As
part of expiring Participant B, associations between writers in
Participant B and readers in Participant A will be scheduled for
removal (in Participant A).  Sending SPDP will restart the discovery
process.  This may cause associations between the readers and writers
to complete for the timer fires that causes the association to be
removed.  Finally, that timer will fire and remove the association
that was just added.

Solution: When adding an association to a reader, cancel any
RemoveAssociationSweeper that will remove the writer.

### DIFF
--- a/dds/DCPS/RemoveAssociationSweeper.h
+++ b/dds/DCPS/RemoveAssociationSweeper.h
@@ -36,6 +36,7 @@ public:
 
   void schedule_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS::WriterInfo>& info, bool callback);
   ReactorInterceptor::CommandPtr cancel_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS::WriterInfo>& info);
+  void cancel_timer(OpenDDS::DCPS::PublicationId writer_id);
 
   // Arg will be PublicationId
   int handle_timeout(const ACE_Time_Value& current_time, const void* arg);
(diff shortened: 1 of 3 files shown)


dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -36,6 +36,7 @@ class RemoveAssociationSweeper : public ReactorInterceptor {
 
   void schedule_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS::WriterInfo>& info, bool callback);
   ReactorInterceptor::CommandPtr cancel_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS::WriterInfo>& info);
+  void cancel_timer(OpenDDS::DCPS::PublicationId writer_id);
```
mitza-oci, line 39:
```suggestion
  void cancel_timer(const OpenDDS::DCPS::PublicationId& writer_id);
```

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -120,6 +121,23 @@ RemoveAssociationSweeper<T>::cancel_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS:
   return execute_or_enqueue(new CancelCommand(this, info));
 }
 
+template <typename T>
+void
+RemoveAssociationSweeper<T>::cancel_timer(OpenDDS::DCPS::PublicationId writer_id)
+{
+  struct Predicate {
+    Predicate(const OpenDDS::DCPS::PublicationId writer_id) : writer_id_(writer_id) {}
```
mitza-oci, line 129:
```suggestion
    explicit Predicate(const OpenDDS::DCPS::PublicationId& writer_id) : writer_id_(writer_id) {}
```

dds/DCPS/RemoveAssociationSweeper.h
mitza-oci, line 39: I know you're following the pattern that's already here, but I don't know why the code in this file keeps obfuscating the interesting parts by repeating the full namespace name on every line (sometimes twice).

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -120,6 +121,23 @@ RemoveAssociationSweeper<T>::cancel_timer(OpenDDS::DCPS::RcHandle<OpenDDS::DCPS:
...
+    OpenDDS::DCPS::PublicationId writer_id_;
+    bool operator() (const RcHandle<WriterInfo>& info) const {
+      return writer_id_ == info->writer_id_;
+    }
+  };
+  OPENDDS_VECTOR(RcHandle<WriterInfo>)::iterator pos = std::find_if(info_set_.begin(), info_set_.end(), Predicate(writer_id));
```
mitza-oci, line 135: This will fail pre-C++11 (I think) due to instantiating a template (`find_if`) with a local class argument.

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -25,6 +25,17 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace DCPS {
+namespace {
+
+struct Predicate {
+  explicit Predicate(const PublicationId writer_id) : writer_id_(writer_id) {}
```
mitza-oci, line 31:
```suggestion
  explicit Predicate(const PublicationId& writer_id) : writer_id_(writer_id) {}
```

PR #1727, opened 2020-06-08
DCPSDefaultAddress parsing

### LINKED ISSUE #1717: Parsing of DCPSDefaultAddress without a port number fails.
I've downloaded and I'm trying the latest (as of a couple of hours ago) master branch. I'm not having much luck; passing the -DCPSDefaultAddress 169.254.5.198 in the command line I'm getting to this in RtpsDiscovery.cpp:

----
         // spdpaddr defaults to DCPSDefaultAddress if set
         if (!TheServiceParticipant->default_address().empty()) {
            ACE_INET_Addr addr;
            if (addr.set(TheServiceParticipant->default_address().c_str())) {
               ACE_ERROR_RETURN((LM_ERROR,
                                          ACE_TEXT("(%P|%t) ERROR: RtpsDiscovery::Config::discovery_config(): ")
                                          ACE_TEXT("failed to parse Service Participant default address %C\n"),
                                          TheServiceParticipant->default_address().c_str()),
                                         -1);
            }
            config->spdp_local_address(addr);
         }
----

Stepping into ACE_INET_Addr::set() the passed in string is "169.254.5.198", and the code is:

----
int
ACE_INET_Addr::string_to_addr (const char s[], int address_family)
{
   ACE_TRACE ("ACE_INET_Addr::string_to_addr");
   int result;
   char *ip_buf = 0;
   char *ip_addr = 0;

   // Need to make a duplicate since we'll be overwriting the string.
   ACE_ALLOCATOR_RETURN (ip_buf,
                                    ACE_OS::strdup (s),
                                    -1);
   ip_addr = ip_buf;
   // We use strrchr because of IPv6 addresses.
   char *port_p = ACE_OS::strrchr (ip_addr, ':');
----

At this point, ip_addr contains "169.254.5.198", and port_p (as you'd probably expect) is 0.

The next bit of code executed is:

----
   if (port_p == 0) // Assume it's a port number.
      {
         char *endp = 0;
         long const port = ACE_OS::strtol (ip_addr, &endp, 10);

         if (*endp == '\0')      // strtol scanned the entire string - all digits
            {
               if (port < 0 || port >  ACE_MAX_DEFAULT_PORT)
                  result = -1;
               else
                  result = this->set (u_short (port), ACE_UINT32 (INADDR_ANY));
            }
         else // port name
            result = this->set (ip_addr, ACE_UINT32 (INADDR_ANY));
      }
----

Has something changed in this area? I've been using -DCPSDefaultAddress=<ip addr>  (with no port number on the end) for a while now, but this code seems to suggest that, if there isn't a port number specified in the string (i.e. no ":", e.g. "169.254.5.198" as opposed to "169.254.5.198:<portno>", then assume it's a port number!

At the "long const port" line the result is that port is assigned the value 169, but endp != '\0' so the this->set call is made into:

----
int
ACE_INET_Addr::set (const char port_name[],
                              ACE_UINT32 inet_address,
                              const char protocol[])
{
   ACE_TRACE ("ACE_INET_Addr::set");

   int const port_number = get_port_number_from_name (port_name, protocol);
   if (port_number == -1)
      {
         ACE_NOTSUP_RETURN (-1);
      }

   return this->set (static_cast<u_short>  (port_number),
                              inet_address, 0);
}
----

with port_name "169.254.5.198", inet_address 0, protocol "tcp". I go into get_port_number_from_name with port_name "169.254.5.198" and protocol "tcp".

at: 
----
   long port_number = ACE_OS::strtol (port_name, &endp, 10);
----

port_number becomes 169, and endp != '\0', so I get to:

----
#else
   port_number = -1;
   servent sentry;
   ACE_SERVENT_DATA buf;
   servent *sp = ACE_OS::getservbyname_r (port_name,
                                                              protocol,
                                                              &sentry,
                                                              buf);
   if (sp != 0)
      port_number = sp->s_port;
#endif /* ACE_LACKS_GETSERVBYNAME */
----

which, because the port_name is an IP Address at this point, causes sp to be a null pointer and -1 is returned. That leads to

----
         ACE_NOTSUP_RETURN (-1);
----

happening in ACE_INET_Addr::set(), making ACE_INET_Addr::string_to_addr() return -1, making the original ACE_INET_Addr::set(const char address[], int address_family) return -1, hence printing the stuff from this:

----
               ACE_ERROR_RETURN((LM_ERROR,
                                          ACE_TEXT("(%P|%t) ERROR: RtpsDiscovery::Config::discovery_config(): ")
                                          ACE_TEXT("failed to parse Service Participant default address %C\n"),
                                          TheServiceParticipant->default_address().c_str()),
                                         -1);
----

Seems odd.

John

mitza-oci:
Use ACE_INET_Addr as the internal representation of DCPSDefaultAddress
Update transports and RtpsDiscovery to match

Fixes #1717

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscovery.cpp
+++ b/dds/DCPS/RTPS/RtpsDiscovery.cpp
@@ -125,16 +125,8 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
       RtpsDiscoveryConfig_rch config = discovery->config();
 
       // spdpaddr defaults to DCPSDefaultAddress if set
-      if (!TheServiceParticipant->default_address().empty()) {
-        ACE_INET_Addr addr;
-        if (addr.set(TheServiceParticipant->default_address().c_str())) {
-          ACE_ERROR_RETURN((LM_ERROR,
-                            ACE_TEXT("(%P|%t) ERROR: RtpsDiscovery::Config::discovery_config(): ")
-                            ACE_TEXT("failed to parse Service Participant default address %C\n"),
-                            TheServiceParticipant->default_address().c_str()),
-                           -1);
-        }
-        config->spdp_local_address(addr);
+      if (TheServiceParticipant->default_address() != ACE_INET_Addr()) {
+        config->spdp_local_address(TheServiceParticipant->default_address());
       }
 
       DCPS::ValueMap values;
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -597,7 +597,13 @@ Service_Participant::parse_args(int &argc, ACE_TCHAR *argv[])
       got_log_verbose = true;
 
     } else if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("-DCPSDefaultAddress"))) != 0) {
-      this->default_address_ = ACE_TEXT_ALWAYS_CHAR(currentArg);
+      if (default_address_.set(u_short(0), currentArg)) {
+        ACE_ERROR_RETURN((LM_ERROR,
+                          ACE_TEXT("(%P|%t) ERROR: Service_Participant::parse_args: ")
+                          ACE_TEXT("failed to parse default address %C\n"),
+                          currentArg),
+                         -1);
+      }
       arg_shifter.consume_arg();
       got_default_address = true;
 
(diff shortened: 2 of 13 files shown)


mitza-oci:
We have also discussed DCPSDefaultAddress's interactions with NetworkConfigMonitor.  I haven't worked on that part yet.  Depending on how it goes it could be added to this PR or done in a separate PR.

PR #1731, opened 2020-06-09
Xtypes: DRAFT: Data Representation Effective QoS

ClaytonCalabrese:
Rather than having the logic to handle blank data representation in every location needed, have a function that interprets and converts.


ClaytonCalabrese:
Apparently the git history is messy here. I'll get it fixed by my next push.

iguessthislldo:
> It is also including other people's changes somehow.

You might have figured this out by now, but you pulled from master instead of xtypes before committing, so these are changes from master that xtypes branch doesn't have. I did the same exact thing with with #1721. I just created a new PR with cherry picked commits, but you could also force push this branch after correcting the history.

PR #1732, opened 2020-06-10
Participants cannot reauthenticate

jrw972:
Suppose participant A has a lease time of 100 seconds and participant
B has a lease time of 50 seconds.  Participant B is suspended for 50+
seconds so that it expires in Participant A.  Participant B is then
resumed.  This will cause Participant A to rediscover Participant
B.  (Participant B does not need to rediscover Participant A because
it has not expired.)  Participant A must reauthenticate with
Participant B.

If Participant A is the requester, then Participant B must restart
authentication when it receives the handshake request.  If Participant
B is the requester, then Participant A must send an auth request and
Participant B must restart the authentication processes.

Solution: Add support for both reauthentication scenarios.

### DIFF
--- a/dds/DCPS/GuidConverter.cpp
+++ b/dds/DCPS/GuidConverter.cpp
@@ -206,14 +206,6 @@ GuidConverter::uniqueId() const
   return id;
 }
 
-#ifdef OPENDDS_SECURITY
-bool
-GuidConverter::isSecure() const
-{
-  return isBuiltinDomainEntity() && guid_.entityId.entityKey[0] == 0xff;
-}
-#endif
-
 } // namespace DCPS
 } // namespace OpenDDS
 
--- a/dds/DCPS/GuidConverter.cpp
+++ b/dds/DCPS/GuidConverter.cpp
@@ -206,6 +206,14 @@ GuidConverter::uniqueId() const
   return id;
 }
 
+#ifdef OPENDDS_SECURITY
+bool
+GuidConverter::isSecure() const
+{
+  return guid_.entityId.entityKey[0] == 0xff;
+}
+#endif
+
 } // namespace DCPS
 } // namespace OpenDDS
 
--- a/dds/DCPS/GuidConverter.cpp
+++ b/dds/DCPS/GuidConverter.cpp
@@ -210,7 +210,7 @@ GuidConverter::uniqueId() const
 bool
 GuidConverter::isSecure() const
 {
-  return guid_.entityId.entityKey[0] == 0xff;
+  return isBuiltinDomainEntity() && guid_.entityId.entityKey[0] == 0xff;
 }
 #endif
 
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -924,6 +924,49 @@ void Sedp::associate_volatile(const Security::SPDPdiscoveredParticipantData& pda
   }
 }
 
+void Sedp::rekey_volatile(const Security::SPDPdiscoveredParticipantData& pdata)
+{
+  using namespace DDS::Security;
+
+  const BuiltinEndpointSet_t& avail = pdata.participantProxy.availableBuiltinEndpoints;
+
+  DDS::Security::SecurityException se = {"", 0, 0};
+  DDS::Security::CryptoKeyFactory_var key_factory = spdp_.get_security_config()->get_crypto_key_factory();
+  const DCPS::RepoId part_guid = make_id(pdata.participantProxy.guidPrefix, ENTITYID_PARTICIPANT);
+
+  if (avail & BUILTIN_PARTICIPANT_VOLATILE_MESSAGE_SECURE_WRITER) {
+    Spdp::ParticipantCryptoInfoPair info = spdp_.lookup_participant_crypto_info(part_guid);
+    if (info.first != DDS::HANDLE_NIL && info.second) {
+      const DCPS::RepoId writer_guid = make_id(part_guid, ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_WRITER);
+      if (!key_factory->rekey_remote_datawriter(remote_writer_crypto_handles_[writer_guid], info.second, se)) {
+        ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::rekey_volatile() - ")
+                   ACE_TEXT("Failure calling rekey_remote_datawriter(). Security Exception[%d.%d]: %C\n"),
+                   se.code, se.minor_code, se.message.in()));
+      }
+    } else {
+      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::rekey_volatile() - ")
+                 ACE_TEXT("Unable to lookup remote participant crypto info.\n")));
+    }
+  }
+  if (avail & BUILTIN_PARTICIPANT_VOLATILE_MESSAGE_SECURE_READER) {
+    Spdp::ParticipantCryptoInfoPair info = spdp_.lookup_participant_crypto_info(part_guid);
+    if (info.first != DDS::HANDLE_NIL && info.second) {
+      const DCPS::RepoId reader_guid = make_id(part_guid, ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_READER);
+      if (!key_factory->rekey_remote_datareader(remote_reader_crypto_handles_[reader_guid], info.second, se)) {
+        ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::rekey_voltatile() - ")
+                   ACE_TEXT("Failure calling rekey_remote_datawriter(). Security Exception[%d.%d]: %C\n"),
+                   se.code, se.minor_code, se.message.in()));
+      }
+    } else {
+      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::rekey_volatile() - ")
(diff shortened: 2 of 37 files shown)


dds/DCPS/GuidConverter.cpp
```diff
@@ -206,6 +206,14 @@ GuidConverter::uniqueId() const
   return id;
 }
 
+#ifdef OPENDDS_SECURITY
+bool
+GuidConverter::isSecure() const
```
mitza-oci, line 211: Should this also check that entityKind is a built-in kind?  Otherwise we don't know how entityKey[0] is assigned.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1335,6 +1411,30 @@ Sedp::disassociate(const ParticipantData_t& pdata)
...
+Sedp::replay_durable_data_for(const DCPS::RepoId& remote_sub_id)
+{
+  DCPS::GuidConverter conv(remote_sub_id);
+  ACE_DEBUG((LM_DEBUG, "Sedp::replay_durable_data_for %C\n", OPENDDS_STRING(conv).c_str()));
+  if (remote_sub_id.entityId == ENTITYID_SEDP_BUILTIN_PUBLICATIONS_READER) {
+    write_durable_publication_data(remote_sub_id, false);
+  } else if (remote_sub_id.entityId == ENTITYID_SEDP_BUILTIN_SUBSCRIPTIONS_READER) {
+    write_durable_subscription_data(remote_sub_id, false);
+  } else if (remote_sub_id.entityId == ENTITYID_P2P_BUILTIN_PARTICIPANT_MESSAGE_READER) {
+    write_durable_participant_message_data(remote_sub_id);
```
mitza-oci, line 1424: I'm not sure why there's a "write_durable" for this one (and its secure version), since the endpoints are not durable.

dds/DCPS/security/CryptoBuiltInImpl.h
```diff
@@ -77,6 +77,11 @@ class DdsSecurity_Export CryptoBuiltInImpl
     bool relay_only,
     DDS::Security::SecurityException& ex);
 
+  virtual bool rekey_remote_datareader(
```
mitza-oci, line 80: Could this method be removed and "folded in" to register_matched_remote_datareader?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2735,8 +2835,8 @@ Sedp::association_complete(const RepoId& localId,
     write_durable_dcps_participant_secure(remoteId);
   } else if (remoteId.entityId == ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_READER) {
 
-    if (associated_volatile_readers_.insert(remoteId).second) {
-
+    if (associated_volatile_readers_.insert(remoteId).second && spdp_.remote_is_requester(remoteId)) {
+      // TODO: Will reliability resend these?
```
mitza-oci, line 2839: (Not directly related to the "TODO" comment above)

Take a look at the order of `Sedp::associate_volatile` (which gets the SharedSecret from Auth to Crypto and makes the transport aware of it) relative to sending the Final part of the handshake in `Spdp::handle_handshake_message`.

In order to know that key exchange on the "volatile" built-in endpoints will be able to decrypt messages, the installation of the SharedSecret on the "requester" side should occur before sending the Final handshake message.

On the "replier" side, keys can be sent directly after receiving the Final and installing the SharedSecret.

On the "requester" side, keys can be sent after receiving at least one key from the peer.  To guarantee this is possible, the participant key could be changed from conditionally to unconditionally generated.  Still need to check how the spec handles this, what it can/should guarantee.

dds/DCPS/RTPS/Sedp.cpp
jrw972: It is durable in the code.  We can defer a fix.

dds/DCPS/security/CryptoBuiltInImpl.h
jrw972: To do so, we would need to commit to either 1) caching the DatareaderCryptoHandle in a way that we can find it again on the next call to register_matched_remote_datareader or 2) generate a new crypto handle and distribute it everywhere that it needs to be.

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: Yep, I'll make a note to come back to it.

dds/DCPS/security/CryptoBuiltInImpl.h
mitza-oci: Just to confirm, this method is only used for the BuiltinParticipantVolatileMessageSecure endpoints, correct?

I was not thinking of case (2) and it doesn't seem like what we want.

I think (1) isn't difficult and we should do it.  I could work on this if you'd like.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -866,7 +865,15 @@ Spdp::handle_auth_request(const DDS::Security::ParticipantStatelessMessage& msg)
...
     pending_remote_auth_tokens_[guid] = msg.message_data[0];
   } else {
-    iter->second.remote_auth_request_token_ = msg.message_data[0];
+    DiscoveredParticipant& dp = iter->second;
+    if (!(dp.remote_auth_request_token_ == msg.message_data[0])) {
+      dp.remote_auth_request_token_ = msg.message_data[0];
+      if (dp.is_requester_) {
+        dp.handshake_state_= DCPS::HANDSHAKE_STATE_EXPECTING_REQUEST;
+      } else {
+        send_handshake_request(guid, dp);
```
simpsont-oci, line 874: Doesn't this need to check the auth state / handshake state before sending off a handshake request? Currently, it seems like this path can circumvent calling validate_remote_identity before sending the handshake request, which seems contrary to the state machine in Figure 9 of the security spec.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -62,7 +62,7 @@ bool bitmapNonEmpty(const OpenDDS::RTPS::SequenceNumberSet& snSet)
       }
       for (int bit = 31; bit >= 0; --bit) {
```
simpsont-oci, line 63: This should probably just create a mask and do a single comparison rather than loop over individual bits.

dds/DCPS/RTPS/Spdp.cpp
jrw972: Correct.  Fixed.

dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -1173,6 +1173,21 @@ DataLink::network_change() const
...
+{
+  GuidConverter local(local_pub_id);
+  GuidConverter remote(remote_sub_id);
+  ACE_DEBUG((LM_DEBUG, "Replay durable data %C -> %C\n", OPENDDS_STRING(local).c_str(), OPENDDS_STRING(remote).c_str()));
+  TransportSendListener_rch send_listener = send_listener_for(local_pub_id);
+  if (send_listener) {
+    ACE_DEBUG((LM_DEBUG, "Replay durable data %C -> %C\n", OPENDDS_STRING(local).c_str(), OPENDDS_STRING(remote).c_str()));
+    send_listener->replay_durable_data_for(remote_sub_id);
+  } else {
+    ACE_DEBUG((LM_DEBUG, "Replay durable data %C -> %C no send_listener\n", OPENDDS_STRING(local).c_str(), OPENDDS_STRING(remote).c_str()));
```
iguessthislldo, line 1187: Debug logs

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -2684,7 +2829,19 @@ Sedp::received_volatile_message_secure(DCPS::MessageId /* message_id */,
...
+    if (spdp_.handle_participant_crypto_tokens(msg)) {
+      const DCPS::RepoId remote_volatile_reader = make_id(msg.message_identity.source_guid, ENTITYID_P2P_BUILTIN_PARTICIPANT_VOLATILE_SECURE_READER);
+      if (associated_volatile_readers_.count(remote_volatile_reader) != 0) {
+        ACE_DEBUG((LM_DEBUG, "Sedp::received_volatile_message_secure calling send_participant_crypto_token\n"));
+        spdp_.send_participant_crypto_tokens(remote_volatile_reader);
+        send_cached_crypto_tokens(remote_volatile_reader);
+        send_builtin_crypto_tokens(spdp_.get_participant_data(remote_volatile_reader));
+        resend_user_crypto_tokens(remote_volatile_reader);
+      } else {
+        ACE_DEBUG((LM_DEBUG, "Sedp::received_volatile_message_secure deferring send_participant_crypto_token until asociation complete %C\n", OPENDDS_STRING(DCPS::GuidConverter(remote_volatile_reader)).c_str()));
```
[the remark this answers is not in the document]
mitza-oci: no, it's not necessary.  it may be preferred in some cases though

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: I'll note it for later.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3224,6 +3446,23 @@ void Spdp::process_participant_ice(const ParameterList& plist,
     }
   }
 }
+
+bool Spdp::remote_is_requester(const DCPS::RepoId& guid) const
+{
+  // TODO:  Locking?
```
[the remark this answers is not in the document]
mitza-oci: It is already locked. I'll remove the comment.

PR #1736, opened 2020-06-10
CMake: support C++11 mapping when building shared libs

mitza-oci:
Also updated run_ci_tests.pl to attempt C++11 test and shared libs for all 3 tests

### DIFF
--- a/tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
+++ b/tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
@@ -17,6 +17,9 @@ set(idl "${target_prefix}idl")
 add_library(${idl})
 OPENDDS_TARGET_SOURCES(${idl} "${src}/Idl/Messenger.idl" OPENDDS_IDL_OPTIONS "-Lc++11")
 target_link_libraries(${idl} PUBLIC OpenDDS::Dcps)
+set_target_properties(${idl} PROPERTIES
+  RUNTIME_OUTPUT_DIRECTORY "${dst}/Idl"
+)
 
 # Assert the mapping used was C++11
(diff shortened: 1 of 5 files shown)


iguessthislldo:
I guess that makes sense that the C++11 messenger wouldn't work on C++03 builds. We could pass an argument to the script on those builds to exclude it from the list.

mitza-oci:
> 
> 
> I guess that makes sense that the C++11 messenger wouldn't work on C++03 builds. We could pass an argument to the script on those builds to exclude it from the list.

It should "know" from config.cmake, I'll look in to how that could work.

tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ set(idl "${target_prefix}idl")
 add_library(${idl})
 OPENDDS_TARGET_SOURCES(${idl} "${src}/Idl/Messenger.idl" OPENDDS_IDL_OPTIONS "-Lc++11")
 target_link_libraries(${idl} PUBLIC OpenDDS::Dcps)
+set_target_properties(${idl} PROPERTIES
+  RUNTIME_OUTPUT_DIRECTORY "${dst}/Idl"
```
iguessthislldo, line 21:
```suggestion
  LIBRARY_OUTPUT_DIRECTORY "${dst}/Idl"
```

mitza-oci: I based this on CMake's docs. https://cmake.org/cmake/help/v3.18/manual/cmake-buildsystem.7.html#runtime-output-artifacts

iguessthislldo: Odd... but alright, as long as it works.

PR #1737, opened 2020-06-11
Allow Built-in Security Plugin to be Reloaded in a Static Build

iguessthislldo:
Fix for failing to load security plugin during restarts of static builds of OpenDDS. This makes `BuiltInPluginLoader` check if it needs to act if called again instead of just acting once. Then I added that a call to the loader to the Restart test.

An unrelated issue:
The Restart test here and on master appears to have a chance to segfault at the end of an Application object.
I have put a band aid over it here by adding a second sleep after creating each Application object.

A Question:
If we're static, is it wise to try to reload the built-in security plugin automatically from `SecurityRegistry::load_security_plugin_lib`? I'm thinking `BuiltInPluginLoader` could regisiter itself with the `SecurityRegistry` when it's called the first time. As it is now the user application will have to call on `BuiltInPluginLoader` on restart to make sure the security plugin is working.


dds/DCPS/security/BuiltInPluginLoader.cpp
```diff
@@ -22,17 +23,19 @@ static const std::string PLUGIN_NAME("BuiltIn");
...
+    plugin = DCPS::make_rch<BuiltInSecurityPluginInst>();
+    TheSecurityRegistry->register_plugin(PLUGIN_NAME, plugin);
+  }
+
+  if (TheSecurityRegistry->has_no_configs()) {
+    SecurityConfig_rch default_config =
+      TheSecurityRegistry->create_config(SecurityRegistry::DEFAULT_CONFIG_NAME,
+                                         plugin);
+    TheSecurityRegistry->default_config(default_config);
+  }
```
iguessthislldo, line 38: I guess if I'm creating a new `BuiltInPluginLoader` object, then there's not much of a point of doing all this.

bin/dcps_tests.lst
```diff
@@ -123,6 +123,7 @@ tests/DCPS/Reconnect/run_test.pl restart_pub: !DCPS_MIN !OPENDDS_SAFETY_PROFILE
 tests/DCPS/Reconnect/run_test.pl bp_timeout: !DCPS_MIN !OPENDDS_SAFETY_PROFILE !DDS_NO_OWNERSHIP_PROFILE
 tests/DCPS/Reconnect/run_test.pl sub_init_crash: !DCPS_MIN !OPENDDS_SAFETY_PROFILE !DDS_NO_OWNERSHIP_PROFILE
 tests/DCPS/Restart/run_test.pl: RTPS !OPENDDS_SAFETY_PROFILE
+tests/DCPS/Restart/run_test.pl --secure: RTPS !OPENDDS_SAFETY_PROFILE
```
mitza-oci, line 126: Should this be in `security_tests.lst`?

tests/DCPS/Restart/RestartTest.cpp
```diff
@@ -74,28 +124,55 @@ struct Application {
...
-int main()
+int main(int argc, char* argv[])
 {
-  {
-    Application a1;
-    {
-      Application a2;
+  const std::string usage =
+    std::string("usage: ") + argv[0] + " [--help|-h] [--secure]\n";
+  for (int i = 1; i < argc; i++) {
```
mitza-oci, line 145:
```suggestion
  for (int i = 1; i < argc; ++i) {
```

tests/DCPS/Restart/RestartTest.cpp
```diff
@@ -74,28 +124,55 @@ struct Application {
...
+        ACE_OS::sleep(1);
+      }
+      {
+        Application a3;
+        ACE_OS::sleep(1);
+      }
     }
+    Application a4;
+    ACE_OS::sleep(1);
+  } catch (std::string error) {
```
mitza-oci, line 172:
```suggestion
  } catch (const std::string& error) {
```

tests/DCPS/Restart/RestartTest.cpp
```diff
@@ -74,28 +124,55 @@ struct Application {
...
+    ACE_ERROR((LM_ERROR, "Caught error: %C\n", error.c_str()));
```
mitza-oci, line 173:
```suggestion
    ACE_ERROR((LM_ERROR, "Caught Error: %C\n", error.c_str()));
```
(scoreboard will highlight it)

dds/DCPS/security/framework/SecurityRegistry.cpp
```diff
@@ -444,10 +447,12 @@ SecurityRegistry::get_plugin_inst(const OPENDDS_STRING& plugin_name)
     // Try to find it again
     if (find(registered_plugins_, plugin_name, plugin_inst) != 0) {
 #endif
-      ACE_ERROR((LM_ERROR,
-                 ACE_TEXT("(%P|%t) SecurityRegistry::create_inst: ")
-                 ACE_TEXT("plugin_type=%C is not registered.\n"),
-                 plugin_name.c_str()));
+      if (log_error) {
```
mitza-oci, line 450: Could this just be removed?  The error can be logged by the caller if needed (and looks like it already is).

dds/DCPS/security/BuiltInPluginLoader.cpp
mitza-oci: Please explain this a bit more

iguessthislldo: So the issue is that the security config is getting removed from the security register, but for a static build it's not getting replaced when needed. In the restart test this happens after the first three Applications are finished but before the forth one was begun. During this point there are no Applications and I guess the security registry gets cleared.

At first I thought it was because the loader could only populate the security register once, but was still getting called multiple times. After I changed it to do it when when the security registry has no configs, it still wasn't working, because the function wasn't getting called again. It was only getting called when ACE was calling it at the beginning of the program. I didn't remove the changes I made to the loader though and just added a loader with a manual call to the restart test, which does work.

If the security plugin calls this automatically though, then this is needed and will eliminate the need to add a manual call in the user application, which is my question in the description of this PR. It might be easier to make that change and show that verses trying to explain it.

bin/dcps_tests.lst
iguessthislldo: It could be. I was following the lead of Messenger which is the only other test with optional security.

tests/DCPS/Restart/RestartTest.cpp
iguessthislldo: It highlights "Error", but not "error"?

dds/DCPS/security/framework/SecurityRegistry.cpp
iguessthislldo: I didn't notice, yes I will remove it.

dds/DCPS/security/BuiltInPluginLoader.cpp
iguessthislldo: > During this point there are no Applications and I guess the security registry gets cleared.

Small correction, I'm certain now the security registry singleton is getting destroyed.

> I guess if I'm creating a new BuiltInPluginLoader object, then there's not much of a point of doing all this.

I've had a change of heart on this. The change to `BuiltInPluginLoader` isn't doing much now, but assuming I've done it right I think I've made the function more robust. For example init was called again and the builtin plugin was still registered, but there were no configs set up for some reason, then this would go ahead and make sure a config existed.

iguessthislldo:
It looks like I accidentally turned the builtin security plugin loader into a non-terminating recursive function when it's not static.

iguessthislldo:
I don't quite know what's going on Windows builds, but I don't thinks it's related to my changes.

mitza-oci:
> 
> 
> I don't quite know what's going on Windows builds, but I don't thinks it's related to my changes.

If the problems are only on VS2019 and not the other versions, I think that's also on other PRs so it's something in the Azure Pipelines image.

mitza-oci:
> 
> 
> > I don't quite know what's going on Windows builds, but I don't thinks it's related to my changes.
> 
> If the problems are only on VS2019 and not the other versions, I think that's also on other PRs so it's something in the Azure Pipelines image.

I was able to reproduce it locally with the same combination of options used by the test script.  Looks like it may be an issue in the combination of -DCPSDefaultAddress with DCPSInfoRepo's Built-In Topics

PR #1741, opened 2020-06-12
CMake: Remove Non-IDL-Related parts from OPENDDS_TARGET_SOURCES

iguessthislldo:
Moved OpenDDS library preprocessor macro definitions from `OPENDDS_TARGET_SOURCES`, which should just be IDL related, to the OpenDDS library definitions where they belong. I think this is good enough for now, but I added TODO comments about possible future improvements regarding how macros definitions should be handled.


iguessthislldo:
I took another look at `OPENDDS_TARGET_SOURCES` and saw it was calling `include_directories`. This is even more egregious than the macros because it sets include paths globally and it doesn't even really need to do this because it was already being done correctly on the library targets.

PR #1745, opened 2020-06-15
XTypes: Serialization functions for the data types in TypeObject.h

mitza-oci:


dds/DCPS/TypeObject.h
```diff
@@ -721,36 +758,36 @@ const DCPS::Encoding& get_typeobject_encoding();
   typedef Sequence<AppliedAnnotationParameter> AppliedAnnotationParameterSeq;
 
   struct AppliedAnnotation {
-    TypeIdentifierPtr                   annotation_typeid;
+    TypeIdentifier                   annotation_typeid;
```
sonndinh, line 761: There are multiple spaces between the type and field name here and several other struct definitions. I'm not sure if it is our coding convention to do that.

mitza-oci: You are correct that it's not our coding convention.  These came from the spec's IDL file and were just preserved that way when they made their way into the header (by @jrw972 I think).  We could decide to reformat now, I'd be OK with it either way.

sonndinh: I'm also good with either way. I think it's not critical at this point.

iguessthislldo: I'd like it to be fixed and am willing to fix it when I have the time.

mitza-oci: I started on it already...

iguessthislldo:
It looks like we will have to go back through all this and add code for appendable/mutable, unless this is something you're planning on doing here?

mitza-oci:
> It looks like we will have to go back through all this and add code for appendable/mutable, unless this is something you're planning on doing here?

I wanted to get the API changes in.  You're correct that we will need more implementation changes after this.  Some of it could potentially be scripted based on opendds_idl's output from processing `etc/xtypes/opendds_typeobject.idl`

PR #1748, opened 2020-06-15
Fix Unused Function Warning in Attributes Test

iguessthislldo:
Also remove negative exit statuses.

### DIFF
--- a/tests/security/attributes/publisher.cpp
+++ b/tests/security/attributes/publisher.cpp
@@ -238,7 +238,8 @@ int run_test(int argc, ACE_TCHAR *argv[], Args& my_args)
 int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
 {
   Args my_args;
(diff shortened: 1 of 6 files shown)


tests/security/attributes/publisher.cpp
```diff
@@ -238,12 +238,9 @@ int run_test(int argc, ACE_TCHAR *argv[], Args& my_args)
 int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
 {
   Args my_args;
-
-  int result = run_test(argc, argv, my_args);
-  if (result == my_args.expected_result_) {
-    return 0;
-  } else {
+  if (run_test(argc, argv, my_args) != my_args.expected_result_) {
     std::cerr << "Publisher exiting with unexpected result: " << result << std::endl;
```
simpsont-oci, line 242: looks like a compile error here... need to add result back in for the error message.

PR #1754, opened 2020-06-17
IDL4 anonymous types

lij-oci:
anonymous type support


tests/DCPS/Compiler/anonymous_types/anonymous_types.mpc
```diff
@@ -0,0 +1,7 @@
+project: dcps {
+  idlflags += -as
```
mitza-oci, line 2: This should move to a base project so all projects can inherit the setting.

tests/DCPS/Compiler/anonymous_types/anonymous_types.mpc
```diff
@@ -0,0 +1,7 @@
...
+  dcps_ts_flags += -as
```
mitza-oci, line 3: This shouldn't be needed.  Remove it, and then if the test breaks make any changes needed in the opendds_idl source code.

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,24 @@
+
```
mitza-oci, line 1: Add the line `#include <tao/LongSeq.pidl>` so the generated code can compile on Visual C++.

dds/idl/anonymous.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef anonymous_H
+#define anonymous_H
+
+#include "dds_generator.h"
+#include "utl_scoped_name.h"
+#include "ast.h"
+
+#include <string>
+
+struct Field {
```
mitza-oci, line 17: This header only defines a class called `Field`, but the header is not `field.h`.  I think we should find a different name for `Field`.  Is it used for all Fields or just sequence/array fields?

dds/idl/anonymous.cpp
```diff
@@ -0,0 +1,134 @@
...
+  ast_type(field.field_type()),
+  arr(AST_Array::narrow_from_decl(ast_type)),
+  seq(AST_Sequence::narrow_from_decl(ast_type)),
+  ast_elem(arr ? arr->base_type() : (seq ? seq->base_type() : nullptr)),
+  name(field.local_name()->get_string()),
+  cls(CL_UNKNOWN), elem_sz(0), n_elems(1)
+{
+  type = ast_elem ? ("_" + name) : name;
+  if (seq) { type += "_seq"; }
+  std::string ftn = scoped(ast_type->name());
```
mitza-oci, line 50:
```suggestion
  const std::string ftn = scoped(ast_type->name());
```

dds/idl/anonymous.cpp
```diff
@@ -0,0 +1,134 @@
...
+}
+
+bool Field::is_anonymous_type(AST_Type& field)
+{
+  return is_anonymous_array(field) || is_anonymous_sequence(field);
+}
+
+std::string Field::get_anonymous_type_name(const std::string& scoped_name)
+{
+  std::size_t i = scoped_name.find("::") + 2;
```
mitza-oci, line 29:
```suggestion
  const std::size_t i = scoped_name.find("::") + 2;
```

dds/idl/anonymous.cpp
```diff
@@ -0,0 +1,134 @@
...
+  return scoped_name.substr(0, i) + "_" + scoped_name.substr(i);
+}
+
+std::string Field::get_type_name(AST_Type& field)
+{
+  std::string n = scoped(field.name());
```
mitza-oci, line 35:
```suggestion
  const std::string n = scoped(field.name());
```

dds/idl/anonymous.h
```diff
@@ -0,0 +1,51 @@
...
+  static bool is_anonymous_array(AST_Type& field);
+  static bool is_anonymous_sequence(AST_Type& field);
+  static bool is_anonymous_type(AST_Type& field);
+  static std::string get_anonymous_type_name(const std::string& scoped_name);
+  static std::string get_type_name(AST_Type& field);
+
+  AST_Type* ast_type;
```
mitza-oci, line 24: guidelines are that these should end in an underscore

dds/idl/anonymous.h
```diff
@@ -0,0 +1,51 @@
...
+  std::string elem;
+  std::size_t n_elems;
+  std::string length;
+  std::string ref;
+  std::string const_ref;
+  std::string arg;
+  std::string unwrap;
+  std::string const_unwrap;
+
+  Field(AST_Field& field); //for anonymous types
```
mitza-oci, line 44:
```suggestion
  explicit Field(AST_Field& field); //for anonymous types
```

dds/idl/metaclass_generator.cpp
```diff
@@ -64,19 +65,15 @@ namespace {
...
   {
     const size_t n = fieldName.size() + 1 /* 1 for the dot */;
-    const std::string fieldType = scoped(field->field_type()->name());
+    const std::string fieldType = Field::get_type_name(*(field->field_type()));
     be_global->impl_ <<
-      "    if (std::strncmp(field, \"" << fieldName << ".\", " << n
-      << ") == 0) {\n"
-      "      return getMetaStruct<" << fieldType << ">().getValue("
-      << firstArg << ", field + " << n << ");\n"
+      "    if (std::strncmp(field, \"" << fieldName << ".\", " << n << ") == 0) {\n"
```
mitza-oci, line 70: revert changes that only re-arrange the existing code (change how it's broken into different lines)

dds/idl/metaclass_generator.cpp
```diff
@@ -64,19 +65,15 @@ namespace {
     const std::string& firstArg, bool skip = false)
   {
     const size_t n = fieldName.size() + 1 /* 1 for the dot */;
-    const std::string fieldType = scoped(field->field_type()->name());
+    const std::string fieldType = Field::get_type_name(*(field->field_type()));
```
mitza-oci, line 68:
```suggestion
    const std::string fieldType = Field::get_type_name(*field->field_type());
```

dds/idl/anonymous.h
lij-oci: It was AnonymousField for anonymous sequence/array. Now it handles non-anonymous fields so the generation code (marshal, meta class, etc) can be shared. And hopefully, the c++11 mapping can be done in one place here for all.

tests/DCPS/Compiler/anonymous_types/anonymous_types.mpc
iguessthislldo: These lines were a workaround until https://github.com/DOCGroup/ACE_TAO/pull/1135 was merged. Since it was merged both lines can be removed if using DOCGroup master.

tests/DCPS/Compiler/anonymous_types/anonymous_types.mpc
iguessthislldo: Same as above.

dds/idl/anonymous.h
```diff
@@ -0,0 +1,51 @@
...
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef anonymous_H
+#define anonymous_H
+
+#include "dds_generator.h"
+#include "utl_scoped_name.h"
+#include "ast.h"
```
iguessthislldo, line 13:
```suggestion

#include <utl_scoped_name.h>
#include <ast.h>
```

Since these are in another directory.

tests/DCPS/Compiler/anonymous_types/anonymous_types.mpc
mitza-oci: We should still have `-as` for tao_idl since some versions will need it.

dds/idl/anonymous.h
mitza-oci: My comment about the file name vs. the class name stands.

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,40 @@
...
+  short a_s3arr[3];
+  short a_s3arr_2nd[3];
+  short a_s2x3arr[2][3];
+  double a_d2x3x4arr[2][3][4];
+
+  sequence<long> a_lseq;
+  sequence<long> a_lseq_2nd;
+  sequence<long, 3> a_l3seq;
+  sequence<long, 3> a_l3seq_2nd;
+  sequence<long, 5> a_l5seq;
```
jwillemsen, line 39: bounded strings seem to be missing as test case

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,45 @@
...
+  string<200> a_bstr;
```
jwillemsen, line 44: Would recommend to also add a test case which the anonymous type is used for a IDL keyword and a C++ keyword to make sure all escaping is done correctly

mitza-oci:
Before doing too much more with C++11, I suggest getting the C++03 build running (see Azure Pipelines Linux GCC6_CPP03):
```
/home/vsts/work/1/s/dds/idl/field_info.cpp:54:69: error: ‘nullptr’ was not declared in this scope
   as_base_ = arr_ ? arr_->base_type() : (seq_ ? seq_->base_type() : nullptr);
                                                                     ^~~~~~~
/home/vsts/work/1/s/dds/idl/field_info.cpp:81:15: error: ‘to_string’ is not a member of ‘std’
     length_ = std::to_string(n_elems_);
```

MPC/config/dcps_ts_defaults.mpb
```diff
@@ -31,7 +31,7 @@ project : dds_macros, taobaseidldefaults {
   // Also To avoid multiple unnecessary warnings, have tao_idl ignore
```
iguessthislldo, line 31:
```suggestion
  // Also to avoid multiple unnecessary warnings, have tao_idl ignore
```

This is my mistake, but it'd be great if you could fix it.

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,116 @@
...
+  Long3Sequence l3seq;
+  BoundedString bstr;
+  XYZSequence xyzseq;
+
+  // prefix a_ denotes anonymous type fields
+
+  short a_s3arr[3];
+  short a_s3arr_2nd[3];
+  short a_s2x3arr[2][3];
+  double a_d2x3x4arr[2][3][4];
```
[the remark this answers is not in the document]
lij-oci: I changed Line 48 to:   @key short a_s2x3arr[2][3]; and it worked. 
When I changed Line 53 to:  @key sequence<long, 3> a_l3seq; the build process stuck. I'm working on it. Any help would be appreciated.

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,202 @@
...
+    case AST_PredefinedType::PT_ulong: size = 4; return "ACE_CDR::ULong";
+    case AST_PredefinedType::PT_longlong: size = 8; return "ACE_CDR::LongLong";
+    case AST_PredefinedType::PT_ulonglong: size = 8; return "ACE_CDR::ULongLong";
+    case AST_PredefinedType::PT_short: size = 2; return "ACE_CDR::Short";
+    case AST_PredefinedType::PT_ushort: size = 2; return "ACE_CDR::UShort";
+    case AST_PredefinedType::PT_float: size = 4; return "ACE_CDR::Float";
+    case AST_PredefinedType::PT_double: size = 8; return "ACE_CDR::Double";
+    case AST_PredefinedType::PT_longdouble: size = 16; return "ACE_CDR::LongDouble";
+    case AST_PredefinedType::PT_char: size = 1; return "ACE_CDR::Char";
+    case AST_PredefinedType::PT_wchar: size = 1; return "ACE_CDR::WChar"; // encoding of wchar length is 1 byte
```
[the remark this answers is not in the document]
lij-oci: Yeah, at first I thought it was a typo, but the comment indicated otherwise. 
      case AST_PredefinedType::PT_wchar:
        size = 1; // encoding of wchar length is 1 byte
        return "ACE_CDR::WChar";

Then later I forgot to ask this question. @mitza-oci could you please verify this?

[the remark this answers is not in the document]
lij-oci: Adam confirmed it's correct.

tests/DCPS/Compiler/anonymous_types/test.idl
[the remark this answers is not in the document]
lij-oci: Since sequence is not used as key, this is considered resolved.

dds/idl/field_info.cpp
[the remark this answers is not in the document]
mitza-oci: See Serializer.inl `operator<<(Serializer&, ACE_OutputCDR::from_wchar)`
Note that the comment is talking about how the "length of the wchar" is encoded (not the wchar itself)

tests/DCPS/Compiler/anonymous_types/test.idl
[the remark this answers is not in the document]
iguessthislldo: Yes, sequences can't be key as far as I know.

dds/idl/field_info.cpp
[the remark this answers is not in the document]
iguessthislldo: I know about the length byte because I removed it in #1668. I guess a better question to ask is how is this size of one being used? Right now glancing at the code it looks like we don't use it, so unless I missing something, so why are we saying the size is 1 byte?

Also I'm this is something I missed in #1668, so more work probably needs to be done on the two byte wchar in XTypes, at least in `metaclass_generator::gen_typedef`.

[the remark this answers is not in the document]
mitza-oci: It is used in gen_field_getValueFromSerialized (metaclass generator).  Scalar types that are not strings but are wide (that is, wchar) have an additional block of code generated for "skip over."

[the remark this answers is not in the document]
iguessthislldo: It doesn't use size there, I want to know why does `to_cxx_type` set the size of wchar at all.

Going off of `metaclass_generator.cpp` on master, this is how call the functions that can call `to_cxx_type` behave when the type is a scalar wchar:

- `gen_field_getValueFromSerialized` calls `to_cxx_type` doesn't use size unless it's a scalar that's not a string or wide.
- `gen_typedef` doesn't call  `to_cxx_type`.
- The `gen_union` callback for `generateSwitchForUnion`, `func` also doesn't call `to_cxx_type`.

So the answer is that it doesn't use the `size = 1` at all. It might have been used (or envisioned to be used) at some point, but it's confusing because you have to get the size of the wchar regardless. It looks like it does a similar thing for string, but it doesn't look like that's used either. Both of these should act like the default case, which doesn't set the size because the size isn't actually known.

XTypes will fix this by making wchar a fixed primitive again, so it will need the size, but I say the `size = 4` should be removed from the string case. Maybe the size information should be broken out into a `scalar_primitive_serialized_size` function.

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,116 @@
...
+struct StructOfMultiDimArray {
+  Str2x3x4x5Array str2x3x4x5arr;
+  Short3Array_x2 s3arr_x2arr;
+};
+
+typedef sequence<long> LongSequence;
+typedef sequence<long, 3> Long3Sequence;
+typedef string<100> BoundedString;
+
+typedef struct XYZ_ {
```
mitza-oci, line 19: Does this need to be "typedef struct" instead of just "struct" in order to test something in the compiler?
Otherwise just make it a normal struct.

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,116 @@
...
+  double x;
+  double y;
+  double z;
+} XYZ;
+
+typedef sequence<XYZ> XYZSequence;
+
+@topic
+struct TestStruct {
+  char  c;
```
mitza-oci, line 29:
```suggestion
  char c;
```

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,116 @@
...
+
+  //test C++ keyword used as anonymous type field
+  //to see all escaping is done correctly
+  sequence<long, 3> _long; // _tao_seq_CORBA_Long__3 _cxx_long;
+  sequence<long, 3> break; // _tao_seq_CORBA_Long__3 _cxx_break;
+  sequence<long, 3> class; // _tao_seq_CORBA_Long__3 _cxx_class;
+};
+
+// ========== ========== ========== ========== ==========
+//from Relay.idl
```
mitza-oci, line 67: Why is this here?  It doesn't seem to use anonymous IDL types.  Is it needed to test other features or compatibility of the generated code?

dds/idl/field_info.h
```diff
@@ -0,0 +1,70 @@
...
+#include "dds_generator.h"
+#include <utl_scoped_name.h>
+#include <ast.h>
+
+#include <string>
+
+struct FieldInfo {
+  struct EleLen {
+    AST_Type* ele_;
+    std::size_t len_;
```
mitza-oci, line 20: I'm not sure this is correct for multidimensional arrays.  Is there a need to distinguish among `octet[2][3]`, `octet[3][2]`, `octet[6]`, etc.?  Please add some of these kinds of cases to the test.

tests/DCPS/Compiler/anonymous_types/test.idl
lij-oci: We have quite a few normal structs, StructOfMultiDimArray, EntityId_t, GUID_t, RelayHeader, and all the topic structs. Here I wanted to have some more coverage and to see where XYZ_ and XYZ appear in the generated code. 

I created struct NS and sequence<NS, 5> a_ns5seq to test a normal struct used in an anonymous field.

dds/idl/field_info.h
lij-oci: The screening set EleLenSet is not used for anonymous arrays. It's true that EleLenSet would be incorrect to screen multidimensional anonymous arrays. We would need something like a string to store 2x3 or 3x2, etc. Right now EleLenSet is only used to screen anonymous sequences. I will change the variable name accordingly as you pointed out in another comment.

dds/idl/marshal_generator.h
```diff
@@ -25,6 +26,8 @@ class marshal_generator : public dds_generator {
                  const std::vector<AST_UnionBranch*>& branches,
                  AST_Type* discriminator,
                  const char* repoid);
+private:
+  FieldInfo::EleLenSet eleLen_;
```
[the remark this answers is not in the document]
lij-oci: changed to anonymous_seq_generated_

dds/idl/field_info.h
mitza-oci: Got it, I must have misread that part of how it's used.

tests/DCPS/Compiler/anonymous_types/test.idl
lij-oci: Yes, the Relay.idl caught most errors that the anonymous type support code messed up the non-anonymous type support code. If we want to further factor out the code for both anonymous and non-anonymous type support, we need to test at least this much of the original Relay.idl to make sure they both work correctly. 

I just added the following in module RtpsRelay to test anonymous fields in module and topic:
  // test anonymous fields in module and topic
  @topic
  struct AnonymousFields {
    @key GUID_t guid;
    @key short a_s2x3arr[2][3];
    sequence<long, 5> a_l5seq;
  };

dds/idl/field_info.cpp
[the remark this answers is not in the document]
mitza-oci: Yes, I see the point with 'size' here being potentially confusing.  Some of that is beyond the scope of this PR.

For this PR what I'm going to suggest is to move `to_cxx_type` back to `metaclass_generator.cpp` since that is the only place it is used.  Each `*_generator.cpp` is its own "pass" over the AST that has its own local concerns.  These don't need to be moved to other files unless they are needed by multiple generators.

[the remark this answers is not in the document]
lij-oci: to_cxx_type is used in FieldInfo to set elem_ and elem_sz_. The duplicated method set_element() is removed, since to_cxx_type is used for both Field and element. That's the reason to_cxx_type is taken out of its original compilation unit scope. If we factor out all the common code and use FieldInfo wherever appropriate, to_cxx_type may not need to be static. Also we would have all the c++11 (or future) checks wrapped in one class.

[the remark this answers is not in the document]
mitza-oci: I went through it yesterday and tracked where the outputs of to_cxx_type were used, and it was only for metaclass_generator.  That resulted in my request that it be moved back there.

tests/DCPS/Compiler/anonymous_types/test.idl
iguessthislldo: I think you should boil down the IDL causing the error you were seeing and try to replicated them using the generic test IDL.

azure-pipelines.yml
```diff
@@ -9,7 +9,7 @@ schedules:
 
 jobs:
 - job: VisualStudio2019
-  timeoutInMinutes: 90
+  timeoutInMinutes: 120
```
mitza-oci, line 12: These changes can be reverted now that we know what the problem was

tests/DCPS/Compiler/anonymous_types/test.idl
```diff
@@ -0,0 +1,155 @@
+#include <tao/LongSeq.pidl>
```
mitza-oci, line 1: Add the equivalent `#include` for ShortSeq

mitza-oci:
```
tests\DCPS\Compiler\anonymous_types\cpp11\testTypeSupportImpl.cpp(0,0): Error C1128: number of sections exceeded object file format limit: compile with /bigobj
```
You can use the `msvc_bigobj` MPC base project.

dds/idl/marshal_generator.cpp
```diff
@@ -1646,6 +2044,18 @@ bool marshal_generator::gen_struct(AST_Structure* node,
     }
   }
 
+  FieldInfo::EleLenSet anonymous_seq_generated;
+  for (size_t i = 0; i < fields.size(); ++i) {
+    if (fields[i]->field_type()->anonymous()) {
+      FieldInfo af(*(fields[i]));
```
mitza-oci, line 2050:
```suggestion
      FieldInfo af(*fields[i]);
```

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+  init();
+}
+
+void FieldInfo::init()
+{
+  act_ = resolveActualType(type_);
+  cls_ = classify(act_);
+  arr_ = AST_Array::narrow_from_decl(type_);
+  seq_ = AST_Sequence::narrow_from_decl(type_);
```
iguessthislldo, line 89:
```suggestion
  arr_ = dynamic_cast<AST_Array*>(type_);
  seq_ = dynamic_cast<AST_Sequence*>(type_);
```
TAO_IDL narrows are just dynamic casts, so the actual cast should be used when possible. Also you're resolving the type of `type_`, but not using it here?

dds/idl/langmap_generator.cpp
```diff
@@ -1379,6 +1387,27 @@ struct Cxx11Generator : GeneratorBase
       "using " << nm << " = " << array << elem_type << bounds.str() << ";\n";
   }
 
+  void gen_anonymous_type(FieldInfo& af)
```
iguessthislldo, line 1390: This function share the same basic elements with `gen_array` and `gen_sequence`. Maybe create a `gen_array_i` and `gen_sequence_i` from the common parts?

dds/idl/marshal_generator.cpp
```diff
@@ -759,7 +971,7 @@ namespace {
       if (elem_cls & CL_PRIMITIVE) {
         string suffix;
         for (unsigned int i = 1; i < arr->n_dims(); ++i)
-          suffix += use_cxx11 ? accessor : "[0]";
+          suffix += use_cxx11 ? "->data()" : "[0]";
```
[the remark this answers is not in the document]
lij-oci: The first operator is a dot operator. That's defined in Line 970:
      const std::string accessor = use_cxx11 ? ".data()" : ".in()";
and the actual writing out is in Line 977:
          << "_array(arr" << accessor << suffix << ", " << n_elems << ");\n";

The result is like this:
  return strm.write_short_array(arr.data()->data()->data(), 24);

dds/idl/metaclass_generator.cpp
```diff
@@ -116,70 +115,36 @@ namespace {
...
     const Classification cls = classify(type);
-    if (cls & CL_ENUM) {
-      size = 4;
-      return "ACE_CDR::ULong";
-    }
-    if (cls & CL_STRING) {
-      size = 4; // encoding of str length is 4 bytes
-      return string_type(cls);
-    }
+    if (cls & CL_ENUM) { size = 4; return "ACE_CDR::ULong"; }
```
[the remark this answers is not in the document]
lij-oci: I agree. These two compilation-unit-local functions to_cxx_type() and string_type() were once moved to FieldInfo and modified there to support sharing code in both typedef and anonymous arrays and sequences. When this approach encountered a lot of problems, separate functions gen_anonymous_array() and gen_anonymous_sequence() were made in the marshal_generator, and later to_cxx_type() and string_type() were moved back to their original places. At this point, the changes made in the parameter type size_t, the temporary variable AST_Type* t, and the throw statement are still valid, but the reformatting changes just add unnecessary work in merging and code review. They are changed back.

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+bool FieldInfo::is_new(EleLenSet& el_set)
+{
+  return cxx11() || el_set.insert(EleLen(*this)).second;
+}
+
+bool FieldInfo::anonymous_array() const
+{
+  return type_->anonymous() && arr_ && (cls_ & CL_ARRAY);
```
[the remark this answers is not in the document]
lij-oci: I would think arr_ is sufficient. During the development, these 3 methods anonymous_array(), anonymous_sequence(), and anonymous() were factored out from different places, where some checked the non-null pointer and some checked classification constant also. The AST_Type::anonymous() was seen giving the incorrect values in some place. In order to fix those problems, I replaced most of such checks with one of these 3 (probably over-kill) checking methods. Since this is the code-generation code, not the runtime application code, an extra comparison won't hurt much.

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,168 @@
...
+      scoped_type_ = struct_name_ + "::" + type_name_;
+      underscored_ = underscore(scoped_type_);
+    } else {
+      type_name_ = scoped_type_.substr(i + 2);
+    }
+  } else {
+    if (scoped_type_.empty()) {
+      scoped_type_ = scoped(type_->name());
+    }
+    //name_
```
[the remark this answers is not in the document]
lij-oci: changed to     //name_ is empty
This is a note for the typedef arrays and sequences, in case future changes or merging may use this field in more placed than where currently used.

mitza-oci:
```
The job running on agent Azure Pipelines 2 ran longer than the maximum time of 90 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134
```
Do we need to adjust the timeout?

dds/idl/langmap_generator.cpp
```diff
@@ -1463,6 +1492,18 @@ struct Cxx11Generator : GeneratorBase
     const char* const nm = name->last_component()->get_string();
     gen_common_strunion_pre(nm);
 
+    bool has_anonymous_field = false;
+    for (size_t i = 0; i < fields.size(); ++i) {
```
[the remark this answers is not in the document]
lij-oci: Resolved the first part (process anonymous in gen_struct_members).

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,168 @@
...
+  return s;
+}
+
+std::string FieldInfo::ref(const std::string& scoped_t, const std::string& underscored_t, const std::string& const_s)
+{
+  return "IDL::DistinctType<" + const_s + scoped_t + ", " + underscored_t + "_tag>";
+}
+
+// for anonymous types
+FieldInfo::FieldInfo(AST_Field& field) :
```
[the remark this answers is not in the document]
mitza-oci: One example is in GuidBuilder.cpp
```
GuidBuilder::GuidBuilder()
  : guid_cxx_(create())
  , guid_(guid_cxx_)
{
```

dds/idl/field_info.cpp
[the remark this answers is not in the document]
lij-oci: Removed anonymous_array() and anonymous_sequence().

dds/idl/field_info.cpp
lij-oci: The cast changes were decided as optional in a meeting with Fred and Adam. Both act_, the resolved the type of type_, and its classification cls_ are used in generators.

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+const std::string FieldInfo::scope_op = "::";
+
+bool FieldInfo::cxx11()
+{
+  return be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+}
+
+std::string FieldInfo::at_pfx() {
+  return cxx11() ? "T_" : "_";
```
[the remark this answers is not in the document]
lij-oci: The prefix T_ is replaced with AnonymousType_.

dds/idl/langmap_generator.cpp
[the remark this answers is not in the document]
lij-oci: A Trello card is created for "Support for IDL4 Anonymous Types in union".

dds/idl/langmap_generator.cpp
lij-oci: Factored out gen_array and gen_sequence; removed gen_anonymous_type.

dds/idl/metaclass_generator.cpp
```diff
@@ -178,8 +178,7 @@ namespace {
       case AST_PredefinedType::PT_octet:
         size = 1;
         return "ACE_CDR::Octet";
-      default:
-        break;
+      default: throw std::invalid_argument("Unknown PRIMITIVE type");
```
mitza-oci, line 181:
```suggestion
      default:
        throw std::invalid_argument("Unknown PRIMITIVE type");
```

dds/idl/metaclass_generator.cpp
```diff
@@ -248,11 +247,11 @@ namespace {
         pre = "IDL::DistinctType<";
         post = ", " + dds_generator::scoped_helper(type->name(), "_") + "_tag>";
       }
+      std::string ptr = field->field_type()->anonymous() ?
```
mitza-oci, line 250:
```suggestion
      const std::string ptr = field->field_type()->anonymous() ?
```

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const std::string FieldInfo::scope_op = "::";
+
+bool FieldInfo::cxx11()
+{
+  return be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+}
+
+std::string FieldInfo::at_pfx() {
```
mitza-oci, line 30:
```suggestion
std::string FieldInfo::at_pfx()
{
```

dds/idl/field_info.cpp
```diff
@@ -0,0 +1,115 @@
...
+ * See: http://www.opendds.org/license.html
+ */
+
+#include "field_info.h"
+
+#include <sstream>
+
+using namespace AstTypeClassification;
+
+FieldInfo::EleLen::EleLen(const FieldInfo& af) : ele_(af.as_base_), len_(af.n_elems_)
```
mitza-oci, line 14:
```suggestion
FieldInfo::EleLen::EleLen(const FieldInfo& af)
  : ele_(af.as_base_)
  , len_(af.n_elems_)
```

PR #1768, opened 2020-06-29
preventing min / max  macro expansion

kuznetsovmoci:


iguessthislldo:
I don't think this should be necessary because ACE should be telling Windows not to define these macros. I would much rather see a real fix than have a any work around like this.

mitza-oci:
> 
> 
> I don't think this should be necessary because ACE should be telling Windows not to define these macros. I would much rather see a real fix than have a any work around like this.

I'm still tracking down why Mike's build was different, but I'd rather us be conservative about this when it's in a public header.

PR #1777, opened 2020-07-06
Various Coverity Fixes

iguessthislldo:
Attempts to fix the last four medium impact issues reported by Coverity. Prompted by Coverity issue caused by changes in #1737.

### DIFF
--- a/performance-tests/bench_2/builder/Common.cpp
+++ b/performance-tests/bench_2/builder/Common.cpp
@@ -79,9 +79,11 @@ bool operator==(const TimeStamp& lhs, const TimeStamp& rhs)
 std::ostream&
 operator<<(std::ostream& out, const TimeStamp& ts)
 {
-  std::streamsize ssize = out.precision();
+  const std::ios_base::fmtflags out_flags(out.flags());
+  const std::streamsize ssize = out.precision();
   out << std::setprecision(3) << std::fixed <<
     static_cast<double>(ts.sec) + (static_cast<double>(ts.nsec) / 1.0e9) << std::setprecision(ssize) << std::flush;
+  out.flags(out_flags);
   return out;
 }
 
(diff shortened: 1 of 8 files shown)


performance-tests/bench_2/builder/Common.cpp
```diff
@@ -79,9 +79,11 @@ bool operator==(const TimeStamp& lhs, const TimeStamp& rhs)
 std::ostream&
 operator<<(std::ostream& out, const TimeStamp& ts)
 {
-  std::streamsize ssize = out.precision();
+  const std::ios_base::fmtflags out_flags(out.flags());
+  const std::streamsize ssize = out.precision();
   out << std::setprecision(3) << std::fixed <<
     static_cast<double>(ts.sec) + (static_cast<double>(ts.nsec) / 1.0e9) << std::setprecision(ssize) << std::flush;
+  out.flags(out_flags);
```
iguessthislldo, line 86: It'd be nice if there was an RAII way of saving and restoring format state of the `ostream`.

mitza-oci: I think it there's one in dds/DCPS/DataSampleHeader.cpp

iguessthislldo: Cool. I was going to move that to a common but I found actually there's a `dds/DCPS/RestoreOutputStreamState.h`. It doesn't do fill and both that and the one in `DataSampleHeader.cpp` don't do precision. I update `RestoreOutputStreamState.h` and use it in both places.

simpsont-oci: I agree RAII would be nice. In general, we should probably revisit the builder library & bench's use of timestamps, since the goal was never to have a separate set of classes / functions... I just needed something that was easy to serialize for coordinating process timing between machines. If we can get that kind of functionality out of OpenDDS types, we can remove a bunch of this.

PR #1780, opened 2020-07-08
Correctly handle duplicate underlying array types within idl2jni

simpsont-oci:


java/idl2jni/codegen/im_jni.cpp
```diff
@@ -19,6 +19,26 @@
...
+    return str + "_var";
+  }
+
+  return str;
+}
+
+//add in "_forany" if it's not already there
+string append_forany(const string &str)
+{
+  if (str.size() <= 4 || str.substr(str.size() - 7) != "_forany") {
```
mitza-oci, line 35:
```suggestion
  if (str.size() <= 7 || str.substr(str.size() - 7) != "_forany") {
```

PR #1688, opened 2020-05-13
XTypes: Data Representation Refactor

ClaytonCalabrese:
Currently assumes empty annotation is all data representations are valid and empty qos is XCDR1.  This may change depending on our design choices.  This PR should also fix a multitude of tests.

PR #1781, opened 2020-07-08
XTypes: Encoding decoding modes

sonndinh:


dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,40 @@ DataWriterImpl_T
...
-          nonencap_kind != Encoding::KIND_UNKNOWN) {
-        break;
+    } else { // Non-encapsulation
+      for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
+        if (repr_to_encoding_kind(repIds[i], encoding_kind) &&
+            Encoding::KIND_UNALIGNED_CDR == encoding_kind) {
+          encoding_mode_ = EncodingMode(encoding_kind, swap_bytes());
+          success = true;
+          break;
+        }
```
iguessthislldo, line 365: So the problem with this (and the code from before) is tied to the "How does Data Representation QoS and Annotations work with Unaligned CDR?" ticket, which is deciding on how much to involved unaligned CDR in the XTypes system. Right now the default representation QoS is just XCDR, which is what the spec says it should be, but that doesn't match how OpenDDS actually works. This means this code breaks all the tests that require unaligned CDR as none of them specify they need it. This is the reason the code from before was wrapped in a `#if 0`, so the tests wouldn't break when merged with #1688.

For now just change it to always be `Encoding::KIND_UNALIGNED_CDR` if there's no encapsulation.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,40 @@ DataWriterImpl_T
...
+    DDS::DataRepresentationIdSeq repIds = get_effective_data_rep_qos(qos_.representation.value);
+    bool success = false;
+    Encoding::Kind encoding_kind;
+    if (cdr_encapsulation()) {
+      for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
+        // TODO(sonndinh): Do we need to return failure when encountering an invalid representation
+        if (repr_to_encoding_kind(repIds[i], encoding_kind) &&
+            (Encoding::KIND_XCDR1 == encoding_kind ||
+             Encoding::KIND_XCDR2 == encoding_kind)) {
+          encoding_mode_ = EncodingMode(encoding_kind, swap_bytes());
```
iguessthislldo, line 353: Maybe set `encoding_kind` instead and move this down below the `success` check, so it's shared with the non-encapsulated case?

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,40 @@ DataWriterImpl_T
...
-      get_effective_data_rep_qos(qos_.representation.value);
-    for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
-      const DDS::DataRepresentationId_t repr = repIds[i];
-      const Encoding::Kind kind =
-        repr_ext_to_encoding_kind(repr, MarshalTraitsType::extensibility());
-      switch (kind) {
-      case Encoding::KIND_CDR_UNALIGNED:
-        if (nonencap_kind == Encoding::KIND_UNKNOWN) {
-          nonencap_kind = kind;
+    DDS::DataRepresentationIdSeq repIds = get_effective_data_rep_qos(qos_.representation.value);
```
iguessthislldo, line 344:
```suggestion
    const DDS::DataRepresentationIdSeq repIds = get_effective_data_rep_qos(qos_.representation.value);
```

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,40 @@ DataWriterImpl_T
...
+    bool success = false;
+    Encoding::Kind encoding_kind;
+    if (cdr_encapsulation()) {
+      for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
+        // TODO(sonndinh): Do we need to return failure when encountering an invalid representation
```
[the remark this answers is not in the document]
sonndinh: Okay. May the data representations in `qos_.representation.value` contains some invalid values different than XCDR, XCDR2, XML, and UNALIGNED?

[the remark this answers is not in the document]
iguessthislldo: It shouldn't, but I don't think it hurts to assume it's possible.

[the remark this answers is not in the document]
sonndinh: In that case, should it stop and return error, or just skip the invalid data representations and keep going?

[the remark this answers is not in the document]
iguessthislldo: It looks like my code returned an error, but I don't think it matters much as long as it can differentiate unsupported valid representations from completely invalid representations.

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: Actually `repr_to_encoding_kind` is setting the encoding kind, so just break.

sonndinh: Per your review above for the non-encapsulated case, the for-loop is removed and there is no duplicate code now.

dds/DCPS/DataWriterImpl_T.h
sonndinh: It is using unaligned CDR without iterating the list of data representations now.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3282,56 @@ DataReaderImpl::get_ice_endpoint()
   return TransportClient::get_ice_endpoint();
 }
 
+DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
```
iguessthislldo, line 3285: This still needs to be called. See where `DataWriterImpl` is calling `setup_serialization`.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3282,56 @@ DataReaderImpl::get_ice_endpoint()
...
+    for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
+      if (repr_to_encoding_kind(repIds[i], encoding_kind)) {
+        if (Encoding::KIND_XCDR1 == encoding_kind ||
+            Encoding::KIND_XCDR2 == encoding_kind) {
+          if (decoding_modes_.find(encoding_kind) == decoding_modes_.end()) {
+            Encoding encoding(encoding_kind, swap_bytes());
+            decoding_modes_.insert(std::make_pair(encoding_kind, encoding));
+            success = true;
+          }
+        } else if (::OpenDDS::DCPS::DCPS_debug_level >= 2) {
```
iguessthislldo, line 3301: Since this code isn't in a template anymore, there's no reason for the full name. Same for the other log level references.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3282,56 @@ DataReaderImpl::get_ice_endpoint()
...
+  if (!success) {
+    if (::OpenDDS::DCPS::DCPS_debug_level) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                 ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                 ACE_TEXT("Could not find a valid data representation\n")));
+    }
+    return DDS::RETCODE_ERROR;
+  }
+
+  // TODO(sonndinh): Do we need to set up allocator? I guess no.
```
iguessthislldo, line 3330: Remove?

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3282,56 @@ DataReaderImpl::get_ice_endpoint()
...
+        // Invalid data representation
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                   ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                   ACE_TEXT("Encounter invalid data representation\n")));
+        return DDS::RETCODE_ERROR;
+      }
+    }
+  } else {
+    Encoding encoding(Encoding::KIND_UNALIGNED_CDR, swap_bytes());
+    decoding_modes_.insert(std::make_pair(Encoding::KIND_UNALIGNED_CDR, encoding));
```
iguessthislldo, line 3318: This still needs to be used in decoding.

dds/DCPS/DataReaderImpl.h
```diff
@@ -566,7 +566,10 @@ class OpenDDS_Dcps_Export DataReaderImpl
 
   virtual ICE::Endpoint* get_ice_endpoint();
 
-protected:
+  // TODO(sonndinh): public method?
```
iguessthislldo, line 569: No I don't think it needs to be public.

sonndinh: I changed it to protected. `setup_serialization` in `DataWriterImpl_T.h` is currently public. Maybe we should also change it to protected or private if it's not going to be called from outside?

iguessthislldo: Sure that sounds fine. Also change the pure virtual in DataWriterImpl.h, if it's public.

sonndinh: Actually it is called by some tests. So I'm gonna leave it as public.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,54 @@ DataReaderImpl::get_ice_endpoint()
...
+          ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+                     ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                     ACE_TEXT("Skip %C data representation\n"),
+                     Encoding::kind_to_string(encoding_kind).c_str()));
+        }
+      } else if (DCPS_debug_level) {
+        // Invalid data representation
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                   ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                   ACE_TEXT("Encounter invalid data representation\n")));
```
mitza-oci, line 3317:
```suggestion
                   ACE_TEXT("Encountered invalid data representation\n")));
```

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,54 @@ DataReaderImpl::get_ice_endpoint()
...
+          }
+        } else if (DCPS_debug_level >= 2) {
+          //Valid but incompatible data representation
+          ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+                     ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                     ACE_TEXT("Skip %C data representation\n"),
+                     Encoding::kind_to_string(encoding_kind).c_str()));
+        }
+      } else if (DCPS_debug_level) {
+        // Invalid data representation
```
mitza-oci, line 3314:
```suggestion
```
log message is already telling us this

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,54 @@ DataReaderImpl::get_ice_endpoint()
...
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                   ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                   ACE_TEXT("Encounter invalid data representation\n")));
+        return DDS::RETCODE_ERROR;
+      }
+    }
+  } else {
+    Encoding encoding(Encoding::KIND_UNALIGNED_CDR, swap_bytes());
```
mitza-oci, line 3322:
```suggestion
    const Encoding encoding(Encoding::KIND_UNALIGNED_CDR, swap_bytes());
```

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,54 @@ DataReaderImpl::get_ice_endpoint()
   return TransportClient::get_ice_endpoint();
 }
 
+DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
+{
+  const DDS::DataRepresentationIdSeq repIds =
+    get_effective_data_rep_qos(qos_.representation.value);
+  bool success = false;
+  if (cdr_encapsulation()) {
+    Encoding::Kind encoding_kind;
```
mitza-oci, line 3296: could this object's scope be reduced (to inside the for loop)?

sonndinh: Sure.

dds/DCPS/DataReaderImpl.cpp
sonndinh: Fixed

dds/DCPS/DataReaderImpl.cpp
sonndinh: Removed.

dds/DCPS/DataReaderImpl.cpp
sonndinh: Updated.

dds/DCPS/DataReaderImpl.cpp
sonndinh: I added a few lines in `dds_demarshal` to check if the encoding kind of the received data sample also appears in `decoding_modes_`. Please let me know if that's all we need to do with `decoding_modes_` or if there is any other logic we need to do with it.

iguessthislldo: `lookup_instance` will also need to be updated. In addition to that I think the `decoding_modes_` can just be a `std::set<Encoding::Kind>`, since the actual `Encoding` object isn't going to be used.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -993,6 +993,15 @@ namespace OpenDDS {
       if (!encap.to_encoding(encoding, MarshalTraitsType::extensibility())) {
         return;
       }
+
+      if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
+          ACE_TEXT("%CDataReaderImpl::dds_demarshal: ")
+          ACE_TEXT("Encoding kind of the received sample does not ")
```
iguessthislldo, line 1000: Can you say what the encoding is?

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -993,6 +993,15 @@ namespace OpenDDS {
...
+          ACE_TEXT("match the ones specified by DataReader.\n"),
+          TraitsType::type_name()));
+        return;
+      }
```
iguessthislldo, line 1004: Can you put in a debug message saying what encoding is being used for the deserialization?

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,53 @@ DataReaderImpl::get_ice_endpoint()
...
+  }
+  if (!success) {
+    if (DCPS_debug_level) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                 ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                 ACE_TEXT("Could not find a valid data representation.\n")));
+    }
+    return DDS::RETCODE_ERROR;
+  }
+
```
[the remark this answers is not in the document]
sonndinh: Added.

dds/DCPS/DataReaderImpl_T.h
sonndinh: Added.

dds/DCPS/DataReaderImpl_T.h
sonndinh: I added a debug message for this.

dds/DCPS/DataReaderImpl.cpp
sonndinh: `lookup_instance` is updated now.

dds/DCPS/DataReaderImpl.h
iguessthislldo: Ah, must have I forgot about that.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3282,6 +3287,57 @@ DataReaderImpl::get_ice_endpoint()
...
+    return DDS::RETCODE_ERROR;
+  }
+  ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+             ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+             ACE_TEXT("Setup successfully with data representations: ")));
+  OPENDDS_SET(Encoding::Kind)::iterator it = decoding_modes_.begin();
+  for (; it != decoding_modes_.end(); ++it) {
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("%C "), Encoding::kind_to_string(*it).c_str()));
+  }
+  ACE_DEBUG((LM_DEBUG, ACE_TEXT(".\n")));
```
iguessthislldo, line 3336: This should be conditional on `DCPS_debug_level >= 2`.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -915,6 +915,22 @@ namespace OpenDDS {
...
         return;
       }
+
+      if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
+          ACE_TEXT("%CDataReaderImpl::lookup_instance: ")
+          ACE_TEXT("Encoding kind of the received sample (%C) does not ")
+          ACE_TEXT("match the ones specified by DataReader.\n"),
+          TraitsType::type_name(),
+          Encoding::kind_to_string(encoding.kind()).c_str()));
```
iguessthislldo, line 925: Looking at this, it would probably be better as a warning that depends on `DCPS_debug_level`.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -915,6 +915,22 @@ namespace OpenDDS {
...
+        return;
+      }
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+        ACE_TEXT("%CDataReaderImpl::lookup_instance: ")
+        ACE_TEXT("Deserializing with encoding kind %C.\n"),
+        TraitsType::type_name(),
+        Encoding::kind_to_string(encoding.kind()).c_str()));
```
iguessthislldo, line 932: This should be conditional on `DCPS_debug_level >= 8`.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,54 @@ DataWriterImpl_T
...
+                   TraitsType::type_name()));
       }
       return DDS::RETCODE_ERROR;
     }
-#endif
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) "),
+      ACE_TEXT("%CDataWriterImpl::setup_serialization: "),
+      ACE_TEXT("Setup successfully with %C data representation.\n"),
+      TraitsType::type_name(),
+      Encoding::kind_to_string(encoding_mode_.encoding().kind()).c_str()));
```
iguessthislldo, line 391: This should be conditional on `DCPS_debug_level >= 2`.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -993,6 +1009,22 @@ namespace OpenDDS {
...
         return;
       }
+
+      if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
+        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
+          ACE_TEXT("%CDataReaderImpl::dds_demarshal: ")
+          ACE_TEXT("Encoding kind %C of the received sample does not ")
+          ACE_TEXT("match the ones specified by DataReader.\n"),
+          TraitsType::type_name(),
+          Encoding::kind_to_string(encoding.kind()).c_str()));
```
iguessthislldo, line 1019: Looking at this, it would probably be better as a warning that depends on `DCPS_debug_level`.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -993,6 +1009,22 @@ namespace OpenDDS {
...
+        return;
+      }
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+        ACE_TEXT("%CDataReaderImpl::dds_demarshal: ")
+        ACE_TEXT("Deserializing with encoding kind %C.\n"),
+        TraitsType::type_name(),
+        Encoding::kind_to_string(encoding.kind()).c_str()));
```
iguessthislldo, line 1026: This should be conditional on `DCPS_debug_level >= 8`.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,54 @@ DataWriterImpl_T
...
+            break;
+          } else if (::OpenDDS::DCPS::DCPS_debug_level >= 2) {
+            // Valid but incompatible data representation
+            ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+                       ACE_TEXT("%CDataWriterImpl::setup_serialization: ")
+                       ACE_TEXT("Skip %C data representation\n"),
+                       TraitsType::type_name(),
+                       Encoding::kind_to_string(encoding_kind).c_str()));
+          }
+        } else if (::OpenDDS::DCPS::DCPS_debug_level){
```
iguessthislldo, line 364:
```suggestion
        } else if (::OpenDDS::DCPS::DCPS_debug_level) {
```

mitza-oci:
Test failure on Jenkins `tests/DCPS/DataRepresentation/run_test.pl rtps_disc`

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -915,6 +915,26 @@ namespace OpenDDS {
       if (!encap.to_encoding(encoding, MarshalTraitsType::extensibility())) {
         return;
       }
+
+      if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
+        if (DCPS_debug_level >= 8) {
+          ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) WARNING ")
```
iguessthislldo, line 921:
```suggestion
        if (DCPS_debug_level >= 1) {
          ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING ")
```

Warnings should use `ACE_DEBUG` as a matter of convention and log level one because they are urgent.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -993,6 +1013,26 @@ namespace OpenDDS {
...
```
iguessthislldo, line 1019:
```suggestion
        if (DCPS_debug_level >= 1) {
          ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING ")
```

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -341,82 +341,56 @@ DataWriterImpl_T
...
       }
       return DDS::RETCODE_ERROR;
     }
-#endif
+    if (::OpenDDS::DCPS::DCPS_debug_level >= 2) {
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) "),
+      ACE_TEXT("%CDataWriterImpl::setup_serialization: "),
+      ACE_TEXT("Setup successfully with %C data representation.\n"),
+      TraitsType::type_name(),
+      Encoding::kind_to_string(encoding_mode_.encoding().kind()).c_str()));
```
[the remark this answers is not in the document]
sonndinh: Sure. Somehow I missed fixing the indentation for this.

sonndinh:
> Test failure on Jenkins `tests/DCPS/DataRepresentation/run_test.pl rtps_disc`

This test also got segfault in my local machine. I'm looking at it now.

sonndinh:
> > Test failure on Jenkins `tests/DCPS/DataRepresentation/run_test.pl rtps_disc`
> 
> This test also got segfault in my local machine. I'm looking at it now.

The test failure was due to some test cases specifying XML as the data representation for data writer or data reader, but currently `setup_serialization` and `setup_deserialization` return error when they encounter XML data representation since we're not supporting XML. This causes creating data writer and data reader to fail and thus the test cases report failure. So I think we need to decide whether encountering XML data representation constitutes an error and change the `DataRepresentation` test accordingly.

iguessthislldo:
> So I think we need to decide whether encountering XML data representation constitutes an error and change the DataRepresentation test accordingly.

Yes. It is only an error when it's the only valid data representation specified. If OpenDDS is handling that case correctly, which it sounds like it is, then I think there are two changes that need to be made to the DataRepresentation test:

1. It needs to report that kind of error. It's reporting `-DCPSSecurityDebugLevel 9`, but nothing else, so something seems wrong with how logging is setup.
2. It needs to expect DataReader and DataWriter creation to fail in these cases. I think @ClaytonCalabrese already changed the test to expect that in some cases but it sounds like the expected results have to be tweaked.

sonndinh:
> > So I think we need to decide whether encountering XML data representation constitutes an error and change the DataRepresentation test accordingly.
> 
> Yes. It is only an error when it's the only valid data representation specified. If OpenDDS is handling that case correctly, which it sounds like it is, then I think there are two changes that need to be made to the DataRepresentation test:
> 
> 1. It needs to report that kind of error. It's reporting `-DCPSSecurityDebugLevel 9`, but nothing else, so something seems wrong with how logging is setup.
> 2. It needs to expect DataReader and DataWriter creation to fail in these cases. I think @ClaytonCalabrese already changed the test to expect that in some cases but it sounds like the expected results have to be tweaked.

`setup_serialization` and `setup_deserialization` currently return an error immediately when they encounter a data representation different than XCDR, XCDR2, or UNALIGNED_CDR. I guess I need to change them to return an error only when there is no one of those three data representations found? 

I think some test cases that involve `xml_repr` which only contains XML representation need to be changed.

PR #1792, opened 2020-07-15
Prefer Unicast For INFO_DST Submessages

simpsont-oci:
As I mentioned previously, this seems to be a pretty straightforward change and has performance improvements in the test environment for small & medium scale tests. However, at larger test scales it also seemed slightly less stable. I can't think of any good reasons why that would be the case, unless it's simply a matter of tying up the reactor thread with a larger number of individual writes. I'm hopeful that one of you might spot something obvious that I'm missing.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -557,7 +557,7 @@ RtpsUdpDataLink::add_locators(const RepoId& remote_id,
     narrow_address.addr_to_string(narrow_addr_buff, 256);
     ACE_TCHAR wide_addr_buff[256] = {};
     wide_address.addr_to_string(wide_addr_buff, 256);
-    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locators %C is now at %s or %s\n"), LogGuid(remote_id).c_str(), narrow_addr_buff, wide_addr_buff));
+    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locators %C is now at %s and %s\n"), LogGuid(remote_id).c_str(), narrow_addr_buff, wide_addr_buff));
   }
 }
 
@@ -2202,14 +2202,14 @@ RtpsUdpDataLink::build_meta_submessage_map(MetaSubmessageVec& meta_submessages, 
   AddrSet addrs;
   // Sort meta_submessages by address set and destination
   for (MetaSubmessageVec::iterator it = meta_submessages.begin(); it != meta_submessages.end(); ++it) {
-    const bool undirected = it->dst_guid_ == GUID_UNKNOWN;
-    if (undirected) {
-      addrs = get_addresses_i(it->from_guid_); // This will overwrite, but addrs should always be empty here
-    } else {
+    const bool directed = it->dst_guid_ != GUID_UNKNOWN;
+    if (directed) {
       accumulate_addresses(it->from_guid_, it->dst_guid_, addrs, true);
+    } else {
+      addrs = get_addresses_i(it->from_guid_); // This will overwrite, but addrs should always be empty here
     }
     for (RepoIdSet::iterator it2 = it->to_guids_.begin(); it2 != it->to_guids_.end(); ++it2) {
-      accumulate_addresses(it->from_guid_, *it2, addrs, !undirected);
+      accumulate_addresses(it->from_guid_, *it2, addrs, directed);
     }
     if (addrs.empty()) {
       continue;
@@ -4009,7 +4009,7 @@ RtpsUdpDataLink::get_addresses_i(const RepoId& local) const {
 
 void
 RtpsUdpDataLink::accumulate_addresses(const RepoId& local, const RepoId& remote,
-                                      AddrSet& addresses, bool narrow) const {
+                                      AddrSet& addresses, bool prefer_narrow) const {
   ACE_UNUSED_ARG(local);
   OPENDDS_ASSERT(local != GUID_UNKNOWN);
   OPENDDS_ASSERT(remote != GUID_UNKNOWN);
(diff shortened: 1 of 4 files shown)


simpsont-oci:
The checks for this didn't run. Going to try to amend last commit and force-push to retrigger checks.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2199,13 +2202,14 @@ RtpsUdpDataLink::build_meta_submessage_map(MetaSubmessageVec& meta_submessages,
   AddrSet addrs;
   // Sort meta_submessages by address set and destination
   for (MetaSubmessageVec::iterator it = meta_submessages.begin(); it != meta_submessages.end(); ++it) {
-    if (it->dst_guid_ == GUID_UNKNOWN) {
+    const bool undirected = it->dst_guid_ == GUID_UNKNOWN;
```
mitza-oci, line 2205: Would this be clearer if it was `directed`?

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -244,32 +244,54 @@ RtpsUdpTransport::use_datalink(const RepoId& local_id,
...
+  bool uc_addr_valid = false;
+  bool mc_addr_valid = false;
   for (CORBA::ULong i = 0; i < locators.length(); ++i) {
     ACE_INET_Addr addr;
     // If conversion was successful
     if (locator_to_address(addr, locators[i], false) == 0) {
       // if this is a unicast address, or if we are allowing multicast
-      if (!addr.is_multicast() || config().use_multicast_) {
-        return addr;
+      if (!addr.is_multicast()) {
```
mitza-oci, line 269:
```suggestion
      if (addr.is_multicast()) {
```
and flip the two then/else blocks

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -244,32 +244,54 @@ RtpsUdpTransport::use_datalink(const RepoId& local_id,
...
+          uc_addr = addr;
+          uc_addr_valid = true;
+        }
+      } else {
+        if (mc_addr == ACE_INET_Addr()) {
+          mc_addr = addr;
+          mc_addr_valid = true;
+        }
+      }
+      if (uc_addr_valid && (!config().use_multicast_ || mc_addr_valid)) {
```
mitza-oci, line 280: should this be outside the loop?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -231,10 +231,11 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
...
   struct RemoteInfo {
-    RemoteInfo() : addr_(), requires_inline_qos_(false) {}
-    RemoteInfo(const ACE_INET_Addr& addr, bool iqos)
-      : addr_(addr), requires_inline_qos_(iqos) {}
-    ACE_INET_Addr addr_;
+    RemoteInfo() : narrow_addr_(), wide_addr_(), requires_inline_qos_(false) {}
+    RemoteInfo(const ACE_INET_Addr& narrow_addr, const ACE_INET_Addr& wide_addr, bool iqos)
+      : narrow_addr_(narrow_addr), wide_addr_(wide_addr), requires_inline_qos_(iqos) {}
+    ACE_INET_Addr narrow_addr_;
+    ACE_INET_Addr wide_addr_;
```
jrw972, line 238: `unicast_addr_` and `multicast_addr_`?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -544,17 +544,20 @@ RtpsUdpDataLink::leave_multicast_group(const DCPS::NetworkInterface& nic)
...
 
   if (DCPS::DCPS_debug_level > 3) {
-    ACE_TCHAR addr_buff[256] = {};
-    address.addr_to_string(addr_buff, 256);
-    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locator %C is now at %s\n"), LogGuid(remote_id).c_str(), addr_buff));
+    ACE_TCHAR narrow_addr_buff[256] = {};
+    narrow_address.addr_to_string(narrow_addr_buff, 256);
+    ACE_TCHAR wide_addr_buff[256] = {};
+    wide_address.addr_to_string(wide_addr_buff, 256);
+    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locators %C is now at %s or %s\n"), LogGuid(remote_id).c_str(), narrow_addr_buff, wide_addr_buff));
```
jrw972, line 560:
```suggestion
    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) RtpsUdpDataLink::add_locators %C is now at %s and %s\n"), LogGuid(remote_id).c_str(), narrow_addr_buff, wide_addr_buff));
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: or `prefer_multicast` or `prefer_unicast`?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -728,7 +729,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
   EndpointSecurityAttributesMap endpoint_security_attributes_;
 #endif
 
-  void accumulate_addresses(const RepoId& local, const RepoId& remote, AddrSet& addresses) const;
+  void accumulate_addresses(const RepoId& local, const RepoId& remote, AddrSet& addresses, bool narrow = false) const;
```
jrw972, line 732: Keeping my theme, I would change `narrow` to `prefer_unicast` or `prefer_multicast`.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
simpsont-oci: This is the early return case so we don't have loop over everything if we already have both addresses (or if we have a unicast address and multicast is disabled).

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
simpsont-oci: I think the reason I didn't just say unicast and multicast is that when multicast is disabled they're both the unicast address. Also, if for whatever reason we only have multicast addresses to chose from, the unicast would be multicast. So you could be debugging and see that multicast_addr_ was actually a unicast address and you might think it was a bug or something. But maybe narrow and wide are just as confusing...

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: In the context of this function, I prefer directed. For RemoteInfo & accumulate_addresses I'm open to switching to different names if there's consensus.

PR #1793, opened 2020-07-15
Fix potential leak in Sedp::Msg by checking for ownership

simpsont-oci:
And deleting owned objects in destructor.


simpsont-oci:
Checks didn't run for this one either. Going to --amend --no-edit last commit and force push to retrigger checks.

PR #1794, opened 2020-07-15
XTypes: WChar Support in Metaclass Generator

ClaytonCalabrese:
I did a check by moving the new test files to an old build of OpenDDS and running the test. As expected, it fails because it is not certain the length of a wchar. 

 The new build of OpenDDS with the metastruct changes passes because a wchar is always 16bits.

### DIFF
--- a/tests/DCPS/MetaStruct/MetaStructTest.cpp
+++ b/tests/DCPS/MetaStruct/MetaStructTest.cpp
@@ -2,6 +2,7 @@
 #include <ace/OS_NS_string.h>
 #include <tao/SystemException.h>
 #include <dds/DCPS/FilterEvaluator.h>
+#include <dds/DCPS/Serializer.h>
 
 #include "MetaStructTestTypeSupportImpl.h"
 
@@ -12,8 +13,8 @@
 
 using namespace OpenDDS::DCPS;
 
-int check(const TAO::String_Manager& lhs, const TAO::String_Manager& rhs,
-          const char* name)
+int checkVal(const TAO::String_Manager& lhs, const TAO::String_Manager& rhs,
+             const char* name)
 {
   if (std::strcmp(lhs, rhs)) {
     std::cout << "ERROR: target's " << name << " " << lhs << " != "
@@ -23,7 +24,7 @@ int check(const TAO::String_Manager& lhs, const TAO::String_Manager& rhs,
   return 0;
 }
 
-int check(const float& lhs, const float& rhs, const char* name)
+int checkVal(const float& lhs, const float& rhs, const char* name)
 {
   if (std::fabs((lhs - rhs) / lhs) > .001) {
     std::cout << "ERROR: target's " << name << " " << lhs << " != "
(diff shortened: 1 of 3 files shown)


tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -59,6 +60,30 @@ int check(const T& lhs, const T& rhs, const char* name)
...
 
+template<typename T, typename T2>
+int check(const T& lhs, const T& rhs, const char* name, ACE_Message_Block* amb, Value::Type type,
+          const MetaStruct& ms, T2 Value::*ptrmbr)
+{
+  if (checkVal(lhs, rhs, name)) return 1;
+
+  Message_Block_Ptr mb (amb->duplicate());
+  const Encoding encoding(OpenDDS::DCPS::Encoding::KIND_UNALIGNED_CDR, OpenDDS::DCPS::ENDIAN_NATIVE);
+  OpenDDS::DCPS::Serializer ser(mb.get(), encoding);
```
iguessthislldo, line 71:
```suggestion
  const Encoding encoding(Encoding::KIND_UNALIGNED_CDR);
  Serializer ser(mb.get(), encoding);
```

Native is the default argument and there's a `using namespace OpenDDS::DCPS` at the top.

tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -75,6 +100,8 @@ int run_test(int, ACE_TCHAR*[])
   Source src;
   src.rhs_a.s = "hello";
   src.rhs_a.l = 42;
+  src.rhs_a.w = 0xdead;
```
iguessthislldo, line 103: This will work, but I think it's fun to use non-ASCII when dealing with wide characters and strings. Me and @simpsont-oci have done this in the past. For example maybe U+263A (☺) or something else in [the Unicode Miscellaneous Symbols block](https://en.wikipedia.org/wiki/Miscellaneous_Symbols)?

```suggestion
  src.rhs_a.w = L'☺';
```

tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -96,21 +123,30 @@ int run_test(int, ACE_TCHAR*[])
...
   }
-
-  return check(tgt.lhs_a.s, src.rhs_a.s, "lhs_a.s")
-    + check(tgt.lhs_a.l, src.rhs_a.l, "lhs_a.l")
-    + check(tgt.lhs_sa[0], src.rhs_sa[0], "lhs_sa[0]")
-    + check(tgt.lhs_sa[1], src.rhs_sa[1], "lhs_sa[1]")
-    + check(tgt.lhs_sa[2], src.rhs_sa[2], "lhs_sa[2]")
+  const Encoding encoding(OpenDDS::DCPS::Encoding::KIND_UNALIGNED_CDR, OpenDDS::DCPS::ENDIAN_NATIVE);
+  OpenDDS::DCPS::Message_Block_Ptr data(new ACE_Message_Block(OpenDDS::DCPS::serialized_size(encoding, src)));
+  OpenDDS::DCPS::Serializer ser(data.get(), encoding);
```
iguessthislldo, line 128:
```suggestion
  const Encoding encoding(Encoding::KIND_UNALIGNED_CDR);
  Message_Block_Ptr data(new ACE_Message_Block(serialized_size(encoding, src)));
  Serializer ser(data.get(), encoding);
```

tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -96,21 +123,30 @@ int run_test(int, ACE_TCHAR*[])
...
+  if(!(ser<<src)) {
```
iguessthislldo, line 129:
```suggestion
  if (!(ser<<src)) {
```

tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -96,21 +123,30 @@ int run_test(int, ACE_TCHAR*[])
...
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Tests::MetaStructTest::run_test: ")
+      ACE_TEXT("Failed to serialize source.\n")));
```
iguessthislldo, line 131:
```suggestion
      ACE_TEXT("Failed to serialize source.\n")));
    return 1;
```

iguessthislldo:
QuerryCondition test failed, which can't be ignored given `MetaStruct` is used for queries and filtering.

PR #1800, opened 2020-07-20
Fix Data Fragmentation and Bundled Submessages in Security

iguessthislldo:
Fix issues related to packet size and data fragmentation and bundled submessages when using various protection schemes in security. Currently will just cover worse-case size scenarios and will need to be optimized in the future.

TODO:

- [x] Update LargeSample test processes to use separate security identities.
- [x] Remove debug statements.
- [ ] Check that `max_message_size` and `max_bundle_size` aren't too small?
- [x] Resolve empty fully-encrypted submessage Issue.
- [ ] Fix "secure RTPS processing failed: decode_rtps_message failed" error


dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -713,6 +713,22 @@ RtpsUdpSendStrategy::stop_i()
...
 
+size_t RtpsUdpSendStrategy::max_message_size(TransportQueueElement* element) const
+{
+  size_t size = max_message_size_;
+#ifdef OPENDDS_SECURITY
+  if (element) {
+    // Temp Dummy Fixed Value. This should be enough for the PREFIX (24) and
+    // POSTFIX (24), and the bytes being added to the payload, but it doesn't
+    // seem to be working.
+    size -= 100;
```
simpsont-oci, line 724: I think size can be zero, meaning unlimited. Better check for that here, or you'll get negative / underflow sizes.

simpsont-oci:
Also, more generally, does it make sense to just add a variably sized type to the security attributes test and just increase that in order to prompt fragmentation so you can easily test all the different security flags?

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
iguessthislldo: I know that's valid for tcp, but is it valid for rtps?

simpsont-oci: It defaults to a big number, but it's also configurable so it can be whatever a user wants to make it.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2272,7 +2272,7 @@ struct BundleHelper {
     prev_size_ = size_;
     prev_padding_ = padding_;
     gen_find_size(val, size_, padding_);
-    if ((size_ + padding_) > max_bundle_size_) {
+    if ((size_ + padding_) > (max_bundle_size_ - 100)) {
```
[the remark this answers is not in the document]
iguessthislldo: > Message bundling does not currently handle DATA or DATA_FRAG submessages (they use a different path through the code).

Yes. I put this PR rather quickly and missed this change which I did more to "mark" where I needed to make the real change later. Changes from https://github.com/iguessthislldo/OpenDDS/pull/12 addresses this more properly. I will also update the title.

> If the bundling size needs to get reduced to accommodate security prefixes & suffixes, I'd prefer to see this done on a per-writer basis, based on whether those additional submessages would be added, rather than just a blanket reduction everywhere.

Absolutely agree. I started to work on that somewhat for DATA/FRAG side of things, but unfortunately this is just a quick fix.

iguessthislldo:
> Also, more generally, does it make sense to just add a variably sized type to the security attributes test and just increase that in order to prompt fragmentation so you can easily test all the different security flags?

I'm not sure. It's true it would be easy to test all the combinations, but I don't think we need to test them all, just a subset of combinations of the three protection types. In addition to that I think we should enable security in normal tests as the need comes up to improve test coverage.

tests/DCPS/ManyToMany/DataReaderListener.h
```diff
@@ -64,10 +64,10 @@ class DataReaderListenerImpl
   mutable ACE_Thread_Mutex mutex_;
   const Options& options_;
   DDS::DataReader_var reader_;
-  typedef std::set< ::CORBA::Long> Counts;
-  typedef std::map< ::CORBA::Long, Counts> WriterCounts;
-  typedef std::map< ::CORBA::Long, WriterCounts> ParticipantWriters;
-  typedef std::map<std::string, ParticipantWriters> ProcessParticipants;
+  typedef std::set<::CORBA::Long> Counts;
```
mitza-oci, line 67:
```suggestion
  typedef std::set< ::CORBA::Long> Counts;
```
C++03 doesn't allow the `<` and `::` tokens next to each other without whitespace.
Also I usually just use "int" since it's the same thing.  Or drop the initial `::`.

dds/DCPS/transport/framework/TransportSendStrategy.cpp
```diff
@@ -1065,7 +1065,7 @@ TransportSendStrategy::send(TransportQueueElement* element, bool relink)
         if (max_message_size) { // fragmentation enabled
           const size_t avail = this->space_available();
           if (element_length > avail) {
-            VDBG_LVL((LM_TRACE, "(%P|%t) DBG:   Fragmenting\n"), 0);
+            VDBG_LVL((LM_TRACE, "(%P|%t) DBG:   Fragmenting %u > %u\n", element_length, avail), 0);
```
mitza-oci, line 1068: `%B` is used for `size_t` logging.  It does seem to be needed so that the `va_arg` is 64-bit

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2246,50 +2246,79 @@ bool RtpsUdpDataLink::separate_message(EntityId_t entity)
 namespace {
 
 struct BundleHelper {
+  const size_t initial_size =
```
mitza-oci, line 2249:
```suggestion
  static const size_t initial_size =
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2246,50 +2246,79 @@ bool RtpsUdpDataLink::separate_message(EntityId_t entity)
...
+    meta_submessage_bundle_sizes_.push_back(size_);
+    size_ = initial_size;
   }
 
   template <typename T>
-  void push_to_next_bundle(const T&) {
-    meta_submessage_bundle_sizes_.push_back(prev_size_ + prev_padding_);
-    size_ -= prev_size_; padding_ -= prev_padding_; prev_size_ = 0; prev_padding_ = 0;
-  }
+  bool add_to_bundle(const T& val, size_t& submessage_length) {
```
mitza-oci, line 2279:
```suggestion
  bool add_to_bundle(const T& val, size_t& submessage_length)
  {
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2246,50 +2246,79 @@ bool RtpsUdpDataLink::separate_message(EntityId_t entity)
 namespace {
 
 struct BundleHelper {
+  const size_t initial_size =
+#ifdef OPENDDS_SECURITY
+    RtpsUdpSendStrategy::MaxSecureFullMessageLeadingSize;
+#else
+    0
```
mitza-oci, line 2253:
```suggestion
    0;
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2246,50 +2246,79 @@ bool RtpsUdpDataLink::separate_message(EntityId_t entity)
...
 
-  size_t prev_size_diff() const {
-    return size_ - prev_size_;
-  }
+    const size_t prev_prev_size = size_;
 
-  size_t prev_padding_diff() const {
-    return padding_ - prev_padding_;
+#ifdef OPENDDS_SECURITY
+    // If this a Data submessage or something that is encrypted.
```
mitza-oci, line 2291:
```suggestion
    // Could be an encoded submessage (encoding happens later)
```
Will not be Data

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2246,50 +2246,79 @@ bool RtpsUdpDataLink::separate_message(EntityId_t entity)
...
 
-  template <typename T>
-  bool add_to_bundle(const T& val) {
-    prev_size_ = size_;
-    prev_padding_ = padding_;
-    gen_find_size(val, size_, padding_);
-    if ((size_ + padding_) > max_bundle_size_) {
-      push_to_next_bundle(val);
-      return false;
+    if (first_add_to_bundle_) {
```
mitza-oci, line 2283: What happens differently based on `first_add_to_bundle_`?

tests/DCPS/LargeSample/publisher.cpp
```diff
@@ -57,23 +66,43 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
...
+      append(props, DDS::Security::Properties::AuthIdentityCA,
+        "file:../../security/certs/identity/identity_ca_cert.pem");
+      append(props, DDS::Security::Properties::AuthIdentityCertificate,
+        "file:../../security/certs/identity/test_participant_01_cert.pem");
+      append(props, DDS::Security::Properties::AuthPrivateKey,
+        "file:../../security/certs/identity/test_participant_01_private_key.pem");
+      append(props, DDS::Security::Properties::AccessPermissionsCA,
+        "file:../../security/certs/permissions/permissions_ca_cert.pem");
+      append(props, DDS::Security::Properties::AccessGovernance,
+        "file:./governance_signed.p7s");
```
mitza-oci, line 91:
```suggestion
        "file:governance_signed.p7s");
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
iguessthislldo: It's just for logging that it is the first add in the series of bundles so it's easy to see in the logs. This will be removed along with the extra debug messages.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
iguessthislldo: Correct me if I'm wrong, but it seems to me that a `const` and a `static const` class member would have the same result.

mitza-oci: It's static because it's a property of the class and not a specific object.  Only static data members that are integral/enum types can be initialized in the class definition.

iguessthislldo: That is to say that `const` would imply the information is common to instances of the class, which is same thing as `static`. I'm not opposed to always saying `static const` and I'll change it, but I'm just curious. I guess it could be stored with each object, but that seems like a massive optimization for a compiler to miss.

iguessthislldo: Actually I guess if it's a const member that initalized by the constructor, then I guess it's not common to all instances. Then I think it's a optimization the compiler should still make if that's not the case but I guess it's not as simple as I was thinking.

mitza-oci: As long as what you're proposing compiles on all supported platforms, I'd be OK with it.  But I don't think it will.

The use of `static` here is different than just `const`.  A `const` member can be assigned in the constructor (can depend on constructor arguments).  A `static const` (which is integral/enum and in-class initialized) is completely determined at compile time.

iguessthislldo: I never reloaded the page, so I didn't see you're first response, so I changed my mind just thinking though about it, but the point about the initialization in definition is enough. I added `static` to the code locally and will try to keep this in mind going forward.

iguessthislldo:
I just noticed these errors which where hidden locally: https://prb.objectcomputing.com/jenkins/job/OpenDDS_Pull_Requests/2506/testReport/junit/(root)/Autobuild_Tests/tests_DCPS_LargeSample_run_test_pl_rtps_disc_sec/

PR #1801, opened 2020-07-20
Serializer tests

adamsj-oci:
Draft PR to run Serializer tests.

### DIFF
--- a/tests/DCPS/UnitTests/Serializer.cpp
+++ b/tests/DCPS/UnitTests/Serializer.cpp
@@ -318,6 +318,29 @@ TEST(serializer_test, Serializer__Serializer_ACE_Message_Block_Kind_bool)
   EXPECT_EQ(0, ser.encoding().to_string().compare(0, 20, "XCDR2, little-endian"));
 }
 
+TEST(serializer_test, Serializer__set_endianness)
+{
+  ACE_Message_Block amb;
+  Encoding enc(Encoding::KIND_UNALIGNED_CDR, ENDIAN_LITTLE);
+  Serializer ser(&amb, enc);
+
+  EXPECT_TRUE(ser.good_bit());
+  EXPECT_EQ(0, ser.encoding().to_string().compare(0, 28, "Unaligned CDR, little-endian"));
+
+  ser.endianness(ENDIAN_BIG);
+  EXPECT_EQ(0, ser.encoding().to_string().compare(0, 25, "Unaligned CDR, big-endian"));
+}
+
+TEST(serializer_test, Serializer__swap_bytes_endianness)
+{
+  ACE_Message_Block amb;
+  Encoding enc;
+  Serializer ser(&amb, enc);
+
+  ser.swap_bytes(true);
+  EXPECT_EQ(ser.endianness(), ENDIAN_NONNATIVE);
+}
+
 int main(int argc, char* argv[])
 {
   ::testing::InitGoogleTest(&argc, argv);
--- a/tests/DCPS/UnitTests/Serializer.cpp
+++ b/tests/DCPS/UnitTests/Serializer.cpp
@@ -176,6 +176,121 @@ TEST(serializer_test, EncapsulationHeader__from_encoding_UNALIGNED_CDR_LITTLE_MU
   EXPECT_STREQ("CDR/XCDR1 Big Endian Plain", eh.to_string().c_str());
 }
 
+TEST(serializer_test, EncapsulationHeader__to_encoding_MUTABLE)
+{
+  EncapsulationHeader eh;
+  Encoding enc;
+  EXPECT_FALSE(eh.to_encoding(enc,Extensibility::MUTABLE));
+  EXPECT_STREQ("CDR/XCDR1 Big Endian Plain", eh.to_string().c_str());
+}
+
+TEST(serializer_test, EncapsulationHeader__to_encoding_NOT_MUTABLE)
+{
+  EncapsulationHeader eh;
+  Encoding enc;
+  EXPECT_TRUE(eh.to_encoding(enc,Extensibility::FINAL));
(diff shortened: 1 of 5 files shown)


tests/DCPS/UnitTests/Serializer.cpp
```diff
@@ -0,0 +1,298 @@
+#include <dds/DCPS/Serializer.h>
+
+#include <gtest/gtest.h>
+
+using namespace OpenDDS::DCPS;
+
+TEST(serializer_test, Encoding__Encoding)
+{
+  Encoding enc;
+  EXPECT_EQ(0,enc.to_string().compare(0,24,"CDR/XCDR1, little-endian"));
```
iguessthislldo, line 10:
```suggestion
  EXPECT_EQ(0, enc.to_string().compare(0, 24,"CDR/XCDR1, little-endian"));
```

Maybe use the `EXPECT_STREQ` macro instead like you do latter in the file? I assume that gets better error reporting out of Google Test.

tests/DCPS/UnitTests/Serializer.cpp
```diff
@@ -0,0 +1,298 @@
...
+TEST(serializer_test, Encoding__Encoding_XCDR2_No_Swap)
+{
+  Encoding enc(Encoding::Kind::KIND_XCDR2,false);
+  EXPECT_EQ(0,enc.to_string().compare(0,20,"XCDR2, little-endian"));
+}
+
+TEST(serializer_test, Encoding__Encoding_XCDR1_max_align)
+{
+  Encoding enc(Encoding::Kind::KIND_XCDR1,Endianness::ENDIAN_BIG);
+  EXPECT_EQ(8,enc.max_align());
```
iguessthislldo, line 64:
```suggestion
  EXPECT_EQ(8, enc.max_align());
```

And all the rest of the places.

tests/DCPS/UnitTests/Serializer.cpp
```diff
@@ -0,0 +1,298 @@
...
+  EXPECT_TRUE(eh.to_encoding(enc,Extensibility::MUTABLE));
+  EXPECT_STREQ("XCDR2 Little Endian Parameter List", eh.to_string().c_str());
+}
+
+TEST(serializer_test, EncapsulationHeader__to_encoding_XML)
+{
+  EncapsulationHeader eh;
+  eh.kind(EncapsulationHeader::Kind::KIND_XML);
+  Encoding enc;
+  EXPECT_FALSE(eh.to_encoding(enc,Extensibility::MUTABLE));
```
iguessthislldo, line 290: I think this reports an warning message, so we might have to either suppress the scoreboard from reporting this specific warning or set the debug level to hide it.

tests/DCPS/UnitTests/Serializer.cpp
adamsj-oci: So the reason I did it this way is because the string also contains "native" and "non-native". My feeling was that this could swap depending on the CI tests run but the endian prints wouldn't. If this is an incorrect assumption, I can definitely switch to the `EXPECT_STREQ` checks and check the full string output.

tests/DCPS/UnitTests/Serializer.cpp
adamsj-oci: > I think this reports an warning message, so we might have to either suppress the scoreboard from reporting this specific warning or set the debug level to hide it.

It's an `ERROR` level. I wasn't aware the scoreboard would check the log levels. I'll look into suppressing it.

tests/DCPS/UnitTests/Serializer.cpp
iguessthislldo: Good point. It's unlikely and I'm unaware of any system we test on that is actually big endian, but it is still technically possible.

iguessthislldo: Alternatively maybe change what string to expect based on the system byte over?

tests/DCPS/UnitTests/Serializer.cpp
adamsj-oci: Added `DCPS_debug_level` to suppress output.

tests/DCPS/UnitTests/Serializer.cpp
adamsj-oci: Do we care that it correctly reflects native vs non-native? It's definitely more complete. Maybe good to know that it's correct to know that swapping is working correctly. I vote to leave as is but if you feel strongly, I can change it.

tests/DCPS/UnitTests/Serializer.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+TEST(serializer_test, Encoding__Encoding_XCDR1_ENDIAN_BIG)
+{
+  Encoding enc(Encoding::KIND_XCDR1,Endianness::ENDIAN_BIG);
+  EXPECT_EQ(0,enc.to_string().compare(0,21,"CDR/XCDR1, big-endian"));
+}
+
+TEST(serializer_test, Encoding__Encoding_UNALIGNED_CDR_ENDIAN_BIG)
+{
+  Encoding enc(Encoding:::KIND_UNALIGNED_CDR,Endianness::ENDIAN_BIG);
```
iguessthislldo, line 21:
```suggestion
  Encoding enc(Encoding::KIND_UNALIGNED_CDR, ENDIAN_BIG);
```

adamsj-oci: fixed

tests/DCPS/UnitTests/Serializer.cpp
iguessthislldo: I don't feel strongly about it either way, so it's fine as is.

PR #1817, opened 2020-07-31
XTypes: Serialization position

sonndinh:


dds/DCPS/Serializer.cpp
```diff
@@ -474,7 +474,7 @@ Serializer::read_string(ACE_CDR::WChar*& dest,
   //
   // Extract the string size.
   //
-  ACE_CDR::ULong bytecount; // includes the null
+  ACE_CDR::ULong bytecount; // Not include the null
```
iguessthislldo, line 477:
```suggestion
  ACE_CDR::ULong bytecount;
```

I would just remove it entirely, since if there's no null it you don't have to say that it's not included in the size.

sonndinh:
> This looks like what we need. In `tests/DCPS/Serializer`, I would like to see a basic check for the position and testing for what happens when we try to read or write past the capacity of the `ACE_Message_Block`.

Do we want to do the test in this PR?

dds/DCPS/Serializer.cpp
sonndinh: Okay. I'll remove it. Just wanted to fix the previous comment that may cause confusion.

iguessthislldo:
> Do we want to do the test in this PR?

The simple position check should probably be done in this PR. I would like the buffer overrun test to be in this PR, but it doesn't absolutely need to be done here.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -418,10 +554,15 @@ checkArrayValues(const ArrayValues& expected, const ArrayValues& observed)
...
+  const Encoding& encoding, const bool checkPos)
 {
   ACE_Message_Block* testchain = getchain(sizeof(chaindefs)/sizeof(chaindefs[0]), chaindefs);
-  const char* out = encoding.endianness() ? "" : "OUT";
+  const char* out =
+#ifdef ACE_LITTLE_ENDIAN
+    encoding.endianness() == OpenDDS::DCPS::ENDIAN_BIG ? "" : "OUT";
+#else
+    encoding.endianness() == OpenDDS::DCPS::ENDIAN_LITTLE ? "" : "OUT";
+#endif
```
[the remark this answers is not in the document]
sonndinh: That's right.

PR #1824, opened 2020-08-04
XTypes: XCDR2 must_understand flag

lij-oci:
Implement must_understand flag in XCDR2 on the reader side.

https://trello.com/c/EtGCQW8X/86-xcdr-parameter-list-flags

### DIFF
--- a/dds/DCPS/Serializer.cpp
+++ b/dds/DCPS/Serializer.cpp
@@ -546,7 +546,7 @@ bool Serializer::read_parameter_id(unsigned& id, size_t& size, bool& must_unders
     }
 
     // TODO(iguessthislldo): handle PID flags
-    must_understand = false;
+    must_understand = false; // Placeholder
 
     // If extended, get the "long" id and size
     if (short_id == pid_extended) {
@@ -572,7 +572,6 @@ bool Serializer::read_parameter_id(unsigned& id, size_t& size, bool& must_unders
       return false;
     }
 
-    // TODO(iguessthislldo): Handle Must Understand Flag
     must_understand = emheader & emheader_must_understand;
 
     // Get Size
--- a/dds/DCPS/Serializer.cpp
+++ b/dds/DCPS/Serializer.cpp
@@ -527,7 +527,7 @@ Serializer::read_string(ACE_CDR::WChar*& dest,
   return length;
 }
 
-bool Serializer::read_parameter_id(unsigned& id, size_t& size)
+bool Serializer::read_parameter_id(unsigned& id, size_t& size, bool& must_understand)
 {
   const Encoding::XcdrVersion xcdr = encoding().xcdr_version();
   if (xcdr == Encoding::XCDR_VERSION_1) {
@@ -546,6 +546,7 @@ bool Serializer::read_parameter_id(unsigned& id, size_t& size)
     }
 
     // TODO(iguessthislldo): handle PID flags
+    must_understand = false;
 
     // If extended, get the "long" id and size
     if (short_id == pid_extended) {
@@ -572,6 +573,7 @@ bool Serializer::read_parameter_id(unsigned& id, size_t& size)
     }
 
     // TODO(iguessthislldo): Handle Must Understand Flag
+    must_understand = emheader & emheader_must_understand;
 
     // Get Size
     // TODO(iguessthislldo) LC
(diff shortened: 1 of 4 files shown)


tests/DCPS/Compiler/xcdr/xcdr.cpp
```diff
@@ -476,6 +476,54 @@ TEST(mutable_tests, to_additional_field_xcdr2_test)
...
+  }
+
+  // Deserialize and Compare C++ Values
+  {
+    ACE_Message_Block buffer(1024);
+    buffer.copy((const char*)expected, sizeof(expected));
+    Serializer serializer(&buffer, xcdr2);
+
+    AdditionalFieldMutableStruct result;
+    EXPECT_TRUE(serializer >> result);
```
iguessthislldo, line 519: Looking at this I'm thinking this should probably be:

```suggestion
    ASSERT_TRUE(serializer >> result);
```

In Google test, this stops the test, where `EXPECT_*` keeps going. I'm thinking this because the value comparison below shouldn't happen if deserialization failed, but I know this is copied from my general test code, so it's the same problem through out the file.

tests/DCPS/Compiler/xcdr/xcdr.cpp
```diff
@@ -476,6 +476,54 @@ TEST(mutable_tests, to_additional_field_xcdr2_test)
   /// value?) if we decide on that.
 }
 
+TEST(mutable_tests, to_additional_field_mustunderstatnd_test)
+{
+  const unsigned char expected[] = {
```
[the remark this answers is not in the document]
lij-oci: replaced by additional_mustunderstatnd

tests/DCPS/Compiler/xcdr/xcdr.cpp
lij-oci: Based on the behaviors of EXPECT_TRUE and ASSERT_TRUE, your suggested change makes perfect sense. Similar changes were also made in two other places (Lines 142 & 149).

tests/DCPS/Compiler/xcdr/xcdr.cpp
```diff
@@ -476,6 +476,54 @@ TEST(mutable_tests, to_additional_field_xcdr2_test)
   /// value?) if we decide on that.
 }
 
+TEST(mutable_tests, from_additional_field_misunderstand_test)
+{
+  const unsigned char additional_mustunderstatnd[] = {
```
iguessthislldo, line 481:
```suggestion
  const unsigned char additional_field_must_understand[] = {
```

tests/DCPS/Compiler/xcdr/xcdr.cpp
```diff
@@ -476,6 +476,54 @@ TEST(mutable_tests, to_additional_field_xcdr2_test)
   /// value?) if we decide on that.
 }
 
+TEST(mutable_tests, from_additional_field_misunderstand_test)
```
iguessthislldo, line 479:
```suggestion
TEST(mutable_tests, from_additional_field_must_understand_test)
```

This one's my fault, I let chrome fix the spelling in the suggestion to "misunderstand".

dds/DCPS/Serializer.cpp
```diff
@@ -546,6 +546,7 @@ bool Serializer::read_parameter_id(unsigned& id, size_t& size)
     }
 
     // TODO(iguessthislldo): handle PID flags
+    must_understand = false; // Placeholder
```
sonndinh, line 549: We can set `must_understand` to the second bit of `pid` in case it's not PID_EXTENDED.

tests/DCPS/Compiler/xcdr/xcdr.cpp
lij-oci: I thought you meant that it is correct that this test misunderstands the must_understand field since it EXPECT_FALSE. :) my bad.

dds/DCPS/Serializer.cpp
iguessthislldo: Yes, but we're deferring further XCDR1 support for this iteration.

sonndinh: Gotcha. Also, I haven't seen changes for `write_parameter_id`. Is this function supposed to get an input parameter for must_understand from the caller? Or does it need to get that flag itself?

lij-oci: Writer side support is also deferred. https://trello.com/c/EtGCQW8X/86-xcdr-parameter-list-flags

sonndinh: I see.

PR #1831, opened 2020-08-07
Improve Efficiency of Waiting on Pending Writes

iguessthislldo:
If multiple `DataWriter`s are being deleted in a `DomainParticipant` or `Publisher`, use PendingTimeout for a single deadline to finish pending data and control writes for all the writers, instead of waiting for at least a separate data and control deadline for every single writer. Also start instance unregisters for all writers before setting and waiting on the deadline.

These changes should dramatically reduce the amount of time `Publisher::delete_contained_entities` and `DomainParticipant::delete_contained_entities` block for multiple writers, especially when there are a large amount of them.

### DIFF
--- a/dds/DCPS/DataWriterImpl.cpp
+++ b/dds/DCPS/DataWriterImpl.cpp
@@ -2528,6 +2528,19 @@ DataWriterImpl::prepare_to_delete()
   this->set_deleted(true);
   this->stop_associating();
   this->terminate_send_if_suspended();
+
+#ifndef OPENDDS_NO_PERSISTENCE_PROFILE
+  // Trigger data to be persisted, i.e. made durable, if so
+  // configured. This needs be called before unregister_instances
+  // because unregister_instances may cause instance dispose.
+  if (!persist_data() && DCPS_debug_level >= 2) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: DataWriterImpl::prepare_to_delete: ")
+      ACE_TEXT("failed to make data durable.\n")));
+  }
+#endif
+
+  // Unregister all registered instances prior to deletion.
+  unregister_instances(SystemTimePoint::now().to_dds_time());
 }
 
 PublicationInstance_rch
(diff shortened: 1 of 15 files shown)


dds/DCPS/PublisherImpl.cpp
```diff
@@ -328,11 +314,53 @@ PublisherImpl::lookup_datawriter(const char* topic_name)
...
-  set_deleted(true);
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, pi_lock_, false);
+  bool result = true;
+  const DataWriterMap::iterator end = datawriter_map_.end();
+  for (DataWriterMap::iterator i = datawriter_map_.begin(); i != end; ++i) {
+    DataWriterImpl* const writer = dynamic_cast<DataWriterImpl*>(i->second.in());
+    if (!writer) {
+      result = false;
+    }
+    writer->prepare_to_delete();
```
mitza-oci, line 327:
```suggestion
    if (writer) {
      writer->prepare_to_delete();
    } else {
      result = false;
    }
```

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -2679,20 +2692,12 @@ DataWriterImpl::persist_data()
...
-
-void
-DataWriterImpl::wait_pending()
+void DataWriterImpl::wait_pending()
 {
   if (!TransportRegistry::instance()->released()) {
-    data_container_->wait_pending();
+    data_container_->wait_pending(wait_pending_deadline_);
+    controlTracker.wait_messages_pending(
+      "DataWriterImpl::wait_control_pending", wait_pending_deadline_);
```
mitza-oci, line 2700:
```suggestion
    controlTracker.wait_messages_pending("DataWriterImpl::wait_pending", wait_pending_deadline_);
```

dds/DCPS/PublisherImpl.cpp
iguessthislldo: I'm not sure why I did this one differently from the same situation from the method below...

PR #1834, opened 2020-08-11
XTypes: DHeader

ClaytonCalabrese:
A little branch off from tryconstruct, which applies the DHeader to Union, Array, and Sequence types (This is needed for XCDR2 in addition to tryconstruct).  It also contains a small refactor of the struct code to reduce generated code.


dds/idl/marshal_generator.cpp
```diff
@@ -1830,14 +1967,21 @@ bool marshal_generator::gen_struct(AST_Structure* node,
...
-        "    }\n"
-        "  }\n";
+        indent << "size_t total_size = 0;\n" <<
+        indent << "serialized_size(encoding, total_size, stru);\n" <<
+        indent << "if (!strm.write_delimiter(total_size)) {\n" <<
+        indent << "  return false;\n" <<
+        indent << "}\n";
+        if (not_only_delimited) {
+          be_global->impl_ << "  }\n";
+        }
```
iguessthislldo, line 1984:
```suggestion
      if (not_only_delimited) {
        be_global->impl_ << "  }\n";
      }
```

dds/idl/marshal_generator.cpp
```diff
@@ -389,7 +389,40 @@ namespace {
...
+  // may_be_delimited determines if the dheader is ever needed for this type.
+  // not_only_delimited determines if the dheader code is conditional based on the encoding.
+  // code contains the lines of code to generate
+  // is_ser_func specifies whether this is a serialization operation
+  void generate_dheader_code(bool may_be_delimited, bool not_only_delimited,
+                             const std::vector<std::string>& code, bool is_ser_func = true)
+  {
+    if (may_be_delimited) {
+      string indent = "  ";
+      if(is_ser_func) {
```
mitza-oci, line 403:
```suggestion
      if (is_ser_func) {
```

dds/idl/marshal_generator.cpp
```diff
@@ -389,7 +389,40 @@ namespace {
...
+        if(is_ser_func) {
+          be_global->impl_ <<
+            "  if (strm.encoding().xcdr_version() == Encoding::XCDR_VERSION_2) {\n";
+        } else {
+          be_global->impl_ <<
+            "  if (encoding.xcdr_version() == Encoding::XCDR_VERSION_2) {\n";
+        }
+        indent = "    ";
+      }
+      for (size_t i = 0; i < code.size(); i++) {
```
mitza-oci, line 416:
```suggestion
      for (size_t i = 0; i < code.size(); ++i) {
```

dds/idl/marshal_generator.cpp
```diff
@@ -389,7 +389,40 @@ namespace {
...
+  // Helper function to conditionally generate dheader-related code which is passed as the
+  // "code" argument.
+  // may_be_delimited determines if the dheader is ever needed for this type.
+  // not_only_delimited determines if the dheader code is conditional based on the encoding.
+  // code contains the lines of code to generate
+  // is_ser_func specifies whether this is a serialization operation
+  void generate_dheader_code(bool may_be_delimited, bool not_only_delimited,
+                             const std::vector<std::string>& code, bool is_ser_func = true)
+  {
+    if (may_be_delimited) {
```
mitza-oci, line 401: Would it be more readable for the calling functions to have this conditional check and not call this function if it's not needed?

ClaytonCalabrese: That would lead to more repeated code.  I think it is fine the way it is as the dheader would only apply if it could be delimited.

dds/idl/marshal_generator.cpp
```diff
@@ -405,6 +438,19 @@ namespace {
 
     AST_Type* elem = resolveActualType(seq->base_type());
     Classification elem_cls = classify(elem);
+
+    AST_Decl* node = td;
+    if (td == 0) {
+      node = seq;
+    }
```
iguessthislldo, line 445:
```suggestion
    AST_Decl* node = td ? td : seq;
```

dds/idl/marshal_generator.cpp
```diff
@@ -405,6 +438,19 @@ namespace {
...
+    const ExtensibilityKind exten = be_global->extensibility(node);
+    const OpenDDS::DataRepresentation repr =
+      be_global->data_representations(node);
```
iguessthislldo, line 448: I'm not sure we should be checking `@data_representation` on non-topic types, since officially it's just for constraining QoS.

ClaytonCalabrese: I believe the most recent commit (4324ed7) should fix this for code that I added. There is at least 1 problem in gen_union that already existed and many in gen_struct that will need to be resolved.  I'm not sure that should be part of this PR considering how long it has been in review.

dds/idl/marshal_generator.cpp
```diff
@@ -2237,29 +2292,18 @@ bool marshal_generator::gen_struct(AST_Structure* node,
...
-
   {
     Function insertion("operator<<", "bool");
     insertion.addArg("strm", "Serializer&");
     insertion.addArg("stru", "const " + cxx + "&");
     insertion.endArgs();
 
-    if (may_be_delimited || may_be_parameter_list) {
-      be_global->impl_ <<
-        "  const Encoding& encoding = strm.encoding();\n";
```
iguessthislldo, line 2249: `encoding` is also used for getting field sizes for mutable, not just for the struct's delimiter.

dds/idl/marshal_generator.cpp
```diff
@@ -2202,10 +2258,9 @@ bool marshal_generator::gen_struct(AST_Structure* node,
         "  size_t xcdr1_pl_running_total = 0;\n";
     }
 
-    if (may_be_delimited) {
```
iguessthislldo, line 2205: This is checking if a delimiter is needed.

dds/idl/marshal_generator.cpp
```diff
@@ -2319,24 +2363,13 @@ bool marshal_generator::gen_struct(AST_Structure* node,
     extraction.addArg("stru", cxx + "&");
     extraction.endArgs();
     string intro;
-    if (may_be_delimited) {
-      be_global->impl_ <<
-        "  size_t total_size = 0;\n";
-      const char* indent = "  ";
-      if (not_only_delimited) {
```
iguessthislldo, line 2326: This is an example of an optimization that I'd like to keep.

PR #1836, opened 2020-08-12
Add support for multiple participants per domain.

jonesc-oci:


dds/DCPS/DomainParticipantFactoryImpl.cpp
```diff
@@ -6,7 +6,7 @@
  */
 
 #include "DCPS/DdsDcps_pch.h" //Only the _pch include should start with DCPS/
-
+#include "dds/DCPS/transport/framework/TransportRegistry.h"
```
iguessthislldo, line 9:
```suggestion
#include "transport/framework/TransportRegistry.h"
```

Also please more this below, as it says in the guidelines.

dds/DCPS/Service_Participant.cpp
```diff
@@ -1923,6 +1925,9 @@ Service_Participant::load_domain_configuration(ACE_Configuration_Heap& cf,
         } else {
           reg->domain_default_config(domainId, tc);
         }
+      } else {
+        // store domain and associated config for template or multiple participant domain use
+        TheTransportRegistry->associate_domain_to_config(domainId, global_transport_config_.c_str());
```
simpsont-oci, line 1930: WChar builds are failing here since associate_domain_to_config is expecting std::string and global_transport_config_ is an ACE_TString (wrap with ACE_TEXT_ALWAYS_CHAR).

dds/DCPS/Service_Participant.cpp
```diff
@@ -2188,6 +2193,19 @@ int Service_Participant::configure_domain_range_instance(DDS::DomainId_t domainI
   return 0;
 }
 
+
+bool
+Service_Participant::belongs_to_domain_range(DDS::DomainId_t domainId)
```
simpsont-oci, line 2198: This can be a const function (using const_iterator, etc)

dds/DCPS/Service_Participant.h
```diff
@@ -396,6 +396,13 @@ class OpenDDS_Dcps_Export Service_Participant {
   int load_configuration(ACE_Configuration_Heap& cf,
                          const ACE_TCHAR* filename);
 
+  /**
+   * Used by TransportRegistry to determine if a domain ID
+   * is part of a [DomainRange]
+   */
+  bool belongs_to_domain_range(DDS::DomainId_t domainId);
```
simpsont-oci, line 403: I'd make this a const function.

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -150,6 +149,13 @@ TransportRegistry::load_transport_configuration(const OPENDDS_STRING& file_name,
                                 transport_id.c_str()),
                                -1);
             }
+
+            // store the transport info
+            TransportEntry entry;
+            entry.transport_name = transport_id;
+            entry.transport_info = values;
+            transports_.push_back(entry);
```
simpsont-oci, line 157: Would any of the failure cases below require us to purge / skip this transport, or are they all severe enough it won't matter that we're modifying transports_ here?

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -621,11 +636,155 @@ bool TransportRegistry::has_type(const TransportType_rch& type) const
   return type_map_.count(name);
 }
 
-bool TransportRegistry::has_transport_template() const
+bool TransportRegistry::create_new_transport_instance_for_participant(DDS::DomainId_t id, OPENDDS_STRING& config_name)
 {
-  return transport_templates_.size() > 0;
+  TransportConfig_rch cfg = get_config(domain_config_map_[id]);
+
+  OPENDDS_STRING inst_name = cfg->instances_[0]->name() + "_" + config_name;
```
simpsont-oci, line 643: Are we sure domain_config_map_ will always have an entry for "id" and we're sure that it will always have at least one instance in it? Otherwise maybe have some sanity checks here to return false if those conditions aren't true.

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -621,11 +636,155 @@ bool TransportRegistry::has_type(const TransportType_rch& type) const
...
+
+    for (OPENDDS_MAP(OPENDDS_STRING, OPENDDS_STRING)::const_iterator it = tr_inst.transport_info.begin();
+         it != tr_inst.transport_info.end(); ++it) {
+      // customization.
+      OPENDDS_MAP(OPENDDS_STRING, OPENDDS_STRING)::const_iterator idx = tr_inst.customizations.find(it->first);
+      if (idx != tr_inst.customizations.end()) {
+        OPENDDS_STRING addr = it->second;
+
+        // only add_domain_id_to_ip_addr and add_domain_id_to_port are supported at this time.
+        if (idx->second.find("add_domain_id_to_ip_addr") != std::string::npos) {
```
simpsont-oci, line 668: I know it's a pain, but I would prefer to see us parsing out the individual strings (looking for ",") so we can complain when we find something we don't expect. For example, if they misspell something, we can complain rather than just ignore it since we didn't find the exact string we were looking for.

As it stands, it looks like this will currently accept "thisiswrong_add_domain_id_to_port,totally_bogus" because it will find "add_domain_id_to_port".

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -621,11 +636,155 @@ bool TransportRegistry::has_type(const TransportType_rch& type) const
...
+    inst->load(ach, sect_key);
+    config->instances_.push_back(inst);
+
+    return true;
+  }
+
+  return false;
+}
+
+void TransportRegistry::associate_domain_to_config(DDS::DomainId_t id, OPENDDS_STRING cfg)
```
simpsont-oci, line 782:
```suggestion
void TransportRegistry::associate_domain_to_config(DDS::DomainId_t id, const OPENDDS_STRING& cfg)
```

dds/DCPS/transport/framework/TransportRegistry.h
```diff
@@ -157,15 +167,30 @@ class OpenDDS_Dcps_Export TransportRegistry {
...
+    ValueMap transport_info;
+  };
+
+  OPENDDS_VECTOR(TransportEntry) transports_;
+
+  bool get_transport_info(const ACE_TString& config_name, TransportEntry& inst);
+
+  bool has_transports() const;
+
+  OPENDDS_MAP(DDS::DomainId_t, OPENDDS_STRING) domain_config_map_;
```
simpsont-oci, line 193: Coming back to this now, since these are config names, maybe this structure ought to hold ACE_TStrings instead of OPENDDS_STRINGs?

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -848,6 +1008,43 @@ TransportRegistry::get_transport_template_info(const ACE_TString& config_name, T
   return ret;
 }
 
+bool TransportRegistry::has_transport_templates() const
+{
+  return transport_templates_.size() > 0;
```
simpsont-oci, line 1013:
```suggestion
  return !transport_templates_.empty();
```

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -848,6 +1008,43 @@ TransportRegistry::get_transport_template_info(const ACE_TString& config_name, T
...
+               ACE_TEXT("%C config %s\n"),
+               ret ? "found" : "did not find", config_name.c_str()));
+  }
+
+  return ret;
+}
+
+bool TransportRegistry::has_transports() const
+{
+  return transports_.size() > 0;
```
simpsont-oci, line 1045:
```suggestion
  return !transports_.empty();
```

dds/DCPS/transport/framework/TransportRegistry.h
```diff
@@ -125,6 +126,15 @@ class OpenDDS_Dcps_Export TransportRegistry {
...
   OPENDDS_STRING get_config_instance_name(DDS::DomainId_t id);
 
+  bool create_new_transport_instance_for_participant(DDS::DomainId_t id, OPENDDS_STRING& config_name);
+
+  // The TransportRegistry needs to know the domains for a given
+  // config in order to know when to create new transport instances
+  // for separate participants in the same domain. This is called
+  // from the Service_Particpant since it knows if the GlobalTransportConfig
+  // has been set or overriden by a command line option.
+  void associate_domain_to_config(DDS::DomainId_t id, OPENDDS_STRING cfg);
```
simpsont-oci, line 136:
```suggestion
  void associate_domain_to_config(DDS::DomainId_t id, const OPENDDS_STRING& cfg);
```
And maybe switch to TString?

dds/DCPS/transport/framework/TransportRegistry.h
```diff
@@ -157,15 +167,30 @@ class OpenDDS_Dcps_Export TransportRegistry {
...
 
   bool get_transport_template_info(const ACE_TString& config_name, TransportTemplate& inst);
 
-  bool has_transport_template() const;
+  bool has_transport_templates() const;
+
+  struct TransportEntry
+  {
+    OPENDDS_STRING transport_name;
+    OPENDDS_STRING config_name;
```
simpsont-oci, line 183: TString since we're comparing these with what comes out of the configuration heap?

dds/DCPS/transport/framework/TransportRegistry.cpp
jonesc-oci: Yes, I agree. I'll make this change.

dds/DCPS/Service_Participant.cpp
```diff
@@ -2204,6 +2201,12 @@ Service_Participant::belongs_to_domain_range(DDS::DomainId_t domainId) const
   return false;
 }
 
+ACE_TString&
+Service_Participant::get_global_transport_name()
```
jwillemsen, line 2205: const?

jonesc-oci: Added. Thanks.

PR #1840, opened 2020-08-13
XTypes: Type Lookup Service implementation

kuznetsovmoci:
Added TypeLookupService class, instantiation, and use by EndpointManager / Sedp, as well as TypeAssignability. TypeObjects are cached by the TypeIdentifier key, and RPC is used during matching process to request missing TypeObjects from the remote participants. PmfSporadicTask is used for lost request data cleanup. Used 'TLS_TODO' comment to tag 'risky' areas. Modified UnitTests_XtypesAssignability test to use new TypeAssignability API.


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3661,20 +3678,46 @@ Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& ty
...
-                                           DCPS::Serializer& ser,
-                                           XTypes::TypeLookup_Reply& type_lookup_reply)
+                                           DCPS::Serializer& ser)
 {
-  // TODO: uncomment; anything else to do here?
-  // ser >> type_lookup_reply;
+  XTypes::TypeLookup_Reply type_lookup_reply;
+  // TLS_TODO: verify reply processing
+  if (!(ser >> type_lookup_reply)) {
+    return DDS::RETCODE_ERROR;
```
iguessthislldo, line 3711: Same question about logging and encapsulation.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3985,19 +4028,31 @@ Sedp::Reader::data_received(const DCPS::ReceivedDataSample& sample)
...
       }
     } else if (entity_id == ENTITYID_TL_SVC_REPLY_WRITER) {
-      // TODO: process reply
-      XTypes::TypeLookup_Reply type_lookup_reply;
-      if (!sedp_.type_lookup_reply_reader_->take_tl_reply(sample, ser, type_lookup_reply)) {
+      // TLS_TODO: verify reply processing
+      if (!sedp_.type_lookup_reply_reader_->take_tl_reply(sample, ser)) {
         ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Sedp::Reader::data_received - ")
-          ACE_TEXT("failed to take type lookup reply\n")));
+                  ACE_TEXT("failed to take type lookup reply\n")));
```
iguessthislldo, line 4055:
```suggestion
                   ACE_TEXT("failed to take type lookup reply\n")));
```

One more space if you're going to match the indent with the start of the arguments.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,94 @@
...
+    return it_object->second;
+  }
+  return to_empty;
+}
+
+
+void TypeLookupService::AddTypeObjectsToCache(TypeIdentifierTypeObjectPairSeq& types)
+{
+  for (ACE_UINT32 i = 0; i < types.length(); ++i)
+  {
```
iguessthislldo, line 67:
```suggestion
  for (ACE_UINT32 i = 0; i < types.length(); ++i) {
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
```
iguessthislldo, line 1: By convention we wrap pragma-onces in a ACE check.

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+#ifndef OPENDDS_DCPS_TYPE_LOOKUP_SERVICE_H
+#define OPENDDS_DCPS_TYPE_LOOKUP_SERVICE_H
+
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
```
iguessthislldo, line 8: License notice should go at the top of the file.

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+namespace XTypes {
+
+class OpenDDS_Dcps_Export TypeLookupService : public virtual DCPS::RcObject {
+public:
+  TypeLookupService();
+  ~TypeLookupService();
+
+  // For TypeAssignability
+  const TypeObject& GetTypeObject(const TypeIdentifier& type_id) const;
+  void AddTypeObjectsToCache(const TypeIdentifier& ti, const TypeObject& tobj);
```
iguessthislldo, line 27: Does the spec specify that the functions have to be capitalized? If not could you make them either snake or lowercase camel case.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1200,253 @@ namespace OpenDDS {
         }
       }
 
+      void
+      remove_expired_endpoints(const DCPS::MonotonicTimePoint& /*now*/)
+      {
+        MatchingDataIter it;
+        for (it = matching_data_buffer.begin(); it != matching_data_buffer.end(); it++) {
+          if (MonotonicTimePoint::now() - it->second.time_added_to_map >= max_reply_period) {
```
iguessthislldo, line 1208: Use the "now" from the argument?

dds/DCPS/DiscoveryBase.h
```diff
@@ -892,6 +913,34 @@ namespace OpenDDS {
         return -1;
       }
 
+
+
```
iguessthislldo, line 917:
```suggestion
```

dds/DCPS/RTPS/Sedp.h
```diff
@@ -165,6 +165,10 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
   void signal_liveliness(DDS::LivelinessQosPolicyKind kind);
   void signal_liveliness_unsecure(DDS::LivelinessQosPolicyKind kind);
 
+  // TypeLookup
```
iguessthislldo, line 168:
```suggestion
```

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -119,7 +85,20 @@ class OpenDDS_Dcps_Export TypeAssignability {
...
 private:
-  TypeLookup typelookup_;
+  TypeLookupService* tl_service;
+
+  const MinimalTypeObject& lookup_minimal(const TypeIdentifier& ti) const
+  {
+    return tl_service->GetTypeObject(ti).minimal;
+  }
+
+public:
```
iguessthislldo, line 95: I don't think you need to make a new public section.

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -119,7 +85,20 @@ class OpenDDS_Dcps_Export TypeAssignability {
   bool get_union_member(const MinimalTypeObject*& ret, const CommonStructMember& m) const;
 
 private:
-  TypeLookup typelookup_;
+  TypeLookupService* tl_service;
```
iguessthislldo, line 88:
```suggestion
  TypeLookupService* tl_service_;
```

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -119,7 +85,20 @@ class OpenDDS_Dcps_Export TypeAssignability {
   bool get_union_member(const MinimalTypeObject*& ret, const CommonStructMember& m) const;
 
 private:
```
iguessthislldo, line 87:
```suggestion
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                                  TypeIdentifierPairSeq& complete_to_minimal,
+                                                  TypeIdentifierSeq& not_found)
+{
+  for (CORBA::ULong i = 0; i < type_ids.length(); ++i) {
+    TypeObjectMap::iterator it_object = type_object_map.find(type_ids[i]);
+    if (it_object != type_object_map.end()) {
+      TypeIdentifierTypeObjectPair new_to_pair(it_object->first, it_object->second);
+      types.append(new_to_pair);
+    }
+    else {
```
iguessthislldo, line 38:
```suggestion
    } else {
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+  void AddTypeObjectsToCache(TypeIdentifierTypeObjectPairSeq& types);
+
+  bool TypeObjectInCache(const TypeIdentifier& ti);
+
+  OpenDDS::DCPS::SequenceNumber rpc_sequence_number_;
+
+protected:
+  // Only minimal Type Objects for now
+  typedef std::map<TypeIdentifier, TypeObject> TypeObjectMap;
+  TypeObjectMap type_object_map;
```
iguessthislldo, line 42:
```suggestion
  TypeObjectMap type_object_map_;
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+
+  typedef std::map<TypeIdentifier, TypeIdentifierWithSizeSeq> TypeIdentifierWithSizeSeqMap;
+  TypeIdentifierWithSizeSeqMap type_dependencies_map;
```
iguessthislldo, line 45:
```suggestion
  TypeIdentifierWithSizeSeqMap type_dependencies_map_;
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+
+  void CollectTypesInfoFromCache(const TypeIdentifierSeq& type_ids,
+    TypeIdentifierTypeObjectPairSeq& types,
+    TypeIdentifierPairSeq& complete_to_minimal,
+    TypeIdentifierSeq& not_found);
+
+  TypeObject to_empty;
```
iguessthislldo, line 52:
```suggestion
  TypeObject to_empty_;
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,62 @@
...
+};
+
+typedef DCPS::RcHandle<TypeLookupService> TypeLookupService_rch;
```
iguessthislldo, line 55:
```suggestion
typedef DCPS::RcHandle<TypeLookupService> TypeLookupService_rch;

```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1200,253 @@ namespace OpenDDS {
         }
       }
 
+      void
+      remove_expired_endpoints(const DCPS::MonotonicTimePoint& /*now*/)
+      {
+        MatchingDataIter it;
+        for (it = matching_data_buffer.begin(); it != matching_data_buffer.end(); it++) {
+          if (MonotonicTimePoint::now() - it->second.time_added_to_map >= max_reply_period) {
+            // TLS_TODO: sanity check on locks
```
iguessthislldo, line 1209: A lock on the `matching_data_buffer` would probably be a good idea.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3661,20 +3678,46 @@ Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& ty
...
+                                               XTypes::TypeLookup_Reply& type_lookup_reply)
 {
-  // TODO: uncomment; anything else to do here?
-  // ser >> type_lookup_request;
-  return DDS::RETCODE_OK;
+  // TLS_TODO: verify request processing
+  XTypes::TypeLookup_Request type_lookup_request;
+  if (!(ser >> type_lookup_request)) {
+        return DDS::RETCODE_ERROR;
+    }
```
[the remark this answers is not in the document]
kuznetsovmoci: It looks like encapsulation deserialization is handled by the caller, data_received()

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -119,7 +85,20 @@ class OpenDDS_Dcps_Export TypeAssignability {
...
+    return tl_service->GetTypeObject(ti).minimal;
+  }
+
+public:
+  void insert_entry(const TypeIdentifier& ti, const TypeObject& tobj)
+  {
+    tl_service->AddTypeObjectsToCache(ti, tobj);
+  }
+
+  static std::map<unsigned int, MinimalTypeObject> table_;
```
sonndinh, line 101: `table_` can be removed. It served to emulate the cache for `TypeIdentifier` to `TypeObject` mapping, but now we have a real one.

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -213,6 +213,11 @@ namespace XTypes {
     {
       std::memcpy(hash, a_hash, sizeof hash);
     }
+
+    bool operator<(const TypeObjectHashId& other) const
+    {
+      return (kind < other.kind) && hash < other.hash;
```
sonndinh, line 219: Minor comment but this could be `return (kind < other.kind) && (hash < other.hash);` for consistency.

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: Isn't MonotonicTimePoint::now() a static operation?

kuznetsovmoci: Or maybe I should ask, what would be the difference of using it from the argument?

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: Implemented - needs verification

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1200,253 @@ namespace OpenDDS {
...
+                (s_iter->second.is_submessage_protected || s_iter->second.is_payload_protected)) {
+                create_and_send_datawriter_crypto_tokens(dwch, writer, drch, reader);
+              }
+            }
+          }
+        }
+#endif
+
+        // Copy reader and writer association data prior to releasing lock
+#ifdef __SUNPRO_CC
```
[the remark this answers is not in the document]
mitza-oci: Yes, please remove

dds/DCPS/DiscoveryBase.h
mitza-oci: Do you want to "sample" the system clock again, or use a previous sample?

dds/DCPS/DiscoveryBase.h
```diff
@@ -212,14 +213,30 @@ namespace OpenDDS {
         , publication_counter_(0)
         , subscription_counter_(0)
         , topic_counter_(0)
+        , reactor_task_(false)
+        , max_reply_period(5)  // 5 sec sufficient?
```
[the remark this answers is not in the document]
mitza-oci: It should be configurable using the RtpsDiscovery object

dds/DCPS/DiscoveryBase.h
iguessthislldo: > Do you want to "sample" the system clock again, or use a previous sample?

Yes, that's what I really meant by my question. I don't actually know the answer, I was just wondering if it was considered.

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
kuznetsovmoci: #ifdef __SUNPRO_CC section?

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: My intention was to sample the system clock again, and check if the time expired has exceeded the predefined value

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
kuznetsovmoci: Implemented

dds/DCPS/DiscoveryBase.h
```diff
@@ -1249,24 +1251,30 @@ namespace OpenDDS {
...
+            drQos = md->second.drQos;
+            subQos = md->second.subQos;
+            wTls = md->second.wTls;
+            rTls = md->second.rTls;
+            writer_type_info = md->second.writer_type_info;
+            reader_type_info = md->second.reader_type_info;
+            cfProp = md->second.cfProp;
+            matching_data_buffer_.erase(writer);
+          }
+          else {
```
iguessthislldo, line 1271:
```suggestion
          } else {
```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1101,39 +1101,41 @@ namespace OpenDDS {
...
+            md.reader = reader;
+            md.time_added_to_map = MonotonicTimePoint::now();
+
+            // store in the map by writer
+            MatchingDataIter md_it = matching_data_buffer_.find(writer);
+            if (md_it != matching_data_buffer_.end()) {
+              // TLS_TODO: is this scenario possible?
+              md_it->second = md;
+            }
+            else {
```
iguessthislldo:
```suggestion
            } else {
```

dds/DCPS/DomainParticipantImpl.h
```diff
@@ -372,6 +372,8 @@ class OpenDDS_Dcps_Export DomainParticipantImpl
   void add_adjust_liveliness_timers(DataWriterImpl* writer);
   void remove_adjust_liveliness_timers();
 
+  XTypes::TypeLookupService_rch GetTypeLookupService() { return type_lookup_service_; }
```
iguessthislldo, line 375:
```suggestion
  XTypes::TypeLookupService_rch getTypeLookupService() { return type_lookup_service_; }
```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1201,263 @@ namespace OpenDDS {
...
+#else
+             "", "",
+#endif
+             cfProp->expressionParameters, octet_seq_type_info_reader };
+
+        DDS::OctetSeq octet_seq_type_info_writer;
+        XTypes::serialize_type_info(*writer_type_info, octet_seq_type_info_writer);
+        const WriterAssociation wa =
+        { add_security_info(*wTls, writer, reader), writer, *pubQos, *dwQos, octet_seq_type_info_writer };
+#endif
```
iguessthislldo, line 1422: Continuation of [this](https://github.com/objectcomputing/OpenDDS/pull/1840/files#r470381737)

Here's the change, also with some formatting tweaks to the non-sun code since it's off:

```suggestion
        DDS::OctetSeq octet_seq_type_info_reader;
        XTypes::serialize_type_info(*reader_type_info, octet_seq_type_info_reader);
        const ReaderAssociation ra = {
          add_security_info(*rTls, writer, reader), reader, *subQos, *drQos,
#ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
          cfProp->filterClassName, cfProp->filterExpression,
#else
          "", "",
#endif
          cfProp->expressionParameters,
          octet_seq_type_info_reader
        };

        DDS::OctetSeq octet_seq_type_info_writer;
        XTypes::serialize_type_info(*writer_type_info, octet_seq_type_info_writer);
        const WriterAssociation wa = {
          add_security_info(*wTls, writer, reader), writer, *pubQos, *dwQos,
          octet_seq_type_info_writer
        };
```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1201,263 @@ namespace OpenDDS {
...
+          if (call_reader) {
+            thr.activate();
+          }
+          dwr->add_association(writer, ra, writer_active);
+          if (call_reader) {
+            thr.writer_done();
+          }
+
+        }
+        else if (call_reader) {
```
iguessthislldo, line 1442:
```suggestion
        } else if (call_reader) {
```

dds/DCPS/Discovery.h
```diff
@@ -22,6 +22,8 @@
 #include "dds/DCPS/PoolAllocator.h"
 #include "dds/DCPS/PoolAllocationBase.h"
 
+#include "dds/DCPS/XTypes/TypeLookupService.h"
```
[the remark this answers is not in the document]
iguessthislldo: Still needs to be done

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
iguessthislldo: Yes, this looks like a workaround for the compiler used on Sun's Solaris operating system, which we don't support anymore.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3661,20 +3688,51 @@ Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& ty
...
-  // ser >> type_lookup_request;
-  return DDS::RETCODE_OK;
+  // TLS_TODO: verify request processing
+  XTypes::TypeLookup_Request type_lookup_request;
+
+  if (!(ser >> type_lookup_request)) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Sedp::TypeLookupRequestReader::take_tl_request - ")
+              ACE_TEXT("failed to deserialize type lookup request\n")));
+    return DDS::RETCODE_ERROR;
+    }
```
iguessthislldo, line 3700:
```suggestion
  }
```

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
iguessthislldo: [Suggested Change](https://github.com/objectcomputing/OpenDDS/pull/1840#discussion_r473516431)

dds/DCPS/XTypes/TypeLookupService.h
iguessthislldo: https://github.com/objectcomputing/OpenDDS/blob/3354da116616da4f05476a964ae97c83a191bad5/dds/DCPS/Serializer.h#L43-L45

If you need an example.

dds/DCPS/XTypes/TypeLookupService.h
iguessthislldo: This still isn't done.

iguessthislldo: Never mine I was looking at the wrong place.

dds/DCPS/Discovery.h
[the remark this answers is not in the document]
iguessthislldo: Nevermind.

dds/DCPS/XTypes/TypeLookupService.h
iguessthislldo: Yet again nevermind.

dds/DCPS/DiscoveryBase.h
iguessthislldo: I think that's probably good.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1231,244 @@ namespace OpenDDS {
         }
       }
 
+      void
+      remove_expired_endpoints(const DCPS::MonotonicTimePoint& /*now*/)
```
iguessthislldo, line 1235:
```suggestion
      remove_expired_endpoints(const MonotonicTimePoint& /*now*/)
```

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1231,244 @@ namespace OpenDDS {
...
+      {
+        // TLS_TODO: sanity check on locks
```
iguessthislldo, line 1237: I think it's good.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1231,244 @@ namespace OpenDDS {
...
+        ACE_GUARD(ACE_Thread_Mutex, g1, matching_data_buffer_lock_);
+        MatchingDataIter it;
+        for (it = matching_data_buffer_.begin(); it != matching_data_buffer_.end(); it++) {
+          if (MonotonicTimePoint::now() - it->second.time_added_to_map >= max_type_lookup_service_reply_period_) {
+            matching_data_buffer_.erase(it->first);
+          }
+        }
+        // TLS_TODO: any other cleanup?
```
iguessthislldo, line 1245: I don't think so. Besides anything you're doing I don't think there's anything to cleanup, because all the action is in `match_continue`.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3160,6 +3163,22 @@ Sedp::signal_liveliness_unsecure(DDS::LivelinessQosPolicyKind kind)
   }
 }
 
+
+bool Sedp::send_type_lookup_request(XTypes::TypeIdentifierSeq& type_ids,
+                                    const DCPS::RepoId& reader)
+{
+  // TLS_TODO: verify reader
```
iguessthislldo, line 3170: Like that the guid is a reader? I'm not sure that's necessary here, but you can use `GuidConverter(reader).isReader()` to do that.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3349,7 +3368,7 @@ Sedp::SedpWriter::write_participant_message(const ParticipantMessageData& pmd,
 
 #ifdef OPENDDS_SECURITY
 DDS::ReturnCode_t
-Sedp::SedpWriter::write_stateless_message(const DDS::Security::ParticipantStatelessMessage& msg,
+Sedp::SecurityWriter::write_stateless_message(const DDS::Security::ParticipantStatelessMessage& msg,
                                           const RepoId& reader,
```
iguessthislldo, line 3372: The indent is off here and in changes below this.

dds/DCPS/RTPS/TypeLookup.h
```diff
@@ -2,11 +2,13 @@
 #define OPENDDS_RTPS_TYPE_LOOKUP_H_
 
 /*
-* Distributed under the OpenDDS License.
-* See: http://www.opendds.org/license.html
-*/
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
```
iguessthislldo, line 7: This comment is usually the very first thing in file if the file has one.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1051,170 +1121,68 @@ namespace OpenDDS {
...
+
+          // TLS_TODO: rpc communication needs to be tested when XTypes are ready
+          if (!writer_local) {
+            if (type_lookup_service_ && !type_lookup_service_->type_object_in_cache(writer_type_info->minimal.typeid_with_size.type_id)) {
+              XTypes::TypeIdentifierSeq type_ids;
+              type_ids.append(writer_type_info->minimal.typeid_with_size.type_id);
+              md.rpc_sequence_number = ++(type_lookup_service_->rpc_sequence_number_);
+              MatchingDataIter md_it = matching_data_buffer_.find(MatchingPair(writer, reader));
+              if (md_it != matching_data_buffer_.end()) {
+                // TLS_TODO: is this scenario possible?
```
iguessthislldo, line 1149: I'm not sure I see the problem. The data has to get into `matching_data_buffer_`, so there would be a point where it's not there.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1231,244 @@ namespace OpenDDS {
...
+        }
+        // TLS_TODO: any other cleanup?
+      }
+
+      void
+      match_continue(OpenDDS::DCPS::SequenceNumber rpc_sequence_number)
+      {
+        // TLS_TODO: sanity check on locks
+        // TLS_TODO: match_continue(it->second.writer, it->second.reader)
+        // will unlock '_lock'. Should it be locked here?
```
iguessthislldo, line 1253: I see two reverse locks, is that what you're talking about?

dds/DCPS/DiscoveryBase.h
```diff
@@ -130,7 +132,7 @@ namespace OpenDDS {
     };
 
     template <typename DiscoveredParticipantData_>
-    class EndpointManager {
+    class EndpointManager : public virtual DCPS::RcEventHandler {
```
mitza-oci, line 135: Changing the base class of `EndpointManager` may require more changes than I see in this PR.  Making it an RcObject (via RcEventHandler) would impact how instances are created, use of RcHandle, etc.

kuznetsovmoci: Is there a better (easier) way to implement a timer (PmfSporadicTask?) without changing the base class of EndpointManager?

mitza-oci: This is more of a design question, which is usually difficult to answer in this little text box in a PR review.  We can certainly discuss with the team outside of this review.

I haven't reviewed the related code yet, so I'm not sure what the "receiver" of this sporadic task needs to be -- maybe we could move it.

Also I'm not saying that changing the base class of EndpointManager is necessarily wrong, just that it has to be carried through to other uses of EndpointManager including derived classes.

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: Probably both... In one thread they get locked in match(), and unlocked in the 'lower level' match_continue(), which is also called from this function. The callers of this function (in Sedp) do not lock (at least I don't see that). So in this thread lower level match_continue() would unlock objects that were never locked. That's why I am locking them in this function. So I wanted to verify if this makes sense.

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: I am just wondering if we should expect a situation when the same two endpoints try to match more than once.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1289,6 +1289,15 @@ DataReaderImpl::enable()
     XTypes::TypeInformation type_info;
     typesupport->to_type_info(type_info);
 
+    XTypes::TypeIdentifierTypeObjectPairSeq ti_to_pair_seq;
+    XTypes::TypeIdentifierTypeObjectPair ti_to_pair;
+    ti_to_pair.type_identifier = type_info.minimal.typeid_with_size.type_id;
+    ti_to_pair.type_object = typesupport->getMinimalTypeObject();
+    ti_to_pair_seq.append(ti_to_pair);
+    // TLS_TODO: verify implementation
```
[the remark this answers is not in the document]
kuznetsovmoci: What I meant is "is this the right place to add type objects to cache?" and "Should there be other places in the code where they are added?"

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: I am more concerned about possible structures that existed prior to this project. For example, when a new endpoint has been discovered and may have been added to some sort of 'structure' for further processing, but now does not reply to the type lookup service request and is essentially 'useless'. Do such 'structures' exist, and if so should the discovered endpoint be removed from them?

dds/DCPS/DiscoveryBase.h
```diff
@@ -1051,170 +1120,68 @@ namespace OpenDDS {
...
+              type_ids.append(reader_type_info->minimal.typeid_with_size.type_id);
+              md.rpc_sequence_number = ++(type_lookup_service_->rpc_sequence_number_);
+              MatchingDataIter md_it = matching_data_buffer_.find(MatchingPair(writer, reader));
+              if (md_it != matching_data_buffer_.end()) {
+                // TLS_TODO: is this scenario possible?
+                md_it->second = md;
+              } else {
+                matching_data_buffer_.insert(std::make_pair(MatchingPair(writer, reader), md));
               }
+              //send_type_lookup_request(type_ids, reader);
```
mitza-oci, line 1170: Why is this commented-out?

dds/DCPS/DiscoveryBase.h
```diff
@@ -1263,6 +1230,244 @@ namespace OpenDDS {
...
+        if (reader_local && !writer_local) {
+          add_assoc_i(reader, lsi->second, writer, dpi->second);
+        }
+
+        if (!call_writer && !call_reader) {
+          return; // nothing more to do
+        }
+
+#ifdef OPENDDS_SECURITY
+        if (is_security_enabled()) {
```
mitza-oci, line 1339: refactor this block to its own method?

mitza-oci (changes requested):
still not a complete review

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -659,6 +664,32 @@ namespace XTypes {
...
+      case XTypes::EK_MINIMAL:
+        for (size_t i = 0; i < 14; ++i) {
+          if (equivalence_hash()[i] > other.equivalence_hash()[i]) {
+            return false;
+          } else if (equivalence_hash()[i] < other.equivalence_hash()[i]) {
+            return true;
+          }
+        }
+        return false;
+      default:
```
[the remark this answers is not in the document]
kuznetsovmoci: I thought they would not have type object associated with them. Or is it just for completeness?

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -118,8 +89,12 @@ class OpenDDS_Dcps_Export TypeAssignability {
   bool get_struct_member(const MinimalTypeObject*& ret, const CommonStructMember& m) const;
   bool get_union_member(const MinimalTypeObject*& ret, const CommonStructMember& m) const;
 
-private:
-  TypeLookup typelookup_;
+  TypeLookupService* tl_service_;
```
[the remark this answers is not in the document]
kuznetsovmoci: Do you mean to ask if `TypeLookupService` is designed to be copied?

dds/DCPS/RTPS/Sedp.h
```diff
@@ -563,7 +590,7 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
 
     DDS::ReturnCode_t take_tl_request(const DCPS::ReceivedDataSample& sample,
       DCPS::Serializer& ser,
-      XTypes::TypeLookup_Request& type_lookup_request);
+      XTypes::TypeLookup_Reply& type_lookup_reply);
```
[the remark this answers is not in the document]
kuznetsovmoci: It populates TypeLookup_Reply structure after deserializing the request

dds/DCPS/XTypes/TypeAssignability.h
```diff
@@ -17,55 +18,25 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
-};
-
 // Set of pairs of members with each pair contains members from
 // two structure types that have the same member ID and name
 typedef std::pair<const MinimalStructMember*, const MinimalStructMember*> MemberPair;
 typedef OPENDDS_VECTOR(MemberPair) MatchedSet;
 
 class OpenDDS_Dcps_Export TypeAssignability {
 public:
+  explicit TypeAssignability(XTypes::TypeLookupService_rch tls) : tl_service_(tls) {};
```
sonndinh, line 28:
```suggestion
  explicit TypeAssignability(XTypes::TypeLookupService_rch tls) : tl_service_(tls) {}
```

dds/DCPS/XTypes/TypeAssignability.h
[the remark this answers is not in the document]
sonndinh: `TypeAssignability` doesn't have any internal state other than `TypeLookupService`, so I think whether copying is allowed depends on how we're going to use `TypeLookupService`.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -378,25 +387,22 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
...
+      size_t size,
+      DCPS::MessageId id,
+      DCPS::SequenceNumber seq = DCPS::SequenceNumber());
+
   private:
     DCPS::SequenceNumber seq_;
   };
 
-  class SedpWriter : public Writer {
+  class SecurityWriter : public Writer {
```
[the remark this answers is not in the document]
kuznetsovmoci: I've looked into that earlier, but it seemed less beneficial than writer refactoring. `assoc()`, `data_received()`, and `remove_associations`  probably belong to the base Reader class. `notify_subscription_disconnected`, `notify_subscription_reconnected`, `notify_subscription_lost`, (and `remove_associations`) are empty in sedp. If we decide going deeper, refactoring may impact TransportReceiveListener and DataReaderImpl, which would be more invasive than it was for the writer. So I was not sure if was worth the effort, but if it is I'd like to have a design discussion first.

dds/DCPS/DiscoveryBase.h
kuznetsovmoci: After uncommenting many Jenkins tests fail:
https://prb.objectcomputing.com/jenkins/job/OpenDDS_Pull_Requests/2834/.
I re-ran some of them locally and they pass. Not sure yet if it's due to some Type Identifiers not being added to the map as they should, or because the comparison operator does not work.

dds/DCPS/DiscoveryBase.h
```diff
@@ -892,6 +913,54 @@ namespace OpenDDS {
...
+        const DDS::SubscriberQos* subQos;
+        TransportLocatorSeq* wTls;
+        TransportLocatorSeq* rTls;
+        XTypes::TypeInformation* reader_type_info;
+        XTypes::TypeInformation* writer_type_info;
+        const ContentFilterProperty_t* cfProp;
+        MonotonicTimePoint time_added_to_map;
+      };
+
+      struct MatchingPair {
```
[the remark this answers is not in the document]
kuznetsovmoci: std::pair does not seem to have 'less than' operator. We can derive our matching pair class from std::pair and add 'less than' operator, but I am not sure it would be much better, than what we have (?).

dds/DCPS/DiscoveryBase.h
```diff
@@ -1051,170 +1120,68 @@ namespace OpenDDS {
...
-                  local_reader_security_attribs_.find(reader);
-                // Yes, this is different for remote datawriters than readers (see 8.8.9.3 vs 8.8.9.2)
-                if (s_iter != local_reader_security_attribs_.end() && s_iter->second.is_submessage_protected) {
-                  create_and_send_datareader_crypto_tokens(drch, reader, dwch, writer);
-                }
+          MatchingData md;
+          ACE_GUARD(ACE_Thread_Mutex, g1, matching_data_buffer_lock_);
+
+          // if the type object is not in cache, send RPC request
+          md.dwQos = dwQos;
```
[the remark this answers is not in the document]
kuznetsovmoci: I've made a fix for those members that have local scope. It looks like other members' lifetime is sufficient. Please review.

dds/DCPS/XTypes/TypeAssignability.h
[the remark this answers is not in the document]
mitza-oci: > Do you mean to ask if `TypeLookupService` is designed to be copied?

I didn't.

Unless I missed something, the TypeAssignability class is copyable but contains a pointer to TypeLookupService.  This is a problem because it assumes the lifetime of TypeLookupService exceeds that of (all copies of) TypeAssignability.  Using a smart pointer would help this.  Or if TypeAssignability's use was more constrained (as in, couldn't be copied and we can make guarantees about its lifetime relative to TypeLookupService), this could be a reference or at least a const pointer.

dds/DCPS/RTPS/Sedp.h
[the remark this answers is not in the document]
mitza-oci: Using the name "take" tells me that it has the semantics of a DataReader's "take" method.  This is doing more, so maybe the name should be something like "process", as in as service processes a request in order to generate a reply.

dds/DCPS/XTypes/TypeAssignability.h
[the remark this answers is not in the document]
kuznetsovmoci: I've changed it to the smart pointer.

dds/DCPS/RTPS/Sedp.h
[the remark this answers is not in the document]
mitza-oci: I agree with the points you've made here.  However I think if we look deeper at what the current Reader does, `Sedp::Reader::data_received` has a large if/else-if chain to determine what type of writer sent the data.  If we split up the Reader, this wouldn't be needed.  Each type of Reader would either know type to expect, or have a small list (2 or 3) options to select from.

So the larger point, does this belong with this PR?  I think it does since this PR is doing the same split on the Writer side.  It will help readability/maintainability going forward.

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
mitza-oci: https://en.cppreference.com/w/cpp/utility/pair/operator_cmp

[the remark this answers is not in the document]
kuznetsovmoci: Interesting... When I tried it yesterday, compiler complained and when I looked at the class source I did not see it there. I have implemented it now and it compiles.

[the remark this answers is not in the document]
kuznetsovmoci: I misspoke - it still does not compile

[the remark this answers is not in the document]
kuznetsovmoci: 1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): error C2678:  binary '<': no operator found which takes a left-hand operand of type 'const _Ty1' (or there is no acceptable conversion)
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): error C2678:         with
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): error C2678:         [
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): error C2678:             _Ty1=OpenDDS::DCPS::RepoId
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): error C2678:         ]
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\system_error(263,24): message :  could be 'bool std::operator <(const std::error_condition &,const std::error_condition &) noexcept'
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\system_error(258,24): message :  or       'bool std::operator <(const std::error_code &,const std::error_code &) noexcept'
1>C:\Users\kuznetsovm\source\repos\OpenDDSX\dds\DCPS\TimeDuration.h(83,26): message :  or       'bool OpenDDS::DCPS::operator <(const OpenDDS::DCPS::TimeDuration &,const OpenDDS::DCPS::TimeDuration &)' [found using argument-dependent lookup]
1>C:\Users\kuznetsovm\source\repos\OpenDDSX\dds\DCPS\Time_Helper.h(61,6): message :  or       'bool OpenDDS::DCPS::operator <(const DDS::Duration_t &,const DDS::Duration_t &)' [found using argument-dependent lookup]
1>C:\Users\kuznetsovm\source\repos\OpenDDSX\dds\DCPS\Time_Helper.h(84,6): message :  or       'bool OpenDDS::DCPS::operator <(const DDS::Time_t &,const DDS::Time_t &)' [found using argument-dependent lookup]
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): message :  while trying to match the argument list '(const _Ty1, const _Ty1)'
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): message :         with
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): message :         [
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): message :             _Ty1=OpenDDS::DCPS::RepoId
1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.22.27905\include\utility(265,1): message :         ]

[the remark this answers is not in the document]
mitza-oci: Right, it would delegate to `operator<` for the GUID which would need to be provided just like the Duration_t/Time_t ones that the compiler is pointing out.

This is not essential, so if it becomes more trouble than it's worth we can go with the code as-is.  However I'd like to rely on the standard library types as much as we can.

dds/DCPS/XTypes/TypeObject.h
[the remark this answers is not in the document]
kuznetsovmoci: Implemented. Please review.

dds/DCPS/DiscoveryBase.h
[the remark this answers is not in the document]
kuznetsovmoci: Implemented for all `md` members

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3661,20 +3687,51 @@ Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& ty
 DDS::ReturnCode_t
 Sedp::TypeLookupRequestReader::take_tl_request(const DCPS::ReceivedDataSample& sample,
                                                DCPS::Serializer& ser,
-                                               XTypes::TypeLookup_Request& type_lookup_request)
+                                               XTypes::TypeLookup_Reply& type_lookup_reply)
 {
-  // TODO: uncomment; anything else to do here?
-  // ser >> type_lookup_request;
-  return DDS::RETCODE_OK;
+  // TLS_TODO: verify request processing
```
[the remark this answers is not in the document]
kuznetsovmoci: I will remove these comments before final check-in

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3661,20 +3687,51 @@ Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& ty
...
 DDS::ReturnCode_t
 Sedp::TypeLookupReplyReader::take_tl_reply(const DCPS::ReceivedDataSample& sample,
-                                           DCPS::Serializer& ser,
-                                           XTypes::TypeLookup_Reply& type_lookup_reply)
+                                           DCPS::Serializer& ser)
 {
-  // TODO: uncomment; anything else to do here?
-  // ser >> type_lookup_reply;
+  XTypes::TypeLookup_Reply type_lookup_reply;
+  // TLS_TODO: verify reply processing
```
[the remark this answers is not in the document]
kuznetsovmoci: I will remove 'TODO' comments before final cleanup

dds/DCPS/XTypes/TypeObject.h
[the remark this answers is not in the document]
mitza-oci: Yes, I see it in the latest revision.  My only thought on it at the moment is that there are two examples of `return (EXPR);` that can be just `return EXPR;`.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3152,6 +3155,20 @@ Sedp::signal_liveliness_unsecure(DDS::LivelinessQosPolicyKind kind)
...
+                                    const DCPS::RepoId& reader)
+{
+  const DCPS::RepoId remote_reader = make_id(reader, ENTITYID_TL_SVC_REPLY_READER);
+  DCPS::SequenceNumber sequence = 0;
+
+  return type_lookup_request_writer_->send_type_lookup_request(type_ids,
+                                                               remote_reader,
+                                                               sequence,
+                                                               type_lookup_service_->rpc_sequence_number_,
+                                                               participant_id_)  == DDS::RETCODE_OK;
```
mitza-oci, line 3169:
```suggestion
                                                               participant_id_) == DDS::RETCODE_OK;
```
Also - I try to avoid the "arguments falling over a cliff" code formatting.  We may not have a better option here.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3619,12 +3642,14 @@ Sedp::TypeLookupRequestWriter::send_type_lookup_request(XTypes::TypeIdentifierSe
...
-Sedp::TypeLookupReplyWriter::send_type_lookup_reply(XTypes::TypeLookup_Reply& type_lookup_reply,
-                                                    const RepoId& reader,
-                                                    DCPS::SequenceNumber& sequence,
-                                                    DDS::RPC::RemoteExceptionCode_t exception_code)
+Sedp::TypeLookupReplyWriter::send_tl_reply(const DCPS::ReceivedDataSample& sample,
+                                           XTypes::TypeLookup_Reply& type_lookup_reply)
 {
-  type_lookup_reply.header.remote_ex = exception_code;
+  DCPS::RepoId reader = sample.header_.publication_id_;
+  reader.entityId = ENTITYID_TL_SVC_REPLY_READER;
```
mitza-oci, line 3649: Use `make_id`?

dds/DCPS/XTypes/External.h
```diff
@@ -48,6 +48,8 @@ class External {
 
   bool operator==(const External& other) const { return ptr_ == other.ptr_; }
   bool operator!=(const External& other) const { return ptr_ != other.ptr_; }
+  bool operator<(const External& other) const { return ptr_ < other.ptr_; }
+  bool operator>(const External& other) const { return ptr_ > other.ptr_; }
```
mitza-oci, line 52: Is `operator>` needed for any of these?  The idiom we inherit from the standard library is to use `operator<` for associative container comparator.  If you need `a > b` you can write `b < a`.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -550,38 +632,49 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
...
 
     virtual ~TypeLookupReplyReader();
 
-    DDS::ReturnCode_t take_tl_reply(const DCPS::ReceivedDataSample& sample,
+  private:
+    virtual void data_received_i(const DCPS::ReceivedDataSample& sample,
+      const DCPS::EntityId_t& entity_id,
       DCPS::Serializer& ser,
-      XTypes::TypeLookup_Reply& type_lookup_reply);
+      const DCPS::Extensibility extensibility);
```
mitza-oci, line 675:
```suggestion
      DCPS::Extensibility extensibility);
```
multiple occurrences of this one

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3763,240 +3838,300 @@ Sedp::Reader::data_received(const DCPS::ReceivedDataSample& sample)
...
-                   ACE_TEXT("failed to deserialize data\n")));
-        return;
-      }
+void
+Sedp::LivelinessReader::data_received_i(const DCPS::ReceivedDataSample& sample,
+  const DCPS::EntityId_t& entity_id,
+  DCPS::Serializer& ser,
+  const DCPS::Extensibility extensibility)
+{
+  ACE_UNUSED_ARG(extensibility);
```
mitza-oci, line 3856:
```suggestion
```
In this case simply remove the name from the method's parameter list.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3763,240 +3838,300 @@ Sedp::Reader::data_received(const DCPS::ReceivedDataSample& sample)
...
-                   ACE_TEXT("failed to deserialize data\n")));
-        return;
-      }
+void
+Sedp::SecurityReader::data_received_i(const DCPS::ReceivedDataSample& sample,
+  const DCPS::EntityId_t& entity_id,
+  DCPS::Serializer& ser,
+  const DCPS::Extensibility extensibility)
+{
+ACE_UNUSED_ARG(sample);
```
mitza-oci, line 3891:
```suggestion
  ACE_UNUSED_ARG(sample);
```
These could be moved to the preprocessor "else" block

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,64 @@
...
+  void get_type_objects(const TypeIdentifierSeq& type_ids,
+    TypeIdentifierTypeObjectPairSeq& types);
+  void add_type_objects_to_cache(TypeIdentifierTypeObjectPairSeq& types);
+
+  // For adding local endpoint types
+  void add_type_objects_to_cache(const DCPS::TypeSupportImpl& typesupport);
+
+  bool type_object_in_cache(const TypeIdentifier& ti);
+
+  OpenDDS::DCPS::SequenceNumber rpc_sequence_number_;
```
mitza-oci, line 43: Does this need to be public?
Also, don't add `OpenDDS::` qualification

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace XTypes {
+
+TypeLookupService::TypeLookupService()
+{
+  to_empty_.minimal.kind = TK_NONE;
+  to_empty_.complete.kind = TK_NONE;
+  rpc_sequence_number_ = 0;
```
mitza-oci, line 18: Use constructor's member initializer list

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+TypeLookupService::~TypeLookupService()
+{
+}
+
+
+void TypeLookupService::get_type_objects(const TypeIdentifierSeq& type_ids,
+                                         TypeIdentifierTypeObjectPairSeq& types)
+{
+  for (CORBA::ULong i = 0; i < type_ids.length(); ++i) {
+    TypeObjectMap::iterator it_object = type_object_map_.find(type_ids[i]);
```
mitza-oci, line 31:
```suggestion
    const TypeObjectMap::iterator it_object = type_object_map_.find(type_ids[i]);
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (it_object != type_object_map_.end()) {
+      TypeIdentifierTypeObjectPair new_to_pair(it_object->first, it_object->second);
```
mitza-oci, line 33: const it, or just combine it with the next line (direct constructor call to make a temporary object).

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+      types.append(new_to_pair);
+    }
+  }
+}
+
+
+const TypeObject& TypeLookupService::get_type_objects(const TypeIdentifier& type_id) const
+{
+  TypeObjectMap::const_iterator it_object = type_object_map_.find(type_id);
```
mitza-oci, line 42:
```suggestion
  const TypeObjectMap::const_iterator it_object = type_object_map_.find(type_id);
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return it_object->second;
+  }
+  return to_empty_;
+}
+
+
+void TypeLookupService::add_type_objects_to_cache(TypeIdentifierTypeObjectPairSeq& types)
+{
+  for (ACE_UINT32 i = 0; i < types.length(); ++i) {
+    TypeObjectMap::iterator it_type_id_with_size_seq = type_object_map_.find(types[i].type_identifier);
```
mitza-oci, line 53:
```suggestion
    const TypeObjectMap::iterator it_type_id_with_size_seq = type_object_map_.find(types[i].type_identifier);
```
I'm going to stop looking for these and let you take over at this point

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (it_type_id_with_size_seq == type_object_map_.end()) {
+      type_object_map_.insert(std::pair<TypeIdentifier, TypeObject>(types[i].type_identifier, types[i].type_object));
```
mitza-oci, line 55: Would `make_pair()` be cleaner?

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
kuznetsovmoci: I am trying to keep same `virtual data_received_i()` function for all derived classes, as they are  called from the base class function.

dds/DCPS/XTypes/TypeLookupService.h
kuznetsovmoci: It has to be public the way code is using it now. Do you want me to change it to private with access functions?

dds/DCPS/XTypes/TypeLookupService.cpp
kuznetsovmoci: Is it possible for `to_empty_.minimal.kind = TK_NONE;`?

dds/DCPS/XTypes/External.h
kuznetsovmoci: I believe this was a result of me using `>` operator in `TypeObject.h` functions, like:

    bool operator<(const PlainMapLTypeDefn& other) const
    {
      if (header < other.header) return true;
      if (header > other.header) return false;
      if (bound < other.bound) return true;
      if (bound > other.bound) return false;
      if (element_identifier < other.element_identifier) return true;
      if (element_identifier > other.element_identifier) return false;
      if (key_flags < other.key_flags) return true;
      if (key_flags > other.key_flags) return false;
      if (key_identifier < other.key_identifier) return true;
      if (key_identifier > other.key_identifier) return false;
      return false;
    }
  };

If I rewrite them without using `>` operators this one may not be needed.
Would you like me to do that?

kuznetsovmoci: Implemented as suggested

dds/DCPS/XTypes/TypeLookupService.h
kuznetsovmoci: Made private and added public API

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: This is not related to parameter names.  The idea is that if you have function definition `void foo(int a) {}`, the language doesn't need the `a` to be there, since it's not used.  This avoids a warning and doesn't need an ACE MACRO.  The presence or absence of a name is not part of a function's type signature and doesn't impact callers or virtual overrides.

dds/DCPS/XTypes/TypeLookupService.cpp
mitza-oci: No, I'm only "pointing to" the single line.  GitHub does a terrible job of rendering it...

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // For Type Lookup Service request / reply
+  void get_type_objects(const TypeIdentifierSeq& type_ids,
+    TypeIdentifierTypeObjectPairSeq& types);
+  void add_type_objects_to_cache(TypeIdentifierTypeObjectPairSeq& types);
+
+  // For adding local endpoint types
+  void add_type_objects_to_cache(const DCPS::TypeSupportImpl& typesupport);
+
+  bool type_object_in_cache(const TypeIdentifier& ti);
```
mitza-oci, line 41:
```suggestion
  bool type_object_in_cache(const TypeIdentifier& ti) const;
```

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -0,0 +1,74 @@
...
+  TypeLookupService();
+  ~TypeLookupService();
+
+  // For TypeAssignability
+  const TypeObject& get_type_objects(const TypeIdentifier& type_id) const;
+  void add_type_objects_to_cache(const TypeIdentifier& ti, const TypeObject& tobj);
+
+  // For Type Lookup Service request / reply
+  void get_type_objects(const TypeIdentifierSeq& type_ids,
+    TypeIdentifierTypeObjectPairSeq& types);
```
mitza-oci, line 35:
```suggestion
  void get_type_objects(const TypeIdentifierSeq& type_ids,
    TypeIdentifierTypeObjectPairSeq& types) const;
```

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.h
```diff
@@ -592,7 +592,7 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
     virtual void data_received_i(const DCPS::ReceivedDataSample& sample,
       const DCPS::EntityId_t& entity_id,
       DCPS::Serializer& ser,
-      DCPS::Extensibility extensibility);
+      DCPS::Extensibility);
```
mitza-oci, line 595: The method declaration shouldn't change for this, it's an implementation detail

dds/DCPS/DiscoveryBase.h
iguessthislldo: I'm not sure. It sounds like it shouldn't be possible, but I don't think how you're handling it is going to hurt anything.

dds/DCPS/DiscoveryBase.h
iguessthislldo: No, I don't think so. I would only expect things like that to happen for builtin endpoints. Regular entities shouldn't do anything until they've gotten the go ahead that they've been matched.

dds/DCPS/DiscoveryBase.h
iguessthislldo: Yea, I don't know about this one. I can say it does seem weird to grab a lock just for a function we're calling to unlock it. It doesn't help that `lock_` implies it's just one object-wide lock.

iguessthislldo (approved):

dds/DCPS/XTypes/TypeLookupService.h
sonndinh: This hasn't been changed.

dds/DCPS/XTypes/TypeLookupService.h
sonndinh: Also this.

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -306,6 +323,15 @@ namespace XTypes {
...
       , element_flags(a_element_flags)
     {}
+
+    bool operator<(const PlainCollectionHeader& other) const
+    {
+      if (equiv_kind < other.equiv_kind) return true;
+      if (other.equiv_kind < equiv_kind) return false;
+      if (element_flags < other.element_flags) return true;
+      if (other.element_flags < element_flags) return false;
+      return false;
```
sonndinh, line 333:
```suggestion
      if (equiv_kind < other.equiv_kind ||
          (equiv_kind == other.equiv_kind && element_flags < other.element_flags)) {
        return true;
      }
      return false;
```

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -322,6 +348,18 @@ namespace XTypes {
...
+
+    bool operator<(const PlainSequenceSElemDefn& other) const
+    {
+      if (header < other.header) return true;
+      if (other.header < header) return false;
+      if (bound < other.bound) return true;
+      if (other.bound < bound) return false;
+      if (element_identifier < other.element_identifier) return true;
+      if (other.element_identifier < element_identifier) return false;
+      return false;
```
sonndinh, line 360:
```suggestion
      if (header < other.header ||
          (header == other.header && bound < other.bound) ||
          (header == other.header && bound == other.bound && element_identifier < other.element_identifier)) {
        return true;
      }
      return false;
```

sonndinh: I suggest similar changes to `operator<` of `PlainSequenceLElemDefn`, `PlainArraySElemDefn`, `PlainArrayLElemDefn`, `PlainMapSTypeDefn`, `PlainMapLTypeDefn`

dds/DCPS/XTypes/TypeObject.h
iguessthislldo: Return the if condition?

dds/DCPS/XTypes/TypeObject.h
iguessthislldo: Return the if condition?

sonndinh: It returns true or false for `operator<`.

iguessthislldo: I meant this:

```
return header < other.header || (header == other.header && bound < other.bound) ||
  (header == other.header && bound == other.bound && element_identifier < other.element_identifier);
```

sonndinh: Yea, that works.

dds/DCPS/DiscoveryBase.h
jrw972: Locking seems to be correct on review.  May remove secondary lock in the future.

dds/DCPS/XTypes/TypeObject.h
kuznetsovmoci: I based current implementation based on suggestion from @mitza-oci . My understanding is that it is a more 'standard' structure comparison algorithm. My vote would be to keep it, unless it gives wrong result.

dds/DCPS/XTypes/TypeLookupService.h
kuznetsovmoci: I could not change it, because this function needs to change its object(s).

dds/DCPS/XTypes/TypeLookupService.h
kuznetsovmoci: I could not change it, because this function needs to change its object(s).

mitza-oci (approved):

PR #1846, opened 2020-08-19
XTypes: Update max_serialized_size and Bounded Checks for Appendable and Mutable Types

iguessthislldo:
To avoid the pre-allocation optimization taken by `DataWriter` on these types since we're deferring proper calculation of `max_serialized_size`. This is still a WIP.

Also since they are quite different, differentiate `max_serialized_size*` functions for primitives from `max_serialized_size` for topic types by renaming the former to `primitive_serialized_size*` to better reflect what those functions are doing.


dds/DCPS/DataWriterImpl_T.h
```diff
@@ -41,47 +41,26 @@ DataWriterImpl_T
...
-        buffer_size_ += header_size_;
-      }
-      if (key_only_bounded_) {
-        const KeyOnlyType key_only_x(x);
-        TraitsType::max_serialized_size(
-          encoding_, key_only_buffer_size_, key_only_x);
-        key_only_buffer_size_ += header_size_;
-      }
+      bound_ = MarshalTraitsType::serialized_size_bound(encoding_);
+      key_only_bound_ = MarshalTraitsType::key_only_serialized_size_bound(encoding_);
```
mitza-oci, line 63: Can these move to the initializer list?

dds/DCPS/TypeSupportImpl.h
```diff
@@ -36,6 +38,41 @@ template <typename Message> struct DDSTraits;
...
 
+class OpenDDS_Dcps_Export SerializedSizeBound {
+public:
+  SerializedSizeBound()
+  : bounded_(false)
+  , bound_(0)
+  {
+  }
+
+  SerializedSizeBound(size_t bound)
```
mitza-oci, line 49:
```suggestion
  explicit SerializedSizeBound(size_t bound)
```

mitza-oci (changes requested):

jrw972 (approved):

dds/DCPS/RTPS/TypeLookup.cpp
```diff
@@ -349,11 +349,11 @@ namespace DCPS {
   void serialized_size(const Encoding& encoding, size_t& size,
     const XTypes::OctetSeq32& seq)
   {
-    DCPS::serialized_size_ulong(encoding, size);
+    DCPS::primitive_serialized_size_ulong(encoding, size);
```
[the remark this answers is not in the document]
iguessthislldo: In this case no.

dds/DCPS/DataSampleHeader.cpp
```diff
@@ -407,9 +405,7 @@ DataSampleHeader::split(const ACE_Message_Block& orig, size_t size,
     hdr.message_length_ = 0; // no room for payload data
     *head << hdr;
     const size_t avail = size - head->length() - 4 /* sequence length */;
-    size_t guid_size = 0;
-    max_serialized_size(encoding, guid_size, GUID_t());
-    const CORBA::ULong n_entries = static_cast<CORBA::ULong>(avail / guid_size);
+    const CORBA::ULong n_entries = static_cast<CORBA::ULong>(avail / guid_cdr_size);
```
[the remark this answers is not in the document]
iguessthislldo: No idea, but the existing code seems to assume it is. I don't think it would hurt to round down though since we're not sure.

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: Yes

dds/DCPS/TypeSupportImpl.h
iguessthislldo: I think this is one of the rare case where we would want an implicit conversion, since the class is basically an optional integer value. In addition to that and probably supporting that is the fact that it would be a bit of a pain to go through and add all constructors to the tests were they would be needed if it was explicit.

tests/DCPS/KeyTest/KeyMarshalling.cpp
```diff
@@ -31,6 +31,23 @@ void print_hex(void* d)
...
+  std::cout << "  bound = " << actual_bound.to_string() << std::endl;
+  if (actual_bound != expected_bound) {
+    ACE_ERROR((LM_ERROR, "ERROR: expected bound to be %C, but it is %C\n",
+       expected_bound.to_string().c_str(), actual_bound.to_string().c_str()));
+    TEST_CHECK(false);
+  }
+
+  OpenDDS::RTPS::KeyHash_t hash;
+  OpenDDS::RTPS::marshal_key_hash(value, hash);
+  print_hex(hash.value);
```
iguessthislldo, line 48: So I just simplified what the test was doing, but shouldn't it be checking that the result of the hash is right?

dds/DCPS/DataSampleHeader.cpp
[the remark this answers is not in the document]
iguessthislldo: Actually isn't the division already rounding it down to the number of whole GUIDs?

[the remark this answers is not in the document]
mitza-oci: Yes, it must round down so the result fits in the space available.

tests/DCPS/KeyTest/KeyMarshalling.cpp
mitza-oci: I'm OK with this not being in scope.  Key hashes are not used.

dds/DCPS/TypeSupportImpl.h
```diff
@@ -36,6 +38,41 @@ template <typename Message> struct DDSTraits;
...
+  : bounded_(true)
```
mitza-oci, line 50: If `bound` is 0, should this be `false`?

iguessthislldo: If a 0 is passed in, then it's bounded to 0. Setting `bounded_` to `false` would make it unbounded.

mitza-oci: I guess I was thinking this was one of the cases where 0 indicates unbounded.  If it's not, nothing to do here...

mitza-oci (approved):

dds/DCPS/TypeSupportImpl.h
iguessthislldo: Yeah the reason behind this class to enforce the distinction, because the existence of that rule in addition to a separate bounded function was kinda weird.

sonndinh (approved):

PR #1847, opened 2020-08-20
Bench Stats Calcuation Bug Fix For Median & MAD Stats

simpsont-oci:
Median output buffers weren't always getting output correctly, and per-worker median stats had an off-by-one error in their calculation.

### DIFF
--- a/performance-tests/bench_2/common/PropertyStatBlock.cpp
+++ b/performance-tests/bench_2/common/PropertyStatBlock.cpp
@@ -187,7 +187,7 @@ void PropertyStatBlock::finalize()
   Builder::DoubleSeq ds;
   ds.length(static_cast<CORBA::ULong>(count));
   for (size_t i = 0; i < count; ++i) {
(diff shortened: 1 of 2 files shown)


performance-tests/bench_2/common/PropertyStatBlock.cpp
```diff
@@ -196,9 +196,9 @@ void PropertyStatBlock::finalize()
     // calculate median
     std::sort(&median_buffer_[0], &median_buffer_[count - 1]);
     if (count % 2 == 0) {
-      median_result = (median_buffer_[count / 2] + median_buffer_[(count / 2) + 1]) / 2.0;
+      median_result = (median_buffer_[(count / 2) - 1] + median_buffer_[count / 2]) / 2.0;
     } else {
-      median_result = median_buffer_[(count / 2) + 1];
+      median_result = median_buffer_[(count / 2)];
     }
     // calculate median absolute deviation (median of absolute valute of data deviation from median)
```
mitza-oci, line 203:
```suggestion
    // calculate median absolute deviation (median of absolute value of data deviation from median)
```

PR #1851, opened 2020-08-21
Observer Interface

lij-oci:
This PR provides an interface for OpenDDS applications to record events and data samples between DataWriter and DataReader. The events currently supported are: DataWriter/DataReader ENABLED, DELETED, QOS_CHANGED, ASSOCIATED, DISASSOCIATED; and SAMPLE_SENT, SAMPLE_RECEIVED, SAMPLE_READ, and SAMPLE_TAKEN. An application can register an observer with these events at the individual DataWriter/DataReader level, at the Publisher/Subscriber level, or at the DomainParticipant level. OpenDDS applications provide the implementation of the Observer interface to take specific actions when these events occur, e.g. logging, or saving to a database.


jrw972 (approved):

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1638,6 +1659,12 @@ DataReaderImpl::data_received(const ReceivedDataSample& sample)
         sample.header_.message_id_));
     break;
   }
+
+  Observer::Rch observer = get_observer(Observer::e_SAMPLE_RECEIVED);
+  if (observer) {
+    Observer::Sample s(sample, instance ? instance->instance_handle_ : 0);
```
jwillemsen, line 1665: Shouldn't 0 be replaced with DDS::HANDLE_NIL

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1152,6 +1170,12 @@ namespace OpenDDS {
   } else {
     const RakeData item = group_coherent_ordered_data_.get_data();
     results.insert_sample(item.rde_, item.si_, item.index_in_instance_);
+    if (observer) {
+      typename InstanceMap::iterator i = instance_map_.begin();
+      const DDS::InstanceHandle_t handle = (i != instance_map_.end()) ? i->second : 0;
```
jwillemsen, line 1175: DDS::HANDLE_NIL?

tests/DCPS/Observer/run_test.pl
```diff
@@ -0,0 +1,27 @@
...
+
+$test->start_process('subscriber');
+$test->start_process('publisher');
+
+my $result = $test->finish(30);
+if ($result != 0) {
+  print STDERR "ERROR: test returned $result\n";
+}
+
+exit $result;
```
mitza-oci, line 27: This test needs to be added to the overall test list.

dds/DCPS/EntityImpl.h
```diff
@@ -59,6 +60,10 @@ class OpenDDS_Dcps_Export EntityImpl
 
   virtual RcHandle<EntityImpl> parent() const { return RcHandle<EntityImpl>(); }
 
+  void set_observer(Observer::Rch observer, const Observer::Event e);
```
mitza-oci, line 63:
```suggestion
  void set_observer(Observer::Rch observer, Observer::Event e);
```

dds/DCPS/EntityImpl.h
```diff
@@ -59,6 +60,10 @@ class OpenDDS_Dcps_Export EntityImpl
...
+
+  Observer::Rch get_observer(const Observer::Event e);
```
mitza-oci, line 65:
```suggestion
  Observer::Rch get_observer(Observer::Event e);
```

dds/DCPS/Observer.cpp
```diff
@@ -0,0 +1,73 @@
...
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+CORBA::ULong Observer::Sample::to_instance_state(const char message_id)
+{
+  switch (message_id) {
+    case UNREGISTER_INSTANCE:
+      return DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE;
```
mitza-oci, line 24:
```suggestion
  switch (message_id) {
  case UNREGISTER_INSTANCE:
    return DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE;
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+class ReceivedDataSample;
+class ReceivedDataElement;
+
+class OpenDDS_Dcps_Export Observer
+  : public virtual RcObject
+{
+public:
+  typedef Observer* Ptr;
+  typedef TAO_Objref_Var_T<Observer> Var;
+  typedef TAO_Objref_Out_T<Observer> Out;
```
mitza-oci, line 39: Are these used?

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+  typedef RcHandle<Observer> Rch;
+
+  typedef unsigned long Event;
+  enum {
```
mitza-oci, line 43: Why not just `enum Event {` here?

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+  };
+
+  struct OpenDDS_Dcps_Export Sample
+  {
+    DDS::InstanceHandle_t instance_;
+    CORBA::ULong instance_state_;
+    DDS::Time_t timestamp_;
+    SequenceNumber seq_n_;
+    const void* const data_;
+    static CORBA::ULong to_instance_state(const char message_id);
```
mitza-oci, line 64:
```suggestion
    static CORBA::ULong to_instance_state(char message_id);
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+    Sample(const DDS::InstanceHandle_t i, const DataSampleElement& e, const DDS::Time_t& t);
```
mitza-oci, line 65:
```suggestion
    Sample(DDS::InstanceHandle_t i, const DataSampleElement& e, const DDS::Time_t& t);
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+    Sample(const ReceivedDataSample& s, const DDS::InstanceHandle_t i = 0);
```
mitza-oci, line 66:
```suggestion
    explicit Sample(const ReceivedDataSample& s, DDS::InstanceHandle_t i = 0);
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+    Sample(const ReceivedDataElement& s, const DDS::InstanceHandle_t i = 0, const CORBA::ULong instance_state = 0);
```
mitza-oci, line 67:
```suggestion
    explicit Sample(const ReceivedDataElement& s, DDS::InstanceHandle_t i = 0, CORBA::ULong instance_state = 0);
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+    Sample(const DDS::InstanceHandle_t i, const CORBA::ULong instance_state,
```
mitza-oci, line 68:
```suggestion
    Sample(DDS::InstanceHandle_t i, const CORBA::ULong instance_state,
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+      const DDS::Time_t& t, const SequenceNumber& sn, const void* const data);
```
mitza-oci, line 69:
```suggestion
      const DDS::Time_t& t, const SequenceNumber& sn, const void* data);
```

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+  // 4) Sample received by DataReader
+  //    - ReaderId, Same data fields as "sample sent" above
+  virtual void on_sample_received(const DDS::DataReader_ptr, const Sample&) {}
+
+  // 5) Sample received by application (read/take from DataReader)
+  //    - ReaderId, Same data fields as "sample sent" above, read/take context
+  virtual void on_sample_read(const DDS::DataReader_ptr, const Sample&) {}
+  virtual void on_sample_taken(const DDS::DataReader_ptr, const Sample&) {}
+
+  virtual ~Observer() {}
```
mitza-oci, line 106: Typically these are defined in the .cpp file since it can determine where typeinfos and vtables are generated (depending on the compiler)

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+protected:
+  Observer() {}
+};
+
+// comply with specification
```
mitza-oci, line 111: Which specification?

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -320,7 +320,12 @@ DataReaderImpl::add_association(const RepoId& yourId,
   //is held here anyway
   guard.release();
 
-  if (!associate(data, active)) {
+  if (associate(data, active)) {
+    Observer::Rch observer = get_observer(Observer::e_ASSOCIATED);
```
mitza-oci, line 324:
```suggestion
    const Observer::Rch observer = get_observer(Observer::e_ASSOCIATED);
```
There are many of these so I haven't marked each one.  Please add `const` to local variables that aren't changed after initialization.

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,122 @@
...
+class ReceivedDataElement;
+
+class OpenDDS_Dcps_Export Observer
+  : public virtual RcObject
+{
+public:
+  typedef Observer* Ptr;
+  typedef TAO_Objref_Var_T<Observer> Var;
+  typedef TAO_Objref_Out_T<Observer> Out;
+  typedef RcHandle<Observer> Rch;
```
mitza-oci, line 40: Is there another example of this style of typedef in the codebase?

mitza-oci (changes requested):

dds/DCPS/Observer.h
lij-oci: At this high level interface, it's probably a good thing to provide these commonly used definitions. Currently these three of the four definitions are only used for defining their DDS-spec-complied aliases.

dds/DCPS/Observer.h
lij-oci: If we define enum Event {A=1, B=2, C=4}; we can not use Event as a function parameter or returned type, unless we only pass or return A, B, or C by itself. For example:
void register(Observer o, Event e); we can not do this: register(o, A | B); 
because (A | B) is not a valid value in enum Event {A, B, C}.

dds/DCPS/DataReaderImpl.cpp
lij-oci: I strongly agree that a variable (not only local) unchanged after initialization should be defined as const. This also entails that only const methods of that object are accessible. In this particular case, we can make const Observer::Rch observer only if we also make all the three groups of interface methods const. That was what I was not sure about in the design phase, i.e. how these interface methods and most importantly their derived implementations will be used. For example, in the Observer test, TestObserver needs to update a specific statistics when a specific event handler is called. It will be difficult to take a specific action like that if these interface methods are const. In other words, we want to make our code robust, but we don't want to limit its usefulness.

dds/DCPS/Observer.h
lij-oci: Sorry, my mistake. I saw things like SomeClass_rch, or DDS::StatusCondition_var, almost everywhere, so I just assumed either DDS spec or C++ mapping may have specified these *_var, *_ptr, *_rch, etc. I thought since this is a high level interface, I'd better follow suit. I would gladly delete these four lines here if they are not required. I like things that belong to a class be defined in that class like the standard library does. So permission to delete these?

dds/DCPS/Observer.h
lij-oci: I did not see in our codebase yet, but the idea is consistently used in the standard library, and I used it for many years for defining custom types, iterators, exceptions, etc.

dds/DCPS/Observer.h
mitza-oci: The `_var`, `_ptr`, and `_out` types are from IDL-to-C++.  There is no IDL here.  Let's remove these.

dds/DCPS/DataReaderImpl.cpp
mitza-oci: You can add const here without changing other code. It's the `RcHandle` that's being constrained, not the thing being pointed to.  This is the same way that `shared_ptr` works.  There are some examples in the code already.
```
TransportClient.cpp:677:  const DataLink_rch link = found->second;
```

lij-oci: I see. 
It's very confusing that after this line: const DataLink_rch link = found->second;
the link is then used to call these mutating methods: 
link->release_reservations(peerId, repo_id_, released);
link->remove_listener(repo_id_);

These behaviors are made possible by RcHandle's const operator->, that is: 
T* operator->() const
{
    return this->ptr_;
}

Technically, this changes the meaning of the const keyword in variable definition here, if not completely defeats its purpose.

Changes are made as suggested.

mitza-oci: It is working as designed, and just as shared_ptr does.  This is the difference between `const T*` and `T* const`.

dds/DCPS/Observer.h
mitza-oci: This doesn't seem to be resolved in the latest code.

dds/DCPS/Observer.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "DataSampleElement.h"
+#include "ReceivedDataElementList.h"
+#include <dds/DCPS/transport/framework/ReceivedDataSample.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+CORBA::ULong Observer::Sample::to_instance_state(char message_id)
```
mitza-oci, line 20:
```suggestion
DDS::InstanceStateKind Observer::Sample::to_instance_state(char message_id)
```

dds/DCPS/Observer.cpp
```diff
@@ -0,0 +1,92 @@
...
+  : instance_(i)
+  , instance_state_(to_instance_state(instance_state))
+  , timestamp_(s.source_timestamp_)
+  , seq_n_(s.sequence_)
+  , data_(s.registered_data_)
+{}
+
+Observer::Sample::Sample(
+  DDS::InstanceHandle_t i,
+  CORBA::ULong instance_state,
```
mitza-oci, line 59:
```suggestion
  InstanceStateKind instance_state,
```

dds/DCPS/Observer.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+void Observer::on_associated(DDS::DataWriter_ptr, const GUID_t& /* readerId */) {}
+void Observer::on_associated(DDS::DataReader_ptr, const GUID_t& /* writerId */) {}
+void Observer::on_disassociated(DDS::DataWriter_ptr, const GUID_t& /* readerId */) {}
+void Observer::on_disassociated(DDS::DataReader_ptr, const GUID_t& /* writerId */) {}
+
+void Observer::on_sample_sent(const DDS::DataWriter_ptr, const Sample&) {}
+void Observer::on_sample_received(const DDS::DataReader_ptr, const Sample&) {}
+void Observer::on_sample_read(const DDS::DataReader_ptr, const Sample&) {}
+void Observer::on_sample_taken(const DDS::DataReader_ptr, const Sample&) {}
```
mitza-oci, line 85: Do these need to be in the .cpp or could they just be inline in the class definition?

mitza-oci (changes requested):

dds/DCPS/Observer.h
lij-oci: The first three (Ptr, Var, Out) were deleted based on your another comment above. Rch is used. As for this style of typedef in the codebase, it's true that there are so many typedef's in the style of ClassName_var, but there are actually this style of typedef's in class declaration, too. For example, the TypeSupportVar in the original tests/DCPS/DataRepresentation/DataRepresentation.cpp (and the _var_type in TypeSupportType):

template<typename Type>
class RegisteredType {
public:
  typedef typename OpenDDS::DCPS::DDSTraits<Type> TypeTraits;
  typedef typename TypeTraits::TypeSupportType TypeSupport;
  typedef typename TypeSupport::_var_type TypeSupportVar;
  typedef typename TypeTraits::TypeSupportTypeImpl TypeSupportImpl;

  TypeSupportVar type_support_var_;

  RegisteredType()
  : type_support_var_(0)
  {
  }
//...

As long as no spec requires otherwise, I would like to suggest using this in-class-declaration style, since it's consistent with the standard library and with the more recent C++ practice. (e.g. before STL, an iterator of ClassX was most likely to be defined as class ClassX_Iterator; now it's simple ClassX::iterator.)

I hope the in-class typedef RcHandle<Observer> Rch; is acceptable here. 
If not, I can replace it with outside-class typedef RcHandle<Observer> Observer_rch; and replace all the use of Observer::Rch with Observer_rch.

dds/DCPS/Observer.cpp
lij-oci: They were inline before, and moved here based on your previous comment:
[
mitza-oci 5 days ago  Member
Typically these are defined in the .cpp file since it can determine where typeinfos and vtables are generated (depending on the compiler)
]

mitza-oci: The previous comment was only about the destructor.  This is a confusing part of the GitHub interface.  When the comment refers to a larger block of code, you'll see something like "Comment on lines +70 to +85" (as this one does).  The other one was referring to a single line, even though in the "Conversation" view GitHub shows multiple lines for context.

dds/DCPS/Observer.h
mitza-oci: We should follow the established patterns for how RcObject/RcHandle are used.

mitza-oci (approved):

dds/DCPS/Observer.h
```diff
@@ -0,0 +1,99 @@
...
+  virtual void on_associated(DDS::DataWriter_ptr, const GUID_t& /* readerId */) {}
+  virtual void on_associated(DDS::DataReader_ptr, const GUID_t& /* writerId */) {}
+  virtual void on_disassociated(DDS::DataWriter_ptr, const GUID_t& /* readerId */) {}
+  virtual void on_disassociated(DDS::DataReader_ptr, const GUID_t& /* writerId */) {}
+
+  // Group 3: Sample sent, received, read, taken
+  virtual void on_sample_sent(const DDS::DataWriter_ptr, const Sample&) {}
+  virtual void on_sample_received(const DDS::DataReader_ptr, const Sample&) {}
+  virtual void on_sample_read(const DDS::DataReader_ptr, const Sample&) {}
+  virtual void on_sample_taken(const DDS::DataReader_ptr, const Sample&) {}
```
mitza-oci, line 85:
```suggestion
  virtual void on_sample_sent(DDS::DataWriter_ptr, const Sample&) {}
  virtual void on_sample_received(DDS::DataReader_ptr, const Sample&) {}
  virtual void on_sample_read(DDS::DataReader_ptr, const Sample&) {}
  virtual void on_sample_taken(DDS::DataReader_ptr, const Sample&) {}
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #1853, opened 2020-08-24
Allow Fragmentation in RTPS Durability Resends

iguessthislldo:
Fixes issue with durable samples bigger than the max UDP message size failing because they were not being fragmented.

Also correct error logs in `RtpsUdpTransport.cpp`.

### DIFF
--- a/dds/DCPS/transport/framework/TransportCustomizedElement.h
+++ b/dds/DCPS/transport/framework/TransportCustomizedElement.h
@@ -23,18 +23,23 @@ class OpenDDS_Dcps_Export TransportCustomizedElement
   : public TransportQueueElement {
 
 public:
-  TransportCustomizedElement(TransportQueueElement* orig,
-                             bool fragment);
+  TransportCustomizedElement(TransportQueueElement* orig);
+
+  void set_fragment(TransportQueueElement* orig);
 
   virtual RepoId publication_id() const;
   void set_publication_id(const RepoId& id);
+
+  RepoId subscription_id() const;
+  void set_subscription_id(const RepoId& id);
+
+  SequenceNumber sequence() const;
+  void set_sequence(const SequenceNumber& value);
 
   virtual const ACE_Message_Block* msg() const;
   void set_msg(Message_Block_Ptr m);
 
   virtual const ACE_Message_Block* msg_payload() const;
-
-  virtual SequenceNumber sequence() const;
 
   virtual bool owned_by_transport() { return false; }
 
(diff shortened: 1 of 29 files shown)


dds/DCPS/transport/framework/TransportSendStrategy.cpp
```diff
@@ -1917,6 +1917,39 @@ TransportSendStrategy::mb_to_iov(const ACE_Message_Block& msg, iovec* iov)
   return num_blocks;
 }
 
+bool TransportSendStrategy::fragmentation_helper(
+  TransportQueueElement* original_element, TqeVector& elements_to_send)
+{
+  const size_t space = space_available();
```
mitza-oci, line 1923: `space_available()` depends on the current state of the Send Strategy (`header_.length_`) so I don't think it can be used here.  Not sure if that would be enough to fix the failing test.

dds/DCPS/transport/framework/TransportCustomizedElement.h
```diff
@@ -23,19 +23,24 @@ class OpenDDS_Dcps_Export TransportCustomizedElement
   : public TransportQueueElement {
 
 public:
-  TransportCustomizedElement(TransportQueueElement* orig,
-                             bool fragment);
+  TransportCustomizedElement(TransportQueueElement* orig);
```
mitza-oci, line 26:
```suggestion
  explicit TransportCustomizedElement(TransportQueueElement* orig);
```

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -716,12 +722,12 @@ RtpsUdpSendStrategy::stop_i()
 size_t RtpsUdpSendStrategy::max_message_size() const
 {
   // TODO: Make this conditional on if the message actually needs to do this.
-  return max_message_size_
+  return max_message_size_ ? max_message_size_
```
mitza-oci, line 725: When would `max_message_size_` be 0?

iguessthislldo:
I still need to work on the DelayedDurable test.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
iguessthislldo: I'm not completely sure, but there are checks for it in `send` and `get_packet_elems_from_queue`.

mitza-oci: If those are in the framework, that would be for other transport types.

iguessthislldo: Yea they are. It's probably for TCP. If that's the case I can remove it.

mitza-oci:
Replaced by #1884

PR #1857, opened 2020-08-26
Implement Wait-for-Discovery-Complete for Bench Worker

wilsongOCI:
Adds a delay to Bench 2 worker until all readers and writers are discovered or times out.

### DIFF
--- a/performance-tests/bench_2/worker/WorkerDataReaderListener.h
+++ b/performance-tests/bench_2/worker/WorkerDataReaderListener.h
@@ -8,6 +8,7 @@
 #include "dds/DCPS/DisjointSequence.h"
 
 #include <unordered_map>
+#include <condition_variable>
 
 namespace Bench {
 
--- a/performance-tests/bench_2/worker/WorkerDataReaderListener.h
+++ b/performance-tests/bench_2/worker/WorkerDataReaderListener.h
@@ -36,14 +36,10 @@ public:
   void set_datareader(Builder::DataReader& datareader) override;
   void unset_datareader(Builder::DataReader& datareader) override;
 
-  size_t get_match_count() { return match_count_; }
-  size_t get_expected_match_count() { return expected_match_count_; }
-
-  std::condition_variable dr_expected_match_cv;
-  std::mutex dr_expected_match_cv_mutex;
+  bool wait_for_expected_match(const std::chrono::system_clock::time_point& deadline) const;
 
 protected:
-  std::mutex mutex_;
+  mutable std::mutex mutex_;
   bool durable_{false};
   bool reliable_{false};
   bool history_keep_all_{false};
@@ -55,6 +51,7 @@ protected:
   Builder::DataReader* datareader_{0};
   DataDataReader_var data_dr_;
   std::vector<DataHandler*> handlers_;
+  mutable std::condition_variable expected_match_cv;
 
   Builder::PropertyIndex last_discovery_time_;
   Builder::PropertyIndex lost_sample_count_;
--- a/performance-tests/bench_2/worker/WorkerDataReaderListener.h
+++ b/performance-tests/bench_2/worker/WorkerDataReaderListener.h
@@ -35,7 +35,7 @@ public:
 
   void set_datareader(Builder::DataReader& datareader) override;
   void unset_datareader(Builder::DataReader& datareader) override;
-  
+
   size_t get_match_count() { return match_count_; }
   size_t get_expected_match_count() { return expected_match_count_; }
 
--- a/performance-tests/bench_2/worker/WorkerDataReaderListener.h
+++ b/performance-tests/bench_2/worker/WorkerDataReaderListener.h
@@ -35,6 +35,12 @@ public:
 
   void set_datareader(Builder::DataReader& datareader) override;
   void unset_datareader(Builder::DataReader& datareader) override;
+  
+  size_t get_match_count() { return match_count_; }
+  size_t get_expected_match_count() { return expected_match_count_; }
+
+  std::condition_variable dr_expected_match_cv;
+  std::mutex dr_expected_match_cv_mutex;
 
 protected:
   std::mutex mutex_;
--- a/performance-tests/bench_2/worker/main.cpp
+++ b/performance-tests/bench_2/worker/main.cpp
@@ -306,7 +306,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
             }
           }
         }
-               
+
         auto writeMap = process.get_writer_map();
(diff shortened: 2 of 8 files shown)


performance-tests/bench_2/worker/WorkerDataReaderListener.h
```diff
@@ -36,6 +36,12 @@ class WorkerDataReaderListener : public Builder::DataReaderListener {
   void set_datareader(Builder::DataReader& datareader) override;
   void unset_datareader(Builder::DataReader& datareader) override;
 
+  size_t get_match_count() { return match_count_; }
+  size_t get_expected_match_count() { return expected_match_count_; }
+
+  std::condition_variable dr_expected_match_cv;
+  std::mutex dr_expected_match_cv_mutex;
```
simpsont-oci, line 43: I think we should move the wait code inside the listener classes so we can make use of their internal mutex while we're checking match_count_ & expected_match_count_. So we wouldn't need the new mutex or the two new methods, but instead we'd add something like:

```
bool wait_for_expected_match(const std::chrono::system_clock::time_point& deadline) const {
  std::unique_lock<std::mutex> lock(mutex_);
  while (match_count_ != expected_match_count_) {
    if (wait_until(lock, timeout_time) == std::cv_status::timeout) {
      return match_count == expected_match_count_;
    }
  }
  return true;
}

```
Which would return true if the expected match was reached before deadline (otherwise false). Same for the writer listener. We'll need to make mutex_ mutable for the const to work, but that's fine. And then we just call those with our timeout and let the listeners worry about match counts etc.

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
...
     do_wait(config.start_time, "start");
 
+    Log::log() << "Starting Discovery Check." << std::endl;
+
+    process_start_discovery_time = Builder::get_time();
+
+    if (config.wait_for_discovery) {
+      if (config.wait_for_discovery_seconds > 0) {
+
+        std::chrono::seconds timeoutPeriod(config.wait_for_discovery_seconds);
```
simpsont-oci, line 279:
```suggestion
        const std::chrono::seconds timeoutPeriod(config.wait_for_discovery_seconds);
```

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
...
+        std::chrono::system_clock::time_point timeout_time;
```
simpsont-oci, line 280: I think we want to be calculating one global absolute timeout up front (currently it recalculates per reader / writer) and then we can pass that timeout in to each reader / writer listener. So this can be const and can do the calculation below (timeout_time = now + timeoutPeriod).

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
...
+        if (readMap.size() > 0) {
+          typedef std::map<std::string, std::shared_ptr<Builder::DataReader>>::iterator ReadMapIt;
+          std::shared_ptr<Builder::DataReader> dtRdrPtr(nullptr);
+          size_t dr_expected_match_count;
+          size_t dr_match_count;
+
+          for (ReadMapIt it = readMap.begin(); it != readMap.end(); ++it) {
+            dtRdrPtr = it->second;
+            Bench::WorkerDataReaderListener* wdrl = dynamic_cast<Bench::WorkerDataReaderListener*>(dtRdrPtr->get_dds_datareaderlistener().in());
+            std::unique_lock<std::mutex> expected_lock(wdrl->dr_expected_match_cv_mutex);
```
simpsont-oci, line 292: If we make the method described above, all this can change to:
```
if (wdrl->wait_for_expected_match(timeout_time)) {
  // Matched Case
} else {
  // Fail Case (This should log some form of the word "Error" so our test scripts can pick up on the issue)
}
```

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
...
+        if (writeMap.size() > 0) {
+          typedef std::map<std::string, std::shared_ptr<Builder::DataWriter>>::iterator WriteMapIt;
+          std::shared_ptr<Builder::DataWriter> dtWtrPtr(nullptr);
+          size_t dw_expected_match_count;
+          size_t dw_match_count;
+
+          for (WriteMapIt it = writeMap.begin(); it != writeMap.end(); ++it) {
+            dtWtrPtr = it->second;
+            Bench::WorkerDataWriterListener* wdwl = dynamic_cast<Bench::WorkerDataWriterListener*>(dtWtrPtr->get_dds_datawriterlistener().in());
+            std::unique_lock<std::mutex> expected_lock(wdwl->wr_expected_match_cv_mutex);
```
simpsont-oci, line 320: See comments above.

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
 
     do_wait(config.start_time, "start");
 
+    Log::log() << "Starting Discovery Check." << std::endl;
```
simpsont-oci, line 272: Let's move this whole section up a few lines before the "do_wait" for start. The wait for start and the wait for discovery to complete almost fill the same role for most tests, but if we need to wait for some reason after discovery but before we start our test actions (say, for durable data to finish being sent) that's what I envisioned the start_wait being use for.

performance-tests/bench_2/worker/main.cpp
```diff
@@ -268,6 +269,79 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
...
+
+    process_start_discovery_time = Builder::get_time();
```
simpsont-oci, line 274: These are now get_sys_time();

performance-tests/bench_2/worker/main.cpp
wilsongOCI: So are you looking to pass the timeoutPeriod to the wait_for_expected_match and then perform the now + timeoutPeriod in wait_for_expected_match?

simpsont-oci: No, I want to calculate it once up front (line 278?) and use the same (const?) timeout time for every call. No need to recalculate it for every entity.

PR #1858, opened 2020-08-26
XTypes: Fix Default Allowed Data Representation and Other Misc Fixes

iguessthislldo:
Set the allowed data representation of types without a annotation to all of them, or all true. I might have gotten mixed up with the fact that in `DataRepresentationAnnotation::node_value_exists`, when there is an annotation, the starting value has to be all false and then the annotations add values to it.

Also added some misc minor fixes, mostly cherry-picked from #1846 so that they might get merged faster.

### DIFF
--- a/dds/DdsDcpsCore.idl
+++ b/dds/DdsDcpsCore.idl
@@ -77,28 +77,33 @@ module DDS {
     };
     typedef sequence<BinaryProperty_t> BinaryPropertySeq;
 
-    //@appendable
+    // This is for when we support XCDR1.
+    // @appendable
     struct PropertyQosPolicy {
         PropertySeq        value;
         BinaryPropertySeq  binary_value;
     };
 
-    //@appendable
+    // This is for when we support XCDR1.
+    // @appendable
     struct UserDataQosPolicy {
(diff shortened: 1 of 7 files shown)


iguessthislldo:
Okay, so the problem with this is the core types I marked as mutable are now deserialized as parameter lists, even if unaligned encoding is being used. This could work, but it's never been tested and it's hitting a divide by zero when it goes to do an alignment.

iguessthislldo:
So I went ahead and fixed the immediate issue since it was simple, but there are almost certainly other things keeping mutable unaligned cdr from working, so I'm keeping those annotations in DdsDcpsCore.idl commented out.

dds/DdsDcpsCore.idl
```diff
@@ -77,28 +77,28 @@ module DDS {
     };
     typedef sequence<BinaryProperty_t> BinaryPropertySeq;
 
-    @appendable
+    //@appendable
```
mitza-oci, line 80: Add a comment for why this is here.

I think these need to be XCDR1 if/when we support that (Annex D IDL, comment at the top).

PR #1861, opened 2020-08-31
XTypes: Support recursive types

jrw972:


dds/idl/annotations.cpp
```diff
@@ -172,16 +208,15 @@ std::string KeyAnnotation::name() const
...
+  return value_from_appl(appl, default_value());
 }
 
 // @topic ====================================================================
 
 TopicAnnotation::TopicAnnotation()
 {
-  platforms_.insert("*");
-  platforms_.insert("DDS");
-  platforms_.insert("OpenDDS");
```
iguessthislldo, line 184: Where are you putting these back in? I think I've missed it.

dds/idl/be_global.cpp
```diff
@@ -748,12 +757,108 @@ ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
...
+  autoid_annotation->node_value_exists(node, value);
+  return value;
+}
+
+bool BE_GlobalData::id(AST_Decl* node, ACE_CDR::ULong& value) const
+{
+  IdAnnotation* id_annotation =
+    dynamic_cast<IdAnnotation*>(builtin_annotations_["::@id"]);
+  return id_annotation->node_value_exists(node, value);
+}
```
iguessthislldo, line 775: `BE_GlobalData::get_id`?

dds/idl/be_global.cpp
```diff
@@ -748,12 +757,108 @@ ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
...
+{
+  MustUnderstandAnnotation* must_understand_annotation =
+    dynamic_cast<MustUnderstandAnnotation*>(
+      builtin_annotations_["::@must_understand"]);
+  bool value = must_understand_annotation->absent_value;
+  must_understand_annotation->node_value_exists(node, value);
+  return value;
+}
+
+bool BE_GlobalData::is_key(AST_Decl* node) const
```
iguessthislldo, line 804: `BE_GlobalData::check_key`?

dds/idl/be_global.cpp
```diff
@@ -646,15 +646,22 @@ bool BE_GlobalData::is_topic_type(AST_Decl* node)
...
+  TopicValue value;
+  return topic->node_value_exists(node, value);
 }
 
 bool BE_GlobalData::is_nested(AST_Decl* node)
 {
+  // @topic overrides @nested and @default_nested.
+  if (is_topic_type(node)) {
+    return false;
+  }
```
iguessthislldo, line 658: You should revert these lines. `is_topic_type` is being called from other places in the compiler. The naming is probably a bit poor, but `is_nested` was just meant to be a helper to `is_topic_type`. If you need to know if something is "nested" you can just invert the result of `is_topic_type`.

dds/idl/annotations.cpp
jrw972: Why are they necessary?

dds/idl/be_global.cpp
jrw972: I need this to support hashid.

dds/idl/be_global.cpp
jrw972: Didn't want the extra parameter.

dds/idl/be_global.cpp
jrw972: I disagree.  According to the spec `@topic` overrides `@nested` and `@default_nested`.  I could define a non-topic type with `@nested`.  I don't know if that is relevant or meaningful, but it does seem to be legal.

dds/idl/be_global.cpp
iguessthislldo: `get_id` should be doing that, there's a todo comment for that.

dds/idl/be_global.cpp
iguessthislldo: The extra parameter is significant.

dds/idl/be_global.cpp
iguessthislldo: >  I could define a non-topic type with @nested.

That's just one of the definitions of a non-topic type. I think the spec handles this very poorly. These are all sides of the same coin, which is if a type is a topic type. By doing this you're breaking the existing functionality.

dds/idl/annotations.cpp
iguessthislldo: They are part of the definition of `@topic`, though I admit they're not really all that important. Still I don't see a reason to remove them.

jrw972: The previous code only allowed topics with a platform in that set.  I think my interpretation of the spec is different.  The platform is a string or pattern that indicates that the topic is specifically for DDS or for any system (*).  This perhaps allows for optimization, or security, or something else that the spec writers dreamed about.

I'm okay putting it back it but I want to clarify the intended behavior.  Specifically for the error case.  If a specify a platform that is not in that list, should I see an error or a warning?

dds/idl/be_global.cpp
jrw972: The support for hashid and others is done piece-wise in the typeobject generator.  Hence, I wanted a very specific function that just looked at `@id` and nothing else.  If we want to move that logic over because we think it is generally useful, then that's the right answer.

dds/idl/be_global.cpp
jrw972: There are three possibilities:  (1) the field is annotated with `@key` and a value, (2) the field is annotated with `@key` and no value, or (3) the field does not have the `@key` annotation.  My function returns false for (3) and (1) when the value is false.  This is adequate for my purposes.

dds/idl/be_global.cpp
jrw972: See corrected comment above.
You can have (1) nested non-topic types, (2) non-nested topic types, and (3) non-nested non-topic types.
How are we supposed to handle (3)?
If we need to handle (3) and want to preserve existing functionality, then we need a compelling reason that is well documented and communicated to the user.  Otherwise, I think we should implement the spec correctly and change the code accordingly.

iguessthislldo: I'm assuming you mean by these cases you mean annotations since there is no such thing as a "non-nested non-topic type".

A type with no `@nested` annotation or `@topic` defaults to the "default nested" property of the module which can go up to the root module's "default nested" property which can be set by the user using a command line argument.

I don't know what you mean by the last paragraph but this is all documented in the devguide in section 2.1.1.4
"Topic Types vs. Nested Types".

dds/idl/be_global.cpp
iguessthislldo: Okay, I'm still not familiar with type object, but you're talking about `MemberFlag.IS_KEY` just being a boolean option? That is troubling because the way our keys work is a little more complicated than that.

dds/idl/be_global.cpp
iguessthislldo: Then maybe `get_id` should be made up of these more primitive functions you're adding, if I understand what you're saying correctly.

dds/idl/annotations.cpp
iguessthislldo: > indicates that the topic is specifically for DDS or for any system (*)

By "topic" do you mean the topic it's talking about for the "name"? If so than maybe, but it's not giving an indication the two parameters are related.

jrw972: I agree that name and platform are not related unless a code generator takes actions based on the two, i.e., a generator that sees DDS for the platform always generates a sample reader and writer using the given topic name.

So, back to my original question:  Why are we whitelisting the platforms?  Why are we not printing an error for platforms that are not in the list?

dds/idl/be_global.cpp
jrw972: Look at https://github.com/objectcomputing/OpenDDS/pull/1861/files#diff-67dba08556ea6d6a4cef262508eaa880R1145 and then let me know if you want to move it to `get_id`.

dds/idl/be_global.cpp
jrw972: No.  What I'm saying is that we cannot assume that a non-topic type is nested.  Again, I don't know why a non-topic non-nested type would be useful, but a user can declare such a thing.  Thus, if code was checking for nested by looking at !topic, then we need to go change the code to actually check for nested.

dds/idl/be_global.cpp
jrw972: I'm assuming that something in the spec says that `@nested(FALSE)` is equivalent to `@topic`.  Where is that located?  If its not in there, then why did we make that assumption?

jrw972: So @oschwaldp-oci pointed me to 7.2.2.5.  I think what the spec writers intended was that type annotated with `@nested(TRUE)` are optimized because they must be nested, type annotated with `@nested(FALSE)` are candidates for being topics, and types annotated with `@topic` should definitely be treated as topics with the name and platform and such.  Thus, treating `@nested(FALSE)` as `@topic` will work but it may be more than the spec writers intended.

I'll correct `is_topic_type` and `is_nested`.

dds/idl/annotations.cpp
iguessthislldo: We don't always follow this rule, but the IDL spec says that annotations shouldn't really cause errors or warnings. In addition to that, I figured when this was set to something we don't recognize, it should be interpreted as an intentional action by the user. Although, like with unknown annotations, I wouldn't be against a configurable warning in that defaults to on in this case though.

dds/idl/be_global.cpp
iguessthislldo: I honestly don't think was that thought out. `@topic` was only added after I started working on the annotations. Adam filed an issue because the spec assumed topic types were the default and it was obvious it would be helpful to have something like `@topic` if we were keeping the original behavior in OpenDDS.

MPC/config/opendds_util.mpb
```diff
@@ -0,0 +1,6 @@
+project: dds_taolib {
+  after += OpenDDSUtil
```
mitza-oci, line 2:
```suggestion
  after += OpenDDS_Util
```
Might as well be consistent with the lib name.

dds/util.mpc
```diff
@@ -0,0 +1,25 @@
+project(OpenDDSUtil): acelib, install {
```
mitza-oci, line 1:
```suggestion
project(OpenDDS_Util): acelib, install {
```
consistent naming pattern with OpenDDS_Rtps, etc.

dds/DCPS/SafetyProfileStreams.h
```diff
@@ -20,16 +18,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
-OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(::CORBA::UShort to_convert);
+OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned short to_convert);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(int to_convert);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned int to_convert, bool as_hex = false);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(long to_convert);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(long long to_convert);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned long long to_convert, bool as_hex = false);
 OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned long to_convert, bool as_hex = false);
 
-OpenDDS_Dcps_Export const char* retcode_to_string(DDS::ReturnCode_t value);
```
iguessthislldo, line 31: Just a heads up. There are some tests that include this file to get this that might break.

dds/idl/annotations.h
```diff
@@ -76,105 +76,136 @@ class Annotation {
...
+   * If node does not have the annotation, this sets value to absent_value()
    * and returns false.
    */
   virtual bool node_value_exists(AST_Decl* node, T& value) const
   {
     AST_Annotation_Appl* appl = find_on(node);
-    value = value_from_appl(appl);
-    return appl;
-  }
+    if (!appl) { return false; }
```
iguessthislldo, line 122: This isn't setting a value like it should (or at least what the comment says).

dds/idl/be_global.cpp
```diff
@@ -770,17 +884,20 @@ OpenDDS::DataRepresentation BE_GlobalData::data_representations(
...
-    value = index;
+  std::string hash_id;
+  if (id(field, member_id)) {
+    // @id
+    return member_id++;
+  } else if (hashid(field, hash_id)) {
+    // @hashid
+    return OpenDDS::XTypes::hash_member_name_to_id(hash_id);
+  } else if (auto_id == autoidkind_hash) {
+    // auto_id == autoidkind_hash
```
iguessthislldo, line 897:
```suggestion
```

dds/idl/be_global.cpp
```diff
@@ -748,12 +766,108 @@ ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
...
+  TryConstructFailAction value = try_construct_annotation->absent_value;
+  try_construct_annotation->node_value_exists(node, value);
+  return value;
+}
+
+TryConstructFailAction BE_GlobalData::union_discriminator_try_construct(AST_Union* node)
+{
+  TryConstructAnnotation* try_construct_annotation =
+    dynamic_cast<TryConstructAnnotation*>(
+                                          builtin_annotations_["::@try_construct"]);
```
iguessthislldo, line 869:
```suggestion
    dynamic_cast<TryConstructAnnotation*>(builtin_annotations_["::@try_construct"]);
```

iguessthislldo (changes requested):

dds/idl/annotations.h
jrw972: That was a commenting fail.

MPC/config/opendds_util.mpb
```diff
@@ -3,4 +3,6 @@ project: dds_taolib {
   libs += OpenDDS_Util
   includes += $(DDS_ROOT)
   libpaths += $(DDS_ROOT)/lib
+  macros   += NOTAO
```
iguessthislldo, line 6: `OPENDDS_NOTAO`?

iguessthislldo: Nevermind, the library is just for `opendds_idl` right?

jrw972: Right.  Eventually, I think this will get renamed but I think we'll have a better idea of what that is in the future.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -308,7 +308,9 @@ DataReaderImpl::add_association(const RepoId& yourId,
 
   //get message block from octet seq then deser to a type info
   XTypes::TypeInformation ti;
-  XTypes::deserialize_type_info(ti, writer.serializedTypeInfo);
+  DCPS::OctetSeq seq;
+  copy(seq, writer.serializedTypeInfo);
```
jwillemsen, line 312: Looks costly to each time copy the buffer just to pass it to a different operation which expects a OctetSeq in a different namespace

dds/DCPS/Discovery.cpp
```diff
@@ -13,6 +13,7 @@
 #include "DdsDcpsCoreC.h"
 #include "Marked_Default_Qos.h"
 #include "dds/DCPS/SafetyProfileStreams.h"
+#include "DCPS_Utils.h"
```
mitza-oci, line 16: This is the only line changed in this file.  Is it needed?

dds/DCPS/DataReaderImpl.cpp
mitza-oci: It's not ideal but this is not "critical path"

dds/DCPS/MultiTopicDataReaderBase.cpp
```diff
@@ -14,6 +14,7 @@
 #include "Marked_Default_Qos.h"
 #include "SubscriberImpl.h"
 #include "TypeSupportImpl.h"
+#include "DCPS_Utils.h"
```
mitza-oci, line 17: Is this include needed?

dds/DCPS/Serializer.cpp
```diff
@@ -25,6 +27,32 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
+}
+
+void string_free(ACE_CDR::Char* ptr)
+{
+  std::free(ptr);
+}
+
+ACE_CDR::WChar* wstring_alloc(ACE_CDR::ULong size)
+{
+  return static_cast<ACE_CDR::WChar*>(std::malloc(size));
```
mitza-oci, line 45: The `size` parameter to `wstring_alloc` is a measure of characters, not bytes.

dds/DCPS/SubscriberImpl.cpp
```diff
@@ -27,6 +27,7 @@
 #include "Util.h"
 #include "dds/DCPS/transport/framework/TransportImpl.h"
 #include "dds/DCPS/transport/framework/DataLinkSet.h"
+#include "DCPS_Utils.h"
```
mitza-oci, line 30: Is this needed?

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -11,22 +11,41 @@
...
-
-
-namespace DDS {
-  class OctetSeq;
-}
 namespace OpenDDS {
 namespace XTypes {
 
+  template<typename T, typename T_slice, typename TAG>
+  class Fake_TAO_Array_Forany_T {
```
mitza-oci, line 24: Rename?  Reformat?

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -814,6 +737,10 @@ namespace XTypes {
     NameHash name_hash;
 
     MinimalMemberDetail() {}
+    MinimalMemberDetail(const NameHash& a_name_hash)
+      {
```
mitza-oci, line 741: de-indent

mitza-oci (changes requested):
Partial review

dds/idl/be_global.cpp
jrw972: @iguessthislldo Please resolve this conversation or let me know what other information you need.

dds/DCPS/Discovery.cpp
jrw972: Yes.  `retcode_to_string` moved to DCPS_Utils.h.

dds/DCPS/MultiTopicDataReaderBase.cpp
jrw972: Yes.  Same as `Discovery.cpp`.

dds/DCPS/SubscriberImpl.cpp
jrw972: Same as `Discovery.cpp`.

dds/idl/annotations.h
iguessthislldo: This still isn't consistent with the comment.

iguessthislldo: Shouldn't it be setting the absent value?

dds/idl/annotations.h
```diff
@@ -222,6 +273,67 @@ class HashidAnnotation : public AnnotationWithValue<std::string> {
...
+  {}
+
+  std::string definition() const;
+  std::string name() const;
+
+  virtual const bool* default_value() const
+  {
+    static bool v = true;
+    return &v;
+  }
```
iguessthislldo, line 300: I don't understand what these `default_value` functions are supposed to be doing. If the annotation is there it has the value, if it's not then you use what you're calling `absent_value` (what I think used to be my `default_value`).

dds/idl/be_global.cpp
```diff
@@ -748,12 +766,107 @@ ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
...
+}
+
+bool BE_GlobalData::is_key(AST_Decl* node) const
+{
+  KeyAnnotation* key_annotation =
+    dynamic_cast<KeyAnnotation*>(
+      builtin_annotations_["::@key"]);
+  bool value = key_annotation->absent_value;
+  key_annotation->node_value_exists(node, value);
+  return value;
```
iguessthislldo, line 820:
```suggestion
  bool node_is_key;
  check_key(node, node_is_key);
  return value;
```

iguessthislldo (changes requested):

mitza-oci (changes requested):
Partial review: I haven't reviewed the `opendds_idl` parts yet

dds/idl/annotations.h
jrw972: If a field has a `@key` annotation, will it have a parameter list where the first parameter is a boolean true?  If so, all of the default stuff can go away.

How would you reword the comment to make it clearer?

dds/idl/annotations.h
jrw972: See comment about default_value above.  I wrote the code assume that annotations with default values would not have a parameter if no value was supplied.  Thus, we had to provide a default value.  Maybe this was just poor naming.  Thus, default value should return the default value as defined in the annotation.

dds/idl/be_global.cpp
jrw972: I prefer the exist coding as it is structurally more similar to similar functions.

dds/idl/annotations.h
iguessthislldo: > If a field has a @key annotation, will it have a parameter list where the first parameter is a boolean true? If so, all of the default stuff can go away.

Yes. An `AST_Annotation_Appl` will always have values for its fields, coming from either the defaults in the annotation definition or the annotation application. If an annotation application in IDL doesn't define all the non-defaulted fields of the definition, then it's a compiler error.

> How would you reword the comment to make it clearer?

The comment is saying it's setting the default value if the annotation isn't there, but it just returns if the node doesn't have the annotation. It used to be done by `value_from_appl`, but that's not happening anymore. I'm still not sure why you did that, but either the part in the comment has to be fixed or preferably you have the function set `value` when  the annotation is missing.

dds/idl/annotations.h
iguessthislldo: > See comment about default_value above. I wrote the code assume that annotations with default values would not have a parameter if no value was supplied. Thus, we had to provide a default value.

Okay that explains quite a bit. No that's not the case.

> Maybe this was just poor naming. Thus, default value should return the default value as defined in the annotation.

Why would you want that? Or is this part of the misunderstanding?

If it helps I think absent_value is a much better name for my `default_value`, since it can't be confused with the default values of the annotations themselves.

dds/idl/be_global.cpp
iguessthislldo: The similar functions don't have a near duplicate existing function like this one does. If you don't want to use `check_key` directly, please define this function in terms of `check_key`. It doesn't make sense for there more than one code path for this.

dds/DCPS/SafetyProfileStreams.h
```diff
@@ -20,16 +18,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
-OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(::CORBA::UShort to_convert);
+OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned short to_convert);
```
jwillemsen, line 21: Why not use the ACE_CDR typedefs?

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -11,22 +11,54 @@
...
 
 #include <algorithm>
 #include <cstring>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+namespace OpenDDS {
 
+namespace XTypes {
+  struct TypeInformation;
+};
```
mitza-oci, line 24:
```suggestion
}
```

dds/DCPS/SafetyProfileStreams.h
```diff
@@ -20,16 +20,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
-OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(::CORBA::UShort to_convert);
+OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(ACE_CDR::UShort to_convert);
```
mitza-oci, line 23:
```suggestion
OpenDDS_Dcps_Export OPENDDS_STRING to_dds_string(unsigned short to_convert);
```
These aren't related to CDR.

mitza-oci (changes requested):
Partial review: I haven't reviewed the `opendds_idl` parts yet

dds/DCPS/XTypes/TypeObject.h
mitza-oci: Marking this as "resolved" since we will handle it in a separate PR.

The approach I'm suggesting is to use `IDL::DistinctType<T>` from `Serializer.h`.  I think that change could be made here without impacting other uses.  We could probably do it when true mutable/appendable support is added for these types (Real Soon Now).

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+  case AST_Expression::EV_short:
+    return ev.u.sval;
+  case AST_Expression::EV_ushort:
+    return ev.u.usval;
+  case AST_Expression::EV_long:
+    return ev.u.lval;
+  case AST_Expression::EV_ulong:
+    return ev.u.ulval;
+  case AST_Expression::EV_longlong:
+    return ev.u.llval;
```
mitza-oci, line 67: What are the implications of narrowing/truncation here?

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+  default:
+    return 0;
+  }
+}
+
+void
+print(const OpenDDS::XTypes::TypeIdentifier& ti);
+
+void
+print_sbound(const OpenDDS::XTypes::SBound& bound)
```
mitza-oci, line 89: calling code may be simplified by making this an `operator<<`

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+  be_global->impl_ << ", " << scc.scc_length << ", " << scc.scc_index;
+  be_global->impl_ << ")";
+}
+
+void
+print(const OpenDDS::XTypes::TypeIdentifier& ti)
+{
+  switch (ti.kind()) {
+  case OpenDDS::XTypes::TK_NONE:
+    be_global->impl_ << "XTypes::TypeIdentifier(XTypes::TK_NONE)";
```
mitza-oci, line 299: Move the common prefix to before the `switch`?  Could avoid repetition in each branch.

Also should we have a function in TypeObject.h that gives you the stringized `TK_*`?

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+      if (be_global->is_nested(n)) {
+        to.minimal.struct_type.struct_flags |= OpenDDS::XTypes::IS_NESTED;
+      }
+
+      if (auto_id == autoidkind_hash) {
+        to.minimal.struct_type.struct_flags |= OpenDDS::XTypes::IS_AUTOID_HASH;
+      }
+
+      // TODO: Support inheritance.
+      to.minimal.struct_type.header.base_type = OpenDDS::XTypes::TypeIdentifier(OpenDDS::XTypes::TypeIdentifier(OpenDDS::XTypes::TK_NONE));
```
mitza-oci, line 1121:
```suggestion
      to.minimal.struct_type.header.base_type = OpenDDS::XTypes::TypeIdentifier(OpenDDS::XTypes::TK_NONE);
```

mitza-oci (changes requested):

dds/idl/annotations.cpp
```diff
@@ -104,25 +116,37 @@ AST_Expression::AST_ExprValue* get_annotation_member_ev(
...
 {
-  return get_annotation_member_ev(appl, member_name)->u.ulval;
+  AST_Expression::AST_ExprValue* ev = get_annotation_member_ev(appl, member_name);
+  if (!ev) {
+    idl_global->err()->misc_error(
+      (std::string("Found null pointer while getting value of member \"") +
+        member_name + "\" of annotation \"" +
+        appl->local_name()->get_string() + "\"").c_str(),
+      appl);
+    BE_abort();
```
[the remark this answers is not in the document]
iguessthislldo: Better yet moved the error back to `get_annotation_member_ev` where it used to be. It's not something specific to the type variants.

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
 #include "utl_identifier.h"
 
 #include "topic_keys.h"
+#include "dds_visitor.h"
+#include "dds/DCPS/Hash.h"
```
iguessthislldo, line 15:
```suggestion
#include "utl_identifier.h"
#include "topic_keys.h"
#include "dds_visitor.h"

#include <dds/DCPS/Hash.h>
```

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+  sorted_dependencies_.push_back(Element(type, dependencies_group, name));
+  in_sorted_dependencies_.insert(type);
+
+  return dependencies_group;
+}
+
+void
+typeobject_generator::generate_minimal_type_identifier(AST_Type* type)
+{
+  // Generate the minimal identifier ()and type object) and cache.
```
iguessthislldo, line 1006:
```suggestion
  // Generate the minimal identifier (and type object) and cache.
```

dds/idl/typeobject_generator.cpp
```diff
@@ -11,339 +11,1593 @@
...
+    for (SortedDependencies::const_iterator group_pos = group_begin; group_pos != group_end; ++group_pos) {
+      generate_minimal_type_identifier(group_pos->type);
+      seq.append(minimal_type_object_map_[group_pos->type]);
+    }
+
+    const OpenDDS::DCPS::Encoding& encoding = OpenDDS::XTypes::get_typeobject_encoding();
+    size_t size = serialized_size(encoding, seq);
+    ACE_Message_Block buff(size);
+    OpenDDS::DCPS::Serializer ser(&buff, encoding);
+    ser << seq;
```
iguessthislldo, line 1527: Check serialization result?

iguessthislldo (changes requested):

dds/idl/typeobject_generator.cpp
jrw972: This function is use for computing case labels so it would have to be a type that can be used as a union discriminator.
I'll change the code to abort for types that cannot be represented by a long.

dds/idl/annotations.cpp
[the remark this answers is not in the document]
jrw972: I'll do both.

dds/idl/typeobject_generator.cpp
jrw972: Agreed.

dds/idl/typeobject_generator.cpp
jrw972: Yes on the switch.

It should be easy to pull out when we need such a function.

dds/idl/typeobject_generator.cpp
mitza-oci: Seems like a bug in the spec that cases for `long long` discriminators can't be represented

jrw972: Agreed.  The relevant line from the spec is `typedef sequence<long> UnionCaseLabelSeq`.

dds/DCPS/TypeSupportImpl.cpp
```diff
@@ -49,8 +49,12 @@ TypeSupportImpl::get_type_name()
 
 void TypeSupportImpl::to_type_info(XTypes::TypeInformation& type_info) const
 {
-  const XTypes::TypeObject& minTypeObject = getMinimalTypeObject();
-  type_info.minimal.typeid_with_size.type_id = XTypes::makeTypeIdentifier(minTypeObject);
+  const XTypes::TypeIdentifier& minTypeId = getMinimalTypeIdentifier();
+  const XTypes::TypeMap& minTypeMap = getMinimalTypeMap();
+  XTypes::TypeMap::const_iterator pos = minTypeMap.find(minTypeId);
+  const XTypes::TypeObject& minTypeObject = pos->second;
```
mitza-oci, line 55: Is `find()` guaranteed to work, or should this be checked?

dds/DCPS/TypeSupportImpl.cpp
```diff
@@ -49,8 +49,12 @@ TypeSupportImpl::get_type_name()
 
 void TypeSupportImpl::to_type_info(XTypes::TypeInformation& type_info) const
 {
-  const XTypes::TypeObject& minTypeObject = getMinimalTypeObject();
-  type_info.minimal.typeid_with_size.type_id = XTypes::makeTypeIdentifier(minTypeObject);
+  const XTypes::TypeIdentifier& minTypeId = getMinimalTypeIdentifier();
+  const XTypes::TypeMap& minTypeMap = getMinimalTypeMap();
+  XTypes::TypeMap::const_iterator pos = minTypeMap.find(minTypeId);
```
mitza-oci, line 54:
```suggestion
  const XTypes::TypeMap::const_iterator pos = minTypeMap.find(minTypeId);
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -48,39 +48,35 @@ const TypeObject& TypeLookupService::get_type_objects(const TypeIdentifier& type
...
 void TypeLookupService::add_type_objects_to_cache(const DCPS::TypeSupportImpl& typesupport)
 {
-  XTypes::TypeInformation type_info;
-  typesupport.to_type_info(type_info);
-
-  const TypeObjectMap::iterator it_type_id_with_size_seq = type_object_map_.find(type_info.minimal.typeid_with_size.type_id);
-  if (it_type_id_with_size_seq == type_object_map_.end()) {
-    type_object_map_.insert(std::make_pair(type_info.minimal.typeid_with_size.type_id, typesupport.getMinimalTypeObject()));
-  }
+  // TODO: This populates the map N times instead of 1.
```
mitza-oci, line 61: TODO in this PR or later?

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -1288,10 +1237,16 @@ namespace XTypes {
...
+
+    MinimalCollectionElement() {}
+    MinimalCollectionElement(const CommonCollectionElement& a_common) : common(a_common) {}
   };
 
   struct CommonCollectionHeader {
     LBound bound;
+
+    CommonCollectionHeader() {}
+    CommonCollectionHeader(LBound a_bound) : bound(a_bound) {}
```
mitza-oci, line 1249:
```suggestion
    explicit CommonCollectionHeader(LBound a_bound) : bound(a_bound) {}
```

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -1775,70 +1759,65 @@ namespace XTypes {
...
+  void hash_member_name(NameHash& name_hash, const std::string& name);
 
-template<> OpenDDS_Dcps_Export
-XTypes::TypeIdentifier getMinimalTypeIdentifier<ACE_CDR::ULong>();
+  OpenDDS_Dcps_Export
+  bool is_fully_descriptive(const TypeIdentifier& ti);
 
-template<> OpenDDS_Dcps_Export
-XTypes::TypeIdentifier getMinimalTypeIdentifier<ACE_CDR::ULongLong>();
+  typedef std::map<TypeIdentifier, TypeObject> TypeMap;
```
mitza-oci, line 1794: In general I would suggest using the allocator-aware instantiations of the STL containers from the start.  At this point maybe we need to do it as a separate pass.

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -159,34 +204,524 @@ TypeIdentifier makeTypeIdentifier(const TypeObject& type_object)
...
+  case XTypes::TI_PLAIN_MAP_LARGE:
+    compute_dependencies(type_map, type_identifier.map_ldefn(), dependencies);
+    break;
+  case XTypes::TI_STRONGLY_CONNECTED_COMPONENT:
+    compute_dependencies(type_map, type_identifier.sc_component_id(), dependencies);
+    break;
+  case XTypes::EK_COMPLETE:
+  case XTypes::EK_MINIMAL:
+    {
+      TypeMap::const_iterator pos = type_map.find(type_identifier);
```
mitza-oci, line 719:
```suggestion
      const TypeMap::const_iterator pos = type_map.find(type_identifier);
```

dds/idl/typeobject_generator.cpp
```diff
@@ -6,499 +6,1740 @@
...
+std::ostream&
+operator<<(std::ostream& out, const Printer& printer)
+{
+  return printer.print_on(out);
+}
 
-    return;
-  }
+template <typename T>
+struct ValuePrinter : public Printer {
```
mitza-oci, line 94:
```suggestion
struct ValuePrinter : Printer {
```

dds/idl/typeobject_generator.cpp
```diff
@@ -6,499 +6,1740 @@
...
+  ValuePrinter(const T& a_value) : value(a_value) {}
```
mitza-oci, line 95: explicit? Not sure if you need implicit conversions here

dds/idl/typeobject_generator.cpp
mitza-oci: (applies to a lot of these)

dds/idl/typeobject_generator.cpp
mitza-oci: (applies to a lot of these)

dds/idl/typeobject_generator.cpp
```diff
@@ -6,499 +6,1740 @@
...
+struct AliasMemberFlagPrinter : public MemberFlagPrinter {
+  AliasMemberFlagPrinter(const OpenDDS::XTypes::AliasMemberFlag a_value) : MemberFlagPrinter(a_value) {}
+};
+
+struct TypeFlagPrinter : public ValuePrinter<OpenDDS::XTypes::TypeFlag> {
+  TypeFlagPrinter(const OpenDDS::XTypes::EquivalenceKind a_value) : ValuePrinter(a_value) {}
+
+  std::ostream& print_on(std::ostream& out) const
+  {
+    out << "0";
```
mitza-oci, line 195: This is a little kludgy, but I'm OK with it.  Another approach would be to keep track of when the first output happens and only add `|` after that.

mitza-oci (changes requested):

dds/idl/typeobject_generator.cpp
mitza-oci: New spec issue created: DDSXTY14-36, you can add a comment here with that identifier.

dds/idl/annotations.cpp
```diff
@@ -92,29 +91,29 @@ AST_Expression::AST_ExprValue* get_annotation_member_ev(
...
-    idl_global->err()->misc_error(
-      (std::string("Found null pointer while getting value of member \"") +
-        member_name + "\" of annotation \"" +
-        appl->local_name()->get_string() + "\"").c_str(),
-      appl);
-    BE_abort();
+    be_util::misc_error_and_abort(std::string("Found null pointer while getting value of member \"") +
+                                  member_name + "\" of annotation \"" +
+                                  appl->local_name()->get_string() + "\"",
+                                  appl);
```
iguessthislldo, line 116: I still don't understand why all the callers of `get_annotation_member_ev` have to do the same check before was just done in `get_annotation_member_ev`.

mitza-oci: @iguessthislldo I'm working on a version of this that passes an `AST_Expression::ExprType` to `get_annotation_member_ev()` so it can be checked.  Does that seem right to you?

mitza-oci (approved):

dds/idl/annotations.cpp
iguessthislldo: Yes, how it looks now is fine.

iguessthislldo (approved):

PR #1868, opened 2020-09-02
Support Padding Marker in Encapsulation Options

kuznetsovmoci:
XTypes 1.3 7.6.3.1.2 says we should set the 2 LSB of the second byte of the options to be the number of bytes required to make the serialized size of the payload align to 4.

### DIFF
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -450,8 +450,8 @@ private:
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
       char* wr = tmp_mb->wr_ptr();
-      EncapsulationHeader encap;
       if (encapsulated) {
+        EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
           return 0;
         }
@@ -471,7 +471,7 @@ private:
         return 0;
       }
       if (encapsulated) {
-        encap.set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
+        EncapsulationHeader::set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
       }
     } else { // OpenDDS::DCPS::FULL_MARSHALING
       ACE_NEW_MALLOC_RETURN(tmp_mb,
@@ -494,8 +494,8 @@ private:
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
       char* wr = tmp_mb->wr_ptr();
-      EncapsulationHeader encap;
       if (encapsulated) {
+        EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
           return 0;
         }
@@ -515,7 +515,7 @@ private:
         return 0;
       }
       if (encapsulated) {
-        encap.set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
+        EncapsulationHeader::set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
       }
     }
 
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -449,6 +449,7 @@ private:
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
+      char* wr = tmp_mb->wr_ptr();
       if (encapsulated) {
         EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
@@ -469,7 +470,11 @@ private:
           TraitsType::type_name()));
         return 0;
       }
-
+      if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
+        || encoding.kind() == Encoding::KIND_XCDR2)) {
+        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
+        wr[3] |= (padding & 0x03);
+      }
     } else { // OpenDDS::DCPS::FULL_MARSHALING
       ACE_NEW_MALLOC_RETURN(tmp_mb,
         static_cast<ACE_Message_Block*>(
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -417,6 +417,8 @@ public:
 
 private:
 
+  static const int PADDING_MARKER_BYTE_INDEX = 3;
+
   /**
    * Serialize the instance data.
    *
@@ -449,7 +451,7 @@ private:
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
-      char* wr = tmp_mb->wr_ptr();
+      char* const wr = mb->wr_ptr();
       if (encapsulated) {
         EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
@@ -471,7 +473,7 @@ private:
         return 0;
       }
       if (encapsulated) {
-        EncapsulationHeader::set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
+        EncapsulationHeader::set_padding_marker(wr[PADDING_MARKER_BYTE_INDEX], mb->wr_ptr() - wr);
       }
     } else { // OpenDDS::DCPS::FULL_MARSHALING
       ACE_NEW_MALLOC_RETURN(tmp_mb,
@@ -493,7 +495,7 @@ private:
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
-      char* wr = tmp_mb->wr_ptr();
+      char* const wr = mb->wr_ptr();
       if (encapsulated) {
         EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
@@ -515,7 +517,7 @@ private:
         return 0;
       }
       if (encapsulated) {
-        EncapsulationHeader::set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
+        EncapsulationHeader::set_padding_marker(wr[PADDING_MARKER_BYTE_INDEX], mb->wr_ptr() - wr);
       }
     }
 
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -450,8 +450,8 @@ private:
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
       char* wr = tmp_mb->wr_ptr();
+      EncapsulationHeader encap;
       if (encapsulated) {
-        EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
           return 0;
         }
@@ -470,10 +470,8 @@ private:
           TraitsType::type_name()));
         return 0;
       }
-      if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
-        || encoding.kind() == Encoding::KIND_XCDR2)) {
-        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
-        wr[serializer.swap_bytes() ? 0 : 3] |= (padding & 0x03);
+      if (encapsulated) {
+        encap.set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
       }
     } else { // OpenDDS::DCPS::FULL_MARSHALING
       ACE_NEW_MALLOC_RETURN(tmp_mb,
@@ -496,8 +494,8 @@ private:
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
       char* wr = tmp_mb->wr_ptr();
+      EncapsulationHeader encap;
       if (encapsulated) {
-        EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
           return 0;
         }
@@ -516,10 +514,8 @@ private:
           TraitsType::type_name()));
         return 0;
       }
-      if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
-          || encoding.kind() == Encoding::KIND_XCDR2)) {
-        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
-        wr[serializer.swap_bytes() ? 0 : 3] |= (padding & 0x03);
+      if (encapsulated) {
+        encap.set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
       }
     }
 
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -492,7 +492,6 @@ private:
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
-      char* const wr = mb->wr_ptr();
       if (encapsulated) {
         EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -473,7 +473,7 @@ private:
       if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
         || encoding.kind() == Encoding::KIND_XCDR2)) {
         unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
-        wr[3] |= (padding & 0x03);
+        wr[serializer.swap_bytes() ? 0 : 3] |= (padding & 0x03);
       }
     } else { // OpenDDS::DCPS::FULL_MARSHALING
       ACE_NEW_MALLOC_RETURN(tmp_mb,
@@ -519,7 +519,7 @@ private:
       if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
           || encoding.kind() == Encoding::KIND_XCDR2)) {
         unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
-        wr[3] |= (padding & 0x03);
+        wr[serializer.swap_bytes() ? 0 : 3] |= (padding & 0x03);
       }
     }
 
--- a/dds/DCPS/DataWriterImpl_T.h
+++ b/dds/DCPS/DataWriterImpl_T.h
@@ -417,8 +417,6 @@ public:
 
 private:
 
-  static const int PADDING_MARKER_BYTE_INDEX = 3;
-
   /**
    * Serialize the instance data.
    *
@@ -451,7 +449,6 @@ private:
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
-      char* const wr = mb->wr_ptr();
       if (encapsulated) {
         EncapsulationHeader encap;
         if (!encap.from_encoding(encoding, MarshalTraitsType::extensibility())) {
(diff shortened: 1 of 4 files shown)


iguessthislldo:
We will also need this for key only serialization.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -469,7 +470,11 @@ DataWriterImpl_T
           TraitsType::type_name()));
         return 0;
       }
-
+      if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
+        || encoding.kind() == Encoding::KIND_XCDR2)) {
```
iguessthislldo, line 474:
```suggestion
      if (encapsulated) {
```

The spec doesn't make this conditional on a specific encoding.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -469,7 +470,11 @@ DataWriterImpl_T
...
+        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
+        wr[serializer.swap_bytes() ? 0 : 3] |= (padding & 0x03);
```
iguessthislldo, line 476:
```suggestion
        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
        wr[3] |= (padding & 0x03);
```
The encapsulation header is a series of bytes, it's not effected by endianness.

Also please move this to logic a new function inside `EncapsulationHeader`.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -469,7 +470,11 @@ DataWriterImpl_T
           TraitsType::type_name()));
         return 0;
       }
-
+      if (encapsulated && (encoding.kind() == Encoding::KIND_XCDR1
+        || encoding.kind() == Encoding::KIND_XCDR2)) {
+        unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;
```
iguessthislldo, line 475: @mitza-oci Should we be worried about chains?

iguessthislldo (changes requested):

dds/DCPS/DataWriterImpl_T.h
kuznetsovmoci: Unless I am reading this wrong, encapsulation header serialization is byte order sensitive:

ACE_INLINE void
Serializer::write_array(const char* x, size_t size, ACE_CDR::ULong length)
{
  this->write_array(x, size, length, this->swap_bytes_);
}

ACE_INLINE void
Serializer::write_array(const char* x, size_t size,
                        ACE_CDR::ULong length, bool swap)
{
  if (!swap || size == 1) {
    //
    // No swap, copy direct.
    //
    this->buffer_write(x, size * length, false);

  } else {
    //
    // Swapping _must_ be done at 'size' boundaries, so we need to spin
    // through the array element by element.
    // NOTE: This assumes that there is _no_ padding between the array
    //       elements.  If this is not the case, do not use this
    //       method.
    //
    while (length-- > 0) {
      this->buffer_write(x, size, true);
      x += size;
    }
  }
}

dds/DCPS/DataWriterImpl_T.h
kuznetsovmoci: This is what it says:
"The RTPS specification does not define any settings for the 2 byte options field and further states
that a receiver should not interpret it when it reads the options field. This DDS-XTYPES
specification changes this defining the use of some bits in the options field"
So I assumed it is 'XTYPES' specific. Is this incorrect interpretation?

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: That code is for an array of `unsigned long`. The encapsulation header uses an `octet` array:

https://github.com/objectcomputing/OpenDDS/blob/045bb54025689ce8b2aeb1017968f1a73708fbf1/dds/DCPS/Serializer.cpp#L244-L249

Which doesn't rely on endianness:

https://github.com/objectcomputing/OpenDDS/blob/045bb54025689ce8b2aeb1017968f1a73708fbf1/dds/DCPS/Serializer.inl#L671-L676

dds/DCPS/DataWriterImpl_T.h
kuznetsovmoci: Just to clarify: the third 'kind' is:
    /**
     * This is the classic encoding of OpenDDS used when there is no RTPS
     * transport being used. It has no padding bytes and no XCDR behavior.
     */
    KIND_UNALIGNED_CDR,

So it looked to me these rules would not be applicable to it, but I might be wrong.

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: Sorry no, it actually looks like the code you posted is for generic arrays, and encap header is using it. Size of  `byte_cdr_size` is one, so it's skipping doing anything about swapping because it doesn't apply.

kuznetsovmoci: OK, that makes sense

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: > So I assumed it is 'XTYPES' specific. Is this incorrect interpretation?

It doesn't say any where it's just limited to the encoding defined by XTypes, it just says the XTypes spec is changing the definition of the options field.

> So it looked to me these rules would not be applicable to it, but I might be wrong.

They do not, but unaligned cdr isn't encapsulated and since the change is for all encapsulated encodings, it doesn't make sense to add any more conditions unto it.

dds/DCPS/DataWriterImpl_T.h
kuznetsovmoci: Do you prefer this line be within the EncupsulationHeader member function or handled by the caller?
`unsigned char padding = (tmp_mb->wr_ptr() - wr) % 4;`

iguessthislldo: Yes, at least the padding part of it. Also I just realized this expression is wrong. It is the number of bytes the value is past the alignment, not the number of bytes to the next or current alignment. Please use: `align(value, 4) - value`. That will get the padding.

kuznetsovmoci: Do I get the `value` as 
`size_t value = serialized_size(encoding, instance_data);`
or this also would be past the alignment?

iguessthislldo: That depends on how we want to do it (see the discussion in the ticket), but `serialized_size` + the header size theoretically should be the same as `mb->wr_ptr() - wr`. Serialized size and the stream position I talked about are more "proper" than pointer arithmetic, but this is simpler.

kuznetsovmoci: So neither `mb->wr_ptr() - wr` nor `serialized_size + the header size` gives us what we need, as both numbers include padding  - correct?

iguessthislldo: Yes, but that is the whole point of this, is to mark how my bytes would be needed to bring the sample into 4 byte alignment.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -449,8 +449,9 @@ DataWriterImpl_T
       mb.reset(tmp_mb);
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
+      char* wr = tmp_mb->wr_ptr();
```
mitza-oci, line 452:
```suggestion
      char* const wr = mb->wr_ptr();
```

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -469,7 +470,9 @@ DataWriterImpl_T
           TraitsType::type_name()));
         return 0;
       }
-
+      if (encapsulated) {
+        encap.set_padding_marker(wr[3], tmp_mb->wr_ptr() - wr);
```
mitza-oci, line 474:
```suggestion
        encap.set_padding_marker(wr[SOME_NAMED_CONSTANT], mb->wr_ptr() - wr);
```

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -450,8 +450,8 @@ DataWriterImpl_T
 
       OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
       char* wr = tmp_mb->wr_ptr();
+      EncapsulationHeader encap;
       if (encapsulated) {
-        EncapsulationHeader encap;
```
iguessthislldo, line 453: `set_padding_marker` can be made static so you don't have to do this.

dds/DCPS/Serializer.cpp
```diff
@@ -196,6 +196,13 @@ bool EncapsulationHeader::to_encoding(
   return true;
 }
 
+void EncapsulationHeader::set_padding_marker(char& options, size_t size)
+{
+  size_t aligned = size;
+  align(aligned, 4);
+  options |= ((aligned - size) & 0x03);
```
mitza-oci, line 203: I don't think we need to call `align()` here, since the running total is never needed.  How about `CONSTANT_FOR_4 - size % CONSTANT_FOR_4`?

mitza-oci (changes requested):

dds/DCPS/Serializer.cpp
```diff
@@ -196,6 +196,13 @@ bool EncapsulationHeader::to_encoding(
   return true;
 }
 
+static const int FOUR_BYTE_ALIGNMENT = 4;
```
mitza-oci, line 199: Could this be one of the existing constants in Serializer.h?  Maybe `ALIGN_XCDR2`?

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -417,6 +417,8 @@ DataWriterImpl_T
 
 private:
 
+  static const int PADDING_MARKER_BYTE_INDEX = 3;
```
mitza-oci, line 420: Move to Serializer?

This could keep more encoding details out of DataWriterImpl_T.  How about passing the message block itself to `EncapsulationHeader::set_encapsulation_options`?  The `rd_ptr` of the message block already points to the start of the block.  The `length` of the message block is number of bytes actually used.  That is enough to determine the options.

mitza-oci (changes requested):

dds/DCPS/Serializer.cpp
iguessthislldo: The padding marker is not specific to any encoding, so I don't think we should use `ALIGN_XCDR2`.

dds/DCPS/DataWriterImpl_T.h
iguessthislldo: Yes I agree. Technically we should be doing this everything a encapsulation header is being used, so that would make it easier to set the padding marker in other places if we wished.

dds/DCPS/Serializer.cpp
iguessthislldo: If it's going to be a constant outside a function, it should probably go in `EcapsulationHeader`.

dds/DCPS/Serializer.cpp
```diff
@@ -196,6 +196,11 @@ bool EncapsulationHeader::to_encoding(
   return true;
 }
 
+void EncapsulationHeader::set_encapsulation_options(Message_Block_Ptr& mb)
```
iguessthislldo, line 199: How about `set_padding_marker`, Since that's all it's doing right now.

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/Serializer.h
```diff
@@ -243,6 +244,8 @@ class OpenDDS_Dcps_Export EncapsulationHeader {
   };
 
   const static size_t serialized_size = 4;
+  const static size_t padding_marker_byte_index = 3;
+  const static int padding_marker_alignmet = 4;
```
mitza-oci, line 248:
```suggestion
  const static int padding_marker_alignment = 4;
```

dds/DCPS/Serializer.cpp
```diff
@@ -196,6 +196,11 @@ bool EncapsulationHeader::to_encoding(
...
+{
+  mb->rd_ptr()[padding_marker_byte_index] |= ((padding_marker_alignmet - mb->length() % padding_marker_alignmet) & 0x03);
```
mitza-oci, line 201: Consider adding a check to make sure `mb` has sufficient length.

Add a test of this function to `tests/DCPS/Serializer`

mitza-oci (changes requested):

dds/DCPS/Serializer.cpp
kuznetsovmoci: What should we use as a test criteria in the test function?

mitza-oci: I think that would be based on the requirements in the spec.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
...
+  const OpenDDS::DCPS::EncapsulationHeader encap;
+  OpenDDS::DCPS::unique_ptr<OpenDDS::DCPS::MessageBlockAllocator> mb_allocator_;
+
+  for (unsigned int i = 1; i <= OpenDDS::DCPS::EncapsulationHeader::padding_marker_alignment; i++) {
+    OpenDDS::DCPS::Message_Block_Ptr mb;
+    ACE_Message_Block* tmp_mb;
+
+    ACE_NEW_RETURN(tmp_mb,
+      ACE_Message_Block(4 + i, ACE_Message_Block::MB_DATA),
+      0);
```
iguessthislldo, line 712:
```suggestion
      false);
```

dds/DCPS/Serializer.h
```diff
@@ -243,6 +244,8 @@ class OpenDDS_Dcps_Export EncapsulationHeader {
   };
 
   const static size_t serialized_size = 4;
+  const static size_t padding_marker_byte_index = 3;
+  const static int padding_marker_alignment = 4;
```
iguessthislldo, line 248:
```suggestion
  const static size_t padding_marker_alignment = 4;
```

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
...
+    mb.reset(tmp_mb);
+
+    OpenDDS::DCPS::Serializer serializer(mb.get(), encodings[5]);
+    serializer << encap;
+    serializer.write_octet_array(arr, i);
```
iguessthislldo, line 717: Check the return value of these functions. They can be the same check.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
...
+  for (unsigned int i = 1; i <= OpenDDS::DCPS::EncapsulationHeader::padding_marker_alignment; i++) {
+    OpenDDS::DCPS::Message_Block_Ptr mb;
+    ACE_Message_Block* tmp_mb;
+
+    ACE_NEW_RETURN(tmp_mb,
+      ACE_Message_Block(4 + i, ACE_Message_Block::MB_DATA),
+      0);
+    mb.reset(tmp_mb);
+
+    OpenDDS::DCPS::Serializer serializer(mb.get(), encodings[5]);
```
iguessthislldo, line 715: I suggest creating a local encoding and also passing it to `EncapsulationHeader` to be consistent.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
...
+    serializer << encap;
+    serializer.write_octet_array(arr, i);
+
+    if (!OpenDDS::DCPS::EncapsulationHeader::set_encapsulation_options(mb)) {
+      std::cerr << "EncapsulationHeader::set_encapsulation_options failed" << std::endl << "size: " << mb->length();
```
iguessthislldo, line 720: It would be helpful for the scoreboard if this was one line, also you're missing a new line at the end.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
...
+      return false;
+    }
+
+    unsigned int padding_marker_alignment = mb->rd_ptr()[OpenDDS::DCPS::EncapsulationHeader::padding_marker_byte_index] & 0x03;
+    if (padding_marker_alignment != (OpenDDS::DCPS::EncapsulationHeader::padding_marker_alignment - i)) {
+      std::cerr << "EncapsulationHeader::set_encapsulation_options failed for "
+        << i << " bytes, padding marker alignment: " << padding_marker_alignment << std::endl;
```
iguessthislldo, line 727: Return false? Maybe set a status if you want to check the whole loop.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +696,41 @@ bool runOverrunTest()
   return true;
 }
 
+bool runEncapsulationOptionsTest()
+{
+  std::cerr << "\nRunning encapsulation options tests...\n";
+  const CORBA::Octet arr[4] = {0};
+  const OpenDDS::DCPS::EncapsulationHeader encap;
+  OpenDDS::DCPS::unique_ptr<OpenDDS::DCPS::MessageBlockAllocator> mb_allocator_;
```
iguessthislldo, line 704: ```suggestion```

iguessthislldo (changes requested):

tests/DCPS/Serializer/SerializerTest.cpp
iguessthislldo: I don't know why it showed up like this, but this isn't being used and the suggestion is to remove the line.

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +698,51 @@ bool runOverrunTest()
...
+    mb.reset(tmp_mb);
+
+    OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
+    if (!(serializer << encap) || !serializer.write_octet_array(arr, i)) {
+      std::cerr << "Serialization failed in runEncapsulationOptionsTest" << std::endl;
+        return false;
+    }
+
+    if (!OpenDDS::DCPS::EncapsulationHeader::set_encapsulation_options(mb)) {
+      std::cerr << "EncapsulationHeader::set_encapsulation_options failed. " << "Size: " << mb->length() << std::endl;
```
iguessthislldo, line 731:
```suggestion
      std::cerr << "EncapsulationHeader::set_encapsulation_options failed. Size: " << mb->length() << std::endl;
```

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -695,6 +698,51 @@ bool runOverrunTest()
...
+
+    ACE_NEW_RETURN(tmp_mb,
+      ACE_Message_Block(4 + i, ACE_Message_Block::MB_DATA),
+      false);
+    mb.reset(tmp_mb);
+
+    OpenDDS::DCPS::Serializer serializer(mb.get(), encoding);
+    if (!(serializer << encap) || !serializer.write_octet_array(arr, i)) {
+      std::cerr << "Serialization failed in runEncapsulationOptionsTest" << std::endl;
+        return false;
```
iguessthislldo, line 727:
```suggestion
      return false;
```

tests/DCPS/Serializer/SerializerTest.cpp
```diff
@@ -1,4 +1,6 @@
 #include "common.h"
+#include "dds/DCPS/dcps_export.h"
+#include "dds/DCPS/DCPS_Utils.h"
```
iguessthislldo, line 3:
```suggestion

#include <dds/DCPS/DCPS_Utils.h>
```

Why did you include the export header?

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -469,7 +469,9 @@ DataWriterImpl_T
           TraitsType::type_name()));
         return 0;
       }
-
+      if (encapsulated) {
+        EncapsulationHeader::set_encapsulation_options(mb);
```
mitza-oci, line 473: check return value

mitza-oci (approved):

PR #1870, opened 2020-09-02
Make participant message datawriters and readers non-durable

jrw972:


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -445,17 +445,17 @@ Sedp::init(const RepoId& guid,
 #endif
 
   if (spdp_.available_builtin_endpoints() & DISC_BUILTIN_ENDPOINT_PARTICIPANT_ANNOUNCER) {
-    participant_message_writer_->enable_transport_using_config(reliable, durable, transport_cfg_);
```
simpsont-oci, line 448: Doesn't this violate 8.4.13.3 of the RTPS spec?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -445,17 +445,17 @@ Sedp::init(const RepoId& guid,
...
+    participant_message_writer_->enable_transport_using_config(reliable, nondurable, transport_cfg_);
   }
-  participant_message_reader_->enable_transport_using_config(reliable, durable, transport_cfg_);
+  participant_message_reader_->enable_transport_using_config(reliable, nondurable, transport_cfg_);
 
 #ifdef OPENDDS_SECURITY
   participant_message_secure_writer_->set_crypto_handles(spdp_.crypto_handle());
   participant_message_secure_reader_->set_crypto_handles(spdp_.crypto_handle());
   if (spdp_.available_builtin_endpoints() & DDS::Security::BUILTIN_PARTICIPANT_MESSAGE_SECURE_WRITER) {
-    participant_message_secure_writer_->enable_transport_using_config(reliable, durable, transport_cfg_);
```
simpsont-oci, line 456: And this seems to violate 7.4.4.1 of the Security spec (see last bullet in particular, stating that it's the reason the volatile readers / writers were necessary).

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4292,29 +4270,6 @@ Sedp::write_subscription_data_secure(
 }
 #endif
 
-DDS::ReturnCode_t
```
simpsont-oci, line 4295: We're never going to write a participant message? Don't we use these for liveliness?

jrw972: Liveliness uses a different method.  I didn't see this method being called anywhere but the CI might tell me otherwise.

dds/DCPS/RTPS/Sedp.cpp
jrw972: I'll defer to @mitza-oci

dds/DCPS/RTPS/Sedp.cpp
jrw972: I'll defer to @mitza-oci

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: Currently the implementation of this built-in endpoint in SEDP is not Durable, but we are enabling Durability in the transport.

The part of the spec you reference does say it should be Durable.  A peer process wouldn't be able to tell that we're not enabling Durability here, it would just appear that there are no data samples in the cache for a late-joining reader.

We could also consider implementing Durability, but it seems to have limited value for the BuiltinParticipantMessageWriter.

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: 7.4.4.1 is non-normative.  It provides background on why DCPSParticipantVolatileMessageSecure exists in the spec.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -445,17 +445,17 @@ Sedp::init(const RepoId& guid,
 #endif
 
   if (spdp_.available_builtin_endpoints() & DISC_BUILTIN_ENDPOINT_PARTICIPANT_ANNOUNCER) {
```
mitza-oci, line 447:
```suggestion
  if (spdp_.available_builtin_endpoints() & BUILTIN_ENDPOINT_PARTICIPANT_MESSAGE_DATA_WRITER) {
```
Need to review use of these constants.

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: In particular we found that nothing adds elements to the `local_participant_messages_` collection.

However, after more discussions I think the current plan is to keep it durable.

PR #1877, opened 2020-09-08
Fix Coverity issues.

jonesc-oci:


dds/DCPS/Service_Participant.cpp
```diff
@@ -1249,7 +1249,17 @@ Service_Participant::get_discovery(const DDS::DomainId_t domain)
       cf.open();
       ACE_Configuration_Section_Key k;
       cf.open_section(cf.root_section(), RTPS_SECTION_NAME, 1 /*create*/, k);
-      this->load_discovery_configuration(cf, RTPS_SECTION_NAME);
+
+      int status = this->load_discovery_configuration(cf, RTPS_SECTION_NAME);
```
simpsont-oci, line 1253:
```suggestion
      int status = load_discovery_configuration(cf, RTPS_SECTION_NAME);
```

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -157,7 +157,8 @@ TransportRegistry::load_transport_configuration(const OPENDDS_STRING& file_name,
             // Create the TransportInst object and load the transport
             // configuration in ACE_Configuration_Heap to the TransportInst
             // object.
-            TransportInst_rch inst = create_inst(ACE_TEXT_ALWAYS_CHAR(transport_id.c_str()), transport_type);
+            const OPENDDS_STRING tid_str = ACE_TEXT_ALWAYS_CHAR(transport_id.c_str());
```
simpsont-oci, line 160: I think Coverity will still complain about this, since the result of c_str() is still directly passed into the constructor and/or operator=() of basic_string, either of which will dereference it. I think it just wants to see you checking for null before using it?

Perhaps something like:
`const OPENDDS_STRING tid_str = transport_id.c_str() ? ACE_TEXT_ALWAYS_CHAR(transport_id.c_str()) : OPENDDS_STRING();`

simpsont-oci (changes requested):

dds/DCPS/transport/framework/TransportRegistry.cpp
simpsont-oci: Or an if statement after declaring it, if that feels cleaner to you. I realize having OPENDDS_STRING() in there looks hokey.

iguessthislldo: How about `const OPENDDS_STRING tid_str = transport_id.c_str() ? ACE_TEXT_ALWAYS_CHAR(transport_id.c_str()) : "";`?

simpsont-oci (approved):

PR #1879, opened 2020-09-10
User cannot dynamically toggle RtpsRelay and ICE

jrw972:
The decision to use the RtpsRelay (or to only use the RtpsRelay) or
ICE can only be made at start time.  Users of OpenDDS want the ability
to dynamically turn off the RtpsRelay and ICE for testing purposes.

Solution: Make the corresponding configuration dynamic.  The
configuration for the RtspRelay is straightforward as it can be picked
up on the next send.  The configuration of ICE is more difficult since
ICE needs to be restarted for participants that have already been
discovered.

The code attempts to reset flags in the ParticpantLocationTopic.  The
ParticipantLocationTopic in a remote participant will be updated after
the corresponding entries time out.  To a certain extent, these
controls influence the ways in which other participants can discover a
given participant.  For example, switching to RtpsRelay only means
that a remote participant will eventually say that the given
participant can only be reached via the RtpsRelay.  The local
participant may still be able to discover the remote locally.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
@@ -52,9 +52,25 @@ ICE::Endpoint*
 RtpsUdpTransport::get_ice_endpoint()
 {
 #ifdef OPENDDS_SECURITY
-  return (config().use_ice_) ? &ice_endpoint_ : 0;
+  return (config().use_ice()) ? &ice_endpoint_ : 0;
 #else
   return 0;
+#endif
+}
+
+void
+RtpsUdpTransport::use_ice_now(bool after)
+{
+  // Update the config.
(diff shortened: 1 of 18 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -52,12 +52,28 @@ ICE::Endpoint*
...
+  return (config().use_ice()) ? &ice_endpoint_ : 0;
 #else
   return 0;
 #endif
 }
 
+void
+RtpsUdpTransport::use_ice_now(bool after)
+{
+  // Update the config.
```
mitza-oci, line 64:
```suggestion
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #1880, opened 2020-09-10
XTYPE-124: TypeConsistencyEnforcement QoS Policy (initial implementation)

adamsj-oci:
- Determine values for the enum “kind” and the 5 booleans in TypeConsistencyEnforcementQosPolicy (Annex D)
- These would be constants for our initial implementation, the user will not be able to change them
- Send these values in the SEDP Built-in Subscription endpoint

### DIFF
--- a/dds/DCPS/RTPS/ParameterListConverter.cpp
+++ b/dds/DCPS/RTPS/ParameterListConverter.cpp
@@ -246,12 +246,12 @@ namespace {
   {
     DDS::TypeConsistencyEnforcementQosPolicy def_tceqp =
         TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
-    return ((tceqp.kind != def_tceqp.kind) ||
-      (tceqp.ignore_sequence_bounds != def_tceqp.ignore_sequence_bounds) ||
-      (tceqp.ignore_string_bounds != def_tceqp.ignore_string_bounds) ||
-      (tceqp.ignore_member_names != def_tceqp.ignore_member_names) ||
-      (tceqp.prevent_type_widening != def_tceqp.prevent_type_widening) ||
-      (tceqp.force_type_validation != def_tceqp.force_type_validation));
+    return (tceqp.kind != def_tceqp.kind ||
+      tceqp.ignore_sequence_bounds != def_tceqp.ignore_sequence_bounds ||
+      tceqp.ignore_string_bounds != def_tceqp.ignore_string_bounds ||
+      tceqp.ignore_member_names != def_tceqp.ignore_member_names ||
+      tceqp.prevent_type_widening != def_tceqp.prevent_type_widening ||
+      tceqp.force_type_validation != def_tceqp.force_type_validation);
   }
 
   void normalize(DDS::Duration_t& dur)
--- a/dds/DCPS/RTPS/ParameterListConverter.cpp
+++ b/dds/DCPS/RTPS/ParameterListConverter.cpp
@@ -1253,8 +1253,6 @@ bool to_param_list(const DCPS::DiscoveredReaderData& reader_data,
     add_param(param_list, param);
   }
 
-
-
   CORBA::ULong i;
   CORBA::ULong locator_len = reader_data.readerProxy.allLocators.length();
   // Serialize from allLocators, rather than the unicastLocatorList
--- a/dds/DCPS/RTPS/ParameterListConverter.cpp
+++ b/dds/DCPS/RTPS/ParameterListConverter.cpp
@@ -242,6 +242,23 @@ namespace {
     return qos.value.length();
   }
 
+  bool not_default(const DDS::TypeConsistencyEnforcementQosPolicy& tceqp)
+  {
+    DDS::TypeConsistencyEnforcementQosPolicy def_tceqp =
+        TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
+    if ((tceqp.kind != def_tceqp.kind) ||
+      (tceqp.ignore_sequence_bounds != def_tceqp.ignore_sequence_bounds) ||
+      (tceqp.ignore_string_bounds != def_tceqp.ignore_string_bounds) ||
+      (tceqp.ignore_member_names != def_tceqp.ignore_member_names) ||
+      (tceqp.prevent_type_widening != def_tceqp.prevent_type_widening) ||
+      (tceqp.force_type_validation != def_tceqp.force_type_validation)) {
+      return true;
+    } else {
+      return false;
+    }
+
+  }
+
   void normalize(DDS::Duration_t& dur)
   {
     // Interoperability note:
@@ -1238,6 +1255,13 @@ bool to_param_list(const DCPS::DiscoveredReaderData& reader_data,
   {
     Parameter param;
     param.representation(reader_data.ddsSubscriptionData.representation);
+    add_param(param_list, param);
+  }
+
+  if (not_default(reader_data.ddsSubscriptionData.type_consistency))
+  {
+    Parameter param;
+    param.type_consistency(reader_data.ddsSubscriptionData.type_consistency);
     add_param(param_list, param);
   }
 
--- a/dds/DCPS/RTPS/ParameterListConverter.cpp
+++ b/dds/DCPS/RTPS/ParameterListConverter.cpp
@@ -246,17 +246,12 @@ namespace {
   {
     DDS::TypeConsistencyEnforcementQosPolicy def_tceqp =
         TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
-    if ((tceqp.kind != def_tceqp.kind) ||
+    return ((tceqp.kind != def_tceqp.kind) ||
       (tceqp.ignore_sequence_bounds != def_tceqp.ignore_sequence_bounds) ||
       (tceqp.ignore_string_bounds != def_tceqp.ignore_string_bounds) ||
       (tceqp.ignore_member_names != def_tceqp.ignore_member_names) ||
       (tceqp.prevent_type_widening != def_tceqp.prevent_type_widening) ||
-      (tceqp.force_type_validation != def_tceqp.force_type_validation)) {
-      return true;
-    } else {
-      return false;
-    }
-
+      (tceqp.force_type_validation != def_tceqp.force_type_validation));
   }
 
   void normalize(DDS::Duration_t& dur)
--- a/dds/DCPS/RTPS/ParameterListConverter.cpp
+++ b/dds/DCPS/RTPS/ParameterListConverter.cpp
@@ -1253,12 +1253,7 @@ bool to_param_list(const DCPS::DiscoveredReaderData& reader_data,
     add_param(param_list, param);
   }
 
-  if (not_default(reader_data.ddsSubscriptionData.type_consistency))
-  {
-    Parameter param;
-    param.type_consistency(reader_data.ddsSubscriptionData.type_consistency);
-    add_param(param_list, param);
-  }
+
 
   CORBA::ULong i;
   CORBA::ULong locator_len = reader_data.readerProxy.allLocators.length();
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -4447,7 +4447,6 @@ Sedp::write_subscription_data_unsecure(
       result = DDS::RETCODE_ERROR;
     }
 
-    //add TypeConsistencyEnforcementQosPolicy to param list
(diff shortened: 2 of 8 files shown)


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4079,6 +4079,12 @@ Sedp::populate_discovered_reader_msg(
     drd.readerProxy.associatedWriters.length(len + 1);
     drd.readerProxy.associatedWriters[len] = *writer;
   }
+  drd.ddsSubscriptionData.type_consistency.kind = DDS::ALLOW_TYPE_COERCION;
+  drd.ddsSubscriptionData.type_consistency.ignore_sequence_bounds = true;
+  drd.ddsSubscriptionData.type_consistency.ignore_string_bounds = true;
+  drd.ddsSubscriptionData.type_consistency.ignore_member_names = false;
+  drd.ddsSubscriptionData.type_consistency.prevent_type_widening = false;
+  drd.ddsSubscriptionData.type_consistency.force_type_validation = false;
```
adamsj-oci, line 4087: @ClaytonCalabrese and @iguessthislldo, These values are the defaults from the spec. Are they also the correct fixed/const values we want to use for now?

mitza-oci:
This error is buried in the Travis CI log:
```
Would remove java/dds/DDS/ALLOW_TYPE_COERCION.java
Would remove java/dds/DDS/DISALLOW_TYPE_COERCION.java
Would remove java/dds/DDS/TypeConsistencyEnforcementQosPolicy.java
Would remove java/dds/DDS/TypeConsistencyEnforcementQosPolicyHelper.java
Would remove java/dds/DDS/TypeConsistencyEnforcementQosPolicyHolder.java
```
These are based on Java code generation from IDL.  You can add them to relevant `.gitignore`, or we could decide that this IDL should not be in the user API and just put it in `RtpsCore.idl` with other RTPS Discovery details.

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -242,6 +242,23 @@ namespace {
...
+        TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
+    if ((tceqp.kind != def_tceqp.kind) ||
+      (tceqp.ignore_sequence_bounds != def_tceqp.ignore_sequence_bounds) ||
+      (tceqp.ignore_string_bounds != def_tceqp.ignore_string_bounds) ||
+      (tceqp.ignore_member_names != def_tceqp.ignore_member_names) ||
+      (tceqp.prevent_type_widening != def_tceqp.prevent_type_widening) ||
+      (tceqp.force_type_validation != def_tceqp.force_type_validation)) {
+      return true;
+    } else {
+      return false;
```
mitza-oci, line 257:
```
if (EXPR) {
  return true;
} else {
  return false;
}
```
is simply `return EXPR;`

dds/DdsDcpsCore.idl
```diff
@@ -272,22 +272,27 @@ module DDS {
       DISALLOW_TYPE_COERCION,
       ALLOW_TYPE_COERCION
     };
-
+    */
+    //constants for TypeConsistencyEnforcementQosPolicy::kind field
+    const short DISALLOW_TYPE_COERCION = 1;
+    const short ALLOW_TYPE_COERCION = 2;
```
mitza-oci, line 278: Use a typedef for `short` (both here and in the struct) to make it clear what this is.

dds/DdsDcpsCore.idl
```diff
@@ -471,9 +474,7 @@ module DDS {
       @id(0x002E) TopicDataQosPolicy topic_data;
       @id(0x002D) GroupDataQosPolicy group_data;
       @id(0x0073) DataRepresentationQosPolicy representation;
-      /* TODO
       @id(0x0074) TypeConsistencyEnforcementQosPolicy type_consistency;
-      */
```
mitza-oci, line 476: Looks like this may need to get populated in SEDP?  Do we need the ParameterListConverter to be able to translate from the Parameter to the structure?

mitza-oci (changes requested):

dds/DdsDcpsCore.idl
adamsj-oci: > Looks like this may need to get populated in SEDP?

I also had to add it to `RtpsCore.idl` if that's what you mean for SEDP. The way I did it, I need the conversion for parameter list in `ParameterListConverter.cpp`.

mitza-oci: I'm asking how these structures get their values "populated."

ParameterListConverter has two directions of conversion:
- "to ParameterList" is for sending discovery info about endpoints in the current participant
- "from ParameterList" is for receiving discovery info about endpoints in other participants

When "this" Sedp receives an announcement on the network and transforms it "from ParameterList", where does the resulting type_consistency info end up?

One possible answer is just that we don't support this in the current iteration.  If so, it shouldn't appear in the user API.

adamsj-oci: I haven't handled the "from" ParameterList. It was a question that came up when Mike and I discussed. There was some confusion based on the mention that the policy only applied to the DataReader. It should be easy to do the reverse to a point. The question I'll have to figure out is how far I can take it until we need more supporting code to make it useful.

mitza-oci: Is the "from ParameterList" more clear based on the comment above?  If not we can discuss outside of this PR review.

dds/DCPS/RTPS/Sedp.cpp
ClaytonCalabrese: Ignore member names is screwed up in the spec and I wrote an issue about that in our doc for spec problems.  The others make sense to me.

adamsj-oci: Are you saying that the default for `ignore_member_names` should be `true`? What about the constant value we we wish to use for `ignore_member_names` for our initial setup?

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -242,6 +242,18 @@ namespace {
     return qos.value.length();
   }
 
+  bool not_default(const DDS::TypeConsistencyEnforcementQosPolicy& tceqp)
+  {
+    DDS::TypeConsistencyEnforcementQosPolicy def_tceqp =
+        TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
+    return ((tceqp.kind != def_tceqp.kind) ||
```
mitza-oci, line 249:
```suggestion
    return tceqp.kind != def_tceqp.kind ||
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4446,6 +4446,16 @@ Sedp::write_subscription_data_unsecure(
                  ACE_TEXT("to ParameterList\n")));
       result = DDS::RETCODE_ERROR;
     }
+
+    //add TypeConsistencyEnforcementQosPolicy to param list
```
mitza-oci, line 4450:
```suggestion
```

dds/DdsDcpsCore.idl
```diff
@@ -272,22 +272,27 @@ module DDS {
       DISALLOW_TYPE_COERCION,
       ALLOW_TYPE_COERCION
     };
-
+    */
+    //constants for TypeConsistencyEnforcementQosPolicy::kind field
```
mitza-oci, line 276:
```suggestion
```

mitza-oci (changes requested):

dds/DdsDcpsCore.idl
```diff
@@ -272,22 +272,26 @@ module DDS {
       DISALLOW_TYPE_COERCION,
       ALLOW_TYPE_COERCION
     };
-
+    */
+    typedef short TypeConsistencyEnforcementQosPolicyKind_t;
+    const TypeConsistencyEnforcementQosPolicyKind_t DISALLOW_TYPE_COERCION = 1;
+    const TypeConsistencyEnforcementQosPolicyKind_t ALLOW_TYPE_COERCION = 2;
+    /*
```
mitza-oci, line 279: As it stands here, we are leaving the `POLICY_ID` and `POLICY_NAME` constants (below) commented-out.  Should they be included now?  Even if weren't not setting these policies directly, the constants could be useful when integrating with Discovery.

mitza-oci (approved):

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1241,6 +1253,8 @@ bool to_param_list(const DCPS::DiscoveredReaderData& reader_data,
     add_param(param_list, param);
   }
 
+
+
```
ClaytonCalabrese, line 1257: delete these two blank lines.

dds/DdsDcpsCore.idl
adamsj-oci: I'm still waiting on the tests in this PR to rerun. If something happens there that I have to rework, I think it would be OK to uncomment assuming it wouldn't cause tests to fail or some other large issue (which it shouldn't). If the tests pass, then I'd prefer to do it in my next issue so we can close this issue out.

dds/DCPS/Service_Participant.cpp
```diff
@@ -787,6 +787,14 @@ Service_Participant::initialize()
   initial_SubscriberQos_.partition = initial_PartitionQosPolicy_;
   initial_SubscriberQos_.group_data = initial_GroupDataQosPolicy_;
   initial_SubscriberQos_.entity_factory = initial_EntityFactoryQosPolicy_;
+
+  initial_TypeConsistencyEnforcementQosPolicy_.kind = DDS::ALLOW_TYPE_COERCION;
+  initial_TypeConsistencyEnforcementQosPolicy_.prevent_type_widening = false;
+  initial_TypeConsistencyEnforcementQosPolicy_.ignore_sequence_bounds = true;
+  initial_TypeConsistencyEnforcementQosPolicy_.ignore_string_bounds = true;
+  initial_TypeConsistencyEnforcementQosPolicy_.ignore_member_names = false;
```
ClaytonCalabrese, line 795: ignore_member_names should be treated like its name suggests.  In our case, we do want to ignore the names of members for assignability so we should set it to true.

The spec has an issue regarding this setting so we should treat it as the name suggests rather than how the spec suggests until we get further clarification from OMG.

dds/DdsDcpsCore.idl
mitza-oci: I agree we don't need to do it in this PR.  I marked it "approved."

dds/DCPS/Service_Participant.cpp
adamsj-oci: @ClaytonCalabrese These are the defaults from the spec. I don't think we need to change them. It's meant to be used to check in the event we don't want to send the parameter.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4446,6 +4446,15 @@ Sedp::write_subscription_data_unsecure(
                  ACE_TEXT("to ParameterList\n")));
       result = DDS::RETCODE_ERROR;
     }
+
+    DDS::TypeConsistencyEnforcementQosPolicy tceqp = TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
+    tceqp.ignore_member_names = true;
```
sonndinh, line 4451: The behavior of type assignability will need to change if `ignore_member_names` is set to true because it currently doesn't ignore member names.

adamsj-oci: @sonndinh @ClaytonCalabrese I understand your point. There is the option to interpret the spec with an intended meaning or we can take it literally. I don't really care as long as we're consistent but I think we need to make record of the resolution of this issue. @mitza-oci Any suggestions?

dds/DCPS/Service_Participant.cpp
ClaytonCalabrese: Son is correct I am wrong. Both the default and what we send should be false.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4446,6 +4446,14 @@ Sedp::write_subscription_data_unsecure(
                  ACE_TEXT("to ParameterList\n")));
       result = DDS::RETCODE_ERROR;
     }
+
+    DDS::TypeConsistencyEnforcementQosPolicy tceqp = TheServiceParticipant->initial_TypeConsistencyEnforcementQosPolicy();
```
adamsj-oci, line 4450: @sonndinh @ClaytonCalabrese @mitza-oci Sending `ignore_member_names` set to `false`.

PR #1884, opened 2020-09-14
Support RTPS large data for durability resends (from #1853)

mitza-oci:


tests/DCPS/DelayedDurable/rtps_disc.ini
```diff
@@ -12,3 +12,6 @@ ResendPeriod=2
 [transport/the_rtps_transport]
 transport_type=rtps_udp
 use_multicast=0
+heartbeat_period=500
```
jrw972, line 15: Does the test work with the default settings?

jrw972 (approved):

tests/DCPS/DelayedDurable/rtps_disc.ini
mitza-oci: It did pass, but I'd like to keep the adjusted times in there.

PR #1885, opened 2020-09-14
Fix coverity warning

jrw972:


dds/DCPS/RTPS/Sedp.h
```diff
@@ -209,7 +209,7 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t>, public DCPS::RcObj
 
   class MsgParticipantData : public MsgBase {
   public:
-    MsgParticipantData(DCPS::WeakRcHandle<Sedp> sedp,
+    MsgParticipantData(const DCPS::WeakRcHandle<Sedp>& sedp,
                        DCPS::MessageId id,
                        const ParticipantData_t data)
```
mitza-oci, line 214: This is the line that coverity didn't like

jrw972: Right.  But it seems like it will start complaining about all of the others so I might as well fix them.

mitza-oci: OK, but we need to update line 214 ("this is the line"... I guess GitHub doesn't make it clear) before merging this.

mitza-oci (approved):

PR #1886, opened 2020-09-14
XTypes: Enum bit_bound assignability rule

sonndinh:


dds/DCPS/XTypes/TypeAssignability.cpp
```diff
@@ -1067,6 +1067,15 @@ bool TypeAssignability::assignable_enum(const MinimalTypeObject& ta,
     return false;
   }
 
+  // T1.bit_bound and T2.bit_bound are in the same equivalence set
```
mitza-oci, line 1070:
```suggestion
  // T1.bit_bound and T2.bit_bound are in the same equivalence set (DDSXTY14-34)
```
provide tracking to this spec issue, since this is not yet actually part of the spec

mitza-oci (approved):

dds/DCPS/XTypes/TypeAssignability.cpp
```diff
@@ -1067,6 +1067,15 @@ bool TypeAssignability::assignable_enum(const MinimalTypeObject& ta,
     return false;
   }
 
+  // T1.bit_bound and T2.bit_bound are in the same equivalence set (DDSXTY14-34)
+  const BitBound ta_bit_bound = ta.enumerated_type.header.common.bit_bound;
+  const BitBound tb_bit_bound = tb.enumerated_type.header.common.bit_bound;
+  if (((ta_bit_bound >= 1 && ta_bit_bound <= 8) && !(tb_bit_bound >= 1 && tb_bit_bound <= 8)) ||
+      ((ta_bit_bound >= 9 && ta_bit_bound <= 16) && !(tb_bit_bound >= 9 && tb_bit_bound <= 16)) ||
+      ((ta_bit_bound >= 17 && ta_bit_bound <= 32) && !(tb_bit_bound >= 17 && tb_bit_bound <= 32))) {
```
iguessthislldo, line 1075: Since it's were I took this idea from, can this check be common code with bitmask?

sonndinh: I think no. For bitmask, it is a check between TK_UINTxx with bit_bound, so its code is different.

iguessthislldo: After looking at the bitmask code I agree. It's similar, but how its gets the range to check is different.

iguessthislldo (approved):

PR #1888, opened 2020-09-15
default data representation XCDR2

lij-oci:
The default DataRepresentationQosPolicy's value is an empty sequence. The meaning of such empty sequence is implementation-defined. In OpenDDS, it is defined as DDS::XCDR2_DATA_REPRESENTATION for DataWriter, and as DDS::XCDR2_DATA_REPRESENTATION and DDS::XCDR_DATA_REPRESENTATION for DataReader.


dds/DCPS/DCPS_Utils.cpp
```diff
@@ -401,13 +401,14 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
 }
 
-DDS::DataRepresentationIdSeq get_effective_data_rep_qos(DDS::DataRepresentationIdSeq qos) {
-  DDS::DataRepresentationIdSeq qos_ids = qos;
-  if (qos_ids.length() == 0) {
-    qos_ids.length(1);
-    qos_ids[0] = DDS::XCDR_DATA_REPRESENTATION;
+DDS::DataRepresentationIdSeq get_effective_data_rep_qos(const DDS::DataRepresentationIdSeq& qos) {
+  if (qos.length() == 0) {
+    DDS::DataRepresentationIdSeq ids = qos;
```
mitza-oci, line 406:
```suggestion
    DDS::DataRepresentationIdSeq ids(1);
```

mitza-oci (changes requested):

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -401,13 +401,28 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
+      Encoding::Kind kind;
+      if (repr_to_encoding_kind(qos[i], kind) && Encoding::KIND_XCDR1 == kind) {
+        // erase XCDR1
+        const CORBA::ULong e = qos.length() - 1;
+        DDS::DataRepresentationIdSeq ids(e);
+        ids.length(e);
+        for (CORBA::ULong j = 0; j < e; ++j) {
+          ids[j] = qos[j < i ? j : (j + 1)];
+        }
+        return ids;
```
mitza-oci, line 421: Is it possible that `qos` contains more than one `XCDR1`?  I guess this wouldn't make a lot of sense but it could happen.

If we want this to be a general purpose "filter" step, it could just accumulate results into a result sequence.

iguessthislldo:
This PR assumes XCDR1 should be avoided. We can't avoid XCDR1, because it is required for compatibility with non-XTypes DDS systems.

mitza-oci (changes requested):

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1147,11 +1147,10 @@ DataReaderImpl::get_matched_publication_data(
...
 DataReaderImpl::enable()
 {
-  //According spec:
-  // - Calling enable on an already enabled Entity returns OK and has no
-  // effect.
+  // According to spec:
+  // - Calling enable on an already enabled Entity has no effect and returns OK.
   // - Calling enable on an Entity whose factory is not enabled will fail
-  // and return PRECONDITION_NOT_MET.
+  //   and return PRECONDITION_NOT_MET.
```
mitza-oci, line 1153: Why are these lines changed in this PR?  Doesn't seem related to this task.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1221,10 +1221,8 @@ DataReaderImpl::enable()
         " Cached_Allocator_With_Overflow %x with %d chunks\n",
         rd_allocator_.get(), n_chunks_));
 
-  if ((qos_.liveliness.lease_duration.sec !=
-      DDS::DURATION_INFINITE_SEC) &&
-      (qos_.liveliness.lease_duration.nanosec !=
-          DDS::DURATION_INFINITE_NSEC)) {
+  if ((qos_.liveliness.lease_duration.sec != DDS::DURATION_INFINITE_SEC) &&
+      (qos_.liveliness.lease_duration.nanosec != DDS::DURATION_INFINITE_NSEC)) {
```
mitza-oci, line 1225: Why are these lines changed in this PR? Doesn't seem related to this task.

mitza-oci (changes requested):
Does discovery need to change to complete this?

mitza-oci:
> The default DataRepresentationQosPolicy's value is an empty sequence. The meaning of such empty sequence is implementation-defined. In OpenDDS, it is defined as DDS::XCDR2_DATA_REPRESENTATION.

Please update this PR's description.

dds/DCPS/DataReaderImpl.cpp
lij-oci: Just a little typo and format correction for better understanding of the code.

dds/DCPS/DataReaderImpl.cpp
lij-oci: for easier to read when I was combing through the code for potential relevant changes.

mitza-oci: I don't think this needs to change here.  We should try to minimize churn so that someone looking at `git blame` can see when the code actually changes.

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -747,6 +742,15 @@ bool from_param_list(const ParameterList& param_list,
 
 // OpenDDS::DCPS::DiscoveredWriterData
 
+void add_DataRepresentationQos(ParameterList& param_list, const DDS::DataRepresentationIdSeq& ids, bool reader)
+{
+  DDS::DataRepresentationQosPolicy dr_qos;
+  dr_qos.value = DCPS::get_effective_data_rep_qos(ids, reader);
+  Parameter param;
+  param.representation(dr_qos);
+  add_param(param_list, param);
```
mitza-oci, line 751: As with most other parameters, if the value is the "network message default" then the parameter should be omitted.  

In this case we have the added complexity that the "network message default" is different from the default QoS (empty sequence) and also different from the interpretation of the empty sequence.

But that shouldn't prevent this from using the common pattern of "if not (network message) default then add the parameter."

mitza-oci (changes requested):

dds/DCPS/RTPS/ParameterListConverter.cpp
iguessthislldo: So you're saying we should omit the parameter if we would otherwise be sending just XCDR1?

mitza-oci: Yeah, I guess that's a shorter way of putting it.  Hopefully the context above is helpful.

iguessthislldo: Yes, I have no problems with that.

lij-oci: @iguessthislldo @mitza-oci  I'm sure what I need to do on this.

tests/DCPS/DataRepresentation/DataRepresentation.cpp
```diff
@@ -22,594 +22,441 @@
...
-    }
-    TheServiceParticipant->shutdown();
-  } catch (const CORBA::Exception& e) {
-    e._tao_print_exception("Caught in main()");
-    exit_status = 1;
-  }
-
-  return exit_status;
+    Test test(argc, argv);
+    ret = test.run();
```
iguessthislldo, line 459:
```suggestion
    ret = test.run() ? 1 : 0;
```

It's not generally a good idea to return counts of any sort from main.

iguessthislldo (changes requested):

tests/DCPS/DataRepresentation/DataRepresentation.cpp
```diff
@@ -22,594 +22,441 @@
...
         rc = reader->get_subscription_matched_status(reader_match);
         if (rc != DDS::RETCODE_OK) {
-          ACE_ERROR((LM_ERROR, ACE_TEXT("%N:%l check_match() ERROR: ")
-            ACE_TEXT("get_subscription_matched_status failed: %C\n"),
+          ACE_ERROR((LM_ERROR, ACE_TEXT("%N:%l check_match() ERROR: get_subscription_matched_status failed: %C\n"),
             retcode_to_string(rc)));
-          return check_match_result_error;
+          return check_match_error;
         }
-
```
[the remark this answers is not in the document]
lij-oci: forgive me please :)

mitza-oci (approved):
Once tests are passing and @iguessthislldo has reviewed, I'm good

iguessthislldo (changes requested):

PR #1892, opened 2020-09-18
XTypes: Fix Segfault Caused By Misplaced #endif in Sedp

iguessthislldo:
I didn't have security enabled in this build so these rc handles were invalid.


jrw972 (approved):
This can be merged immediately.

PR #1893, opened 2020-09-18
Fix warnings and compile errors on older MSVC

mitza-oci:


dds/DCPS/transport/framework/TransportQueueElement.cpp
```diff
@@ -20,7 +20,7 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
-const TqePair null_tqe_pair(0, 0);
+const TqePair null_tqe_pair;
```
iguessthislldo, line 23: Why was this an warning or error and is it guaranteed the object will initialize with null values?

mitza-oci: > 
> 
> Why was this an warning or error and

It's an error because there's a bug in that version of MSVC

> is it guaranteed the object will initialize with null values?

Absolutely.  It has static storage duration.

jrw972 (approved):

PR #1895, opened 2020-09-19
NAT bindings for RtpsRelay may timeout without return data

jrw972:
A client was observed sending SEDP to the relay and the relay was
observed sending SEDP to the client.  The client's messages were
received by the relay but the relay's messages were not received by
the client.  Simultaneously, there was no issue with the SPDP and STUN
messages to the SEDP port.  The key difference between the SEDP port
and SPDP/STUN is that the latter has a periodic return of data.  Thus,
the motivation for this change is to provide a mechanism for the
periodic return of data from the relay to a client.

Solution
--------

Each port on the relay acts as a STUN server.  Clients of the relay
periodically send binding requests to refresh any NAT bindings.
Client send binding indications between requests.  The timing of the
requests is configured through the ICE setting for server reflexive
addresses.

The STUN messages to the RtpsRelay were extended with the GUID prefix
of the client.  This was necessary to setup and maintain the routing
tables on the relay.

A new builtin topic (Connection Record) was created to monitor the
connection with the relay.

Future Work
-----------

The Connection Record topic should replace the Participant Location
Topic.

When the RtpsRelay and ICE are used simultaneously, the server
reflexive task in ICE should be disabled and the results of the relay
STUN task used instead.

The next possible step in this design is to implement STUN via RTPS.
One would most likely take the route of STUN readers and writers and
secure versions of the same.  One challenge is that the ports in the
relay are not associated with a participant.  One option is for the
relay to impersonate the application participant for the purpose of
reading and writing STUN.


dds/DCPS/transport/framework/DataLink.h
```diff
@@ -410,9 +410,11 @@ class OpenDDS_Dcps_Export DataLink
   typedef OPENDDS_MAP_CMP(RepoId, LocalAssociationInfo, GUID_tKeyLessThan) AssocByLocal;
   AssocByLocal assoc_by_local_;
 
+ protected:
```
mitza-oci, line 413: Is this needed?  Could `impl()` be used?

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.h
```diff
@@ -73,6 +75,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpReceiveStrategy
                                       const ACE_SOCK_Dgram& socket,
                                       ACE_INET_Addr& remote_address,
                                       ICE::Endpoint* endpoint,
+                                      RtpsUdpTransport* tport,
```
mitza-oci, line 78: Is this needed?  Doesn't the Data Link already have it?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2530,122 +2526,128 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
...
+      ICE::Endpoint* endpoint = get_ice_endpoint();
+      if (endpoint) {
+        ICE::Agent::instance()->receive(endpoint, local, remote, message);
       }
-    } else if (!submessageLength) {
-      break; // submessageLength of 0 indicates the last submessage
     }
   }
+#endif
+# endif
```
mitza-oci, line 2650: Indent preprocessor lines consistently (or we could just give up and left-align all of them)

mitza-oci (changes requested):
Take a look at logging (within core libs) to see if some of these need to be conditional.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.h
jrw972: The helper is a free function.

dds/DCPS/RTPS/ICE/Stun.h
```diff
@@ -83,6 +87,7 @@ struct OpenDDS_Rtps_Export Attribute {
     ACE_UINT32 fingerprint; // FINGERPRINT
     ACE_UINT32 priority; // PRIORITY
     ACE_UINT64 ice_tie_breaker; // ICE_CONTROLLED, ICE_CONTROLLING
+    unsigned char guid_prefix[12]; // GUID_PREFIX
```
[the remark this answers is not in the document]
mitza-oci: This line wasn't updated by the latest commit

dds/DCPS/RTPS/ICE/Ice.h
```diff
@@ -225,6 +226,206 @@ class OpenDDS_Rtps_Export Agent {
...
+    SRSM_Change
+  };
+
+  ServerReflexiveStateMachine()
+    : indication_count_(0)
+  {}
+
+  // Return Unset if transitioning from a determined SRA to an undetermined SRA.
+  // Return None otherwise.
+  StateChange send(const ACE_INET_Addr& address, size_t indication_count_limit, const DCPS::GuidPrefix_t& guid_prefix)
```
mitza-oci, line 244: A few of these methods seem to be too large to be defined inline (in the class definition).

dds/DCPS/RTPS/ICE/Ice.cpp
```diff
@@ -104,11 +104,191 @@ Candidate make_peer_reflexive_candidate(const ACE_INET_Addr& address, ACE_UINT32
...
+    return SRSM_None;
+  }
+
+  ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) ServerReflexiveStateMachine::receive: WARNING Unknown STUN message class %d\n"), message.class_));
+  return SRSM_None;
+}
+
+ServerReflexiveStateMachine::StateChange
+ServerReflexiveStateMachine::start(const ACE_INET_Addr& address,
+                                     const DCPS::GuidPrefix_t& guid_prefix)
```
mitza-oci, line 165:
```suggestion
                                   const DCPS::GuidPrefix_t& guid_prefix)
```

dds/DCPS/RTPS/ICE/Ice.cpp
```diff
@@ -104,11 +104,191 @@ Candidate make_peer_reflexive_candidate(const ACE_INET_Addr& address, ACE_UINT32
...
+    return SRSM_None;
+  }
+
+
+  if (!message.has_error_code()) {
+    ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) ServerReflexiveStateMachine::error_response: WARNING No error code\n")));
+    return SRSM_None;
+  }
+
+  ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) ServerReflexiveStateMachine::error_response: WARNING STUN error response code=%d reason=%s\n"), message.get_error_code(), message.get_error_reason().c_str()));
```
mitza-oci, line 279: * split very long line
* %s vs. %C

mitza-oci (changes requested):

mitza-oci (approved):

PR #1899, opened 2020-09-22
XTypes: TypeLookup getTypeDependencies

sonndinh:
Implement getTypeDependencies operation for TypeLookup service.


dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -69,7 +69,15 @@ void TypeLookupService::get_type_dependencies(const TypeIdentifierSeq& type_ids,
...
   for (; iter != tmp.end(); ++i, ++iter) {
-    dependencies[i] = *iter;
+    // TODO(sonndinh): do we include TypeIdentifiers that don't have TypeObjects?
+    // Currently, we include only the ones that have TypeObjects.
+    const TypeObjectMap::const_iterator tobj_it = type_object_map_.find(*iter);
+    if (tobj_it != type_object_map_.end()) {
+      dependencies[i].type_id = *iter;
+      size_t size = 0;
+      DCPS::serialized_size(get_typeobject_encoding(), size, tobj_it->second);
+      dependencies[i].typeobject_serialized_size = size;
```
iguessthislldo, line 79:
```suggestion
      dependencies[i].typeobject_serialized_size =
        DCPS::serialized_size(get_typeobject_encoding(), tobj_it->second);
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3671,25 +3682,92 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+      }
+    }
+
+    DCPS::RepoId remote_guid = sample.header_.publication_id_;
+    if (!sedp_.send_type_lookup_request(req_type_ids, remote_guid, XTypes::TypeLookup_getTypes_HashId)) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Sedp::TypeLookupReplyReader::process_get_dependencies_reply - ")
+                 ACE_TEXT("failed to send type lookup request\n")));
+      return DDS::RETCODE_ERROR;
+    }
+    sedp_.type_lookup_reply_deadline_processor_->schedule(sedp_.max_type_lookup_service_reply_period_);
```
kuznetsovmoci, line 3765: `sedp_.type_lookup_reply_deadline_processor_->schedule(sedp_.max_type_lookup_service_reply_period_);`
This may not be needed, as this part of code does not create additional entries in the `MatchingDataMap` to be removed if the reply times out.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3671,25 +3682,92 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+  default:
+    return DDS::RETCODE_UNSUPPORTED;
+  }
+
+  if (DDS::RETCODE_OK == retcode) {
     DCPS::SequenceNumber rpc_sequence;
     rpc_sequence.setValue(type_lookup_reply.header.related_request_id.sequence_number.high,
-      type_lookup_reply.header.related_request_id.sequence_number.low);
+                          type_lookup_reply.header.related_request_id.sequence_number.low);
+    // TODO(sonndinh): Is this a correct place to call match_continue?
```
kuznetsovmoci, line 3728: We can't call match_continue() until we receive both Type Objects and ALL Dependencies (which becomes even more interesting, as getting dependencies may result in subsequent request for the Type Objects). Somehow we should keep track (counter? - as a suggestion) if any Type Lookup Service requests are outstanding and proceed with match_continue(), only if the counter is 0.

kuznetsovmoci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3560,24 +3562,33 @@ Sedp::TypeLookupReplyReader::~TypeLookupReplyReader()
...
-
   type_lookup_request.header.request_id.sequence_number.high = rpc_sequence.getHigh();
   type_lookup_request.header.request_id.sequence_number.low = rpc_sequence.getLow();
 
   // As per chapter 7.6.3.3.4 of XTypes spec
   const OPENDDS_STRING instance_name = OPENDDS_STRING("dds.builtin.TOS.") + DCPS::to_string(participant_id);
   type_lookup_request.header.instance_name = instance_name.c_str();
+  type_lookup_request.data.kind = tl_kind;
+
+  if (tl_kind == XTypes::TypeLookup_getTypes_HashId) {
```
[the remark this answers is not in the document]
ClaytonCalabrese: I think you will have repeated code either way due to repeating the typelookuprequest code for both tl_kinds.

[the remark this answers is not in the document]
sonndinh: Using `switch` is fine, but the checking at the beginning allows it to return right away for unsupported `tl_kind`s without going through the initialization for `type_lookup_request`. If it uses `switch`, it would be better to move the `switch` block to the beginning, before the initialization of `type_lookup_request`. But even with that, I don't think there is a clear advantage of using `switch`.

dds/DCPS/RTPS/Sedp.cpp
sonndinh: This is handled now.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1243,14 +1229,61 @@ namespace OpenDDS {
...
       void
       remove_expired_endpoints(const MonotonicTimePoint& /*now*/)
       {
         ACE_GUARD(ACE_Thread_Mutex, g, lock_);
+        MonotonicTimePoint now = MonotonicTimePoint::now();
         MatchingDataIter end_iter = matching_data_buffer_.end();
 
         for (MatchingDataIter iter = matching_data_buffer_.begin(); iter != end_iter; ) {
-          if (MonotonicTimePoint::now() - iter->second.time_added_to_map >= max_type_lookup_service_reply_period_) {
+          if (now - iter->second.time_added_to_map >= max_type_lookup_service_reply_period_) {
```
kuznetsovmoci, line 1286: Do we need to add cleanup for `orig_seq_numbers_`?

kuznetsovmoci (changes requested):

ClaytonCalabrese (changes requested):

dds/DCPS/DiscoveryBase.h
```diff
@@ -1243,14 +1229,61 @@ namespace OpenDDS {
...
+
+          // Get the dependent TypeIdentifiers of the topic type
+          send_type_lookup_request(type_ids, remote_id, is_discovery_protected, false);
+        } else {
+          type_ids.length(type_info->minimal.dependent_typeid_count + 1);
+          type_ids[0] = type_info->minimal.typeid_with_size.type_id;
+          for (size_t i = 1; i <= (size_t)type_info->minimal.dependent_typeid_count; ++i) {
+            type_ids[i] = type_info->minimal.dependent_typeids[i].type_id;
+          }
+          has_all_dependencies_ = true;
```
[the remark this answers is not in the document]
sonndinh: Good catch. After the discussion today, this flag isn't needed anymore. The internal data for getTypeDependencies operations use collections now, with each entry contains data for a pair of endpoints.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,23 +3823,164 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+      type_lookup_reply._cxx_return.kind == XTypes::TypeLookup_getTypes_HashId &&
+      sedp_.has_all_dependencies_) {
+    DCPS::SequenceNumber seq_num;
+    seq_num.setValue(type_lookup_reply.header.related_request_id.sequence_number.high,
+                     type_lookup_reply.header.related_request_id.sequence_number.low);
+    const OrigSeqNumberMap::const_iterator it = sedp_.orig_seq_numbers_.find(seq_num);
+    if (it != sedp_.orig_seq_numbers_.end()) {
+      // Cleanup data
+      const XTypes::TypeIdentifier remote_ti = sedp_.orig_seq_numbers_[seq_num].first;
+      OrigSeqNumberMap::iterator it = sedp_.orig_seq_numbers_.begin();
```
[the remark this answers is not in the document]
sonndinh: This code is removed now.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,23 +3823,166 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+    break;
+  case XTypes::TypeLookup_getDependencies_HashId:
+    retcode = process_get_dependencies_reply(sample, type_lookup_reply, is_discovery_protected);
+    break;
+  default:
+    return DDS::RETCODE_UNSUPPORTED;
+  }
+
+  if (DDS::RETCODE_OK == retcode &&
+      type_lookup_reply._cxx_return.kind == XTypes::TypeLookup_getTypes_HashId &&
```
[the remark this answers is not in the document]
sonndinh: As we discussed today, implementing it in two phases (get all dependencies first and then get all type objects) eliminates those out-of-order cases.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,23 +3823,166 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+  // Store an entry for either the final getTypes or next getTypeDependencies request
+  sedp_.orig_seq_numbers_.insert(std::make_pair(++sedp_.type_lookup_service_sequence_number_,
+                                                it->second));
+
+  if (data.continuation_point.length() == 0) {
+    // All dependencies have been received
+    sedp_.has_all_dependencies_ = true;
+
+    // Get the type objects of all dependencies
+    if (!sedp_.send_type_lookup_request(deps, remote_id, is_discovery_protected, true)) {
```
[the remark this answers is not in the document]
sonndinh: In the updated code, `deps` now has at least the TypeIdentifier of topic type.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,23 +3823,166 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+    // Optimization - only store TypeIdentifiers for which TypeObjects haven't
+    // already been in the type objects cache yet
+    if (!sedp_.type_lookup_service_->type_object_in_cache(ti)) {
+      deps.append(ti);
+    }
+  }
+  dependencies_[guid_pref][remote_ti].first = data.continuation_point;
+
+  // Store an entry for either the final getTypes or next getTypeDependencies request
+  sedp_.orig_seq_numbers_.insert(std::make_pair(++sedp_.type_lookup_service_sequence_number_,
```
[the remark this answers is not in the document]
sonndinh: This is updated now.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -760,24 +765,65 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
...
+      {
+        return std::memcmp(&prefix_[0], &other.prefix_[0], sizeof(GuidPrefix_t)) < 0;
+      }
+
+      GuidPrefix_t prefix_;
+    };
+
+    // Map from each remote participant to the data stored for its types.
+    typedef OPENDDS_MAP(GuidPrefixWrapper, RemoteDependencies) DependenciesMap;
+    DependenciesMap dependencies_;
```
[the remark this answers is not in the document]
sonndinh: Yes, it is cleaned up at the end of `process_type_lookup_reply` in the success case. I'll still need to add the cleanup for when replies expire though.

dds/DCPS/DiscoveryBase.h
sonndinh: Yes, it is handled. I'll also add the cleanup for it when replies expire.

dds/DCPS/TypeSupportImpl.cpp
```diff
@@ -70,6 +70,27 @@ void TypeSupportImpl::to_type_info(XTypes::TypeInformation& type_info) const
...
+  XTypes::compute_dependencies(type_map, type_id, dependencies);
+
+  OPENDDS_SET(XTypes::TypeIdentifier)::const_iterator it = dependencies.begin();
+  for (; it != dependencies.end(); ++it) {
+    XTypes::TypeMap::const_iterator iter = type_map.find(*it);
+    if (iter != type_map.end()) {
+      const size_t tobj_size = serialized_size(XTypes::get_typeobject_encoding(), iter->second);
+      deps_with_size.append({*it, static_cast<ACE_CDR::ULong>(tobj_size)});
+    } else {
+      ACE_DEBUG((LM_WARNING,
```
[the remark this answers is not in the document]
sonndinh: That works for me.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,26 +3820,196 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+
 DDS::ReturnCode_t
-Sedp::TypeLookupReplyReader::process_tl_reply(DCPS::Serializer& ser)
+Sedp::TypeLookupRequestReader::process_get_dependencies_request(const XTypes::TypeLookup_Request& request,
+  XTypes::TypeLookup_Reply& reply)
+{
+  // Send all dependencies of the requested types
+  sedp_.type_lookup_service_->get_type_dependencies(request.data.getTypeDependencies.type_ids,
+    reply._cxx_return.getTypeDependencies.result.dependent_typeids);
+  if (reply._cxx_return.getTypeDependencies.result.dependent_typeids.length() > 0) {
```
mitza-oci, line 3837: Are there cases where length == 0 is expected?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,26 +3820,196 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+    break;
+  case XTypes::TypeLookup_getDependencies_HashId:
+    retcode = process_get_dependencies_reply(sample, type_lookup_reply, seq_num, is_discovery_protected);
+    break;
+  default:
+    return DDS::RETCODE_UNSUPPORTED;
+  }
+
+  if (XTypes::TypeLookup_getTypes_HashId == type_lookup_reply._cxx_return.kind) {
+    DCPS::SequenceNumber key_seq_num = seq_num_it->second.seq_number;
```
mitza-oci, line 3916:
```suggestion
    const DCPS::SequenceNumber key_seq_num = seq_num_it->second.seq_number;
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,26 +3820,196 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+
   return DDS::RETCODE_OK;
 }
 
+void Sedp::cleanup_type_lookup_data(const XTypes::TypeIdentifier& ti)
+{
+#ifdef OPENDDS_SECURITY
+  type_lookup_reply_secure_reader_->cleanup(ti);
+#else
+  type_lookup_reply_reader_->cleanup(ti);
```
mitza-oci, line 4009: `type_lookup_reply_reader_` can also be used in a security-enabled build

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3693,25 +3696,38 @@ Sedp::TypeLookupReplyReader::~TypeLookupReplyReader()
...
+  XTypes::TypeLookup_Request type_lookup_request;
   type_lookup_request.header.request_id.writer_guid = get_repo_id();
-
   type_lookup_request.header.request_id.sequence_number.high = rpc_sequence.getHigh();
   type_lookup_request.header.request_id.sequence_number.low = rpc_sequence.getLow();
 
-  // As per chapter 7.6.3.3.4 of XTypes spec
-  const OPENDDS_STRING instance_name = OPENDDS_STRING("dds.builtin.TOS.") + DCPS::to_string(participant_id);
+  // As per chapter 7.6.3.3.4 of XTypes spec.
+  const OPENDDS_STRING instance_name = OPENDDS_STRING("dds.builtin.TOS.") +
```
mitza-oci, line 3717: Is there already code for checking the `instance_name` of an incoming request?

mitza-oci (changes requested):

dds/DCPS/unique_ptr.h
```diff
@@ -130,6 +130,11 @@ class unique_ptr {
     std::swap(ptr_, b.ptr_);
   }
 
+  bool operator<(const unique_ptr& other) const
+  {
+    return *ptr_ < *other.ptr_;
```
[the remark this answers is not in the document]
sonndinh: I think comparing pointer values is not what we want. So `operator<` in `TypeIdentifier` relies on a number of `operator<` for member types, for example `PlainSequenceSElemDefn::operator<`, which calls `External<TypeIdentifier>::operator<` which finally calls `unique_ptr<TypeIdentifier>::operator<`. Since we're using `TypeIdentifier` as keys in some maps, if `unique_ptr<TypeIdentifier>::operator<` returns the comparison of pointer values, we may not be able to find an entry with a particular `TypeIdentifier` object value even though it's there in the map. So I think comparing using the actual contents of the objects is what we want in the case of `TypeIdentifier`.

The weakness is that all types that use `unique_ptr` now must provide `operator<`. That was a reason I thought about modifying the `operator<` operators of member types of `TypeIdentifier` directly so that they don't rely on `unique_ptr::operator<`.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1243,19 +1220,71 @@ namespace OpenDDS {
         }
       }
 
+      void save_matching_data_and_get_typeobjects(const XTypes::TypeInformation* type_info,
+                                                  MatchingData& md, const MatchingPair mp,
```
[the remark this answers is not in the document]
sonndinh: Yes, that works.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1243,19 +1220,71 @@ namespace OpenDDS {
...
+            type_info->minimal.dependent_typeids.length() < (CORBA::ULong)type_info->minimal.dependent_typeid_count) {
+          type_ids.append(type_info->minimal.typeid_with_size.type_id);
+
+          // Get dependencies of topic type
+          send_type_lookup_request(type_ids, remote_id, is_discovery_protected, false);
+        } else {
+          type_ids.length(type_info->minimal.dependent_typeid_count + 1);
+          type_ids[0] = type_info->minimal.typeid_with_size.type_id;
+          for (size_t i = 1; i <= (size_t)type_info->minimal.dependent_typeid_count; ++i) {
+            type_ids[i] = type_info->minimal.dependent_typeids[i].type_id;
```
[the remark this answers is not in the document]
sonndinh: Thanks!

dds/DCPS/DiscoveryBase.h
```diff
@@ -1612,6 +1641,31 @@ namespace OpenDDS {
       DDS::BuiltinTopicKey_t pub_bit_key_, sub_bit_key_;
       XTypes::TypeLookupService_rch type_lookup_service_;
 
+      struct GuidPrefixWrapper {
```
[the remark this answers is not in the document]
sonndinh: I think there is no advantage. I wanted to use `GuidPrefix_t` as key so that it's more clear that each entry in the map is for a remote participant. But using GUID works too. Do you want to change it to use GUID as key?

dds/DCPS/RTPS/Sedp.cpp
sonndinh: It's possible. There may be a request for TypeIdentifier that has no dependencies. I'll fix this.

dds/DCPS/unique_ptr.h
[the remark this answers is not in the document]
mitza-oci: It needs to match the standard library definition. If we want the caller to compare the pointed-to objects, then that code should change.

dds/DCPS/RTPS/Sedp.cpp
sonndinh: No, we haven't done anything with `instance_name` of incoming request. The spec doesn't say how it is handled as far as I know.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -760,27 +765,48 @@ class Sedp : public DCPS::EndpointManager<ParticipantData_t> {
...
+    typedef std::pair<XTypes::OctetSeq32, XTypes::TypeIdentifierSeq> ContinuationPair;
 
-  TypeLookupRequestReader_rch type_lookup_request_reader_;
+    // Map from each remote type to the most recent continuation_point
+    // and all dependencies received for that type so far.
+    typedef OPENDDS_MAP(XTypes::TypeIdentifier, ContinuationPair) RemoteDependencies;
+
+    // Map from each remote participant to the data stored for its types.
+    typedef OPENDDS_MAP(GuidPrefixWrapper, RemoteDependencies) DependenciesMap;
+    DependenciesMap dependencies_;
```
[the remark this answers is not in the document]
sonndinh: Do you mean the parent `Reader` class? If so, I think no since this map is specific to type lookup service (more specifically to getTypeDependencies of type lookup service).

dds/DCPS/unique_ptr.h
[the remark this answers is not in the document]
sonndinh: So should I change the operators `operator<` in `TypeIdentifier`?

dds/DCPS/RTPS/Sedp.h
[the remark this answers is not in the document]
kuznetsovmoci: How about TypeLookupService class?

dds/DCPS/DiscoveryBase.h
```diff
@@ -1243,25 +1220,75 @@ namespace OpenDDS {
...
+      {
+        md.rpc_sequence_number = ++type_lookup_service_sequence_number_;
+        MatchingDataIter md_it = matching_data_buffer_.find(mp);
+        if (md_it != matching_data_buffer_.end()) {
+          md_it->second = md;
+        } else {
+          matching_data_buffer_.insert(std::make_pair(mp, md));
+        }
+        // Store an entry for the first request
+        const TypeIdOrigSeqNumber orig_req_data = {type_info->minimal.typeid_with_size.type_id,
```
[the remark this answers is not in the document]
kuznetsovmoci: OK, that should work.

dds/DCPS/RTPS/Sedp.h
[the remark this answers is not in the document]
sonndinh: Note (from standup): we'll keep it in this class for now.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3803,26 +3817,184 @@ Sedp::TypeLookupRequestReader::process_get_types_request(const XTypes::TypeLooku
...
+    }
+  }
+
+  return DDS::RETCODE_OK;
+}
+
+void Sedp::cleanup_type_lookup_data(const DCPS::GuidPrefix_t& guid_prefix,
+                                    const XTypes::TypeIdentifier& ti)
+{
+#ifdef OPENDDS_SECURITY
```
kuznetsovmoci, line 3989: Would this work?

#ifdef OPENDDS_SECURITY
  type_lookup_reply_secure_reader_->cleanup(guid_prefix, ti);
#endif
  type_lookup_reply_reader_->cleanup(guid_prefix, ti);

You may also need to check if secure discovery is enabled. Is caller in a better position to know which reader to use (and pass it as as flag to this function)? If not - an easier option might be to move dependencies_ map to the type lookup service class :-). That will make cleanup unambiguous.

kuznetsovmoci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
sonndinh: For this function, calling `cleanup()` on both secure and non-secure readers works when security is enabled because there will be at most 1 reader that has the targeted data and the other reader will do nothing for cleanup. It just doesn't follow the pattern used by other functions:
```if (remote_entity is secure)
       type_lookup_reply_secure_reader_->cleanup(guid_prefix, ti);
     else if (remote_entity is non-secure)
       type_lookup_reply_reader_->cleanup(guid_prefix, ti);```

Comment (kuznetsovmoci) [reply dds/DCPS/RTPS/Sedp.cpp:3989]: Technically, it would work, but calling two functions, because we don't know which object should do the job is not the "cleanest" approach.

Comment (sonndinh) [reply dds/DCPS/RTPS/Sedp.cpp:3990]: Alright. It's calling the correct object now.

File: dds/DCPS/RTPS/Sedp.cpp (hunk 15 above)
Comment (sonndinh) [reply dds/DCPS/RTPS/Sedp.cpp]: This has been updated.

File: dds/DCPS/RTPS/Sedp.cpp [hunk 24]
```diff
@@ -3787,13 +3779,18 @@ Sedp::TypeLookupRequestReader::process_type_lookup_request(DCPS::Serializer& ser
...
   if (!(ser >> type_lookup_request)) {
     ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Sedp::TypeLookupRequestReader::process_type_lookup_request - ")
                ACE_TEXT("failed to deserialize type lookup request\n")));
     return DDS::RETCODE_ERROR;
   }
 
+  if (OPENDDS_STRING(type_lookup_request.header.instance_name) != instance_name_) {
+    ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::TypeLookupRequestReader::process_type_lookup_request - ")
+               ACE_TEXT("received request with wrong instance name\n")));
+    return DDS::RETCODE_PRECONDITION_NOT_MET;
```
mitza-oci, line 3791: Isn't this an expected case?  If so, remove the log message and return "OK."

kuznetsovmoci (approved):

dds/DCPS/TypeSupportImpl.h
```diff
@@ -25,6 +25,10 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 
 namespace OpenDDS {
+namespace XTypes{
+class TypeLookupService;
```
mitza-oci, line 29:
```suggestion
namespace XTypes {
  class TypeLookupService;
```

dds/DCPS/TypeSupportImpl.h
```diff
@@ -98,9 +102,8 @@ class OpenDDS_Dcps_Export TypeSupportImpl
 
   void to_type_info(XTypes::TypeInformation& type_info) const;
 
-  void get_dependencies(const XTypes::TypeMap& type_map,
-                        const XTypes::TypeIdentifier& type_id,
-                        XTypes::TypeIdentifierWithSizeSeq& deps_with_size) const;
+  //  void populate_dependencies(const XTypes::TypeLookupService_rch& tls) const;
+  void populate_dependencies(const RcHandle<XTypes::TypeLookupService>& tls) const;
```
mitza-oci, line 106: Why do we have both of these?

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -267,7 +267,11 @@ namespace XTypes {
 
     bool operator<(const TypeObjectHashId& other) const
     {
-      return (kind < other.kind) && (hash < other.hash);
+      if (kind < other.kind) return true;
+      if (other.kind < kind) return false;
+      if (std::memcmp(hash, other.hash, sizeof hash) < 0) return true;
+      if (std::memcmp(other.hash, hash, sizeof hash) < 0) return false;
```
mitza-oci, line 273: Only needs a single memcmp call

mitza-oci (changes requested):

dds/DCPS/TypeSupportImpl.h
sonndinh: Looks like there is no indentation inside namespaces in this file.

mitza-oci: It's described in the guidelines.  Not a big deal though...

mitza-oci (approved):

dds/DCPS/TypeSupportImpl.h
sonndinh: Okay, I see that in the guidelines and have updated this. I think we may add to the rule to say that it applies to the innermost namespace scope.

PR #1910, opened 2020-09-29
Use Bit Masking in RTPS::bitmapNonEmpty()

simpsont-oci:
Previous implementation was iterating over both the bitmap unsigned integer indices and individual bits, which is unnecessary.

This change pulls two of the helper functions out into common library utility locations and adds a unit test for RTPS::bitmapNonEmpty(), as well as modifying the implementation to avoid the double loop.


dds/DCPS/RTPS/Util.cpp
```diff
@@ -0,0 +1,37 @@
+#include "dds/DCPS/RTPS/Util.h"
```
iguessthislldo, line 1:
```suggestion
#include "Util.h"
```

dds/DCPS/RTPS/Util.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#ifndef OPENDDS_RTPS_UTIL_H
+#define OPENDDS_RTPS_UTIL_H
+
+#if !defined (ACE_LACKS_PRAGMA_ONCE)
+#pragma once
+#endif /* ACE_LACKS_PRAGMA_ONCE */
+
+#include "dds/DCPS/RTPS/rtps_export.h"
```
iguessthislldo, line 15:
```suggestion
#include "rtps_export.h"
```

dds/DCPS/Util.h
```diff
@@ -8,6 +8,9 @@
 #ifndef OPENDDS_DCPS_UTIL_H
 #define OPENDDS_DCPS_UTIL_H
 
+#include "dds/DCPS/dcps_export.h"
+#include "dds/DCPS/SequenceNumber.h"
```
iguessthislldo, line 12:
```suggestion
#include "dcps_export.h"
#include "SequenceNumber.h"
```

dds/DCPS/RTPS/Util.h
```diff
@@ -0,0 +1,30 @@
...
+#include "RtpsCoreC.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace RTPS {
+
+OpenDDS_Rtps_Export bool bitmapNonEmpty(const OpenDDS::RTPS::SequenceNumberSet& snSet);
```
iguessthislldo, line 23:
```suggestion
OpenDDS_Rtps_Export bool bitmapNonEmpty(const SequenceNumberSet& snSet);
```

dds/DCPS/RTPS/Util.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include "dds/DCPS/Definitions.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace RTPS {
+
+bool bitmapNonEmpty(const OpenDDS::RTPS::SequenceNumberSet& snSet)
```
iguessthislldo, line 10:
```suggestion
bool bitmapNonEmpty(const SequenceNumberSet& snSet)
```

dds/DCPS/Util.cpp
```diff
@@ -0,0 +1,21 @@
...
+/// Return the number of CORBA::Longs required for the bitmap representation of
+/// sequence numbers between low and high, inclusive (maximum 8 longs).
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+CORBA::ULong
+bitmap_num_longs(const OpenDDS::DCPS::SequenceNumber& low,
+                 const OpenDDS::DCPS::SequenceNumber& high)
```
iguessthislldo, line 12:
```suggestion
CORBA::ULong bitmap_num_longs(const SequenceNumber& low, const SequenceNumber& high)
```

dds/DCPS/Util.cpp
```diff
@@ -0,0 +1,21 @@
+#include "Util.h"
+
+/// Return the number of CORBA::Longs required for the bitmap representation of
+/// sequence numbers between low and high, inclusive (maximum 8 longs).
```
iguessthislldo, line 4: Move this to the header file so it documents for Doxygen?

dds/DCPS/Util.h
```diff
@@ -197,6 +200,8 @@ bool set_intersect(SetA& sA, const SortedB& sB, LessThan lessThan)
   return !sA.empty();
 }
 
+OpenDDS_Dcps_Export CORBA::ULong bitmap_num_longs(const OpenDDS::DCPS::SequenceNumber& low, const OpenDDS::DCPS::SequenceNumber& high);
```
iguessthislldo, line 203:
```suggestion
OpenDDS_Dcps_Export CORBA::ULong bitmap_num_longs(const SequenceNumber& low, const SequenceNumber& high);
```

dds/DCPS/RTPS/Util.cpp
```diff
@@ -0,0 +1,37 @@
...
+  if (num_ulongs == 0) {
+    return false;
+  }
+
+  const size_t last_index = num_ulongs - 1;
+  if (last_index > 0) {
+    for (CORBA::ULong i = 0; i < last_index; ++i)
+    if (snSet.bitmap[i]) {
+      return true;
+    }
```
iguessthislldo, line 25:
```suggestion
    for (CORBA::ULong i = 0; i < last_index; ++i) {
      if (snSet.bitmap[i]) {
        return true;
      }
    }
```

dds/DCPS/RTPS/Util.cpp
simpsont-oci: Without the path, isn't it ambiguous whether we're talking about DCPS/Util.h and DCPS/RTPS/Util.h? Maybe these need to be named differently, but I'd rather not leave it down to order of search path.

dds/DCPS/RTPS/Util.cpp
simpsont-oci: whoops!

dds/DCPS/RTPS/Util.cpp
iguessthislldo: No, `#include "file.h"` always searches the directory of the including file first, then goes on to the directories like that ones passed by `-I`.

iguessthislldo: There is also `#include <file.h>` which skips the local check, and just checks the non-local include paths.

dds/DCPS/RTPS/Util.cpp
```diff
@@ -0,0 +1,38 @@
...
+  if (last_index > 0) {
+    for (CORBA::ULong i = 0; i < last_index; ++i) {
+      if (snSet.bitmap[i]) {
+        return true;
+      }
+    }
+  }
+
+  const CORBA::ULong mod = snSet.numBits % 32;
+  const CORBA::ULong mask = mod ? ~((0x1u << (32 - mod)) - 1) : 0xFFFFFFFF;
```
mitza-oci, line 30:
```suggestion
  const CORBA::ULong mask = mod ? -(1u << (32 - mod)) : 0xFFFFFFFF;
```
I think this gives the same values...

dds/DCPS/Util.h
```diff
@@ -197,6 +200,10 @@ bool set_intersect(SetA& sA, const SortedB& sB, LessThan lessThan)
   return !sA.empty();
 }
 
+/// Return the number of CORBA::Longs required for the bitmap representation of
+/// sequence numbers between low and high, inclusive (maximum 8 longs).
+OpenDDS_Dcps_Export CORBA::ULong bitmap_num_longs(const SequenceNumber& low, const SequenceNumber& high);
```
mitza-oci, line 205: Does this need to be in the core library?  Even though it doesn't use RTPS types, it depends on the RTPS details,  Or maybe it could go with DisjointSequence which does deal with the bitmap representation.

dds/DCPS/RTPS/Util.h
```diff
@@ -0,0 +1,30 @@
+/*
```
mitza-oci, line 1: Combine with BaseMessageUtils.h/cpp?

mitza-oci (changes requested):

dds/DCPS/RTPS/Util.cpp
mitza-oci, line 30: Another way (not sure it's better) to handle the special case of `mod == 0`.  The entire integer is used, so it could just be effectively `++last_index` in the loop above.

simpsont-oci: On my build, the generated assembly winds up being the same, but it's closer to your version so I'll change my code over.

simpsont-oci: I'm not entirely sure either. I think if we were optimizing for the mod == 0 case extending the loop would be ideal, but most of the time (probably much less than 1/32, since the majority of bitmaps will be short) we'll be dealing with a non-zero mod value. As it stands, I think I prefer the clarity of 1) loop & test for beginning ulongs 2) mask & test for final ulong.

mitza-oci: Right, I don't know if one is clearly better than the other but I guess I'm leaning towards fewer operators (especially if we handle `mod==0` somewhere else).

dds/DCPS/RTPS/BaseMessageUtils.cpp
```diff
@@ -199,6 +199,28 @@ DCPS::TimeDuration rtps_duration_to_time_duration(const Duration_t& rtps_duratio
   }
 }
 
+bool bitmapNonEmpty(const OpenDDS::RTPS::SequenceNumberSet& snSet)
```
mitza-oci, line 202:
```suggestion
bool bitmapNonEmpty(const SequenceNumberSet& snSet)
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #1911, opened 2020-09-29
XTypes end-to-end test

kuznetsovmoci:
End-to-end test for XTypes.

A test fully encompassing all parts of an xtypes project. this would mean creating types, writers, and readers to interact with each other.

These interactions with eachother will traverse our discovery and assignability code.

Following that, the communication between parties will traverse our serialization and tryconstruct.


tests/DCPS/XTypes/desktop.ini
```diff
@@ -0,0 +1,3 @@
+[LocalizedFileNames]
+RtpsTypeLookupService.cpp=@RtpsTypeLookupService.cpp,0
+RtpsTypeLookupService.h=@RtpsTypeLookupService.h,0
```
iguessthislldo, line 3: You should remove this.

kuznetsovmoci: This file is in this project is probably by mistake, due to some git manipulations. Can it be removed?

tests/DCPS/XTypes/run_test.pl
```diff
@@ -12,33 +12,44 @@
...
 GetOptions(
   "verbose" => \$verbose,
+  "test|f=s"   => \$test_name,
 );
 
-my @common_args = ('-DCPSConfigFile', 'rtps_disc.ini');
-push(@common_args, "--verbose") if ($verbose);
+my $common_args = ('-DCPSConfigFile rtps_disc.ini -ORBDebugLevel 1 -DCPSDebugLevel 6');
+if ($verbose) {
+    $common_args = ("$common_args, --verbose");
```
iguessthislldo, line 23:
```suggestion
  push(@common_args, "--verbose");
```

iguessthislldo: Otherwise the result is `"-DCPSConfigFile rtps_disc.ini -ORBDebugLevel 1 -DCPSDebugLevel 6, --verbose"`

tests/DCPS/XTypes/run_test.pl
```diff
@@ -12,33 +12,44 @@
...
-$test->start_process('writer2');
+if ($test_name eq '') {
+  while (my ($k, $v) = each %params) {
+    my @reader_args = ("$common_args -ORBLogFile publisher_$k.log --reader --type $v->{reader_type} $v->{expect_to_fail}");
+    $test->process("reader_$k", 'XTypes', join(' ', @reader_args));
+    $test->start_process("reader_$k");
+
+    my @writer_args = ("$common_args -ORBLogFile subscriber_$k.log --writer --type $v->{writer_type} $v->{expect_to_fail}");
+    $test->process("writer_$k", 'XTypes', join(' ', @writer_args));
+    $test->start_process("writer_$k");
```
iguessthislldo, line 39: Make this into a function to share with the case that a specific test case was requested?

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -50,94 +53,93 @@ ReturnCode_t check_additional_field_value(T1 data, AdditionalFieldValue expected
...
+}
+
+template<typename T1, typename T2>
+ReturnCode_t read_tryconstruct_struct(DataReader_var dr, T1 pdr, T2& data, const std::string& expected_value)
+{
+  ReturnCode_t ret = RETCODE_OK;
+  if ((ret = read_i(dr, pdr, data)) == RETCODE_OK) {
     if (data.length() != 1) {
-      ACE_DEBUG((LM_DEBUG, "reader: unexpected data length: %d", data.length()));
+      ACE_DEBUG((LM_ERROR, "ERROR: reader: unexpected data length: %d", data.length()));
```
jwillemsen, line 92: Use also ACE_ERROR

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -92,10 +92,10 @@ ReturnCode_t read_tryconstruct_struct(DataReader_var dr, T1 pdr, T2& data, const
       ACE_ERROR((LM_ERROR, "ERROR: reader: unexpected data length: %d", data.length()));
       ret = RETCODE_ERROR;
     } else if (ACE_OS::strcmp(data[0].trim_string, expected_value.c_str()) != 0) {
-      ACE_ERROR((LM_ERROR, "ERROR: reader: expected key value: %s, received: %s\n", expected_value.c_str(), data[0].trim_string));
+      ACE_ERROR((LM_ERROR, "ERROR: reader: expected key value: %s, received: %s\n", expected_value.c_str(), data[0].trim_string.in()));
```
jwillemsen, line 95: Logging an ascii string should be %C

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,656 @@
...
+      } else if (type == "ModifiedMutableUnion") {
+        write_modified_mutable_union(dw);
+      } else if (type == "Trim64Struct") {
+        write_trim64_struct(dw);
+      } else if (type == "AppendableStructWithDependency") {
+        write_appendable_struct_with_dependency(dw);
+      } else {
+        ACE_ERROR((LM_ERROR, "ERROR: Type %s is not supported\n", type.c_str()));
+        failed = true;
+      }
```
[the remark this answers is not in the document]
kuznetsovmoci: Maybe I misunderstood, but I intended to catch a situation, when in the test file, or in the command line the `type` parameter is incorrect or missing. Wouldn't this code do the job?

tests/DCPS/XTypes/XTypes.idl
```diff
@@ -0,0 +1,141 @@
...
+struct Property_2 {
+  @key long key;
+  string value;
+};
+
+// Final structues
+
+@topic
+@final
+@OpenDDS::data_representation(XCDR2)
```
iguessthislldo, line 20: The `data_representation` annotations can be removed

tests/DCPS/XTypes/run_test.pl
```diff
@@ -0,0 +1,66 @@
...
+  "ModifiedMutableUnion"                  => {reader_type => "ModifiedMutableUnion", writer_type => "ModifiedMutableUnion", expect_to_fail => "", reg_type => "ModifiedMutableUnionT"},
+);
+
+
+sub run_test {
+  my $v = $_[0];
+  my $test_name_param = $_[1];
+
+  if ($v->{reg_type}) {
+  push(@common_args, "--type_r $v->{reg_type}");
```
iguessthislldo, line 43:
```suggestion
    push(@common_args, "--type_r $v->{reg_type}");
```

tests/DCPS/XTypes/run_test.pl
```diff
@@ -0,0 +1,66 @@
...
+use strict;
+
+my $test = new PerlDDS::TestFramework();
+$test->enable_console_logging();
+
+my $verbose = 0;
+my $test_name = "";
+GetOptions(
+  "verbose" => \$verbose,
+  "test|f=s"   => \$test_name,
```
iguessthislldo, line 18:
```suggestion
  "test|f=s" => \$test_name,
```

Also why `-f`? Did you mean `-t`?

tests/DCPS/XTypes/run_test.pl
```diff
@@ -0,0 +1,66 @@
...
+);
+
+my @common_args = ('-DCPSConfigFile rtps_disc.ini -ORBDebugLevel 1 -DCPSDebugLevel 6');
+if ($verbose) {
+  push(@common_args, "--verbose");
+}
+
+my %params = (
+  "FirstTest"                             => {reader_type => "Property_1", writer_type => "Property_1", expect_to_fail => ""},
```
iguessthislldo, line 27: Remove some of the space between the keys and the values. Also maybe make `expected_to_fail` a boolean (0 or 1 in Perl).

tests/DCPS/XTypes/run_test.pl
```diff
@@ -0,0 +1,66 @@
...
+  $test->process("writer_$test_name_param", 'XTypes', join(' ', @writer_args));
+  $test->start_process("writer_$test_name_param");
+}
+
+if ($test_name eq '') {
+  while (my ($k, $v) = each %params) {
+    run_test ($v, $k);
+    sleep 10;
+  }
+} else {
```
iguessthislldo, line 62:
```suggestion
}
else {
```

iguessthislldo (changes requested):

tests/DCPS/XTypes/XTypes.idl
kuznetsovmoci: In all definitions?

iguessthislldo: Yes

tests/DCPS/XTypes/run_test.pl
kuznetsovmoci: Will do later, when all tests are in the list and I know how long is the longest name.

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,668 @@
...
+      } else if (type == "AppendableStructWithDependency") {
+        write_appendable_struct_with_dependency(dw);
+      }
+    }
+
+    ACE_OS::sleep(ACE_Time_Value(3, 0));
+
+    if (failed) {
+      ACE_ERROR((LM_ERROR, "ERROR: Writer failed for type %s\n", type.c_str()));
+    }
```
sonndinh, line 609: I think this hasn't handled all possible cases for `failed` and `expect_to_fail`. It looks to me that the test fails when `failed != expect_to_fail`. Also `failed == true` doesn't always indicate a failed test as implied by the if block at lines 607-609, that is if `failed` is `true` and `expect_to_fail` is also `true`, then the test should be considered passed. 

So I think we could do something like, if `failed != expect_to_fail` then the test failed. Otherwise, if `failed == expect_to_fail` and both are `false`, then perform write/read message depending on whether it is a writer or reader; if `failed == expect_to_fail` but both are `true`, then the test passed for this negative test case and does nothing else.

Same comment applies to the similar logic for the reader case.

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,668 @@
...
+  std::string registered_type_name = "";
+
+  for (int i = 1; i < argc; ++i) {
+    ACE_TString arg(argv[i]);
+    if (arg == ACE_TEXT("--verbose")) {
+      verbose = true;
+    } else if (arg == ACE_TEXT("--writer")) {
+      writer = true;
+    } else if (arg == ACE_TEXT("--reader")) {
+      reader = true;
```
sonndinh, line 465: I think the test is designed to create either a writer or a reader, but not both, for each process, right? If so, maybe check if the user enters both `--writer` and `--reader` options, and print an error message in that case.

tests/DCPS/XTypes/XTypes.cpp
kuznetsovmoci: Part of this logic is handled within `check_inconsistent_topic_status` function, but I've made some changes, and also replaced `expect_to_fail` flag name with 'expect_to_match', as Adam suggested earlier, for clarity.

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,670 @@
...
+      }
+    } else if (arg == ACE_TEXT("--expect_to_fail")) {
+      expect_to_match = false;
+    } else {
+      ACE_ERROR((LM_ERROR, "ERROR: Invalid argument: %s\n", argv[i]));
+      return 1;
+    }
+  }
+
+  if (!(writer ^ reader)) {
```
sonndinh, line 495: Using bitwise operator probably works in this case, but I'd prefer boolean operators because it's safer.

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,670 @@
...
+    ACE_ERROR((LM_ERROR, "ERROR: Invalid endpoint arguments\n"));
```
sonndinh, line 496: Please use a more concrete error message to say user must specify either writer or reader but not both.

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,670 @@
...
+      } else if (type == "Trim20Struct") {
+        failed = !(read_trim20_struct(dr) == RETCODE_OK);
+      } else if (type == "AppendableStructWithDependency") {
+        failed = !(read_appendable_struct_with_dependency(dr) == RETCODE_OK);
+      }
+    }
+  } else {
+    ACE_ERROR((LM_ERROR, "ERROR: Must pass either --writer or --reader\n"));
+    failed = true;
+  }
```
sonndinh, line 663: Because of the newly added check above, this won't be reached and can be deleted.

tests/DCPS/XTypes/XTypes.cpp
iguessthislldo:
```suggestion
  if (writer == reader) {
```

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,667 @@
...
+}
+
+
+void write_mutable_union(const DataWriter_var& dw)
+{
+  MutableUnionDataWriter_var typed_dw = MutableUnionDataWriter::_narrow(dw);
+
+  MutableUnion mu;
+  mu.key(key_value);
```
sonndinh, line 372: Set the discriminator too?

tests/DCPS/XTypes/XTypes.cpp
```diff
@@ -0,0 +1,667 @@
...
+  }
+}
+
+
+void write_modified_mutable_union(const DataWriter_var& dw)
+{
+  ModifiedMutableUnionDataWriter_var typed_dw = ModifiedMutableUnionDataWriter::_narrow(dw);
+
+  ModifiedMutableUnion mmu;
+  mmu.key(key_value);
```
sonndinh, line 385: Set discriminator?

tests/DCPS/XTypes/XTypes.cpp
mitza-oci: The methods in the union that are named for its members always set the discriminator.

sonndinh: That's nice.

sonndinh (approved):

bin/dcps_tests.lst
```diff
@@ -406,6 +406,8 @@ tests/FACE/Compiler/idl_test1_main/run_test.pl: !DCPS_MIN !WCHAR !NO_BUILT_IN_TO
 tests/FACE/Compiler/idl_test3_main/run_test.pl: !DCPS_MIN !WCHAR !NO_BUILT_IN_TOPICS
 tests/FACE/Compiler/idl_test_fixed/run_test.pl: !DCPS_MIN !WCHAR !NO_BUILT_IN_TOPICS
 
+tests/DCPS/XTypes/run_test.pl
```
ClaytonCalabrese, line 409: I think !DCPS_MIN would be a fitting argument for this.

ClaytonCalabrese (approved):

iguessthislldo (approved):

PR #1912, opened 2020-09-29
Statistic Update for RtpsRelay

brattonj-oci:
The multiple publish and report flags have been consolidated so that
there is a single flag for both publishing and reporting.

The RtpsRelay no longer subscribes to statistics topics, it just sends
samples.  An external collector will be created to process the stats.

Participant stats have been seperated from the Handler stats, and the
stats have been changed slightly.  Error and governor stats have been
added, and participant stats are sent in groups to prevent too many
stats from being sent at one time.


jrw972:
Replaced by #1930

PR #1917, opened 2020-09-30
Permissions handles are not returned to Access Control Plugin

jrw972:
Problem
-------

The DDS Security Specification provides no way to return the handles
returned by validate_local_permissions and
validate_remote_permissions.  This creates a resource leak as local
and remote partipants are created and destroyed.

Solution
--------

Add a method return_permissions_handle to free the resources backing
the handle.

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -121,6 +121,20 @@ DomainParticipantImpl::DomainParticipantImpl(
 
 DomainParticipantImpl::~DomainParticipantImpl()
 {
+#ifdef OPENDDS_SECURITY
+  if (security_config_ && perm_handle_ != DDS::HANDLE_NIL) {
+    Security::AccessControl_var access = security_config_->get_access_control();
+    DDS::Security::SecurityException se;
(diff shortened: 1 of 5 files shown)


mitza-oci:
Spec issue (just created): DDSSEC12-89

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -121,6 +121,20 @@ DomainParticipantImpl::DomainParticipantImpl(
 
 DomainParticipantImpl::~DomainParticipantImpl()
 {
+#ifdef OPENDDS_SECURITY
+  if (security_config_ && perm_handle_ != DDS::HANDLE_NIL) {
+    Security::AccessControl_var access = security_config_->get_access_control();
+    DDS::Security::SecurityException se;
+    const bool okay = access->return_permissions_handle(perm_handle_, se);
+    if (!okay) {
+      ACE_ERROR((LM_ERROR,
```
mitza-oci, line 130: should this log statement be conditional like the one below?

jrw972: Yes.

mitza-oci (approved):

PR #1918, opened 2020-10-01
Extensible serialization for TypeObject.h types

mitza-oci:
- Types that are mutable/appendable in Annex B IDL now actually do those things
- Additional bug fixes in TypeIdentifier serialization
- Completed the set of serialized_size and stream insertion/extraction operators for all types


dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -51,30 +49,45 @@ void TypeIdentifier::activate(const TypeIdentifier* other)
...
+  case TK_INT32:
+  case TK_INT64:
+  case TK_UINT16:
+  case TK_UINT32:
+  case TK_UINT64:
+  case TK_FLOAT32:
+  case TK_FLOAT64:
+  case TK_FLOAT128:
+  case TK_CHAR8:
+  case TK_CHAR16:
```
sonndinh, line 89: There are also `TK_INT8` and `TK_UINT8`.

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -97,26 +110,41 @@ void TypeIdentifier::reset()
...
```
sonndinh, line 146: Also `TK_INT8` and `TK_UINT8` here.

iguessthislldo (approved):

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -725,6 +757,19 @@ void compute_dependencies(const TypeMap& type_map,
   }
 }
 
+bool write_empty_xcdr2_nonfinal(DCPS::Serializer& strm)
+{
+  return strm.write_delimiter(0);
```
sonndinh, line 762: `write_delimiter()` function is a bit confusing; I think its argument is the total size that includes 4 bytes of the Dheader itself. So the return statement would be `return strm.write_delimiter(4);` I think.

iguessthislldo: Ah yes, I forgot about that. Almost all uses of it would need the subtraction it's doing, but yea it's a bit confusing if you're doing something manually instead of using `serialized_size`.

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -1021,30 +1140,79 @@ bool operator>>(Serializer& strm, XTypes::CompleteTypeDetail& stru)
...
+  serialized_size_delimiter(encoding, size);
   serialized_size(encoding, size, stru.base_type);
   serialized_size(encoding, size, stru.detail);
 }
 
 bool operator<<(Serializer& strm, const XTypes::CompleteStructHeader& stru)
 {
+  size_t total_size = 0;
+  serialized_size(strm.encoding(), total_size, stru);
+  if (!strm.write_delimiter(total_size)) {
```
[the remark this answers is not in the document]
mitza-oci: I think it's already accounted for in `total_size`.  The code here is adapted from what `opendds_idl` generates, but simplified to always assume XCDR2.  @iguessthislldo please check as well.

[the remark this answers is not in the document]
sonndinh: Oh yes, I forgot the updated `serialized_size` has it already. Please ignore this and other similar comments.

sonndinh (approved):

dds/DCPS/XTypes/TypeObject.cpp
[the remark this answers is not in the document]
iguessthislldo: Yes, this is this is the use case `write_delimiter` was written for. Adding the delimiter size manually here would be wrong.

dds/DCPS/StaticDiscovery.cpp
```diff
@@ -532,7 +532,7 @@ StaticEndpointManager::writer_exists(const RepoId& writerid, const RepoId& reade
     wa.writerQos = writer_pos->second.qos;
 #else
     const WriterAssociation wa =
-      {writer_pos->second.trans_info, writerid, writer_pos->second.publisher_qos, writer_pos->second.qos};
+      {writer_pos->second.trans_info, writerid, writer_pos->second.publisher_qos, writer_pos->second.qos, 0};
```
jrw972, line 535: Thank you.

jrw972 (approved):

PR #1923, opened 2020-10-02
XTypes Type Lookup Service compiler warning fixes

kuznetsovmoci:


dds/DCPS/RTPS/TypeLookup.cpp
```diff
@@ -216,8 +216,8 @@ namespace DCPS {
 
   bool operator>>(Serializer& strm, XTypes::TypeLookup_Call& stru)
   {
-    ACE_CDR::Octet kind;
-    if (!(strm >> ACE_InputCDR::to_octet(kind))) {
+    CORBA::ULong kind;
```
mitza-oci, line 219:
```suggestion
    CORBA::Long kind;
```
From 7.6.3.3.3

mitza-oci (changes requested):

dds/DCPS/RTPS/TypeLookup.cpp
kuznetsovmoci: Hash IDs are unsigned in the spec:
// computed from @hashid("getTypes")
const unsigned long TypeLookup_getTypes_HashId = 0x018252d3;
// computed from @hashid("getDependencies");
const unsigned long TypeLookup_getDependencies_HashId = 0x05aafb31;

I must be missing something...

mitza-oci: The part of the code we're looking at is reading the union's discriminator.  The union's discriminator type is `long`.

kuznetsovmoci: They use `switch(long)`, indeed, but is it inconsistent with the hash id type?

mitza-oci: Maybe an issue somewhere, but not right here.

I don't think it's really an issue since it's just specifying the bit pattern and it's a positive 32-bit signed integer so it has the exact same representation as an unsigned 32-bit integer.  But it would be nice for the spec to just be consistent.

mitza-oci (changes requested):

dds/DCPS/RTPS/TypeLookup.cpp
```diff
@@ -298,8 +298,8 @@ namespace DCPS {
 
   bool operator>>(Serializer& strm, XTypes::TypeLookup_Return& stru)
   {
-    ACE_CDR::Octet kind;
-    if (!(strm >> ACE_InputCDR::to_octet(kind))) {
+    CORBA::Long kind;
```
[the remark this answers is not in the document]
kuznetsovmoci: Could you be looking at the older version? It has been fixed...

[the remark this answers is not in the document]
kuznetsovmoci: Never mind, I see it elsewhere

mitza-oci (approved):

PR #1926, opened 2020-10-05
Extensibility-aware serialization for Type Lookup Service

mitza-oci:


dds/DCPS/RTPS/TypeLookup.cpp
```diff
@@ -8,384 +8,654 @@
...
-        return true;
-      }
+
+    bool must_understand = false;
+    if (!strm.read_parameter_id(member_id, field_size, must_understand)) {
       return false;
     }
+
+    switch (member_id) {
+    case 41962193:
```
jrw972, line 151: Magic number and elsewhere.

jrw972 (changes requested):

dds/DCPS/RTPS/TypeLookup.cpp
mitza-oci: It's generated by `opendds_idl` that way.

kuznetsovmoci (approved):

dds/DCPS/RTPS/TypeLookup.cpp
iguessthislldo: Yea, I didn't know where to put them. It seems like it'd be more awkward to do anything else.

iguessthislldo (approved):

PR #1932, opened 2020-10-07
XTypes: Enum bit_bound assignability refinement

sonndinh:
Update the assignability of enum types so that 2 enum types with different bit_bounds are not assignable.


mitza-oci (approved):

mitza-oci:
Travis CI has an error about .gitignore:
```
ERROR These files should be gitignored
Would remove tests/DCPS/UnitTests/UnitTests_XTypesAssignability
```

PR #1934, opened 2020-10-08
Fixed compiler warnings

pollocktoci:
These changes are to fix several compiler warnings.


.gitignore
```diff
@@ -44,3 +44,4 @@ lib*.dylib
 .vs/
 /tags
 /host_tools.mwc*
+*.recipe
```
mitza-oci, line 47: Could this be in local configuration instead of in the repo?

pollocktoci:
Most likely.  Is that a separate .gitignore file?

.gitignore
mitza-oci: https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

tools/rtpsrelay/RelayHandler.h
```diff
@@ -246,7 +246,7 @@ class VerticalHandler : public RelayHandler {
 
   void populate_address_map(AddressMap& address_map, const GuidSet& to);
 
-  virtual uint32_t local_active_participants() const override { return guid_addr_set_map_.size(); }
+  virtual uint32_t local_active_participants() const override { return (uint32_t)guid_addr_set_map_.size(); }
```
mitza-oci, line 249: static_cast?

mitza-oci (changes requested):
There are merge conflicts due to concurrent changes in rtpsrelay

mitza-oci (approved):

PR #1940, opened 2020-10-10
BIT readers have spec-defined memory leak

jrw972:
Problem
-------

The DDS Specification says that the members of the
ReaderDataLifecycleQosPolicy are set to infinite which means a
participant will never purge the data stored in the builtin topics.  A
long-running participant will eventually run out of memory.

It is possible to manually change the qos for all of the builtin
participant data readers.  However, the addition of a new builtin
topic would render such code useless and the new builtin topic would
create another memory leak.

Solution
--------

Provide a mechanism to configure the ReaderDataLifecycleQosPolicy for
the builtin topic data readers.

### DIFF
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -204,7 +204,9 @@ Service_Participant::Service_Participant()
     monitor_enabled_(false),
     shut_down_(false),
     shutdown_listener_(0),
-    default_configuration_file_(ACE_TEXT(""))
+    default_configuration_file_(ACE_TEXT("")),
+    bit_autopurge_nowriter_samples_delay_({DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC}),
+    bit_autopurge_disposed_samples_delay_({DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC})
 {
   initialize();
 }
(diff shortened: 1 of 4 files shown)


jrw972:
As an alternative, we could provide a function that sets the durations on all of the readers.

mitza-oci:
Spec issue DDS15-46 https://issues.omg.org/issues/lists/dds-rtf5#issue-14097

dds/DCPS/Service_Participant.cpp
```diff
@@ -204,7 +204,9 @@ Service_Participant::Service_Participant()
     monitor_enabled_(false),
     shut_down_(false),
     shutdown_listener_(0),
-    default_configuration_file_(ACE_TEXT(""))
+    default_configuration_file_(ACE_TEXT("")),
+    bit_autopurge_nowriter_samples_delay_({DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC}),
+    bit_autopurge_disposed_samples_delay_({DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC})
```
mitza-oci, line 209: This will fail on older compilers

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -203,6 +203,11 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
   }
 #endif
 
+  const DDS::Duration_t one_minute = { 0, 0 };
```
mitza-oci, line 206:
```suggestion
  const DDS::Duration_t one_minute = { 60, 0 };
```

mitza-oci (changes requested):

tools/rtpsrelay/RtpsRelay.cpp
jrw972: Oops.

mitza-oci (approved):

PR #1941, opened 2020-10-10
Deque is memory inefficient

jrw972:
Problem
-------

Currently, a typical DDS application might involve 50+
ReactorInterceptors.  Each ReactorInterceptor includes a deque.  A
deque is built from internal nodes.  The internal nodes typically have
a minimum size which is the lesser of the size of the value type and
some constant.  This constant can be much bigger than the size of the
value type.  Consequently, a deque that contains a only a few entries,
such as one for a ReactorInterceptor, is highly memory inefficient.
If an empty deque is 1K, then a typical DDS application will have 50K+
of memory that is not actually being used.

Solution
--------

Change ReactorInterceptor and Job queue to use vectors that are
swapped with empty vectors as they are being processed.

### DIFF
--- a/dds/DCPS/JobQueue.h
+++ b/dds/DCPS/JobQueue.h
@@ -38,7 +38,7 @@ public:
   {
     ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
     const bool empty = job_queue_.empty();
-    job_queue_.push(job);
+    job_queue_.push_back(job);
     if (empty) {
       guard.release();
       reactor()->notify(this);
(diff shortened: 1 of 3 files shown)


simpsont-oci (approved):
Overall, no issue switching away from deque, but have you profiled this at all? I'm a little worried vector will behave poorly if we're growing it up and then swapping back to zero all the time.  Also see the queue used in ThreadPerConnectionSendTask, which I think contains an OPENDDS_DEQUE internally.

jrw972:
> Overall, no issue switching away from deque, but have you profiled this at all? I'm a little worried vector will behave poorly if we're growing it up and then swapping back to zero all the time. Also see the queue used in ThreadPerConnectionSendTask, which I think contains an OPENDDS_DEQUE internally.

I think that is the only other use of deque but I was unwilling to change it since I don't know the expected behavior for that queue.  I have not profiled it but I do know that the empty deques were consuming significant memory on Linux.  We can easily try a list if we find that vector is not performant.

PR #1942, opened 2020-10-12
The TAO DevGuide can be requested for free

jwillemsen:
* README.md:

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ details.
 - The OpenDDS Developer's Guide is freely downloadable at:
 http://download.objectcomputing.com/OpenDDS/
 
-- The TAO Developer's Guide book set may also be purchased from
+- The TAO Developer's Guide book can be requested for free from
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ details.
 - The OpenDDS Developer's Guide is freely downloadable at:
 http://download.objectcomputing.com/OpenDDS/
 
-- The TAO Developer's Guide book set may also be purchased from
+- The TAO Developer's Guide book can be requested for free from
```
mitza-oci, line 50: Should it move to the ACE/TAO section of the document?

jwillemsen: I would just keep it here

iguessthislldo: I would agree with moving it. We could maybe also just remove it entirely if we have other links were someone could find it through if they wanted to know more about ACE/TAO.

jwillemsen: Moved

iguessthislldo (approved):

PR #1944, opened 2020-10-13
Updating Bench Improved Discovery, CPU, Memory Stats

simpsont-oci:
Problem:
Bench 2 does not allow for comparative resource usage between test runs, nor does it allow for visibility into minimum, mean, median DDS entity discovery times (concerning itself currently with only the worse-case maximum).

Solution:
Add CPU and Memory utilization statistics to the node_controller process and promote "discovery time delta" to first-class statistic, similar to latency & jitter.


simpsont-oci:
> Does have a return value which you should check

Agreed. This was pulled in from another developer's changes. I'll add better error handling before this gets out of draft.

simpsont-oci:
For @sonndinh & @kuznetsovmoci, this is still a draft PR so some of the details for getting CPU & Memory Stats still need work (the Linux variant is completely unimplemented... not even sure we'll add a MacOS version). But hopefully doing this review will help get you more familiar with Bench 2.

performance-tests/bench_2/node_controller/main.cpp
```diff
@@ -649,7 +714,7 @@ void wait_for_full_scenario(
     if (allocated_scenario.scenario_id != TAO::String_Manager() &&
         initial_attempt + std::chrono::seconds(30) < std::chrono::system_clock::now())
     {
```
sonndinh, line 716:
```suggestion
    if (allocated_scenario.scenario_id != TAO::String_Manager() &&
        initial_attempt + std::chrono::seconds(30) < std::chrono::system_clock::now()) {
```

performance-tests/bench_2/node_controller/ProcessStatsCollector.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    last_time_ = time;
+    last_sys_time_ = sys_time;
+    last_user_time_ = user_time;
+  }
+#endif
+  return percent;
+}
+
+double ProcessStatsCollector::get_virtual_mem_usage() noexcept
```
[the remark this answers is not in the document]
sonndinh: I saw your comment above. I guess this will be implemented sometime later.

kuznetsovmoci (approved):

performance-tests/bench_2/node_controller/ProcessStatsCollector.cpp
```diff
@@ -0,0 +1,294 @@
...
+    std::string token;
+    while (meminfofile >> token)
+    {
+      if (token == "MemTotal:")
+      {
+        meminfofile >> total_mem_;
+        break;
+      }
+      if (token == "VmallocTotal:")
+      {
```
iguessthislldo, line 162:
```suggestion
      if (token == "VmallocTotal:") {
```

PR #1946, opened 2020-10-15
Xtypes: Extensibility test fixes

ClaytonCalabrese:
Adding exten test to test list and removing cases that shouldn't have existed in the first place.

### DIFF
--- a/tests/DCPS/Compiler/XtypesExensibility/run_test.pl
+++ b/tests/DCPS/Compiler/XtypesExensibility/run_test.pl
@@ -1,0 +1,22 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
+     & eval 'exec perl -S $0 $argv:q'
+     if 0;
+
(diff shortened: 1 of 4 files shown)


iguessthislldo (approved):

tests/DCPS/Compiler/XtypesExensibility/run_test.pl
```diff
@@ -0,0 +1,22 @@
...
+use Env (ACE_ROOT);
+use lib "$ACE_ROOT/bin";
+use PerlDDS::Run_Test;
+use strict;
+
+my $name = 'Compiler_XtypesExtensibility';
+my $status = 0;
+my $TESTDVR = PerlDDS::create_process ($name);
+
+my $status = $TESTDVR->SpawnWaitKill (300);
```
jwillemsen, line 16: 300 seconds, when there would be a problem the build waits 5 minutes before killing the process, is that really required to be so long?

PR #1950, opened 2020-10-16
fix and test for enum union cpp11

ClaytonCalabrese:
In CPP11 if a union had an enum as a union discriminator, it would cause a seg fault in idl if all cases of the enum had a branch in the union.


dds/idl/langmap_generator.cpp
```diff
@@ -213,8 +213,13 @@ struct GeneratorBase
         {
           AST_Enum* e = AST_Enum::narrow_from_decl(the_union->disc_type());
           if (be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11) {
-            first_label << scoped(e->name()) << "::"
-              << e->value_to_name(dv.u.enum_val)->last_component()->get_string();
+            first_label << scoped(e->name()) << "::";
+            UTL_ScopedName * default_name = e->value_to_name(dv.u.enum_val);
```
iguessthislldo, line 217:
```suggestion
            UTL_ScopedName* default_name = e->value_to_name(dv.u.enum_val);
```

dds/idl/langmap_generator.cpp
```diff
@@ -213,8 +213,13 @@ struct GeneratorBase
...
+            //default_name can be null when there is not default branch
+            if (default_name == 0) {
+              default_name = e->value_to_name(0);
```
mitza-oci, line 220: This doesn't seem to match the spec's requirement "In all other cases it is initialized to the first discriminant value specified in IDL"

Would it be easier to detect that the `dv` is not in the valid range of the enum before calling `value_to_name`?

mitza-oci (changes requested):

dds/idl/langmap_generator.cpp
ClaytonCalabrese: Isn't the value that maps to 0 always the first element?

mitza-oci: Which one is default in this case?  The spec seems to require `B`.
```
enum E {A, B};
union U switch (E) {
  case B: octet o1;
  case A: octet o2;
};
```

mitza-oci: Please extend the test to cover this case (or something similar).

dds/idl/langmap_generator.cpp
```diff
@@ -219,9 +219,29 @@ struct GeneratorBase
...
+              if (pos == std::string::npos) {
+                prefix = "";
+              } else {
+                prefix = prefix.substr(0, pos) + "::";
+              }
+            } else {
+              prefix += "::";
+            }
+            first_label << prefix;
+            UTL_ScopedName* default_name = e->value_to_name(dv.u.enum_val);
```
mitza-oci, line 236: Would it be easier to detect that the dv is not in the valid range of the enum before calling value_to_name?

mitza-oci (approved):

PR #1951, opened 2020-10-17
Secure discovery is slow

jrw972:
Problem 1
---------

The default configuration delays an SPDP resend for 3 seconds.
However, a participant A receiving an initial SPDP announcement from B
will attempt authentication with B.  However, B will ignore the
authentication messages because it has not discovered A.

Solution 1
----------

Send an SPDP announcement when attempting authentication with a newly
discovered participant.

Problem 2
---------

`match_authenticated` is called before the FINAL in the auth handshake
is sent.  This is problematic because `match_authenticated` is slow,
e.g., takes 1 second to complete.  The other participant is idle while
waiting for the FINAL and will call `match_authenticated` when it gets
the FINAL.

Solution 2
----------

Send the FINAL before calling `match_authenticated`.  While it doesn't
address the reason that `match_authenticated` is so slow, it helps
because both participants can be in `match_authenticated` at the same
time.

Problem 3
---------

`match_authenticated` associates all of the builtin readers and
writers.  At more-or-less the same time, it is sending crypto tokens
for the builtin endpoints.  This creates a race where the heartbeats
and acknacks for reliability can get dropped because the crypto tokens
are not in place to decrypt them.  This means discovery has to wait
one or two rounds of reliability which is 1 or 2 seconds by default.

Solution 3
----------

Defer associating the builtin endpoints until the crypto tokens have
been received for the remote.  This suggests that a similar race
condition may exist for user endpoints.

### DIFF
--- a/dds/DCPS/RTPS/SecurityHelpers.h
+++ b/dds/DCPS/RTPS/SecurityHelpers.h
@@ -37,12 +37,6 @@ const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_WRITER = {{0x
 const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_READER = {{0xff, 0x01, 0x01}, 0xc7};
 ///@}
 
-inline bool
-is_secure(const DCPS::RepoId& id)
-{
-  return id.entityId.entityKey[0] == 0xff;
-}
-
 const DDS::Security::ParticipantSecurityInfo PARTICIPANT_SECURITY_ATTRIBUTES_INFO_DEFAULT = {0, 0};
 const DDS::Security::EndpointSecurityInfo ENDPOINT_SECURITY_ATTRIBUTES_INFO_DEFAULT = {0, 0};
 
--- a/dds/DCPS/RTPS/SecurityHelpers.h
+++ b/dds/DCPS/RTPS/SecurityHelpers.h
@@ -37,6 +37,12 @@ const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_WRITER = {{0x
 const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_READER = {{0xff, 0x01, 0x01}, 0xc7};
 ///@}
 
+inline bool
+is_secure(const DCPS::RepoId& id)
+{
+  return id.entityId.entityKey[0] == 0xff;
+}
+
 const DDS::Security::ParticipantSecurityInfo PARTICIPANT_SECURITY_ATTRIBUTES_INFO_DEFAULT = {0, 0};
 const DDS::Security::EndpointSecurityInfo ENDPOINT_SECURITY_ATTRIBUTES_INFO_DEFAULT = {0, 0};
 
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -816,7 +816,7 @@ create_association_data_proto(DCPS::AssociationData& proto,
 
 #ifdef OPENDDS_SECURITY
 void
-Sedp::associate_preauth(const Security::SPDPdiscoveredParticipantData& pdata)
+Sedp::associate_preauth(Security::SPDPdiscoveredParticipantData& pdata)
 {
   // First create a 'prototypical' instance of AssociationData.  It will
   // be copied and modified for each of the (up to) four SEDP Endpoints.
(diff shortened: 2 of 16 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -632,7 +632,15 @@ Spdp::handle_participant_data(DCPS::MessageId id,
         iter->second.property_qos_ = pdata.ddsParticipantDataSecure.base.property;
         iter->second.security_info_ = pdata.ddsParticipantDataSecure.base.security_info;
 
+        // The remote needs to see our SPDP before attempting authentication.
+        if (from_relay) {
+          tport_->write_i(guid, SpdpTransport::SEND_TO_RELAY);
+        } else {
+          tport_->write_i(guid, SpdpTransport::SEND_TO_LOCAL);
+        }
```
mitza-oci, line 640:
```suggestion
        tport_->write_i(guid, from_relay ? SpdpTransport::SEND_TO_RELAY : SpdpTransport::SEND_TO_LOCAL);
```

dds/DCPS/RTPS/SecurityHelpers.h
```diff
@@ -37,6 +37,12 @@ const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_WRITER = {{0x
 const EntityId_t ENTITYID_SPDP_RELIABLE_BUILTIN_PARTICIPANT_SECURE_READER = {{0xff, 0x01, 0x01}, 0xc7};
 ///@}
 
+inline bool
+is_secure(const DCPS::RepoId& id)
+{
+  return id.entityId.entityKey[0] == 0xff;
```
mitza-oci, line 43: Check for builtin 'kind'?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1024,110 +1032,149 @@ void Sedp::remove_remote_crypto_handle(const RepoId& participant, const EntityId
...
+  }
+}
+
+void Sedp::associate_secure_reader_to_writer(const RepoId& remote_writer)
+{
+  using namespace DDS::Security;
+
+  ParticipantData_t& pdata = spdp_.get_participant_data(remote_writer);
+
+  using namespace DDS::Security;
```
mitza-oci, line 1095:
```suggestion
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1510,32 +1517,31 @@ Spdp::handle_participant_crypto_tokens(const DDS::Security::ParticipantVolatileM
...
   }
 
-  send_our_tokens = !dp.is_requester_ && !dp.seen_some_crypto_tokens_;
-  dp.seen_some_crypto_tokens_ = true;
   if (dp.handshake_state_ == DCPS::HANDSHAKE_STATE_WAITING_FOR_TOKEN) {
     dp.handshake_state_ = DCPS::HANDSHAKE_STATE_DONE;
     purge_handshake_deadlines(iter);
   }
 
+  // TODO: Does a secure remote always send a participant token?
```
mitza-oci, line 1525: TODO

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -977,11 +984,12 @@ void Sedp::rekey_volatile(const Security::SPDPdiscoveredParticipantData& pdata)
 
 #endif // OPENDDS_SECURITY
 
-void Sedp::disassociate_helper(const BuiltinEndpointSet_t& avail, const CORBA::ULong flags,
+void Sedp::disassociate_helper(BuiltinEndpointSet_t& avail, const CORBA::ULong flags,
                                const RepoId& id, const EntityId_t& ent, DCPS::TransportClient& client)
 {
   if (avail & flags) {
     client.disassociate(make_id(id, ent));
+    avail ^= flags;
```
jrw972, line 992: This is incorrect.

mitza-oci: I was wondering about that... is there a need to change `avail` at all here?

mitza-oci (approved):

dds/DCPS/security/AccessControlBuiltInImpl.h
```diff
@@ -260,20 +260,27 @@ class DdsSecurity_Export AccessControlBuiltInImpl
...
+      time_t expiration;
+
+      Entry() : handle(0), expiration(0) {}
+      Entry(DDS::Security::PermissionsHandle a_handle,
+            time_t a_expiration)
+        : handle(a_handle)
+        , expiration(a_expiration)
+      {}
+    };
+    typedef OPENDDS_MAP(DDS::Security::PermissionsHandle, Entry*) EntryMap;
```
mitza-oci, line 281: Does this need to be a pointer?

mitza-oci (changes requested):

dds/DCPS/security/AccessControlBuiltInImpl.h
jrw972: It works better with `schedule_timer`.

mitza-oci (approved):

PR #1957, opened 2020-10-20
Minor changes to BaseMessageUtils to eliminate compile warnings.

pollocktoci:


dds/DCPS/RTPS/BaseMessageUtils.cpp
```diff
@@ -209,15 +209,15 @@ bool bitmapNonEmpty(const SequenceNumberSet& snSet)
     return false;
   }
 
-  const size_t last_index = num_ulongs - 1;
+  const CORBA::ULong last_index = num_ulongs - 1;
```
mitza-oci, line 212: Will this generate a new warning for 64-bit to 32-bit truncation?

mitza-oci (changes requested):

dds/DCPS/RTPS/BaseMessageUtils.cpp
pollocktoci: Right.  I'll revert that.

mitza-oci (changes requested):
There are still warnings here:
```
BaseMessageUtils.cpp(212,49): warning C4267: 'initializing': conversion from 'size_t' to 'const CORBA::ULong', possible loss of data
BaseMessageUtils.cpp(214,8): warning C4267: 'argument': conversion from 'size_t' to 'CORBA::ULong', possible loss of data
```

dds/DCPS/RTPS/BaseMessageUtils.cpp
pollocktoci: Looking at other usages of SequenceNumberSet, such as in the RtpsCoreTypeSupportImpl RtpsCoreTypeSupportImpl::operator<< function, Corba::ULong is used instead of size_t.  Should I follow that example and change bitmapNonEmpty to use Corba::ULong for num_ulongs and in the for loop?  I think that might be the correct change.

mitza-oci: Yes, I think so.  The reasoning is that IDL sequences use 32-bit unsigned integers for their lengths and subscripting functions.  Especially the way these particular sequences are used, there is no chance that we'll have more than 4.3 billion elements.

pollocktoci: Ok.  I'll make num_ulongs and the for loop use CORBA::ULong.

mitza-oci (approved):

dds/idl/be_produce.cpp
```diff
@@ -138,18 +138,14 @@ string to_macro(const char* fn)
 #ifdef max
 #undef max
 #endif
```
iguessthislldo, line 140: I think you can get rid of these as well, unless `max` is being used somewhere below.

pollocktoci: Done

mitza-oci (approved):

mitza-oci:
```
Running trailing_whitespaces check
Error: ./dds/idl/be_produce.cpp:140: found trailing whitespace
```

dds/idl/be_produce.cpp
```diff
@@ -135,21 +135,15 @@ string to_macro(const char* fn)
...
 
   unsigned int seed = static_cast<unsigned int>(msec);
-#ifdef max
-#undef max
-#endif
-  const float MAX_VAL = static_cast<float>(numeric_limits<char>::max());
-  const float coefficient = static_cast<float>(MAX_VAL / (RAND_MAX + 1.0f));
 
   if (ret[ret.size() - 1] != '_') ret += '_';
+
```
mitza-oci, line 140:
```suggestion

```

PR #1959, opened 2020-10-20
Added XTYPES switch to RTPS Discovery configuration

kuznetsovmoci:
Added XTYPES switch to RTPS Discovery configuration. Made Type Information addition to the parameter list conditional.


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -497,6 +495,8 @@ Sedp::init(const RepoId& guid,
   }
 #endif
 
+  max_type_lookup_service_reply_period_ = disco.config()->max_type_lookup_service_reply_period();
```
mitza-oci, line 498: Why is this line changing?

mitza-oci (changes requested):
Should this also control the `from_param_list` direction?  If we don't change that code, the local participant can receive an endpoint announcement from a peer that includes type information and process it -- doesn't seem like what we want.

dds/DCPS/RTPS/Sedp.cpp
kuznetsovmoci: I moved it from above, where it was in the middle of writers /  readers setup.

mitza-oci (approved):
.

PR #1960, opened 2020-10-20
TransportSendBuffer leaks memory

jrw972:
Problem
-------

Entries are inserred into the `destinations_` map but never removed.

Solution
--------

Remove entries from `destinations_` when the corresponding entry is
removed from `buffers_`.

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendBuffer.cpp
+++ b/dds/DCPS/transport/framework/TransportSendBuffer.cpp
@@ -132,6 +132,7 @@ SingleSendBuffer::release_i(BufferMap::iterator buffer_iter)
     }
   }
 
+  destinations_.erase(buffer_iter->first);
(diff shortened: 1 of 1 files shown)


dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -162,6 +163,7 @@ SingleSendBuffer::remove_i(BufferMap::iterator buffer_iter, BufferVec& removed)
     }
   }
 
+  destinations_.erase(buffer_iter->first);
```
simpsont-oci, line 166: Since release_i() and remove_i() clean this up now, I think you can get rid of remove destinations_.erase(it->first) in check_capacity_i() which calls remove_i right after.

simpsont-oci (approved):

PR #1961, opened 2020-10-20
XTypes: opendds_idl command-line options

sonndinh:
Suppport default settings for several XTypes annotations (extensibility, autoid, try-construct) in opendds_idl.


dds/idl/be_global.cpp
```diff
@@ -434,6 +434,20 @@ BE_GlobalData::parse_args(long& i, char** av)
           av[i + 1]));
         idl_global->parse_args_exit(1);
       }
+    } else if (!std::strcmp(av[i], "--default-data-representation")) {
+      default_data_representation_.set_all(false);
```
iguessthislldo, line 438: I missed this part. So yeah this would work, but you could only choose one representation, where as the annotation allows you to chose multiple representations.

sonndinh: Yes, it only allows one default representation. I can change to let it allow more than one representation, but I don't know if providing that flexibility is really necessary. I'll leave this out as we discussed.

iguessthislldo: It would be necessary to be able to specify multiple representations if we were really doing it, because that's what the annotation does, but yes leave it out for now.

dds/idl/be_global.cpp
```diff
@@ -422,16 +424,49 @@ BE_GlobalData::parse_args(long& i, char** av)
     } else if (!ACE_OS::strncasecmp(av[i], NO_DCPS_DATA_TYPE_WARNINGS_FLAG, NO_DCPS_DATA_TYPE_WARNINGS_FLAG_SIZE)) {
       warn_about_dcps_data_type_ = false;
     } else if (!std::strcmp(av[i], "--default-extensibility")) {
-      if (!std::strcmp(av[i + 1], "final")) {
+      if (av[i + 1] == nullptr) {
```
iguessthislldo, line 427:
```suggestion
      if (av[++i] == 0) {
```

Same with other arguments.

dds/idl/be_global.cpp
```diff
@@ -422,16 +424,49 @@ BE_GlobalData::parse_args(long& i, char** av)
...
+        ACE_ERROR((LM_ERROR, ACE_TEXT("No argument for --default-extensibility\n")));
+        idl_global->parse_args_exit(1);
+      }
+      if (!std::strcmp(av[++i], "final")) {
```
iguessthislldo, line 431:
```suggestion
      } else if (!std::strcmp(av[i], "final")) {
```

Same with other arguments.

dds/idl/be_util.cpp
```diff
@@ -116,28 +116,33 @@ be_util::usage()
...
     ACE_TEXT("to RapidJSON JavaScript objects\n")
-    ACE_TEXT(" --[no-]default-nested\tTopic types must be declared. True by default.\n")
+    ACE_TEXT(" --[no-]default-nested\ttopic types must be declared. True by default.\n")
     ACE_TEXT(" --no-dcps-data-type-warnings\t\tdon't warn about #pragma DCPS_DATA_TYPE\n")
-    ACE_TEXT(" -Wb,export_macro=<macro name>\t\tsets export macro ")
+    ACE_TEXT(" --default-extensibility\t\tset default XTypes extensibility kind")
+    ACE_TEXT(" (appendable if not set)\n")
+    ACE_TEXT(" --default-autoid\t\t\tset default XTypes autoid approach (sequential if not set)\n")
+    ACE_TEXT(" --default-try-construct\t\tset default XTypes try-construct behavior")
+    ACE_TEXT(" (discard if not set)\n")
```
iguessthislldo, line 125: It would be good to list what the valid arguments for these are.

dds/idl/be_util.cpp
```diff
@@ -116,28 +116,33 @@ be_util::usage()
     ACE_TEXT("\t\t\t\t-Wb,v8 is an alternative form for this option\n")
     ACE_TEXT(" -Grapidjson\t\tgenerate TypeSupport for converting data samples ")
     ACE_TEXT("to RapidJSON JavaScript objects\n")
-    ACE_TEXT(" --[no-]default-nested\tTopic types must be declared. True by default.\n")
+    ACE_TEXT(" --[no-]default-nested\ttopic types must be declared. True by default.\n")
```
iguessthislldo, line 119:
```suggestion
    ACE_TEXT(" --[no-]default-nested\ttopic types must be declared, true by default\n")
```

iguessthislldo (changes requested):

iguessthislldo:
Could you also add a null check to `-o`?

sonndinh:
> Could you also add a null check to `-o`?

Sure. I just added it.

iguessthislldo (approved):

PR #1968, opened 2020-10-23
Check sample validity in SimpleLatency performance test

simpsont-oci:
Make test script executable and check sample validity in SimpleLatency performance test before making calculations on sample data.


sonndinh (approved):
I noticed that some `if/else` blocks in this test don't follow the style guidelines. It's nice to have them fixed here or maybe that can be done later.

PR #1969, opened 2020-10-24
fixed warnings, mostly integer truncation

mitza-oci:


sonndinh (approved):

tests/DCPS/MetaStruct/MetaStructTest.cpp
```diff
@@ -100,7 +105,7 @@ int run_test(int, ACE_TCHAR*[])
   Source src;
   src.rhs_a.s = "hello";
   src.rhs_a.l = 42;
-  src.rhs_a.w = L'☺';
+  src.rhs_a.w = L'\x263a';
```
iguessthislldo, line 108: What's the use of wide characters if we can't use emojis? /s

iguessthislldo (approved):

ClaytonCalabrese (approved):

tests/DCPS/MetaStruct/MetaStructTest.cpp
mitza-oci: It's still "used" but just not visible directly in the source code.
To avoid warnings we need to avoid characters that aren't in the "basic source character set"
https://en.cppreference.com/w/cpp/language/translation_phases

sonndinh: I'm curious what hex input the compiler sees from the "smiley face" of the old code. Isn't it also encoded by `0x263a`?

PR #1980, opened 2020-10-28
Add Bench 2 Documentation

simpsont-oci:


iguessthislldo (changes requested):
I don't think we should add HTML documentation to the repo.

mitza-oci:
> 
> 
> I don't think we should add HTML documentation to the repo.

Put it on `gh-pages` instead?

simpsont-oci:
> I don't think we should add HTML documentation to the repo.

I guess I don't get this.

Having the HTML in the repo means the documentation (and accompanying images) is viewable by anyone with a browser (i.e. everyone), even when they're offline. If you're worried about file churn when the HTML gets re-generated, perhaps we can check it in with pretty-printing turned on (one element per line).

We've also already got HTML committed in a few places:
- java (a few folders)
- tools/modeling (numerous)
- performance-tests/Bench/doc/userguide.html (Bench v1)

iguessthislldo:
Yeah and I'm trying to get rid of those other HTML files and have gotten of a couple of them. There's no reason to add HTML when other things like Markdown and RST exist.

simpsont-oci:
> There's no reason to add HTML when other things like Markdown and RST exist.

Does the average user have anything to preview these files with images included? I'd say that's a pretty good reason to include the rendered HTML. It seems like some browsers have plugins for MD and RST, but AFAIK it's not supported by default.

iguessthislldo:
The end result is not that this would be 100% perfect for users to read out of the box if they just have the repo. The end result is something that can be put on the website and pdf and is easy for us to maintain in a repo. HTML only meets one of those requirements and I never considered it as a candidate for documentation.

simpsont-oci:
> HTML only meets one of those requirements and I never considered it as a candidate for documentation.

Right. But I'm not advocating that HTML becomes our primary candidate for documentation. I'm pointing out that it's handy to still have the HTML in the repo for users (alongside MD or RST or whatever we're using to render the HTML, if it exists), especially if images are involved. And as long as we have a script for generating up-to-date HTML (again, if the source document is in the repo), there's no reason the two should ever be out of sync (and it's an easy fix, if they are... we could even make it part of the fuzz checks).

iguessthislldo:
> And as long as we have a script for generating up-to-date HTML (again, if the source document is in the repo), there's no reason the two should ever be out of sync (and it's an easy fix, if they are... we could even make it part of the fuzz checks).

This isn't something I'm interesting in doing. I think the generated documentation should be hosted elsewhere.

PR #1989, opened 2020-10-29
Sequence keys need to report errors (not infinite loop in opendds_idl)

mitza-oci:


iguessthislldo:
I don't know how it would report the error as you have it, but we could add a case to `TopicKeys::Iterator::operator++()` to report it properly.

iguessthislldo (approved):

PR #1993, opened 2020-10-30
RTPS Transport does send durable data for first ACKNACK

jrw972:
Problem
-------

The "association complete" callback for a reliable RTPS writer is
called when the writer receives the first acknack for a reader.  If
the writer is durable, it causes the writer to send the durable data
for the new reader.  This is currently done and the end of the ACKNACK
processing.  Durable data is sent in response to an ACKNACK.
Consequently, a reader must send one ACKNACK to enqueue the durable
data and another to retrieve it.

Solution
--------

Invoke the callback earlier so the durable data will be available for
sending in response to the first ACKNACK.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -2776,25 +2776,20 @@ RtpsUdpDataLink::RtpsWriter::process_acknack(const RTPS::AckNackSubmessage& ackn
     return;
   }
 
-  bool first_ack = false;
-
   if (!ri->second.handshake_done_) {
     ri->second.handshake_done_ = true;
-    first_ack = true;
-  }
-
-  if (first_ack) {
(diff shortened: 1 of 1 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2778,11 +2776,27 @@ RtpsUdpDataLink::RtpsWriter::process_acknack(const RTPS::AckNackSubmessage& ackn
...
   }
 
+  bool first_ack = false;
+
   if (!ri->second.handshake_done_) {
     ri->second.handshake_done_ = true;
     first_ack = true;
   }
 
+  if (first_ack) {
```
mitza-oci, line 2786: Is `first_ack` needed?

jrw972: Good point.

mitza-oci (approved):

PR #2010, opened 2020-11-05
ReactorTask has a busy loop

jrw972:
Problem
-------

The thread monitoring code in ReactorTask runs for one cycle and then
degenerates into a busy loop.

Solution
--------

Reset the time given to run the event loop.


dds/DCPS/ReactorTask.cpp
```diff
@@ -181,21 +180,17 @@ OpenDDS::DCPS::ReactorTask::svc()
         key += " (" + name_ + ")";
       }
 
-      ACE_Time_Value t = timeout_.value();
-
       while (state_ == STATE_RUNNING) {
+        ACE_DEBUG((LM_DEBUG, "### Loop 3\n"));
```
simpsont-oci, line 184:
```suggestion
```

PR #2017, opened 2020-11-06
durability_resend memory leak fix

pollocktoci:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3388,7 +3388,8 @@ RtpsUdpDataLink::RtpsWriter::acked_by_all_helper_i(TqeSet& to_deliver)
 
 void RtpsUdpDataLink::durability_resend(TransportQueueElement* element)
 {
-  static const RTPS::FragmentNumberSet none = {{0}, 0, RTPS::LongSeq8()};
+  CORBA::Long buffer[8];
```
simpsont-oci, line 3391:
```suggestion
  static const CORBA::Long buffer[8];
```

pollocktoci: That results in:

error C2440: '<function-style-cast>': cannot convert from 'initializer list' to 'OpenDDS::RTPS::LongSeq8'

Because of how the LongSeq8 ctor is defined:

OpenDDS::RTPS::LongSeq8::LongSeq8 (
    ::CORBA::ULong length,
    ::CORBA::Long * buffer,
    ::CORBA::Boolean release
  )

mitza-oci: No `const` here

pollocktoci: Changed to:

static CORBA::Long buffer[8];

mitza-oci (approved):

PR #1633, opened 2020-04-17
XTypes Branch Integration

mitza-oci:

PR #2024, opened 2020-11-09
Crypto handles and security attributes are leaked

jrw972:
Problem
-------

SEDP and the RTPS transport both maintain maps mapping GUIDs to crypto
handles and security attributes.  Neither cleanup the maps.

Solution
--------

Factor out the maps to a shared component in the security config and
add appropriate logic for cleaning up crypto handles.


dds/DCPS/security/HandleRegistryImpl.cpp
```diff
@@ -0,0 +1,320 @@
...
+HandleRegistryImpl::~HandleRegistryImpl()
+{
+  if (DCPS::security_debug.bookkeeping) {
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) {bookkeeping} ")
+               ACE_TEXT("HandleRegistryImpl::~HandleRegistryImpl local datareader %d local datawriter %d remote participant %d remote datareader %d remote datawriter %d\n"),
+               local_datareader_crypto_handles_.size(),
+               local_datawriter_crypto_handles_.size(),
+               remote_participant_crypto_handles_.size(),
+               remote_datareader_crypto_handles_.size(),
+               remote_datawriter_crypto_handles_.size()));
```
mitza-oci, line 19: 32/64-bit mismatch between log format string and parameter

dds/DCPS/DiscoveryBase.h
```diff
@@ -1398,6 +1422,17 @@ namespace OpenDDS {
       {
         return crypto_key_exchange_;
       }
+
+      inline void set_handle_registry(OpenDDS::Security::HandleRegistry* hr)
```
mitza-oci, line 1426: remove redundant namespace qualification

mitza-oci (changes requested):
not a complete review

dds/DCPS/DiscoveryBase.h
```diff
@@ -526,13 +520,34 @@ namespace OpenDDS {
...
 
+#ifdef OPENDDS_SECURITY
+      void cleanup_secure_writer(const RepoId& publicationId)
+      {
+        using namespace DDS::Security;
+
+        const DatawriterCryptoHandle dwch = get_handle_registry()->get_local_datawriter_crypto_handle(publicationId);
+        SecurityException ex = {"", 0, 0};
+        if (!get_crypto_key_factory()->unregister_datawriter(dwch, ex)) {
+          ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: Sedp::cleanup_secure_writer() - ")
```
iguessthislldo, line 531: Sounds closer to an error than a warning to me.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1344,12 +1316,12 @@ Sedp::send_builtin_crypto_tokens(
   const DCPS::RepoId dst = make_id(dstParticipant, dstEntity);
   if (DCPS::GuidConverter(src).isReader()) {
     create_and_send_datareader_crypto_tokens(
```
iguessthislldo, line 1318: I don't understand the purpose of listing the arguments this way. Could you move the first argument up to this line? Same with the datawriter case?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4778,65 +4674,59 @@ void Sedp::resend_user_crypto_tokens(const RepoId& id)
...
+      if (!DCPS::GuidPrefixEqual()(reader_pos->guidPrefix, remote_participant.guidPrefix)) {
+        continue;
+      }
+      const DDS::Security::DatareaderCryptoHandle drch =
+        get_handle_registry()->get_remote_datareader_crypto_handle(*reader_pos);
+      if (drch == DDS::HANDLE_NIL) {
+        continue;
       }
+      create_and_send_datawriter_crypto_tokens(dwch, writer_pos->first, drch, *reader_pos);
     }
```
iguessthislldo, line 4730: The whole function look better. It isn't two solid blocks of if and for loop text anymore.

dds/DCPS/RcHandle_T.h
```diff
@@ -307,6 +307,12 @@ RcHandle<T> make_rch(U0 const& u0, U1 const& u1, U2 const& u2, U3 const& u3, U4
   return RcHandle<T>(new T(unwrap_reference(u0), unwrap_reference(u1), unwrap_reference(u2), unwrap_reference(u3), unwrap_reference(u4), unwrap_reference(u5), unwrap_reference(u6), unwrap_reference(u7)), keep_count());
 }
 
+template <typename T, typename U0, typename U1, typename U2, typename U3, typename U4, typename U5, typename U6, typename U7, typename U8>
+RcHandle<T> make_rch(U0 const& u0, U1 const& u1, U2 const& u2, U3 const& u3, U4 const& u4, U5 const& u5, U6 const& u6, U7 const& u7, U8 const& u8)
```
iguessthislldo, line 311: We will be able to eliminate these in C++14 right?

dds/DCPS/security/HandleRegistry.h
```diff
@@ -0,0 +1,76 @@
+/*
+*
+*
+* Distributed under the OpenDDS License.
+* See: http://www.OpenDDS.org/license.html
+*/
+
+#ifndef DDS_DCPS_HANDLE_REGISTRY_H
+#define DDS_DCPS_HANDLE_REGISTRY_H
```
iguessthislldo, line 9:
```suggestion
#ifndef OPENDDS_DCPS_HANDLE_REGISTRY_H
#define OPENDDS_DCPS_HANDLE_REGISTRY_H
```

dds/DCPS/security/HandleRegistryImpl.h
```diff
@@ -0,0 +1,88 @@
+/*
+*
+*
+* Distributed under the OpenDDS License.
+* See: http://www.OpenDDS.org/license.html
+*/
+
+#ifndef DDS_DCPS_HANDLE_REGISTRY_IMPL_H
+#define DDS_DCPS_HANDLE_REGISTRY_IMPL_H
```
iguessthislldo, line 9:
```suggestion
#ifndef OPENDDS_DCPS_HANDLE_REGISTRY_IMPL_H
#define OPENDDS_DCPS_HANDLE_REGISTRY_IMPL_H
```

iguessthislldo (changes requested):

dds/DCPS/security/framework/SecurityConfig.h
```diff
@@ -111,6 +118,7 @@ class OpenDDS_Dcps_Export SecurityConfig : public DCPS::RcObject {
   CryptoKeyExchange_var key_exchange_plugin_;
   CryptoKeyFactory_var key_factory_plugin_;
   CryptoTransform_var transform_plugin_;
+  HandleRegistry* handle_registry_plugin_;
```
simpsont-oci, line 121: Can you even do this? Isn't the security config across the whole application? What if you have two secure local participants talking to the same secure remote participant? Won't the handle lookups clash?

dds/DCPS/security/HandleRegistryImpl.cpp
jrw972: What's the correct way to do this?  Cast up to 64-bit int?

dds/DCPS/RcHandle_T.h
jrw972: I sure hope so.

dds/DCPS/security/framework/SecurityConfig.h
jrw972: Can, yes.  Should, no...for the exact reason you mentioned.  Interestingly enough, the existing approach to security "works" by using opaque handles so it is safe to share that state between participants.

I think what I'll do is create other class called `SecurityContext` to hold a participants security-related state that is 1) not explicitly handled in the SecurityConfig and 2) must be shared across module boundaries (the main one being security and transport).  This creates the following problem (which goes back to the "design" of the Transport framework):  who owns the SecurityContext?  Discovery?  The transport?  The transport framework?  Or does the SecurityConfig gain an additional map of participant to SecurityContext.  I think I'll go with the last one unless there is a strong objection.

dds/DCPS/security/HandleRegistryImpl.cpp
mitza-oci: There's a log format character for `size_t`

dds/DCPS/DiscoveryBase.h
```diff
@@ -526,13 +520,35 @@ namespace OpenDDS {
...
 
+#ifdef OPENDDS_SECURITY
+      void cleanup_secure_writer(const RepoId& publicationId)
+      {
+        using namespace DDS::Security;
+
+        const DatawriterCryptoHandle dwch = get_handle_registry()->get_local_datawriter_crypto_handle(publicationId);
+        SecurityException ex = {"", 0, 0};
+        if (!get_crypto_key_factory()->unregister_datawriter(dwch, ex)) {
+          ACE_ERROR((LM_ERROR,
```
jwillemsen, line 531: Dependent on some log level?

iguessthislldo: Since it's security related, I will change it to use a `security_debug` flag.

mitza-oci:
Closing this one.  Changes that are needed for this feature are already a part of the xtypes branch so they'll be in #1633.

PR #2026, opened 2020-11-09
Adding SIO_UDP_CONNRESET flag to unicast socket in RtpsUdpDataLink

simpsont-oci:
In order to fix intermittently failing tests with "connection reset" error message.


simpsont-oci:
My bad, this was a 3.14 merge issue. Not sure why there would be recent scoreboard failures for this, but I can't see a reason for this PR anymore.

PR #2028, opened 2020-11-09
Bench2: Support For Definition of Content Filtered Topics, Worker "Action" to Adjust Content Filter Parameters

kuznetsovmoci:


performance-tests/bench_2/builder/TopicManager.cpp
```diff
@@ -29,5 +29,13 @@ DDS::ContentFilteredTopic_var TopicManager::get_content_filtered_topic_by_name(c
   return result;
 }
 
+bool TopicManager::enable(bool throw_on_error) {
```
iguessthislldo, line 32:
```suggestion
bool TopicManager::enable(bool throw_on_error)
{
```

performance-tests/bench_2/builder_idl/Builder.idl
```diff
@@ -297,6 +297,7 @@ module Builder {
   struct DataReaderConfig {
     string name;
     string topic_name;
+    string cft_name;
```
simpsont-oci, line 300: I think I would prefer to use just the single topic_name and sort out where it belongs internally. I realize this will make the lookup a bit more complicated, but after taking another look at the spec, it seems like all the topic classes (normal, content filtered, multitopic) inherit from TopicDescription which holds info about the participant, name, and type name. This makes me think a function like "lookup_topicdescription" on the DDS participant is secretly dictating that normal topics, content filtered topics, and multitopics for the same participant can't share a name... and so I'd rather have our API reflect that.

Also worth noting that the multitopic class (which we may one day support in bench) doesn't differentiate between normal and content filtered topics in its create method, it just takes an SQL-like expression that lists topic names as strings... we'll have to decide internally if they're normal or content filtered.

simpsont-oci (changes requested):
Looks good, for the most part. Biggest change is that I think we'll need to switch to using a single topic name for the DataReaderConfig and figure things out internally. See the comment on the IDL file.

performance-tests/bench_2/worker/ForwardAction.cpp
```diff
@@ -13,10 +13,10 @@ namespace Bench {
 ForwardAction::ForwardAction(ACE_Proactor& proactor) : proactor_(proactor), started_(false), stopped_(false), prevent_copy_(false), force_copy_(false), copy_threshold_(0), queue_first_(0), queue_last_(0) {
 }
 
-bool ForwardAction::init(const ActionConfig& config, ActionReport& report, Builder::ReaderMap& readers, Builder::WriterMap& writers) {
-
+bool ForwardAction::init(const ActionConfig& config, ActionReport& report,
+  Builder::ReaderMap& readers, Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map) {
```
iguessthislldo, line 17:
```suggestion
    Builder::ReaderMap& readers, Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map) {
```

performance-tests/bench_2/idl/Bench.idl
```diff
@@ -35,6 +35,7 @@ module Bench {
     Builder::PropertySeq params;
     Builder::StringSeq readers;
     Builder::StringSeq writers;
+    string content_filtered_topic;
```
simpsont-oci, line 38: Let's not add this here, I wouldn't expect most actions to know about content filtered topics.

performance-tests/bench_2/worker/SetCftParametersAction.cpp
```diff
@@ -0,0 +1,137 @@
+#include "SetCftParametersAction.h"
+
+#include "MemFunHandler.h"
+
+namespace {
+
+uint32_t one_at_a_time_hash(const uint8_t* key, size_t length) {
```
simpsont-oci, line 7: Fine to reuse this, but let's move it to a common header / cpp location (bench2/Common?) so it's not duplicated.

performance-tests/bench_2/worker/SetCftParametersAction.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+bool SetCftParametersAction::init(const ActionConfig& config, ActionReport& report,
+  Builder::ReaderMap& readers, Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map)
+{
+
+  std::unique_lock<std::mutex> lock(mutex_);
+  Action::init(config, report, readers, writers, cft_map);
+
+  if (!content_filtered_topic_) {
```
simpsont-oci, line 40: Instead of expecting the common Action::Init() to populate a content filtered topic name, let's check for a config parameter called "content_filtered_topic_name" (or something like that) here and then check the map. Alternatively, you could even have the action attach a reader and use get_topicdescription on the reader to get a handle to the CFT.

performance-tests/bench_2/worker/SetCftParametersAction.cpp
```diff
@@ -0,0 +1,141 @@
...
+  param_count_ = param_count;
+
+  if (!param_count_) {
+    std::stringstream ss;
+    ss << "SetCftParametersAction '" << config.name << "' is missing a valid param_count parameter" << std::flush;
+    throw std::runtime_error(ss.str());
+  }
+
+  // First check frequency as double (seconds)
+  auto write_frequency_prop = get_property(config.params, "write_frequency", Builder::PVK_DOUBLE);
```
simpsont-oci, line 73: These parameter names probably ought to get changed from "write_XXX" to "change_XXX" or something similar.

performance-tests/bench_2/worker/SetCftParametersAction.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+  std::unique_lock<std::mutex> lock(mutex_);
+  if (started_ && !stopped_) {
+    if (max_count_ == 0 || set_call_count_ < max_count_) {
+      ++set_call_count_;
+
+      DDS::StringSeq params(param_count_);
+      params.length(param_count_);
+      for (size_t i = 0; i < param_count_; i++) {
+        int new_val = mt_();
```
simpsont-oci, line 129: Hmm, we will probably need to think about a useful way of setting a bounds on this. Simply having random numbers as our parameters means we'll (almost) never match the actual content. Probably the simplest thing to do will be to have a list of acceptable values to for each parameter. We can also add ranges of values, though that makes more sense for numbers than strings. We might also want to add the ability to configure whether we're iterating over the acceptable values or randomly choosing one (a "change_pattern" parameter).

performance-tests/bench_2/worker/configs/simple_action_content_filtered_topics_config.json
```diff
@@ -0,0 +1,111 @@
...
+        "qos": { "entity_factory": { "autoenable_created_entities": false } },
+        "qos_mask": { "entity_factory": { "has_autoenable_created_entities": false } },
+
+        "topics": [
+          { "name": "topic_01",
+            "type_name": "Bench::Data",
+            "content_filtered_topics": [
+              {
+                "cft_name": "cft_1",
+                "cft_expression": "msg_count > %0",
```
simpsont-oci, line 46: Looking at the Bench::Data IDL structure, it doesn't look like there's that much we can usefully filter against. Let's add a field to it called filter_class or msg_class or something (ulong is fine to keep the size down) and then update the WriteAction to populate it. We may need to have a little bit of logic in there for usefully being able to iterate over over "acceptable" values for it (similar to what I described for the CFT action, but not as complex).

performance-tests/bench_2/worker/configs/simple_action_content_filtered_topics_config.json
```diff
@@ -0,0 +1,111 @@
+{
+  "enable_time": { "sec": -1, "nsec": 0 },
+  "start_time": { "sec": -3, "nsec": 0 },
+  "stop_time": { "sec": -10, "nsec": 0 },
```
simpsont-oci, line 4: Let's add an actual run_test.pl to the worker directory that launches the worker with whatever config is specified (e.g. "./worker configs/simple_action_content_filtered_topics_config.json") as if we were going to add it to dcps_test.lst). We'll eventually want to show that the reader received a message to check for correctness, but don't worry about that just yet.  Let's just get it so that your code is exercised. Once we figure out a nice efficient way of checking for success within the worker (we're not quite there yet), we'll add the test to dcps_tests.lst.

simpsont-oci (changes requested):

performance-tests/bench_2/builder/BuilderProcess.cpp
```diff
@@ -8,12 +8,13 @@ BuilderProcess::BuilderProcess(const ProcessConfig& config)
   : config_sections_(std::make_shared<ConfigSectionManager>(config.config_sections))
   , discoveries_(std::make_shared<DiscoveryManager>(config.discoveries))
   , instances_(std::make_shared<TransportInstanceManager>(config.instances))
-  , participants_(std::make_shared<ParticipantManager>(config.participants, report_.participants, reader_map_, writer_map_)) {
+  , participants_(std::make_shared<ParticipantManager>(config.participants, report_.participants, reader_map_, writer_map_, cft_map_)) {
```
sonndinh, line 11:
```suggestion
  , participants_(std::make_shared<ParticipantManager>(config.participants, report_.participants, reader_map_, writer_map_, cft_map_))
{
```

performance-tests/bench_2/builder/BuilderProcess.cpp
```diff
@@ -8,12 +8,13 @@ BuilderProcess::BuilderProcess(const ProcessConfig& config)
...
 }
 
 BuilderProcess::~BuilderProcess() {
```
sonndinh, line 14:
```suggestion
BuilderProcess::~BuilderProcess()
{
```

performance-tests/bench_2/builder/DataReaderManager.cpp
```diff
@@ -2,7 +2,8 @@
 
 namespace Builder {
 
-DataReaderManager::DataReaderManager(const DataReaderConfigSeq& configs, DataReaderReportSeq& reports, DDS::Subscriber_var& subscriber, const std::shared_ptr<TopicManager>& topics, ReaderMap& reader_map) {
+DataReaderManager::DataReaderManager(const DataReaderConfigSeq& configs, DataReaderReportSeq& reports, DDS::Subscriber_var& subscriber,
+  const std::shared_ptr<TopicManager>& topics, ReaderMap& reader_map, const ContentFilteredTopicMap& cft_map) {
```
sonndinh, line 6:
```suggestion
  const std::shared_ptr<TopicManager>& topics, ReaderMap& reader_map, const ContentFilteredTopicMap& cft_map)
{
```

performance-tests/bench_2/builder/ParticipantManager.cpp
```diff
@@ -2,10 +2,11 @@
 
 namespace Builder {
 
-ParticipantManager::ParticipantManager(const ParticipantConfigSeq& configs, ParticipantReportSeq& reports, ReaderMap& reader_map, WriterMap& writer_map) {
+ParticipantManager::ParticipantManager(const ParticipantConfigSeq& configs, ParticipantReportSeq& reports,
+  ReaderMap& reader_map, WriterMap& writer_map, ContentFilteredTopicMap& cft_map) {
```
sonndinh, line 6:
```suggestion
  ReaderMap& reader_map, WriterMap& writer_map, ContentFilteredTopicMap& cft_map)
{
```

performance-tests/bench_2/common/util.h
```diff
@@ -8,9 +8,30 @@
 #include <ace/ace_wchar.h> // For ACE_TCHAR
 #include <ace/Default_Constants.h> // For ACE_DIRECTORY_SEPARATOR_*
 #include <ace/SString.h>
+#include "ace/Time_Value.h"
 
 #include "Bench_Common_Export.h"
 
+namespace {
+
+  uint32_t one_at_a_time_hash(const uint8_t* key, size_t length) {
```
sonndinh, line 17:
```suggestion
  uint32_t one_at_a_time_hash(const uint8_t* key, size_t length)
  {
```

performance-tests/bench_2/worker/ForwardAction.cpp
sonndinh, line 17:
```suggestion
  Builder::ReaderMap& readers, Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map)
{
```

performance-tests/bench_2/worker/WriteAction.cpp
```diff
@@ -30,10 +31,11 @@ WriteAction::WriteAction(ACE_Proactor& proactor)
 {
 }
 
-bool WriteAction::init(const ActionConfig& config, ActionReport& report, Builder::ReaderMap& readers, Builder::WriterMap& writers) {
+bool WriteAction::init(const ActionConfig& config, ActionReport& report, Builder::ReaderMap& readers,
+  Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map) {
```
sonndinh, line 35:
```suggestion
  Builder::WriterMap& writers, const Builder::ContentFilteredTopicMap& cft_map)
{
```

performance-tests/bench_2/builder/DataReader.h
```diff
@@ -8,7 +8,8 @@ namespace Builder {
 
 class DataReader : public ListenerFactory<DDS::DataReaderListener> {
 public:
-  explicit DataReader(const DataReaderConfig& config, DataReaderReport& report, DDS::Subscriber_var& subscriber, const std::shared_ptr<TopicManager>& topics);
+  explicit DataReader(const DataReaderConfig& config, DataReaderReport& report, DDS::Subscriber_var& subscriber,
```
mitza-oci, line 11:
```suggestion
  DataReader(const DataReaderConfig& config, DataReaderReport& report, DDS::Subscriber_var& subscriber,
```

performance-tests/bench_2/worker/SetCftParametersAction.cpp
```diff
@@ -0,0 +1,156 @@
+#include "SetCftParametersAction.h"
+
+#include "MemFunHandler.h"
+#include "util.h"
+
+namespace Bench {
+
+  SetCftParametersAction::SetCftParametersAction(ACE_Proactor& proactor)
```
mitza-oci, line 8:
```suggestion
SetCftParametersAction::SetCftParametersAction(ACE_Proactor& proactor)
```

performance-tests/bench_2/builder/DataReader.cpp
```diff
@@ -93,7 +87,28 @@ DataReader::DataReader(const DataReaderConfig& config, DataReaderReport& report,
     enable_time_->value.time_prop(create_time_->value.time_prop());
   }
 
-  datareader_ = subscriber_->create_datareader(topic_, qos_, listener_, listener_status_mask_);
+  auto topic_ptr = topics->get_topic_by_name(topic_name_);
+  if (topic_ptr) {
+    DDS::Topic_var topic = topic_ptr->get_dds_topic();
+    datareader_ = subscriber_->create_datareader(topic, qos_, listener_, listener_status_mask_);
+  } else {
+#ifdef _DDS_CONTENTFILTEREDTOPIC__TRAITS_
```
mitza-oci, line 95:
```suggestion
#ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
```

mitza-oci (dismissed):

performance-tests/bench_2/builder/Topic.cpp
```diff
@@ -83,6 +83,7 @@ Topic::Topic(const TopicConfig& config, DDS::DomainParticipant_var& participant,
     throw std::runtime_error(ss.str());
   }
 
+#ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
   for (unsigned int i = 0; i != content_filtered_topics_.length(); ++i) {
     DDS::ContentFilteredTopic_var cft =
       participant_->create_contentfilteredtopic(content_filtered_topics_[i].cft_name,
```
simpsont-oci, line 89: It dawns on me that we're not currently supporting delayed enable() (based the autoenable_created_entities QoS policy being set to false). I think the easiest thing to do here is to call enable all content filtered topics for the current topic in Topic::enable().

performance-tests/bench_2/common/util.h
```diff
@@ -14,21 +14,22 @@
 
 namespace {
```
simpsont-oci, line 15: This was only in an anonymous namespace to protect it from symbol collisions. Now that it's being upgraded to a proper utility, let's move it into the Bench namespace below (along with ZERO_TIME) and put the implementation for one_at_a_time_hash in the .cpp and not the header.

performance-tests/bench_2/worker/configs/simple_action_content_filtered_topics_config.json
```diff
@@ -115,8 +115,14 @@
         },		
         { "name": "set_frequency",
           "value": { "_d": "PVK_DOUBLE", "double_prop": 2.0 }
-        }
-      ]
+        },
+        { "name": "random_order",
+          "value": { "_d": "PVK_ULL", "ull_prop": 1 }
+        },		
+		{ "name": "acceptable_param_values",
```
simpsont-oci, line 122: Clean up the tabs in here

performance-tests/bench_2/worker/run_test.pl
```diff
@@ -0,0 +1,38 @@
...
+
+$test->{dcps_debug_level} = 0;
+$test->{dcps_transport_debug_level} = 0;
+$test->{add_transport_config} = 0;
+$test->{add_orb_log_file} = 0;
+$test->{add_pending_timeout} = 0;
+
+my $is_rtps_disc = 1;
+
+$test->setup_discovery("-ORBDebugLevel 1 -ORBLogFile DCPSInfoRepo.log") unless $is_rtps_disc;
```
simpsont-oci, line 32: Can we just remove this, or does it still want to start the InfoRepo? I can't remember how PerlDDS defaults off the top of my head.

simpsont-oci (changes requested):

performance-tests/bench_2/builder/Topic.cpp
kuznetsovmoci: I am getting this error:

error C2039: 'enable': is not a member of 'DDS::ContentFilteredTopic'
Does ContentFilteredTopic have its own enable function?

mitza-oci: Only Entities can be enabled.  CFT is not an Entity.

simpsont-oci: Ah, I assumed they were still entities. Nevermind!

simpsont-oci (approved):

PR #2030, opened 2020-11-09
XTypes: Fixes for Older C++03 Only Compilers

iguessthislldo:
Includes Errors and Warnings

### DIFF
--- a/dds/DCPS/LinuxNetworkConfigMonitor.cpp
+++ b/dds/DCPS/LinuxNetworkConfigMonitor.cpp
@@ -51,6 +51,7 @@ bool LinuxNetworkConfigMonitor::open()
     nlmsghdr header;
     rtgenmsg msg;
   } request;
+  memset(&request, 0, sizeof(request));
 
   // Request a dump of the links.
   request.header.nlmsg_len = NLMSG_LENGTH(sizeof(request.msg));
--- a/dds/DCPS/LinuxNetworkConfigMonitor.cpp
+++ b/dds/DCPS/LinuxNetworkConfigMonitor.cpp
@@ -50,7 +50,7 @@ bool LinuxNetworkConfigMonitor::open()
   struct {
     nlmsghdr header;
     rtgenmsg msg;
-  } request = {};
+  } request;
 
   // Request a dump of the links.
   request.header.nlmsg_len = NLMSG_LENGTH(sizeof(request.msg));
--- a/dds/DCPS/RTPS/TypeLookup.h
+++ b/dds/DCPS/RTPS/TypeLookup.h
@@ -82,7 +82,7 @@ struct TypeLookup_getTypeDependencies_Result { //IDL: union switch (long)
 };
 
 struct TypeLookup_Call { //IDL: union switch (long)
-  ACE_CDR::Long kind;
+  ACE_CDR::ULong kind;
 
   // case TypeLookup_getTypes_Hash:
   TypeLookup_getTypes_In getTypes;
@@ -102,7 +102,7 @@ struct TypeLookup_Request {
 };
 
 struct TypeLookup_Return { //IDL: union switch (long)
-  ACE_CDR::Long kind;
+  ACE_CDR::ULong kind;
 
   // case TypeLookup_getTypes_Hash:
   TypeLookup_getTypes_Result getType;
--- a/dds/DCPS/RTPS/TypeLookup.h
+++ b/dds/DCPS/RTPS/TypeLookup.h
@@ -82,7 +82,7 @@ struct TypeLookup_getTypeDependencies_Result { //IDL: union switch (long)
 };
 
 struct TypeLookup_Call { //IDL: union switch (long)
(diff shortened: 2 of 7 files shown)


dds/DCPS/LinuxNetworkConfigMonitor.cpp
```diff
@@ -50,7 +50,7 @@ bool LinuxNetworkConfigMonitor::open()
   struct {
     nlmsghdr header;
     rtgenmsg msg;
-  } request = {};
+  } request;
```
mitza-oci, line 53: does this need a memset or similar?

dds/DCPS/RTPS/TypeLookup.h
```diff
@@ -82,7 +82,7 @@ struct TypeLookup_getTypeDependencies_Result { //IDL: union switch (long)
 };
 
 struct TypeLookup_Call { //IDL: union switch (long)
-  ACE_CDR::Long kind;
+  ACE_CDR::ULong kind;
```
mitza-oci, line 85: It was that way to match the IDL in 7.6.3.3.3 in the spec.

mitza-oci (changes requested):

dds/DCPS/RTPS/TypeLookup.h
iguessthislldo: I did it because I got this warning:

```
Sedp.cpp: In member function ‘DDS::ReturnCode_t OpenDDS::RTPS::Sedp::TypeLookupReplyReader::process_type_lookup_reply(const OpenDDS::DCPS::ReceivedDataSample&, OpenDDS::DCPS::Serializer&, bool)’:
Sedp.cpp:4031: warning: comparison between signed and unsigned integer expressions
```

https://github.com/objectcomputing/OpenDDS/blob/2168fc3ac3ca1acea8e6f4b162baea1e3ab2254b/dds/DCPS/RTPS/Sedp.cpp#L4031

I assumed it was a typo on our part. Since it's in the spec, then the spec is contradicting itself. It declares the id constants as `unsigned long`s, but the union switch type is `long`. Unsigned would be correct since they're member ids.

mitza-oci: It's signed in the RPC spec, so I assume that's where it came from.
I guess I'd rather have the constants match the spec and use a cast in the code.

dds/DCPS/LinuxNetworkConfigMonitor.cpp
iguessthislldo: Maybe? That's part of the reason I asked for Justin's review. At least for my system's `nlmsghdr`:

```c
struct nlmsghdr {
	__u32		nlmsg_len;	/* Length of message including header */
	__u16		nlmsg_type;	/* Message content */
	__u16		nlmsg_flags;	/* Additional flags */
	__u32		nlmsg_seq;	/* Sequence number */
	__u32		nlmsg_pid;	/* Sending process port ID */
};
```

We're setting everything but `nlmsg_seq`. The [man page](https://man7.org/linux/man-pages/man7/netlink.7.html) does seem to suggest that we should set it and track it.

My system's `rtgenmsg` just has `rtgen_family` that we're setting.

dds/DCPS/RTPS/TypeLookup.h
iguessthislldo: I'd rather fix the spec if it's contradicting it self. I don't really know anything about the RPC spec, but if it's supposed to be `long`, then it sounds like the constants have to change.

dds/DCPS/LinuxNetworkConfigMonitor.cpp
mitza-oci: `nlmsg_seq` seems to be a distinct issue.
For this change in particular I think the safest thing to do is to memset to 0 the number of bytes based on sizeof.  That makes it compatible with different system headers.  The man page is less than clear on what it expects.

iguessthislldo: Yes, that's fine with me.

jrw972: I agree with the `memset` recommendation.

dds/DCPS/RTPS/TypeLookup.h
jrw972: I agree with both of you.  I recommend:

1. The IDL should match the spec.
2. Create a ticket against the spec for the contradiction.
3. Put cast in code to silence warning.
4. Create internal ticket to revisit spec in future.

  Can we create an issue for the spec to work out the

jrw972 (changes requested):
Same changes as @mitza-oci

dds/DCPS/RTPS/TypeLookup.h
iguessthislldo: I'd be fine with using a cast as long as an issue is filed.

mitza-oci (approved):

PR #2031, opened 2020-11-10
Bench2: Add tag-based stats

sonndinh:


performance-tests/bench_2/test_controller/main.cpp
```diff
@@ -146,7 +198,17 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
           << "--overwrite-result           Write the result when using --result-id, even if" << std::endl
-          << "                             it would overwrite another result." << std::endl;
+          << "                             it would overwrite another result." << std::endl
+          << "--override-bench-partition-suffix SUFFIX     Override the system-wide partition suffix." << std::endl
+          << "--override-create-time N                     Override the system-wide creation time." << std::endl
+          << "--override-enable-time N                     Override the system-wide enabling time." << std::endl
+          << "--override-start-time N                      Override the system-wide starting time." << std::endl
+          << "--override-stop-time N                       Override the system-wide stopping time." << std::endl
+          << "--override-destruction-time N                Override the system-wide destruction time." << std::endl
+          << "--tags N [TAGS...]                           List of tags for which the user wants to collect" << std::endl
```
simpsont-oci, line 208: I think the count as a parameter is a little clunky. How about just allowing "--tag <tag name>" to be specified multiple times? It's more verbose, but (I hope) less prone to usage errors.

simpsont-oci: Reading through your PR, it dawns on me that what we'll probably really want is some way to specify tag "expressions" from the command line... something like "tagA && !(tagB || tagC)", but let's not worry about that for now. Tag names on their own are a good start.

performance-tests/bench_2/test_controller/main.cpp
```diff
@@ -528,53 +629,95 @@ int handle_reports(const Bench::NodeController::ReportSeq& nc_reports,
...
-
   result_out << std::endl;
 
   result_out << "Data Timing Stats:" << std::endl;
-
   result_out << std::endl;
 
-  consolidated_latency_stats.pretty_print(result_out, "latency");
-
+  consolidated_latency_stats.pretty_print(result_out, "latency", "  ", 1);
```
simpsont-oci, line 659: Any particular reason these got indented but not the ones above?

performance-tests/bench_2/test_controller/main.cpp
```diff
@@ -35,7 +37,53 @@ using namespace Bench::TestController;
...
+{
+  for (CORBA::ULong i = 0; i < reported_tags.length(); ++i) {
+    const std::string tag(reported_tags[i]);
+    if (input_tags.find(tag) != input_tags.end()) {
+      details[tag] += detail;
+    }
+  }
+}
+
+void consolidate_tagged_stats(std::unordered_map<std::string, Bench::SimpleStatBlock>& stats,
```
simpsont-oci, line 75: Is this a good candidate for the common library? Maybe once we figure out how to handle tag "expressions" or however we want to handle queries involving multiple tags?

simpsont-oci (changes requested):
Overall, looks good. Let's change the options to allow multiple --tag parameters with a single tag name, and then let's create a simple test case with multiple tags and add it to run_test.pl.

performance-tests/bench_2/test_controller/main.cpp
iguessthislldo: You could make use of yard to do something similar as what we're doing for queries.

iguessthislldo: I agree with specifying the count being clunky. I think multiple `--tag` options would be better. Other alternatives that would also work would be something like [what I did for security debug logging tags which are comma separated](https://github.com/objectcomputing/OpenDDS/blob/94fb01271f49be83e38ce08d39ca6d42a4a5815b/dds/DCPS/debug.cpp#L53-L92) or requiring that `--tags` be the last option if it's passed and the rest of the positional arguments are treated as tags. Depending on how often tag are used since `test_controller` has a fixed number of positional arguments you could just treat the rest of the non-option arguments after the scenario as tags: `test_controller TEST_CONTEXT SCENARIO_ID [TAGS...]`.

performance-tests/bench_2/test_controller/main.cpp
sonndinh: It looks to me that the stats info for latency, jitter, round-trip latency, round-trip jitter all fall under the same "Data Timing Stats:" section. Without an additional indentation level for them, it looks like "Data Timing Stats:" is its own section without any data.

performance-tests/bench_2/test_controller/main.cpp
sonndinh: I don't know if there are other places that will use this function. But yes, we can move it if there are.

performance-tests/bench_2/test_controller/main.cpp
```diff
@@ -146,7 +198,17 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
           << "                             incrementing numbers are assigned, but ID doesn't" << std::endl
           << "                             have to be a valid number, just a valid file name." << std::endl
           << "--overwrite-result           Write the result when using --result-id, even if" << std::endl
-          << "                             it would overwrite another result." << std::endl;
+          << "                             it would overwrite another result." << std::endl
+          << "--override-bench-partition-suffix SUFFIX     Override the system-wide partition suffix." << std::endl
```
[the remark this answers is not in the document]
sonndinh: Feel free to fix the text if it doesn't correctly explain what the options are supposed to do.

performance-tests/bench_2/test_controller/main.cpp
sonndinh: I'll change to use multiple "--tag" options for now.

iguessthislldo:
Can you add `--tag` and any documentation it needs to [`docs/internal/bench2.rst`](https://github.com/objectcomputing/OpenDDS/blob/master/docs/internal/bench2.rst)?

sonndinh:
> Can you add `--tag` and any documentation it needs to [`docs/internal/bench2.rst`](https://github.com/objectcomputing/OpenDDS/blob/master/docs/internal/bench2.rst)?

Sure, I'll update it.

docs/internal/bench2.rst
```diff
@@ -741,6 +750,11 @@ This is a subset of the options. Use ``--help`` option to see all the options.
     Overwrite individual worker configs to start their test actions (writes &
     forwards) N seconds from now (absolute time reference)
 
+.. option:: --tag TAG
```
iguessthislldo, line 753:
```suggestion
.. option:: --tag TAG

```

I don't know why but it gets messed up if there isn't a blank line:
Screenshot from 2020-11-13 01-07-21

simpsont-oci (approved):

PR #2038, opened 2020-11-10
Monitor Test Failure due to strong RcHandle

ClaytonCalabrese:
The RcHandle in sporadictask.h was causing failures for the monitor task.  It has been changed to a WeakRcHandle instead, and the functions that use it now have a lock to grab a strong RcHandle.  Also added an error if that lock is not found.

### DIFF
--- a/dds/DCPS/SporadicTask.h
+++ b/dds/DCPS/SporadicTask.h
@@ -32,8 +32,7 @@ public:
     RcHandle<ReactorInterceptor> interceptor = interceptor_.lock();
     if (interceptor) {
       interceptor->execute_or_enqueue(new ScheduleCommand(this, delay));
-    }
-    else {
+    } else {
       ACE_ERROR((LM_ERROR, "(%P|%t) SporadicTask::schedule"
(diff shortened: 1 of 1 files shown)


dds/DCPS/SporadicTask.h
```diff
@@ -29,24 +29,42 @@ class SporadicTask : public RcEventHandler {
 
   void schedule(const TimeDuration& delay)
   {
-    interceptor_->execute_or_enqueue(new ScheduleCommand(this, delay));
+    RcHandle<ReactorInterceptor> interceptor = interceptor_.lock();
+    if (interceptor) {
+      interceptor->execute_or_enqueue(new ScheduleCommand(this, delay));
+    } else {
+      ACE_ERROR((LM_ERROR, "(%P|%t) SporadicTask::schedule"
```
jrw972, line 36: Downgrade these to warnings.  Eliminate the %p since errno will most likely not be a factor.

jrw972 (changes requested):

jrw972 (approved):

PR #2042, opened 2020-11-12
RTPS direct heartbeats unnecessarily

jrw972:
Problem
-------

When a reliable writer writes a sample, all of the reader will now be
lagging.  The writer only needs to send a single non-directed
heartbeat instead of sending a directed heartbeat to each reader.

Solution
--------

Send a non-directed heartbeat for the identified case which is not
applicable to the participant volatile secure writer.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -3646,6 +3646,29 @@ RtpsUdpDataLink::send_heartbeats(const DCPS::MonotonicTimePoint& /*now*/)
 }
 
 void
+RtpsUdpDataLink::RtpsWriter::expire_durable_data(const ReaderInfo_rch& reader,
+                                                 const RtpsUdpInst& cfg,
+                                                 const MonotonicTimePoint& now,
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3645,6 +3645,29 @@ RtpsUdpDataLink::send_heartbeats(const DCPS::MonotonicTimePoint& /*now*/)
...
+  if (!reader->durable_data_.empty()) {
+    const MonotonicTimePoint expiration = reader->durable_timestamp_ + cfg.durable_data_timeout_;
+    if (now > expiration) {
+      typedef OPENDDS_MAP(SequenceNumber, TransportQueueElement*)::iterator dd_iter;
+      for (dd_iter it = reader->durable_data_.begin(); it != reader->durable_data_.end(); ++it) {
+        pendingCallbacks.push_back(it->second);
+      }
+      reader->durable_data_.clear();
+      if (Transport_debug_level > 3) {
+        VDBG_LVL((LM_INFO, "(%P|%t) RtpsUdpDataLink::gather_heartbeats - "
```
mitza-oci, line 3663: Update log with new function name

mitza-oci (changes requested):

PR #2058, opened 2020-11-14
Minor bench2 cleanup

jwillemsen:


performance-tests/bench_2/worker/WorkerTopicListener.h
```diff
@@ -3,24 +3,22 @@
...
 
   void on_inconsistent_topic(DDS::Topic_ptr the_topic, const DDS::InconsistentTopicStatus& status) override;
   void set_topic(Builder::Topic& topic) override;
 
 protected:
-  std::mutex mutex_;
-  Builder::Topic* topic_{0};
-  size_t inconsistent_count_{0};
+  Builder::Topic* topic_{};
+  std::atomic<size_t> inconsistent_count_{0};
```
simpsont-oci, line 21: I think I would prefer this stay a mutex, since the intent was for it to protect everything in the listener class, not just inconsistent_count_. It's a mistake, I think, that the mutex wasn't locked in set_topic(). But for now this is fine since the class is still basically a stub and it doesn't reduce functionality. If/when we add in stats collection for topic listeners we can figure out the best approach for the full implementation.

simpsont-oci:
Also, did you mean to include the non-bench changes in this PR as well as PR #2056 ?

performance-tests/bench_2/worker/WorkerTopicListener.h
jwillemsen: Ok, but the overhead of the mutex is much more as the atomic, why not optimize for the best case in a performance test, at the moment someone adds functionality later he can always change back to a mutex

simpsont-oci (approved):

Issue #2057, opened 2020-11-14
Using two different datatypes for the same topic_name triggers on_inconsistent_topic but also publication/subscribed matched


jwillemsen:
Initial testing shows that when publishing two different data types using the same topic_name does trigger `on_inconsistent_topic` in the publisher and subscriber application but also `on_publication_matched` and `on_subscription_matched` which seems incorrect. 

It looks there is no test in OpenDDS to validate the behaviour when using the same topic_name and two different datatypes

jwillemsen:
Extended existing test to reproduce this, see pull request #2060

jwillemsen:
The extended test as part of #2060 can be used to reproduce this issue, the application code receives incorrect callbacks in case of an inconsistent topic

jwillemsen:
[pubsub.log](https://github.com/objectcomputing/OpenDDS/files/5627582/pubsub.log)

jwillemsen:
Closing as fixed, after the xtypes merge this looks to work correct

PR #2060, opened 2020-11-16
Extend this test with a check of the subscription matched status

jwillemsen:
* tests/DCPS/InconsistentTopic/pubsub.cpp:

See issue #2057


jwillemsen:
Our testing showed that after the x-types merge the AXCIOMA test that uncovered this works without problems. Updated our proposed test extensions after resolving the merge conflicts.

mitza-oci (approved):

Issue #2062, opened 2020-11-16
Problem with find_topic/delete_topic


jwillemsen:
Within DDS4CCM there can be multiple connectors which all use the same topic, these are all deployed separately within the same process. Each connector creates its own DDS entities and for the topic we use `find_topic` to find if the topic is already there within the domain participant. On shutdown each connector now destroys the DDS entities it has created and calls `delete_topic` on the topic returned with `find_topic`. We now see that this fails with `DDS::RETCODE_PRECONDITION_NOT_MET` where it does return `DDS::RETCODE_OK` with RTI Connext DDS. 

For example we deploy 2 connectors, first does a `find_topic` which returns null, so calls `create_topic`, second uses `find_topic`, both on the same domain participant. As each connector is independent of each other we would expect that after connector 2 has deleted datawriter 2 referring to topic 2 the `delete_topic` returns `DDS::RETCODE_OK` because the datawriter 2 that refers to the topic 2 returned by `find_topic` has been deleted. The fact that there is connector 1 that also has used `create_topic` returning topic 1 shouldn't prevent the second connector to cleanup its DDS entities. At the moment now connector 1 does it cleanup by removing its datawriter and topic, we try to delete the domain participant but that will fail because the `delete_topic` for the connector 2 still has to be done.

OpenDDS should return `DDS::RETCODE_OK` at the moment we have a call to `delete_topic` on a topic retrieved by `find_topic` to which no datareaders/datawriters/ContentFilteredTopic/MultiTopi are attached attached anymore

jwillemsen:
The delete_topic is even more problematic than I initially thought. The combination of find_topic/delete_topic can only be cleaned at the moment delete_topic is called multiple times after all attached DDS entities are deleted. At the moment connector 1 calls the delete_topic when connector 2 has still DDS entities attached the delete_topic will return RETCODE_PRECONDITION_NOT_MET and at that moment the reference in the TopicMap will not be decremented with one, the decrement will only happen when there is no entities attached. 

Only at the moment there are no entities attached connector 1 and connector 2 can call delete_topic, first call will decrement the refcount only, second call will really destroy the topic.

jwillemsen:
This doesn't seem to work correctly yet

jwillemsen:
I need to dig more into this, but looks the remaining issue now at our side, we have now two different topic pointers both having the same instance handle.

PR #2067, opened 2020-11-17
Added some logging to delete_topic_i so that when the topic is not re…

jwillemsen:
…moved from the map we can see that in the log, related to issue #2062


jwillemsen:
@mitza-oci Maybe this one also for 3.15?

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -620,10 +618,9 @@ DomainParticipantImpl::delete_topic_i(
         return DDS::RETCODE_ERROR;
       }
 
-      --entry->client_refs_;
+      const CORBA::ULong client_refs = --entry->client_refs_;
 
-      if (remove_objref == true ||
-          0 == entry->client_refs_) {
+      if (remove_objref == true || 0 == client_refs) {
```
mitza-oci, line 623:
```suggestion
      if (remove_objref || 0 == client_refs) {
```

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -650,9 +647,16 @@ DomainParticipantImpl::delete_topic_i(
           }
           return DDS::RETCODE_ERROR;
 
-        } else
+        } else {
           return DDS::RETCODE_OK;
-
+        }
+      } else {
+        if (DCPS_debug_level >= 1) {
```
mitza-oci, line 654: Should this be logged at a higher level?  I guess I'm not familiar with the issue to understand why someone would be interested in this log message.

mitza-oci (changes requested):

dds/DCPS/DomainParticipantImpl.cpp
jwillemsen: We could do that, which level, 4?

iguessthislldo: I've never been quite sure what [it means by "administrative interfaces"](https://github.com/objectcomputing/OpenDDS/blob/c3e21a83403b91ea06e532506932bc9f8339722e/dds/DCPS/debug.h#L28).

mitza-oci: My guess would be things that are in between Entity creation/deletion and per-sample.  So other Entity "controls" like set_qos?

iguessthislldo (approved):

mitza-oci (approved):

mitza-oci:
Merge conflict from #2181

PR #2070, opened 2020-11-17
configure and CI: when using DOCGroup/ACE_TAO from GitHub only the ace6tao2 branch is supported

mitza-oci:


jwillemsen:
Why not have some builds use ACE/TAO/MPC master to make sure everything keeps working? Looks to be also a good test for the work we are doing on ACE/TAO master

mitza-oci:
> 
> 
> Why not have some builds use ACE/TAO/MPC master to make sure everything keeps working? Looks to be also a good test for the work we are doing on ACE/TAO master

We are not going to support an additional ACE/TAO version at this time.  When we have OpenDDS 4 we'll standardize on a newer ACE.

jwillemsen:
> We are not going to support an additional ACE/TAO version at this time. When we have OpenDDS 4 we'll standardize on a newer ACE.

Ok, we will keep our OpenDDS builds on ACE/TAO master, the integration within AXCIOMA will be based on ACE7/TAO3.

mitza-oci:
Our near-term goals are to get 3.15 (master branch) and 3.16 (xtypes branch) done and released.  Having additional ACE versions isn't going to accelerate that process, and we've already seen that CI breaks when using ACE master.  

After the 3.16 we can consider changing how 3.17+ uses ACE.

jwillemsen:
We normally use the Latest_ACETAO_Micro branch for our CI, a stable branch so that the CI for TAOX11/AXCIOMA. We temporarily switched to master because we are migrating some additions back to the main repo and as a result do some cleanup. As soon as possible we switch back to a fixed tag for ACE/TAO/MPC. Maybe something to consider also for OpenDDS, just use a latest tag.

mitza-oci:
> 
> 
> We normally use the Latest_ACETAO_Micro branch for our CI, a stable branch so that the CI for TAOX11/AXCIOMA. We temporarily switched to master because we are migrating some additions back to the main repo and as a result do some cleanup. As soon as possible we switch back to a fixed tag for ACE/TAO/MPC. Maybe something to consider also for OpenDDS, just use a latest tag.

We do already have builds using the `Latest_Micro` ref in the repository.  Maybe only a theoretical problem, but those would have to temporarily "held back" if this starts referring to ACE7.

jwillemsen:
> We do already have builds using the `Latest_Micro` ref in the repository. Maybe only a theoretical problem, but those would have to temporarily "held back" if this starts referring to ACE7.

At that moment we could add a latest ace7tao3 tags so that we keep both branches independent

jwillemsen:
See https://github.com/DOCGroup/ACE_TAO/pull/1298 for a proposed PR to make Latest_ACE7TAO3_ branches/tags

PR #2072, opened 2020-11-17
Make all domain participant logging dependent on the log level

jwillemsen:


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -127,10 +127,12 @@ DomainParticipantImpl::~DomainParticipantImpl()
     DDS::Security::SecurityException se;
     if (!access->return_permissions_handle(perm_handle_, se)) {
       if (DCPS::security_debug.auth_warn) {
-        ACE_ERROR((LM_ERROR,
-                   ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::~DomainParticipantImpl: ")
-                   ACE_TEXT("Unable to return permissions handle. SecurityException[%d.%d]: %C\n"),
-                   se.code, se.minor_code, se.message.in()));
+        if (DCPS_debug_level > 0) {
```
iguessthislldo, line 130: `DCPS::security_debug` are log settings that are separate from `DCPSDebugLevel` like the transport debug level is. Please revert this.

iguessthislldo (changes requested):

dds/DCPS/DomainParticipantImpl.cpp
jwillemsen: Sorry, missed that, reverted

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -194,18 +196,22 @@ DomainParticipantImpl::delete_publisher(
...
   if (!the_servant->is_clean()) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: ")
-               ACE_TEXT("DomainParticipantImpl::delete_publisher, ")
-               ACE_TEXT("The publisher is not empty.\n")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: ")
+                ACE_TEXT("DomainParticipantImpl::delete_publisher, ")
+                ACE_TEXT("The publisher is not empty.\n")));
```
iguessthislldo, line 213:
```suggestion
                 ACE_TEXT("(%P|%t) ERROR: ")
                 ACE_TEXT("DomainParticipantImpl::delete_publisher, ")
                 ACE_TEXT("The publisher is not empty.\n")));
```

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -217,10 +223,12 @@ DomainParticipantImpl::delete_publisher(
...
   if (OpenDDS::DCPS::remove(publishers_, pair) == -1) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::delete_publisher, ")
-               ACE_TEXT("%p\n"),
-               ACE_TEXT("remove")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::delete_publisher, ")
+                ACE_TEXT("%p\n"),
+                ACE_TEXT("remove")));
```
iguessthislldo, line 230:
```suggestion
                 ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::delete_publisher, ")
                 ACE_TEXT("%p\n"),
                 ACE_TEXT("remove")));
```

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -174,10 +174,12 @@ DomainParticipantImpl::create_publisher(
...
   if (OpenDDS::DCPS::insert(publishers_, pair) == -1) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_publisher, ")
-               ACE_TEXT("%p\n"),
-               ACE_TEXT("insert")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_publisher, ")
+                ACE_TEXT("%p\n"),
+                ACE_TEXT("insert")));
```
iguessthislldo, line 181:
```suggestion
                 ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_publisher, ")
                 ACE_TEXT("%p\n"),
                 ACE_TEXT("insert")));
```

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -263,10 +271,12 @@ DomainParticipantImpl::create_subscriber(
...
   if (OpenDDS::DCPS::insert(subscribers_, pair) == -1) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_subscriber, ")
-               ACE_TEXT("%p\n"),
-               ACE_TEXT("insert")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_subscriber, ")
+                ACE_TEXT("%p\n"),
+                ACE_TEXT("insert")));
```
iguessthislldo, line 278:
```suggestion
                 ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::create_subscriber, ")
                 ACE_TEXT("%p\n"),
                 ACE_TEXT("insert")));
```

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -283,27 +293,33 @@ DomainParticipantImpl::delete_subscriber(
...
   if (!the_servant->is_clean()) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: ")
-               ACE_TEXT("DomainParticipantImpl::delete_subscriber, ")
-               ACE_TEXT("The subscriber is not empty.\n")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: ")
+                ACE_TEXT("DomainParticipantImpl::delete_subscriber, ")
+                ACE_TEXT("The subscriber is not empty.\n")));
```
iguessthislldo, line 310:
```suggestion
                 ACE_TEXT("(%P|%t) ERROR: ")
                 ACE_TEXT("DomainParticipantImpl::delete_subscriber, ")
                 ACE_TEXT("The subscriber is not empty.\n")));
```

And I can see there are more after this.

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1251,9 +1296,11 @@ DomainParticipantImpl::ignore_topic(
...
                            dp_id_,
                            ignoreId)) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::ignore_topic, ")
-               ACE_TEXT(" Could not ignore topic.\n")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::ignore_topic, ")
+                ACE_TEXT(" Could not ignore topic.\n")));
+    }
```
iguessthislldo, line 1303: I think the convention is that it's unconditional if the error doesn't have an error return. Is that right @mitza-oci?

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1646,17 +1700,21 @@ DomainParticipantImpl::enable()
...
 #ifdef OPENDDS_SECURITY
   if (TheServiceParticipant->get_security() && !security_config_) {
-    ACE_ERROR((LM_ERROR,
-               ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::enable, ")
-               ACE_TEXT("DCPSSecurity flag is set, but unable to load security plugin configuration.\n")));
+    if (DCPS_debug_level > 0) {
+      ACE_ERROR((LM_ERROR,
+                ACE_TEXT("(%P|%t) ERROR: DomainParticipantImpl::enable, ")
+                ACE_TEXT("DCPSSecurity flag is set, but unable to load security plugin configuration.\n")));
+    }
```
iguessthislldo, line 1717: This one and a lot of the ones below it are security related errors, so they should really be controlled by an appropriate `security_debug` variable.

iguessthislldo (changes requested):

dds/DCPS/DomainParticipantImpl.cpp
jwillemsen: As user I would expect OpenDDS to be silent with logging when there is no logging level set, in order to work around some OpenDDS issues we for example call some operations which could fail because for example some other module already has created the topic, at that moment we take a different path, same in cleanup

dds/DCPS/DomainParticipantImpl.cpp
jwillemsen: Which level to be used? Maybe all security logging needs to be reviewed at some point?

iguessthislldo: Yeah, it's a long term goal of mine to get the security logging and really all logging under tighter control. Looking at the [security logging tags](https://github.com/objectcomputing/OpenDDS/blob/5d1f84624d60f8a989e0d5d68e69c94d34b0c6e4/dds/DCPS/debug.h#L66-L84) it doesn't look like there is an appropriate one in this file, but that's okay since it's intended to be able to add new ones relatively easily. Besides here and debug.h, there are only a few places it needs to be added in debug.cpp. Maybe something like `entity_init`. It could also be two separate ones for the two functions.

jwillemsen: For TAOX11/AXCIOMA we have a new logger that uses ACE internally but uses log modules which can be enabled/disabled independently and uses the insertion operations with optional log formatters. It uses C++11 features so maybe something to look at for OpenDDS4?

That makes it much easier to perform logging, no %c/%d usage anymore, see for example https://github.com/RemedyIT/taox11/tree/master/tao/x11/logger and https://github.com/RemedyIT/taox11/blob/master/tests/log_test/log_test.cpp

jwillemsen: What is now the next best step, for our usage we would like OpenDDS to be silent, do you make a general issue for checking all security logging at a later point in time?

dds/DCPS/DomainParticipantImpl.cpp
mitza-oci: Do we want this case to be `RETCODE_OK`?  If it returned an error it would probably have no need to log (the discovery implementation can log any internal problems).

jwillemsen: Not sure, maybe something to test and change separately?

jwillemsen: Looks I misinterpreted your first comment, now see what you exactly mean

jwillemsen: Shouldn't at the moment the ignore_topic calls on discovery fails the addition to ignored_topics_ be reverted?

jwillemsen: Created new issue #2079 for this specific line

dds/DCPS/DomainParticipantImpl.cpp
iguessthislldo: > It uses C++11 features so maybe something to look at for OpenDDS4?

We were already planning on heading in that general direction.

> What is now the next best step, for our usage we would like OpenDDS to be silent, do you make a general issue for checking all security logging at a later point in time?

I think you are free to make an issue, but it's going to be a while before we can truly address this and make it possible to silence OpenDDS without silencing ACE.

jwillemsen: When we find more problematic areas we will make more PRs so that we can move forward with our effort

iguessthislldo: This isn't resolved yet

jwillemsen: Changing them to use DCPS::security_debug.auth_warn, only in this file, will keep this in mind when I update more later on

jwillemsen:
Updated security related logging to check DCPS::security_debug.auth_warn in this file

iguessthislldo:
> Updated security related logging to check DCPS::security_debug.auth_warn in this file

[`auth_warn`](https://github.com/objectcomputing/OpenDDS/blob/e1cf7869a5f0257159e3ea735a288254164f2d4d/dds/DCPS/debug.h#L70-L72) is for the authentication handshake when the participants associate securely, which is not what is happening here. These logging statements are for creating a Particiant and a Topic, so please create a new flag that called something like `new_entity`.

jwillemsen:
Ok, will do that tomorrow

iguessthislldo:
`new_entity_error` and `new_entity_warn` actually since the participant ones are errors and the topic ones are warnings

iguessthislldo (approved):

PR #2074, opened 2020-11-17
Bench2: Add raw JSON output to Test Controller

simpsont-oci:


iguessthislldo:
Does it have to take an id separate from `--result-id`? I think it would be better to reuse `result_id` for the filename with a ".json" extension tacked on.

simpsont-oci:
> Does it have to take an id separate from `--result-id`? I think it would be better to reuse `result_id` for the filename with a ".json" extension tacked on.

I was waffling on this, but wondered if someone might want to name them distinctly. After implementing it this way, I was annoyed that I had to keep coming up with new json file names or add --overwrite-result. I'll change it.

iguessthislldo:
Interesting. Maybe `--overwrite-result` should be made the default? I imagined the test results as mostly immutable, but maybe that's not really important as I thought it was.

mitza-oci (approved):

PR #2077, opened 2020-11-18
Bench2: Add ability to specify node for allocation

sonndinh:


iguessthislldo:
If this uses the `--name` option of node controller, it would probably be good to change the documentation of that option since it says it's just for display purposes and that it has no other effect.

sonndinh:
> If this uses the `--name` option of node controller, it would probably be good to change the documentation of that option since it says it's just for display purposes and that it has no other effect.

Sure, will do that.

performance-tests/bench_2/test_controller/ScenarioManager.cpp
```diff
@@ -201,107 +202,254 @@ void ScenarioManager::customize_configs(std::map<std::string, std::string>& work
...
-  for (unsigned protoworker_i = 0; protoworker_i < scenario_prototype.any_node.length(); protoworker_i++) {
-    maximum_nodes += scenario_prototype.any_node[protoworker_i].count;
   }
 
-  // Get Minimum Node Count
-  unsigned minimum_nodes = 0;
-  minimum_nodes = exclusive_nodes;
-  if (maximum_nodes - exclusive_nodes) { // If there are any non-exclusive node configs
-    minimum_nodes++; // Then there just has to be at least one node for them
+  struct NcMetaData {
```
simpsont-oci, line 234: Let's move this and the queue typedef into a new header.

performance-tests/bench_2/test_controller/ScenarioManager.cpp
```diff
@@ -201,107 +202,254 @@ void ScenarioManager::customize_configs(std::map<std::string, std::string>& work
...
+    // Maximum probabilty that it is used for an exclusive node
+    double max_prob;
+
+    // Number of node prototypes that can assign a node to it
+    unsigned overlap_count;
+
+    // Whether it is occupied by an exclusive node
+    bool exclusive_occupied;
+
+    bool operator()(const NcMetaData& a, const NcMetaData& b) const
```
simpsont-oci, line 252: Interesting, this puts the lowest probability with lowest overlap count at the top? Am I reading that right? Maybe a comment here to explain the goal of using this as the sort operator (rather than std::less & operator<).

performance-tests/bench_2/test_controller/ScenarioManager.cpp
```diff
@@ -201,107 +202,254 @@ void ScenarioManager::customize_configs(std::map<std::string, std::string>& work
...
+  allocated_scenario.configs.length(0);
+
+  // Node controllers allocated to exclusive nodes
+  std::set<NodeController::NodeId, OpenDDS::DCPS::GUID_tKeyLessThan> exclusive_ncs;
+
+  // Allocate exclusive node configs first
+  for (unsigned i = 0; i < scenario_prototype.nodes.length(); ++i) {
+    const NodePrototype& nodeproto = scenario_prototype.nodes[i];
+    if (nodeproto.exclusive && nodeproto.count > 0) {
+      // Node controllers that are less likely to be used by some other node prototypes
```
simpsont-oci, line 310: Ah, here's the comment I was looking for earlier. Makes sense. Maybe this gets moved (or copied) into the new header.

performance-tests/bench_2/test_controller/ScenarioManager.cpp
```diff
@@ -201,107 +202,254 @@ void ScenarioManager::customize_configs(std::map<std::string, std::string>& work
...
+{
+  if (pat.empty() ||
+      (!OpenDDS::DCPS::is_wildcard(pat.c_str()) && pat == str) ||
+      (OpenDDS::DCPS::is_wildcard(pat.c_str()) && ACE::wild_match(str.c_str(), pat.c_str(), true, true))) {
+    return true;
+  }
+  return false;
+}
+
+AllocatedScenario ScenarioManager::allocate_scenario(const ScenarioPrototype& scenario_prototype,
```
simpsont-oci, line 215: I think this is complicated enough that it needs a unit test. For now, let's just make a test app local to the test_controller directory. Feel free to use googletest and make it a required library, but you'll need to update the mpc accordingly. Some of the security tests use googletest already, if you're looking for examples of how we do that.

We should test for a few basic scenarios, including failures, and then maybe one or two more complicated scenarios.

simpsont-oci: If it makes sense to separate the choice of node_controller for each worker config out from the reading in of worker config files that happens in this function in order to make testing easier, I'm fine with that.

simpsont-oci (changes requested):

performance-tests/bench_2/test_controller/ScenarioManager.cpp
sonndinh: That's correct. I'll add a comment for it.

performance-tests/bench_2/test_controller/ScenarioManager.cpp
sonndinh: Should it be moved to a header? It just contains extra data needed to compute the allocation, and I don't think it'll be used somewhere else.

simpsont-oci: In general, I prefer not to see local structs and this function's already pretty large. Perhaps you can pull some of this logic into a separate helper class (AllocationHelper, etc) that's easier to create a test against, and this struct can be part of that helper class?

performance-tests/bench_2/test_controller/unit_tests/ScenarioAllocationTest.cpp
```diff
@@ -0,0 +1,129 @@
+#include "../ScenarioManager.h"
```
simpsont-oci, line 1: Is there an mpc for this?

sonndinh: Yes, I'm adding one. Will push soon.

performance-tests/bench_2/test_controller/DdsEntities.h
```diff
@@ -13,6 +13,8 @@
 class DdsEntities {
 public:
   DdsEntities(DDS::DomainParticipantFactory_var dpf, int domain);
+  DdsEntities() : dpf_{nullptr}, participant_{nullptr}, status_reader_{nullptr} {}
```
jwillemsen, line 16: No need to pass nullptr, {} will initialise a pointer with nullptr

performance-tests/bench_2/test_controller/unit_tests/unit_tests.mpc
```diff
@@ -0,0 +1,17 @@
+project (*ScenarioAllocationTest): dcps, dcpsexe, dcps_test, dcps_rapidjson, googletest {
```
simpsont-oci, line 1: Something's still not quite right here, some of the VS2017 and VS2019 debug builds are failing. Take a look at the bench_builder_exe and bench_exe mpb files in performance-tests/bench_2 for hints on what might be missing?

simpsont-oci (changes requested):

simpsont-oci (approved):

PR #2084, opened 2020-11-20
Fix memory leaks detected in the LivelinessTimeout test

lij-oci:
added historic flag in write_participant_message

### DIFF
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -3191,7 +3191,7 @@ Sedp::Writer::write_participant_message(const ParticipantMessageData& pmd,
             (ser << pmd);
 
   if (ok) {
-      send_sample(payload, size, reader, sequence);
+    send_sample(payload, size + padding, reader, sequence, reader != GUID_UNKNOWN);
 
   } else {
       result = DDS::RETCODE_ERROR;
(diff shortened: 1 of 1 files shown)


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -3191,7 +3191,7 @@ Sedp::Writer::write_participant_message(const ParticipantMessageData& pmd,
             (ser << pmd);
 
   if (ok) {
-      send_sample(payload, size, reader, sequence);
+    send_sample(payload, size + padding, reader, sequence, reader != GUID_UNKNOWN);
```
mitza-oci, line 3194: If this uses `size+padding`, does line 3184 need to change as well?

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
simpsont-oci: I think we're potentially banking on the fact that there's no padding in this message (is that true)? But yeah, either change it both places or don't change it in either place.

lij-oci: Actually, there are 4 places like this, plus 3 places of new ACE_Message_Block(size). They were all changed to size + padding in my original draft PR#1952, together with DCPS::Message_Block_Ptr replacing DCPS::Message_Block and separate delete statements. Should I bring them in this PR?

lij-oci: In write_stateless_message(), line 3221: new ACE_Message_Block(size + padding)); but line 3232: send_sample(payload, size, reader, sequence);

mitza-oci (approved):

simpsont-oci (approved):

PR #2085, opened 2020-11-20
Failed To Join Multicast (IPv6)

ClaytonCalabrese:
Proposed fix for failing to join multicast due to simultaneous calls.  This is occurring in windows 7 and remedy fedora builds.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -2942,25 +2942,29 @@ Spdp::SpdpTransport::join_multicast_group(const DCPS::NetworkInterface& nic,
                  all_interfaces ? "all interfaces" : nic.name().c_str()));
     }
 
-    if (0 == multicast_socket_.join(multicast_address_, 1, all_interfaces ? 0 : ACE_TEXT_CHAR_TO_TCHAR(nic.name().c_str()))) {
-      joined_interfaces_.insert(nic.name());
+    static ACE_Thread_Mutex ipv4_static_lock;
+    {
+      ACE_GUARD(ACE_Thread_mutex, g2, ipv4_static_lock);
+      if (0 == multicast_socket_.join(multicast_address_, 1, all_interfaces ? 0 : ACE_TEXT_CHAR_TO_TCHAR(nic.name().c_str()))) {
+        joined_interfaces_.insert(nic.name());
 
-      if (reactor()->register_handler(multicast_socket_.get_handle(),
-                                    this, ACE_Event_Handler::READ_MASK) != 0) {
-        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) Spdp::SpdpTransport::join_multicast_group failed to register multicast input handler\n")));
-        return;
+        if (reactor()->register_handler(multicast_socket_.get_handle(),
+                                      this, ACE_Event_Handler::READ_MASK) != 0) {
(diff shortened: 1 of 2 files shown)


dds/DCPS/RTPS/Spdp.h
```diff
@@ -339,7 +339,15 @@ class OpenDDS_Rtps_Export Spdp : public DCPS::LocalParticipant<Sedp>
     void execute()
     {
       if (action_ == CMG_JOIN) {
-        tport_->join_multicast_group(nic_, all_interfaces_);
+        bool success = false;
+        for (int count = 0; count <= 1 && !success; ++count) {
+          success = tport_->join_multicast_group(nic_, all_interfaces_);
```
simpsont-oci, line 344: This just calls the join again? Does this actually fix the broken behavior, or are we crossing our fingers here and hoping the second time behaves better than the first? At the best, this seems it would reduce a N% failure rate to a SQRT(N)% failure rate.

It would be good to understand why the failure happened in these cases so we can actually prevent it if possible. Have you been able to catch the error in a debugger? Was the interface name wrong? Does this version of Windows / Linux need a special flag or system call to prevent interruption? Or perhaps we need to check for a special return status in order to know when to try again?

ClaytonCalabrese: The problem is that you have multiple transports on different threads hitting the join at the same time.  As far as I can tell this happens only in windows 7, but it might be related to the failed to join multicast group issue also appearing in fedora.  

Barring a lock which would only be useful for windows 7, while causing a slowdown for other OS, Adam and I agreed this was probably the least impactful solution.

ClaytonCalabrese:
Finally got the failure for ipv4 on mosquito replicated, these changes do NOT fix the ipv4 multicast issues.  I'm looking into a solution now.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2942,25 +2942,29 @@ Spdp::SpdpTransport::join_multicast_group(const DCPS::NetworkInterface& nic,
                  all_interfaces ? "all interfaces" : nic.name().c_str()));
     }
 
-    if (0 == multicast_socket_.join(multicast_address_, 1, all_interfaces ? 0 : ACE_TEXT_CHAR_TO_TCHAR(nic.name().c_str()))) {
-      joined_interfaces_.insert(nic.name());
+    static ACE_Thread_Mutex ipv4_static_lock;
+    {
```
mitza-oci, line 2946: Does this require an "artificial" scope here?  (It's hard to tell from the diffs on GitHub, if the code would be equivalent without it then leave it out.)

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2942,25 +2942,29 @@ Spdp::SpdpTransport::join_multicast_group(const DCPS::NetworkInterface& nic,
                  all_interfaces ? "all interfaces" : nic.name().c_str()));
     }
 
-    if (0 == multicast_socket_.join(multicast_address_, 1, all_interfaces ? 0 : ACE_TEXT_CHAR_TO_TCHAR(nic.name().c_str()))) {
-      joined_interfaces_.insert(nic.name());
+    static ACE_Thread_Mutex ipv4_static_lock;
```
mitza-oci, line 2945: This is an unusual use, so add a comment describing why we want `static` (if we do?)

mitza-oci (changes requested):

ClaytonCalabrese:
The IPv4 failed_to_join multicast is a separate issue.  This PR is ready for review.

mitza-oci (approved):

PR #2088, opened 2020-11-22
Thrasher test lacks durability scenarios

jrw972:
Problem
--------

The Thrasher tests create T reliable writers in T participants and 1
reliable subscriber.  The writers write N messages and the reader
checks that all messages are received.  The test stresses
discovery (especially RTPS) and reliability including features like
waiting for acknowledgements and waiting for match status.  Durability
is implicitly tested through discovery.  The test could be used to
diagnose problems with durability if it were extended.

Solution
--------

Add flags to enable durability.  When the writer is durable, it either
1) waits for a match and then writes data or 2) writes data and then
waits for a match.

### DIFF
--- a/bin/dcps_tests.lst
+++ b/bin/dcps_tests.lst
@@ -294,6 +294,22 @@ tests/DCPS/Thrasher/run_test.pl low rtps: !DCPS_MIN RTPS !LYNXOS
 tests/DCPS/Thrasher/run_test.pl medium rtps: !DCPS_MIN RTPS !LYNXOS
 tests/DCPS/Thrasher/run_test.pl high rtps: !DCPS_MIN RTPS !LYNXOS
 tests/DCPS/Thrasher/run_test.pl aggressive rtps: !DCPS_MIN RTPS !LYNXOS
+tests/DCPS/Thrasher/run_test.pl single durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl double durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl triangle durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl default durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl low durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl medium durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl high durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl aggressive durable: !DCPS_MIN !LYNXOS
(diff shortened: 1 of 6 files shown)


bin/dcps_tests.lst
```diff
@@ -294,6 +294,22 @@ tests/DCPS/Thrasher/run_test.pl low rtps: !DCPS_MIN RTPS !LYNXOS
...
 tests/DCPS/Thrasher/run_test.pl high rtps: !DCPS_MIN RTPS !LYNXOS
 tests/DCPS/Thrasher/run_test.pl aggressive rtps: !DCPS_MIN RTPS !LYNXOS
+tests/DCPS/Thrasher/run_test.pl single durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl double durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl triangle durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl default durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl low durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl medium durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl high durable: !DCPS_MIN !LYNXOS
+tests/DCPS/Thrasher/run_test.pl aggressive durable: !DCPS_MIN !LYNXOS
```
mitza-oci, line 304: The equivalent lines on master have changed a bit since you made this PR.  The tests that are not `rtps` have an additional ` !OPENDDS_SAFETY_PROFILE` here.

tests/DCPS/Thrasher/ParticipantTask.h
```diff
@@ -13,7 +13,8 @@
 class ParticipantTask : public ACE_Task_Base
 {
 public:
-  explicit ParticipantTask(const std::size_t& samples_per_thread);
+  explicit ParticipantTask(std::size_t samples_per_thread,
```
mitza-oci, line 16:
```suggestion
  ParticipantTask(std::size_t samples_per_thread,
```

tests/DCPS/Thrasher/ParticipantTask.cpp
```diff
@@ -156,12 +160,16 @@ ParticipantTask::svc()
     OpenDDS::DCPS::DataWriterImpl* impl =
       dynamic_cast<OpenDDS::DCPS::DataWriterImpl*>(writer.in());
 
+    const bool wait_early = durable_ ? rand() % 2 == 0 : true;
```
mitza-oci, line 163: Using `rand()` here may be more trouble than it's worth.  If we go with `rand()`, log the seed value used and provide a way to override it in order to have repeatable tests.

mitza-oci (changes requested):

mitza-oci (approved):

PR #2106, opened 2020-11-24
Configure QueueTaskBase ACE_Condition to use monotonic time.

jonesc-oci:
This fixes an internal thread status monitoring high cpu usage issue.


iguessthislldo:
@mitza-oci Would it be possible for us to wrap/inherit `ACE_Condition` to make this easier in the future?

mitza-oci:
> @mitza-oci Would it be possible for us to wrap/inherit `ACE_Condition` to make this easier in the future?

Sure, why not?

mitza-oci (approved):

PR #2111, opened 2020-11-30
Fix for AccessControl test read access violation

pollocktoci:
Removed call to cancel_and_wait from RevokePermissionsTask dtor


mitza-oci (dismissed):

mitza-oci:
Similar to the prior PR (since the code is the same): while this change may be the right thing to do, it also crashes the unit test.  @jrw972

mitza-oci:
Replaced by #2145

PR #2114, opened 2020-11-30
XTypes Branch: Wrap and Deprecate ACE_Condition

iguessthislldo:
Prompted by https://github.com/objectcomputing/OpenDDS/pull/2106. Do a similar thing to `ACE_Condition` as what was done to `ACE_Time_Value` in https://github.com/objectcomputing/OpenDDS/pull/1256, which is to wrap and deprecate it. In wrapping it we can force monotonic time behavior and make it so it only accepts `MonotonicTimePoint` and `TimeDuration`.

Fixed include order along the way, including a pretty bad case in the `RecorderImpl` files where the header depended on the include order in the source file.

### DIFF
--- a/dds/DCPS/Condition.h
+++ b/dds/DCPS/Condition.h
@@ -1,0 +1,96 @@
+#ifndef OPENDDS_DCPS_CONDITION_H
+#define OPENDDS_DCPS_CONDITION_H
+
+#include "TimeTypes.h"
+
+#include <ace/Condition_Thread_Mutex.h>
+#include <ace/Condition_Recursive_Thread_Mutex.h>
+#include <ace/Condition_T.h>
+#include <ace/Condition_Attributes.h>
+#include <ace/OS_NS_Thread.h>
+
+#ifndef ACE_LACKS_PRAGMA_ONCE
+#  pragma once
+#endif
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+/**
+ * ACE_Condition wrapper based on std::condition_variable that enforces
+ * monotonic time behavior.
+ *
+ * Besides the fact that it only works with ACE Mutexes, the major difference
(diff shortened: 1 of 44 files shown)


dds/DCPS/Condition.h
```diff
@@ -0,0 +1,96 @@
...
+    Timeout,
+    WaitError
+  };
+
+  WaitStatus wait()
+  {
+    if (impl_.wait() == 0) {
+      return NoTimeout;
+    }
+    ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Condition::wait: %p\n"));
```
jwillemsen, line 49: Logging based on log level?

tests/DCPS/Deadline/DataWriterListenerImpl.h
```diff
@@ -55,7 +59,7 @@ class DataWriterListenerImpl
 
 private:
   mutable ACE_Thread_Mutex mutex_;
-  mutable ACE_Condition<ACE_Thread_Mutex> matched_condition_;
+  mutable  matched_condition_;
```
mitza-oci, line 62: What type is this supposed to have?

mitza-oci (changes requested):

dds/DCPS/MessageTracker.cpp
```diff
@@ -72,39 +71,50 @@ void MessageTracker::wait_messages_pending(const char* caller)
...
+    case CvStatus_NoTimeout:
+      break;
+
+    case CvStatus_Error:
+    default:
+      if (DCPS_debug_level) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: MessageTracker::wait_messages_pending: "
+          "error in wait_until\n"));
+      }
+      loop = false;
```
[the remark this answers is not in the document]
iguessthislldo: It should keep trying even if it got an error?

mitza-oci:
No, the line after it is a "return" so this is a dead store

mitza-oci (approved):

PR #2118, opened 2020-12-02
A Build & Test Workflow for GitHub Actions

simpsont-oci:


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,899 @@
...
+        report_paths: OpenDDS/workspace/Tests_JUnit.xml
+
+  # Windows
+
+  windows-2019_defaults_artifact:
+
+    runs-on: windows-2019
+
+    steps:
+    - name: install openssl-windows & xerces-c
```
jwillemsen, line 314: Would recommend to use run-vcpkg, that uses caching, as far as I know the default vcpkg doesn't do that, see https://github.com/DOCGroup/ACE_TAO/blob/a72619e2ffcb37bc821c76b1e30b6886c2664003/.github/workflows/windows.yml#L99

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,899 @@
...
+    - name: configure OpenDDS
+      shell: cmd
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        cd OpenDDS
+        configure --no-tests --security --ace=%GITHUB_WORKSPACE%/ACE_TAO/ACE --tao=%GITHUB_WORKSPACE%/ACE_TAO/TAO --mpc=%GITHUB_WORKSPACE%/MPC --xerces3=C:\vcpkg\installed\x64-windows --openssl=C:\vcpkg\installed\x64-windows
+    - name: build OpenDDS
+      shell: cmd
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
jwillemsen, line 359: Easier is msvc-dev-cmd, see https://github.com/RemedyIT/taox11/blob/bbf9ec8d864aae68e77d9107a84cbd02f974a054/.github/workflows/windows.yml#L57

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,899 @@
...
+    - name: check build configuration
+      shell: cmd
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        cd OpenDDS
+        call setenv.cmd
+        tools\scripts\show_build_config.pl
+    - name: build OpenDDS
+      shell: cmd
+      run: |
```
jwillemsen, line 549: Or use setup-msbuild, see https://github.com/DOCGroup/ACE_TAO/blob/a72619e2ffcb37bc821c76b1e30b6886c2664003/.github/workflows/windows.yml#L131

jwillemsen:
Also MinGW64 can now be used on github actions, see https://github.com/DOCGroup/ACE_TAO/blob/c62ecd5955c1350fa25f95deb9bce2c81bc70f1c/.github/workflows/windows.yml#L137

mitza-oci:
> 
> 
> Also MinGW64 can now be used on github actions, see https://github.com/DOCGroup/ACE_TAO/blob/c62ecd5955c1350fa25f95deb9bce2c81bc70f1c/.github/workflows/windows.yml#L137

MinGW (last time we checked) doesn't work.  So I suggest we not expand platform support as part of this effort.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,908 @@
...
+      uses: mikepenz/action-junit-report@v1
+      if: always()
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        check_name: "Test Results: macos-10_15_security"
+        report_paths: OpenDDS/workspace/Tests_JUnit.xml
+
+  macos-10_15_java_no-bit:
+
+    runs-on: macos-10.15
```
jwillemsen, line 780: There is also a macos-11.0 environment, see https://github.com/actions/virtual-environments#available-environments

.github/workflows/build_and_test.yml
simpsont-oci: I've tried it, and it looks like it had a cache miss for the case when the cache key was identical, so it really didn't change the overall time by much. Perhaps I just need to run the workflow again with the same cache key, since it looks like updating the cache is sometimes post-build step?

I am, however, glad that I read the documentation for actions/cache, which is probably what we want to be using rather than artifacts to copy ACE/TAO files between jobs in the workflow. This is especially true if it will let you pull cached samples between workflow runs and lets you pull the cached values from master/main, which is how I understand the documentation. So I'll be looking into using that to avoid having to build the artifacts at all as part of normal pull request runs, which was part of the motivation for setting it up this way.

jwillemsen: First time run-vcpkg it makes the cache, in next runs it saves time, see the ACE-TAO actions

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,907 @@
+name: "Build & Test"
+
+on:
+  push:
+    branches:
+      - master
+      - branch-DDS-3.15
+      - branch-DDS-3.14
+      - branch-DDS-3.13
+      - gh_workflow
```
iguessthislldo, line 10:
```suggestion
      - gh_workflow
    paths:
      - '**'
      - '!docs/**'
      - '!README.md'
      - '!INSTALL.md'
      - '!NEWS.md'
```

This makes it not run the workflow if those are the only files that changed. We can add more if needed.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,907 @@
...
+  pull_request:
+    branches:
+      - master
+      - branch-DDS-3.15
+      - branch-DDS-3.14
+      - branch-DDS-3.13
+      - gh_workflow
```
iguessthislldo, line 17:
```suggestion
      - gh_workflow
    paths:
      - '**'
      - '!docs/**'
      - '!README.md'
      - '!INSTALL.md'
      - '!NEWS.md'
```

.github/workflows/build_and_test.yml
```diff
@@ -1538,7 +1544,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         repository: DOCGroup/ACE_TAO
-        ref: 61338c7 # Update to 6_5_13 when available
+        ref: 9eae9c11f1bcd07b48c74820d8cb4c67aca9ef23 # Update to 6_5_13 when available
```
iguessthislldo, line 1541:
```suggestion
        ref: ace6tao2
```

jwillemsen: Or the  ACE+TAO-6_5_12 tag?

simpsont-oci:
OSX 11.0 didn't exist yet.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,2093 @@
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        check_name: "Test Results: ubuntu-16_04_java_no-bit"
+        report_paths: OpenDDS/workspace/Tests_JUnit.xml
+
+  # Windows 2019
+
+  windows-2019_defaults_artifact:
+
+    runs-on: windows-2019
```
jwillemsen, line 895: why duplicate everything for windows-2016 or windows-2019, using a matrix there is less duplication, see for example https://github.com/RemedyIT/taox11/blob/454f313787885b345a26a3960bcb49abb76ec432/.github/workflows/windows.yml#L27

.github/workflows/build_and_test.yml
simpsont-oci: OSX 11.0 support didn't exist for these.

.github/workflows/build_and_test.yml
```diff
@@ -78,7 +78,7 @@ jobs:
     - name: upload ACE_TAO artifact
       uses: actions/upload-artifact@v2
       with:
-        name: ubuntu_ace_artifact
+        name: ubuntu-20_04_defaults_ace_artifact
```
jwillemsen, line 81: Will it not work when you use the job information, something like {{ job.name }}, would make it much easier for maintenance. You can also refer to the properties another job, see https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname.

dds/InfoRepo/DCPSInfoRepo.mpc
```diff
@@ -1,4 +1,4 @@
-project(*Lib): dcpslib, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
+project(*Lib): dcpslib, dcps_tcp, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
```
mitza-oci, line 1: Is this change necessary?  I thought it was already covered by `DCPSInfoRepo_bits`.

simpsont-oci: Hmm, I thought I reverted these...

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,2073 @@
+name: "Build & Test"
+
+on:
+  push:
+    branches:
+      - master
+      - branch-DDS-3.15
+      - branch-DDS-3.14
+      - branch-DDS-3.13
```
mitza-oci, line 9: Does it allow wildcard expressions?  We don't necessarily need to do it now but maybe next iteration of this script could use them.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,2073 @@
...
+    - name: checkout OpenDDS
+      uses: actions/checkout@v2
+      with:
+        path: OpenDDS
+        submodules: true
+    - name: checkout ACE_TAO
+      uses: actions/checkout@v2
+      with:
+        repository: DOCGroup/ACE_TAO
+        ref: ACE+TAO-6_5_12
```
mitza-oci, line 68:
```suggestion
        ref: ace6tao2
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #2121, opened 2020-12-02
Migrate Travis CI configurations to Azure Pipelines

pollocktoci:


azure-pipelines.yml
```diff
@@ -247,9 +249,17 @@ jobs:
       GCC6:
         ConfigOpts: --compiler=g++-6 --xerces --features=uses_wchar=1 --no-debug
         PackageDeps: libxerces-c-dev g++-6
+      GCC8:
+        ConfigOpts: --compiler=g++-8 --xerces --no-inline --java=$(JAVA_HOME_12_X64)
+        TEST_JAVA_MESSENGER: true
```
simpsont-oci, line 254: We should probably run the JAVA messenger tests for the other JAVA builds as well (Debug & Release)

simpsont-oci (changes requested):

azure-pipelines.yml
```diff
@@ -247,9 +249,17 @@ jobs:
...
+        PackageDeps: libxerces-c-dev g++-8
       GCC10:
         ConfigOpts: --compiler=g++-10 --no-inline
         PackageDeps: g++-10
+      GCC10_Xerces:
+        ConfigOpts: --compiler=g++-10 --xerces --no-inline
+        TEST_FACE_MESSENGER: true
```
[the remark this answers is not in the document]
pollocktoci: Done.

azure-pipelines.yml
pollocktoci: Done.

simpsont-oci (approved):

mitza-oci (changes requested):
Please add the check for `.gitignore` from Travis CI

mitza-oci (changes requested):
This adds 2 new build jobs: GCC8, and GCC10_Xerces.  Since the builds already take way too long, I'd like to avoid adding new jobs and just try to cover the same feature set with existing jobs.

pollocktoci:
I eliminated the GCC10_Xerces job, merging it into the GCC10 job.  I kept the GCC8 job, since where was not other job that used g++ 8.

azure-pipelines.yml
```diff
@@ -106,6 +106,9 @@ jobs:
       cd DevGuideExamples\DCPS\Messenger
       perl integration_run_test.pl -ExeSubDir %SUBDIR_FOR_MESSENGER%
     displayName: Run Examples DCPS Messenger
+  - script: |
+      ! git clean -nd -e ext | cut -d ' ' -f 3 | xargs -I {} echo "##vso[task.logissue type=error]This file should be gitignored: '{}'" | exit 1
+    displayName: Check validity of .gitignore
```
mitza-oci, line 111: This doesn't necessarily need to run on the Windows builds.  Just having it on some builds will be helpful.  Ideally those would include optional features like Java and DDS Security.

mitza-oci (approved):

PR #2130, opened 2020-12-04
XTypes: Fix Deadlock in Type Lookup Service

iguessthislldo:


iguessthislldo:
@mitza-oci @jrw972 Is there a way to assert the ACE lock is held like [std::unique_lock<Mutex>::owns_lock](https://en.cppreference.com/w/cpp/thread/unique_lock/owns_lock)?

kuznetsovmoci (approved):

mitza-oci:
> 
> 
> @mitza-oci @jrw972 Is there a way to assert the ACE lock is held like [std::unique_lock::owns_lock](https://en.cppreference.com/w/cpp/thread/unique_lock/owns_lock)?

`ACE_Guard::locked` ?

iguessthislldo:
The guard I would be interested is in another function way down in the stack. I might also be thinking about this wrong.

mitza-oci:
> The guard I would be interested is in another function way down in the stack. I might also be thinking about this wrong.

`ACE_Thread_Mutex::tryacquire`

mitza-oci (approved):

PR #2132, opened 2020-12-05
Durable RTPS writers may send gaps

jrw972:
Problem
-------

The `send_and_gather_nack_replies` function attempts to not
send (undirected) gaps if a durable reader is detected.  This is
flawed in two ways.  First, a durable writer should never send an
undirected gap because it cannot know when a potential reader will
come online.  Second, the existing logic only checks readers that need
a reply instead of all of the readers when making the determination.

Solution
--------

Re-write `send_and_gather_nack_replies` so that naks for directed data
receive a directed reply (gap or data) and naks for undirected data
receive receive an undirected reply.  Directed data includes durable
resends and messages from the participant volatile secure writer.  The
code for responding to naks for durable data will/should not be
executed until the logic that processes acknacks is updated.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -3044,7 +3044,6 @@ RtpsUdpDataLink::RtpsWriter::send_and_gather_nack_replies(MetaSubmessageVec& met
           // Non-directed gap.
           consolidated_recipients.insert(addrs.begin(), addrs.end());
           consolidated_gaps.insert(seq);
-          continue;
         }
       }
     }
@@ -3061,6 +3060,7 @@ RtpsUdpDataLink::RtpsWriter::send_and_gather_nack_replies(MetaSubmessageVec& met
           consolidated_recipients.insert(addrs.begin(), addrs.end());
           consolidated_fragment_requests[seq].insert(rf->second.bitmapBase.value, rf->second.numBits,
                                                      rf->second.bitmap.get_buffer());
+          continue;
         } else if (destination != reader->id_){
           // Directed at another reader.
           gaps.insert(seq);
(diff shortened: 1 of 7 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3048,78 +3048,131 @@ RtpsUdpDataLink::RtpsWriter::send_and_gather_nack_replies(MetaSubmessageVec& met
...
-                     "resend data %q-%q\n", ranges[i].first.getValue(),
-                     ranges[i].second.getValue()));
-        }
-        sb.resend_i(ranges[i], &gaps);
-      }
+  // Send the consolidated requests.
+  const OPENDDS_VECTOR(SequenceRange) ranges = consolidated_requests.present_sequence_ranges();
+  ACE_GUARD(TransportSendBuffer::LockType, guard, send_buff_->strategy_lock());
+  const RtpsUdpSendStrategy::OverrideToken ot =
+    link->send_strategy()->override_destinations(consolidated_recipients);
```
mitza-oci, line 3155: Can the scope of both of these objects be reduced to only over the `resend_i` phase?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3048,78 +3048,131 @@ RtpsUdpDataLink::RtpsWriter::send_and_gather_nack_replies(MetaSubmessageVec& met
...
+  for (OPENDDS_VECTOR(SequenceRange)::const_iterator pos = ranges.begin(), limit = ranges.end();
+       pos !=limit; ++pos) {
```
mitza-oci, line 3157:
```suggestion
       pos != limit; ++pos) {
```

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I think I resolved this.  There is still one loop where the lock and token are outside of the loop.  I'm not sure there is an advantage moving them inside the loop.

mitza-oci (approved):

PR #2133, opened 2020-12-05
Add missing virtual

jwillemsen:
* dds/DCPS/TopicImpl.h:

### DIFF
--- a/dds/DCPS/TopicImpl.h
+++ b/dds/DCPS/TopicImpl.h
@@ -83,7 +83,7 @@ public:
 
   virtual void transport_config(const TransportConfig_rch& cfg);
 
-  void inconsistent_topic(int count);
+  virtual void inconsistent_topic(int count);
 
 private:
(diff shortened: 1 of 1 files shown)


iguessthislldo (commented):
It's not "missing". It's not necessary in inherited virtual methods.

jwillemsen:
But it will make it clear they are virtual, with C++11 I would have added override

jwillemsen:
All other overridden methods in this class do use virtual

sonndinh:
> All other overridden methods in this class do use virtual

I don't think adding `virtual` makes it clear that the function is overridden because a `virtual` member function doesn't mean it overrides a member from its base classes. Maybe we could go the other way, that is remove `virtual` from functions that won't be overridden.

mitza-oci:
I guess it's not explicitly in the style guide, but we have not been adding `virtual` to these.

PR #2143, opened 2020-12-08
Do full cleanup for RtpsReaders in RtpsUdpDatalink::client_stop()

simpsont-oci:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -762,7 +762,37 @@ void RtpsUdpDataLink::client_stop(const RepoId& localId)
...
+
+    } else {
+      for (WriterToSeqReadersMap::iterator w = writer_to_seq_best_effort_readers_.begin();
+          w != writer_to_seq_best_effort_readers_.end(); ++w) {
+        if (w != writer_to_seq_best_effort_readers_.end()) {
+          RepoIdSet::iterator r = w->second.readers.find(localId);
+          if (r != w->second.readers.end()) {
+            w->second.readers.erase(r);
+            if (w->second.readers.empty()) {
+              writer_to_seq_best_effort_readers_.erase(w);
```
simpsont-oci, line 790: Hmm, I copied and modified this from release_reservations, where it was a single iterator... this should probably use the post-increment trick from above as well when it erases.

mitza-oci (approved):

PR #2147, opened 2020-12-09
ManualAssertLiveliness test memory leak fix

pollocktoci:
This may not fix the memory leak associated with the ManualAssertLiveliness test, but it does resolve (at least when testing locally) an issue with the test shutting down before all of the callbacks have arrived.


pollocktoci:
No longer seeing this leak.  Closing.

PR #2155, opened 2020-12-10
Log the instance_state as readable string which says something for a …

jwillemsen:
…user

    * tests/DCPS/Messenger/DataReaderListener.cpp:


iguessthislldo:
I've always just sorta tuned those, so yeah this this good.

iguessthislldo (approved):

PR #2158, opened 2020-12-10
Bench2: Make worker's WriteAction use infinite wait_for_acks on stop()

simpsont-oci:
... but make it configurable so we can change as needed when testing needs demand it.

### DIFF
--- a/performance-tests/bench_2/worker/WriteAction.cpp
+++ b/performance-tests/bench_2/worker/WriteAction.cpp
@@ -15,6 +15,7 @@ WriteAction::WriteAction(ACE_Proactor& proactor)
  , max_count_(0)
  , new_key_count_(0)
  , new_key_probability_(0)
+ , final_wait_for_ack_{DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC}
 {
 }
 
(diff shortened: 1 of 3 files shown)


performance-tests/bench_2/worker/WriteAction.cpp
```diff
@@ -173,10 +182,9 @@ void WriteAction::stop()
   if (started_ && !stopped_) {
     stopped_ = true;
     proactor_.cancel_timer(*handler_);
-    const DDS::Duration_t timeout = { 0, 0 };
-    data_dw_->wait_for_acknowledgments(timeout);
+    data_dw_->wait_for_acknowledgments(final_wait_for_ack_);
     data_dw_->unregister_instance(data_, instance_);
-    data_dw_->wait_for_acknowledgments(timeout);
+    data_dw_->wait_for_acknowledgments(final_wait_for_ack_);
```
sonndinh, line 187: I'm curious why `wait_for_acknowledgments` is called twice here?

performance-tests/bench_2/worker/WriteAction.cpp
```diff
@@ -15,6 +15,7 @@ WriteAction::WriteAction(ACE_Proactor& proactor)
  , max_count_(0)
  , new_key_count_(0)
  , new_key_probability_(0)
+ , final_wait_for_ack_{DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC}
```
sonndinh, line 18: So if the config file doesn't specify a value, it may wait forever on a call to `stop()`?

performance-tests/bench_2/worker/WriteAction.cpp
simpsont-oci: To allow data to propagate before unregistering the instance (potentially gapping data that needs to be resent), and then to allow the instance unregister to propagate before exiting. In general, the instance stuff in WriteAction probably needs a bit of work. I was just lazy here and didn't want to subdivide the final_wait_for_ack_ time period manually here so it uses the same period twice.

sonndinh (approved):

PR #2159, opened 2020-12-10
clang warning fixes

kuznetsovmoci:
OpenDDS/tests/DCPS/TcpReconnect/publisher.cpp:92:32: warning: format string is not a string literal (potentially insecure)
OpenDDS/performance-tests/bench_2/test_controller/main.cpp:10:32: warning: unknown warning group '-Wclass-memaccess'
OpenDDS/performance-tests/bench_2/worker/main.cpp:16:32: warning: unknown warning group '-Wclass-memaccess', ignored [-Wunknown-warning-option]

### DIFF
--- a/tests/DCPS/TcpReconnect/publisher.cpp
+++ b/tests/DCPS/TcpReconnect/publisher.cpp
@@ -89,7 +89,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
       ACE_TString command_line = stubCmd + ACE_TEXT(" ") + stubArgs;
       ACE_DEBUG((LM_INFO, ACE_TEXT("stub command line: %s\n"), command_line.c_str()));
 
-      if (options.command_line(ACE_TEXT("%s"), command_line.c_str()) != 0)
+      if (options.command_line("%s", ACE_TEXT_ALWAYS_CHAR(command_line.c_str())) != 0)
         ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT ("%p\n"), ACE_TEXT("set options")) ,-1);
 
       if (stub_ready_filename.empty())
--- a/tests/DCPS/TcpReconnect/publisher.cpp
+++ b/tests/DCPS/TcpReconnect/publisher.cpp
@@ -89,7 +89,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
       ACE_TString command_line = stubCmd + ACE_TEXT(" ") + stubArgs;
       ACE_DEBUG((LM_INFO, ACE_TEXT("stub command line: %s\n"), command_line.c_str()));
 
-      if (options.command_line(command_line.c_str()) != 0)
+      if (options.command_line(ACE_TEXT_CHAR_TO_TCHAR(command_line.c_str())) != 0)
         ACE_ERROR_RETURN((LM_ERROR, ACE_TEXT ("%p\n"), ACE_TEXT("set options")) ,-1);
 
       if (stub_ready_filename.empty())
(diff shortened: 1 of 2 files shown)


tests/DCPS/TcpReconnect/publisher.cpp
```diff
@@ -89,7 +89,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
       ACE_TString command_line = stubCmd + ACE_TEXT(" ") + stubArgs;
       ACE_DEBUG((LM_INFO, ACE_TEXT("stub command line: %s\n"), command_line.c_str()));
 
-      if (options.command_line(ACE_TEXT_CHAR_TO_TCHAR(command_line.c_str())) != 0)
+      if (options.command_line(ACE_TEXT(command_line.c_str())) != 0)
```
iguessthislldo, line 92: This isn't going to compile.

iguessthislldo: On wchar builds that is.

iguessthislldo: To satisfy the warning, the format string needs to be a string literal:

```suggestion
      if (options.command_line("%s", ACE_TEXT_ALWAYS_CHAR(command_line.c_str())) != 0)
```

kuznetsovmoci: I pushed a new fix before I saw your comments. Do you think this version would work too?

jwillemsen: Instead of converting the command_line string to plain ascii you can also use the correct format string, based on wchar or not, use `%s` or `%ls`, see for example https://github.com/DOCGroup/ACE_TAO/blob/368d8bff6c4b9889eb67d6337b3ebfaf5a6bca2c/ACE/tests/Process_Manager_Test.cpp#L84

kuznetsovmoci: The version I wrote is based on similar code in TAO_IDL be_interface.cpp and utl_global.cpp. I welcome and appreciate other  suggestions. Now, when we have more than one, which one would be the best?

iguessthislldo: > Now, when we have more than one, which one would be the best?

More than one of what?

kuznetsovmoci: "suggestions". We have three options now (unless the one in the PR now is wrong - then we have two).

tests/DCPS/TcpReconnect/publisher.cpp
```diff
@@ -89,7 +89,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
       ACE_TString command_line = stubCmd + ACE_TEXT(" ") + stubArgs;
       ACE_DEBUG((LM_INFO, ACE_TEXT("stub command line: %s\n"), command_line.c_str()));
 
-      if (options.command_line(ACE_TEXT(command_line.c_str())) != 0)
+      if (options.command_line(ACE_TEXT("%s"), command_line.c_str()) != 0)
```
iguessthislldo, line 92: So `command_line` is an `ACE_TString`, which means `c_str()` will be `const char*` normally, but will be `const wchar*` if ACE is built in its weird "wchar preferred" mode and that won't work with `%s`. You can either do what I suggested and use `ACE_TEXT_ALWAYS_CHAR` that makes sure it works with `%s`, which will only take up one or do what Johnny suggested, to switch the format string to using `%ls` conditionally.

tests/DCPS/TcpReconnect/publisher.cpp
iguessthislldo: Personally I'd prefer the one that minimizes the involvement of `ACE_USES_WCHAR` because I don't like anything about the whole system, since it leads to confusion like this. However what Johnny suggested is the "correct" way to use the system since it allows for the possibility of wide characters in `command_line`. It's up to you to choose.

tests/DCPS/TcpReconnect/publisher.cpp
kuznetsovmoci: This makes sense. I also agree with your comments above about the "correct" way, but would go with your suggestion, as it looks "lighter".

iguessthislldo (approved):

PR #2170, opened 2020-12-11
fix failure in safety profile base

lij-oci:
This is one of the tests that failed safety profile.


simpsont-oci:
I don't think this fixes anything. The C++ code change seems to reduce an internal wait time, but the Perl code change seems to increase the total test run time. And 600 seconds is 10 minutes, which is an awfully long time to wait for a test to fail. If the test didn't complete after 5 minutes, I'm guessing 10 minutes isn't going to matter.

PR #2171, opened 2020-12-11
Segmentation Fault Due to Missing Export Android Api30

ClaytonCalabrese:
A seg fault would occur when trying to run rtps on android api 30 due to a missing export. This adds that export.


iguessthislldo (approved):
I wonder how many more times we'll have to do this.

PR #2173, opened 2020-12-12
Remove usage of localtime

jwillemsen:
Depends on a perl module which is not part a default RHEL installation. Didn't add real functionality to the test, the progress is already visible because of the test count is shown

### DIFF
--- a/tests/security/attributes/run_test.pl
+++ b/tests/security/attributes/run_test.pl
@@ -13,8 +13,6 @@ use lib "$ACE_ROOT/bin";
 use PerlDDS::Run_Test;
 use strict;
 
-use Time::Piece;
-use Time::Seconds;
 use Getopt::Long;
 
 my $scenario;
@@ -486,13 +484,8 @@ foreach my $gov_file (@gov_files) {
         #if ($status != 0) {
           $current_test_num++;
 
-          my $total_test_percent = (100.0 * $current_test_num) / $total_test_count;
-          my $current_time = localtime;
-          my $elapsed_time = $current_time - $test_start_time;
-          my $estimate = $elapsed_time->seconds * ($total_test_count - $current_test_num) / ($current_test_num);
-
           if ($total_test_count != 1) {
-            print "\ntest #$current_test_num of $total_test_count. Estimating $estimate seconds remaining.\n";
+            print "\ntest #$current_test_num of $total_test_count.\n";
           }
 
           #print "$gov_file $pub_perm_file $sub_perm_file $topic_name $status\n\n-----------\n\n";


iguessthislldo:
You could change it be optional based on checking if it could import the module.

jwillemsen:
> You could change it be optional based on checking if it could import the module.

Is that worth the effort, the fact that you can see the total amount of tests and the current test should be enough, at least the test shows progress

iguessthislldo:
It's not much more than the effort it takes to remove the code.

iguessthislldo:
Or you could just install the missing module.

jwillemsen:
We don’t have full control on all systems, so I prefer a change which makes it possible to run this script without having to install additional perl modules

iguessthislldo:
I agree that this case doesn't really matter and the code could be removed without any real loss. However I'd like to make the case that at least in my experience it's hard to tell what Perl modules can be expected to be on most systems. These particular modules don't look like they're OS-specific and are listed as ["core modules"](https://perldoc.perl.org/modules), so I would have expected it to be safe. On top of that we do have cases like [`Win32::Process::Info`](https://stackoverflow.com/questions/64320132/what-is-win32-process-pm-in-inc) which isn't a "code module", but is recommended by the Perl documentation and which we actually do need a Windows user to install if their Perl doesn't have it included and they want to run tests. I guess what I'm trying to say and the reason I'm resistant to this change is that I don't want us to play wack-a-mole with Perl modules.

mitza-oci:
Both of these things are true:
- all of Perl is required, not some subset defined by package maintainers (on RHEL/CentOS you can `yum install perl-core`)
- the use of `localtime` in this particular script may not be all that useful (@simpsont-oci your thoughts?)

simpsont-oci:
I originally wrote the security attributes test with the thought that it would be good to occasionally iterate over all possible permutations of the security governance and topic attributes in order to check for valid behaviors. For a number of reasons, this wound up not being practical, not least of which is the time involved to run such a test. This timing code was originally written in order to give the test-runner some semblance of an idea of the time remaining for these long, exhaustive tests.

Currently, we never run the security attribute tests this way, nor can I envision us ever really wanting to do so. It would be incredibly tedious to try to categorize all the expected failure points within the test application for every possible combination of security attributes, and surely there are more "failure" cases than "success" cases in the global space of attribute combinations. We could possibly try to write code to predict where the failures would be for each permutation, but in the end I think that would wind up being just as complicated (if not more so) than the internal algorithms.

I'm fine with removing this aspect of the code as long as it doesn't impact the current way we're using the security attributes tests (running individual test scenarios with expected outcomes).

simpsont-oci (approved):

PR #2175, opened 2020-12-12
Fixed unitialized warning on rhel75

jwillemsen:
* dds/DCPS/security/CryptoBuiltInImpl.cpp:


iguessthislldo (approved):

mitza-oci:
For the record the compiler is wrong in warning, there's no way for this to be used before being initialized.

PR #2177, opened 2020-12-13
Thrasher test: added time and thread id in output

lij-oci:
Modified Thrasher test output as a pre-step of fixing its failure in safety profile base.


mitza-oci:
Now that #2172 is in, merge master to the source branch of this PR to resolve conflicts

tests/DCPS/Thrasher/DataReaderListenerImpl.cpp
```diff
@@ -69,10 +69,12 @@ void DataReaderListenerImpl::on_sample_lost(DDS::DataReader_ptr,
 
 bool DataReaderListenerImpl::wait_received(const OpenDDS::DCPS::TimeDuration& duration, size_t target)
 {
+  ACE_DEBUG((LM_INFO, ACE_TEXT("%T(%t) sub->wait received_samples %d\n"), received_samples_));
```
jrw972, line 72: %T is unnecessary because the test can be configured with OrbVerboseLogging.  If you are going to add %t, then you should probably have %P|%t.

tests/DCPS/Thrasher/ParticipantTask.cpp
```diff
@@ -39,9 +40,9 @@ ParticipantTask::~ParticipantTask()
 int
 ParticipantTask::svc()
 {
-  try {
-    ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t)    -> PUBLISHER STARTED\n")));
-
+  std::string pfx("%T pub");
```
jrw972, line 43: I disagree with the approach of a dynamic format string.  Use a static format for all logging statements and inject the appropriate variables.

tests/DCPS/Thrasher/ParticipantTask.cpp
```diff
@@ -168,8 +180,8 @@ ParticipantTask::svc()
     // pathways related to publication; we should be especially dull
     // and write only one sample at a time per writer.
 
-    ProgressIndicator progress("(%P|%t)       PUBLISHER %d%% (%d samples sent)\n",
-                               samples_per_thread_);
+    const std::string fmt(pfx + "  %d%% (%d samples sent)\n");
+    ProgressIndicator progress(fmt.c_str(), samples_per_thread_);
```
jrw972, line 184: Make the progress indicator take a thread id to avoid the dynamic format string.

jrw972 (changes requested):

tests/DCPS/Thrasher/DataReaderListenerImpl.cpp
jrw972: Please add the OrbVerboseLogging to this test as that will help in the future.

tests/DCPS/Thrasher/ParticipantTask.cpp
lij-oci: I tried but changed back, because the progress indicator is shared by both pub and sub. The sub does not need thread id since it only has a main thread and a listener thread, while the pub can have a variable number of threads. Making indicator save a thread id as member not only makes no sense for sub, but also the format string is still needed since they are different. One is "(%P|%t) sub %d%% (%d samples received)\n", and the other is "(%P|%t) pub%d  %d%% (%d samples sent)\n". So wrapping the pub thread id in the string seems to be a way to simplify this.

tests/DCPS/Thrasher/ParticipantTask.cpp
lij-oci: See the comment below about the format string.

PR #2178, opened 2020-12-14
Allow Transport Template Customization of IP Without Port

simpsont-oci:


jonesc-oci (changes requested):
I think this check can be removed, since there's an identical check above.

jonesc-oci (approved):

PR #2190, opened 2020-12-15
Only use {{}} as array initializer for arrays of pod types

jwillemsen:
* dds/idl/langmap_generator.cpp:
    * tests/DCPS/Compiler/idl_test1_lib/FooDef.idl:

Matches the functionality as we have in TAOX11, see https://github.com/RemedyIT/taox11/blob/9719377a200a0859755c1dd4a064a2ebc9e21e2f/ridlbe/c%2B%2B11/templates/cli/hdr/struct_post.erb#L48

### DIFF
--- a/dds/idl/langmap_generator.cpp
+++ b/dds/idl/langmap_generator.cpp
@@ -1474,7 +1474,12 @@ struct Cxx11Generator : GeneratorBase
       }
     } else {
       if (af.cls_ & CL_ARRAY) {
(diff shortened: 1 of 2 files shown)


jwillemsen:
Tested manually on a rhel75 system, tests/DCPS/Compiler/C++11/idl_test1_lib is now warning free

mitza-oci (approved):

Issue #2167, opened 2020-12-11
selected google test version doesn't compile with cmake on RHEL 7.5


jwillemsen:
I am setting up a new build on rhel 7.5 and it looks the google test module used by OpenDDS doesn't work there, cmake fails, see https://buildlogs.remedy.nl/rhel75_opendds_versioned/index.html.

Looks googletest 1.10.x requires C++11 support which shouldn't be a problem on rhel75. Not sure about the correct path forward with this dependency on googletest and the need to support C++03

jwillemsen:
I validated that googletest 1.8.0 does work (see https://github.com/google/googletest/releases/tag/release-1.8.0)

iguessthislldo:
> https://buildlogs.remedy.nl/rhel75_opendds_versioned.

This link leads to a 404 page.

> Not sure about the correct path forward with this dependency on googletest and the need to support C++03

I don't think there is a path forward for now. If memory serves correct we are using the version we're using because it's the last to support C++03.

jwillemsen:
Sorry, internal link, it should be https://buildlogs.remedy.nl/rhel75_opendds_versioned/index.html

mitza-oci:
This doesn't seem like something RHEL would break between 7.4 and 7.5.  Aren't they supposed to be providing stability?

jwillemsen:
We always explicitly validate each minor release, there is always the risk of a breaking change

mitza-oci:
I'm not questioning the approach of trying out new platform versions.  Since this is an error from cmake and not from the code, is there a workaround?

jwillemsen:
A downgrade to googletest 1.8.0 did resolve the cmake error. Are there any features that we really need from 1.8.1, if not, maybe merge #2168, at the moment C++11 support becomes mandatory we could upgrade to a newer googletest

jwillemsen:
Looks this specific compiler triggers some more warning to look at

mitza-oci:
Which version of CMake is this using?

mitza-oci:
Worked for me using a current CMake release.  I'll update documentation since I'm doing a refresh in preparation for 3.15.

jwillemsen:
rhel 7.5 gets shipped with `cmake version 2.8.12.2`

mitza-oci:
https://github.com/objectcomputing/OpenDDS/pull/2183

PR #2193, opened 2020-12-15
Add requires for googletest, makes it possible to disable the compila…

jwillemsen:
…tion of all tests dependent on googletest using googletest=0 through default.features

    * MPC/config/googletest.mpb:

### DIFF
--- a/MPC/config/googletest.mpb
+++ b/MPC/config/googletest.mpb
@@ -1,4 +1,6 @@
 project {
+  requires += googletest
+
   expand(GTEST_ROOT) {
     $GTEST_ROOT
     $(DDS_ROOT)/tests/googletest/build/install


jwillemsen:
Makes it possible to disable googletest as dependent library because of https://github.com/objectcomputing/OpenDDS/issues/2167

jwillemsen:
MPC features are default 1 so this will not impact the default case

mitza-oci:
As of now GoogleTest is a requirement when building with tests enabled.  This is documented in docs/dependencies.md.

This risk of changing it is that there would be different subsets of testing.  This is not desirable.

GoogleTest doesn't buy us all that much since we are not using advanced features.  What it does can be done with a simple assertion macro.  If GoogleTest is too onerous a requirement then it should be completely removed.

simpsont-oci:
I'm pretty sure the security access control tests use google mock.

jwillemsen:
I disabled OpenDDS security for RHEL 7.5 because of the problems compiling google test there, looks only DCPS/tests/SequenceIteratorTest.cpp uses google test, looks that one can be easily rewritten to not use googletest

mitza-oci:
> I disabled OpenDDS security for RHEL 7.5 because of the problems compiling google test there, looks only DCPS/tests/SequenceIteratorTest.cpp uses google test, looks that one can be easily rewritten to not use googletest

Google Test can be built on RHEL 7.5 with a modern CMake version.  I just did it myself yesterday.

jwillemsen:
Reworked unit test in #2195. Did you build cmake from sources or used a specific RHEL channel?

mitza-oci:
> Reworked unit test in #2195. Did you build cmake from sources or used a specific RHEL channel?

I downloaded CMake from the link that cmake.org provided, which I think goes to GitHub

jwillemsen:
> > Reworked unit test in #2195. Did you build cmake from sources or used a specific RHEL channel?
> 
> I downloaded CMake from the link that cmake.org provided, which I think goes to GitHub

Ok, a manual build from sources, something I can't do on some of our build slaves. Hopefully at the moment we need DDS security support we are at RHEL8 which has much more modern support for various dev tools

mitza-oci:
> > > Reworked unit test in #2195. Did you build cmake from sources or used a specific RHEL channel?
> > 
> > 
> > I downloaded CMake from the link that cmake.org provided, which I think goes to GitHub
> 
> Ok, a manual build from sources, something I can't do on some of our build slaves. Hopefully at the moment we need DDS security support we are at RHEL8 which has much more modern support for various dev tools

No, it was a download of a Linux binary.

jwillemsen:
> > > > Reworked unit test in #2195. Did you build cmake from sources or used a specific RHEL channel?
> > > 
> > > 
> > > I downloaded CMake from the link that cmake.org provided, which I think goes to GitHub
> > 
> > 
> > Ok, a manual build from sources, something I can't do on some of our build slaves. Hopefully at the moment we need DDS security support we are at RHEL8 which has much more modern support for various dev tools
> 
> No, it was a download of a Linux binary.

Ok, I will try to find some time to try to install that as part of the job into a local user directory in case we build opendds with security enabled.

mitza-oci:
It's just a wget of `https://github.com/Kitware/CMake/releases/download/v3.19.1/cmake-3.19.1-Linux-x86_64.tar.gz`, untar, and run cmake using an absolute path

iguessthislldo:
> GoogleTest doesn't buy us all that much since we are not using advanced features. What it does can be done with a simple assertion macro. If GoogleTest is too onerous a requirement then it should be completely removed.

The XTypes tests make extensive use of Google Test. It would not be simple to remove.

jwillemsen:
> It's just a wget of `https://github.com/Kitware/CMake/releases/download/v3.19.1/cmake-3.19.1-Linux-x86_64.tar.gz`, untar, and run cmake using an absolute path

Next build of our rhel75 opendds build installs cmake 3.19.1 and uses that for googletest

jwillemsen:
cmake 3.19 gives the following warnings on googletest, latest versions of googletest do require C++11 so upgrading googletest is not an option right now, something for OpenDDS 4.0. A downgrade to cmake 3.18.5 gets rid of these warnings, so we are using 3.18.5 for the moment

```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
Compatibility with CMake < 2.8.12 will be removed from a future version of
CMake.
```

PR #2196, opened 2020-12-15
Removed dependency from googletest for NetworkConfig test

jwillemsen:
* tests/DCPS/UnitTests/NetworkConfig.cpp:
    * tests/DCPS/UnitTests/UnitTests.mpc:


sonndinh:
What is the reason for removing googletest for this test?

jwillemsen:
#2167, why use googletest for just calling 4 operations in a main?

iguessthislldo:
> #2167 

That issue is closed and this specific test isn't mentioned in that issue.

sonndinh:
> #2167, why use googletest for just calling 4 operations in a main?

If the removal is because of the CMake build error for gtest in #2167, I think Adam mentioned the latest CMake works for him.

iguessthislldo:
Closing because of #2167

PR #2208, opened 2020-12-16
DRAFT: IPv6 Failed to Join Multicast Group

ClaytonCalabrese:
The previous change seems to have fixed tests where it was one process with multiple participants, such as the DCPSInfoRepo test.  This change should make tests with multiple processes more reliable. 

The same problem is that multiple things are accessing the join at the same time.


ClaytonCalabrese:
Putting into draft until Mike's changes are merged into ace.

PR #2222, opened 2020-12-18
Participant BIT and location BIT not populated

jrw972:
Problem
-------

A secure remote participant was never published to the participant BIT
and, consequently, the location topic was never published.

Solution
--------

When security is enabled, publish the participant in the BIT when the
secure SPDP message arrives.  Add the ability to publish the location
topic even when there are no updates.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -464,16 +464,10 @@ void Spdp::process_location_updates_i(DiscoveredParticipantIter iter, bool force
 {
   // We have the global lock.
 
-  if (iter->second.bit_ih_ == DDS::HANDLE_NIL) {
-    // Not in the built-in topics.
-    return;
-  }
-
-  const RepoId guid = iter->first;
-
   DiscoveredParticipant::LocationUpdateList location_updates;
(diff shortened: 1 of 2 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1836,27 +1849,6 @@ Spdp::match_authenticated(const DCPS::RepoId& guid, DiscoveredParticipantIter& i
...
-        bit->store_synthetic_data(pbtd, DDS::NEW_VIEW_STATE);
-      rg.release();
-      iter = participants_.find(guid);
-      if (iter == participants_.end()) {
-        return false;
-      }
-    }
-  }
-#endif /* DDS_HAS_MINIMUM_BIT */
-
```
iguessthislldo, line 1859: Why remove this part? Isn't this when a secure participant without secure user data would get the participant bit data or is this happening somewhere else?

jrw972: Secure user data dictates what you send, not what you receive.  Reception-wise, a secure participant should only trust the data received from the secure/reliable SPDP writer.  The code in `handle_participant_data` has now been expanded to write to the BIT when this data arrives the first time.

PR #2224, opened 2020-12-19
Fix MultiDPTest for Safety Profile

sonndinh:
- Remove the use of temporary files to synchronize publisher and subscriber because of a potential race
- Create a transport instance for each participant in the subscriber


tests/DCPS/MultiDPTest/DataReaderListener.cpp
```diff
@@ -8,158 +8,125 @@
...
         reader, foo.x, foo.y, foo.data_source));
       ACE_DEBUG((LM_DEBUG,
         ACE_TEXT("(%P|%t) SampleInfo.sample_rank = %d \n"),
         si.sample_rank));
+    } else if (status == ::DDS::RETCODE_NO_DATA) {
+      ACE_ERROR_RETURN((LM_ERROR,
+        ACE_TEXT("(%P|%t) ERROR: reader received ::DDS::RETCODE_NO_DATA!\n")), -1);
+    } else {
+      ACE_ERROR_RETURN((LM_ERROR,
+        ACE_TEXT("(%P|%t) ERROR: read  foo: Error: %d\n"), status), -1);
```
jwillemsen, line 36: Isn't there a helper to log status as string?

sonndinh: This is updated.

tests/DCPS/MultiDPTest/publisher.cpp
```diff
@@ -160,165 +127,131 @@ void init ()
...
-            delete writers[i];
-          }
+    // let the Service_Participant (in above line) strip out -DCPSxxx parameters
+    // and then get application specific parameters.
+    parse_args(argc, argv);
+
+    init();
+
+    Allocator* allocator;
+    ACE_NEW_RETURN(allocator, Allocator(mmap_file), -1);
```
mitza-oci, line 174: Is this going to work on all supported platforms?  It's unclear from the diffs on GitHub where this class is coming from.

The test is set up to use temporary files since they are more widely supported.  They can also support running tests across different hosts - this has been used with embedded targets where one side (perhaps publisher) is on the target and the other (perhaps subscriber) is on the host system.  Using a shared filesystem like NFS allows the temp files to work across hosts.

mitza-oci (changes requested):

tests/DCPS/MultiDPTest/publisher.cpp
sonndinh: `Allocator` is a typedef of an `ACE_Malloc_T` template class. My intention was to replace the temporary files with shared memory for synchronization between publisher and subscriber. But this wouldn't work if the publisher and subscriber run on different hosts. 

The reason I changed to use shared memory is because one of the temporary files was also used to communicate the number of timed-out writes. But it's possible that the subscriber may fail to read that number even though the publisher has written that number (but it isn't flushed to the file yet). But I think that could be fixed to keep using temporary files.

mitza-oci (approved):

PR #2225, opened 2020-12-19
Reauthentication fails

jrw972:
Problem
-------

The main challenge in reauthentication is re-keying the volatile
reader and writer.  The existing approach of disassociating and
associating does not work with the new and improved handling of crypto
tokens.  Specifically, the association does not result in the new
shared secret being installed.

Solution
--------

After disassociating, remove the crypto tokens for the remote volatile
reader and writer so that the call to associate will install the new
shared secret.


jrw972:
@iguessthislldo We probably want to pull this into #2024.

kuznetsovmoci (approved):

mitza-oci (approved):

PR #2232, opened 2020-12-22
README.md and other docs: updates for 3.15

mitza-oci:


README.md
```diff
@@ -150,12 +149,11 @@ This release of OpenDDS has been tested under the following platforms:
 
 Linux family:
 * Red Hat EL and CentOS 6.6, 6.8, 6.9 (x86\_64)
-* Red Hat EL and CentOS 7.2, 7.3, 7.4 (x86\_64)
+* Red Hat EL and CentOS 7.2, 7.3, 7.4, 7.5 (x86\_64)
 * Fedora 24 and 31 (x86\_64)
```
jwillemsen, line 153: Our build is now fc33

README.md
```diff
@@ -124,8 +124,7 @@ Perl is used for the configure script, running the automated tests and examples
 included in this source tree, and generating Makefiles or Visual Studio project
 files.
 
-On Windows we recommend the use of [ActiveState Perl](
-https://www.activestate.com/activeperl).
+On Windows we recommend the use of [Straweberry Perl](https://strawberryperl.com).
```
iguessthislldo, line 127: If we're changing this, [INSTALL.md](https://github.com/objectcomputing/OpenDDS/blob/master/INSTALL.md#compiling) and [configure.cmd](https://github.com/objectcomputing/OpenDDS/blob/4f3eba4be2a7f4d7412d4d041852bb2cbd79719a/configure.cmd#L8) also reference ActiveState Perl.

jwillemsen: Our builds use activestate without problems

iguessthislldo: It's still possible to use it, but we've had a few minor issues with it. I think our main problems with it are that [it doesn't come with `Win32::Process`](https://stackoverflow.com/questions/64320132/what-is-win32-process-pm-in-inc/64321097) included and we couldn't use it build OpenDDS for Android on Windows, so [the instructions for that already say to use Strawberry Perl for that](https://github.com/objectcomputing/OpenDDS/blob/master/docs/android.md#building-on-windows).

jwillemsen: Ok, maybe mention this somewhere in the docu that ActiveState could be problematic?

PR #2235, opened 2020-12-24
Add a union test with array members

jwillemsen:
* tests/DCPS/Compiler/idl_test1_lib/FooDef.idl:


jrw972 (approved):

mitza-oci:
@jwillemsen was the intent here just to check that the generated code compiles?  If so we can move it back out of draft state.
Or does it need some additional C++ code in the test?

jwillemsen:
Yes, just check that it all compiles

PR #2237, opened 2020-12-31
Android API30+ Triage due to Loss of NetlinkSockets

ClaytonCalabrese:
**Problem**: Android API30+ added security measures which result in netlink sockets failing when called from an application.  

**Consequence**: LinuxNetworkConfigMonitor is unable to detect network changes, and is therefore unfit to be used on android30+.

**Proposal**: Find a way to reconnect when a network change happens.  There are android NetworkCallbacks which allow for calls when the android device detects network changes.  We would need to use this plus NetworkConfigModifier to have a monitor on android api30+.  This is similar to how we operate on iOS. 

**Implementation**: If LinuxNetworkConfigMonitor attempts run but fails to setup, it will fail into a NetworkConfigModifier.  This condition fits with the behavior currently happening on the phone, and gets us a NetworkConfigModifier to work with.  
Upon a user doing the JNICALL of "Java_OpenDDS_DCPS_TheParticipantFactory_WithArgs" on an android device, the context will be obtained, allowing us to set up network callbacks.  Currently these callbacks are onAvailable, onLost, and onCapabilitiesChanged.  

The following corresponds to this documentation: https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback

onAvailable is the trigger I used to add an interface/address, and corresponds to "Called when the framework connects and has declared a new network ready for use." 

onLost is one of the triggers I used to remove an interface/address, and corresponds to "Called when a network disconnects or otherwise no longer satisfies this request or callback."

Initially, I thought these two would be sufficient, but for some reason onLost is called before the network has properly been removed from the ifaddrs causing my implementation to miss timing. This lead to me adding onCapabilitiesChanged.

onCapabilitiesChanged is the other trigger I used to remove an interface/address, and corresponds to "Called when the network corresponding to this request changes capabilities but still satisfies the requested criteria.".

By having both onLost and onCapabilitiesChanged, the removal works as desired.  For some reason, just having one or the other doesn't work.  onCapabilitiesChanged has the downside of being pretty spammy, so there are definitely some wasted checks occuring.

Although I have been saying addition and removal are separate, they are done from the same function, but the timing of the call leads to a logical result of either additional or removal.  If onAvailable was the trigger, then there should have been an addition but no remove.  If the others were the trigger, there should have been a removal, but not an addition.  

**Desired Changes before merge:** 
1. In NetworkConfigModifier.h the declaration of NetworkConfigModifier was hidden behind "if defined(ACE_HAS_GETIFADDRS) && !defined(OPENDDS_SAFETY_PROFILE) && !defined (ACE_LINUX)".  This was blocking Android for some reason. Obviously this ifdef existed for a reason, so I'd like to add back as much as possible without blocking the android use. 

2. Java_OpenDDS_DCPS_NetworkConfigModifier__1jni_1fini has a TODO that I'm not sure what to do with.

3. I probably need to add some more debug statements somewhere.


dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+      // Skip the ANY address
+      if (!IN6_IS_ADDR_UNSPECIFIED(&addr->sin6_addr)) {
+        net_names.insert(std::make_pair(p_if->ifa_name, p_if));
+      }
+    }
+# endif /* ACE_HAS_IPV6 */
+  }
+  // Remove interfaces that are no longer active
+  NetworkInterfaces nis = get_interfaces();
+  for (OpenDDS::DCPS::NetworkInterfaces::iterator iter = nis.begin(); iter < nis.end(); ++iter) {
```
iguessthislldo, line 142:
```suggestion
  for (NetworkInterfaces::iterator iter = nis.begin(); iter != nis.end(); ++iter) {
```

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+
+  // Add interfaces that are new
+  nis = get_interfaces();
+  for (std::map<std::string, ifaddrs*>::iterator iter = net_names.begin(); iter != net_names.end(); ++iter) {
+    NetworkInterfaces::iterator pos = std::find_if(nis.begin(), nis.end(), NetworkInterfaceName(iter->first));
+    if (pos == nis.end()) {
+      ifaddrs* ifa = iter->second;
+      ACE_INET_Addr address;
+      if (ifa->ifa_addr->sa_family == AF_INET) {
+        struct sockaddr_in* addr = reinterpret_cast<sockaddr_in*> (ifa->ifa_addr);
```
iguessthislldo, line 156:
```suggestion
        sockaddr_in* addr = reinterpret_cast<sockaddr_in*>(ifa->ifa_addr);
```

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+        address.set((u_short) 0, addr->sin_addr.s_addr, 0);
+        NetworkInterface ni(ACE_OS::if_nametoindex(ifa->ifa_name), ifa->ifa_name, ifa->ifa_flags & (IFF_MULTICAST | IFF_LOOPBACK));
```
iguessthislldo, line 158: Factor out `ifa->ifa_flags & (IFF_MULTICAST | IFF_LOOPBACK)` here and below into common constant?

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+        ni.add_address(address);
+        NetworkConfigMonitor::add_interface(ni);
+      }
+# if defined (ACE_HAS_IPV6)
+      else if (ifa->ifa_addr->sa_family == AF_INET6) {
+        struct sockaddr_in6 *addr = reinterpret_cast<sockaddr_in6 *> (ifa->ifa_addr);
+        address.set(reinterpret_cast<struct sockaddr_in *> (addr), sizeof(sockaddr_in6));
```
iguessthislldo, line 165:
```suggestion
        sockaddr_in6* addr = reinterpret_cast<sockaddr_in6*>(ifa->ifa_addr);
        address.set(reinterpret_cast<sockaddr_in*>(addr), sizeof(sockaddr_in6));
```

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+        NetworkInterface ni(ACE_OS::if_nametoindex(ifa->ifa_name), ifa->ifa_name, ifa->ifa_flags & (IFF_MULTICAST | IFF_LOOPBACK));
+        ni.add_address(address);
+        NetworkConfigMonitor::add_interface(ni);
+      }
+# endif /* ACE_HAS_IPV6 */
+    }
+  }
+  ::freeifaddrs (p_ifa);
```
iguessthislldo, line 173:
```suggestion
  ::freeifaddrs(p_ifa);
```

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -100,6 +100,79 @@ bool NetworkConfigModifier::close()
...
+  NetworkInterfaces nis = get_interfaces();
+  for (OpenDDS::DCPS::NetworkInterfaces::iterator iter = nis.begin(); iter < nis.end(); ++iter) {
+    if (net_names.find(iter->name()) == net_names.end()) {
+      remove_interface(iter->index());
+    }
+  }
+
+  // Add interfaces that are new
+  nis = get_interfaces();
+  for (std::map<std::string, ifaddrs*>::iterator iter = net_names.begin(); iter != net_names.end(); ++iter) {
```
iguessthislldo, line 150:
```suggestion
  for (Names::iterator iter = net_names.begin(); iter != net_names.end(); ++iter) {
```

Or whatever the type is.

dds/DCPS/NetworkConfigModifier.h
```diff
@@ -10,8 +10,6 @@
 
 #include "ace/config.h"
 
-#if defined(ACE_HAS_GETIFADDRS) && !defined(OPENDDS_SAFETY_PROFILE) && !defined (ACE_LINUX)
-
```
iguessthislldo, line 14: This should be restored probably. Safety profile build is failing. Also ACE treats Android a separate platform as far as `ACE_LINUX` goes if that's why you removed it.

configure
```diff
@@ -107,6 +107,8 @@ my %platforminfo =
                   'usage' => [
                       "Use --macros=ANDROID_ABI=<ARCH> to specify the",
                       "target architecture.",
+                      "Use --macros=ANDROID_API_PATH=<API_PATH> to specifify",
```
iguessthislldo, line 110:
```suggestion
                      "Use --macros=ANDROID_API_PATH=<API_PATH> to specify",
```

dds/DCPS/Service_Participant.cpp
```diff
@@ -2774,11 +2774,28 @@ NetworkConfigMonitor_rch Service_Participant::network_config_monitor()
...
+        network_config_monitor_ = make_rch<NetworkConfigModifier>();
+        if (network_config_monitor_ && !network_config_monitor_->open()) {
+          open_failed = true;
+        }
+#else
+        open_failed = true;
+#endif
+        if (open_failed) {
+          ACE_ERROR((LM_ERROR,
+            ACE_TEXT("(%P|%t) ERROR: Service_Participant::network_config_monitor could not open network config monitor\n ")));
```
iguessthislldo, line 2794:
```suggestion
            ACE_TEXT("(%P|%t) ERROR: Service_Participant::network_config_monitor could not open network config monitor\n")));
```

java/dds/OpenDDS/DCPS/AndroidNetworkCallback.java
```diff
@@ -0,0 +1,53 @@
...
+import android.net.ConnectivityManager;
+import android.net.ConnectivityManager.NetworkCallback;
+import android.net.Network;
+import android.net.NetworkCapabilities;
+import android.net.NetworkRequest;
+import android.os.Build;
+
+import OpenDDS.DCPS.NetworkConfigModifier;
+import OpenDDS.DCPS.TheServiceParticipant;
+
```
iguessthislldo, line 21:
```suggestion
```

I don't think you have to import something if it's in the current package.

iguessthislldo (changes requested):

iguessthislldo:
Testing it on the [Android matrix](https://github.com/iguessthislldo/OpenDDS-Android/actions/runs/460261250). Forced this run to all be Java, so we'll have to see if that works out because normally only a few builds are Java-enabled. It also has to work when it's not Java, but I might do that later.

iguessthislldo:
[This is the run that's actually doing something](https://github.com/iguessthislldo/OpenDDS-Android/runs/1641874347). Problem is that it's failing. [See this comment](https://github.com/objectcomputing/OpenDDS/pull/2237/files#r551111292).

dds/DCPS/NetworkConfigModifier.h
iguessthislldo: So [`ACE_LACKS_GETIFADDRS` is being defined for APIs < 24](https://github.com/DOCGroup/ACE_TAO/blob/78006062ece6be12eba3e81471961c74743ba071/ACE/ace/config-android.h#L74-L76).

iguessthislldo:
I also forgot to set up the Android SDK.

iguessthislldo:
I added the capability of downloading the Android SDK along with a platform which is what has `android.jar` and passing that to the configure script. [This is the latest run.](https://github.com/iguessthislldo/OpenDDS-Android/runs/1642565395)

dds/DCPS/NetworkConfigModifier.h
ClaytonCalabrese: Had to do a workaround for this as ACE is not properly setting ACE_HAS_GETIFADDERS for android.

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -105,72 +105,77 @@ void NetworkConfigModifier::update_interfaces()
...
+        names.insert(std::make_pair(p_if->ifa_name, p_if));
       }
     }
 # endif /* ACE_HAS_IPV6 */
   }
   // Remove interfaces that are no longer active
   NetworkInterfaces nis = get_interfaces();
-  for (OpenDDS::DCPS::NetworkInterfaces::iterator iter = nis.begin(); iter < nis.end(); ++iter) {
-    if (net_names.find(iter->name()) == net_names.end()) {
+  for (OpenDDS::DCPS::NetworkInterfaces::iterator iter = nis.begin(); iter != nis.end(); ++iter) {
```
iguessthislldo, line 146:
```suggestion
  for (NetworkInterfaces::iterator iter = nis.begin(); iter != nis.end(); ++iter) {
```

dds/DCPS/NetworkConfigModifier.cpp
```diff
@@ -105,72 +105,77 @@ void NetworkConfigModifier::update_interfaces()
...
 
   // Add interfaces that are new
   nis = get_interfaces();
-  for (std::map<std::string, ifaddrs*>::iterator iter = net_names.begin(); iter != net_names.end(); ++iter) {
+  for (Names::iterator iter = names.begin(); iter != names.end(); ++iter) {
     NetworkInterfaces::iterator pos = std::find_if(nis.begin(), nis.end(), NetworkInterfaceName(iter->first));
     if (pos == nis.end()) {
       ifaddrs* ifa = iter->second;
       ACE_INET_Addr address;
+      bool can_multicast = ifa->ifa_flags & (IFF_MULTICAST | IFF_LOOPBACK);
```
iguessthislldo, line 159:
```suggestion
      const bool can_multicast = ifa->ifa_flags & (IFF_MULTICAST | IFF_LOOPBACK);
```

dds/DCPS/NetworkConfigModifier.h
```diff
@@ -10,6 +10,9 @@
 
 #include "ace/config.h"
 
+// ACE_HAS_GETIFADDRS is not set on android but is available in API >= 24
+#if ((!defined (ACE_LINUX) && defined(ACE_HAS_GETIFADDRS))  || (defined(ACE_ANDROID) && __ANDROID_API__ >= 24)) && !defined(OPENDDS_SAFETY_PROFILE)
```
iguessthislldo, line 14: So I think this is almost right. I would say `(defined(ACE_ANDROID) && !defined ACE_LACKS_IF_NAMEINDEX)` because [ACE 6's condition is different](https://github.com/DOCGroup/ACE_TAO/blob/f1733f33df42f11ceade9575a010cc5a0495a8af/ACE/ace/config-android.h#L137) from the ACE master one I accidentally linked before.

configure
```diff
@@ -107,7 +107,7 @@ my %platforminfo =
                   'usage' => [
                       "Use --macros=ANDROID_ABI=<ARCH> to specify the",
                       "target architecture.",
-                      "Use --macros=ANDROID_API_PATH=<API_PATH> to specifify",
+                      "Use --macros=ANDROID_API_PATH=<API_PATH> to specify",
```
iguessthislldo, line 110: Two things:
- New make variable names like this should probably be lowercase to match other Android make ACE variables. `ANDROID_ABI` will still work for some time, but it has already been basically replaced with `android_abi` in ACE. I was going to change references when I updated OpenDDS to build using the NDK directly, but haven't gotten around to it yet.
- So I think we should split this into two different variables: `android_sdk` and `android_target_api`. It's probably unlikely, but theoretically a user might want to use Java and target a platform lower than 24, so doesn't care about the API 30 feature. That way we can check and disable `AndroidNetworkCallback.java` if the API doesn't support it. On top of that the user doesn't really have to know about the location of `android.jar` within the SDK.

This is separate and doesn't have to be done now, but it would have been helpful when getting this set up for the matrix to check that the `android.jar` actually exists and print an error if it doesn't. Otherwise it just has a bunch of "undefined" errors and it's not immediately clear what the issue is. It happened to me because I was downloading the minimum API platform instead of the targeted API platform, but could happen if the user a number of ways.

iguessthislldo (changes requested):

java/dds/OpenDDS_DCPS_jni.cpp
```diff
@@ -135,6 +157,43 @@ jstring JNICALL Java_OpenDDS_DCPS_TheServiceParticipant_get_1unique_1id
...
+  ncm->_remove_ref();
+}
+
+// NetworkConfigModifier::update_interfaces
+void JNICALL Java_OpenDDS_DCPS_NetworkConfigModifier_update_1interfaces
+(JNIEnv * jni, jobject jthis)
+{
+  OpenDDS::DCPS::NetworkConfigModifier* ncm = recoverCppObj<OpenDDS::DCPS::NetworkConfigModifier>(jni, jthis);
+  ncm->update_interfaces();
+}
```
iguessthislldo, line 194: References to `NetworkConfigModifier` are [causing errors](https://github.com/iguessthislldo/OpenDDS-Android/runs/1647110278). These should probably be NOPs if  `OPENDDS_NETWORK_CONFIG_MODIFIER` isn't defined.

java/dds/dcps_java.mpc
```diff
@@ -82,13 +83,22 @@ project: idl2jni, javah, dcpslib, optional_jni_check, dcps_java_optional, dcps_t
     OpenDDS/DCPS/transport/UnableToCreateException.java
   }
 
+  verbatim (gnuace, top, 1) {
+    opendds_android_network_callback = $(if $(android_target_api)>23,AndroidNetworkCallback.java,FakeAndroidNetworkCallback.java)
```
iguessthislldo, line 87: So I haven't thought about this before, but I don't think GNU make can't do integer comparisons. `$(if)` just checks the string length of the condition.

```make
all:
	echo $(if 1>3,ERROR,Good)
```

Results in `ERROR`

I think the only real option is use something like:

```make
has_android_network_callback ?=
ifdef android_target_api
  ifeq ($(shell expr $(android_target_api) '>=' 24),1)
    has_android_network_callback = 1
  endif
endif
all:
	echo $(if $(has_android_network_callback),Yes,No)
```

```
$ make -f test_makefile android_target_api=23
No
$ make -f test_makefile android_target_api=24
Yes
```

docs/android.md
```diff
@@ -153,7 +153,7 @@ those sections before configuring and building OpenDDS.
 
 ```Shell
 ./configure --doc-group --target=android --macros=ANDROID_ABI=armeabi-v7a
-    --macros=ANDROID_API_PATH=$SDK/platforms/android-$API
+    --macros=android_sdk=$SDK --macros=android_target_api=$API
```
iguessthislldo, line 156: This is just a minimal example and doesn't do anything because `--java` isn't being passed. If you want to keep it I would suggest moving this to the "Network Permissions and Availability" section.

docs/android.md
```diff
@@ -449,6 +449,14 @@ OpenDDS connections automatically when the device switches from one network
 (cellular or wifi) to another. Apps that need to know when a network change
 occurs can register with the [ConnectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager) for network callback events.
 
+As of API 30+, LinuxNetworkConfigMonitor can no longer be used, as Netlink 
+sockets are blocked by OS for security reasons. Instead, NetworkConfigModifier
+is utilized.  As a concequence of this, two new variables are required from the
```
iguessthislldo, line 454:
```suggestion
is utilized.  As a consequence of this, two new variables are required from the 
```

docs/android.md
```diff
@@ -449,6 +449,14 @@ OpenDDS connections automatically when the device switches from one network
 (cellular or wifi) to another. Apps that need to know when a network change
 occurs can register with the [ConnectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager) for network callback events.
 
+As of API 30+, LinuxNetworkConfigMonitor can no longer be used, as Netlink 
+sockets are blocked by OS for security reasons. Instead, NetworkConfigModifier
+is utilized.  As a concequence of this, two new variables are required from the 
+user, android_sdk, and android_target_api.  These correspond to the location of your
+sdk, likely /home/<username>/Android/Sdk on Linux, and the API number you are building
+for. The NetworkConfigModifier is set up along with the necessary network callbacks
+when the user uses "TheParticipantFactory.WithArgs".
```
iguessthislldo, line 458:
```suggestion
when the user uses `TheParticipantFactory.WithArgs`.
```

docs/android.md
```diff
@@ -449,6 +449,14 @@ OpenDDS connections automatically when the device switches from one network
 (cellular or wifi) to another. Apps that need to know when a network change
 occurs can register with the [ConnectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager) for network callback events.
 
+As of API 30+, LinuxNetworkConfigMonitor can no longer be used, as Netlink 
+sockets are blocked by OS for security reasons. Instead, NetworkConfigModifier
+is utilized.  As a concequence of this, two new variables are required from the 
+user, android_sdk, and android_target_api.  These correspond to the location of your
+sdk, likely /home/<username>/Android/Sdk on Linux, and the API number you are building
+for. The NetworkConfigModifier is set up along with the necessary network callbacks
```
iguessthislldo, line 457:
```suggestion
sdk, likely /home/<username>/Android/Sdk on Linux, and the API number you are targeting.
The NetworkConfigModifier is set up along with the necessary network callbacks
```

iguessthislldo (changes requested):

iguessthislldo:
After you fix the target api check I will run my matrix on your branch again.

java/dds/dcps_java.mpc
```diff
@@ -84,7 +84,7 @@ project: idl2jni, javah, dcpslib, optional_jni_check, dcps_java_optional, dcps_t
   }
 
   verbatim (gnuace, top, 1) {
-    opendds_android_network_callback = $(if $(android_target_api)>23,AndroidNetworkCallback.java,FakeAndroidNetworkCallback.java)
+    opendds_android_network_callback = $(if $(shell [ $(android_target_api) -ge 24 ] && echo "ok"),AndroidNetworkCallback.java,FakeAndroidNetworkCallback.java)
```
iguessthislldo, line 87: This doesn't work if `android_target_api` isn't defined.

dds/DCPS/NetworkConfigMonitor.cpp
```diff
@@ -138,6 +138,7 @@ void NetworkConfigMonitor::add_interface(const NetworkInterface& nic)
   for (Listeners::const_iterator pos = listeners.begin(), limit = listeners.end(); pos != limit; ++pos) {
     const RcHandle<NetworkConfigListener> listener(pos->lock());
     if (listener) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) NetworkConfigMontior::add_interface(%s)\n"), nic.name().c_str()));
```
iguessthislldo, line 141: Add debug level to this and the other `ACE_DEBUG`s here.

docs/android.md
```diff
@@ -447,6 +448,20 @@ OpenDDS connections automatically when the device switches from one network
...
+sockets are blocked by OS for security reasons. Instead, NetworkConfigModifier
+is utilized.  As a consequence of this, two new variables are required from the
+user, android_sdk, and android_target_api.  These correspond to the location of your
+sdk, likely /home/<username>/Android/Sdk on Linux, and the API number you are targeting.
+The NetworkConfigModifier is set up along with the necessary network callbacks
+when the user uses `TheParticipantFactory.WithArgs`.
+
+```Shell
+./configure --doc-group --target=android --macros=ANDROID_ABI=armeabi-v7a
+    --macros=android_sdk=$SDK --macros=android_target_api=$API --java
```
iguessthislldo, line 461: So the target API passed here will almost certainly be different from the minimum API that is passed to `make_standalone_toolchain.py`. I would suggest using a different variable name.

iguessthislldo (changes requested):
I think that's it after these two things. I will run my matrix and comment if there's any issues it catches.

docs/android.md
iguessthislldo: `$API` that is.

iguessthislldo:
[Builds are clean](https://github.com/iguessthislldo/OpenDDS-Android/actions/runs/464967726)

iguessthislldo (approved):

dds/DCPS/NetworkConfigMonitor.cpp
```diff
@@ -152,6 +155,9 @@ void NetworkConfigMonitor::remove_interface(int index)
     NetworkInterfaces::iterator pos = std::find_if(network_interfaces_.begin(), network_interfaces_.end(), NetworkInterfaceIndex(index));
     if (pos != network_interfaces_.end()) {
       nic = *pos;
+      if (DCPS_debug_level > 4) {
+        ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) NetworkConfigMontior::remove_interface(%s)\n"), nic.name().c_str()));
```
jwillemsen, line 159: Logging an ascii string should be done using %C

dds/DCPS/NetworkConfigMonitor.cpp
```diff
@@ -171,6 +177,9 @@ void NetworkConfigMonitor::remove_interface(const OPENDDS_STRING& name)
     NetworkInterfaces::iterator pos = std::find_if(network_interfaces_.begin(), network_interfaces_.end(), NetworkInterfaceName(name));
     if (pos != network_interfaces_.end()) {
       nic = *pos;
+      if (DCPS_debug_level > 4) {
+        ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) NetworkConfigMontior::remove_interface(%s)\n"), nic.name().c_str()));
```
jwillemsen, line 181: Logging an ascii string should be done using %C

PR #2238, opened 2021-01-04
Increase finish timeout in delayed durable test

pollocktoci:
When running the DelayedDurable test multiple time in a loop, changing “60” in the "finish" call to a larger number resulted in more iterations of the test passing before eventually erroring on a “Reader Missing Instance” error.  I'm thinking that the test might be shutting down before it has a chance to fully clean itself up (and that is just a guess).  Several other similar tests use a timeout of 300, so I used that value.


jwillemsen:
5 minutes to wait on a finish is very very long, that is a long time a build system is doing nothing

pollocktoci:
I also thought that was a long time to wait, but I noticed that 26 other tests use "finish(300)" and figured there might be some reason why such long timeouts were chosen (maybe to offset some anomaly in the test framework?).  I really think the system should always clean itself up no matter what the timeout value.

jrw972:
DelayedDurable only seems to be failing for large samples (fragmentation).  This includes the ASAN failure.  In general, I believe this is an issue with flow control, i.e., sender is overwhelming the receiver which is why increasing the timeout sometimes works.

PR #2241, opened 2021-01-04
Update DataSampleElement's pointers in its related lists

sonndinh:
- Some enqueue/dequeue functions in the `InstanceDataSampleList`, `WriterDataSampleList`, and `SendStateDataSampleList` classes currently haven't properly updated the corresponding pointers in the affected `DataSampleElement` nodes.
- This also removes redundant logic and does reformatting for those list classes.

### DIFF
--- a/dds/DCPS/SendStateDataSampleList.inl
+++ b/dds/DCPS/SendStateDataSampleList.inl
@@ -6,6 +6,7 @@
  */
 
 #include "DataSampleElement.h"
+
 #include <algorithm>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
@@ -130,10 +131,12 @@ ACE_INLINE
 void
 SendStateDataSampleList::remove(DataSampleElement* stale)
 {
-  if (stale->previous_send_sample_)
+  if (stale->previous_send_sample_) {
     stale->previous_send_sample_->next_send_sample_ = stale->next_send_sample_;
-  if (stale->next_send_sample_)
+  }
+  if (stale->next_send_sample_) {
     stale->next_send_sample_->previous_send_sample_ = stale->previous_send_sample_;
+  }
 }
 
 ACE_INLINE
--- a/dds/DCPS/WriterDataSampleList.inl
+++ b/dds/DCPS/WriterDataSampleList.inl
@@ -6,6 +6,7 @@
  */
 
 #include "DataSampleElement.h"
+
 #include <algorithm>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
@@ -58,14 +59,14 @@ WriterDataSampleList::enqueue_tail(const DataSampleElement* sample)
   // changed to accommodate const-correctness throughout.
   DataSampleElement* mSample = const_cast<DataSampleElement*>(sample);
 
-  //sample->previous_writer_sample_ = 0;
-  //sample->next_writer_sample_ = 0;
+  mSample->next_writer_sample_ = 0;
 
   ++size_ ;
 
   if (head_ == 0) {
     // First sample in the list.
     head_ = tail_ = mSample ;
+    mSample->previous_writer_sample_ = 0;
 
   } else {
     // Add to existing list.
--- a/dds/DCPS/InstanceDataSampleList.cpp
+++ b/dds/DCPS/InstanceDataSampleList.cpp
@@ -11,7 +11,7 @@
 #include "Definitions.h"
 #include "PublicationInstance.h"
 
-#include "dds/DCPS/transport/framework/TransportSendListener.h"
+#include "transport/framework/TransportSendListener.h"
 
 #if !defined (__ACE_INLINE__)
 #include "InstanceDataSampleList.inl"
--- a/dds/DCPS/InstanceDataSampleList.h
+++ b/dds/DCPS/InstanceDataSampleList.h
@@ -8,10 +8,11 @@
 #ifndef OPENDDS_DCPS_INSTANCEDATASAMPLELIST_H
 #define OPENDDS_DCPS_INSTANCEDATASAMPLELIST_H
 
-#include "dds/DdsDcpsInfoUtilsC.h"
 #include "Definitions.h"
-#include "transport/framework/TransportDefs.h"
 #include "Dynamic_Cached_Allocator_With_Overflow_T.h"
+
+#include "transport/framework/TransportDefs.h"
+#include "dds/DdsDcpsInfoUtilsC.h"
 
 #include <iterator>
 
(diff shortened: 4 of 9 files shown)


dds/DCPS/SendStateDataSampleList.inl
```diff
@@ -130,10 +131,12 @@ ACE_INLINE
 void
 SendStateDataSampleList::remove(DataSampleElement* stale)
```
kuznetsovmoci, line 132: Should this function decrement `size_`?

kuznetsovmoci (commented):
In general, I like 'enqueue' functions style in SendStateDataSampleList better, than in two other lists. To me it was more readable. Not sure if it's worth changing in this PR though.

dds/DCPS/SendStateDataSampleList.inl
sonndinh: This method is static and its description in the header file says it removes `stale` from whatever list object it belongs to. But you're right that `size_` of the affected list object must be decreased somewhere. I'm not sure if the callers of this method do the decrement. From a quick search, it seems to me that `size_` is not updated where this method is called. So this may be something to look further into.

sonndinh:
> In general, I like `enqueue` functions style in `SendStateDataSampleList` better, than in two other lists. To me it was more readable. Not sure if it's worth changing in this PR though.

Which specific enqueue function in `SendStateDataSampleList` do you think looks better? To me they're quite similar.

dds/DCPS/SendStateDataSampleList.inl
simpsont-oci: It looks like this method is only ever called by the WriteDataContainer when it can't figure out which list a "delivered" or "dropped" DataSampleElement came from... which I hope is a corner case we don't actually hit very often. Note in the comment description it admits that this invalidates the list this element comes from. As far as I'm concerned, this method shouldn't exist.

@mitza-oci Can you think of any reason why we couldn't remove this method (and it's calls) from the WriteDataContainer? Surely if we can't find the list a delivered/dropped sample came from, we should just log & ignore it rather than try to remove it from some list we don't own and/or that doesn't contain this sample anymore.

dds/DCPS/WriterDataSampleList.inl
```diff
@@ -58,14 +59,14 @@ WriterDataSampleList::enqueue_tail(const DataSampleElement* sample)
   // changed to accommodate const-correctness throughout.
   DataSampleElement* mSample = const_cast<DataSampleElement*>(sample);
 
-  //sample->previous_writer_sample_ = 0;
-  //sample->next_writer_sample_ = 0;
+  mSample->next_writer_sample_ = 0;
```
simpsont-oci, line 62: I'm wondering about why these lines were commented out. Is it possible anyone is using this function to enqueue multiple samples onto the tail by chaining them before passing them in (e.g. concatenating two lists)? If so, this would break that usage.

sonndinh: This function is called in `WriteDataContainer` and I think the input is a single element. Maybe these lines were commented out by mistake?

mitza-oci: @simpsont-oci I don't think the multiple samples use case would be expected due to `++size_`
@sonndinh You could check the revision history for when they were commented out.

dds/DCPS/SendStateDataSampleList.inl
mitza-oci: The comments are referring to `get_resend_data()` which to me looks different than "can't find the list".  But I agree with the general point, if we want a different way to handle this it is only used in two places in WriteDataContainer.

kuznetsovmoci:
> > In general, I like `enqueue` functions style in `SendStateDataSampleList` better, than in two other lists. To me it was more readable. Not sure if it's worth changing in this PR though.
> 
> Which specific enqueue function in `SendStateDataSampleList` do you think looks better? To me they're quite similar.

For example, two lists make an assumption:
 mSample->next_instance_sample_ = 0;
 
 and then conditionally override it.
 
 Whereas SendStateDataSampleList handles the assignments 'symmetrically' for both cases:
  if (head_ == 0) {
    head_ = tail_ = mSample;
    sample->next_send_sample_ = sample->previous_send_sample_ = 0;
  } else {
    sample->previous_send_sample_ = tail_;
    sample->next_send_sample_ = 0;
    tail_->next_send_sample_ = mSample;
    tail_ = mSample;
  }
  
  Could be just a personal preference, of course.

dds/DCPS/WriterDataSampleList.inl
sonndinh: The oldest commit I can find on GitHub was from Jul 03, 2014 (https://github.com/objectcomputing/OpenDDS/commit/304939ca0212b4ed75b2f215a86d085db3f7eee5#). Looks like these two lines were already commented out before the code was imported to GitHub.

mitza-oci: This repository has commits going back to the start of the project, before GitHub existed (probably?).  They were imported from SVN (and CVS before that).  It may be difficult to see where this came from due to renamed files.

oschwaldp-oci: That may be a carry over from when DataSampleList was split up into WriterDataSampleList, SendStateDataSampleList, and InstanceDataSampleList, which I believe may have all been based off of the DataSampleList initially.  Might have been left in, correctly or incorrectly, as a note on the historical implementation in the DataSampleList.

sonndinh:
> > > In general, I like `enqueue` functions style in `SendStateDataSampleList` better, than in two other lists. To me it was more readable. Not sure if it's worth changing in this PR though.
> > 
> > 
> > Which specific enqueue function in `SendStateDataSampleList` do you think looks better? To me they're quite similar.
> 
> For example, two lists make an assumption:
> mSample->next_instance_sample_ = 0;
> 
> and then conditionally override it.
> 
> Whereas SendStateDataSampleList handles the assignments 'symmetrically' for both cases:
> if (head_ == 0) {
> head_ = tail_ = mSample;
> ` sample->next_send_sample_ = sample->previous_send_sample_ = 0;`
> } else {
> `sample->previous_send_sample_ = tail_;`
> `sample->next_send_sample_ = 0;`
> tail_->next_send_sample_ = mSample;
> tail_ = mSample;
> }
> 
> Could be just a personal preference, of course.

The line `mSample->next_instance_sample_ = 0;` is just a common statement for both if and else branches. If it was moved to both the if and else blocks, the function would look the same as the `enqueue_tail` function of `SendStateDataSampleList`. So I think they are not much different and either way is ok to me.

dds/DCPS/SendStateDataSampleList.inl
simpsont-oci: I think it's OK to leave it for now, but only because I'm guessing our future transport framework changes are going to have us revisiting the WriteDataContainer & associated lists as well.

simpsont-oci (approved):
Things look OK to me as they are. I'd love to nuke some of this code and move towards more robust & general-purpose containers, but that's outside the scope of this PR.

kuznetsovmoci (approved):

PR #2247, opened 2021-01-06
[xtypes] Improve Thread Monitoring

iguessthislldo:
When using internal thread status BIT with conditions it's possible for one publish thread status to knock out the previous one before it can be read because the history depth is one and they are in the same instance.

Also:
- Refactor thread monitoring code by making common functions for generating the thread key and updating thread status.
- Create `OpenDDS::DCPS::String` to eventually replace `OPENDDS_STRING`.
- Other minor miscellaneous fixes.

### DIFF
--- a/dds/DCPS/GuidUtils.h
+++ b/dds/DCPS/GuidUtils.h
@@ -200,9 +200,7 @@ inline void assign(GUID_t& dest, const GUID_t& src)
 
 inline void assign(OctetArray16& dest, const GUID_t& src)
 {
-  std::memcpy(&dest[0], &src.guidPrefix, sizeof(GuidPrefix_t));
-  std::memcpy(&dest[sizeof(GuidPrefix_t)], &src.entityId.entityKey, sizeof(EntityKey_t));
-  dest[sizeof(GuidPrefix_t) + sizeof(EntityKey_t)] = src.entityId.entityKind;
+  std::memcpy(&dest[0], &src, sizeof(src));
 }
 
 inline RepoId make_id(const GuidPrefix_t& prefix, const EntityId_t& entity)
--- a/dds/DCPS/GuidUtils.h
+++ b/dds/DCPS/GuidUtils.h
@@ -193,17 +193,29 @@ inline void assign(GuidPrefix_t& dest, const GuidPrefix_t& src)
   std::memcpy(&dest[0], &src[0], sizeof(GuidPrefix_t));
 }
 
-inline DCPS::RepoId make_id(const GuidPrefix_t& prefix, const EntityId_t& entity)
+inline void assign(GUID_t& dest, const GUID_t& src)
 {
-  DCPS::RepoId id;
+  std::memcpy(&dest, &src, sizeof(GUID_t));
+}
+
+inline void assign(OctetArray16& dest, const GUID_t& src)
+{
+  std::memcpy(&dest[0], &src.guidPrefix, sizeof(GuidPrefix_t));
+  std::memcpy(&dest[sizeof(GuidPrefix_t)], &src.entityId.entityKey, sizeof(EntityKey_t));
+  dest[sizeof(GuidPrefix_t) + sizeof(EntityKey_t)] = src.entityId.entityKind;
+}
+
+inline RepoId make_id(const GuidPrefix_t& prefix, const EntityId_t& entity)
+{
+  RepoId id;
   assign(id.guidPrefix, prefix);
   id.entityId = entity;
   return id;
 }
 
-inline DCPS::RepoId make_id(const DCPS::RepoId& participant_id, const EntityId_t& entity)
+inline RepoId make_id(const RepoId& participant_id, const EntityId_t& entity)
 {
-  DCPS::RepoId id = participant_id;
+  RepoId id = participant_id;
   id.entityId = entity;
   return id;
 }
--- a/dds/DCPS/ReactorTask.cpp
+++ b/dds/DCPS/ReactorTask.cpp
@@ -262,13 +262,13 @@ bool ThreadStatus::update(
   return true;
 }
 
-String ThreadStatus::get_key(const char* fallback_tid, const String& name)
+String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
 {
-  ACE_UNUSED_ARG(fallback_tid);
   String key;
 #ifdef OPENDDS_SAFETY_PROFILE
-  key = fallback_tid;
+  key = safety_profile_tid;
 #else
+  ACE_UNUSED_ARG(safety_profile_tid);
 #  ifdef ACE_HAS_MAC_OSX
   unsigned long tid = 0;
   uint64_t u64_tid;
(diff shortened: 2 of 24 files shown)


dds/DCPS/Service_Participant.cpp
```diff
@@ -422,7 +422,7 @@ Service_Participant::get_domain_participant_factory(int &argc,
         }
       }
 
-#if defined OPENDDS_SAFETY_PROFILE && defined ACE_HAS_ALLOC_HOOKS
+#if OPENDD_POOL_ALLOCATOR
```
mitza-oci, line 425: typo here and below

mitza-oci (changes requested):

dds/DCPS/GuidUtils.h
```diff
@@ -193,17 +193,29 @@ inline void assign(GuidPrefix_t& dest, const GuidPrefix_t& src)
...
-inline DCPS::RepoId make_id(const GuidPrefix_t& prefix, const EntityId_t& entity)
+inline void assign(GUID_t& dest, const GUID_t& src)
 {
-  DCPS::RepoId id;
+  std::memcpy(&dest, &src, sizeof(GUID_t));
+}
+
+inline void assign(OctetArray16& dest, const GUID_t& src)
+{
+  std::memcpy(&dest[0], &src.guidPrefix, sizeof(GuidPrefix_t));
```
jrw972, line 203: We know that `sizeof(dest)` has to be `sizeof(GUID_t)`.  Change to single memcpy.

dds/DCPS/ReactorTask.cpp
```diff
@@ -258,4 +237,56 @@ OpenDDS::DCPS::ReactorTask::stop()
...
 
+bool ThreadStatus::update(
+  const String& thread_key, const SystemTimePoint& timestamp)
+{
+  ACE_WRITE_GUARD_RETURN(ACE_Thread_Mutex, g, lock, false);
+  Map::iterator it = map.find(thread_key);
+  const bool is_new = it == map.end();
+  if (is_new) {
+    it = map.insert(Map::value_type(thread_key, Thread())).first;
+  }
```
jrw972, line 248:
```suggestion
  Map::iterator it =map.insert(Map::value_type(thread_key, Thread())).first;

```

dds/DCPS/ReactorTask.h
```diff
@@ -46,7 +60,7 @@ public virtual RcObject {
   virtual ~ReactorTask();
 
 public:
-  int open_reactor_task(void*, TimeDuration timeout = TimeDuration(0), ThreadStatus* thread_stat = 0, OPENDDS_STRING name = "");
+  int open_reactor_task(void*, TimeDuration timeout = TimeDuration(0), ThreadStatus* thread_stat = 0, String name = "");
```
jrw972, line 63:
```suggestion
  int open_reactor_task(void*, TimeDuration timeout = TimeDuration(0), ThreadStatus* thread_stat = 0, const String& name = "");
```

dds/DCPS/GuidUtils.h
iguessthislldo: I know that's how we normally treat it, but `GUID_t` is a struct. Is it guaranteed that it won't have padding inside it?

mitza-oci: In this case, yes I think so.  Everything in it is struct/array of single bytes.

mitza-oci:
Looks like `java/tests/internal_thread_status` needs to be updated

dds/DCPS/GuidUtils.h
iguessthislldo: I guess if it's done like in CDR, still seems weird to me to assume how the compiler is going to layout an unpacked struct.

iguessthislldo: I will make the change and make a unit test that asserts the two sizes are the same.

jrw972 (approved):

mitza-oci:
Jenkins build failed: `libOpenDDS_Dcps.so: undefined reference to `ACE_OS::thr_gettid()'`

iguessthislldo:
> Jenkins build failed: libOpenDDS_Dcps.so: undefined reference to ACE_OS::thr_gettid()

Yeah I saw that. I don't know why it's doing that since https://github.com/DOCGroup/ACE_TAO/pull/1390 was merged.

mitza-oci:
> > Jenkins build failed: libOpenDDS_Dcps.so: undefined reference to ACE_OS::thr_gettid()
> 
> Yeah I saw that. I don't know why it's doing that since [DOCGroup/ACE_TAO#1390](https://github.com/DOCGroup/ACE_TAO/pull/1390) was merged.

The code you're adding to OpenDDS has to work with the latest release tag of ACE.  I think that's what Jenkins has.

iguessthislldo:
The rest of the CI works with `--ace-github-latest`, why is Jenkins different?

mitza-oci:
> The rest of the CI works with `--ace-github-latest`, why is Jenkins different?

Something should check that we don't break things.  It's good that this does for CI (pre-merge).  We also have scoreboard builds to check post-merge.

iguessthislldo:
> Something should check that we don't break things. It's good that this does for CI (pre-merge). We also have scoreboard builds to check post-merge.

Okay, but what if the PR requires changes to ACE/TAO that aren't in a release yet?

mitza-oci:
> > Something should check that we don't break things. It's good that this does for CI (pre-merge). We also have scoreboard builds to check post-merge.
> 
> Okay, but what if the PR requires changes to ACE/TAO that aren't in a release yet?

If that change is critical then a new DOCGroup release and a new OCITAO release have to happen.
If it's not critical to have, use an `#ifdef` and detect ACE's capabilities (we did this with the support for IDL/CDR "fixed" back a few years ago).

iguessthislldo:
The macro that's supposed to detect this feature doesn't work because the function isn't public. This is what I fixed in the PRs for ACE/TAO.

mitza-oci:
> The macro that's supposed to detect this feature doesn't work because the function isn't public. This is what I fixed in the PRs for ACE/TAO.

Since you are adding something new to ACE, you could add a preprocessor macro at the same time.  That way the user of ACE can detect if it's there.

Depending on what you want ACE to do, you could also just bring the relevant code into OpenDDS since this is all not worth the trouble.

dds/FACE/FaceTSS.cpp
```diff
@@ -80,7 +80,7 @@ void Initialize(const CONFIGURATION_RESOURCE configuration_file,
       return_code = INVALID_PARAM;
     } else {
       return_code = RC_NO_ERROR;
-#if defined OPENDDS_SAFETY_PROFILE && defined ACE_HAS_ALLOC_HOOKS
+#if OPENDD_POOL_ALLOCATOR
```
mitza-oci, line 83:
```suggestion
#if OPENDDS_POOL_ALLOCATOR
```

mitza-oci (approved):

PR #2248, opened 2021-01-06
Fix race condition in TransportRegistry::fix_empty_default()

simpsont-oci:
And a build dependency issue for DCPSInfoRepo and the TCP transport.

### DIFF
--- a/dds/InfoRepo/DCPSInfoRepo.mpc
+++ b/dds/InfoRepo/DCPSInfoRepo.mpc
@@ -1,4 +1,4 @@
-project(*Lib): dcpslib, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
+project(*Lib): dcpslib, dcps_tcp, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
   sharedname = OpenDDS_InfoRepoLib
   dynamicflags = OPENDDS_INFOREPOLIB_BUILD_DLL
   includes  += $(DDS_ROOT)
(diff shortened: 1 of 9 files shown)


dds/InfoRepo/DCPSInfoRepo.mpc
```diff
@@ -1,4 +1,4 @@
-project(*Lib): dcpslib, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
+project(*Lib): dcpslib, dcps_tcp, iortable, dcps_inforepodiscovery, DCPSInfoRepo_bits, imr_client, svc_utils, install_lib {
```
mitza-oci, line 1: Why does it need `dcps_tcp` even when built in topics are disabled?

mitza-oci (changes requested):

dds/InfoRepo/DCPSInfoRepo.mpc
simpsont-oci: I'm pretty sure I was getting a link error to TcpInst::get_local_address() without this.

simpsont-oci: There's a line in DCPSInfo_i.cpp: tcp_inst->local_address(listen_str);

mitza-oci: Why is that line compiled when built-in topics are disabled?

simpsont-oci: Because there are no #define guards around it. It doesn't look like init_transport method gets called when there's no support for BITs, but this is still producing a dependency on the tcp transport.

So what's the right fix here? Add the guards for all TCP-specific code and then make the MPC dependency upon tcp transport library contingent on BIT support? Or add the guards to the entirety of init_transport's internals and do the same?

simpsont-oci: Oh, I see the DCPSInfoRepo_bits.mpb now... I'll back this out and put in the #defines.

mitza-oci: Right, the DCPSInfoRepo doesn't need to use any transport implementations when Built-in Topics are excluded at build time.

> MPC dependency upon tcp transport library contingent on BIT support?

This is already done, see `DCPSInfoRepo_bits`

mitza-oci (approved):

PR #2249, opened 2021-01-06
max_bundle_size is unnecessary

jrw972:
Problem
-------

Applications setting max_message_size should set max_bundle_size.
Consequently, max_bundle_size should just be derived from
max_message_size to eliminate a configuration variable.

Solution
--------

Derive max_bundle_size from max_message_size.


jrw972:
This applies to SEDP as well.

simpsont-oci (approved):

iguessthislldo (approved):

PR #2262, opened 2021-01-11
Custom Workspace File Option for configure Script

iguessthislldo:


configure
```diff
@@ -176,6 +176,7 @@ my @specs = # Array of array-refs, each inner array is an option group which
       $argIndent . 'target DDS_ROOT to target ACE_ROOT)',
     'host-tools-only!', 'Just build the host tools (no)',
     'prefix=s', 'Installation prefix (none)',
+    'workspace=s', 'Custom MPC workspace file to copy and use (Use a builtin one)',
```
mitza-oci, line 179: Why is it copied?

iguessthislldo: MPC places the `GNUmakefile` file at the location of the mwc file instead of in OpenDDS root like I want.

mitza-oci: Yes, I was thinking that the user of this script would have already put the file in the current directory.

iguessthislldo: A user can do that too if they name the file `OpenDDS_custom.mwc`.

mitza-oci: Does this work using the changes on this PR?
```
cp /tmp/MyWorkspace.mwc .
./configure --workspace=MyWorkspace.mwc
```
If the answer is no, it seems like it should.
If the answer is yes, I'm not sure why the configure script needs to "copy and use," seems like it's doing more than necessary.

If the user would like to run `./configure --workspace=/somewhere/else/has/the/Workspace.mwc` then I don't see why it shouldn't just go ahead with it.

iguessthislldo: > ```
> cp /tmp/MyWorkspace.mwc .
> ./configure --workspace=MyWorkspace.mwc
> ```

Yes that works as well.

> If the user would like to run ./configure --workspace=/somewhere/else/has/the/Workspace.mwc then I don't see why it shouldn't just go ahead with it.

Without a copy there are a two issues:

- OpenDDS paths would always have to reference to `$(DDS_ROOT)` in the workspace file. This isn't terrible, but a user would have to know to do that since we don't do in our builtin workspace files.

- The `GNUmakefile` created at non-local workspace location needs `setenv` set. Yet again not terrible, but also another thing that's different from normal.

Even ignoring these avoiding creating that out of source make file and saving an extra copy command for something I'm going to use quite a bit seem like more than enough reason to copy the file.

mitza-oci: Right, this would be an advanced use case where the user will have to deal with a lot more details than just using the configure script as-is.  Depending on the sort of build system being used there are different details to be aware of.

As you noted, contents of workspace files can be relative paths.  This is a choice made by the author of the workspace.  If this is used then the workspace is essentially context-dependent.  So copying the file adds additional complexity for understanding the workspace - is it relative to where it was written or where it was copied to?  This complexity is what I'm asking us to consider avoiding in the `configure` script.  Let it do the minimal number of steps to get the functionality we need.

iguessthislldo: >  is it relative to where it was written or where it was copied to?

I think that part is pretty clear, the help message says it's copying it before use.

PR #2263, opened 2021-01-11
Transport receive strategy doesn't cleanup buffers

jrw972:
Problem
-------

The transport receive strategy doesn't cleanup buffers (allocated with
`handle_simple_dds_input`).

Solution
--------

Delete the buffers in the destructor.

### DIFF
--- a/dds/DCPS/transport/framework/TransportReceiveStrategy_T.cpp
+++ b/dds/DCPS/transport/framework/TransportReceiveStrategy_T.cpp
@@ -62,6 +62,15 @@ TransportReceiveStrategy<TH, DSH>::~TransportReceiveStrategy()
                  size));
     }
   }
+
+  for (int index = 0; index < RECEIVE_BUFFERS; ++index) {
+    if (receive_buffers_[index] != 0) {
+      ACE_DES_FREE(
+                   receive_buffers_[index],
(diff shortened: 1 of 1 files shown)


dds/DCPS/transport/framework/TransportReceiveStrategy_T.cpp
```diff
@@ -62,6 +62,15 @@ TransportReceiveStrategy<TH, DSH>::~TransportReceiveStrategy()
                  size));
     }
   }
+
+  for (int index = 0; index < RECEIVE_BUFFERS; ++index) {
```
jwillemsen, line 66:
``` suggestion
 for (size_t index = 0; index < RECEIVE_BUFFERS; ++index) {
```

mitza-oci: This could result in a signed/unsigned comparison warning.  The type of `RECEIVE_BUFFERS` is int.

jwillemsen: It seems both are used in existing code for RECEIVE_BUFFERS, int and size_t

mitza-oci (approved):

PR #2270, opened 2021-01-12
Xtypes xcdr1 warnings

kuznetsovmoci:
We need to have some sort of runtime warning when a user tries to use XCDR1 Appendable and a warning or error when a user tries to use XCDR1 mutable.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -3312,6 +3312,67 @@ DataReaderImpl::get_ice_endpoint()
   return TransportClient::get_ice_endpoint();
 }
 
+DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
+{
+  const DDS::DataRepresentationIdSeq repIds =
+    get_effective_data_rep_qos(qos_.representation.value, true);
+  bool success = false;
+  if (cdr_encapsulation()) {
+    for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
+      Encoding::Kind encoding_kind;
+      if (repr_to_encoding_kind(repIds[i], encoding_kind)) {
+        if (Encoding::KIND_XCDR2 == encoding_kind || Encoding::KIND_XCDR1 == encoding_kind) {
+          if (encoding_kind == Encoding::KIND_XCDR1) {
+            if (get_max_extensibility() == MUTABLE) {
+              if (::OpenDDS::DCPS::DCPS_debug_level) {
+                ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                  ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                  ACE_TEXT("Encountered unsupported combination of XCDR1 encoding and mutable extensibility\n")));
+              }
+              return DDS::RETCODE_ERROR;
+            }
+          }
+          decoding_modes_.insert(encoding_kind);
+          success = true;
+        } else if (DCPS_debug_level >= 2) {
+          // Supported but incompatible data representation
+          ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+                     ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                     ACE_TEXT("Skip %C data representation.\n"),
+                     Encoding::kind_to_string(encoding_kind).c_str()));
+        }
+      } else if (DCPS_debug_level) {
+        // Unsupported or unknown data representation
+        ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: ")
+                   ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                   ACE_TEXT("Encountered unsupported or unknown data representation.\n")));
+      }
+    }
+  } else {
+    decoding_modes_.insert(Encoding::KIND_UNALIGNED_CDR);
+    success = true;
+  }
+  if (!success) {
+    if (DCPS_debug_level) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
+                 ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+                 ACE_TEXT("Could not find a valid data representation.\n")));
+    }
+    return DDS::RETCODE_ERROR;
+  }
+  if (DCPS_debug_level >= 2) {
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) ")
+               ACE_TEXT("DataReaderImpl::setup_deserialization: ")
+               ACE_TEXT("Setup successfully with data representations: ")));
+    OPENDDS_SET(Encoding::Kind)::iterator it = decoding_modes_.begin();
+    for (; it != decoding_modes_.end(); ++it) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("%C "), Encoding::kind_to_string(*it).c_str()));
+    }
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT(".\n")));
+  }
+
+  return DDS::RETCODE_OK;
+}
 
 void DataReaderImpl::accept_sample_processing(const SubscriptionInstance_rch& instance, const DataSampleHeader& header, bool is_new_instance)
 {
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -3312,57 +3312,6 @@ DataReaderImpl::get_ice_endpoint()
   return TransportClient::get_ice_endpoint();
 }
 
-//DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
-//{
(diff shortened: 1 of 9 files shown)


dds/DCPS/DataWriterImpl_T.h
```diff
@@ -333,6 +333,15 @@ DataWriterImpl_T
         if (repr_to_encoding_kind(repIds[i], encoding_kind)) {
           if (Encoding::KIND_XCDR2 == encoding_kind || Encoding::KIND_XCDR1 == encoding_kind) {
             encoding_mode_ = EncodingMode(encoding_kind, swap_bytes());
+            if (encoding_kind == Encoding::KIND_XCDR1) {
+              if (MarshalTraitsType::extensibility_level() == MUTABLE) {
+                ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: ")
```
iguessthislldo, line 338: Needs to check for `DCPS_debug_level` first.

dds/idl/marshal_generator.cpp
```diff
@@ -2642,6 +2680,29 @@ namespace {
         "Unexpected extensibility while generating MarshalTraits", node);
       return false;
     }
+    be_global->header_ << "; }\n";
+
+    /*
+     * This is used for topic type extensibility level.
+     */
+    const ExtensibilityKind ek = extensibility_kind(struct_node != 0 ? struct_node : union_node);
```
iguessthislldo, line 2688:
```suggestion
    const ExtensibilityKind ek = extensibility_kind(node);
```

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -897,6 +897,67 @@ namespace OpenDDS {
     return inst;
   }
 
+  DDS::ReturnCode_t setup_deserialization()
```
iguessthislldo, line 900: I'd like to to avoid adding more functions to headers when possible in the long run because they slow down compilation. Instead you could add a virtual method to `DataReaderImpl` to get the max extensibility.

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
kuznetsovmoci: Have to dynamic cast, but then it would work

dds/idl/marshal_generator.cpp
```diff
@@ -1630,6 +1630,44 @@ namespace {
...
 
+  ExtensibilityKind extensibility_kind(AST_Type* type, bool root = true)
+  {
+    static ExtensibilityKind ek;
+    static std::vector<AST_Type*> type_stack;
+
+    if (root) {
+      type_stack.clear();
+      ek = extensibilitykind_final;
+    }
```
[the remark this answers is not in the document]
kuznetsovmoci: Would it be faster to skip sibling SubData if it has been processed once (there is a check for this), and its extensibility was applied towards calculation of the maximum extensibility?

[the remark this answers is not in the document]
iguessthislldo: I don't think that's an optimization that's worth taking the time to set up.

dds/idl/marshal_generator.cpp
iguessthislldo: Okay, I thought `node` was `AST_Type`. It probably should be, but it's fine how it is for now.

dds/idl/marshal_generator.cpp
```diff
@@ -1630,6 +1630,46 @@ namespace {
...
+      const Fields fields(fld_cls & CL_STRUCTURE ? dynamic_cast<AST_Structure*>(type) : dynamic_cast<AST_Union*>(type));
+      const Fields::Iterator fields_end = fields.end();
+      for (Fields::Iterator i = fields.begin(); i != fields_end; ++i) {
+        if (extensibilitykind_mutable == max_extensibility_kind(type, false)) {
+          return extensibilitykind_mutable;
+        }
+      }
+      return max_exten;
+    } else if (fld_cls & CL_ARRAY || fld_cls & CL_SEQUENCE) {
+      return max_extensibility_kind(type, false);
```
iguessthislldo, line 1667: Need to use base type. As it is it will go into a recursive loop because `type` will always be the same.

```suggestion
    } else if (fld_cls & CL_ARRAY) {
      return max_extensibility_kind(dynamic_cast<AST_Array*>(type)->base_type(), false);
    } else if (fld_cls & CL_SEQUENCE) {
      return max_extensibility_kind(dynamic_cast<AST_Sequence*>(type)->base_type(), false);
```

dds/idl/marshal_generator.cpp
```diff
@@ -2642,6 +2682,29 @@ namespace {
...
       return false;
     }
+    be_global->header_ << "; }\n";
+
+    /*
+     * This is used for topic type extensibility level.
+     */
+    const ExtensibilityKind ek = max_extensibility_kind(dynamic_cast<AST_Type*>(node));
+    be_global->header_ <<
+      "  static Extensibility extensibility_level() { return ";
```
iguessthislldo, line 2692:
```suggestion
      "  static Extensibility max_extensibility_level() { return ";
```

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
```diff
@@ -1630,6 +1630,44 @@ namespace {
     return bounded;
   }
 
+  ExtensibilityKind extensibility_kind(AST_Type* type, bool root = true)
+  {
+    static ExtensibilityKind ek;
```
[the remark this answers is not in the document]
kuznetsovmoci: What would be a better alternative?

dds/idl/marshal_generator.cpp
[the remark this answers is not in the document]
kuznetsovmoci: Added `pops`

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -117,6 +117,41 @@ bool TypeLookupService::type_object_in_cache(const TypeIdentifier& ti) const
...
 
+bool TypeLookupService::extensibility(TypeFlag extensibility_mask, const TypeIdentifier& type_id) const
+{
+  bool result = false;
+  TypeObject to = get_type_objects(type_id);
+
+  if (TK_UNION == to.minimal.kind) {
+    result = to.minimal.union_type.union_flags & extensibility_mask;
+  }
+  else if (TK_STRUCTURE == to.minimal.kind) {
```
iguessthislldo, line 128:
```suggestion
  } else if (TK_STRUCTURE == to.minimal.kind) {
```

dds/idl/marshal_generator.cpp
```diff
@@ -1630,6 +1630,54 @@ namespace {
     return bounded;
   }
 
+  ExtensibilityKind max_extensibility_kind(AST_Type* type, bool root = true)
+  {
+    static ExtensibilityKind max_exten;
+    static std::vector<AST_Type*> type_stack;
+
+    if (root) {
+      type_stack.clear();
```
iguessthislldo, line 1639:
```suggestion
```

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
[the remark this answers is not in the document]
iguessthislldo: Returning the max extensibility like you're already doing.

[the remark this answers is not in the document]
iguessthislldo: You also wouldn't need the root argument if you did that.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3320,6 +3320,16 @@ DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
     for (CORBA::ULong i = 0; i < repIds.length(); ++i) {
       Encoding::Kind encoding_kind;
       if (repr_to_encoding_kind(repIds[i], encoding_kind)) {
+          if (encoding_kind == Encoding::KIND_XCDR1) {
+            if (get_max_extensibility() == MUTABLE) {
```
iguessthislldo, line 3324:
```suggestion
          if (encoding_kind == Encoding::KIND_XCDR1 && get_max_extensibility() == MUTABLE) {
```

Also the indentation seems messed up.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -332,6 +332,17 @@ DataWriterImpl_T
         Encoding::Kind encoding_kind;
         if (repr_to_encoding_kind(repIds[i], encoding_kind)) {
           encoding_mode_ = EncodingMode(encoding_kind, swap_bytes());
+            if (encoding_kind == Encoding::KIND_XCDR1) {
+              if (MarshalTraitsType::max_extensibility_level() == MUTABLE) {
```
iguessthislldo, line 336:
```suggestion
            if (encoding_kind == Encoding::KIND_XCDR1 && MarshalTraitsType::max_extensibility_level() == MUTABLE) {
```

Same thing about the indentation.

iguessthislldo (changes requested):

PR #2271, opened 2021-01-13
Fixed compiler warnings.

pollocktoci:


tests/security/attributes/publisher.cpp
```diff
@@ -95,7 +95,7 @@ int run_test(int argc, ACE_TCHAR *argv[], Args& my_args)
       if (my_args.secure_part_user_data_) {
         part_qos.user_data.value.length(static_cast<unsigned int>(part_user_data_string.size()));
         for (size_t i = 0; i < part_user_data_string.size(); ++i) {
-          part_qos.user_data.value[i] = part_user_data_string[i];
+          part_qos.user_data.value[i] = static_cast<unsigned int>(part_user_data_string[i]);
```
iguessthislldo, line 98: I'm pretty sure the warning was about the `i` in  `part_qos.user_data.value[i]` (`size_t` to `CORBA::ULong`). `part_qos.user_data.value[i] = part_user_data_string[i]` itself shouldn't need a cast.

pollocktoci: Ha!  That's what I meant to do.  Thanks.

iguessthislldo (approved):

PR #2272, opened 2021-01-13
Port for SPDP local address not respected

jrw972:
Problem
-------

SPDP ignores the port of the SPDP local address.  Setting the SPDP
local address to just a port means that the request port number is
effectively ignored.

Solution
--------

Use the port of the SPDP local address so long as it is not zero.


jwillemsen:
Something to test? The TAO portspan test does something similar, test if the requested port is really used

iguessthislldo (approved):

PR #2273, opened 2021-01-13
LinuxNetworkConfigMonitor error is now a warning for android

ClaytonCalabrese:
Reporting it as an error was bothering users as it gave the impression something unexpected had occurred in OpenDDS.

### DIFF
--- a/dds/DCPS/LinuxNetworkConfigMonitor.cpp
+++ b/dds/DCPS/LinuxNetworkConfigMonitor.cpp
@@ -40,8 +40,8 @@ bool LinuxNetworkConfigMonitor::open()
   if (socket_.open(addr, AF_NETLINK, NETLINK_ROUTE) != 0) {
 #ifdef ACE_ANDROID
     if (DCPS_debug_level >= 1) {
-      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: LinuxNetworkConfigMonitor::open: could not open socket"
-        " this is expected for API30+\n")));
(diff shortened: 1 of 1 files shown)


dds/DCPS/LinuxNetworkConfigMonitor.cpp
```diff
@@ -38,7 +38,14 @@ bool LinuxNetworkConfigMonitor::open()
   ACE_Netlink_Addr addr;
   addr.set(pid, RTMGRP_NOTIFY | RTMGRP_LINK | RTMGRP_IPV4_IFADDR | RTMGRP_IPV6_IFADDR);
   if (socket_.open(addr, AF_NETLINK, NETLINK_ROUTE) != 0) {
+#ifdef ACE_ANDROID
+    if (DCPS_debug_level >= 1) {
+      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: LinuxNetworkConfigMonitor::open: could not open socket"
+        " this is expected for API30+\n")));
```
iguessthislldo, line 44: `ACE_TEXT` can't span multiple strings since all it does is expand to `L` or nothing to change if the literal is wide or narrow. Either add another `ACE_TEXT` to the second literal or remove it entirely. Personally I've been leaving them off for a while.

iguessthislldo (changes requested):

dds/DCPS/LinuxNetworkConfigMonitor.cpp
mitza-oci, line 44:
```suggestion
      ACE_DEBUG((LM_WARNING, "(%P|%t) WARNING: LinuxNetworkConfigMonitor::open: could not open socket"
        " this is expected for API30+\n"));
```

PR #2283, opened 2021-01-15
Don't add the directory, at the moment we use a subdirectory for the …

jwillemsen:
…TypeSupport.idl we need to use gendir already (see ReadCondition test)

    * MPC/modules/TYPESUPPORTHelper.pm:


jwillemsen:
Looks windows-2019_defaults_artifact failed related to vcpkg, not our change

PR #2284, opened 2021-01-15
DomainRange test fails on flea

ClaytonCalabrese:
Domain 10 is problematic no matter what on Flea, I couldn't find why, and even after a restart it still fails if domain 10 is attempted to be used. It also appears domain 11 is problematic.

Additionally, having at least 5 readers with 5 domains fails on Flea.


jonesc-oci (approved):
We should consider modifying the test to check for invalid domain and numbers of participants configurations.

simpsont-oci (approved):

PR #2289, opened 2021-01-15
Fix Workflow Permissions Issues and Add ACE_TAO Artifact Caching

simpsont-oci:
The permissions for the previous round of GitHub actions workflows didn't work for pull requests, and according to [this article](https://securitylab.github.com/research/github-actions-preventing-pwn-requests), switching to pull_request_target was problematic for other reasons. This PR adjusts the workflows to avoid security vulnerabilities by splitting it into three separate workflows (cancel previous, build & test, and check results). It also adds ACE_TAO artifact caching for a small performance improvement.


simpsont-oci:
This should be ready. The only two test failures were fan_frag, which is still a problem for CI vms.

PR #2169, opened 2020-12-11
generate correct initializer for arrays in the C++11 mapping, trigge…

jwillemsen:
…rs a warning with gcc 4.8.5

    * dds/idl/langmap_generator.cpp:

Fixes warning given by gcc 4.8.5, for example:
[Details] /home/build/jenkins/workspace/rhel75_opendds_versioned/BUILD/DOC_ROOT/DDS/tests/DCPS/Compiler/C++11/idl_test1_lib/FooDefC.h:293:27: warning: missing initializer for member ‘std::array<std::basic_string<char>, 5ul>::_M_elems’ [-Wmissing-field-initializers]

PR #2293, opened 2021-01-16
Restored previous approach to C++11 array initialization

mitza-oci:
Fixes failing test in #2292 by partially reverting #2190 and #2169.


jwillemsen:
Reverting this will lead to compile warnings with gcc 4.8.x. For a nested array of pod also {{}} should be used.

mitza-oci:
> 
> 
> Reverting this will lead to compile warnings with gcc 4.8.x. Tested this further, looks it can be much simpler
> 
> ```
> if (af.cls_ & CL_ARRAY) {
>   initializer = "{{}}";
> }
> ```

We can have additional changes in a subsequent PR.

PR #2298, opened 2021-01-18
Moved bench2 handle_report and associated functions to common location.

pollocktoci:


performance-tests/bench_2/common/util.cpp
```diff
@@ -137,4 +140,327 @@ uint32_t one_at_a_time_hash(const uint8_t* key, size_t length)
...
+{
+  for (CORBA::ULong i = 0; i < reported_tags.length(); ++i) {
+    const std::string tag(reported_tags[i]);
+    if (input_tags.find(tag) != input_tags.end()) {
+      stats[tag] = consolidate(stats[tag], data.to_simple_stat_block());
+    }
+  }
+}
+
+int handle_report(const Bench::TestController::Report& report,
```
simpsont-oci, line 187: The behavior of handle_reports is unique to the test_controller, as it's writing out results to the screen and returning with an exit code. I don't think this is generic enough to move over.  Instead, I was thinking parts of this function, specifically the parts that loop over the parts of the report, could be factored out into common code for use in a scraping tool. Looking at this now, I'm not sure it will make sense for test_controller to use whatever we factor out since it will be faster to collect all the stats at once if we're interested in all the stats.

But for the scraping tool, I think we may sometimes be interested in one stat at a time, so we could have functions that, for example, loop over all datareader & datawriter reports and consolidates stat blocks of certain names (e.g. "discovery_delta" and "round_trip_latency") that match a tagged expression (e.g. "'signal' AND NOT 'best effort' or 'any'). Or, as another example, we could have a function that loops over all node controller reports and collects stat blocks of a certain name (e.g. "cpu_percent") and returns the individual blocks in a sequence with associated node names, so we can plot them individually.

All that to say, I think we'd better move this back into test_controller until we figure out what our common collection functions need to look like. Otherwise, the rest of this looks fine.

simpsont-oci (changes requested):

performance-tests/bench_2/common/util.cpp
pollocktoci: By "the rest of this looks file", I think you're saying we should move update_stats_for_tags, update_details_for_tags and consolidate_tagged_stats into util, but keep handle_report in main.  Is that correct?

simpsont-oci:
Yes.

performance-tests/bench_2/test_controller/main.cpp
```diff
@@ -785,12 +739,12 @@ int handle_report(const Bench::TestController::Report& report,
...
+    total_undermatched_writers ||
+    total_out_of_order_data_count ||
+    total_duplicate_data_count ||
+    total_missing_data_count ||
+    missing_durable_data) {
     result = EXIT_FAILURE;
   }
   return result;
-}
+}
```
simpsont-oci, line 750:
```suggestion
}

```

simpsont-oci:
The 'bench_2' directory rename to 'bench' already happened on master, so we'll need to resolve those conflicts before we can merge this.

simpsont-oci (approved):

PR #2299, opened 2021-01-18
lint.pl: Additional Checks and Improvements

iguessthislldo:
Continued from #2161 because apparently we can't retarget from a closed branch.

- Added the following checks and changed OpenDDS to pass these checks:

  - `whitespace_before_newline_in_string`: ex: `"This is a string \n"`

  - `missing_include_guard`: Checks IDL and header files for include guards matching one of two patterns. Both start with the file path, take `dds/` off the front, turn anything that's not alphanumeric or an underscore to underscores, uppercase everything and prepend `OPENDDS_`. The difference between them is one converts from camel case to snake case while the other doesn't. For example, `dds/DdsDcps.idl` can be `OPENDDS_DDSDCPS_IDL` or `OPENDDS_DDS_DCPS_IDL`.

  - `nonrelative_include_path`: Checks to see if an include statement path contains the including file's directory in it. For example if `foo/bar.h` has ` #include <foo/util/stuff.h>`, it should be `#include "util/stuff.h"`.

- Added `--try-fix` to try to fix issues. So far this only works for the checks with `strip_fix` property set and the `missing_include_guard` check if there's an existing include guard.

- Way to pass arguments to ACE `fuzz.pl`.

- Add more info to `--help` message.

- :rainbow: ANSI Escape Code Colors in messages :rainbow: Can be controlled using command line arguments and environment variables.

- For checks using `line_matched` that support it, highlight what the offending characters of the line are in a separate error message.

- Rename `not_exactly_one_eof_newline` check to `eof_newline_count`. Made it accept a blank line at the end of text files. Also fixed an issue where it actually wasn't catching when there were no newlines at the end of the file. This means there must be 1 or 2 newline characters at the end of a file now.

- Checks are now listed and happen in consistent order now (alphabetically) and a proper error message is displayed when an invalid check is passed as an argument.

- Ability to check paths outside of OpenDDS, assuming the check isn't OpenDDS specific.


iguessthislldo:
Undid IDL relative includes because https://github.com/DOCGroup/ACE_TAO/pull/1357 and https://github.com/DOCGroup/ACE_TAO/pull/1366 still aren't merged.

mitza-oci (commented):
`dds/DCPS/QOS_XML_Handler/dds_qos.hpp` is a generated file, do we want to change it in this PR?

`dds/DCPS/security/CryptoBuiltIn.idl` has no use for multi-include guards.  No other code includes this.

`dds/DCPS/transport/multicast/Multicast_Export.h` and others like it are also generated

iguessthislldo:
> dds/DCPS/QOS_XML_Handler/dds_qos.hpp is a generated file, do we want to change it in this PR?

I don't consider `DDS_QOS_HPP` to be a sufficient include guard name, so it's be nice if we could change it.

> dds/DCPS/security/CryptoBuiltIn.idl has no use for multi-include guards. No other code includes this.

I can exclude it from the check if you're sure that it's never going to need an include guard.

> dds/DCPS/transport/multicast/Multicast_Export.h and others like it are also generated

Have we ever had to regenerate export headers before?

mitza-oci:
> > dds/DCPS/QOS_XML_Handler/dds_qos.hpp is a generated file, do we want to change it in this PR?
> 
> I don't consider `DDS_QOS_HPP` to be a sufficient include guard name, so it'd be nice if we could change it.

It's not ideal.  I don't mind changing it but we need to be aware of that.  Maybe add a README in this directory, which should have one anyway?

> > dds/DCPS/transport/multicast/Multicast_Export.h and others like it are also generated
> 
> Have we ever had to regenerate export headers before?

No, but we could in the future, in order to adapt to a (hypothetical) change in ACE or in how OpenDDS uses ACE.

Aside from that case, what I was primarily thinking of was that if someone creates a new export header for a new shared library, will it immediately be a lint error?

iguessthislldo:
> > > dds/DCPS/QOS_XML_Handler/dds_qos.hpp is a generated file, do we want to change it in this PR?
> > 
> > 
> > I don't consider `DDS_QOS_HPP` to be a sufficient include guard name, so it'd be nice if we could change it.
> 
> It's not ideal. I don't mind changing it but we need to be aware of that. Maybe add a README in this directory, which should have one anyway?

Yes, that sounds fine with me.

> > > dds/DCPS/transport/multicast/Multicast_Export.h and others like it are also generated
> > 
> > 
> > Have we ever had to regenerate export headers before?
> 
> No, but we could in the future, in order to adapt to a (hypothetical) change in ACE or in how OpenDDS uses ACE.
> 
> Aside from that case, what I was primarily thinking of was that if someone creates a new export header for a new shared library, will it immediately be a lint error?

Yes, it would be an immediate lint error if there isn't a way to make the script use one of the names the script is looking for. I think I'm fine with excluding them from the check because with the exception of the security library all of includes guards seem start with `OPENDDS`. The thing is that as is if someone creates a new export header in `dds` they'll have to exclude it from the check.

mitza-oci:
> > Aside from that case, what I was primarily thinking of was that if someone creates a new export header for a new shared library, will it immediately be a lint error?
> 
> Yes, it would be an immediate lint error if there isn't a way to make the script use one of the names the script is looking for. I think I'm fine with excluding them from the check because with the exception of the security library all of includes guards seem start with `OPENDDS`. The thing is that as is if someone creates a new export header in `dds` they'll have to exclude it from the check.

It will start with `OPENDDS` as long as the parameter given on the command line starts with `OpenDDS`, which it should.  So it seem like there are two alternatives and I'm OK with either:

1. Exclude from lint check
2. Put the steps to create a new export header in `guidelines.md`

iguessthislldo:
I undid most of the export changes, but changed the names of `QOS_XML_Handler` and `DdsSecurity` exports to start with `OpenDDS`. I changed the script to detect and ignore the export headers, so new ones will also be ignored.

dds/DCPS/QOS_XML_Handler/README.md
```diff
@@ -0,0 +1,9 @@
+# QoS XML Handler
+
+Optional utility to allow one to define the QoS of DDS unities using XML.
```
mitza-oci, line 3:
```suggestion
Optional utility to allow one to define the QoS of DDS enities using XML.
```

iguessthislldo: I think I hit the wrong suggestion in the spell check.

mitza-oci (approved):

iguessthislldo:
I don't think I understand the Windows errors.

mitza-oci:
> 
> 
> I don't think I understand the Windows errors.

Update the `dynamicflags` setting in qos_xml_handler.mpc.

PR #2304, opened 2021-01-19
InternalThreadStatus Test: Clean Up Includes

iguessthislldo:


ClaytonCalabrese:
I noticed this test being broken when building for my TC Static PR. It seems like #include "dds/DCPS/StaticIncludes.h" isn't doing what we want anymore.  Another example of this is the BuiltInTopicTest rtps_disc having trouble finding the specified transport.  Might be a more widespread issue?

mitza-oci (approved):

mitza-oci:
> I noticed this test being broken when building for my DataRepresentation Static PR. It seems like #include "dds/DCPS/StaticIncludes.h" isn't doing what we want anymore. Another example of this is the BuiltInTopicTest rtps_disc having trouble finding the specified transport. Might be a more widespread issue?

The comment at the top of `StaticIncludes.h` describes what it does.

PR #2312, opened 2021-01-20
Converting IDL data types to/from JSON requires special codegen

jrw972:
Problem
-------

The existing approach to converting an IDL datatype to/from JSON is
based on code generated by the IDL compiler that is specific to the
JSON conversion.

Solution
--------

Define the ValueReader interface and a JsonValueReader implementation.
(ValueWriter and JsonValueWriter already existed.)  The
ValueReader/ValueWriter interfaces present a generic interface for
deserializers/serializers.  The IDL compiler has the responsiblity of
define generic procedures that use these interfaces to populate a
value (vread) or recite a value (vwrite).

This change creates a number of short-term and long-term
opportunities:

1. The rapidjson pass can be removed from the IDL compiler.

2. The v8 pass can be removed by implementing ValueReader and
ValueWriter for the v8 environment.  Presumably, these implementations
would reside in node-opendds repository.

3. Support for CDR/XCDR (de)serialization can be converted to the
ValueReader/ValueWriter interfaces.

4. (3) supposes manual implemenations of vread and vwrite for type
objects and type identifiers.  These would correspond to the existing
serializer support.

4. A C++ ValueWriter implementation can be defined and used to
generate C++ code that re-creates a value.  This could be used to
great effect in the type object pass to generate the type object and
type identifier output.

5. JsonValueWriter could be applied to type objects and type
identifiers to create an IDL to JSON pass to replace ITL.  Json schema
support may be interesting to consider as well.


dds/DCPS/JsonValueWriter.cpp
```diff
@@ -1,230 +0,0 @@
...
-  // TODO: Escape.
-  out_ << '"' << name << '"';
-}
-
-} // namespace DCPS
-} // namespace OpenDDS
-
-OPENDDS_END_VERSIONED_NAMESPACE_DECL
-
-#endif
```
iguessthislldo, line 230: Why delete the cpp file?

iguessthislldo: Nevermind, I see they're templates now.

iguessthislldo:
You updated the rapidjson submodule, is there is something you needed in a newer version?

jrw972:
> You updated the rapidjson submodule, is there is something you needed in a newer version?

Yes.  The iterative parsing methods are new.

tests/DCPS/UnitTests/JsonValueReader.cpp
```diff
@@ -0,0 +1,528 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include "dds/DCPS/JsonValueReader.h"
+
+#include "gtest/gtest.h"
```
iguessthislldo, line 8:
```suggestion
#include <dds/DCPS/JsonValueReader.h?

#include <gtest/gtest.h>
```

tests/DCPS/UnitTests/JsonValueReader.cpp
```diff
@@ -0,0 +1,528 @@
...
+  StringStream ss(json);
+  JsonValueReader<> jvr(ss);
+  EXPECT_TRUE(jvr.begin_sequence());
+  EXPECT_FALSE(jvr.elements_remaining());
+  EXPECT_TRUE(jvr.end_sequence());
+}
+
+TEST(JsonValueReader, struct_max)
+{
+  const char json[] = "{\"bool\":true,\"byte\":255,\"int8\":127,\"uint8\":255,\"int16\":32767,\"uint16\":65535,\"int32\":2147483647,\"uint32\":4294967295,\"int64\":9223372036854775807,\"uint64\":18446744073709551615,\"float32\":1.25,\"float64\":1.25,\"float128\":1.25,\"char8\":97,\"char16\":97,\"string\":\"a string\",\"enum\":\"kValue1\"}";
```
iguessthislldo, line 100: Break down this line?

iguessthislldo (changes requested):

dds/DCPS/ValueReader.h
```diff
@@ -0,0 +1,182 @@
...
+namespace DCPS {
+
+// A ValueReader produces events and values for the recitation of a
+// value.  To use it, one manually or automatically, e.g., code
+// generation in the IDL compiler, defines a vread function for a
+// given type V.
+//
+//   bool vread(ValueReader& vw, V& value)
+//
+// The vreade function should invoke the appropriate methods of the
```
mitza-oci, line 32:
```suggestion
// The vread function should invoke the appropriate methods of the
```

dds/idl/value_reader_generator.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+  void array_helper(const std::string& expression, AST_Array* array,
+                    size_t dim_idx, const std::string& idx, int level)
+  {
+    const bool use_cxx11 = be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+    const std::string indent(level * 2, ' ');
+    if (dim_idx < array->n_dims()) {
+      const size_t dim = array->dims()[dim_idx]->ev()->u.ulval;
+      be_global->impl_ <<
+        indent << "if (!value_reader.begin_array()) return false;;\n" <<
```
mitza-oci, line 33:
```suggestion
        indent << "if (!value_reader.begin_array()) return false;\n" <<
```

tests/DCPS/UnitTests/JsonValueReader.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+TEST(JsonValueReader, from_json)
+{
+  const char json[] = "{\"bool\":true}";
+  StringStream ss(json);
+  MyStruct s;
+  EXPECT_TRUE(from_json(s, ss));
+}
+
+//bool from_json(T& value, const std::string& sample)
```
mitza-oci, line 521: ?

dds/DCPS/JsonValueReader.h
```diff
@@ -0,0 +1,569 @@
...
+    kStartObject,
+    kKey,
+    kEndObject,
+    kStartArray,
+    kEndArray,
+    kError,
+    kEnd
+  };
+
+  void peek() {
```
mitza-oci, line 122: Should this function return the token's type?  That may simplify the code that calls it.

mitza-oci (changes requested):

dds/DCPS/JsonValueWriter.h
```diff
@@ -74,53 +86,259 @@ class OpenDDS_Dcps_Export JsonValueWriter : public ValueWriter {
...
+}
+
+template <typename Buffer>
+void JsonValueWriter<Buffer>::write_string(const ACE_CDR::Char* value)
+{
+  writer_.String(value);
+}
+
+template <typename Buffer>
+void JsonValueWriter<Buffer>::write_wstring(const ACE_CDR::WChar* value)
```
mitza-oci, line 288:
```suggestion
void JsonValueWriter<Buffer>::write_wstring(const ACE_CDR::WChar*)
```
due to warning

tests/DCPS/UnitTests/JsonValueReader.cpp
```diff
@@ -0,0 +1,528 @@
...
+  EXPECT_TRUE(jvr.end_element());
+
+  EXPECT_TRUE(jvr.begin_element());
+  EXPECT_TRUE(jvr.read_uint32(uint32_value));
+  EXPECT_EQ(uint32_value, 0);
+  EXPECT_TRUE(jvr.end_element());
+
+  EXPECT_TRUE(jvr.begin_element());
+  EXPECT_TRUE(jvr.read_int64(int64_value));
+  EXPECT_EQ(int64_value, -9223372036854775808);
```
mitza-oci, line 316: ` warning: integer constant is so large that it is unsigned`

tests/DCPS/UnitTests/JsonValueReader.cpp
```diff
@@ -0,0 +1,528 @@
...
+  EXPECT_TRUE(jvr.end_element());
+
+  EXPECT_TRUE(jvr.begin_element());
+  EXPECT_TRUE(jvr.read_uint64(uint64_value));
+  EXPECT_EQ(uint64_value, 0);
```
mitza-oci, line 321: `warning: comparison between signed and unsigned integer expressions`

mitza-oci (changes requested):

dds/DCPS/JsonValueReader.h
jrw972: I don't think so.  You are either going to do
`const TokenType x = peek();` which is unnecessary since the logical return type of peek is the member `token_type_` or you are going to do
`if (peek() == X)` which only gains a line a put a modifying statement in an if which I tend to not like.

iguessthislldo (approved):

dds/DCPS/JsonValueReader.h
mitza-oci: I was thinking of this usage:
```
  switch (peek()) {
  case kInt: 
    value = int_value_;
    return consume(kInt);
  case kUint:
    value = uint_value_;
    return consume(kUint);
  }
```

jrw972: That makes sense.

PR #2317, opened 2021-01-21
XTypes test fixes

kuznetsovmoci:


tests/DCPS/XTypes/XTypes.h
```diff
@@ -28,5 +28,36 @@ enum AdditionalFieldValue {
...
 
+template<typename T>
+bool get_topic(T ts, const DomainParticipant_var dp, const std::string& topic_name,
+  Topic_var& topic, const std::string& registered_type_name)
+{
+  if (ts->register_type(dp, registered_type_name.empty() ? "" : registered_type_name.c_str()) != DDS::RETCODE_OK) {
+    ACE_ERROR((LM_ERROR, "ERROR: register_type failed\n"));
+    return false;
+  }
+  CORBA::String_var type_name = (registered_type_name.empty() ? ts->get_type_name() : registered_type_name.c_str());
```
mitza-oci, line 39: This is not addressing the leak found by valgrind/asan.

When using the `?:` operator the compiler needs to come up with a common type between the two expressions, and it's not going to do what we want it to.  

The usual rule for using `_var` for memory management applies: the result of an allocating call (`get_type_name()` in this case) should be directly put in a `_var`.

tests/DCPS/XTypes/run_test.pl
```diff
@@ -100,7 +100,7 @@ sub run_test {
 if ($test_name eq '') {
   while (my ($k, $v) = each %params) {
     run_test ($v, $k);
-    sleep(7);
+    sleep(10);
```
mitza-oci, line 103: Why is any amount of delay needed here?

mitza-oci (changes requested):

tests/DCPS/XTypes/run_test.pl
kuznetsovmoci: The readers' wait sets start timing out if "too many" tests are running simultaneously. Locally 7s works, but Katydid shows this symptom with 7s. Fred and I discussed it and decided to try increasing it to see if it would help. There is an underlying reason that still needs to be investigated.

PR #2321, opened 2021-01-21
Fix a couple of warnings on bench-2

sonndinh:


mitza-oci (approved):

simpsont-oci (approved):
I've got a few warnings cleaned up for bench in PR #2313 too but I think only the static_cast is an overlap.

sonndinh:
> I've got a few warnings cleaned up for bench in PR #2313 too but I think only the static_cast is an overlap.

Oh okay. I wasn't aware of that. This is more specific to the Jira task 128 (https://opendds.atlassian.net/jira/software/projects/OPENDDS/boards/10?selectedIssue=OPENDDS-128). If the fix in this PR is covered in PR #2313, I'm good with closing it.

simpsont-oci:
I think it's OK to merge this and I can handle any others or conflicts in
my PR.

PR #2322, opened 2021-01-22
type_to_default patch for long double

ClaytonCalabrese:
Now creates a temp var that is initialized then assigned to the member.  Context is needed for non unions to prevent multiple declarations.

### DIFF
--- a/dds/idl/dds_generator.cpp
+++ b/dds/idl/dds_generator.cpp
@@ -332,11 +332,13 @@ string type_to_default(const std::string& indent, AST_Type* type, const string& 
       if (use_cxx11) {
         def_val = "0.0L";
       } else {
+		    string temp_val = "ACE_CDR::LongDouble val = ACE_CDR_LONG_DOUBLE_INITIALIZER";
         if (is_union) {
-          def_val = "0";
+          def_val = "val";
+		  return indent + temp_val + ";\n" + indent + name + pre + def_val + post + ";\n";
         } else {
-          def_val = "ACE_CDR_LONG_DOUBLE_ASSIGNMENT(" + name + ", 0)";
-          return indent + def_val + ";\n";
+          def_val += "ACE_CDR_LONG_DOUBLE_ASSIGNMENT(" + name + ", val)";
+          return indent + "{\n  " + indent + temp_val + ";\n  " + indent + def_val + ";\n" + indent + "}\n";
         }
       }
     } else {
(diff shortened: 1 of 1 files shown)


dds/idl/dds_generator.cpp
```diff
@@ -332,11 +332,13 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
       if (use_cxx11) {
         def_val = "0.0L";
       } else {
+		    string temp_val = "ACE_CDR::LongDouble val = ACE_CDR_LONG_DOUBLE_INITIALIZER";
```
mitza-oci, line 335: Indent looks wrong... tabs?

dds/idl/dds_generator.cpp
```diff
@@ -332,11 +332,13 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
...
         if (is_union) {
-          def_val = "ACE_CDR_LONG_DOUBLE_INITIALIZER";
+          def_val = "val";
+		  return indent + temp_val + ";\n" + indent + name + pre + def_val + post + ";\n";
```
mitza-oci, line 338: Indent looks wrong... tabs?

dds/idl/dds_generator.cpp
```diff
@@ -332,11 +332,13 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
...
         } else {
-          def_val = "ACE_CDR_LONG_DOUBLE_ASSIGNMENT(" + name + ", 0)";
-          return indent + def_val + ";\n";
+          def_val += "ACE_CDR_LONG_DOUBLE_ASSIGNMENT(" + name + ", val)";
+          return indent + "{\n  " + indent + temp_val + ";\n  " + indent + def_val + ";\n" + indent + "}\n";
```
mitza-oci, line 341:
```suggestion
          return indent + "{\n"
            "  " + indent + temp_val + ";\n"
            "  " + indent + def_val + ";\n" + 
            indent + "}\n";
```

mitza-oci (changes requested):

dds/idl/dds_generator.cpp
ClaytonCalabrese: It doesn't look like this for me. I don't know how this is happening on github as it isn't what is on my branch.

ClaytonCalabrese: Nevermind, you are right there were tabs which were interpreted differently locally than on github.  Weird.

iguessthislldo: Which I think is the main reason for not using tabs. Tabs are traditionally 8 spaces which Github is doing, but editors can be configured to display them differently.

ClaytonCalabrese: Okay, thank you.

mitza-oci (approved):

PR #2330, opened 2021-01-22
Changed XType test flow to event driven, removing Sleep statements.

kuznetsovmoci:
Reviewed the way that run_test.pl in tests/DCPS/XTypes uses the perl test framework and what it expects in terms of timing of tests.  Sleep is not necessary since the Perl script can determine when the child process(es) exit.

In addition to the Perl part of this issue, replace sleep in the C++ test code with the use of WaitSet and Conditions.

### DIFF
--- a/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
+++ b/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
@@ -263,7 +263,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
 
   DDS::ConditionSeq conditions;
   DDS::Duration_t timeout = { 10, 0 };
-  if ((ws->wait(conditions, timeout) != DDS::RETCODE_OK) && expect_to_match) {
+  if (ws->wait(conditions, timeout) != DDS::RETCODE_OK) {
     ACE_ERROR((LM_ERROR, "ERROR: %C condition wait failed for type %C\n",
       expect_to_match ? "PUBLICATION_MATCHED_STATUS" : "INCONSISTENT_TOPIC_STATUS", type.c_str()));
     failed = 1;
--- a/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
+++ b/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
@@ -237,7 +237,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     AppendableStructWithDependencyTypeSupport_var ts = new AppendableStructWithDependencyTypeSupportImpl;
     failed = !get_topic(ts, dp, topic_name, topic, registered_type_name);
   } else {
-    ACE_ERROR((LM_ERROR, "ERROR: Type %s is not supported\n", type.c_str()));
+    ACE_ERROR((LM_ERROR, "ERROR: Type %C is not supported\n", type.c_str()));
     return 1;
   }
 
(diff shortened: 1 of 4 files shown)


tests/DCPS/XTypes/Sub/XTypesSubscriber.cpp
```diff
@@ -275,9 +275,23 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+  DDS::StatusCondition_var condition = expect_to_match ? dr->get_statuscondition() : topic->get_statuscondition();
+  condition->set_enabled_statuses(expect_to_match ? DDS::SUBSCRIPTION_MATCHED_STATUS : DDS::INCONSISTENT_TOPIC_STATUS);
+  DDS::WaitSet_var ws = new DDS::WaitSet;
+  ws->attach_condition(condition);
 
-  failed = !check_inconsistent_topic_status(topic);
+  DDS::ConditionSeq conditions;
+  DDS::Duration_t timeout = { 10, 0 };
+  if (ws->wait(conditions, timeout) != DDS::RETCODE_OK) {
+    ACE_ERROR((LM_ERROR, "ERROR: wait failed for type %s\n", type.c_str()));
```
iguessthislldo, line 286: Maybe say which wait failed, `SUBSCRIPTION_MATCHED_STATUS` or `INCONSISTENT_TOPIC_STATUS`?

mitza-oci, line 286: ACE logging uses `%C` for `const char*` arguments.  Please check the entire test for this.

iguessthislldo (approved):

tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
```diff
@@ -256,12 +256,27 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+  DDS::WaitSet_var ws = new DDS::WaitSet;
+  ws->attach_condition(condition);
+
+  DDS::ConditionSeq conditions;
+  DDS::Duration_t timeout = { 10, 0 };
+  if ((ws->wait(conditions, timeout) != DDS::RETCODE_OK) && expect_to_match) {
+    ACE_ERROR((LM_ERROR, "ERROR: %C condition wait failed for type %C\n",
+      expect_to_match ? "PUBLICATION_MATCHED_STATUS" : "INCONSISTENT_TOPIC_STATUS", type.c_str()));
+    failed = 1;
+  }
```
sonndinh, line 270:
```suggestion
  if (ws->wait(conditions, timeout) != DDS::RETCODE_OK) {
    ACE_ERROR((LM_ERROR, "ERROR: %C condition wait failed for type %C\n",
      expect_to_match ? "PUBLICATION_MATCHED_STATUS" : "INCONSISTENT_TOPIC_STATUS", type.c_str()));
    failed = 1;
  }
```
It won't print the log message if `expect_to_match == false` and the `wait` call fails.

sonndinh (commented):
Can you remove the command-line options `--writer` and `--reader` on both publisher and subscriber as they are now not needed?

tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
kuznetsovmoci: Good catch - leftover from the original logic. Thanks!

PR #2338, opened 2021-01-25
Volatile reader can ack beyond expected maximum sequence number

jrw972:
Problem
-------

If multiple participants are exchanging tokens at once, a volatile
reader may ack a sequence number beyond what was expected.  This case
is not handled by the current leader/lagger tracking.

Solution
--------

Allow a volatile reader to ack beyond it's expected max and adjust
it's expected max if it does so.


mitza-oci (approved):

ClaytonCalabrese (approved):
looks like it fixed the problem on cricket

PR #2350, opened 2021-01-26
opendds_idl: use a leading :: to qualify generated C++ scoped names

mitza-oci:


ClaytonCalabrese (approved):
Code looks fine but I'm wondering if this might require changes in other places.  It looks like Java has its own idea of scoped_helper.

mitza-oci:
> 
> 
> Code looks fine but I'm wondering if this might require changes in other places. It looks like Java has its own idea of scoped_helper.

There may be some remaining issues with code generated for the Java bindings, but it hasn't been a problem to this point.  We can address it in a different PR if it becomes a problem.

iguessthislldo (approved):

PR #2353, opened 2021-01-27
DDS foundation link replaces OMG Wiki link

jwillemsen:
www.omgwiki.org/dds redirects to the DDS Foundation

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -47,7 +47,8 @@
           <li><a href="/support.html">Support</a></li>
           <li><a href="{{ site.data.opendds.mailing_lists_uri }}">Mailing lists</a></li>
           <li><a href="/faq.html">FAQ</a></li>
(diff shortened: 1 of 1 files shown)


sonndinh (approved):

_includes/footer.html
```diff
@@ -47,7 +47,8 @@ <h3>Community</h3>
           <li><a href="/support.html">Support</a></li>
           <li><a href="{{ site.data.opendds.mailing_lists_uri }}">Mailing lists</a></li>
           <li><a href="/faq.html">FAQ</a></li>
-          <li><a href="https://www.omgwiki.org/dds/">DDS Standard</a></li>
+          <li><a href="https://www.omg.org/spec/DDS/">DDS Standard</a></li>
```
mitza-oci, line 50: I don't think we need the spec link here, it's already on https://opendds.org/documents.

mitza-oci (changes requested):

PR #2358, opened 2021-01-28
Fix Loop in WriteDataContainer::wait_pending

iguessthislldo:


dds/DCPS/ConditionVariable.h
```diff
@@ -59,43 +60,52 @@ class ConditionVariable {
...
     }
     return CvStatus_Error;
   }
 
+  /// Block until woken up or for expire_in. Same as wait() if expire_in is zero.
   CvStatus wait_for(const TimeDuration& expire_in)
   {
+    if (expire_in.is_zero()) {
+      return wait();
+    }
```
jrw972, line 85: This is the only part that worries me but I don't see anything that is calling this function.  Is that correct?

jrw972 (approved):

dds/DCPS/ConditionVariable.h
iguessthislldo: `wait_for`? Yea I don't think anything calls it. It's here because `std::condition_variable` has it, but don't think I ever took the time to see if I could refactor anything to use it. Is the part you're worried about that someone might want 0 to mean to timeout immediately?

iguessthislldo:
Should it and MessageTracker also break out of the loop when there's an error?

mitza-oci:
> 
> 
> Should it and MessageTracker also break out of the loop when there's an error?

Yes, it looks like breaking out of the loop matches the previous version of this function.

PR #2386, opened 2021-02-03
add correctly named news post about xtypes webinar

kroegerk:


adamsj-oci:
Is the weblink going to be open or should it be behind a registration wall?

PR #2387, opened 2021-02-03
Removed special case encoding of ReliabilityQosPolicy's kind in SEDP

kuznetsovmoci:
RTPS spec version 2.3 Table 9.13 has footnote 3 that describes how the "kind" field of ReliabilityQosPolicy needs special treatment.  Currently we are reusing the same struct from DCPS inside RTPS Discovery, however this is not necessary (since it goes through ParameterListConverter anyway) and causes problems for XTypes (TryConstruct is putting special-case checks into the code generator).


mitza-oci:
The encoding on the wire needs to be the same as before this change.  The goal is just to simplify the internal handling of it.

mitza-oci:
Replaced by #2401

PR #2362, opened 2021-01-28
get_minimal_type_map_private call not thread safe for cpp03

ClaytonCalabrese:
The get_minimal_type_map function call in typeobject_generator.cpp was not thread safe and was possibly the cause of an error occurring on bee where TLS would fail.  This is a fix to make that section of code thread safe.  I did not see the TLS error in my testing on bee.

PR #2389, opened 2021-02-04
Initialization of TypeIdentifier is not thread-safe

jrw972:
Problem
-------

Initialization of statics is not thread-safe on all compilers.  See #2362.

Solution
--------

Add lock to serialize initialization of type identifier.

### DIFF
--- a/dds/idl/typeobject_generator.cpp
+++ b/dds/idl/typeobject_generator.cpp
@@ -938,7 +938,7 @@ typeobject_generator::gen_epilogue()
   be_global->impl_ <<
     "const XTypes::TypeMap& get_minimal_type_map()\n"
     "{\n"
-    "  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, TheServiceParticipant->get_tm_lock(), tm);\n"
+    "  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, TheServiceParticipant->get_tm_lock(), XTypes::TypeMap());\n"
     "  static XTypes::TypeMap tm;\n"
     "  if (tm.empty()) {\n"
     "    tm = get_minimal_type_map_private();\n"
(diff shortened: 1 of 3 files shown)


ClaytonCalabrese (commented):
Do we want a different lock for this or is the TypeMap one fine?

jrw972:
I'll defer to @mitza-oci  on that one.

iguessthislldo:
> Initialization of statics is not thread-safe on all compilers. See

See what?

dds/idl/typeobject_generator.cpp
```diff
@@ -1762,7 +1762,11 @@ typeobject_generator::generate(AST_Type* node, UTL_ScopedName* name)
     gti.endArgs();
     const OpenDDS::XTypes::TypeIdentifier ti = get_minimal_type_identifier(node);
     be_global->impl_ <<
-      "  static const XTypes::TypeIdentifier ti = " << ti << ";\n"
+      "  static XTypes::TypeIdentifier ti;\n"
+      "  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, TheServiceParticipant->get_tm_lock(), ti);\n"
+      "  if (ti.kind() == XTypes::TK_NONE) {\n"
```
mitza-oci, line 1767: Move lock within the `if`?

jrw972: One thread could return a partially constructed type identifier if the lock is within the `if`.  If anything, the lock should be first.  I'll fix the other instance as well.

mitza-oci: Sorry, I misread it.

dds/idl/typeobject_generator.cpp
```diff
@@ -1762,7 +1762,11 @@ typeobject_generator::generate(AST_Type* node, UTL_ScopedName* name)
     gti.endArgs();
     const OpenDDS::XTypes::TypeIdentifier ti = get_minimal_type_identifier(node);
     be_global->impl_ <<
-      "  static const XTypes::TypeIdentifier ti = " << ti << ";\n"
+      "  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, TheServiceParticipant->get_tm_lock(), ti);\n"
+      "  static XTypes::TypeIdentifier ti;\n"
```
mitza-oci, line 1766: Now this looks like a problem because (in the generated code) we have ti referenced before it's defined.

mitza-oci:
> 
> 
> > Initialization of statics is not thread-safe on all compilers. See
> 
> See what?

Looks like it was updated after the initial post.

mitza-oci:
> 
> 
> I'll defer to @mitza-oci on that one.

I think it's fine to have a single lock handling this initialization.  We don't expect it to be in the critical path of data transmission, it should just be at startup.  So the goal here is for correctness on all platforms, not necessarily maximum scalability.

The name `get_tm_lock` and comment around it in Service_Participant should be updated.

mitza-oci (approved):

ClaytonCalabrese (approved):

PR #2392, opened 2021-02-04
Tweak delete_datareader Precondition Warning

iguessthislldo:
Mainly to say the topic name in case the reader is builtin, because it would have an unknown guid.

### DIFF
--- a/dds/DCPS/SubscriberImpl.cpp
+++ b/dds/DCPS/SubscriberImpl.cpp
@@ -29,10 +29,10 @@
 #include "transport/framework/TransportImpl.h"
 #include "transport/framework/DataLinkSet.h"
 #include "DCPS_Utils.h"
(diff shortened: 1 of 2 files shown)


dds/DCPS/SubscriberImpl.cpp
```diff
@@ -382,7 +380,7 @@ SubscriberImpl::delete_contained_entities()
   // mark that the entity is being deleted
   set_deleted(true);
 
-  ACE_Vector<DDS::DataReader_ptr> drs;
+  std::vector<DDS::DataReader*> drs;
```
mitza-oci, line 383: Use allocator-enabled container

mitza-oci (changes requested):

mitza-oci (approved):

PR #2396, opened 2021-02-07
Prevent ACE from mangling RtpsRelay's argc value

simpsont-oci:


jrw972:
Out of curiosity, why was this necessary?

jrw972 (approved):

simpsont-oci:
> Out of curiosity, why was this necessary?

On normal char builds, the ACE_Argv_Type_Converter is a pass-through and ACE_Arg_Shifter modifies both the original argc and argv when it finds arguments it cares about. On wchar builds, ACE_Argv_Type_Converter builds up a char* copy of argv, and the ACE_Arg_Shifter modifies the original argc, but modifies the _copy_ of argv... which means that when ServiceParticipant goes to inspect the original argv, the argc is wrong.

There are arguably saner ways to handle this, but this one seemed like the smallest viable change. If we're worried about overlap between RtpsRelay and ServiceParticipant arguments, we may need to make sure the arguments taken out of the copied argv are also taken out of the original argv (and argc updated), but that didn't seem to be an issue here.

PR #2401, opened 2021-02-08
Removed special case encoding of ReliabilityQosPolicy's kind in SEDP

kuznetsovmoci:
RTPS spec version 2.3 Table 9.13 has footnote 3 that describes how the "kind" field of ReliabilityQosPolicy needs special treatment. Currently we are reusing the same struct from DCPS inside RTPS Discovery, however this is not necessary (since it goes through ParameterListConverter anyway) and causes problems for XTypes (TryConstruct is putting special-case checks into the code generator).

### DIFF
--- a/tests/DCPS/UnitTests/ut_ParameterListConverter.cpp
+++ b/tests/DCPS/UnitTests/ut_ParameterListConverter.cpp
@@ -1595,8 +1595,7 @@ ACE_TMAIN(int, ACE_TCHAR*[])
       TEST_ASSERT(to_param_list(writer_data, param_list, true, type_info, map));
       TEST_ASSERT(is_present(param_list, PID_RELIABILITY));
       Parameter param = get(param_list, PID_RELIABILITY);
-      const int reliable_interop = static_cast<int>(RELIABLE_RELIABILITY_QOS) + 1;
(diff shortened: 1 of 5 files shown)


tests/DCPS/UnitTests/ut_ParameterListConverter.cpp
```diff
@@ -1595,8 +1595,7 @@ ACE_TMAIN(int, ACE_TCHAR*[])
       TEST_ASSERT(to_param_list(writer_data, param_list, true, type_info, map));
       TEST_ASSERT(is_present(param_list, PID_RELIABILITY));
       Parameter param = get(param_list, PID_RELIABILITY);
-      const int reliable_interop = static_cast<int>(RELIABLE_RELIABILITY_QOS) + 1;
-      TEST_ASSERT(static_cast<int>(param.reliability().kind) == reliable_interop);
+      TEST_ASSERT(static_cast<int>(param.reliability().kind.value) == RELIABLE);
```
mitza-oci, line 1598: The cast isn't necessary, it's already that type

mitza-oci (changes requested):

mitza-oci (approved):

PR #2402, opened 2021-02-09
RTPS lacks flow control

jrw972:
Problem
-------

The current implementation of reliability for RTPS is timer driven.
When the timer fires, heartbeats are sent to the readers.  The readers
respond with acknacks with nacks leading to data being resent.  If the
network delay exceeds the heartbeat timer period, then multiple
heartbeats and nacks can be in flight simultaneously.  This causes
data to be resent multiple times which leads to further delays in
processing.

Solution
--------

Record the readers that caused non-final heartbeats to be sent and
measure the delay of the network as the time required to receive an
acknack from all of the readers.  If all the acknacks were received,
recompute the heartbeat period as the measured delay increased by a
safety factor.  Otherwise, increase the heartbeat period by a backoff
factor.


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3579,21 +3608,28 @@ RtpsUdpDataLink::send_heartbeats(const DCPS::MonotonicTimePoint& /*now*/)
...
   }
+  heartbeat_period_ = std::min(heartbeat_period_, cfg.heartbeat_period_maximum_);
+  heartbeat_period_ = std::max(heartbeat_period_, cfg.heartbeat_period_minimum_);
+  expected_acks_ = 0;
 
   using namespace OpenDDS::RTPS;
 
   MetaSubmessageVec meta_submessages;
 
+  bool keep_going = false;
```
mitza-oci, line 3625: Should `keep_going` impact the `for` loop two lines down?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3755,7 +3800,10 @@ RtpsUdpDataLink::RtpsWriter::gather_heartbeats(OPENDDS_VECTOR(TransportQueueElem
   if (!preassociation_readers_.empty()) {
     for (ReaderInfoSet::const_iterator pos = preassociation_readers_.begin(), limit = preassociation_readers_.end();
          pos != limit; ++pos) {
-      gather_preassociation_heartbeat_i(meta_submessages, meta_submessage, *pos);
+      const ReaderInfo_rch& reader = (*pos);
```
mitza-oci, line 3803:
```suggestion
      const ReaderInfo_rch& reader = *pos;
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3808,6 +3860,8 @@ RtpsUdpDataLink::RtpsWriter::gather_heartbeats(OPENDDS_VECTOR(TransportQueueElem
     }
     readers_expecting_heartbeat_.clear();
   }
+
+  return !(additional_guids.empty() && preassociation_readers_.empty() && lagging_readers_.empty() && readers_expecting_heartbeat_.empty());
```
mitza-oci, line 3864: Not sure it would matter, but this might be better as `!a.empty() || !b.empty()...` for short-circuiting

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I've pushed code that initializes `keep_going`.  Does that answer your question?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2788,6 +2790,16 @@ RtpsUdpDataLink::RtpsWriter::process_acknack(const RTPS::AckNackSubmessage& ackn
     return;
   }
 
+  if (reader->expecting_ack_) {
+    reader->expecting_ack_ = false;
+    --link->expected_acks_;
```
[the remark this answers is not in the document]
jrw972: Agreed.

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -164,6 +168,28 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
               value.c_str(), rtps_name.c_str()), -1);
           }
           config->quick_resend_ratio(ratio);
+        } else if (name == "SedpHeartbeatBackoffFactor") {
+          const OPENDDS_STRING& value = it->second;
+          double ratio;
+          if (!DCPS::convertToDouble(value, ratio)) {
+            ACE_ERROR_RETURN((LM_ERROR,
+              ACE_TEXT("(%P|%t) RtpsDiscovery::Config::discovery_config(): ")
+              ACE_TEXT("Invalid entry (%C) for  in ")
```
mitza-oci, line 177: Missing a word between "for" and "in" (also on line 188)

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
```diff
@@ -36,8 +36,12 @@ RtpsUdpInst::RtpsUdpInst(const OPENDDS_STRING& name)
   , max_message_size_(RtpsUdpSendStrategy::UDP_MAX_MESSAGE_SIZE)
   , nak_depth_(0)
   , quick_reply_ratio_(0.1)
+  , heartbeat_backoff_factor_(2.0)
+  , heartbeat_safety_factor_(2.0)
```
simpsont-oci, line 40: Should this also be 1.1?

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -595,6 +621,30 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
                               ACE_TEXT("[rtps_discovery/%C] section.\n"),
                               string_value.c_str(), rtps_name.c_str()), -1);
           }
+        } else if (name == "SedpHeartbeatPeriodMinimum") {
```
simpsont-oci, line 624: Does it make sense to assert / complain if minimum > maximum? Or maybe if backoff factor < safety factor?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -80,9 +80,10 @@ RtpsUdpDataLink::RtpsUdpDataLink(RtpsUdpTransport& transport,
   , job_queue_(DCPS::make_rch<DCPS::JobQueue>(reactor_task->get_reactor()))
   , multi_buff_(this, config.nak_depth_)
   , best_effort_heartbeat_count_(0)
-  , nack_reply_(this, &RtpsUdpDataLink::send_nack_replies,
-                config.nak_response_delay_)
-  , heartbeat_(reactor_task->interceptor(), config.heartbeat_period_, *this, &RtpsUdpDataLink::send_heartbeats)
+  , nack_reply_(reactor_task->interceptor(), *this, &RtpsUdpDataLink::send_nack_replies)
+  , expected_acks_(0)
```
simpsont-oci, line 84: Does it make sense to move the timer and expected_acks_ field to the individual writers? I think we're (definitely) going to want it there eventually... without it, a writer publishing high speed data attached to a single reader will have his HB frequency decided by the acks to a writer publishing low speed status updates, etc.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3755,7 +3787,10 @@ RtpsUdpDataLink::RtpsWriter::gather_heartbeats(OPENDDS_VECTOR(TransportQueueElem
   if (!preassociation_readers_.empty()) {
     for (ReaderInfoSet::const_iterator pos = preassociation_readers_.begin(), limit = preassociation_readers_.end();
          pos != limit; ++pos) {
-      gather_preassociation_heartbeat_i(meta_submessages, meta_submessage, *pos);
+      const ReaderInfo_rch& reader = *pos;
+      gather_preassociation_heartbeat_i(meta_submessages, meta_submessage, reader);
+      reader->expecting_ack_ = true;
+      ++link->expected_acks_;
```
simpsont-oci, line 3793: Lock link->heartbeat_mutex_? Or add a helper method for it? Same for increments below.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3808,6 +3847,8 @@ RtpsUdpDataLink::RtpsWriter::gather_heartbeats(OPENDDS_VECTOR(TransportQueueElem
     }
     readers_expecting_heartbeat_.clear();
   }
+
+  return !additional_guids.empty() || !preassociation_readers_.empty() || !lagging_readers_.empty() || !readers_expecting_heartbeat_.empty();
```
simpsont-oci, line 3851: Assuming these structures haven't changed since the top of the method, can this just return true? Not necessarily a suggestion, just a question.

dds/DCPS/transport/rtps_udp/RtpsUdpInst.cpp
jrw972: When I was testing 1.1 seemed to small.

dds/DCPS/RTPS/RtpsDiscovery.cpp
jrw972: The backoff factor vs. safety factor doesn't really matter, they can be anything (should be greater than 1).  I will update the code so they can reverse the limits and it will still work properly.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: `readers_expecting_heartbeat_` may change so I think we need to check each condition.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: It absolutely does but it will require changing static discovery so I think we should take that up later.

simpsont-oci (approved):

PR #2408, opened 2021-02-11
Added a script for running all unit tests; moved utility and test files out of bin

mitza-oci:


tools/scripts/unit_tests.pl
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env perl
```
jrw972, line 1: This should be in `DDS_ROOT/bin` or the other test related files should be moved to `DDS_ROOT/tools/scripts/unit_tests.pl`.

jrw972 (changes requested):

tools/scripts/unit_tests.pl
mitza-oci: I agree with the latter.  All this random stuff doesn't need to be added to the user's PATH.

jwillemsen:
Moving the lst files also needs a change in ACE/TAO scripts, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/diff-builds-and-group-fixed-tests-only.sh

mitza-oci:
> 
> 
> Moving the lst files also needs a change in ACE/TAO scripts, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/diff-builds-and-group-fixed-tests-only.sh

Seems like this would be a good opportunity to remove the reverse dependency.  Should the script be in the autobuild repository?

jwillemsen:
> > Moving the lst files also needs a change in ACE/TAO scripts, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/diff-builds-and-group-fixed-tests-only.sh
> 
> Seems like this would be a good opportunity to remove the reverse dependency. Should the script be in the autobuild repository?

Yes, that could be an alternative location.

jrw972 (approved):

mitza-oci:
> 
> 
> > Moving the lst files also needs a change in ACE/TAO scripts, see https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/diff-builds-and-group-fixed-tests-only.sh
> 
> Seems like this would be a good opportunity to remove the reverse dependency. Should the script be in the autobuild repository?

https://github.com/DOCGroup/ACE_TAO/pull/1429

PR #2410, opened 2021-02-11
Unit testing documentation for maintainers

jrw972:


docs/unit_tests.rst
```diff
@@ -0,0 +1,126 @@
+============
```
iguessthislldo, line 1: Could you move the file into `internal` and add it to `index.rst`?

docs/unit_tests.rst
```diff
@@ -0,0 +1,126 @@
...
+ Unit Tests
+============
+
+The Goals of Unit Testing
+=========================
```
iguessthislldo, line 6: Can you match the headers to [the scheme used by the other rst files](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections)?

docs/unit_tests.rst
```diff
@@ -0,0 +1,126 @@
...
+2. Exercise - The code under test is invoked.
+3. Check - The resulting state of the system and outputs are checked.
+4. Teardown - Any resources allocated in the test are deallocated.
+
+Test cases are grouped into a *test suite*.
+
+Test suites are organized into *a test plan*.
+
+We adopt file boundaries for organizing the unit tests for OpenDDS.
+That is, the unit tests for a file group `dds/DCPS/SomeFile.(h|cpp)` will be located in `unit-tests/dds/DCPS/SomeFile.cpp`.
```
iguessthislldo, line 67: Shouldn't it be under `tests/unit-tests`? I think the root directory is already too crowded.

iguessthislldo (changes requested):

kuznetsovmoci (commented):
Would it be worthwhile mentioning association  of unit tests in OpenDDS with google test, or how they are interfaced with Perl test script? I know this would be on my learning curve path.

I think it's a well written introduction on the unit tests in general. Would it have value for other OCI teams? Would it be worthwhile sharing it or presenting at Tech lunch, or elsewhere?

docs/unit_tests.rst
jrw972: Sure.  If that's the case, we should have
* `tests/unit-tests`
* `tests/performance-tests`
*  etc.

iguessthislldo: Yeah, if we wanted to be consistent. That would mean moving bench again though after just moving it.

iguessthislldo:
I think you accidentally removed the file when you did the move.

iguessthislldo (approved):

PR #2419, opened 2021-02-12
Sedp does not send secure deferred type lookup samples

jrw972:
Problem
-------

The secure type lookup writers are not discharged of deferred samples
when their association is complete.

Solution
--------

Invoke `send_deferred_samples` on the secure type lookup writers when
their association is complete.


iguessthislldo (approved):
Ah, yes, I did mess that up.

PR #2420, opened 2021-02-15
Starting with PR #2411 RtpsRelay has a dependency on rapidjson

mitza-oci:


jwillemsen:
Some code in rtpsdelay does check for `OPENDDS_RAPIDJSON`, when this is now a mandatory feature checked through MPC the check for `OPENDDS_RAPIDJSON` in the C++ code could be removed?

PR #2427, opened 2021-02-18
SingleSendBuffer lacks synchronization

jrw972:
Problem
-------

1. `max_sn_` and the send buffer are not synchronized.

   `max_sn_` is increased before the corresponding sample is place in
   the send buffer.  Between these two activities, the task for
   sending heartbeats and nack responses may access `max_sn_` and the
   send buffer and send incorrect gaps and heartbeats.

2. Various methods in SingleSendBuffer lacked synchronization.

   An inspection of the send buffer revealed that many methods are not
   synchronized.  Many of these methods are used as a group and need
   to be transactional.

Solution
--------

1. Add missing synchronization to SingleSendBuffer.

2. Introduce a proxy object for the SingleSendBuffer that holds the
   lock.

3. Extend SingleSendBuffer with a set of sequence numbers that will
   eventually be inserted.  Sequence numbers in this set cannot be
   gap'ed.

4. Refactor `send_and_gather_nack_replies` to use the proxy object.

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendBuffer.h
+++ b/dds/DCPS/transport/framework/TransportSendBuffer.h
@@ -185,7 +185,7 @@ public:
 
   private:
     SingleSendBuffer& ssb_;
-    Proxy(Proxy&);
+    OPENDDS_DELETED_COPY_MOVE_CTOR_ASSIGN(Proxy)
   };
 
 private:
--- a/dds/DCPS/transport/framework/TransportSendBuffer.h
+++ b/dds/DCPS/transport/framework/TransportSendBuffer.h
@@ -87,20 +87,6 @@ public:
 
   bool resend(const SequenceRange& range, DisjointSequence* gaps = 0);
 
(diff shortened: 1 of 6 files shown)


dds/DCPS/transport/framework/TransportSendBuffer.h
```diff
@@ -110,6 +96,98 @@ class OpenDDS_Dcps_Export SingleSendBuffer
...
+
+    void resend_fragments_i(SequenceNumber sequence,
+                            const DisjointSequence& fragments)
+    {
+      ssb_.resend_fragments_i(sequence, fragments);
+    }
+
+  private:
+    SingleSendBuffer& ssb_;
+    Proxy(Proxy&);
```
mitza-oci, line 188: See `OPENDDS_DELETED_COPY_MOVE_CTOR_ASSIGN` in Definitions.h

iguessthislldo (approved):

jonesc-oci (approved):

PR #2430, opened 2021-02-19
Immediate replies to acknacks leads to shootout

jrw972:
Problem
-------

A heartbeat can elicit an immediate acknack from a reader.  This
acknack in turn could elicit another heartbeat.  This leads to a
downward spiral.

Solution
--------

Resurrect the `readers_expecting_heartbeat_` member and logic to defer
the response to a non-final ack.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1874,6 +1874,7 @@ RtpsUdpDataLink::RtpsWriter::remove_reader(const RepoId& id)
       const SequenceNumber acked_sn = reader->acked_sn();
       const SequenceNumber max_sn = expected_max_sn(reader);
       readers_expecting_data_.erase(reader);
+      readers_expecting_heartbeat_.erase(reader);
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2801,7 +2802,7 @@ RtpsUdpDataLink::RtpsWriter::gather_gaps_i(const ReaderInfo_rch& reader,
 void
 RtpsUdpDataLink::RtpsWriter::process_acknack(const RTPS::AckNackSubmessage& acknack,
                                              const RepoId& src_prefix,
-                                             MetaSubmessageVec& meta_submessages)
+                                             MetaSubmessageVec& /*meta_submessages*/)
```
mitza-oci, line 2805: Should the parameter be removed?

jrw972: It's invoked via function pointer so we can't change the signature.

simpsont-oci:
This seems to add an issue in [StaticDiscoveryReconnect](https://github.com/objectcomputing/OpenDDS/pull/2430/checks?check_run_id=1939979003) that wasn't there before.

simpsont-oci (approved):

PR #2432, opened 2021-02-20
Fix Valgrind Issue In RemoveAssociationSweeper

simpsont-oci:
Problem: The scheduled timeout is doing invalid reads / writes on a WriterInfo object that was previously deleted.

Solution: Use the sweeper's mutex to protect access to info_set_ and make better use of RcHandle objects.

### DIFF
--- a/dds/DCPS/RemoveAssociationSweeper.h
+++ b/dds/DCPS/RemoveAssociationSweeper.h
@@ -58,7 +58,7 @@ public:
   }
 
 
-  int remove_info(WriterInfo* const info);
+  RcHandle<WriterInfo> remove_info(WriterInfo* const info);
 private:
   ~RemoveAssociationSweeper();
 
@@ -111,7 +111,10 @@ RemoveAssociationSweeper<T>::RemoveAssociationSweeper(ACE_Reactor* reactor,
 
 template <typename T>
 RemoveAssociationSweeper<T>::~RemoveAssociationSweeper()
-{ }
+{
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  info_set_.clear();
+}
 
 template <typename T>
 void RemoveAssociationSweeper<T>::schedule_timer(RcHandle<WriterInfo>& info, bool callback)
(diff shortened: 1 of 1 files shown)


dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -111,7 +111,10 @@ RemoveAssociationSweeper<T>::RemoveAssociationSweeper(ACE_Reactor* reactor,
 
 template <typename T>
 RemoveAssociationSweeper<T>::~RemoveAssociationSweeper()
-{ }
+{
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  info_set_.clear();
```
mitza-oci, line 116: Since the `RemoveAssociationSweeper` object is being destroyed, this doesn't seem like it's necessary.
Being a reference-counted object, no reference to this object exists when the destructor starts running.
If this isn't true in practice then there's a bug somewhere else.

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -58,7 +58,7 @@ class RemoveAssociationSweeper : public ReactorInterceptor {
   }
 
 
-  int remove_info(WriterInfo* const info);
+  RcHandle<WriterInfo> remove_info(WriterInfo* const info);
```
mitza-oci, line 61:
```suggestion
  RcHandle<WriterInfo> remove_info(WriterInfo* info);
```

mitza-oci (changes requested):

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -203,33 +218,56 @@ void RemoveAssociationSweeper<T>::ScheduleCommand::execute()
...
+
+  RcHandle<WriterInfo>& info = this->info_;
+
+  if (info->remove_association_timer_ != WriterInfo::NO_TIMER) {
+    const void *arg = 0;
+    sweeper->reactor()->cancel_timer(info->remove_association_timer_, &arg);
+    const TimerArg* ta = reinterpret_cast<const TimerArg*>(arg);
+    if (ta) {
+      delete ta;
+    }
```
mitza-oci, line 263:
```suggestion
    delete ta;
```

dds/DCPS/RemoveAssociationSweeper.h
```diff
@@ -203,33 +218,56 @@ void RemoveAssociationSweeper<T>::ScheduleCommand::execute()
...
-  const void* arg = reinterpret_cast<const void*>(this->info_.in());
-  this->sweeper_->info_set_.push_back(this->info_);
+  RcHandle<RemoveAssociationSweeper<T> > sweeper = this->sweeper_.lock();
+  if (!sweeper) {
+    return;
+  }
+
+  RcHandle<WriterInfo>& info = this->info_;
+
+  const void* arg = new TimerArg(sweeper, info);
```
mitza-oci, line 228: Could this move down to right before it's used?  Or even eliminate the need for a distinct local variable?

mitza-oci (changes requested):

mitza-oci (approved):

PR #2438, opened 2021-02-23
Improvements to RTPS reliability to reduce resends

jrw972:


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -656,19 +651,7 @@ RtpsUdpDataLink::associated(const RepoId& local_id, const RepoId& remote_id,
       if (rr == readers_.end()) {
         pending_reliable_readers_.erase(local_id);
         RtpsUdpDataLink_rch link(this, OpenDDS::DCPS::inc_count());
-        CORBA::Long an_start = 0;
-        CORBA::Long nf_start = 0;
-        CountMapType::iterator anc_it = acknack_counts_.find(local_id.entityId);
```
simpsont-oci, line 661: I thought the purpose of this was to prevent long-running writers from blocking / ignoring short-lived readers that reconnect with the same guid?

jrw972: I'm not sure where the initial nackfrag counter came from but the acknack counter came from StaticDiscovery.  Essentially, a reader can be advertised before it exists so a counter was needed to keep track of the acknack count.  From the code, the nackfrag counter was not used.

The code you are referring to is in `process_acknack` where we check that the count increased but the acknack decreased meaning the reader has gone through rediscovery.  Given the test results, I'll probably have to adjust this code.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1661,23 +1624,49 @@ RtpsUdpDataLink::RtpsReader::process_gap_i(const RTPS::GapSubmessage& gap,
...
 
   DeliverHeldData dhd(rchandle_from(this), src);
+}
 
-  return false;
+void
+RtpsUdpDataLink::send_interesting_ack_nack(const RepoId& writerid,
+                                           const RepoId& readerid,
+                                           CORBA::Long count,
+                                           MetaSubmessageVec meta_submessages)
```
mitza-oci, line 1633:
```suggestion
                                           MetaSubmessageVec& meta_submessages)
```

mitza-oci (changes requested):

mitza-oci (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -640,7 +635,7 @@ RtpsUdpDataLink::associated(const RepoId& local_id, const RepoId& remote_id,
       RtpsWriter_rch writer = rw->second;
       g.release();
       writer->update_max_sn(remote_id, max_sn);
-      writer->add_reader(make_rch<ReaderInfo>(remote_id, remote_durable));
+      writer->add_reader(make_rch<ReaderInfo>(remote_id, remote_durable, remote_context));
```
simpsont-oci, line 638: Can we just rename this variable to participant_flags since that makes it clear what we're dealing with?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1694,60 +1683,57 @@ RtpsUdpDataLink::received(const RTPS::HeartBeatSubmessage& heartbeat,
...
   if (wi == remote_writers_.end()) {
-    return false;
+    return;
   }
 
   const WriterInfo_rch& writer = wi->second;
 
   if (!compare_and_update_counts(heartbeat.count.value, writer->heartbeat_recvd_count_)) {
-    return false;
+    VDBG((LM_WARNING, "(%P|%t) RtpsUdpDataLink::process_heartbeat_i "
```
simpsont-oci, line 1734: Does VDBG buy us anything here? We weren't historically using it in RtpsUdpDatalink.

We really need to get a better logging setup for RTPS/UDP... ability to filter by message type & builtin vs user entity at the very least...

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -589,7 +592,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
 
   struct CountKeeper {
     OPENDDS_MAP(size_t, IdCountSet) heartbeat_counts_;
-    IdCountSet acknack_counts_;
+    OPENDDS_MAP_CMP(EntityId_t, IdCountSet, EntityId_tKeyLessThan) acknack_counts_;
```
simpsont-oci, line 595: Make a typedef above for readability / in case we need to use iterators in the future?

simpsont-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I did it to be consistent with similar code.

simpsont-oci (approved):

PR #2444, opened 2021-02-25
SporadicTask schedule is on the reliable write path

jrw972:
Problem
-------

Scheduling a SporadicTask is now in the path of writing a reliable
sample.  As the writing thread is not the same as the SporadicTask
thread, the schedule commands will be queued.

Solution
--------

Optimize by not enqueuing a schedule/cancel command when the
previously enqueued command is the same.

### DIFF
--- a/dds/DCPS/SporadicTask.h
+++ b/dds/DCPS/SporadicTask.h
@@ -21,7 +21,7 @@ public:
   explicit SporadicTask(RcHandle<ReactorInterceptor> interceptor)
     : interceptor_(interceptor)
     , scheduled_(false)
-    , last_command_(NONE)
+    , last_command_(SPORADIC_TASK_NONE)
   {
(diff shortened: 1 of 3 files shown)


dds/DCPS/SporadicTask.h
```diff
@@ -69,17 +72,37 @@ class SporadicTask : public RcEventHandler {
 private:
   WeakRcHandle<ReactorInterceptor> interceptor_;
   bool scheduled_;
+  enum {
+    NONE,
+    SCHEDULE,
+    CANCEL
```
mitza-oci, line 78: Use enumerator names that are less likely to have naming collisions with other identifiers including macros.

mitza-oci (changes requested):

mitza-oci (approved):

PR #2446, opened 2021-02-25
Fix Static Discovery Test Issues and Improve Qos Mismatch Error Logging

simpsont-oci:


dds/DCPS/StaticDiscovery.cpp
```diff
@@ -242,8 +242,10 @@ void StaticEndpointManager::assign_subscription_key(RepoId& rid,
   // Qos in registry will not have the user data so overwrite.
   qos2.user_data = pos->second.qos.user_data;
 
-  if (qos2 != pos->second.qos) {
-    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) WARNING: StaticEndpointManager::assign_subscription_key: dynamic and static QoS differ\n")));
+  DDS::DataReaderQos qos3(pos->second.qos);
```
mitza-oci, line 245: Is this change needed to make it work, or was it just to ease debugging?
I don't mind having it here, just wondering what motivated that change.

simpsont-oci: It was for debugging, but I left the change to match the DataWriterQos case above. For debugging, I had inserted an OPENDDS_ASSERT(qos2 == qos3) above the conditional and gdb didn't want to allow me to print out the value of "pos->second.qos" without an active process but was happy to allow me to print qos3.

mitza-oci (approved):

mitza-oci:
failure in `tests/DCPS/MultiDiscovery` looks like it's related to this change

simpsont-oci:
Hmm, there seems to be message order issues on windows builds still. I will take a look tomorrow.

PR #2451, opened 2021-02-26
Github Actions Update + Migration from Azure/Jenkins

ClaytonCalabrese:
As part of our modernization of our CI/Scoreboard we wanted to migrate all of our CI to github actions.  This PR represents the first step of that.  It has migrated all of our CI builds from Azure and Jenkins to GHA.  Some extraneous builds were removed and others were changed.  

There are documents in confluence and the google drive describing the plan and the problems remaining.

The Safety and Static builds are expected to fail for this PR.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -29,7 +29,7 @@ env:
   TRIGGERING_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
 
 jobs:
-  ACE_TAO_ubuntu-18_04_i0_java12_nojson:
+  ACE_TAO_u18_i0_xer0_js0_j12:
 
     runs-on: ubuntu-18.04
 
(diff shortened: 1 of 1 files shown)


jwillemsen:
Also the badge in the README.md needs to be updated

simpsont-oci (changes requested):
Overall, looks good. I think we should revisit the naming scheme to shorten further them to fit in the actions view window. Also, the ubuntu-18_04_clang10_nojson build probably needs to reflect that it has security in its name.

Only other thought is that it seems to be taking 6+ hours to finish all the builds and testing, which is pretty huge. We may need to remove the tests from some builds or remove a feature or two from some builds to keep the build & test times down (java, I'm looking at you).

simpsont-oci (approved):
Looks good, run time is much better. The "Test Result" pages are missing but the artifacts look fine. I'm guessing it's an issue with the check_test_results workflow and we can try to fix it in a followup PR.

PR #2452, opened 2021-02-27
Ability to Run All Configured Tests and auto_run_test.pl Fixes

iguessthislldo:
By default `auto_run_tests.pl` will now run all configured tests for almost all, if not all, situations by getting information from the configure script.

Also added improvements and fixes from ACE's `auto_run_test.pl` and other new improvements. These include:

 - Now able to interrupt the script with ctl-c when it's running tests. Before it would just skip the currently running test and go on to the next one.
 - A `--help` message that's accurate.
 - The normal list files are now builtin as options: `--no-dcps`, `--security`, `--java`, `--modeling`
 - New options for listing information about the test lists.


iguessthislldo:
Closing this because most of the auto_run_tests improvements were added as part of #2664 and CMake build support in #4203 is going to allow us to automatically run configured and built test inherently when we've CMaked/CTested all the tests.

PR #2453, opened 2021-02-27
Improve Root .gitignore and build_and_test.yml

iguessthislldo:
- Root `.gitignore`:
  - Sorted the entries into categories
  - Added:
    - `*.zip` and `*.tar.gz` files for incomplete ACE/TAO downloads
    - Generic rule for CMake `build` directories and removed existing individual rules
    - Core dumps
    - Valgrind output files

- `build_and_test.yml`:
  - ~Make common ignored file list using YAML anchors. We should probably refactor the whole file to take advantage of these because they look like they could significantly reduce the size of the file.~
  - Added to ignored file list:
    - Any `*.md`, `*.rst`, `*.txt`, `.gitignore`, `.lint_config`, and `README` files.
    - All the remaining things in the root directory that look like they can be ignored.


iguessthislldo:
I didn't write the anchors right, but while looking at the error message I saw this: 

> .github/workflows/build_and_test.yml: Anchors are not currently supported. Remove the anchor 'opendds_common_on'

WHY?!? :fearful:

Here's some links to places where this issue is being tracked, but it doesn't look like Github is going to implement this even though it's standard YAML...

- https://github.community/t/support-for-yaml-anchors/16128
- https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/14
- https://github.com/actions/runner/issues/438

The choices seem to be:

- Revert to copy and paste, which I will do for now.
- Do some sort of prepossessing in the workflow using a compliment YAML tool like what some people are suggesting. I think we where also talking about something in a similar vain, but with having a program generate the file.

mitza-oci (approved):

PR #2454, opened 2021-02-27
Test Making RTPS the Default for the Future

iguessthislldo:
Currently, when not using safety profile, the default discovery is InfoRepo and the default transport is tcp. This attempts to change the default discovery to rtps and the default transport to rtps_udp.

### DIFF
--- a/DDS.mwc
+++ b/DDS.mwc
@@ -19,5 +19,6 @@ workspace {
     tools/modeling/tests
 
     dds/InfoRepo
+    tools/dcpsinfo_dump
   }
 }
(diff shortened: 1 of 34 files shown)


mitza-oci:
Additional context for this one... It's a draft PR because we don't intent to merge it in the near term.  We're testing the changes here to learn what would be required to make the changes for a future version.

iguessthislldo:
That's a good point that I left out. I put something to that affect in the commit message, but left that out of the PR description for some reason. Changed the PR name to more accurately reflect this.

jwillemsen:
This would be a good step forward!

iguessthislldo:
Related: #4779. Also my write up for further work was uploaded to the wiki as https://github.com/OpenDDS/OpenDDS/wiki/Proposal:-Make-RTPS-the-default

jrw972:
Closed by https://github.com/OpenDDS/OpenDDS/pull/4963

PR #2460, opened 2021-03-03
Single timer used for heartbeats

jrw972:
Problem
-------

A single timer is used for sending heartbeats.  This means there is a
single heartbeat period which can penalize writers without slow
readers.  Furthermore, the support static discovery is intertwined.

Solution
--------

Use a timer per-writer.


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1324,6 +1322,53 @@ bool RtpsUdpDataLink::requires_inline_qos(const GUIDSeq_var& peers)
...
+
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g2, heartbeat_mutex_);
+    if (expected_acks_) {
+      // One more readers did not ack since the last heartbeat.
+      heartbeat_period_ *= cfg.heartbeat_backoff_factor_;
+    } else if (last_ack_ != MonotonicTimePoint::zero_value && last_heartbeat_ != MonotonicTimePoint::zero_value) {
+      heartbeat_period_ = (last_ack_ - last_heartbeat_) * cfg.heartbeat_safety_factor_;
+    }
+    heartbeat_period_ = std::min(heartbeat_period_, std::max(cfg.heartbeat_period_maximum_, cfg.heartbeat_period_minimum_));
```
mitza-oci, line 1345: It seems like the relative order of `cfg.heartbeat_period_maximum_, cfg.heartbeat_period_minimum_` should already be known

jrw972: There is nothing that prevents the user from misconfiguring it and I didn't see a fool-proof method for correcting it when they do.

simpsont-oci:
Looks like there are still a number of WaitForAck test failures in GHA (none on Scoreboard) as well as some FooTest* and TransientLocal*Test failures on windows (none on Scoreboard)

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -388,6 +388,18 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
     mutable ACE_Thread_Mutex mutex_;
     mutable ACE_Thread_Mutex elems_not_acked_mutex_;
 
+    typedef PmfSporadicTask<RtpsWriter> Sporadic;
+
+    mutable ACE_Thread_Mutex heartbeat_mutex_;
```
simpsont-oci, line 393: Are you sure need heartbeat_mutex_ and can't just use mutex_? The only reason elems_not_acked_mutex_ exists is because of some weird reentrant behavior with the send buffer, I'm pretty sure. I wouldn't have expected the heartbeat stuff to need something similar if it didn't need it before.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3921,6 +3932,8 @@ RtpsUdpDataLink::RtpsWriter::~RtpsWriter()
       ACE_TEXT("deleting with %d elements left not fully acknowledged\n"),
       elems_not_acked_.size()));
   }
+
+  heartbeat_.cancel_and_wait();
```
simpsont-oci, line 3936: I'm not sure this method should be holding on the the mutexes if it's going to potentially be waiting here. Based on @mitza-oci 's comments in the past I'm not sure locking either one in the destructor is a great idea anyway, since it means somewhere out there someone is accessing a reference counted object without a reference count... so maybe try removing them and fix the incorrect access if it causes problems?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Can you fix the order in the transport constructor so you only have to do it once per run and not for every heartbeat?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1324,6 +1330,53 @@ bool RtpsUdpDataLink::requires_inline_qos(const GUIDSeq_var& peers)
...
+
+  for (size_t i = 0; i < pendingCallbacks.size(); ++i) {
+    pendingCallbacks[i]->data_dropped();
+  }
+
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g2, heartbeat_mutex_);
+    if (keep_going) {
+      heartbeat_.cancel();
+      heartbeat_.schedule(heartbeat_period_);
```
simpsont-oci, line 1371: Hmm, we're canceling and rescheduling every heartbeat for every writer? What's the cancel / reschedule delay for SporadicTask? It's definitely non-zero for MultiTask, which is why I added the cancel_estimate_ field. Maybe you could use MultiTask with the constructed default of the max heartbeat period and just enable with the new heartbeat period every time? I'm not sure if it will be much more efficient, but it seems like it will be more robust to cancelation delays and eliminate the need for a cancel call here.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2748,7 +2801,12 @@ RtpsUdpDataLink::RtpsWriter::process_acknack(const RTPS::AckNackSubmessage& ackn
 
     if (reader->expecting_ack_) {
       reader->expecting_ack_ = false;
-      link->received_expected_ack();
+      --expected_acks_;
+      last_ack_ = MonotonicTimePoint::now();
+      if (expected_acks_ == 0) {
+        heartbeat_.cancel();
+        heartbeat_.schedule(TimeDuration::zero_value);
```
simpsont-oci, line 2808: Ditto cancel estimate question here, maybe even more so since the deadline is zero.

simpsont-oci (changes requested):
The latest GHA builds are showing multiple builds with issues in the WaitForAck test, and one of the Ubuntu 16.04 builds had a double free / invalid pointer error in the TransientLocalMultiInstanceTest. Neither those normally cause issues on the scoreboard, Jenkins PRB, or GHA builds (pretty sure). Overall the code changes seem reasonable, not sure why the tests are failing.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I think the code it question no longer exists.  Don't know why github is not showing outdated.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: The goal here was to get to processing the nack replies as soon as possible, i.e., when a nack from every reader has been received.  Based on the wait for ack test on my machine, the overhead is .000014 seconds.  This will only occur once per heartbeat period (say 100ms).

simpsont-oci (approved):

simpsont-oci:
I think the remaining test failures in GHA are probably issues with the new builds rather than the PR.

PR #2467, opened 2021-03-05
Misc. Logging Changes

iguessthislldo:
- Log when Readers and Writers get a GUID. Also logs the topic type and topic name.
- Fix missing dash before metadata in OpenDDS version logging.
- Make all thread status logging I could find start at 4 to match the update logging point.
- Slap `DCPS_debug_level` on some logging points that were missing them.

### DIFF
--- a/dds/DCPS/DataWriterImpl.cpp
+++ b/dds/DCPS/DataWriterImpl.cpp
@@ -1498,7 +1498,7 @@ DataWriterImpl::enable()
     ACE_DEBUG((LM_DEBUG, "(%P|%t) DataWriterImpl::enable: "
       "got GUID %C, publishing to topic name \"%C\" type \"%C\"\n",
       LogGuid(publication_id_).c_str(),
-      topic_servant_->get_name(), topic_servant_->get_type_name()));
+      topic_servant_->topic_name(), topic_servant_->type_name()));
   }
 
   this->data_container_->publication_id_ = this->publication_id_;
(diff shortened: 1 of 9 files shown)


mitza-oci (approved):

mitza-oci:
Are the errors detected in `tests/security/attributes` (GHPRB) new from this PR?

iguessthislldo:
I think they're expected to fail, but I changed it from a warning to an error to match the other error returns that log in that function.

iguessthislldo:
I will look into more and fix it.

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1482,16 +1485,22 @@ DataWriterImpl::enable()
...
+    }
     data_container_->shutdown_ = true;
     return DDS::RETCODE_ERROR;
   }
 
+  if (DCPS_debug_level >= 2) {
+    ACE_DEBUG((LM_DEBUG, "(%P|%t) DataWriterImpl::enable: "
+      "got GUID %C, publishing to topic name \"%C\" type \"%C\"\n",
+      LogGuid(publication_id_).c_str(),
+      topic_servant_->get_name(), topic_servant_->get_type_name()));
```
mitza-oci, line 1501: Use `_var` or the methods that don't start with `get_*` from `TopicImpl.h`

mitza-oci (changes requested):

mitza-oci (approved):

PR #2469, opened 2021-03-05
GitHub Actions: CMake CI testing

ClaytonCalabrese:
The initial PR for GHA was unable to add CMake integration in a reasonable amount of time.  This is the PR to add that functionality. The cmake for static and safety builds probably require additional changes but they should be relatively minor and will be included with the PR for those two pieces.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -175,6 +175,24 @@ jobs:
         cd OpenDDS
         . setenv.sh
         make -j4
+    - name: build CMake
+      shell: bash
+      run: |
+        cd OpenDDS
+        . setenv.sh
+        cd tests/cmake_integration/Messenger/Messenger_1
+        rm -fr build
+        mkdir build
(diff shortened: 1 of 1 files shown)


ClaytonCalabrese:
Due to some of these builds not running the "Test" job after doing the build, I have put the running of the cmake_integration test during the "Build" job.  

I am undecided on whether this is better than just commenting out the test part of the "Test" job and placing it there for the builds that don't do auto_run_tests.pl.

Also there is a TODO that 8 days from now change the _2 at the end of the name of ACE_TAO_u18_o1d0v1_xer0_j12_2.  It is only there for force a cache clear, and GHA doesn't let us manually clear the cache.  After 7 days the old ACE_TAO artifact for ACE_TAO_u18_o1d0v1_xer0_j12_2 should be cleared, and then I will switch the name back to build_u18_o1d0v1_xer0_j12.

mitza-oci (approved):

PR #2475, opened 2021-03-08
Fix Memory Leak in dds_generator

simpsont-oci:


dds/idl/dds_generator.cpp
```diff
@@ -348,7 +348,9 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
     AST_Enum* enu = dynamic_cast<AST_Enum*>(actual_type);
     UTL_ScopeActiveIterator i(enu, UTL_Scope::IK_decls);
     AST_EnumVal *item = dynamic_cast<AST_EnumVal*>(i.item());
-    def_val = item->name()->get_string_copy();
+    char* temp = item->name()->get_string_copy();
+    def_val = temp;
+    delete [] temp;
```
mitza-oci, line 353:
```suggestion
    def_val = item->local_name()->get_string();
```

mitza-oci (changes requested):

dds/idl/dds_generator.cpp
```diff
@@ -348,9 +348,7 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
     AST_Enum* enu = dynamic_cast<AST_Enum*>(actual_type);
     UTL_ScopeActiveIterator i(enu, UTL_Scope::IK_decls);
     AST_EnumVal *item = dynamic_cast<AST_EnumVal*>(i.item());
-    char* temp = item->name()->get_string_copy();
-    def_val = temp;
-    delete [] temp;
+    def_val = item->local_name()->get_string();
```
simpsont-oci, line 351: @mitza-oci It looks like this is leaving off the "::DDS::" before some constants in dds/DdsDcpsCoreTypeSupportImpl.cpp. I don't have a great feel for the difference between name and local_name. Should I revert this to the safe version of name->get_string_copy() or does this need to look like the "if (use_cxx11)" block just below?

iguessthislldo: `local_name` is the last component of `name`, so changing them would be significant. I think what Adam meant was `def_val = item->name()->get_string();`.

mitza-oci: Ah, I thought it could be just the simple name of the enumerator.

dds/idl/dds_generator.cpp
```diff
@@ -348,7 +348,7 @@ string type_to_default(const std::string& indent, AST_Type* type, const string&
     AST_Enum* enu = dynamic_cast<AST_Enum*>(actual_type);
     UTL_ScopeActiveIterator i(enu, UTL_Scope::IK_decls);
     AST_EnumVal *item = dynamic_cast<AST_EnumVal*>(i.item());
-    def_val = item->name()->get_string_copy();
+    def_val = item->name()->get_string();
```
iguessthislldo, line 351: Okay, looks like I lead you astray. That's what I get for commenting without actually looking at the code. Now that I'm actually paying attention, you can use:

```suggestion
    def_val = scoped(item->name());
```

mitza-oci: Yes... also while we're here let's update the lines right after this so that it's not immediately overwritten in C++11-mode

PR #2476, opened 2021-03-08
Static Build Fix for GitHub Actions

ClaytonCalabrese:
The initial PR for GHA did not have working static builds. This PR fixes them by removing the test build option so that the builds do not run out of disk space.  Additionally dev guide messenger test is built/run so that we have at least a little test coverage in the CI for static builds.

Similar to other GHA PRs, there will need to be a name reversion 8 days after the PR is merged due to cacheing of build artifacts.

To reviewers: The only non uncommenting changes are the ./configure options, and adding DevGuideExamples sections to each.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4684,23 +4684,22 @@ jobs:
         cd DevGuideExamples\DCPS\Messenger
         perl run_test.pl -ExeSubDir Static_Debug
     - name: build CMake
-      shell: bash
+      shell: cmd
       run: |
         cd OpenDDS
-        . setenv.sh
+        call setenv.cmd
         cd tests/cmake_integration/Messenger/Messenger_1
-        rm -fr build
         mkdir build
         cd build
         cmake -DCMAKE_PREFIX_PATH=$OPENDDS_INSTALL_PREFIX ..
         cmake --build .
     - name: run CMake test
-      shell: bash
+      shell: cmd
       run: |
         cd OpenDDS
-        . setenv.sh
+        call setenv.cmd
         cd tests/cmake_integration/Messenger/Messenger_1/build
-        perl run_test.pl rtps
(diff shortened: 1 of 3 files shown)


ClaytonCalabrese:
Hold up on reviewing this for a bit. The CMake stuff isn't playing nice right now.

.github/workflows/build_and_test.yml
```diff
@@ -4684,23 +4684,22 @@ jobs:
...
     - name: run CMake test
-      shell: bash
+      shell: cmd
       run: |
         cd OpenDDS
-        . setenv.sh
+        call setenv.cmd
         cd tests/cmake_integration/Messenger/Messenger_1/build
-        perl run_test.pl rtps
+        perl ../../../../DCPS/Messenger/run_test.pl -ExeSubDir Debug rtps
```
iguessthislldo, line 4702: The `run_test.pl` file is copied to the CMake output directory from what I remember, at least for the `Messenger_1/2` tests. Also what about using `run_ci_tests.pl`? It's building running all the CMake tests, which I don't think is the case on the scoreboard.

ClaytonCalabrese: https://github.com/ClaytonCalabrese/OpenDDS/runs/2070048284?check_suite_focus=true

This is when I was debugging yesterday and did dir of the Messenger_1 and build directories.

ClaytonCalabrese: As for the run_ci_tests.pl I'm trying it on a local branch right now.

ClaytonCalabrese: It looks like run_ci_tests.pl doesn't play nice with static builds.  I ran it in an old static local build and it is experiencing problems.  

I ran it with a local normal build and it runs fine.

I'm making an up to date static build but I have my doubts that it will fix anything.

iguessthislldo (approved):

PR #2478, opened 2021-03-09
Avoid Matching on RTPS Relay

iguessthislldo:
Also remove `GuidPrefixEqual` since nothing was using it as a struct and replaced it with a normal function.

### DIFF
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -1009,6 +1009,10 @@ namespace OpenDDS {
           discovered_endpoints = td.discovered_subscriptions();
         }
 
+        if (local_endpoints.empty()) {
+          return;
+        }
+
         for (RepoIdSet::const_iterator iter = local_endpoints.begin();
              iter != local_endpoints.end(); ++iter) {
           // check to make sure it's a Reader/Writer or Writer/Reader match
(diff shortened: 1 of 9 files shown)


dds/DCPS/DiscoveryBase.h
```diff
@@ -1216,10 +1220,6 @@ namespace OpenDDS {
           return; // Possible and ok, since lock is released
         }
 
-        if (!writer_local && !reader_local) {
-          return;
-        }
-
```
mitza-oci, line 1222: Should this be removed? When there is at least one local endpoint on this topic, should the function continue past this point when both sides are non-local?

mitza-oci (changes requested):

dds/DCPS/DiscoveryBase.h
iguessthislldo: Then what about restoring my previous change in `match_endpoints` where we return before checking non-locals?

mitza-oci: Any change that helps scalability and doesn't hurt correctness is fair game

mitza-oci (approved):

jrw972 (approved):

dds/DCPS/DiscoveryBase.h
```diff
@@ -1009,6 +1009,10 @@ namespace OpenDDS {
           discovered_endpoints = td.discovered_subscriptions();
         }
 
+        if (local_endpoints.empty()) {
+          return;
+        }
```
iguessthislldo, line 1014: It seems that when this function is called `local_endpoints` can be empty and `repoId` can be a local endpoint that has to make it to the `discovered_endpoints` for loop or else it causes problems.

mitza-oci (approved):

PR #2479, opened 2021-03-10
Fix Spdp Shutdown / Dispose Memory Access Issues

simpsont-oci:
There are a couple race conditions causing heap-use-after-free ASAN issues in Spdp/Sedp.
- First, a race condition exists between SpdpTransport::handle_input() and Spdp's destructor. Simply moving to a weak object handle during the call to data_recieved potentially causes the reactor thread in charge of cleaning up its own reactor, so the actual shutdown work for Spdp needed to be moved to a separate method to allow trivial behaviors in ~Spdp().
- Second, the reverse locking was still causing issues with the iterators, but this time it was because the "updated" iterators weren't getting passed back up into calling methods when applicable, and also because we weren't checking to see if the pdata object in disassociate was being invalidated after our reverse lock to call disassociate_helper functions.

This PR should hopefully fix those issues.

### DIFF
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -2022,7 +2022,7 @@ namespace OpenDDS {
         ACE_GUARD(ACE_Thread_Mutex, g, lock_);
         endpoint_manager().ignore(ignoreId);
 
-        const DiscoveredParticipantIter iter = participants_.find(ignoreId);
+        DiscoveredParticipantIter iter = participants_.find(ignoreId);
         if (iter != participants_.end()) {
           remove_discovered_participant(iter);
         }
@@ -2327,8 +2327,11 @@ namespace OpenDDS {
 
       virtual EndpointManagerType& endpoint_manager() = 0;
 
-      void remove_discovered_participant(DiscoveredParticipantIter iter)
+      void remove_discovered_participant(DiscoveredParticipantIter& iter)
       {
+        if (iter == participants_.end()) {
+          return;
+        }
         RepoId part_id = iter->first;
         bool removed = endpoint_manager().disassociate(iter->second.pdata_);
         iter = participants_.find(part_id); // refresh iter after disassociate, which can unlock
@@ -2375,7 +2378,7 @@ namespace OpenDDS {
         }
       }
 
-      virtual void remove_discovered_participant_i(DiscoveredParticipantIter) {}
+      virtual void remove_discovered_participant_i(DiscoveredParticipantIter&) {}
 
 #ifndef DDS_HAS_MINIMUM_BIT
     DCPS::ParticipantBuiltinTopicDataDataReaderImpl* part_bit()
--- a/dds/DCPS/DiscoveryBase.h
+++ b/dds/DCPS/DiscoveryBase.h
@@ -2551,6 +2551,7 @@ namespace OpenDDS {
           participants_.erase(domain);
         }
 
+        participant->shutdown();
         return true;
       }
 
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -1731,6 +1731,11 @@ Sedp::disassociate(ParticipantData_t& pdata)
     result = true;
   }
 
+  BuiltinEndpointSet_t associated_endpoints = pdata.associated_endpoints;
+#ifdef OPENDDS_SECURITY
+  DDS::Security::ExtendedBuiltinEndpointSet_t extended_associated_endpoints = pdata.extended_associated_endpoints;
+#endif
+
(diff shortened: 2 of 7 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2822,7 +2831,12 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
           plist[0]._d(PID_PARTICIPANT_GUID);
         }
 
-        outer_->data_received(data, plist, remote);
+        DCPS::WeakRcHandle<Spdp> wh(*outer_);
+        DCPS::RcHandle<Spdp> outer = wh.lock();
+
+        if (outer) {
```
mitza-oci, line 2837: For WeakRcHandle to work in this way, the type of `outer_` needs to be changed to WeakRcHandle.  That would allow the WeakRcHandle to be initialized at a time when the RcObject is guaranteed to be valid.

mitza-oci (dismissed):

dds/DCPS/RTPS/Spdp.cpp
simpsont-oci: In general, I agree with what you're saying, but I think we can assume outer_ is still valid at this point. With the current design, as long as the reactor thread is still running (which it has to be, here, in handle_input() before data_received has a change to reverse lock anything), the object will still be valid.

I initially started by changing outer_ to a WeakRcHandle but it touches a lot of code with marginal advantage for this usage so I changed back and tried it this way. I haven't seen any complaints (for this section of code) from the address sanitizer so far, even running both reliable and durable versions of Thrasher in a loop.

I see the benefit for maintainability to move to WeakRcHandle, though, so if you still want me to make those changes let me know.

mitza-oci: It's fine to keep it that way, but add the info about the assumptions in a comment.

dds/DCPS/DiscoveryBase.h
```diff
@@ -2375,7 +2378,7 @@ namespace OpenDDS {
         }
```
jrw972, line 2378: Should iter == participants_.end() when this function returns?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1739,57 +1739,62 @@ Sedp::disassociate(ParticipantData_t& pdata)
     remove_from_bit_i(*it);
   }
 
+  BuiltinEndpointSet_t associated_endpoints = pdata.associated_endpoints;
```
jrw972, line 1742:
```suggestion
  BuiltinEndpointSet_t& associated_endpoints = pdata.associated_endpoints;
```

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -1739,57 +1739,62 @@ Sedp::disassociate(ParticipantData_t& pdata)
...
+#ifdef OPENDDS_SECURITY
+  DDS::Security::ExtendedBuiltinEndpointSet_t extended_associated_endpoints = pdata.extended_associated_endpoints;
```
jrw972, line 1744:
```suggestion
  DDS::Security::ExtendedBuiltinEndpointSet_t& extended_associated_endpoints = pdata.extended_associated_endpoints;
```

jrw972 (changes requested):

dds/DCPS/RTPS/Sedp.cpp
simpsont-oci: These can't be references, because pdata can be invalidated while the reverse lock is in effect. See the end, where I write this value back into pdata (if it exists).

simpsont-oci:
I have a few more changes coming to fix the dpshutdown issues, which weren't showing up on Linux but were reproducible on Windows.

dds/DCPS/DiscoveryBase.h
simpsont-oci: I think so, yeah, but I'm not sure if it's possible to reregister while disassociate's reverse lock is in effect.

dds/DCPS/RTPS/Spdp.cpp
simpsont-oci: The same operation needs / needed to happen in SpdpTransport, but because of the ChangeMulticastGroup jobs, which held references to SpdpTransport, the same guarantees couldn't be made. As a result, I opted for the full WeakRcHandle conversion (which touched a lot of lines but wasn't horrible) and removed these comments. The DpShutdown issues seem to be resolved on my windows build and Thrasher seems to run fine without ASAN issues. I'll kick off a new set of tests but this is probably ready for re-review (I'll mark it as such if I don't see anything major).

jrw972 (approved):

PR #2482, opened 2021-03-11
Fix address sanitizer global-buffer-overflow issue in DiffieHellman.cpp

simpsont-oci:
AddressSanitizer is pointing out a global-buffer-overflow issue in the DiffieHellman code where it's technically comparing the length of an octet sequence (of potentially arbitrary length) against fixed-length string constants and using the sequence length to do the comparison. In practice, the lengths will probably be from a small subset of known algorithm names, and since they seem to be null terminated the "overflow" will probably only ever be a single character, but we probably ought to clean it up.

Optionally, we could also investigate why the octet sequences are null terminated and clean that up to avoid the "trimming" math that needs to happen in compare, but the current changes at least resolve the address sanitizer complaints.

### DIFF
--- a/dds/DCPS/security/SSL/DiffieHellman.cpp
+++ b/dds/DCPS/security/SSL/DiffieHellman.cpp
@@ -18,6 +18,15 @@ namespace OpenDDS {
 namespace Security {
 namespace SSL {
 
+namespace {
+  int compare(const DDS::OctetSeq& s1, const char* s2) {
+    size_t s1_len = s1.length();
+    while (s1_len > 0 && s1.get_buffer()[s1_len - 1] == 0) { s1_len--; }
+    const size_t s2_len = std::strlen(s2);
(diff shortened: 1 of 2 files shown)


dds/DCPS/security/SSL/DiffieHellman.cpp
```diff
@@ -18,6 +18,15 @@ namespace OpenDDS {
 namespace Security {
 namespace SSL {
 
+namespace {
+  int compare(const DDS::OctetSeq& s1, const char* s2) {
```
mitza-oci, line 22: This can be simplified:
- we don't need to know relative order, just equality/inequality
- if `s1.length()` is not equal to `strlen(s2) + 1`, the strings must not be equal, bail out
- then defer to memcmp

mitza-oci (changes requested):

mitza-oci (approved):

PR #2493, opened 2021-03-16
Github Actions: Safety Profile

ClaytonCalabrese:
This PR adds the safety builds to github actions


.github/workflows/build_and_test.yml
```diff
@@ -541,72 +541,74 @@ jobs:
...
+        <command name="print_make_version"/>
+        <command name="check_compiler" options="gcc"/>
+        <command name="print_perl_version"/>
+        <command name="print_autobuild_config"/>
+        <command name="auto_run_tests" options="script_path=tests dir=$GITHUB_WORKSPACE/OpenDDS -Config RTPS -Config LINUX -Config XERCES3 -Config CXX11 -Config OPENDDS_SAFETY_PROFILE -Config GH_ACTIONS"/>
+        <command name="log" options="off"/>
+        <command name="process_logs" options="prettify"/>
+        </autobuild>
+        EOF
+        $GITHUB_WORKSPACE/autobuild/autobuild.pl "$GITHUB_WORKSPACE/OpenDDS/build/target/${{ github.job }}_autobuild_workspace/config.xml"
```
simpsont-oci, line 610: These all seem to be missing the "upload artifact" & "check results" steps of the other test jobs, so we won't know if they fail (unless it's a build failure).

simpsont-oci: Also, now that the wait_cancel_previous PR is merged, these should all get a "needs: wait_cancel_previous".

simpsont-oci (changes requested):
See general comments. We need to check / display the autobuild test results, and these should add a "needs: wait_cancel_previous" to help with workflow queuing / cancelation.

simpsont-oci:
The build_u18_gcc8_i0_js0_j looks like it's failing because it's running out of disk space when building bench. This is odd though, since --no-rapidjson should (and normally does) prevent bench from getting built. Can you try to figure out why it's not honoring --no-rapidjson? That alone might solve the build size issue...

ClaytonCalabrese:
> The build_u18_gcc8_i0_js0_j looks like it's failing because it's running out of disk space when building bench. This is odd though, since --no-rapidjson should (and normally does) prevent bench from getting built. Can you try to figure out why it's not honoring --no-rapidjson? That alone might solve the build size issue...

I'm assuming it is because the ACE_TAO isn't built with it.  I'll change that and see what happens.

ClaytonCalabrese:
Unfortunately the deletion of the commented parts of the safety builds has confused the hell out of git and so the changes look pretty sloppy.  Not sure why it is having so much trouble with that.

simpsont-oci:
Since the diffs are screwy I'm going to make general comments here:
- Still a lot of test failures that are "Can't spawn X" or "X doesn't exist" so we don't have the test flags quite right yet.
- Pretty sure you don't want the XERCES test flag for either safety profile build. I'm hoping that's causing most of the trouble.
- Even though they're not used anywhere at the moment, I would add the "SAFETY_BASE" and "SAFETY_EXTENDED" flags to their respective builds. It should make it easier to selectively disable tests if those flags are already set for auto_run_tests.

simpsont-oci (approved):

PR #2495, opened 2021-03-17
CMake GENERATED property is not shared between directories

ClaytonCalabrese:
I have created a test to demonstrate this issue.  The default state of this test is to show the error and as such it has not been added to any of the automated testing. In addition to the test, I created a macro that can be used to add the GENERATED property to an _export.h file.  

A better explanation of the problem is included in the test's README.


tests/cmake_integration/generated_global/run_cmake.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
mitza-oci, line 1: Did you intend to include this file?

cmake/api_macros.cmake
```diff
@@ -86,6 +86,20 @@ macro(_OPENDDS_GENERATE_EXPORT_MACRO_COMMAND  target  output)
   set(${output} ${_output_file})
 endmacro()
 
+macro(OPENDDS_MARK_AS_GENERATED target)
```
mitza-oci, line 89: Please add to documentation

mitza-oci (changes requested):

tests/cmake_integration/generated_global/.gitignore
```diff
@@ -0,0 +1,3 @@
+build
+cpp/build
+idl/build
```
iguessthislldo, line 3: You can remove this file. `build` directories are already covered by the root `.gitignore`.

cmake/api_macros.cmake
```diff
@@ -86,6 +86,20 @@ macro(_OPENDDS_GENERATE_EXPORT_MACRO_COMMAND  target  output)
...
+
+  if (NOT TARGET ${target})
```
iguessthislldo, line 91: Also check if the target is a shared library. If it's not, then there won't be an export header.

cmake/api_macros.cmake
```diff
@@ -86,6 +86,20 @@ macro(_OPENDDS_GENERATE_EXPORT_MACRO_COMMAND  target  output)
...
+    message(FATAL_ERROR "Invalid target '${target}' passed into OPENDDS_MARK_AS_GENERATED")
+  endif()
+  set(arglist ${ARGN})
+  foreach (curarg ${arglist})
+    get_property(dirs DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY INCLUDE_DIRECTORIES)
+    foreach(dir ${dirs})
+      set_source_files_properties("${dir}/${curarg}_export.h" PROPERTIES GENERATED TRUE)
+    endforeach()
+  endforeach()
+endmacro()
```
iguessthislldo, line 101: Is it possible for us to generate the header as part of the configure step? So instead of making running the script part of the dependency tree, generate it in `OPENDDS_TARGET_SOURCES`.

tests/cmake_integration/generated_global/README.md
```diff
@@ -0,0 +1,16 @@
+This test is due to the fact the GENERATED property (and properties in general) are unable to be read from those outside of the directory setting them.
```
iguessthislldo, line 1:
```suggestion
This test is due to the fact the `GENERATED` property (and properties in general) are unable to be read from those outside of the directory setting them.
```

tests/cmake_integration/generated_global/idl/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+find_package(OpenDDS REQUIRED)
+
+
+set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
+
+if (NOT OPENDDS_OWNERSHIP_PROFILE)
+  message(FATAL_ERROR "These tests require OpenDDS to be built with Ownership Profile enabled")
+endif()
+
+set(target_prefix "idl_")
```
iguessthislldo, line 19: Make `target_prefix` unique like the other tests. When using `tests/cmake_integration/CMakeLists.txt` all the targets are in the same set.

iguessthislldo (changes requested):

cmake/api_macros.cmake
ClaytonCalabrese: I'm pretty sure this would run into the same problem where we cannot check if it is a shared library because that is a property set by the library which is in a seperate directory from the one including the library.

I tried a few different ways of doing this and was running into the same NOTFOUND issue as with GENERATED.

I realize the way it is currently doing it is inefficient, but I'm not sure there is a way around that.

iguessthislldo: Okay, then this just seems like more motivation to investigate if we can just generate the export header at configure step instead of the build step. Out of curiosity I went ahead and tried this and it seems to solve the problem, at least for the test you came up with.

```diff
diff --git a/cmake/api_macros.cmake b/cmake/api_macros.cmake
index 2d3d688d5..2325be383 100644
--- a/cmake/api_macros.cmake
+++ b/cmake/api_macros.cmake
@@ -75,13 +75,15 @@ macro(_OPENDDS_GENERATE_EXPORT_MACRO_COMMAND  target  output)
     message(FATAL_ERROR "Failed to find required script 'generate_export_file.pl'")
   endif()
 
-  add_custom_command(
-    OUTPUT ${_output_file}
-    DEPENDS perl
+  execute_process(
     COMMAND ${CMAKE_COMMAND} -E env "DDS_ROOT=${DDS_ROOT}" "TAO_ROOT=${TAO_ROOT}"
-      $<TARGET_FILE:perl> ${_gen_script} ${target} $<ANGLE-R> ${_output_file}
-    VERBATIM
-  )
+      perl ${_gen_script} ${target}
+    OUTPUT_FILE ${_output_file}
+    RESULT_VARIABLE _export_script_exit_status)
+  if(NOT _export_script_exit_status EQUAL "0")
+    message(FATAL_ERROR
+      "Export header script for ${target} exited with ${_export_script_status}")
+  endif()
 
   set(${output} ${_output_file})
 endmacro()
```

cmake/api_macros.cmake
ClaytonCalabrese: I think with the new changes this is no longer necessary as there is no new macro, and the macro is built in to OpenDDS, and thus no longer requires the user to call anything manually.

iguessthislldo (changes requested):
Also I don't know if I mentioned this before, but can you add the new test's directory to [`tests/cmake_integration/CMakeLists.txt`](https://github.com/objectcomputing/OpenDDS/blob/master/tests/cmake_integration/CMakeLists.txt)?

tests/cmake_integration/generated_global/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  "${dst}/Writer.h"
+  "${dst}/Writer.cpp"
+)
+
+set_target_properties(${publisher}
+  PROPERTIES OUTPUT_NAME publisher
+)
+
+get_source_file_property(prop "${CMAKE_CURRENT_BINARY_DIR}/../idl/${messenger}_export.h" GENERATED)
+message("${messenger}_export.h GENERATED = ${prop}")
```
iguessthislldo, line 46: I think you can remove these lines.

```suggestion
```

tests/cmake_integration/generated_global/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  message(FATAL_ERROR "These tests require OpenDDS to be built with Ownership Profile enabled")
+endif()
+
+set(target_prefix "opendds_cmake_generated_global_")
+set(publisher "${target_prefix}publisher")
+set(messenger "${target_prefix}messenger")
+
+include_directories(${CMAKE_CURRENT_BINARY_DIR}/../idl)
+link_directories(${CMAKE_CURRENT_BINARY_DIR}/../idl)
+
```
iguessthislldo, line 24:
```suggestion
```

I don't think these are doing anything. Also these types of commands aren't considered good practice in CMake and the documentation warns against using them.

iguessthislldo (changes requested):
After this I think it will be ready to merge.

iguessthislldo (approved):

PR #2499, opened 2021-03-18
Add Missing Decrement to ZeroCopyDataSeq

iguessthislldo:
If a data sequence object was reused in a scenario involving a previous read of a different instance, the zero-copy count of a sample wasn't being decremented when it should've been. This resulted in an incorrect outstanding zero-copy loan error when trying to delete the data reader.

This fixes that and adds a case to the ZeroCopyRead test to try to simulate that sequence of events.


mitza-oci (approved):

mitza-oci:
`ZeroCopyRead` is failing in GitHub Actions

jrw972 (approved):

Issue #2455, opened 2021-03-01
[opendds.org] Update Article-Intro.html to use annotations


jwillemsen:
https://opendds.org/about/articles/Article-Intro.html needs to be updated to use IDL annotations instead of the deprecated pragma's

mitza-oci:
In general the articles are snapshots of that particular time, but this is a good one to change to use the new annotations.
Source is here https://github.com/objectcomputing/OpenDDS/blob/gh-pages/about/articles/Article-Intro.html

jwillemsen:
More could be done to this article, remove the usage of TimeBase to really simplify it and more code looks to be outdated compared to the example code which is in the repo

PR #2500, opened 2021-03-18
Update article-intro.html

jwillemsen:
* Use annotations
* Use register_instance instead of _cxx_register
* Removed old throw specifications

### DIFF
--- a/about/articles/Article-Intro.html
+++ b/about/articles/Article-Intro.html
@@ -252,25 +252,22 @@ without making any code changes.</p>
 #include "orbsvcs/TimeBase.idl"
 module StockQuoter
 {
-#pragma DCPS_DATA_TYPE "StockQuoter::Quote"
-#pragma DCPS_DATA_KEY "StockQuoter::Quote ticker"
+  @topic  
   struct Quote {
-    string ticker;
+    @key string ticker;
     string exchange;
     string full_name;
     double value;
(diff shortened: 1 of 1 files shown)


jwillemsen:
Fixed #2455

about/articles/Article-Intro.html
```diff
@@ -279,8 +276,8 @@ <h3><a name="IDL_Types"></a>IDL Types</h3>
 <p>We publish two data types: a <em>Quote</em> type for each
 stock quote, and an <em>ExchangeEvent</em> type to indicate when
 the stock exchange is opened, closed, and when trading is suspended or
-resumed. The <code>DCPS_DATA_TYPE</code> pragma marks a type for use with DDS. The
-<code>DCPS_DATA_KEY</code> defined for each type is a unique identifier for each
+resumed. The <code>@topic</code> annotation marks a type for use with DDS. The
+<code>@key</code> annocaton defined for each type is a unique identifier for each
```
mitza-oci, line 280:
```suggestion
<code>@key</code> annotation defined for each type is a unique identifier for each
```

mitza-oci (changes requested):

about/articles/Article-Intro.html
jwillemsen: Thanks!

mitza-oci (approved):

PR #2503, opened 2021-03-19
Use automatically detected CMake compiler for opendds_idl and tao_idl preprocessor

ClaytonCalabrese:
Previously you would need to have the compiler in the path for cmake to work with opendds_Idl and tao_idl, which could lead to errors during build time.  This PR changes the default behavior on 'win32' to instead of 'cl' have the OPENDDS_COMPILER be the path to cl.exe.


cmake/api_macros.cmake
```diff
@@ -191,6 +191,14 @@ macro(OPENDDS_TARGET_SOURCES target)
     list(APPEND _tao_options "-SS")
   endif()
 
+  if(NOT ${OPENDDS_COMPILER} STREQUAL "cl" AND ${OPENDDS_HOST} STREQUAL "win32")
```
mitza-oci, line 194: Why is this conditional?  CMake always provides `CMAKE_CXX_COMPILER` and I don't know why we wouldn't just use it.

mitza-oci (changes requested):

cmake/api_macros.cmake
iguessthislldo: I agree. You can also get rid of the configure script change.

cmake/api_macros.cmake
```diff
@@ -191,6 +191,11 @@ macro(OPENDDS_TARGET_SOURCES target)
...
+    list(APPEND _opendds_options "-Yp,${CMAKE_CXX_COMPILER}")
+    list(APPEND _tao_options "-Yp,${CMAKE_CXX_COMPILER}")
```
iguessthislldo, line 196: To allow for overriding it should check the options before appending, like above for the `-Wb,export_*` options and the `--default-nested` options.

iguessthislldo (changes requested):

cmake/api_macros.cmake
ClaytonCalabrese: You can't get rid of the configure script change because it would result in CL being put into CMAKE_CXX_COMPILER with no path.

configure
```diff
@@ -2449,6 +2449,9 @@ sub write_cmake_file {
       $var =~ s/-/_/g;
       my $val = $opts_{$k};
 
+      if($var eq "OPENDDS_COMPILER" && $val eq "cl") {
+        $val = "\${CMAKE_CXX_COMPILER}";
+      }
```
iguessthislldo, line 2454: `OPENDDS_COMPILER` isn't being used except in the tests, where it's iffy if it should be used.

mitza-oci: If the feature being added by this PR can work without a special case here, that would be preferred.

ClaytonCalabrese: The reason for this change is the tests are replacing CMAKE_CXX_COMPILER with it.  

https://github.com/objectcomputing/OpenDDS/blob/67ec6d2445a0925724fb84935729e5789b4e502a/tests/cmake_integration/Messenger/Messenger_2/CMakeLists.txt#L12

This causes problems with the main part of this change which is in the macro where we change the OpenDDS and Tao options.  

If we want to remove the CMAKE_CXX_COMPILER replacement with OPENDDS_COMPILER from our tests, then we can get rid of the configure change.

iguessthislldo: > This causes problems with the main part of this change which is in the macro where we change the OpenDDS and Tao options.

What are the problems?

> If we want to remove the CMAKE_CXX_COMPILER replacement with OPENDDS_COMPILER from our tests, then we can get rid of the configure change.

I don't really like overriding `CMAKE_CXX_COMPILER`, but looking at the original commit (117f95dcd0ca87ad8aebaa3a92100874a5c2a570 in #1240), it looks like I was trying to account for if we built OpenDDS with a compiler that's not the system compiler or what ever CMake defaults to. So we could remove them, but we'd have to make sure we're always setting the compiler if we need to.

mitza-oci: CMake's documentation of `CMAKE_<LANG>_COMPILER` says we should not use it this way (in `cmake-variables(7)`).

When running the scripts for `tests/cmake_integration`, the `cmake` command line should contain `-DCMAKE_CXX_COMPILER=foo` if that build doesn't use the system's default compiler.

If you know it's using gnuace you could run make to determine how ACE was configured:
```
 make -C $ACE_ROOT/ace -npf GNUmakefile.ACE | grep '^CXX ='
```

OpenDDS's CMake Module could (if we wanted to implement this) attempt to detect a mismatch and at least warn.

ClaytonCalabrese: > 
> What are the problems?
> 

The line

list(APPEND _opendds_options "-Yp,${CMAKE_CXX_COMPILER}")

would be putting something like "-Yp,cl" into the options, rather than "-Yp,<Path To cl.exe>"

iguessthislldo: > When running the scripts for tests/cmake_integration, the cmake command line should contain -DCMAKE_CXX_COMPILER=foo if that build doesn't use the system's default compiler.

Yes, that's what I was referring to with "setting the compiler if we need to".

> OpenDDS's CMake Module could (if we wanted to implement this) attempt to detect a mismatch and at least warn.

This might be possible, but we'd have to do some processing on both of the variables to resolve them to their canonical paths. Also [`CMAKE_CXX_COMPILER` can contain arguments](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html) which makes comparing a little harder.

> would be putting something like "-Yp,cl" into the options, rather than "-Yp,<Path To cl.exe>"

On Linux I think `tao_idl` only gets `g++` by default, so I would think letting it use the `PATH` on Windows would be okay for the tests. If you're worried about it messing with the goal of the PR, it shouldn't, because users shouldn't be using `set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})`. Only our tests should be doing that; [DevGuide Messenger, the "real" CMake example, doesn't have it](https://github.com/objectcomputing/OpenDDS/blob/master/DevGuideExamples/DCPS/Messenger/CMakeLists.txt). The tests should have compiler on the `PATH`, otherwise default `CMAKE_CXX_COMPILER` should be an absolute path, at least as far as I know.

cmake/api_macros.cmake
```diff
@@ -191,6 +191,21 @@ macro(OPENDDS_TARGET_SOURCES target)
     list(APPEND _tao_options "-SS")
   endif()
 
+  list(LENGTH ${CMAKE_CXX_COMPILER} cxx_compiler_length)
+
+  if(${cxx_compiler_length} LESS 1)
```
iguessthislldo, line 196:
```suggestion
  if(${cxx_compiler_length} EQUAL 1)
```

cmake/api_macros.cmake
```diff
@@ -191,6 +191,21 @@ macro(OPENDDS_TARGET_SOURCES target)
...
+    if(NOT "${_opendds_options}" MATCHES "-Yp")
+      list(APPEND _opendds_options "-Yp,${CMAKE_CXX_COMPILER}")
+    endif()
+
+    if(NOT "${_tao_options}" MATCHES "-Yp")
+      list(APPEND _tao_options "-Yp,${CMAKE_CXX_COMPILER}")
+    endif()
+
+  else()
+    message(FATAL_ERROR "Unable to use the default CMake CXX compiler, please add it to the path.")
```
iguessthislldo, line 206: I don't think this is a concern as long as the project is a C++ project. I would change this to say we don't support argument items in `CMAKE_CXX_COMPILER`.

iguessthislldo (changes requested):

cmake/api_macros.cmake
```diff
@@ -191,6 +191,21 @@ macro(OPENDDS_TARGET_SOURCES target)
     list(APPEND _tao_options "-SS")
   endif()
 
+  list(LENGTH ${CMAKE_CXX_COMPILER} cxx_compiler_length)
```
iguessthislldo, line 194:
```suggestion
  list(LENGTH CMAKE_CXX_COMPILER cxx_compiler_length)
```

I played around with this and I don't think it returns the right value otherwise:

```cmake
set(list1 "1")
list(LENGTH ${list1} len1)
message("List is ${len1} ${list1}") # List is 0 1
list(LENGTH list1 len1)
message("List is ${len1} ${list1}") # List is 1 1
```

iguessthislldo (changes requested):

cmake/api_macros.cmake
ClaytonCalabrese: It's 0 or undefined if it isn't a list.  

If CMAKE_CCX_COMPILER is "path/to/cl/exe", then it is 0 or undefined.  

If CMAKE_CXX_COMPILER is "path/to;cl/exe" then it is length 2.

If it is a list of any length (meaning has a ';') we don't want it in the options because it will get put in as something weird like "-Yp,path/to;cl/exe"  which will cause problems.

cmake/api_macros.cmake
ClaytonCalabrese: Yeah, I was having some issues with it when doing testing where it would jump from 0 to 2 in length.

cmake/api_macros.cmake
iguessthislldo: I totally agree with not passing it if it's a list, that's what I'm trying to avoid. This is related to comment about `list(LENGTH...)`, even though I didn't see that at first. If that's fixed then going to be 1 if `CMAKE_CXX_COMPILER` is normal.

iguessthislldo (approved):

mitza-oci (approved):

PR #2506, opened 2021-03-22
GitHub Actions: Wireshark on Windows

mitza-oci:
Also has the PR 2502 commit on this branch

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -5562,7 +5562,8 @@ jobs:
           wsbuild
           wireshark-win64-libs*
           vcpkg-qt/installed
-        key: c01_${{ github.job }}_${{ env.WIRESHARK_COMMIT }}
+          ${{ github.job }}.tar.xz
+        key: c02_${{ github.job }}_${{ env.WIRESHARK_COMMIT }}
     - name: install vcpkg packages
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       id: runvcpkg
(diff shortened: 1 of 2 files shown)


iguessthislldo (approved):

.github/workflows/build_and_test.yml
```diff
@@ -5624,13 +5699,22 @@ jobs:
 
     runs-on: windows-2019
 
-    needs: ACE_TAO_w19-release_j_FM-1f
+    needs: [ACE_TAO_w19-release_j_FM-1f, wireshark_w19-release_j_FM-1f]
```
simpsont-oci, line 5702: Update these build & test job names to have _ws_ in the name.

mitza-oci: @ClaytonCalabrese please let me know what the name should be.  We probably don't need to change the `ACE_TAO` one since nothing there depends on wireshark?

ClaytonCalabrese: That name is fine.  Basically its project type + name. so wireshark_w19-release_j_FM-1f works.

ClaytonCalabrese: Also yes, unless the ACE needs the wireshark passed in the ./configure line, then it doesn't need the wireshark dependency.

simpsont-oci: @ClaytonCalabrese I'm requesting that the build_ and test_ job names change, specifically, because they didn't used to depend on wireshark and now they do. The wireshark_ and ACE_TAO_ names seem fine to me.

simpsont-oci (approved):

PR #2509, opened 2021-03-24
Generated Global blocked from running by --no-shared

ClaytonCalabrese:
The previous change made test unfailable.  This will actually allow the test to fail while blocking it from running when an undesired error would occur.

### DIFF
--- a/tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
+++ b/tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
@@ -1,7 +1,9 @@
 project(OpenDDS_C++11_Messenger CXX)
 cmake_minimum_required(VERSION 3.8.2)
 
-find_package(OpenDDS REQUIRED)
+if(NOT OpenDDS_FOUND)
+  find_package(OpenDDS REQUIRED)
+endif()
 
 set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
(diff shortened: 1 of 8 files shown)


iguessthislldo:
We also need to exclude the test in `tests/cmake_integration/CMakeLists.txt`. I think you could do that by including `find_package(OpenDDS REQUIRED)` and making `add_subdirectory` conditional on `OPENDDS_STATIC`.

iguessthislldo (changes requested):
Should also add a fatal error for `OPENDDS_STATIC` in the project's CMakeLists.txt like there is for `OPENDDS_OWNERSHIP_PROFILE`.

tests/cmake_integration/Messenger/C++11_Messenger/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 project(OpenDDS_C++11_Messenger CXX)
 cmake_minimum_required(VERSION 3.8.2)
 
-find_package(OpenDDS REQUIRED)
+if(NOT OpenDDS_FOUND)
```
iguessthislldo, line 4: I wouldn't have thought this should be necessary, but I tried it locally and do see that it fails on multiple calls. I don't think we can fix this now, so it's fine for now. I'd like this to be possible in the future though since I think I've seen code for this is other CMake modules.

iguessthislldo (approved):

PR #2515, opened 2021-03-25
Use OPENDDS_VERSION_METADATA for opendds_idl --version

mitza-oci:


iguessthislldo (approved):

iguessthislldo (approved):
At some point we should probably should either add `OPENDDS_VERSION_METADATA` to `OPENDDS_VERSION` or add another field with the two combined. With this PR added, this would makes six separate spots ([1](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/docs/conf.py#L49-L51), [2](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/dds/DCPS/Service_Participant.cpp#L432-L435), [3](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/rules.dds.GNU#L72-L76)) where the full version string is getting constructed.

mitza-oci:
> 
> 
> At some point we should probably should either add `OPENDDS_VERSION_METADATA` to `OPENDDS_VERSION` or add another field with the two combined. With this PR added, this would makes six separate spots ([1](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/docs/conf.py#L49-L51), [2](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/dds/DCPS/Service_Participant.cpp#L432-L435), [3](https://github.com/objectcomputing/OpenDDS/blob/8f8a43bf6c99573843e517f62fa8be3ec53e2f2e/rules.dds.GNU#L72-L76)) where the full version string is getting constructed.

That's right, I didn't attempt a refactor here since it's still pretty small and simple.  If we were to refactor, of the 6 spots you mentioned 4 are in C++:
- Service_Participant
- opendds_idl --version
- opendds_idl generated comment
- idl2jni

Since we now have an OpenDDS_Util library that can share code among the core libs and generators, this could be moved there.  Or more of it could be moved to the preprocessor.

PR #2517, opened 2021-03-26
Github Actions: Make Install

ClaytonCalabrese:
This PR adds the make install capability to the u18_o1d0v1_xer0_j12 github actions build, then tests the installed OpenDDS.


.github/workflows/build_and_test.yml
```diff
@@ -318,36 +318,21 @@ jobs:
...
-      run: |
-        cd OpenDDS/ACE_TAO
-        export ACE_COMMIT=$(git rev-parse HEAD)
-        echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
-    - name: Cache Artifact
-      id: cache-artifact
-      uses: actions/cache@v2.1.4
-      with:
-        path: ${{ github.job }}.tar.xz
-        key: ${{ github.job }}_ace6tao2_${{ env.ACE_COMMIT }}
```
mitza-oci, line 350: Why does this change remove caching of ACE_TAO?  Seems like that would result in longer build times

ClaytonCalabrese: It doesn't build a seperate ace anymore.  The separate ace was not properly being brought into the make installed directory.

.github/workflows/build_and_test.yml
```diff
@@ -358,89 +343,60 @@ jobs:
...
+        cd tests/DCPS/Messenger
+        git clean -dfx .
+        $ACE_ROOT/bin/mwc.pl -type gnuace .
+        make
+        perl run_test.pl rtps
+    - name: remove install clutter
+      shell: bash
+      run: |
+        cd OpenDDS
+        rm -rf install_temp install_prefix install_setenv.sh
```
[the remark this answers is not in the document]
ClaytonCalabrese: No, I forgot to remove it. Thanks.

.github/workflows/build_and_test.yml
```diff
@@ -318,36 +318,21 @@ jobs:
   #       cat "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"
   #       if ! grep -q 'Failures: 0' "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"; then echo "::error file=latest.txt,line=0,col=0::Test Failures: see 'Test Results: ${{ github.job }}'"; fi
 
-  ACE_TAO_u18_o1d0v1_xer0_j12_2:
+  ACE_TAO_u18_o1d0v1_xer0_j12:
 
     runs-on: ubuntu-18.04
 
-    needs: wait_cancel_previous
```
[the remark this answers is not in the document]
ClaytonCalabrese: I'm actually not sure how that got removed.

.github/workflows/build_and_test.yml
simpsont-oci: This job should probably get a new name then, since it's not just an ACE_TAO job?
Also, we can remove all the "if: steps.cache-artifact.outputs.cache-hit != 'true'" checks below since we're not caching anything.
And I agree that no caching is sad times... our total build times on GHA are starting to creep upward.

mitza-oci: There is nothing inherent in `make install` that would prevent caching.  We essentially have this with Jenkins where ACE_TAO jobs are distinct from (and used by) OpenDDS jobs.

.github/workflows/build_and_test.yml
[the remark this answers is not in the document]
simpsont-oci: Still missing. You can check the checks "diagram" on the Build & Test summary page to see whether it's got the dependencies correct. wait_cancel_previous should be the only job in the far left column.

.github/workflows/build_and_test.yml
```diff
@@ -421,144 +424,65 @@ jobs:
...
-        tar cvf ../${{ github.job }}.tar setenv.sh
-        git clean -xdfn | cut -d ' ' -f 3- | xargs tar uvf ../${{ github.job }}.tar
-        xz -3 ../${{ github.job }}.tar
-    - name: upload OpenDDS artifact
-      uses: actions/upload-artifact@v2
-      with:
-        name: ${{ github.job }}_artifact
-        path: ${{ github.job }}.tar.xz
-
-  test_u18_o1d0v1_xer0_j12_2:
```
simpsont-oci, line 466: Are we getting rid of the whole test job for this one? Do we have comparable coverage elsewhere? If we are dropping it, should we at least add a java test using the install path?

simpsont-oci (changes requested):

.github/workflows/build_and_test.yml
ClaytonCalabrese: Aside from make install, this is a pretty unremarkable build, so I didn't feel that the full test suite was necessary for it.  I figured that if one or a few tests pass with make install then the rests probably are fine as well.

I am open to readding the full run of tests if you think it is necessary.

.github/workflows/build_and_test.yml
```diff
@@ -397,6 +392,7 @@ jobs:
         export DESTDIR=$(realpath install_temp)
         export INSTALL_PREFIX=$(realpath install_prefix)
         . setenv.sh
+<<<<<<< HEAD
```
iguessthislldo: Conflict markers

simpsont-oci (approved):

PR #2520, opened 2021-03-29
Fix GHA Modeling Tests SEGV Issue

simpsont-oci:


mitza-oci:
Since the configure script is getting to `ACE_ROOT/ace/config.h exists, skipping configuration of ACE+TAO`, it won't be maintaining `default.features` for us.  However we could have the build script append to `default.features` as a separate step.

simpsont-oci:
It looks like this worked as intended. @mitza-oci did you want me to change something here, or is what you're talking about something we can come back to later?

ClaytonCalabrese (approved):

PR #2524, opened 2021-03-30
RtpsRelay clients use relay's local address

jrw972:
Problem
-------

The relay advertises it's local address for SEDP.  Clients of the
relay will attempt to send messages to this address.  If the relay is
behind a firewall, then this local address will most likely not be
routable and the packets sent to it are wasted.

Solution
--------

Provide a configuration option to specify the advertised address for
the SEDP endpoints.


mitza-oci (changes requested):
Some build errors here, possibly related to IPv6

mitza-oci (approved):

PR #2525, opened 2021-03-30
Cleanup Memory Access Inside SEDP BIT Update Reverse-Locking

simpsont-oci:
Problem:
We're accessing a local map iterator within a reverse lock for the "update" case of SEDP's publication and subscription data handling.

Solution:
The "new" case is already making a copy of the relevant data to use during the reverse locking. Update the "update" case to do the same.


mitza-oci (approved):
Looks like there are build errors with built_in_topics=0, otherwise OK

jrw972 (approved):

PR #2526, opened 2021-03-30
Stop overwriting CMAKE_CXX_COMPILER

ClaytonCalabrese:
The second part of this PR is to find the scoreboard builds that need to have CMAKE_CXX_COMPILER set to something besides the default.

This PR exists because we are overwriting CMAKE_CXX_COMPILER with OPENDDS_COMPILER, which causes failures like:

[Details] CMake Error at C:/tao_builds/jenkins/workspace/dds_doc_vs2019_d1o0/OpenDDS/cmake/api_macros.cmake:206 (message):
[Details] CMake Error at C:/tao_builds/jenkins/workspace/dds_doc_vs2019_d1o0/OpenDDS/cmake/api_macros.cmake:206 (message):
[Details] MSBUILD : error MSB1009: Project file does not exist.
[Details] MSBUILD : error MSB1009: Project file does not exist.


iguessthislldo:
There is a GHA path exclude rule I wrote that's keeping this from building:

https://github.com/objectcomputing/OpenDDS/blob/b5e19781ae387c295c15cef323248f31df763816/.github/workflows/build_and_test.yml#L55

I didn't think of CMake files being `.txt` files when I wrote it. Could you remove line that and the other `'!**.txt'` line from the excludes?

https://github.com/objectcomputing/OpenDDS/blob/b5e19781ae387c295c15cef323248f31df763816/.github/workflows/build_and_test.yml#L28

iguessthislldo:
`CMAKE_CXX_COMPILER` should be set when the compiler we used to build OpenDDS wasn't the system compiler.

Here for example:

https://github.com/objectcomputing/OpenDDS/blob/9941a92fd5425cd31483673ca0332f93f723766e/.github/workflows/build_and_test.yml#L1361-L1363

we're using `clang++`, so `CMAKE_CXX_COMPILER` should be set to that.

Also I noticed that when we're trying to use a gcc that's not the system compiler we're setting `gcc` to the version we installed, but not `g++`, so it's building with the `g++` system compiler:

https://github.com/objectcomputing/OpenDDS/blob/9941a92fd5425cd31483673ca0332f93f723766e/.github/workflows/build_and_test.yml#L1799-L1804

.github/workflows/build_and_test.yml
```diff
@@ -1838,7 +1838,7 @@ jobs:
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       run: |
         cd OpenDDS
-        ./configure --compiler=g++ --no-debug --features=uses_wchar=1 --std=c++03 --tests --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
+        ./configure --compiler=gcc --no-debug --features=uses_wchar=1 --std=c++03 --tests --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```
iguessthislldo, line 1841: `gcc` is a C compiler, so I think this will break the build. What I meant was that the `update-alternatives` command should be changed to `g++`, but I guess I didn't make that clear.

iguessthislldo (changes requested):

.github/workflows/build_and_test.yml
mitza-oci: Why do we need `update-alternatives`?

iguessthislldo: I don't know if we do. I was just trying to work with what's already here. I'm guessing @simpsont-oci had a reason for using it?

simpsont-oci: Wasn't me! Then again, I didn't make any of the jobs that use non-default compilers (except ASAN, but that's still just the default version of clang).

iguessthislldo: Sorry, I assumed since it was here and seemed like like frequent pattern, you wrote it. Looking at the actual git history I see now that Clayton added them.

ClaytonCalabrese: I'm not sure what the objection to update-alternatives is? I am using it to specify which version of gcc/g++/clang I want to be used for the build.  Doing prints of gcc --version or g++ --version show that it is using the one I specified.

mitza-oci: CI testing should ensure that however the user specifies the compiler, the combination of `configure` and CMake needs to handle it.  In order to test that this works, the test cases should be more than just the system's default compiler

This seems to be outside the scope of the current PR so maybe just note it and move on.

.github/workflows/build_and_test.yml
```diff
@@ -1840,7 +1838,7 @@ jobs:
...
```
mitza-oci, line 1841:
```suggestion
        ./configure --no-debug --features=uses_wchar=1 --std=c++03 --tests --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```

.github/workflows/build_and_test.yml
```diff
@@ -1912,7 +1910,7 @@ jobs:
     - name: configure OpenDDS
       run: |
         cd OpenDDS
-        ./configure --compiler=g++ --no-debug --features=uses_wchar=1 --std=c++03 --tests --rapidjson --ace=$GITHUB_WORKSPACE/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
+        ./configure --compiler=gcc --no-debug --features=uses_wchar=1 --std=c++03 --tests --rapidjson --ace=$GITHUB_WORKSPACE/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```
mitza-oci, line 1913:
```suggestion
        ./configure --no-debug --features=uses_wchar=1 --std=c++03 --tests --rapidjson --ace=$GITHUB_WORKSPACE/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```

mitza-oci: apply to all of these - no reason to give `configure` a command line option which is also the default

DevGuideExamples/DCPS/Messenger/CMakeLists.txt
```diff
@@ -3,8 +3,6 @@ cmake_minimum_required(VERSION 3.8.2)
 
 find_package(OpenDDS REQUIRED)
 
-set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
```
mitza-oci, line 6: To make this work for systems that don't use the default compiler, the script that invokes `cmake` needs to pass `-DCMAKE_CXX_COMPILER=<something>`.

mitza-oci (changes requested):

.github/workflows/build_and_test.yml
iguessthislldo: Another con of `update-alternatives`, at least in Ubuntu 18.04, is that `CMAKE_CXX_COMPILER` is `/usr/bin/c++` by default, which would have to be to set in `update-alternatives`. We could make `update-alternatives` work by setting `c++` and `g++`/`clang++`, but it'd be more complicated than just passing `clang++-X` or `g++-X` to `configure` and `cmake`.

.github/workflows/build_and_test.yml
```diff
@@ -1291,7 +1289,7 @@ jobs:
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       run: |
         cd OpenDDS
-        ./configure --compiler=clang++ --std=c++11 --tests --security --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
+        ./configure --compiler=clang++ --std=c++11 --tests --security --no-rapidjson --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```
mitza-oci, line 1292: RapidJSON appears to be unrelated - is the change needed here?

.github/workflows/build_and_test.yml
```diff
@@ -1644,26 +1634,12 @@ jobs:
         . setenv.sh
         cd tests/cmake_integration
         perl run_ci_tests.pl --skip-cxx11
-    - name: create OpenDDS tar.xz artifact
```
mitza-oci, line 1647: Why was this part removed?

tests/cmake_integration/run_ci_tests.pl
```diff
@@ -64,6 +66,10 @@ ($)
     "-D", "CMAKE_PREFIX_PATH=$ENV{'DDS_ROOT'}",
     "-D", "CMAKE_VERBOSE_MAKEFILE:BOOL=ON",
   );
+  if($compiler ne "") {
```
mitza-oci, line 69:
```suggestion
  if ($compiler ne "") {
```

mitza-oci (changes requested):

.github/workflows/build_and_test.yml
ClaytonCalabrese: Because we aren't using that build for testing so artifacting doesn't help with anything.

.github/workflows/build_and_test.yml
ClaytonCalabrese: I've found it is best to match both of the configures (ACE + Build).

.github/workflows/build_and_test.yml
mitza-oci: Let's keep that kind of change in its own PR

.github/workflows/build_and_test.yml
```diff
@@ -1644,26 +1634,12 @@ jobs:
         . setenv.sh
         cd tests/cmake_integration
         perl run_ci_tests.pl --skip-cxx11
```
iguessthislldo, line 1636: This and all the other jobs that are using the non default `g++` need the compiler explicitly set because of the issue I mentioned [here](https://github.com/objectcomputing/OpenDDS/pull/2526#discussion_r605277436).

iguessthislldo (changes requested):

mitza-oci (approved):

.github/workflows/build_and_test.yml
```diff
@@ -1034,7 +1030,7 @@ jobs:
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       run: |
         cd OpenDDS
-        ./configure --compiler=clang++ --no-inline --features=uses_wchar=1 --std=c++11 --tests --security --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
+        ./configure --compiler=/usr/bin/clang++-5.0 --no-inline --features=uses_wchar=1 --std=c++11 --tests --security --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```
iguessthislldo, line 1033:
```suggestion
        ./configure --compiler=clang++-5.0 --no-inline --features=uses_wchar=1 --std=c++11 --tests --security --ace=$GITHUB_WORKSPACE/OpenDDS/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
```

I think we can use command name instead of an absolute path. Same thing for CMake script and explicit g++ builds.

Same for all the other places we're explicitly setting the file.

.github/workflows/build_and_test.yml
```diff
@@ -1500,21 +1492,16 @@ jobs:
...
-    - name: install gcc
+    - name: install gcc and g++
       run: |
-        sudo apt-get install -y gcc-6
-        sudo update-alternatives \
-          --install /usr/bin/gcc gcc /usr/bin/gcc-6 100 \
-          --slave /usr/bin/gcc-ar gcc-ar /usr/bin/gcc-ar-6 \
-          --slave /usr/bin/gcc-ranlib gcc-ranlib /usr/bin/gcc-ranlib-6 \
-          --slave /usr/bin/gcov gcov /usr/bin/gcov-6
+        sudo apt-get install gcc-6 g++-6
```
iguessthislldo, line 1504:
```suggestion
        sudo apt-get install g++-6
```

We don't use the gcc C compiler, although it will get installed anyway because the g++-X packages depend on the corresponding gcc-X packages.

.github/workflows/build_and_test.yml
```diff
@@ -2298,14 +2266,10 @@ jobs:
...
-    - name: install gcc
+    - name: install gcc and g++
       run: |
-        sudo apt-get install -y gcc-10
-        sudo update-alternatives \
-          --install /usr/bin/gcc gcc /usr/bin/gcc-10 100 \
-          --slave /usr/bin/gcc-ar gcc-ar /usr/bin/gcc-ar-10 \
-          --slave /usr/bin/gcc-ranlib gcc-ranlib /usr/bin/gcc-ranlib-10 \
-          --slave /usr/bin/gcov gcov /usr/bin/gcov-10
+        echo $PATH
```
iguessthislldo, line 2271:
```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -2150,7 +2116,9 @@ jobs:
     - name: configure OpenDDS
       run: |
         cd OpenDDS
-        ./configure --compiler=g++ --no-inline --features=uses_wchar=1 --tests --rapidjson --ace=$GITHUB_WORKSPACE/ACE_TAO/ACE --mpc=$GITHUB_WORKSPACE/MPC
+        gcc --version
+        g++ --version
```
iguessthislldo, line 2120:
```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -2106,21 +2075,18 @@ jobs:
...
       run: |
-        sudo apt-get install -y gcc-10
-        sudo update-alternatives \
-          --install /usr/bin/gcc gcc /usr/bin/gcc-10 100 \
-          --slave /usr/bin/gcc-ar gcc-ar /usr/bin/gcc-ar-10 \
-          --slave /usr/bin/gcc-ranlib gcc-ranlib /usr/bin/gcc-ranlib-10 \
-          --slave /usr/bin/gcov gcov /usr/bin/gcov-10
+        echo $PATH
+        sudo apt-get install gcc-10 g++-10
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 100 --slave /usr/bin/g++ g++ /usr/bin/g++-10
```
iguessthislldo, line 2089:
```suggestion
        sudo apt-get install g++-10
```

.github/workflows/build_and_test.yml
```diff
@@ -1577,21 +1564,17 @@ jobs:
...
+    - name: install gcc and g++
       run: |
-        sudo apt-get install -y gcc-6
-        sudo update-alternatives \
-          --install /usr/bin/gcc gcc /usr/bin/gcc-6 100 \
-          --slave /usr/bin/gcc-ar gcc-ar /usr/bin/gcc-ar-6 \
-          --slave /usr/bin/gcc-ranlib gcc-ranlib /usr/bin/gcc-ranlib-6 \
-          --slave /usr/bin/gcov gcov /usr/bin/gcov-6
+        sudo apt-get install gcc-6 g++-6
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-6 100 --slave /usr/bin/g++ g++ /usr/bin/g++-6
```
iguessthislldo, line 1577:
```suggestion
```

iguessthislldo (approved):
If we want to merge this today, then that would be okay since I don't see anything here that keeps us from reaching the PR's goal. It's just cleanup.

.github/workflows/build_and_test.yml
mitza-oci: FYI libACE does use `$(CC)` in place of `$(CXX)` for a few translation units.
I agree with letting the apt dependencies sort it out.

iguessthislldo: Huh, I have seen the C support in gnuace, but I didn't think it was actually using it.

iguessthislldo (approved):

PR #2528, opened 2021-03-30
Github Actions: Attempt To Fix 'wireshark_w19-release' Caching

simpsont-oci:


.github/workflows/build_and_test.yml
```diff
@@ -5061,12 +5061,7 @@ jobs:
       id: cache-artifact
       uses: actions/cache@v2.1.4
       with:
-        path: |
-          winflexbison/bin/Debug
-          wsbuild
-          wireshark-win64-libs*
-          vcpkg-qt/installed
-          ${{ github.job }}.tar.xz
+        path: ${{ github.job }}.tar.xz
```
mitza-oci, line 5064: Does this need an "extract-archive" step directly after this one?

In the first revision of this (before dealing with tar files manually), we were depending on the Cache Artifact action to put things back where they were.

mitza-oci (changes requested):

.github/workflows/build_and_test.yml
simpsont-oci: I don't think so. If the cache action is a hit, it should download the tar.xz file and skip every step except uploading it as an artifact. Extracting the wireshark artifact is done in the job 'build_w19-release_j_ws_FM-1f'

mitza-oci (approved):

PR #2531, opened 2021-03-31
Directed SPDP messages are sent via multicast

jrw972:
Problem
-------

In some environments, unicast is scheduled before multicast.  For
completing authentication, this causes a delay since the SPDP message
that is necessary to start authentication on a remote may be delivered
after the authentication messages (even though the code clearly sends
them in the correct order).

Solution
--------

Send directed SPDP messages via unicast.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -663,6 +663,9 @@ Spdp::handle_participant_data(DCPS::MessageId id,
     iter = p.first;
     iter->second.discovered_at_ = now;
     update_lease_expiration_i(iter, now);
+    if (!from_relay && from != ACE_INET_Addr()) {
+      iter->second.local_address_ = from;
+    }
 
 #ifndef DDS_HAS_MINIMUM_BIT
     if (!from_sedp) {
@@ -680,8 +683,8 @@ Spdp::handle_participant_data(DCPS::MessageId id,
     // Since we've just seen a new participant, let's send out our
     // own announcement, so they don't have to wait.
     if (from != ACE_INET_Addr()) {
-      if (from_relay) {
-        tport_->write_i(guid, SpdpTransport::SEND_TO_RELAY);
+      if (!from_relay && from_relay) {
+        tport_->write_i(guid, iter->second.local_address_, SpdpTransport::SEND_RELAY);
       } else {
         tport_->shorten_local_sender_delay_i();
       }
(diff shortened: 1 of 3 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2161,8 +2170,9 @@ bool Spdp::announce_domain_participant_qos()
 }
 
 #if !defined _MSC_VER || _MSC_VER > 1700
-const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_TO_LOCAL;
-const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_TO_RELAY;
+const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_MULTICAST;
+const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_RELAY;
+ const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_DIRECT;
```
mitza-oci, line 2175:
```suggestion
const Spdp::SpdpTransport::WriteFlags Spdp::SpdpTransport::SEND_DIRECT;
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2664,24 +2674,29 @@ Spdp::SpdpTransport::write_i(const DCPS::RepoId& guid, WriteFlags flags)
...
 
-  if ((flags & SEND_TO_LOCAL) && !outer->config_->rtps_relay_only()) {
+  if ((flags & SEND_MULTICAST) && !outer->config_->rtps_relay_only()) {
     typedef OPENDDS_SET(ACE_INET_Addr)::const_iterator iter_t;
     for (iter_t iter = send_addrs_.begin(); iter != send_addrs_.end(); ++iter) {
       send(*iter);
     }
   }
 
+  if (((flags & SEND_DIRECT) || !outer->config_->rtps_relay_only()) &&
```
mitza-oci, line 2693: Why is this an `||`?  If the command is to send it "directly" and the configuration is "relay only," should that prevent this `send()`?

mitza-oci (changes requested):

mitza-oci (approved):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -680,8 +683,8 @@ Spdp::handle_participant_data(DCPS::MessageId id,
     // Since we've just seen a new participant, let's send out our
     // own announcement, so they don't have to wait.
     if (from != ACE_INET_Addr()) {
-      if (from_relay) {
-        tport_->write_i(guid, SpdpTransport::SEND_TO_RELAY);
+      if (!from_relay && from_relay) {
```
iguessthislldo, line 686: Wouldn't this always be false?

iguessthislldo (changes requested):

dds/DCPS/RTPS/Spdp.cpp
jrw972: Good catch.  That probably saved me hours of debugging.

PR #2534, opened 2021-04-01
RTPS sends too many little messages

jrw972:
Problem
-------

For reliability, RTPS sends small messages like heartbeats, acknacks,
and gaps.  When the RtpsUdpDataLink was using consolidate timers,
these messages were bundled together which had the advantage of
reducing the overhead associated with sending a bunch of small
messages.

Solution
--------

Add a send queue to the RtpsUdpDataLink that is sent whenever it is
full (enough) or after a configurable delay.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -81,6 +81,7 @@ RtpsUdpDataLink::RtpsUdpDataLink(RtpsUdpTransport& transport,
   , reactor_task_(reactor_task)
   , job_queue_(make_rch<JobQueue>(reactor_task->get_reactor()))
   , multi_buff_(this, config.nak_depth_)
+  , flush_send_queue_task_(reactor_task->interceptor(), *this, &RtpsUdpDataLink::flush_send_queue)
   , best_effort_heartbeat_count_(0)
   , heartbeat_(reactor_task->interceptor(), *this, &RtpsUdpDataLink::send_heartbeats)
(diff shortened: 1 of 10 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2462,50 +2464,49 @@ RtpsUdpDataLink::bundle_mapped_meta_submessages(
...
+  MetaSubmessageVecVec temp;
+
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, send_queue_lock_);
+
+    send_queue_.push_back(MetaSubmessageVec());
+    send_queue_.back().swap(in);
+    total_submessage_count_ += send_queue_.back().size();
+
+    // Heartbeats and Acknacks are around 8 longs.
```
mitza-oci, line 2496: Worst case Gap and AckNack are about 16.

simpsont-oci: I'm not sure where to put my comment, so I'll just add it here:
- I worry that this PR is removing functionality (the ability to temporarily enforce bundling for a group of actions taken by a thread (e.g. in response to a single packet or in response to a single timer)) which might prove beneficial later.
- I worry that the only division in packets now is time & size, meaning our Wireshark output might start to look a whole lot more "jumbled".
- If we're going to go this route, I would suggest we at least have separate queues for readers & writers for the sake of making packet inspection less chaotic.

mitza-oci:
examples/DCPS/ishapes/main.cpp#L69

‘class OpenDDS::DCPS::RtpsUdpInst’ has no member named ‘handshake_timeout_’

simpsont-oci (approved):

iguessthislldo (approved):

PR #2537, opened 2021-04-01
Delayed duplicate acknack leads to invalid heartbeat

jrw972:
Problem
-------

Assume that two peers are local but also using the RtpsRelay.  A
preassociation acknack can be delivered after a postassociation
acknack.  This may be interpretted as a reset by the writer.  For
normal RTPS writers, this does not present a problem as it sends
another heartbeat and eventually gets the correct ack.

For the volatile writer in security, it presents a problem since the
value used for the heartbeat last is tracked separately.  If the data
has already been acknowledged, then the writer will send an invalid
heartbeat.  The reader will ignore the invalid heartbeat.  The writer
may keep sending the invalid heartbeat.

Solution
--------

Adjust the heartbeat for the volatile writer so that it is correct.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1920,7 +1920,9 @@ RtpsUdpDataLink::RtpsWriter::add_reader(const ReaderInfo_rch& reader)
   ReaderInfoMap::const_iterator iter = remote_readers_.find(reader->id_);
   if (iter == remote_readers_.end()) {
 #ifdef OPENDDS_SECURITY
-    reader->max_pvs_sn_ = max_sn_;
+    if (is_pvs_writer_) {
+      reader->max_pvs_sn_ = max_sn_;
+    }
 #endif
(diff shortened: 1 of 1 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3819,12 +3831,21 @@ RtpsUdpDataLink::RtpsWriter::gather_directed_heartbeat_i(const SingleSendBuffer:
                                                          const ReaderInfo_rch& reader)
 {
   const SequenceNumber first_sn = reader->durable_ ? 1 : non_durable_first_sn(proxy);
-  const SequenceNumber last_sn = expected_max_sn(reader);
+  SequenceNumber last_sn = expected_max_sn(reader);
```
mitza-oci, line 3834: `ACE_UNUSED_ARG` can help avoid a warning when security and asserts are both disabled

mitza-oci: Nevermind... I missed that it was used on another line

mitza-oci (approved):

PR #2538, opened 2021-04-01
Github Actions: Release Builds

ClaytonCalabrese:
Release configuration added for github actions. Due to size of builds, there needed to be removal of tests, which will be added later in another PR>

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4777,7 +4777,7 @@ jobs:
   #       cat "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"
   #       if ! grep -q 'Failures: 0' "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"; then echo "::error file=latest.txt,line=0,col=0::Test Failures: see 'Test Results: ${{ github.job }}'"; fi
 
-  ACE_TAO_w19-release_o1p1_sec_2:
+  ACE_TAO_w19-release_o1p1_sec_3:
 
     runs-on: windows-2019
 
@@ -4859,7 +4859,7 @@ jobs:
 
     runs-on: windows-2019
 
-    needs: ACE_TAO_w19-release_o1p1_sec_2
+    needs: ACE_TAO_w19-release_o1p1_sec_3
 
     steps:
     - name: install openssl-windows & xerces-c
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -4505,7 +4505,7 @@ jobs:
   #       cat "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"
   #       if ! grep -q 'Failures: 0' "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"; then echo "::error file=latest.txt,line=0,col=0::Test Failures: see 'Test Results: ${{ github.job }}'"; fi
 
-  ACE_TAO_w19_p1_cpp03_stat_FM-08_2:
+  ACE_TAO_w19_p1_cpp03_stat_FM-08_3:
```
simpsont-oci, line 4508: Rather than messing with the job names, the cleaner approach is probably just to append a unique string to the cache key for the ACE_TAO job. This forces a cache rebuild, all the jobs and artifact names can stay the same. We will probably need a task / PR to clean up all the suffixes from job names (I think one of the mac names is missing a number as well), so don't feel like you need to do this here... it's just worth noting for future reference.

.github/workflows/build_and_test.yml
```diff
@@ -5117,14 +5126,22 @@ jobs:
...
         path: ${{ github.job }}.tar.xz
 
-
-  ACE_TAO_w19-release_j_FM-1f:
+  ACE_TAO_w19-release_j_ws_FM-1f_2:
 
     runs-on: windows-2019
 
-    needs: wait_cancel_previous
+    needs: [wait_cancel_previous, wireshark_w19-release]
```
simpsont-oci, line 5133: This seems unnecessary. Why does the ACE_TAO job need to be aware of wireshark?

.github/workflows/build_and_test.yml
mitza-oci: Right, see the wireshark one:
```
        key: c01_${{ github.job }}_${{ env.WIRESHARK_COMMIT }}
```
The "c01" is "cache 1" so you can increment it whenever necessary

.github/workflows/build_and_test.yml
ClaytonCalabrese: I've found it best to match the ACE_TAO configure with the build configure.  I don't think this dependency causes any noticeable slowdown.

simpsont-oci (approved):

PR #2540, opened 2021-04-02
Zero-count acknacks may cause reset

jrw972:
Problem
-------

For background, see #2537.  A problem with the initial solution is
that resets may still occur due to the way zero-count acknacks are
handled.

A zero-count acknack is used for preassociation when reflecting
heartbeat counts.  Delivering a duplicate zero-count acknack late
causes a reset.

Solution
--------

Provide special handling for zero-count acknacks.  Specifically, they
need to be processed but they should only be used for completing
association.

This commit also resets the durable data to avoid a memory leak.


mitza-oci (approved):

ClaytonCalabrese (approved):

ClaytonCalabrese:
I don't recall seeing the "u18_o1d0v1_xer0_j12_2" build error before.  It doesn't look to be related to these changes though.

ClaytonCalabrese:
The u18_o1d0v1_xer0_j12_2 failure is weird but probably unrelated to this PR.

PR #2542, opened 2021-04-02
Github Actions: Static Tests

ClaytonCalabrese:
Due to space issues on building all the tests for static builds, I have created a subset of tests to run.  This includes a script to run the subset of build tests, which we pass to auto_run_tests.pl.  I'd like to add the unit tests as well but I need to check how much disk space we have left.


.github/workflows/build_and_test.yml
```diff
@@ -4650,6 +4761,42 @@ jobs:
...
+        perl %ACE_ROOT%\bin\mwc.pl -type vs2019 -static
+    - name: make compiler tests
+      shell: cmd
+      run: |
+        cd OpenDDS
+        call setenv.cmd
+        cd tests\DCPS\Compiler
+        dir
+        msbuild -p:Configuration=Debug,Platform=x64 -m Compiler.sln
+    - name: run Static CI tests
```
simpsont-oci, line 4793: Let's keep the pattern of running the tests in the test job. And especially for the static test jobs, which may build & run separate tests, so they can reuse the core OpenDDS build artifact.

.github/workflows/build_and_test.yml
```diff
@@ -4378,6 +4442,42 @@ jobs:
         call setenv.cmd
         cd tests/cmake_integration/
         perl run_ci_tests.pl --no-shared --build-config=Debug
+    - name: build messenger tests
```
simpsont-oci, line 4445: Should we make a separate mwc file for this subset of tests to match what's in the new test list? We can re-call mwc after the configure step if we don't feel like changing configure just for the static builds.

ClaytonCalabrese: mwc file would be found recursively so I'm hesitant to create a new mwc file rather than just manually running it.

tests/static_ci_tests.lst
```diff
@@ -0,0 +1,40 @@
+tests/DCPS/C++11/Messenger/run_test.pl: !DCPS_MIN CXX11 !STATIC_COMPILER !STATIC_UNIT
```
simpsont-oci, line 1: Is there any reason these aren't specified as single-flag positives (e.g. "STATIC_MESSENGER") rather than multi-flag negatives (e.g. "!STATIC_COMPILER !STATIC_UNIT")?

.github/workflows/build_and_test.yml
```diff
@@ -3612,16 +3610,20 @@ jobs:
...
         path: ${{ github.job }}.tar.xz
 
-  devguide_u18_stat_qt_ws_sec:
+
+  cmake_u18_stat_qt_ws_sec:
 
     runs-on: ubuntu-18.04
 
     needs: build_u18_stat_qt_ws_sec
-    timeout-minutes: 150
```
simpsont-oci, line 3620: Let's keep / add the timeout to all the final test/cmake/etc jobs.

.github/workflows/build_and_test.yml
```diff
@@ -4393,100 +4587,1649 @@ jobs:
...
+        tar cvf ../${{ github.job }}.tar setenv.cmd
+        git clean -xdfn | cut -d ' ' -f 3- | xargs tar uvf ../${{ github.job }}.tar
+        xz -3 ../${{ github.job }}.tar
+    - name: upload OpenDDS artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: ${{ github.job }}_artifact
+        path: ${{ github.job }}.tar.xz
+
+  messenger_w19_re_p1_stat_FM-08:
```
simpsont-oci, line 5150: Any thoughts on why the subscriber isn't getting built for tests/DCPS/C++11/Messenger/run_test.pl on this one? I think one other test run had the same complaint.
```
ERROR: executable "Subscriber/subscriber" does not exist; subdir: Static_Release//; basename: subscriber ; dirname: Subscriber/
```

simpsont-oci (commented):
Not a complete review, but overall it looks good. Here are a few things that probably need to be addressed before merging.

.github/workflows/build_and_test.yml
ClaytonCalabrese: It is in another directory I need another mwc call for it I think.

simpsont-oci (approved):

PR #2546, opened 2021-04-05
Remove SequenceBackInsertIterator for OctetSeq

simpsont-oci:
Problem:
SequenceBackInsertIterator for OctetSeq is particularly inefficient due to repeated resize / memcpy calls for what can usually be done with a single resize & memcpy.

Solution:
Avoid using SequenceBackInsertIterator for OctetSeq.

### DIFF
--- a/dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
+++ b/dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
@@ -84,10 +84,11 @@ bool LocalAuthCredentialData::load_access_permissions(const DDS::Security::Permi
     return false;
   }
 
-  DCPS::SequenceBackInsertIterator<DDS::OctetSeq> back_inserter(access_permissions_);
-  std::copy(cperm, cperm + std::strlen(cperm), back_inserter);
+  size_t len = std::strlen(cperm);
+  access_permissions_.length(len + 1);
+  std::memcpy(&access_permissions_[0], cperm, len);
+  access_permissions_[len] = 0u;
 
-  *back_inserter = 0u;
   return true;
 }
 
--- a/dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
+++ b/dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
@@ -84,10 +84,9 @@ bool LocalAuthCredentialData::load_access_permissions(const DDS::Security::Permi
     return false;
   }
 
-  size_t len = std::strlen(cperm);
-  access_permissions_.length(len + 1);
-  std::memcpy(&access_permissions_[0], cperm, len);
-  access_permissions_[len] = 0u;
+  const size_t len = std::strlen(cperm);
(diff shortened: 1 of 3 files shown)


dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
```diff
@@ -84,10 +84,11 @@ bool LocalAuthCredentialData::load_access_permissions(const DDS::Security::Permi
     return false;
   }
 
-  DCPS::SequenceBackInsertIterator<DDS::OctetSeq> back_inserter(access_permissions_);
-  std::copy(cperm, cperm + std::strlen(cperm), back_inserter);
+  size_t len = std::strlen(cperm);
```
mitza-oci, line 87:
```suggestion
  const size_t len = std::strlen(cperm);
```

dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
```diff
@@ -84,10 +84,11 @@ bool LocalAuthCredentialData::load_access_permissions(const DDS::Security::Permi
...
+  access_permissions_.length(len + 1);
```
mitza-oci, line 88: 32-bit/64-bit mismatch - probably generating a warning somewhere

dds/DCPS/security/Authentication/LocalAuthCredentialData.cpp
```diff
@@ -84,10 +84,11 @@ bool LocalAuthCredentialData::load_access_permissions(const DDS::Security::Permi
...
+  std::memcpy(&access_permissions_[0], cperm, len);
+  access_permissions_[len] = 0u;
```
mitza-oci, line 90:
```suggestion
  std::memcpy(&access_permissions_[0], cperm, len + 1);
```

mitza-oci (changes requested):

dds/DCPS/security/SSL/Certificate.cpp
```diff
@@ -435,6 +434,13 @@ void Certificate::load_cert_bytes(const std::string& path)
 
 #else
   std::ifstream in(path.c_str(), std::ios::binary);
+  const std::ifstream::pos_type begin = in.tellg();
```
mitza-oci, line 437: Isn't this guaranteed to be 0 for a file opened (without "append")?
Does it need to check if the stream is open first?

mitza-oci (changes requested):

dds/DCPS/security/SSL/Certificate.cpp
simpsont-oci: I'm not sure about what's guaranteed, and cppreference isn't giving me a warm fuzzy feeling about it either.
I'll add an earlier check, but when in.fail() is true I'm pretty sure those pos_type values come back as -1 and the rest of it is mostly a no-op.

mitza-oci (approved):

PR #2548, opened 2021-04-05
Cleanup Leak in ReactorTask

simpsont-oci:
Problem:
On reuse of ReactorTask, any existing reactors and/or proactors are potentially leaked.

Solution:
Don't do that.


dds/DCPS/ReactorTask.cpp
```diff
@@ -39,23 +38,37 @@ ReactorTask::ReactorTask(bool useAsyncSend)
 }
 
 ReactorTask::~ReactorTask()
+{
+  cleanup();
+}
+
+void
+ReactorTask::cleanup()
```
iguessthislldo, line 46:
```suggestion
void ReactorTask::cleanup()
```

Rest of the file uses the one line signature style.

dds/DCPS/ReactorTask.cpp
```diff
@@ -64,51 +77,42 @@ int ReactorTask::open_reactor_task(void*, TimeDuration timeout,
...
+    }
 #endif
   } else {
-    reactor_ = new ACE_Reactor(new ACE_Select_Reactor, true);
-    proactor_ = 0;
+    if (!reactor_) {
+      reactor_ = new ACE_Reactor(new ACE_Select_Reactor, true);
+      proactor_ = 0;
+    }
   }
```
iguessthislldo, line 98: It's bothered me for awhile that the else blocks are the same here. Maybe something like this?

```c++
#if defined (ACE_WIN32) && defined (ACE_HAS_WIN32_OVERLAPPED_IO)
  if (use_async_send_ && !reactor_) {
    reactor_ = new ACE_Reactor(new ACE_WFMO_Reactor, 1);

    ACE_WIN32_Proactor* proactor_impl = new ACE_WIN32_Proactor(0, 1);
    proactor_ = new ACE_Proactor(proactor_impl, 1);
    reactor_->register_handler(proactor_impl, proactor_impl->get_handle());
  } else
#endif
  if (!reactor_) {
    reactor_ = new ACE_Reactor(new ACE_Select_Reactor, true);
    proactor_ = 0;
  }
```

iguessthislldo (changes requested):

mitza-oci:
Could we add the execute bit check to lint for `*.pl`?

iguessthislldo:
> Could we add the execute bit check to lint for *.pl?

Yes. I have though about this before, but for the opposite case, where there's a file that has the execute bit that shouldn't.

iguessthislldo (approved):

PR #2550, opened 2021-04-05
Fix uninitialized members

jrw972:


dds/DCPS/AssociationData.h
```diff
@@ -28,6 +28,7 @@ struct AssociationData {
 
   AssociationData()
     : remote_id_(GUID_UNKNOWN)
+    , participant_discovered_at_({ 0, 0})
```
mitza-oci, line 31: I think this fails for older compilers.  It may be easier to have a static function that returns the object.

mitza-oci (changes requested):

dds/DCPS/AssociationData.h
iguessthislldo: `MonotonicTimePoint::zero_value`?

mitza-oci (approved):

PR #2557, opened 2021-04-07
Reuse instance handles

mitza-oci:


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1930,6 +1931,7 @@ DDS::InstanceHandle_t DomainParticipantImpl::lookup_handle(const GUID_t& id) con
 {
   ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, handle_protector_, DDS::HANDLE_NIL);
   const CountedHandleMap::const_iterator iter = handles_.find(id);
+//  OPENDDS_ASSERT(id == GUID_UNKNOWN || iter != handles_.end());
```
jrw972, line 1934: assertion?

mitza-oci: I didn't intend to include that.  It was helpful for initial testing but not really an assertion that must be true in all cases.

iguessthislldo (approved):

PR #2559, opened 2021-04-07
Workaround TAO3 Setting Our SONAME Version

iguessthislldo:
TAO3 added a [`tao_rules` base project](https://github.com/mcorino/ACE_TAO/blob/master/TAO/MPC/config/tao_rules.mpb) in https://github.com/DOCGroup/ACE_TAO/pull/1296. This base project serves the same purpose as our [`dds_macros` base project](https://github.com/objectcomputing/OpenDDS/blob/master/MPC/config/dds_macros.mpb), which is to insert a `rules.*.GNU` file into the final Makefile. The problem is that `rules.tao.GNU` is getting inserted before ours and that's setting `GNUACE_PROJECT_VERSION` first. This results in OpenDDS so libraries getting TAO3's version.

This change moves `dds_macros` before TAO base projects everywhere I could find to try to make sure our rules file is included first.

This is just a workaround though. The underlying issue is that `GNUACE_PROJECT_VERSION` can easily bleed between different projects. This would have to be fixed in gnuace.

### DIFF
--- a/MPC/config/dcps_no_idl.mpb
+++ b/MPC/config/dcps_no_idl.mpb
@@ -1,4 +1,4 @@
-project: dds_taolib, dds_macros, dcps_optional_bidir_giop, dds_any_support, coverage_optional, dds_vc_warnings {
+project: dds_macros, dds_taolib, dcps_optional_bidir_giop, dds_any_support, coverage_optional, dds_vc_warnings {
   libs        += OpenDDS_Dcps
   after       += OpenDDS_Dcps
   libpaths    += $(DDS_ROOT)/lib
--- a/MPC/config/idl2jni.mpb
+++ b/MPC/config/idl2jni.mpb
@@ -2,7 +2,7 @@ feature(!java_pre_jpms) : java {
   javacflags += -classpath $(DDS_ROOT)/lib/i2jrt_corba.jar
 }
 
-project: taolib_with_idl, portableserver, java, dds_macros {
+project: dds_macros, taolib_with_idl, portableserver, java {
 
   includes   += $(DDS_ROOT)/java/idl2jni/runtime
   libpaths   += $(DDS_ROOT)/lib
--- a/dds/DdsDcps.mpc
+++ b/dds/DdsDcps.mpc
@@ -1,5 +1,5 @@
 project(OpenDDS_Dcps): core, coverage_optional, \
-        dcps_optional_features, dcps_optional_safety, dds_macros, \
+        dds_macros, dcps_optional_features, dcps_optional_safety, \
         dds_suppress_any_support, dds_taolib, gen_ostream, install, valgrind, \
         dds_vc_warnings, dds_versioning_idl_defaults, msvc_bigobj, dcps_optional_rapidjson {
   sharedname   = OpenDDS_Dcps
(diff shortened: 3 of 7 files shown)


mitza-oci:
Should TAO3 be changed instead?  Based on the name of that PR it seems like the different order may have been unintentional.

iguessthislldo:
I agree the different different order issue was probably unintentional. Without modifying the OpenDDS base projects order, these are the alternative fixes I can think of:
1. Change gnuace versioning to be a private detail of the libraries, so it's impossible to bleed across projects like this.
2. [Restore the manual includes of the rules files](https://github.com/mcorino/ACE_TAO/commit/24404e88a3eeb4fcd04c0bee215fba6e6978322c#diff-c01f7a1b77887d25ac9fd79c5ca27e34c57dc138a34d0573deb93656ec4ce2fe), but I guessing they removed them because it's preferable not to have child-project specific stuff in a parent project.
3. Add one or more doc-group specific markers after [the existing one](https://github.com/mcorino/ACE_TAO/blob/74d46d603c2f34cbd675287f40ceaff6a21007fa/ACE/bin/MakeProjectCreator/templates/gnu.mpd#L202). This is just the previous one in a different form though.
4. Extend MPC's `verbatim` to be able to specify that a block of text must be last if there are multiple blocks going to the same marker. Or have some sort of number-based precedence like some `*.d` configuration directories have on Unix systems, where TAO would be lower than ours.
5. Taking the previous point further, assuming it has the power and information, make gnuace smart enough to figure out the precedence of the rules files on its own. If it doesn't have the information we might have to have a way to tell it which rule files a rule file being added must precede.

If I could I'd have 1 and 5 magically already done.

mitza-oci:
I was thinking of something like option 3 where another `<%marker(something)%>` should appear in `gnu.mpd`.

iguessthislldo:
Closing because I merged https://github.com/DOCGroup/ACE_TAO/pull/1470.

OpenDDS not using ACE6/TAO2 will need the next ACE7/TAO3 release to produce correct version numbers.

jwillemsen:
Wouldn't it be good to merge this one so that the OpenDDS base projects are the first adding the extension? Maybe it helps users that are combining their own versioned macros with OpenDDS/TAO?

iguessthislldo:
I don't really like it because normally it doesn't matter what order the base projects appear in. However if we wanted to cooperate with the system in ACE7, then yes we should.

jwillemsen:
Can you then merge this also @iguessthislldo

iguessthislldo:
I don't know. I'm wary of breaking other things if the order suddenly matters. I will leave this up to @mitza-oci if I should reopen this PR.

mitza-oci:
@jwillemsen I'm not sure what problem this would be solving, assuming the user has a current ACE_TAO from either the ACE6/TAO2 or ACE7/TAO3 branch.
My recommendation would be to set the version of a downstream project in MPC so that it can work across all project types.  gnuace handles it here:
```
<%if(version)%>
GNUACE_PROJECT_VERSION = <%version%>
<%else%>
GNUACE_PROJECT_VERSION ?= $(ACE_VERSION)
<%endif%>
```

PR #2563, opened 2021-04-08
Use Complete Version for `OPENDDS_VERSION` Macro

iguessthislldo:
- Use complete version string for `OPENDDS_VERSION` macro as proposed in https://github.com/objectcomputing/OpenDDS/pull/2515. Adjust and simplify code that tried to workaround the incomplete version string.

- Allow CMake package to differentiate between micro releases. This allows a `find_package` to require at least a certain micro release.


dds/DCPS/Service_Participant.cpp
```diff
@@ -428,15 +428,10 @@ Service_Participant::get_domain_participant_factory(int &argc,
...
-        String version_string = OPENDDS_VERSION;
-        const char* const version_metadata = "-" OPENDDS_VERSION_METADATA;
-        if (version_metadata[1] != '\0') {
-          version_string += version_metadata;
-        }
         ACE_DEBUG((LM_NOTICE,
                    ACE_TEXT("(%P|%t) NOTICE: Service_Participant::get_domain_participant_factory - ")
                    ACE_TEXT("This is OpenDDS %C using ACE %C.\n"),
-                   version_string.c_str(), ACE_VERSION));
+                   OPENDDS_VERSION, ACE_VERSION));
```
mitza-oci, line 434: Do we want to normalize these to the approach that `be_init.cpp` uses where strings that are known at compile time (and known not have special characters) are directly part of the format string and not a separate parameter?

mitza-oci (approved):

PR #2569, opened 2021-04-08
Github Actions: Names, Moving tests, Caching, old CI removal

ClaytonCalabrese:
This PR cleans up the names that had previously been used for caching.

It also moves the running of tests to the test jobs so that it does not mark builds as failures.

It also removes travis and azure ci from our directory.


simpsont-oci:
For the modeling test failures in test_u20_p1_j8_FM-1f, I think all the SEGV's are a sign that the modeling tests were not built with the right features. My guess is that we need to add the ipv6 feature to the modeling test setup step in the build job.

simpsont-oci:
Otherwise, looks pretty good.

simpsont-oci:
Did not mean to close, just cancel a comment. Sorry!

simpsont-oci:
And now that I think about it, do we expect to have a fix for the cmake test issue any time soon? If we're not sure how long it will take / what's causing it, we should move them to entirely separate test jobs so when they fail they don't interrupt the autobuild test artifacts from getting uploaded. (or figure out how to include them in the autobuild testing).

ClaytonCalabrese:
> And now that I think about it, do we expect to have a fix for the cmake test issue any time soon? If we're not sure how long it will take / what's causing it, we should move them to entirely separate test jobs so when they fail they don't interrupt the autobuild test artifacts from getting uploaded. (or figure out how to include them in the autobuild testing).

I thought about that and had moved the cmake tests below the creating of the autobuild artifact. I'll do another sweep to see if I missed any.

iguessthislldo:
> (or figure out how to include them in the autobuild testing).

That is on the way.

simpsont-oci:
> I thought about that and had moved the cmake tests below the creating of the autobuild artifact.
I don't think that solves the problem, since when the job fails because of the cmake failures the artifacts are abandoned.

simpsont-oci (approved):
Looks good. I might still rename the dev-guide-only test jobs to "devguide_XXX" rather than "test_" but since those jobs don't seem to have the autobuild tests in them, I don't think they're blocking us from seeing any test results when they fail.

PR #2571, opened 2021-04-09
Adjust SPDP Reset Counting Logic & Local Announcement Timings

simpsont-oci:


jrw972 (approved):

mitza-oci:
A few tests failed with this error.  Is it caused by the change in this PR?
```
      subscriber: /home/jenkins/workspace/OpenDDS_Pull_Requests/OpenDDS/dds/DCPS/RcHandle_T.h:112: T* OpenDDS::DCPS::RcHandle<T>::operator->() const [with T = OpenDDS::DCPS::PmfSporadicTask<OpenDDS::RTPS::Spdp::SpdpTransport>]: Assertion `ptr_' failed.
```

simpsont-oci:
Shouldn't be, local_sender_ isn't the SporadicTask. I can take a look later.

simpsont-oci:
Looking into this a bit more, I think this is happening because there are a few places where we have RcHandle<SpdpSporadic> and we're not checking if they're valid before calling into them. I can add guards in this PR or we can resolve it in a separate PR.

PR #2572, opened 2021-04-09
Fix BIT Subscriber Leak In Security Attributes Test

simpsont-oci:


tests/security/attributes/DataReaderListener.cpp
```diff
@@ -54,7 +54,7 @@ void DataReaderListenerImpl::on_data_available(DDS::DataReader_ptr reader)
 #else
     if (reader == part_reader_) {
       DDS::ParticipantBuiltinTopicDataDataReader_var part_reader_impl =
-        DDS::ParticipantBuiltinTopicDataDataReader::_narrow(part_reader_);
+        DDS::ParticipantBuiltinTopicDataDataReader::_narrow(reader);
```
mitza-oci, line 57: Does this change anything? Line 55 says they're equal.  Not that I'm objecting to it.

simpsont-oci: I think just consistency with the narrow below.  Initially I was worried we were inadvertently increasing reference counts on dead objects, but that's not what was going on in this particular issue... it was not capturing / releasing the duplicated Subscriber_ptr from get_builtin_subsriber().

mitza-oci (approved):

PR #2575, opened 2021-04-12
Security attributes test: fixed Perl

mitza-oci:
This scalar needs to be a string, not a reference to a string


iguessthislldo (approved):
I guess I was thinking of Python when I wrote that...

simpsont-oci (approved):

PR #2576, opened 2021-04-12
Better Support for Thread BIT When There are Multiple Participants

iguessthislldo:
- Rename `ThreadStatus` to `ThreadStatusManager` to better reflect what it is.
- Removes finished threads from the global `ThreadStatusManager` so they don't continue to get written in the Thread BIT if there is another participant that's already running or going to run.
- Each participant gets its own local `ThreadStatusManager` to sync with the global one so that each participant can correctly do a single dispose for each finished thread.
- Fix a `ACE_HAS_MINIMUM_BIT` that should be `DDS_HAS_MINIMUM_BIT`.

### DIFF
--- a/dds/DCPS/ReactorTask.cpp
+++ b/dds/DCPS/ReactorTask.cpp
@@ -341,11 +341,12 @@ bool ThreadStatusManager::sync_with_parent(ThreadStatusManager& parent,
         std::strcmp(ci->first.c_str(), pi->first.c_str()) : got_parent ? 1 : -1;
       if (cmp < 0) { // We're behind, this thread was removed
         finished.insert(*ci);
-        map_.erase(ci++);
+        ++ci;
       } else if (cmp > 0) { // We're ahead, this thread was added
-        map_.insert(*pi);
+        running.insert(*pi);
         ++pi;
-      } else { // Same Thread
+      } else { // Same Thread, Update
+        running.insert(*pi);
         ++ci;
         ++pi;
       }
@@ -354,7 +355,7 @@ bool ThreadStatusManager::sync_with_parent(ThreadStatusManager& parent,
     }
   }
 
-  running = map_;
+  map_ = running;
 
   return true;
 }
--- a/dds/DCPS/ReactorTask.cpp
+++ b/dds/DCPS/ReactorTask.cpp
@@ -266,7 +266,7 @@ bool ThreadStatusManager::update(const String& thread_key, ThreadStatus status)
   const SystemTimePoint now = SystemTimePoint::now();
   if (DCPS_debug_level >= 4) {
     ACE_DEBUG((LM_DEBUG, "(%P|%t) ThreadStatus::update: "
-      "update for thread \"%C\" %C @ %d \n",
+      "update for thread \"%C\" %C @ %d\n",
       thread_key.c_str(), status_to_string(status), now.value().sec()));
   }
   switch (status) {
--- a/dds/DCPS/ReactorTask.cpp
+++ b/dds/DCPS/ReactorTask.cpp
@@ -183,7 +183,7 @@ int ReactorTask::svc()
     }
     if (!thread_status_manager_->update(thread_key, ThreadStatus_Finished) &&
         DCPS_debug_level) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: ReactorTask::svc: final updated failed\n"));
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: ReactorTask::svc: final update failed\n"));
     }
   }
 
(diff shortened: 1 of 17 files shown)


dds/DCPS/transport/framework/QueueTaskBase_T.h
```diff
@@ -167,6 +167,17 @@ class QueueTaskBase : public ACE_Task_Base {
...
     }
 
+    if (update_thread_status) {
+      if (DCPS_debug_level >= 4) {
+        ACE_DEBUG((LM_DEBUG, "(%P|%t) QueueTaskBase::svc: "
+          "Updating thread status for the last time\n"));
+      }
+      if (!thread_status_manager->update(thread_key, ThreadStatus_Finished) &&
+          DCPS_debug_level) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) QueueTaskBase::svc: final updated failed\n"));
```
simpsont-oci, line 177:
```suggestion
        ACE_ERROR((LM_ERROR, "(%P|%t) QueueTaskBase::svc: final update failed\n"));
```

iguessthislldo:
You'd think after awhile I'd remember that changing `DataReaderImpl` and `DataWriterImpl` breaks `tests/DCPS/unit`.

dds/DCPS/ReactorTask.cpp
```diff
@@ -272,6 +323,31 @@ String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
...
 
+bool ThreadStatusManager::sync_with_parent(ThreadStatusManager& parent,
+  ThreadStatusManager::Map& running, ThreadStatusManager::Map& finished)
+{
+  ACE_READ_GUARD_RETURN(ACE_Thread_Mutex, g1, parent.lock_, false);
+  ACE_WRITE_GUARD_RETURN(ACE_Thread_Mutex, g2, lock_, false);
+
+  // Check for finished threads
+  for (Map::iterator i = map_.begin(); i != map_.end(); /* Do in loop to support erase */) {
+    if (parent.map_.count(i->first) == 0) {
```
simpsont-oci, line 334: I don't know how big these maps are, but since we're locking two mutexes here this might be a good candidate for walking two iterators over both maps (as sorted "lists") rather than using std::map::count in order to move from O(n*log(m)) to O(n+m).

dds/DCPS/ReactorTask.cpp
```diff
@@ -272,6 +323,31 @@ String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
...
+      finished[i->first] = i->second;
+      map_.erase(i++);
+    } else {
+      ++i;
+    }
+  }
+
+  // Update running threads
+  for (Map::iterator i = parent.map_.begin(); i != parent.map_.end(); ++i) {
+    running[i->first] = i->second;
```
simpsont-oci, line 344: Similar sort of question here.

dds/DCPS/ReactorTask.cpp
iguessthislldo: They are the threads that OpenDDS is using, so they're normally tiny, like 4. I can sorta see how I could use two iterators to do it, especially because I'm trying turn one map into the other while recording the difference. I think it'd be more complicated though to try to keep track of where the two iterators in relation to each other.

dds/DCPS/ReactorTask.cpp
```diff
@@ -272,6 +323,42 @@ String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
   return key;
 }
 
+bool ThreadStatusManager::sync_with_parent(ThreadStatusManager& parent,
+  ThreadStatusManager::Map& running, ThreadStatusManager::Map& finished)
+{
+  ACE_WRITE_GUARD_RETURN(ACE_Thread_Mutex, g1, lock_, false);
+
+  {
+    ACE_READ_GUARD_RETURN(ACE_Thread_Mutex, g2, parent.lock_, false);
```
mitza-oci, line 332: I know this is not new to this PR, but with `ACE_Thread_Mutex` there is no use in distinguishing between `READ` and `WRITE` mode locks.  I am not advocating for changing to a more complex sync object.

dds/DCPS/ReactorTask.cpp
```diff
@@ -272,6 +323,42 @@ String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
...
+
+    Map::iterator ci = map_.begin();
+    Map::iterator pi = parent.map_.begin();
+    bool got_child = ci != map_.end();
+    bool got_parent = pi != parent.map_.end();
+    while (got_child || got_parent) {
+      const int cmp = got_child && got_parent ?
+        std::strcmp(ci->first.c_str(), pi->first.c_str()) : got_parent ? 1 : -1;
```
mitza-oci, line 340: Based on CI annotation, looks like it needs an `#include`.

dds/DCPS/ReactorTask.cpp
```diff
@@ -272,6 +323,42 @@ String ThreadStatus::get_key(const char* safety_profile_tid, const String& name)
...
+{
+  ACE_WRITE_GUARD_RETURN(ACE_Thread_Mutex, g1, lock_, false);
+
+  {
+    ACE_READ_GUARD_RETURN(ACE_Thread_Mutex, g2, parent.lock_, false);
+
+    Map::iterator ci = map_.begin();
+    Map::iterator pi = parent.map_.begin();
+    bool got_child = ci != map_.end();
+    bool got_parent = pi != parent.map_.end();
```
mitza-oci, line 337: I'm not sure keeping these as separate variables is any better than just using comparisons with `end`, since the variables have to be updated at the right time.  I'm also not sure it's any worse.

tests/DCPS/InternalThreadStatus/InternalThreadStatusListenerImpl.cpp
```diff
@@ -16,8 +16,8 @@
 #include <string>
 
 // Implementation skeleton constructor
-InternalThreadStatusListenerImpl::InternalThreadStatusListenerImpl(OPENDDS_STRING id) :
-  id_(id), count_(0)
+InternalThreadStatusListenerImpl::InternalThreadStatusListenerImpl(OpenDDS::DCPS::String id) :
```
mitza-oci, line 19:
```suggestion
InternalThreadStatusListenerImpl::InternalThreadStatusListenerImpl(const OpenDDS::DCPS::String& id) :
```

mitza-oci (changes requested):

dds/DCPS/ReactorTask.cpp
iguessthislldo: Yeah, they're left over from a version of the algorithm where I was doing more with the variables. Taking them away makes the `cmp` expression a bit harder to read, but I'm fine with that.

mitza-oci (approved):

PR #2582, opened 2021-04-14
Generate Typecode identifiers needed by tao_idl when using C++11 Mapping

ClaytonCalabrese:
This PR Allows the user to attempt to generate typecodes in C++11 mapping.  These typecodes will just create empty objects to allow for compilation but not use.  There are new tests I have created located in tests/DCPS/Compiler/typecode and tests/DCPS/Compiler/typecode/C++11, which try to create an instance of the various idl types as type objects, just to confirm the dummies allow compilation.  The base /typecode  test should work by default but the C++11 one requires the changes in this PR.


configure
```diff
@@ -248,6 +248,7 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'object-model-profile!', 'Object Model Profile (yes)',
     'persistence-profile!', 'Persistence Profile (yes)',
     'safety-profile:s', 'Safety Profile: base or extended (none)',
+    'suppress-anys!', 'Suppress Anys (yes))',
```
mitza-oci, line 251:
```suggestion
    'suppress-anys!', 'Suppress Anys (yes)',
```

mitza-oci, line 251: Looks like we need to decide how this should interact with the related MPC features.  I would expect it to have an impact if we're going to make it a configure option.

On the other hand, we want it to only impact how the application's IDL is built and not impact the core, then it shouldn't be in the configure script.  In that case it would be up to the build system (both CMake and MPC) of the client application to set the option (in CMake's case that could be a cache variable or an argument to a function or ??).

mitza-oci (changes requested):

configure
ClaytonCalabrese: configure is what sets up our config.cmake file which is where we had the OPENDDS_SUPPRESS_ANYS flag being set.  This allows the user to turn that flag off.  The other alternatives would be:

1.  The user manually editing the cmake.config file and turning OPENDDS_SUPPRESS_ANYS off after their configure has been run.

2. manually removing the flags in their cmake lists:
list(REMOVE_ITEM OPENDDS_DDS_BASE_IDL_FLAGS -Sa -St)
list(REMOVE_ITEM OPENDDS_TAO_BASE_IDL_FLAGS -Sa -St)

3. Updating our macros to have a macro to strip the -Sa -St flags like in 2. This macro would then need to be used in CMakeLists.txt.

mitza-oci: Let's go back to the question what exactly this option needs to do.  In my previous comment I listed two possible approaches.  I still think that either could work, so we need understand both approaches and their pros/cons.

1. If the user specifies `--no-suppress-anys` to the configure script, the effect of this would be to change how OpenDDS core libraries are built.  Not doing so would be really surprising since that's what the configure script is for.  To make this work, add `dds_suppress_anys=0` to the default.features before configure runs MPC.  However, this wide-reaching of a change may not be needed.  In which case we consider...

2. If the core libraries should not be changed, this kind of functionality could be provided by only changing how the application's IDL is built.  In this case, `configure` shouldn't change at all (which means `config.cmake` doesn't change either).  Instead, for each build system (MPC and CMake) provide (and test) a way for the user to specify that support for Anys and TypeCodes are needed only in these projects.

At this point I'd be more in favor of approach 2 since it minimizes impact (and footprint) of the change.  If it turns out the changing how the core libs are built is indeed necessary, that could be added in a subsequent PR.

ClaytonCalabrese:
The way it is right now it will always create dummy typecodes for C++11.  I don't think this is harmful but I'm open to finding some way of blocking them if we deem it necessary.

MPC/config/dds_no_suppress_anys.mpb
```diff
@@ -0,0 +1,4 @@
+project: dcps, dcps_ts_defaults {
```
mitza-oci, line 1:
```suggestion
project: dcps_ts_defaults {
```
Does it work without the `dcps` base project?  Seems like it wouldn't be needed, also probably doesn't hurt to have it.

dds/idl/langmap_generator.cpp
```diff
@@ -1407,6 +1407,12 @@ struct Cxx11Generator : GeneratorBase
     }
     be_global->add_include("<array>", BE_GlobalData::STREAM_LANG_H);
     be_global->lang_header_ << ind << "using " << type << " = " << array << elem << bounds.str() << ";\n";
+
+    if (!be_global->suppress_typecode()) {
+      be_global->add_include("tao/Basic_Types.h", BE_GlobalData::STREAM_LANG_H);
+      be_global->lang_header_ << "extern ::CORBA::TypeCode_ptr const _tc_" << type << ";\n";
```
mitza-oci, line 1413:
```suggestion
      be_global->lang_header_ << "extern const ::CORBA::TypeCode_ptr _tc_" << type << ";\n";
```
Same on the next line.
Since the same block appears 4 times, refactor to a common function.

cmake/api_macros.cmake
```diff
@@ -65,6 +65,11 @@ macro(OPENDDS_GET_SOURCES_AND_OPTIONS
   endforeach()
 endmacro()
 
+macro(OPENDDS_NO_SUPPRESS_ANYS)
+  list(REMOVE_ITEM OPENDDS_DDS_BASE_IDL_FLAGS -Sa -St)
+  list(REMOVE_ITEM OPENDDS_TAO_BASE_IDL_FLAGS -Sa -St)
```
mitza-oci, line 70: Would this need to split the two args into two different CMake commands to handle `-St -Sa` (or if they appear with other args mixed in)?

tests/DCPS/Compiler/typecode/C++11/run_test.pl
```diff
@@ -0,0 +1,22 @@
...
+use Env (DDS_ROOT);
+use lib "$DDS_ROOT/bin";
+use Env (ACE_ROOT);
+use lib "$ACE_ROOT/bin";
+use PerlDDS::Run_Test;
+use strict;
+
+my $name = 'Compiler_Typecode_C++11';
+my $status = 0;
+my $TESTDVR = PerlDDS::create_process ($name);
```
mitza-oci, line 14: Prefer using TestFramework

tests/DCPS/Compiler/typecode/C++11/typecode.cpp
```diff
@@ -0,0 +1,24 @@
+/***
+ * This test is meant to test the annotations set by typeflags.
+ * Currently IS_AUTOID_HASH is not implemented so that should always return false.
```
mitza-oci, line 3: ?

tests/DCPS/Compiler/typecode/C++11/typecode.cpp
```diff
@@ -0,0 +1,24 @@
...
+ ***/
+#include "typecodeTypeSupportImpl.h"
+
+#include <gtest/gtest.h>
+
+using namespace OpenDDS::DCPS;
+
+TEST(TestTypecodes, can_allocate)
+{
+  my_module::_tc_my_struct;
```
mitza-oci, line 13: Use the expression in a statement (with a gtest assert, for example).

cmake/api_macros.cmake
```diff
@@ -65,6 +65,11 @@ macro(OPENDDS_GET_SOURCES_AND_OPTIONS
   endforeach()
 endmacro()
 
+macro(OPENDDS_NO_SUPPRESS_ANYS)
```
mitza-oci, line 68: Test this?

mitza-oci (changes requested):

cmake/api_macros.cmake
ClaytonCalabrese: I think it handles the values as different entities.

The documentation says: list(REMOVE_ITEM <list> <value> [<value> ...])

Which indicates each value is probably actually done as a seperate call of remove item.

tests/cmake_integration/C++11_typecode/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+project(OpenDDS_C++11_Typecode CXX)
+cmake_minimum_required(VERSION 3.8.2)
+
+if(NOT OpenDDS_FOUND)
+  find_package(OpenDDS REQUIRED)
+endif()
+
+if(OPENDDS_NO_OPENDDS_SAFETY_PROFILE EQUAL OFF)
```
iguessthislldo, line 8:
```suggestion
if(NOT OPENDDS_NO_OPENDDS_SAFETY_PROFILE)
```

tests/cmake_integration/CMakeLists.txt
```diff
@@ -9,3 +9,6 @@ add_subdirectory(Nested_IDL)
 if(NOT OPENDDS_STATIC)
   add_subdirectory(generated_global)
 endif()
+if(OPENDDS_NO_OPENDDS_SAFETY_PROFILE EQUAL NOTFOUND OR OPENDDS_NO_OPENDDS_SAFETY_PROFILE EQUAL ON)
```
iguessthislldo, line 12: `NOTFOUND` and `ON` are both boolean literals (out of many) in CMake. `NOTFOUND` is false and `ON` is true, so I'm a little confused by what you intended, but going the test file, I'm assuming:

```suggestion
if(OPENDDS_NO_OPENDDS_SAFETY_PROFILE)
```

iguessthislldo (changes requested):

dds/idl/langmap_generator.cpp
```diff
@@ -1370,6 +1370,14 @@ struct Cxx11Generator : GeneratorBase
     helpers_[HLP_FIXED_CONSTANT] = "IDL::Fixed_T";
   }
 
+  static void gen_cxx11_typecodes(const std::string& type) {
```
mitza-oci, line 1373:
```suggestion
  static void gen_typecode_ptrs(const std::string& type)
  {
```
This is in a class called `Cxx11Generator`

dds/idl/langmap_generator.cpp
```diff
@@ -1370,6 +1370,14 @@ struct Cxx11Generator : GeneratorBase
...
+    if (!be_global->suppress_typecode()) {
+      be_global->add_include("tao/Basic_Types.h", BE_GlobalData::STREAM_LANG_H);
+      be_global->lang_header_ << "extern ::CORBA::TypeCode_ptr const _tc_" << type << ";\n";
+      be_global->impl_ << "::CORBA::TypeCode_ptr const _tc_" << type << " = 0;\n";
```
mitza-oci, line 1377:
```suggestion
      be_global->lang_header_ << "extern const ::CORBA::TypeCode_ptr _tc_" << type << ";\n";
      be_global->impl_ << "const ::CORBA::TypeCode_ptr _tc_" << type << " = nullptr;\n";
```

tests/DCPS/Compiler/typecode/C++11/typecode.cpp
```diff
@@ -0,0 +1,20 @@
+#include "typecodeTypeSupportImpl.h"
+
+#include <gtest/gtest.h>
+
+using namespace OpenDDS::DCPS;
+
+TEST(TestTypecodes, can_allocate)
+{
+  ASSERT_EQ(my_module::_tc_my_struct, my_module::_tc_my_struct);
```
mitza-oci, line 9:
```suggestion
  ASSERT_EQ(my_module::_tc_my_struct, nullptr);
```

tests/cmake_integration/C++11_typecode/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+endif()
+
+set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
+OPENDDS_NO_SUPPRESS_ANYS()
+set(target_prefix "opendds_cmake_cpp11_typecode_")
+set(src "${CMAKE_CURRENT_SOURCE_DIR}")
+set(dst ${CMAKE_CURRENT_BINARY_DIR})
+set(opendds_libs
+  OpenDDS::Dcps
+)
```
mitza-oci, line 19: Is this used?

tests/cmake_integration/C++11_typecode/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+if(NOT OpenDDS_FOUND)
+  find_package(OpenDDS REQUIRED)
+endif()
+
+if(OPENDDS_NO_OPENDDS_SAFETY_PROFILE EQUAL OFF)
+  message(FATAL_ERROR "This test cannot be built with Safety Profile enabled")
+endif()
+
+set(CMAKE_CXX_COMPILER ${OPENDDS_COMPILER})
+OPENDDS_NO_SUPPRESS_ANYS()
```
mitza-oci, line 13: What's the scope of the change here?  It seems like this belongs in `OPENDDS_TARGET_SOURCES`

tests/cmake_integration/C++11_typecode/typecode.idl
```diff
@@ -0,0 +1,10 @@
+#ifndef __TYPECODE_IDL__
```
mitza-oci, line 1: Don't use reserved identifiers (double underscore)

tests/cmake_integration/run_ci_tests.pl
```diff
@@ -47,12 +48,14 @@ ($)
...
     "no-shared" => \$no_shared,
     );
-
-my @dirs = ('../Nested_IDL', 'Messenger_1', 'Messenger_2');
-push @dirs, '../generated_global' unless $no_shared;
-push @dirs, 'C++11_Messenger' unless $skip_cxx11;
+my @dirs;
+#my @dirs = ('../Nested_IDL', 'Messenger_1', 'Messenger_2');
+#push @dirs, '../generated_global' unless $no_shared;
+#push @dirs, 'C++11_Messenger' unless $skip_cxx11;
```
mitza-oci, line 57: Why is this changing?

mitza-oci (changes requested):

tests/cmake_integration/run_ci_tests.pl
ClaytonCalabrese: Sorry, forgot to clean that up from testing.

tests/cmake_integration/C++11_typecode/CMakeLists.txt
iguessthislldo: I agree. It's changing IDL flags. It should either be a global cache variable, a argument to `OPENDDS_TARGET_SOURCES`, or both.

ClaytonCalabrese: ok

mitza-oci (approved):

tests/DCPS/Compiler/typecode/C++11/.gitignore
```diff
@@ -0,0 +1,14 @@
+Compiler_Typecode_C++11
+typecodeC.cpp
+typecodeC.h
+typecodeC.inl
+typecodeS.cpp
+typecodeS.h
```
iguessthislldo, line 6:
```suggestion
typecodeC.h
```

I don't think these files get generated for the C++11 mapping.

iguessthislldo (changes requested):

iguessthislldo:
Other than those minor things I think this PR is ready.

iguessthislldo (approved):

PR #2585, opened 2021-04-15
Removed unused third-party services from README.md

mitza-oci:


jwillemsen:
Coday is still enabled as far as I can tell

mitza-oci:
> 
> 
> Codacy is still enabled as far as I can tell

It doesn't result in "checks" appearing in the pull requests.  So we don't know if each change to the repository is impacting whatever it is that Codacy is doing.

jwillemsen:
Maybe that is a permission issue in the project settings?

ClaytonCalabrese (approved):

mitza-oci:
> 
> 
> Maybe that is a permission issue in the project settings?

Is it working in ACE_TAO or some other repository we can use as an example?

PR #2588, opened 2021-04-16
Add Mutex Protection For DDS Entity Listeners

simpsont-oci:
Problem:
Setting Listeners, especially setting them to NULL, is not thread safe.

Solution:
Add mutex protection around listener objects.

### DIFF
--- a/dds/DCPS/DataReaderImpl.h
+++ b/dds/DCPS/DataReaderImpl.h
@@ -585,6 +585,10 @@ public:
   void return_handle(DDS::InstanceHandle_t handle);
 
 protected:
+
+  // Perform cast to get extended version of listener (otherwise nil)
+  virtual DataReaderListener_ptr get_ext_listener();
+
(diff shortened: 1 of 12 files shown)


dds/DCPS/DataReaderImpl.h
```diff
@@ -585,6 +585,10 @@ class OpenDDS_Dcps_Export DataReaderImpl
   void return_handle(DDS::InstanceHandle_t handle);
 
 protected:
+
+  // Perform cast to get extended version of listener (otherwise nil)
+  virtual DataReaderListener_ptr get_ext_listener();
```
mitza-oci, line 590: Does it need `virtual`?

mitza-oci (changes requested):

dds/DCPS/DataReaderImpl.h
simpsont-oci: Ah, no, just copied from get_listener(). I'll remove them.

mitza-oci (approved):

PR #2592, opened 2021-04-20
RtpsRelay becomes non-responsive when mass disconnect

jrw972:
Problem
-------

If all of the clients of the RtpsRelay go down at the same time and
they have the same lease duration, then they will timeout in the relay
at the same time.  Given enough clients, the number of unregister
dispose messages floods the network at make the relay instance
unresponsive.

Solution
--------

Distribute the unregisters in time by queuing them and processing them
with a periodic task.  The proposed solution can be improved in at
least three ways:

1. Eliminate redundant code.

2. Use a sporadic task.

3. Make the period configurable and/or come up with a better heuristic.

### DIFF
--- a/tools/rtpsrelay/ParticipantListener.h
+++ b/tools/rtpsrelay/ParticipantListener.h
@@ -5,6 +5,7 @@
 #include "DomainStatisticsReporter.h"
 
 #include <dds/DCPS/DomainParticipantImpl.h>
+#include <dds/DCPS/PeriodicTask.h>
 
 namespace RtpsRelay {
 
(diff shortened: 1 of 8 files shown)


tools/rtpsrelay/ParticipantListener.h
```diff
@@ -14,16 +15,34 @@ class ParticipantListener : public ListenerBase {
...
+  };
 
   const Config& config_;
   OpenDDS::DCPS::DomainParticipantImpl* participant_;
   DomainStatisticsReporter& stats_reporter_;
   ParticipantEntryDataWriter_var writer_;
+  OpenDDS::DCPS::RcHandle<Unregister> unregister_;
+  ACE_Thread_Mutex mutex_;
+  std::list<OpenDDS::DCPS::GUID_t> unregister_queue_;
+  // mutex_
```
mitza-oci, line 45:
```suggestion
```

PR #2598, opened 2021-04-22
RtpsRelay is not scalable

jrw972:
Problem
-------

The existing approach to forwarding messages in the relay relies on
having a complete list of all the readers and writers using the relay,
their associations, and a record indicating which relay is servicing
the associated participant.  This is a large amount of durable data.
Adding a new relay instance means that it must receive all of this
data from its peers.  This places extra load on the existing relay
instances to the point that they start thrashing.

Solution
--------

Reduce the amount of data that must be shared among the relays.  The
new topics are:

* Relay Partitions - Each relay publishes the set of partitions
  corresponding to its clients.

* Relay Partitions Increment - Each relay publishes an additional set
  of partitions for new clients.  The partitions plus the increment
  forms the set of partitions that the relay is servicing.  The
  motivation behind the increment is that it is a smaller and faster
  message which allows all of the relays to be more responsive to new
  clients.  This approach errs on the side of forwarding too many
  messages instead of not sending enough.

* Relay Addresses - Each relay publishes its horizontal addresses.

* SPDP Replay - Each relay publishes requests for replaying SPDP
  messages for new client as they appear.  Currently, this topic is
  reliable but it may make sense to change it to best effort.

Notes
-----

The capability for matching based on topic has been removed.  It could
be restored in one of two ways.  First, instead of publishing
partitions, the relay would publish (partition, topic) pairs for
roughly P x T entries.  Second, the topics could be published
independently for roughly P + T entries and a message is forwarded to
a relay if it service at least one source partition and at least one
source topic.  However, most relays will most likely have all of the
topics so the topic would not actually filter any sends.


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -254,55 +257,57 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+  DDS::Topic_var relay_partitions_increment_topic =
+    relay_participant->create_topic(RELAY_PARTITIONS_INCREMENT_TOPIC_NAME.c_str(),
+                                    relay_partitions_increment_type_name,
+                                    TOPIC_QOS_DEFAULT, nullptr,
+                                    OpenDDS::DCPS::DEFAULT_STATUS_MASK);
 
-  if (!writers_topic) {
-    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: failed to create Writers topic\n")));
+  if (!relay_partitions_increment_topic) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: failed to create Readers topic\n")));
```
mitza-oci, line 292: error doesn't match topic name

tools/rtpsrelay/utility.h
```diff
@@ -89,6 +93,46 @@ inline bool operator<(const Duration_t& x, const Duration_t& y)
   return x.nanosec() < y.nanosec();
 }
 
+  inline OpenDDS::DCPS::RepoId guid_to_repoid(const GUID_t& a_guid)
```
mitza-oci, line 96:
```suggestion
inline OpenDDS::DCPS::RepoId guid_to_repoid(const GUID_t& a_guid)
```

tools/rtpsrelay/GuidPartitionTable.h
```diff
@@ -0,0 +1,212 @@
...
+  RelayPartitionsDataWriter_var relay_partitions_writer_;
+  RelayPartitions relay_partitions_;
+  RelayPartitionsIncrementDataWriter_var relay_partitions_increment_writer_;
+  RelayPartitionsIncrement relay_partitions_increment_;
+  SpdpReplayDataWriter_var spdp_replay_writer_;
+
+  typedef std::map<OpenDDS::DCPS::RepoId, StringSet, OpenDDS::DCPS::GUID_tKeyLessThan> GuidToPartitions;
+  GuidToPartitions guid_to_partitions_;
+
+  typedef std::set<OpenDDS::DCPS::GUID_t, OpenDDS::DCPS::GUID_tKeyLessThan> OGuidSet;
```
mitza-oci, line 203:
```suggestion
  typedef std::set<OpenDDS::DCPS::GUID_t, OpenDDS::DCPS::GUID_tKeyLessThan> OrderedGuidSet;
```
??

mitza-oci (changes requested):

mitza-oci (approved):

PR #2600, opened 2021-04-22
Github Actions: Add ace7tao3 build to GHA

ClaytonCalabrese:
Someday we want to be on ace7tao3 and we want a GHA build to check on our compatibility over time. This PR adds that to the u20_j_qt_ws_sec build.  There are multiple options for ACE7TAO3: major, minor, and micro.  I settled on minor but am open to changing.

Update 4/22/21: Changed the branch to be master instead of any of the 3 aforementioned branches.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -2103,7 +2103,6 @@ jobs:
       uses: actions/checkout@v2.3.4
       with:
         repository: DOCGroup/ACE_TAO
-        ref: master
         path: OpenDDS/ACE_TAO
     - name: get ACE_TAO commit
       shell: bash
@@ -2178,7 +2177,6 @@ jobs:
       uses: actions/checkout@v2.3.4
       with:
         repository: DOCGroup/ACE_TAO
-        ref: master
         path: ACE_TAO
     - name: download ACE_TAO artifact
       uses: actions/download-artifact@v2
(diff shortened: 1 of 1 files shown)


jwillemsen:
I think micro would be the best option, a minor release is each 1.5 to 2 years, major took us 10 years ;-)

.github/workflows/build_and_test.yml
```diff
@@ -2103,7 +2103,7 @@ jobs:
       uses: actions/checkout@v2.3.4
       with:
         repository: DOCGroup/ACE_TAO
-        ref: ace6tao2
+        ref: Latest_ACE7TAO3_Minor
```
mitza-oci, line 2106: This should track master, same concept as ace6tao2.

mitza-oci (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -2103,7 +2103,7 @@ jobs:
       uses: actions/checkout@v2.3.4
       with:
         repository: DOCGroup/ACE_TAO
-        ref: ace6tao2
+        ref: master
```
mitza-oci, line 2106:
```suggestion
```

I assume it just uses the default branch?

mitza-oci (changes requested):

iguessthislldo (approved):

mitza-oci (approved):

PR #2605, opened 2021-04-23
Properly Detect XCDR1 in Combination with Appendable or Mutable

ClaytonCalabrese:
We were incorrectly detecting if XCDR1 was really being used or was just in the list of the QoS. This PR changes to only use the writer QoS which should only have 1 item in it.


dds/DCPS/DiscoveryBase.h
```diff
@@ -1522,22 +1522,16 @@ namespace OpenDDS {
           const XTypes::TypeIdentifier& writer_type_id = writer_type_info->minimal.typeid_with_size.type_id;
           const XTypes::TypeIdentifier& reader_type_id = reader_type_info->minimal.typeid_with_size.type_id;
           if (writer_type_id.kind() != XTypes::TK_NONE && reader_type_id.kind() != XTypes::TK_NONE) {
-            if (!writer_local || !reader_local) {
```
iguessthislldo, line 1525: I think this check is still important because if it's a local association it doesn't matter if it's XCDR1 appendable.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #2606, opened 2021-04-24
Attempt to Choose a Single Coherent Address

simpsont-oci:
Specifically, when the resolve order is unpredictable, choose one sensible address.

### DIFF
--- a/dds/DCPS/transport/framework/NetworkAddress.cpp
+++ b/dds/DCPS/transport/framework/NetworkAddress.cpp
@@ -496,24 +496,24 @@ ACE_INET_Addr choose_single_coherent_address(const ACE_INET_Addr& address, bool 
 #ifdef ACE_HAS_IPV6
     if (temp.get_type() == AF_INET6 && !temp.is_multicast()) {
       if (temp.is_loopback()) {
-        ACE_DEBUG((LM_DEBUG, "(%P|%t) NetworkAddress::choose_single_coherent_address() - "
+        VDBG((LM_DEBUG, "(%P|%t) NetworkAddress::choose_single_coherent_address() - "
                    "Considering Address %C:%d (%C) - ADDING TO IPv6 LOOPBACK LIST\n",
                    temp.get_host_addr(), temp.get_port_number(), temp.get_host_name()));
         set6_loopback.insert(temp);
       } else if (temp.is_ipv4_mapped_ipv6() || temp.is_ipv4_compat_ipv6()) {
 #ifndef IPV6_V6ONLY
-        ACE_DEBUG((LM_DEBUG, "(%P|%t) NetworkAddress::choose_single_coherent_address() - "
+        VDBG((LM_DEBUG, "(%P|%t) NetworkAddress::choose_single_coherent_address() - "
                    "Considering Address %C:%d (%C) - ADDING TO IPv6 MAPPED / COMPATIBLE IPv4 LIST\n",
                    temp.get_host_addr(), temp.get_port_number(), temp.get_host_name()));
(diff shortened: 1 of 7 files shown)


dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -473,6 +473,87 @@ bool open_appropriate_socket_type(ACE_SOCK_Dgram& socket, const ACE_INET_Addr& l
...
+      loop4 = *it;
+      set4.erase(it++);
+    } else {
+      ++it;
+    }
+  }
+
+#ifdef ACE_HAS_IPV6
+  if (loop6 != ACE_INET_Addr()) {
+    return loop6;
```
mitza-oci, line 536: Why is loop preferred over non-loop?

simpsont-oci: If one of the resolved addresses for the hostname in question is loopback, we're asking about ourselves, so just use loopback.

I am not super tied to any of the logic in this method other than sorting, I was just trying to pick something somewhat sensible to start with.

In practice, I am not sure INET_Addr is even this sophisticated. I never saw a mix of IPv4 and IPv6 in my testing, but I saw a IPv4 mapped IPv6 address that didn't seem to get flagged appropriately by that API. So in reality, maybe everything other than sorting can go away?

mitza-oci: The loopback aspect that concerns me is that it's used for `TcpInst::local_address_` which seems like that could impact what the socket binds to.

simpsont-oci: I originally thought about adding an "allow_loopback" flag to the arguments for use with tcp & udp transports' local addresses, but in testing it never seemed to be a problem. I think this is partly because it's rare for hosts to resolve any hostname (including their own) to the loopback address, so I left it out. I can either add the flag and apply it for local addresses, or I can remove loopback addresses entirely.

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -473,6 +473,101 @@ bool open_appropriate_socket_type(ACE_SOCK_Dgram& socket, const ACE_INET_Addr& l
   return socket.open(local_address) == 0;
 #endif
 }
+
+ACE_INET_Addr choose_single_coherent_address(const ACE_INET_Addr& address, bool prefer_loopback)
+{
+// Check ACE_INET_Addr does supports next()
```
mitza-oci, line 479:
```suggestion
// Check that ACE_INET_Addr supports next()
```

dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -473,6 +473,101 @@ bool open_appropriate_socket_type(ACE_SOCK_Dgram& socket, const ACE_INET_Addr& l
...
+  }
+#endif
+
+  if (!set4_loopback.empty()) {
+    return *set4_loopback.begin();
+  }
+
+#endif // !(ACE_MAJOR_VERSION < 6 || (ACE_MAJOR_VERSION == 6 && (ACE_MINOR_VERSION < 3 || (ACE_MINOR_VERSION == 3 && ACE_MICRO_VERSION < 1))))
+  return address;
+}
```
mitza-oci, line 569: Add ACE_UNUSED_ARG for the second argument

mitza-oci (changes requested):

mitza-oci (approved):

Issue #2607, opened 2021-04-24
Module name too close to IDL keyword ?


neoliant:
Hello,

With default OpenDDS-3.15 and 3.16 installation, when I try to compile the following IDL file :

```
module ATTRIBUTE
{
  @topic
  struct TopicA {
    string a;
  };
}; // module ATTRIBUTE
```

with the following commands :
```
tao_idl --idl-version 4 -I/usr/local/include --unknown-annotations ignore -Cw test.idl
opendds_idl -Cw test.idl
tao_idl --idl-version 4 -I/usr/local/include --unknown-annotations ignore -Cw testTypeSupport.idl
```

I have the following error with the last command
Error - tao_idl: "testTypeSupport.idl", line 38: error in lookup of symbol: ATTRIBUTE::TopicA

The issue doesn't appear with a different module name. I understand this is closely related to the IDL keyword, but with a different case. 

Am I doing something silly here? Any hint? Is this a bug? Help and thanks in advance ;-)

mitza-oci:
The IDL language spec doesn't allow identifiers that collide with keywords case-insensitively.

neoliant:
So what is the purpose of providing a flag that changes it from error to warning ? 

The OMG Specification for the OPC UA - DDS Gateway comes with a normative IDL file (dds-opcua_services.idl) that contains such a module name. Who do you think should be contacted in that case ?

jwillemsen:
You should report a formal issue to the OMG at https://issues.omg.org/issues/create-new-issue

neoliant:
I understand and I will do it.

Meanwhile, I tried using "module _ATTRIBUTE {" and can pass the opendds_idl and tao_idl code generation steps. But latter on it doesn't compile.

test.idl :

module Foo {
module _ATTRIBUTE
{
  @topic
  struct TopicA {
    string a;
  };
}; // module ATTRIBUTE
}; // module Foo

main.cpp :
#include "testTypeSupportImpl.h"

int main (int argc, char *argv[])
{
   Foo::ATTRIBUTE::TopicA a;

   return 0;
}

CMakeLists.txt :

cmake_minimum_required(VERSION 3.9.4)
project(testATTRIBUTE CXX)
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(OPENDDS_CMAKE_VERBOSE ON)
set(CMAKE_PREFIX_PATH "/usr/local")

[...]

PR #2608, opened 2021-04-25
Correctly Process Escaped Identifiers in `opendds_idl`

neoliant:
As described in issue #2607. This test is presently failing to compile.


iguessthislldo:
Updated the PR with the fix, so I also changed the title. Summery is:

Change `string dds_generator::to_string(Identifier*)` to add or remove the leading underscore on the id based on the context. Update functions that call `to_string` when the context requires it.

IDL keywords can now be used as identifiers. C++ ones also should generally, except for topic type names because `DCPS_DATA_SEQUENCE_TYPE` doesn't handle this properly in `tad_idl`.

tests/DCPS/Compiler/keywords/cpp11/main.cpp
```diff
@@ -0,0 +1,32 @@
...
+  b.component() = boolean::oneway;
+  ACE_UNUSED_ARG(b);
+  boolean::attributeDataWriter* b_dw = 0;
+  ACE_UNUSED_ARG(b_dw);
+
+  /* TODO(iguessthislldo): See test.idl
+  _cxx_bool::_cxx_class c;
+  c._cxx_else() = _cxx_bool::_cxx_continue;
+  ACE_UNUSED_ARG(c);
+  _cxx_bool::cxx_classDataWriter* c_dw = 0;
```
jwillemsen, line 21: This is not as I would expect it, first the implied IDL rules from the DDS specification should be applied, after that the language mapping escaping on the implied IDL, so it should be `_cxx_bool::classDataWriter`.

iguessthislldo: This was just result of applying the existing code to this, but this sounds right. I could change `EscapeContext_ForGenIdl` to strip `_cxx_` and I think it will happen that way.

iguessthislldo: Alright done and I was able to test using the ACE/TAO PR.

tests/DCPS/Compiler/keywords/classic/keywords_classic.mpc
```diff
@@ -0,0 +1,15 @@
+project: dcps_test {
+  exename = keywords-classic
+  includes += .
```
mitza-oci, line 3: Is this needed?

tests/DCPS/Compiler/keywords/test.idl
```diff
@@ -0,0 +1,39 @@
+// This test was made to catch problems with IDL keywords
+// used as module name and escaped with an appended '_'
```
mitza-oci, line 2:
```suggestion
// used as module name and escaped with a prepended '_'
```

mitza-oci (changes requested):

tests/DCPS/Compiler/keywords/classic/keywords_classic.mpc
iguessthislldo: Yes, [it's used in Multitopic](https://github.com/objectcomputing/OpenDDS/blob/13bc1f9bade4d4b8f9d78ee7eb4311c24867bcdf/tests/DCPS/MultiTopic/classic/MultiTopicTest_Classic.mpc#L11), which is the dual mapping archetype. It adds the generated headers to the include path for `main.cpp` because they're in different directories.

mitza-oci (approved):

ClaytonCalabrese (changes requested):
Looks like the u18_o1d0v1_xer0_j12 build is failing because of this pr.  Possibly due to the _cxx not being stripped properly?

iguessthislldo:
Yes, it's expected, as I mentioned this is [this commit message](https://github.com/objectcomputing/OpenDDS/pull/2608/commits/16a6ac7266437132c4a0212b8901970446806845). These changes require `tao_idl` to have the changes I have linked above for that test to compile correctly.

ClaytonCalabrese (approved):

tests/DCPS/Compiler/keywords/test.idl
```diff
@@ -0,0 +1,39 @@
...
+};
+
+// Identifiers that are both IDL and C++ keywords.
+module _case {
+  @topic
+  struct _struct {
+    @key long _union;
+  };
+  const long _typeid = 0;
+};
```
jwillemsen, line 39:
``` suggestion
};

module Escaping
{
  union _private;
  @topic
  union _private switch (long)
  {
    case 2: short b_03_3;
  };
  struct _public;
  @topic
  struct _public
  {
    string bar;
  };
};
```

jwillemsen:
Added some more example IDL which failed to compile on master, maybe merge this into the PR @neoliant?

jwillemsen:
New ACE/TAO releases are available, these contain the required fix in tao_idl

neoliant:
Thank you so much. I'll give it a try very soon. Thanks again and keep up with this beautiful work ;-)

neoliant:
Alright, I tried this morning and had the following result : the example "module _ATTRIBUTE" idl generates correctly a "ATTRIBUTE" namespace in C++, but in the xxxTypeSupportImpl.cpp when using a member of this namespace in functions (like set_default, operator<<, operator>>, ...) parameters or template names, there's still "_ATTRIBUTE" instead of "ATTRIBUTE" which leads to subsequent compilation failure.

jwillemsen:
Yes, just found #2667 also here also

jwillemsen:
Created #2668 which at least got my code to compile, haven't seen a case where `_ATTRIBUTE` was still used

neoliant:
It's compiling now. Thanks a lot to the 3 of you !

PR #2627, opened 2021-04-29
Typecode and C++11 mapping fixes

mitza-oci:


ClaytonCalabrese (approved):

ClaytonCalabrese:
There needs to be changes to the .cpp for the test as well. It is trying to use the typecodes removed from the idl.

ClaytonCalabrese (changes requested):
The test needs to have its cpp change to not longer reference the removed types.

ClaytonCalabrese (approved):

PR #2630, opened 2021-04-29
Typecode Test Follow Up

ClaytonCalabrese:
The previous PR for typecodes that I made had some flaws that were partially fixed with Fred's PR from yesterday: 
https://github.com/objectcomputing/OpenDDS/pull/2624

and Adam's from today:
https://github.com/objectcomputing/OpenDDS/pull/2627

The main fixes have to do with cleaning up the tests as well as changing how OPENDDS_TARGET_SOURCES decides whether to suppress typecodes.


cmake/api_macros.cmake
```diff
@@ -206,7 +210,7 @@ macro(OPENDDS_TARGET_SOURCES target)
     message(FATAL_ERROR "OpenDDS does not support argument items in CMAKE_CXX_COMPILER.")
   endif()
 
-  if(OPENDDS_SUPPRESS_ANYS)
+  if("${_suppress_typecodes}" MATCHES ON OR (NOT "${_suppress_typecodes}" MATCHES OFF AND OPENDDS_SUPPRESS_ANYS MATCHES ON))
```
iguessthislldo, line 213: I suggest having

```cmake
set(_suppress_typecodes ${OPENDDS_SUPPRESS_ANYS})
```
before `OPENDDS_GET_SOURCES_AND_OPTIONS` and then here:

```suggestion
  if(_suppress_typecodes)
```

iguessthislldo (changes requested):

cmake/api_macros.cmake
```diff
@@ -8,11 +8,13 @@ macro(OPENDDS_GET_SOURCES_AND_OPTIONS
...
   tao_options
-  opendds_options)
+  opendds_options
+  suppress_typecodes)
 
   set(_options_n
     PUBLIC PRIVATE INTERFACE
-    TAO_IDL_OPTIONS OPENDDS_IDL_OPTIONS)
+    TAO_IDL_OPTIONS OPENDDS_IDL_OPTIONS
+    OPENDDS_SUPPRESS_TYPECODES)
```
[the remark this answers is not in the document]
mitza-oci: The `tao_idl` options include both `-Sa` and `-St`, so both anys and typecodes are involved.  The MPC feature is `dds_suppress_anys` although it impacts both.  One could decide to have typecodes without anys (but not the other way around) but maybe this could be handled by directly adding the command line parameters to the rules for the code generators.

tests/DCPS/Compiler/typecode/C++11/typecode.cpp
```diff
@@ -1,17 +1,9 @@
...
-
 using namespace OpenDDS::DCPS;
 
-TEST(TestTypecodes, can_allocate)
-{
-  ASSERT_EQ(my_module::_tc_my_struct, nullptr);
-  ASSERT_EQ(my_module::_tc_my_union, nullptr);
-}
-
 int main(int argc, char* argv[])
```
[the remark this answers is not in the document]
mitza-oci: ACE may still require the names here due to the preprocessor replacement of main.  Or you can turn off that behavior of ACE with a preprocess macro.

mitza-oci (approved):

cmake/api_macros.cmake
```diff
@@ -206,7 +215,7 @@ macro(OPENDDS_TARGET_SOURCES target)
     message(FATAL_ERROR "OpenDDS does not support argument items in CMAKE_CXX_COMPILER.")
   endif()
 
-  if(OPENDDS_SUPPRESS_ANYS)
+  if(${_suppress_anys} MATCHES ON)
```
iguessthislldo, line 218:
```suggestion
  if(${_suppress_anys})
```

cmake/api_macros.cmake
```diff
@@ -44,6 +46,10 @@ macro(OPENDDS_GET_SOURCES_AND_OPTIONS
   set(${tao_options} ${_arg_TAO_IDL_OPTIONS})
   set(${opendds_options} ${_arg_OPENDDS_IDL_OPTIONS})
 
+  if (${_arg_SUPPRESS_ANYS} MATCHES ON OR ${_arg_SUPPRESS_ANYS} MATCHES OFF)
```
iguessthislldo, line 49: DEFINED?

iguessthislldo (changes requested):

cmake/api_macros.cmake
ClaytonCalabrese: I don't think that DEFINED works.  It also appears that what I have currently does not work on certain builds. I'll try to a find a solution.

iguessthislldo (approved):

PR #2632, opened 2021-04-30
Fix Assertion Issues From Inverted GAP Submessage Values

simpsont-oci:
This is the initial fix for the GAP inversion assertion issue. Once a sender-side fix is in place to prevent the invalid GAP submessages, the OPENDDS_ASSERT statements can be removed.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1452,6 +1452,7 @@ RtpsUdpDataLink::RtpsWriter::add_gap_submsg_i(RTPS::SubmessageSeq& msg,
     to_rtps_seqnum(gap_start),
     {to_rtps_seqnum(max_sn_), 0, bitmap}
   };
+  OPENDDS_ASSERT(gap_start < max_sn_);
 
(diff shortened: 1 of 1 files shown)


jrw972 (approved):

jrw972 (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1080,7 +1080,7 @@ RtpsUdpDataLink::RtpsWriter::customize_queue_element_helper(
     max_sn_ = std::max(max_sn_, seq);
     if (!durable_ && !is_pvs_writer() &&
         element->subscription_id() == GUID_UNKNOWN &&
-        previous_max_sn != max_sn_.previous()) {
+        previous_max_sn < max_sn_.previous()) {
```
jrw972, line 1083: The implication is that `previous_max_sn` > `max_sn_.previous()` is possible.  I would like to understand how this may occur.  If it is the for case of `previous-max_sn == max_sn` then that is probably okay.

simpsont-oci: Yeah, that's the only issue I saw, when they were equal, then previous_max_sn != max_sn_.previous() would be true and it would generate an inverted gap.

PR #2634, opened 2021-05-03
Upgrade to vcpkg 2021.04.30

jwillemsen:
* .github/workflows/build_and_test.yml:


mitza-oci:
Looks like in this version, `openssl-windows` should be just `openssl`.  The former will probably still work, but it's deprecated.

mitza-oci (approved):

PR #2635, opened 2021-05-03
GitHub Actions: Wireshark Add apt-get update

ClaytonCalabrese:
Wireshark was failing on our ubuntu builds due to a lack of apt-get update. I took this time to clean up some mess around wireshark that we had.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -2034,7 +2034,7 @@ jobs:
     - name: install xerces
       run: sudo apt-get -y install libxerces-c-dev
     - name: install wireshark
-      run: sudo apt-get -y install wireshark-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -2034,7 +2034,7 @@ jobs:
     - name: install xerces
       run: sudo apt-get -y install libxerces-c-dev
     - name: install wireshark
-      run: sudo apt-get -y install wireshark-dev
+      run: sudo apt-get update -y install wireshark-dev
```
mitza-oci, line 2037: I don't think this works.  `apt-get` accepts one command at a time.

mitza-oci (changes requested):

mitza-oci (approved):

PR #2638, opened 2021-05-04
Fix .gitignore for ManualAssertLiveliness subscriber

simpsont-oci:


mitza-oci:
It was removed in #2568.

PR #2642, opened 2021-05-05
GitHub Actions: Make use of concurrency cancel-in-progress

### LINKED ISSUE #2612: Github Actions: Make use of cancel-in-progress
github concurrency now seems to have support for cancelling any running github actions, no need for a custom action from the marketplace, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idconcurrency

jwillemsen:
Fixed #2612


.github/workflows/cancel_previous_runs.yml
```diff
@@ -1,18 +0,0 @@
-name: Cancel Duplicate Workflow Runs
```
simpsont-oci, line 1: If you're going to remove cancel_previous_runs.yml, you'll also need to remove the wait_cancel_previous job from build_and_test.yml as well as all the dependencies upon it.

simpsont-oci (changes requested):

simpsont-oci (approved):

PR #2646, opened 2021-05-06
Memory Leak Fix: Security Attributes Test

ClaytonCalabrese:
Problem: 

The security attributes test had a memory leak that occurs when the publisher/writer and subscriber/reader are never supposed to match.

Solution:  

I have added better handling for the timeout scenario in Writer.cpp to preempt the ending of the waitset and allow cleanup of the writer.  Currently the find match timeout is set to 5 seconds if a match doesn't happen. Considering we give most of the test scenarios 10 or 20 seconds, I think 5 seconds will work but am open to changing it. 

When this timeout occurs, I added additional cleanup to publisher.cpp which will properly clean up the writer before a CLEAN_ERROR_RETURN which then cleans everything else up.

### DIFF
--- a/tests/security/attributes/Writer.cpp
+++ b/tests/security/attributes/Writer.cpp
@@ -58,13 +58,18 @@ Writer::svc()
     ws->attach_condition(condition);
 
     DDS::Duration_t timeout =
-      { DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC };
+      { 5, DDS::DURATION_INFINITE_NSEC };
 
     DDS::ConditionSeq conditions;
     DDS::PublicationMatchedStatus matches = {0, 0, 0, 0, 0};
 
     do {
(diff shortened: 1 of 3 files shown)


tests/security/attributes/Writer.cpp
```diff
@@ -58,13 +58,18 @@ Writer::svc()
     ws->attach_condition(condition);
 
     DDS::Duration_t timeout =
-      { DDS::DURATION_INFINITE_SEC, DDS::DURATION_INFINITE_NSEC };
+      { 5, DDS::DURATION_INFINITE_NSEC };
```
simpsont-oci, line 61: Rather than some arbitrary time, let's use a GuardCondition here and trigger it in Writer::end(). See performance-tests/bench/test_controller/ScenarioManager.cpp for an example of how to combine GuardCondition with other conditions on a WaitSet.

simpsont-oci (changes requested):

simpsont-oci (approved):

PR #2653, opened 2021-05-12
RtpsUdpDataLink can't handle more than one unicast address

jrw972:
Problem
-------

A participant A with multiple IP addresses may advertise all of them
in the locators of SPDP and SEDP.  The RtpsUdpDataLink of a peer B
will pick the first of each type, unicast and multicast, to determine
the addresses of entities in A.  Thus, participant B may use the wrong
address and fail to communicate with A.

Solution
--------

Use all of the unicast addresses.

Notes
-----

* ICE is useful in these situations as it will determine the correct
  address to use.

* When security is enabled, ICE will not start until secure discovery
  is complete meaning authentication and key exchange may fail without
  this feature.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1464,18 +1464,31 @@ RtpsUdpDataLink::RtpsWriter::add_gap_submsg_i(RTPS::SubmessageSeq& msg,
   msg[i].gap_sm(gap);
 }
 
+void RtpsUdpDataLink::update_last_recv_addr(const RepoId& src, const ACE_INET_Addr& addr)
+{
+  if (addr == config().rtps_relay_address()) {
+    return;
+  }
+
+  ACE_GUARD(ACE_Thread_Mutex, g, locators_lock_);
+  const RemoteInfoMap::iterator pos = locators_.find(src);
+  if (pos != locators_.end()) {
+    pos->second.last_recv_addr_ = addr;
+    pos->second.last_recv_time_ = MonotonicTimePoint::now();
+  }
+}
 
 // DataReader's side of Reliability
 
 void
 RtpsUdpDataLink::received(const RTPS::DataSubmessage& data,
-                          const GuidPrefix_t& src_prefix)
+                          const GuidPrefix_t& src_prefix,
+                          const ACE_INET_Addr& remote_addr)
 {
   const RepoId local = make_id(local_prefix_, data.readerId);
+  const RepoId src = make_id(src_prefix, data.writerId);
 
-  RepoId src;
-  std::memcpy(src.guidPrefix, src_prefix, sizeof(GuidPrefix_t));
-  src.entityId = data.writerId;
+  update_last_recv_addr(src, remote_addr);
 
   OPENDDS_VECTOR(RtpsReader_rch) to_call;
   {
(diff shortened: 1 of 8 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -537,29 +537,34 @@ RtpsUdpDataLink::leave_multicast_group(const NetworkInterface& nic)
...
-                              const ACE_INET_Addr& wide_address,
+                              const AddrSet& narrow_addresses,
+                              const AddrSet& wide_addresses,
                               bool requires_inline_qos)
 {
-  if (narrow_address == ACE_INET_Addr()) {
-    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: RtpsUdpDataLink::add_locators: narrow_address for %C is empty\n"), LogGuid(remote_id).c_str()));
-    return;
+  if (narrow_addresses.empty()) {
+    ACE_ERROR((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: RtpsUdpDataLink::add_locators: narrow_addresses for %C is empty\n"), LogGuid(remote_id).c_str()));
```
mitza-oci, line 545: error or warning?  also on 549.

If it ends up as a warning, should there be an error if both sets are empty?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -537,29 +537,34 @@ RtpsUdpDataLink::leave_multicast_group(const NetworkInterface& nic)
 
 void
 RtpsUdpDataLink::add_locators(const RepoId& remote_id,
-                              const ACE_INET_Addr& narrow_address,
-                              const ACE_INET_Addr& wide_address,
+                              const AddrSet& narrow_addresses,
+                              const AddrSet& wide_addresses,
```
mitza-oci, line 541: narrow/wide may be confusing - should these be renamed?

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: We discussed this with @simpsont-oci when this code first went in.  What was the motivation for not using unicast and multicast?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
jrw972: I'm thinking an error if both are empty and that's it.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: The general motivation was because "multicast_address" makes no sense when multicast is disabled. I agree, the names are still potentially confusing. Perhaps undirected / directed or spread / focused or something similar?

jrw972: My thought is that a multicast address is still a multicast address, even if we are not using it.

I've updated the code is a couple places to accommodate this.  First, when we are extracting locators from the blob, multicast addresses will be skipped in multicast is disabled.  Second, when we get to the point of selecting addresses, we1
1. Use the unicast address if preferred and available.
2. Use the multicast address if available.
3. Fall back to the unicast address.

So, 1 and 3 together means they when multicast is disable, they will always get the unicast locators.

simpsont-oci: > My thought is that a multicast address is still a multicast address, even if we are not using it.

The previous code was setting a variable named "multicast_address" to be a unicast address.

mitza-oci (approved):

PR #2661, opened 2021-05-14
Fix Deadlock in DataReaderImpl's remove_association

simpsont-oci:


mitza-oci (approved):

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -588,6 +589,11 @@ DataReaderImpl::remove_associations_i(const WriterIdSeq& writers,
     }
   }
 
+  while (!removed_writers.empty()) {
+    removed_writers.begin()->second->removed();
+    removed_writers.erase(removed_writers.begin());
+  }
+
```
jrw972, line 596: Since you are erasing in the loop, the complexity of this code is O(n * log(n)).
I believe iterating over the map is O(n) and clearing it (if necessary) is O(n).
Thus, I think it would be more efficient to just iterate over is as opposed to erasing from it until it was empty.

jrw972 (changes requested):

jrw972 (approved):

PR #2663, opened 2021-05-14
handle unknown entityid in sedp.cpp

ClaytonCalabrese:
Previously we were erroring if both is_final and is_mutable were false.  instead we should silently return, and only error if both are true.


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4468,10 +4468,14 @@ Sedp::Reader::data_received(const DCPS::ReceivedDataSample& sample)
...
-        to_string(entity_id).c_str(), is_mutable));
-      break;
+      if (is_mutable) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Sedp::Reader::data_received: "
+          "entity id extensibility error over %C: both is_final and is_mutable are %d\n",
+          to_string(entity_id).c_str(), is_mutable));
+        break;
+      } else {
+        return;
+      }
```
iguessthislldo, line 4478:
```suggestion
      }
      break;
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #2664, opened 2021-05-15
Run CMake Tests with CTest and `auto_run_tests.pl`

iguessthislldo:
Added `add_test`s to the CMake tests. `auto_run_tests.pl` will now be able to use CTest to run them and it will convert the results to look like the existing test results. This is enabled with the `--cmake` option.

Also:
  - Make CMake project and target names consistent.
  - Merge `DevGuideExamples/DCPS/CMakeLists.txt` into `DevGuideExamples/DCPS/Messenger/CMakeLists.txt` and copy files needed for test into the build output. This makes the instructions for using CMake with the example messenger much simpler.
  - Since I'm changing auto_run_tests, I'm including the QOL improvements from https://github.com/objectcomputing/OpenDDS/pull/2452 namely:
    - Now able to interrupt the script with ctl-c when it's running tests. Before it would just skip the currently running test and go on to the next one.
    - A `--help` message that's accurate.
    - The normal list files are now built-in as options: `--no-dcps`, `--security`, `--java`, `--modeling`
    - New options for listing information about the test lists.


tests/auto_run_tests.pl
```diff
@@ -2,60 +2,228 @@
...
+    },
+);
+my %builtin_test_lists_hash = map { $_->{name} => $_ } @builtin_test_lists;
+
+sub print_usage {
+    my $error = shift // 1;
+
+    my $fd = $error ? *STDERR : *STDOUT;
+    print $fd
+        "auto_run_tests.pl [<options> ...] [<list_file> ...]\n" .
```
mitza-oci, line 89: Why the change from using `-l` to specify list files?  This is now diverging from ACE - not necessarily a problem but it may be confusing

tests/auto_run_tests.pl
```diff
@@ -2,60 +2,228 @@
...
+    my %configs = ();
+    foreach my $test_list (@file_list) {
+        my $config_list = new PerlACE::ConfigList;
+        $config_list->load($test_list);
+        for my $config (split(/ /, $config_list->list_configs())) {
+            $configs{$config} = 1;
+        }
+    }
+    for my $config (sort(keys(%configs))) {
+        print("$config\n");
```
mitza-oci, line 217:
```suggestion
        print "$config\n";
```
Just for consistency... there's a few more of these below as well

tests/auto_run_tests.pl
```diff
@@ -2,60 +2,228 @@
...
+    }
+    exit(0);
 }
 
-my $cwd = getcwd();
+if (!$query) {
```
mitza-oci, line 222: It's not clear to my how the `$query` variable differs from just using `$list_tests` since the other two options (show_configs, list_configs) have already existed by this point.

mitza-oci (changes requested):

tests/auto_run_tests.pl
iguessthislldo: ACE's already accepts positional argument list files, and we completely control this one and usage of this script, so it made sense to me to eliminate `-l` if it is no longer needed.

tests/auto_run_tests.pl
iguessthislldo: Yes, you're right. I was curious why I did this, so I looked at #2452. It looks like this is a copy of a similar thing for automatic options that occurs earlier in the script: https://github.com/objectcomputing/OpenDDS/pull/2452/files#diff-53f3fd860cfed00c421271ed5cb9c434a1bd9fc43f6b32bce4e5e53fb8f89451R272-R276

tests/auto_run_tests.pl
```diff
@@ -2,60 +2,228 @@
...
+    }
+    for my $config (sort(keys(%configs))) {
+        print "$config\n";
+    }
+    exit(0);
 }
 
-my $cwd = getcwd();
+if (!$list_tests) {
+    print "Test Lists:", join(',', @file_list), "\n");
```
mitza-oci, line 223:
```suggestion
    print "Test Lists:", join(',', @file_list), "\n";
```

mitza-oci (changes requested):

mitza-oci (approved):

iguessthislldo:
Scoreboard file changes are ready to be pushed when ever this is going to be merged.

PR #2671, opened 2021-05-18
tests/DCPS/unit is poorly named

jrw972:
Problem
-------

tests/DCPS/unit is poorly named.

Solution
--------

Rename to tests/DCPS/MockedTypeSupport.


iguessthislldo (approved):
:+1:

PR #2672, opened 2021-05-18
Update `documents/building.html`

iguessthislldo:


mitza-oci (commented):
These changes are fine but I think we also need to update the OS/compiler lists

documents/building.html
```diff
@@ -83,7 +83,7 @@
       <li>gcc 6.3</li>
       <li>gcc 7.2</li>
       <li>gcc 10.2</li>
```
mitza-oci, line 85:
```suggestion
      <li>gcc 11.1</li>
```

documents/building.html
```diff
@@ -83,7 +83,7 @@
...
-      <li>Clang 10.0.1 (llvm.org) and 11.0.3 (Apple)</li>
+      <li>Clang 11.0.1 (llvm.org) and 11.0.3 (Apple)</li>
```
mitza-oci, line 86:
```suggestion
      <li>Clang 12.0.1 (llvm.org) and 11.0.3 (Apple)</li>
```

mitza-oci (changes requested):
These changes are fine but I think we also need to update the OS/compiler lists

mitza-oci (approved):

PR #2679, opened 2021-05-20
Fix `configure --wireshark-cmake` on Linux

iguessthislldo:
On Linux when using `--wireshark-cmake`, `--wireshark-lib` defaults to empty. Between making that option and now, a check was added for `%optdep` locations that require them not to be empty. This added a field to `%optdep` that allows the value to be empty.

As part of this I also rewrote `%optdep` to use hashes instead of arrays.

### DIFF
--- a/configure
+++ b/configure
@@ -1152,27 +1152,39 @@ sub write_platform_macros {
 }
 
 ## Optional OpenDDS dependencies
-
-my %optdep =
-# %opts key =>    [env var,           sanity check,                    MPC feature]
-  ('java' =>      ['JAVA_HOME',       'include/jni.h',                 'java'],
-   'jboss' =>     ['JBOSS_HOME',      'lib/jboss-common.jar'],
-   'ant' =>       ['ANT_HOME',        'bin/ant'],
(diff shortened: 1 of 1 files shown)


mitza-oci:
Is it feasible to change (or add to) the GitHub Actions configs to have one of the Linux Wireshark-enabled build use `--wireshark-cmake`?  Does that work with the .deb package or does it need the Wireshark source?

iguessthislldo:
It would have to be a Wireshark built using CMake to use `--wireshark-cmake`. Everything else, including an installed Wireshark built using CMake, needs to use `--wireshark`.

mitza-oci:
> 
> 
> It would have to be a Wireshark built using CMake to use `--wireshark-cmake`. Everything else, including an installed Wireshark built using CMake, needs to use `--wireshark`.

The Windows Wireshark setup could be ported over since it's built on cross-platform tools like vcpkg and CMake.  vcpkg may not be needed if the system's package manager is used for development tools and libs.

simpsont-oci (approved):

PR #2689, opened 2021-05-24
[Bench] Attempt to address CI failures with QoS updates

simpsont-oci:
Problem:
For some of the CI tests that send multiple samples, the instance is the same, meaning that with the default history settings a sample can be lost if a subsequent sample overwrites it for the same instance.

Solution:
For these bench tests, we probably ought to be using HISTORY_KEEP_ALL to avoid losing samples in this way. A longer term solution might be to update the logic in Bench's worker to allow missing data when samples are from the same instance and don't force the use of HISTORY_KEEP_ALL, but this solution should help the CI tests to pass again.


simpsont-oci:
Replaced by #2760

PR #2692, opened 2021-05-25
extension to the XTypes test

pmesnier:
Catching some late-emerging scoreboard failures.
1. an inconsistent topic count went to 2 and thus was counted as a failure (addressed)
2. reader still failed to detect a message sent by the writer or the writer failed to detect the reader was still subscribed.

### DIFF
--- a/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
+++ b/tests/DCPS/XTypes/Pub/XTypesPublisher.cpp
@@ -464,6 +464,10 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     return 1;
   }
 
+  if (!wait_for_reader (true, control_dw)) {
+    return 1;
+  }
+
   ACE_DEBUG((LM_DEBUG, "Writer sending echo at %T\n"));
 
   ControlStruct cs;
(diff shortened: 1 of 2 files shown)


mitza-oci:
>     1. an inconsistent topic count went to 2 and thus was counted as a failure (addressed)

Is seeing a count of 2 expected based on how the test is designed?

>     2. reader still failed to detect a message sent by the writer or the writer failed to detect the reader was still subscribed.

Is this one not yet addressed?

pmesnier:
> > ```
> > 1. an inconsistent topic count went to 2 and thus was counted as a failure (addressed)
> > ```
> 
> Is seeing a count of 2 expected based on how the test is designed?

I don't know if it was expected, but it seems like it is a valid case based on the evidence in the log file. It looks like a thread race result as both the reader and writer in this case do not match the topic so it seems that while most of the time a single topic inconsistency is able to trigger the waiter to return, this time both inconsistencies were detected before the topic state was evaluated.

> 
> > ```
> > 2. reader still failed to detect a message sent by the writer or the writer failed to detect the reader was still subscribed.
> > ```
> 
> Is this one not yet addressed?
Not yet. I'm still trying to determine what exactly can happen.

pmesnier:
I'm still not 100% sure I've resolved the timeout failures but I've added another sync point and more log output. I'd like to get this merged so that the main scoreboard can be affected.

ClaytonCalabrese (approved):

pmesnier:
got one approval, merging

PR #2697, opened 2021-05-26
GitHub Actions: renamed and updated configurations of some jobs

lij-oci:


.github/workflows/build_and_test.yml
```diff
@@ -6776,17 +6776,17 @@ jobs:
       shell: cmd
       run: |
         cd OpenDDS
-        configure --no-debug --no-inline --tests --security --ace=%GITHUB_WORKSPACE%/OpenDDS/ACE_TAO/ACE --tao=%GITHUB_WORKSPACE%/OpenDDS/ACE_TAO/TAO --mpc=%GITHUB_WORKSPACE%/MPC --xerces3=%VCPKG_ROOT%/installed/x86-windows --openssl=%VCPKG_ROOT%/installed/x86-windows --mpcopts=-hierarchy
+        configure --no-inline --tests --security --ace=%GITHUB_WORKSPACE%/OpenDDS/ACE_TAO/ACE --tao=%GITHUB_WORKSPACE%/OpenDDS/ACE_TAO/TAO --mpc=%GITHUB_WORKSPACE%/MPC --xerces3=%VCPKG_ROOT%/installed/x86-windows --openssl=%VCPKG_ROOT%/installed/x86-windows --mpcopts=-hierarchy
```
simpsont-oci, line 6779: The 'd0' in the job names signifies the debug feature of MPC is disabled. This doesn't have any real correspondence to the "Debug" configuration for MSVC. I think we probably need to keep the --no-debug flag here, or we need to rename all the w16_x86_d0i0 jobs to reflect this change.

mitza-oci: The `--debug` command line option to the configure script does not correspond to an MPC feature.  If the intent is to build a Visual C++ "Debug" configuration, then `--debug` should be there for the configure script.

lij-oci: Ok, let's keep the 'd0' and close this PR, since the real problem was fixed in the PR for NetworkAddress changes.

lij-oci: If we replace 'configure --no-debug ...' with 'configure --debug ...' here, then should we take out the 'd0' from 'ACE_TAO_w16_x86_d0i0' and 'build_w16_x86_d0i0_sec' ?

simpsont-oci: I don't think the intent was to 'build a Visual C++ "Debug" configuration'. It was to match the code coverage described in CI Coverage spreadsheet. Looking at it again, it's not clear to me which tab we we're claiming to have implemented. On some tabs, these builds are marked as "Debug" configuration and on others they are "Release", but in all cases the "dbg" column is "N" so I think we ought to leave that part alone.

simpsont-oci (approved):

simpsont-oci:
Hey @iguessthislldo Can you take a look at the cmake failures in test_w16_x86_i0_sec (you'll need to open the autobuild_output artifact) and determine whether this is something that's been caused by this PR or something that needs to be addressed separately?

ClaytonCalabrese (approved):

PR #2700, opened 2021-05-27
GitHub Actions: Normalize Calling 'apt-get update' Across Related Jobs

simpsont-oci:
So we don't have a weird mix of xerces / qt / wireshark versions, and so that these larger dependency installations continue to work over time (QT has been causing trouble on other builds).

And upgrade that one cache 2.1.4 action to match the others (2.1.5)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1952,9 +1952,6 @@ jobs:
       with:
         path: ${{ github.job }}.tar.xz
         key: c01_${{ github.job }}_master_${{ env.ACE_COMMIT }}
-    - name: update apt
-      if: steps.cache-artifact.outputs.cache-hit != 'true'
-      run: sudo apt-get update
     - name: install xerces
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       run: sudo apt-get -y install libxerces-c-dev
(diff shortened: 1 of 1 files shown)


mitza-oci (approved):

.github/workflows/build_and_test.yml
```diff
@@ -3533,6 +3540,8 @@ jobs:
     steps:
     - name: install xerces
       run: sudo apt-get -y install libxerces-c-dev
+    - name: update apt
+      run: sudo apt-get update
```
iguessthislldo, line 3544: Shouldn't this go before installing xerces?

simpsont-oci: I originally had it that way (see commit history), but it would be a change from how we were doing it before and, more importantly, it would mean we might potentially get different versions of xerces between what was used by the cached ACE artifact and what's used by the test & build jobs. If we know they'll be ABI compatible, it might not matter, but I was worried it might expect the same version.

On the other hand, I don't know that not upgrading apt would even avoid this issue 100%, since the virtual environment runner can still get upgraded out from underneath us (and so the "default" version of xerces may change). But this approach seemed like it might be the safest bet to sidestep the issue for cached ACE artifacts & xerces versions.  Happy to try it the other way if we don't think there will be any issues...

PR #2704, opened 2021-05-28
Attempt to resolve NetworkAddress warning on vs2017 x86 builds

simpsont-oci:


simpsont-oci:
I think the one multicast failure in messenger is unrelated... it's a failure to join the (correct) multicast group... so I think this one is ready to go.

mitza-oci (approved):

PR #2705, opened 2021-05-30
Mass Spelling Correction

iguessthislldo:
After [seeing it on Hackaday](https://hackaday.com/2021/05/29/spell-checking-your-programming-from-the-linux-command-line/) I wanted to try to use [the typos source code spell checker](https://github.com/crate-ci/typos) on select locations, namely `dds`. I didn't really want to fix things like `docs/design`, `doc/history`, `tools/modeling`, and most of tests. I've gone through them all and they seem correct.

Seems pretty lax and will only go after something it really thinks is a misspelled word. The only false positive I think saw for the entire tree was for stuff inside `docs/logo.svg`.

### DIFF
--- a/dds/DCPS/transport/framework/TransportClient.h
+++ b/dds/DCPS/transport/framework/TransportClient.h
@@ -35,7 +35,7 @@ namespace DCPS {
 
 class EntityImpl;
 class TransportInst;
-class AssocationInfo;
+class AssociationInfo;
 class ReaderIdSeq;
 class WriterIdSeq;
 class SendStateDataSampleList;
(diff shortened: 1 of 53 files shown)


jwillemsen:
Interesting, I will try this next week on some DOCGroup repositories

dds/DCPS/transport/framework/TransportClient.h
```diff
@@ -35,7 +35,7 @@ namespace DCPS {
 
 class EntityImpl;
 class TransportInst;
-class AssocationInfo;
+class AssociationInfo;
```
simpsont-oci, line 38:
```suggestion
```
Evidently we don't need this?

iguessthislldo: No and there's not even a class with that name anymore.

iguessthislldo: Looking at all of the forward declarations in that group, it doesn't look like any of them are useful anymore. I wish the compilers or something else would help us with this.

simpsont-oci (approved):

PR #2708, opened 2021-06-01
GitHub Actions: added compiler_version as part of cache key

lij-oci:


.github/workflows/build_and_test.yml
```diff
@@ -877,7 +877,7 @@ jobs:
       uses: actions/cache@v2.1.6
       with:
         path: ${{ github.job }}.tar.xz
-        key: c01_${{ github.job }}_ace6tao2_${{ env.ACE_COMMIT }}
+        key: "c01_${{ github.job }}_ace6tao2_${{ env.ACE_COMMIT }}_clang++-5.0"
```
simpsont-oci, line 880: We want something here that will change when the compiler version string changes (e.g. the output of ```clang++-5.0 --version```). Since the output of that command is probably not a great candidate for concatenation with the end of the cache key, you should probably use something like ```clang++-5.0 --version | md5sum``` and it will be compact but still be robust against version changes.

If you're not sure about how to append that to the key, take a look at how the 'get ACE_TAO commit' step is writing the ```ACE_COMMIT``` environment variable into ```$GITHUB_ENV``` which is then being used as ```env.ACE_COMMIT``` in the key. I would add another step before `Cache Artifact` that queries the compiler version and writes it into the GitHub Environment similar to `get ACE_TAO commit`.

Ditto for basically every `ACE_TAO` or `build` job.

simpsont-oci (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       run: |
         cd OpenDDS/ACE_TAO
         export ACE_COMMIT=$(git rev-parse HEAD)
-        echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
+        echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV #??
```
iguessthislldo, line 85:
```suggestion
        echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
```

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

.github/workflows/build_and_test.yml
```diff
@@ -872,12 +882,17 @@ jobs:
         cd OpenDDS/ACE_TAO
         export ACE_COMMIT=$(git rev-parse HEAD)
         echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
+    - name: get Compiler version
+      shell: bash
+      run: |
+        export COMPILER_VERSION=$("c++ --version 2>&1 | head -n 1")
```
iguessthislldo, line 888: This build is using installing and using clang, so `c++` isn't going to be the right compiler.

iguessthislldo (changes requested):

.github/workflows/build_and_test.yml
lij-oci: Ok, I'll change it to "clang++ --version 2>&1 | head -n 1"
That will be the same for these two jobs, right?
1) u18_clang5_i0w1_sec
2) u18_clang10_sec_js0

.github/workflows/build_and_test.yml
```diff
@@ -83,12 +83,17 @@ jobs:
         cd OpenDDS/ACE_TAO
         export ACE_COMMIT=$(git rev-parse HEAD)
         echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
+    - name: get Compiler version
```
simpsont-oci, line 86:
```suggestion
    - name: get compiler version
```
Minor nit-pick.

.github/workflows/build_and_test.yml
```diff
@@ -83,12 +83,17 @@ jobs:
...
+      shell: bash
+      run: |
+        export COMPILER_VERSION=$("c++ --version 2>&1 | head -n 1")
```
simpsont-oci, line 89: I don't think this will evaluate correctly with the quotes, and it will grab the whole first line of the version which can have spaces and possibly fill up the cache key more than we need (I believe there's a 256 byte character limit). Maybe something like:
```suggestion
        export COMPILER_VERSION=$(c++ --version |& head -n 1 | md5sum | cut -d ' ' -f 1)
```

.github/workflows/build_and_test.yml
simpsont-oci: You'll want to explicitly add the version to the executable or you'll be getting a system default. For Ubuntu, I believe this looks like: clang++10, clang++9, clang++8, clang++7, clang++6.0, clang++5.0 (note the decimal in 6 and 5).

simpsont-oci (changes requested):

.github/workflows/build_and_test.yml
lij-oci: replaced with "clang++-5.0 --version ..." or "clang++-10 --version ..."

simpsont-oci (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -1705,12 +1735,17 @@ jobs:
         cd OpenDDS/ACE_TAO
         export ACE_COMMIT=$(git rev-parse HEAD)
         echo "ACE_COMMIT=$ACE_COMMIT" >> $GITHUB_ENV
+    - name: get compiler version
+      shell: bash
+      run: |
+        export COMPILER_VERSION=$(c++ --version 2>&1 | head -n 1)
```
simpsont-oci, line 1741:
```suggestion
        export COMPILER_VERSION=$(g++-10 --version 2>&1 | head -n 1)
```

simpsont-oci (approved):

iguessthislldo (approved):

PR #2716, opened 2021-06-03
More Fixes for CMake Tests on Windows

iguessthislldo:
Follow up to #2712

- `auto_run_tests.pl`:
  - Make a separate ` --cmake-build-cfg`  for `ctest --build-config` because using the same value of `-ExeSubDir` as the MPC-based tests seems impossible. Default this to `"Debug"` on Windows for convenience.
  - Default `--python` on Windows to `python`.
- Detect when a required build configuration is missing in `ctest-to-auto-run-tests.py` to give a friendly error instead of a confusing stack trace.

Also:
- Add Python to `docs/dependencies.md`
- Fixed spelling in `tests/cmake`

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -17,6 +17,7 @@
   * [Boost](#boost)
   * [Xerces](#xerces)
   * [OpenSSL](#openssl)
+  * [Python](#python)
 
 ## Required to Build the Core OpenDDS Libraries
 
@@ -26,7 +27,7 @@ downloaded automatically by the configure script by default.**
 ### Perl
 
 Perl is an interpreted language used in the configure script, the tests, and
-any other scripting in OpenDDS codebase. Even if the configure script is not
+most other scripting in OpenDDS codebase. Even if the configure script is not
 used, it is also required to run MPC, so it is required to build OpenDDS.
 
 Testing scripts are written in Perl and use the common PerlACE modules provided
@@ -191,3 +192,15 @@ security XML configuration files.
 
 OpenSSL is used for DDS Security for verifying security configurations and
 encryption and decryption. Version 1.0 and version 1.1 are supported.
+
+### Python
+
+Python is used for some scripts where Perl isn't as suitable. Most notably this
+includes the [Sphinx-based documentation](README.md#sphinx-documentation) and
+processing the results of the CMake tests in `auto_run_tests.pl` if `--cmake`
+is passed.
+
+Unless noted otherwise, Python should be version 3.6 or later. Because it's an
+optional dependency, Python should not be required for any script used for
+building and testing the core functionality of OpenDDS. Right now only Perl can
+be used for situations like that.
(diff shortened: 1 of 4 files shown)


tests/auto_run_tests.pl
```diff
@@ -227,9 +222,12 @@ sub print_help {
 my $list_tests = 0;
 my $cmake = 0;
 my $cmake_build_dir = "$cmake_tests/build";
+my $cmake_build_cfg = windows ? 'Debug' : undef;
 my $ctest = 'ctest';
 my $ctest_args = '';
-my $python = 'python3';
+# Python on Windows is called python from what I can see. On other platforms
+# that are legacy-mindful python is Python 2, but we can rely on python3.
+my $python = windows ? 'python' : 'python3';
```
mitza-oci, line 230: `python3` doesn't appear to be part of Python itself according to https://docs.python.org/3/using/cmdline.html
It may be a change the distro packagers make.  That doesn't really impact what we need to do, but I think this is the other way around from what the comment says.  Some platforms that encourage every single program on the computer to share a common directory location need to resolve naming conflicts.  Other platforms let different things exist in different directories and users can decide what to run explicitly or with `PATH`.

docs/dependencies.md
```diff
@@ -191,3 +192,15 @@ security XML configuration files.
 
 OpenSSL is used for DDS Security for verifying security configurations and
 encryption and decryption. Version 1.0 and version 1.1 are supported.
+
+### Python
+
+Python is used for some scripts where Perl isn't as suitable. Most notably this
+includes the [Sphinx-based documentation](README.md#sphinx-documentation) and
+processing the results of the CMake tests in `auto_run_tests.pl` if `--cmake`
+is passed.
```
mitza-oci, line 201: Can you run `auto_run_tests` with `--cmake` and no Python if you don't need to process the results?

iguessthislldo: Not how it's setup right now. `--python` just manually specifies the Python to use.

tests/auto_run_tests.pl
iguessthislldo: I think that page is more focused on the interpreter arguments. `python3` is definitely part of Python. In fact looking at the builds for Linux I did the other day, there is no `python` there, just `python3` which symlinks to `python3.9`.

docs/dependencies.md
mitza-oci: Is this use case both worthwhile and easy to implement?  The user would be just running the tests interactively, not attempting to have the results understood by Jenkins, autobuild/scoreboard, etc.

iguessthislldo: Actually right now would just say it couldn't find Python and present it as a failed test, so it doesn't cause the script to fail. Is that what you meant?

mitza-oci: I think so, except for the "failed test" part.  But I can see why that would be important for the primary use case.

tests/auto_run_tests.pl
mitza-oci: Looks like `make` builds an interpreter called `python`, but the `make install` process renames it.  It's unfortunate that the page of their documentation that seemed to be the most relevant doesn't cover this.

mitza-oci (approved):

ClaytonCalabrese (approved):

PR #2717, opened 2021-06-04
Bulk partition update in RtpsRelay is inefficient

jrw972:
Problem
-------

A bulk update of the partitions serviced by an RtpsRelay instance has
two problems.  First, it is repetitive and wastes bandwidth since most
of the partitions will already be in the set.  Second, it is subject
to fragmentation which create additional overhead.

Solution
--------

Partition the set of partitions to create additional instances of the
topic.  The goal is that each instance is small enough to fit in a
single UDP message to avoid fragmentation but large enough to fill up
a message.


tools/rtpsrelay/GuidPartitionTable.h
```diff
@@ -104,29 +105,93 @@ class GuidPartitionTable {
...
 
-  void add_globally_new(const StringSet& partitions)
+  void add_new(const StringSet& partitions)
   {
-    std::copy(partitions.begin(), partitions.end(), std::back_inserter(relay_partitions_increment_.partitions()));
-    if (relay_partitions_increment_writer_->write(relay_partitions_increment_, DDS::HANDLE_NIL) != DDS::RETCODE_OK) {
-      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: failed to write Relay Partitions Increment\n")));
+    std::set<size_t> slots_to_write;
+    for (const auto& partition : partitions) {
+      size_t slot = get_free_slot();
```
mitza-oci, line 112:
```suggestion
      const size_t slot = get_free_slot();
```

tools/rtpsrelay/GuidPartitionTable.h
```diff
@@ -104,29 +105,93 @@ class GuidPartitionTable {
...
-      }
-      if (relay_partitions_writer_->write(relay_partitions_, DDS::HANDLE_NIL) != DDS::RETCODE_OK) {
+    write_slots(slots_to_write);
+  }
+
+  void remove_defunct(const StringSet& partitions)
+  {
+    std::set<size_t> slots_to_write;
+    for (const auto& partition : partitions) {
+      size_t slot = partition_to_slot_[partition];
```
mitza-oci, line 124:
```suggestion
      const size_t slot = partition_to_slot_[partition];
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #2719, opened 2021-06-04
Perl Module for Compiler Properties

iguessthislldo:
Added 2 Perl modules. One is called `CompilerProps` that collects the properties of a given compiler. The basic properties are the version, default  C++ standard supported, the target OS, and simplified target architecture, but there are more. This is meant to be used in the configure script and a standalone script at `tools/scripts/compiler_props.pl`.

The other Perl module is called `OsProps` and is meant to hold OS-specific stuff for the configure script and `compiler_props.pl`.


iguessthislldo:
Similar to #2452, after #4203 we can use CMake's built in features to do the same thing but without having to maintain the specifics. Though looking through it I think I was proud of this one, so it hurts a little to let it go.

PR #2723, opened 2021-06-08
DCPSChunkAssociationMutltiplier Spelling Fix

ClaytonCalabrese:
This variable was misspelled a number of places.  This pr fixes that.


iguessthislldo (changes requested):
This option looks like it's been misspelled for a very long time. I don't know how often this option is used, but if we want to handle this properly we should accept the old spelling as well. Otherwise if this options is in a user's ini it will be silently ignored. If it's still going to be valid in OpenDDS 4 we should also mark the old spelling as deprecated so we can remove it in, making a note of it in the NEWS file and warn about it.

dds/DCPS/Service_Participant.cpp
```diff
@@ -1664,6 +1664,8 @@ Service_Participant::load_common_configuration(ACE_Configuration_Heap& cf,
     if (got_chunk_association_multiplier) {
       ACE_DEBUG((LM_NOTICE, message, ACE_TEXT("DCPSChunkAssociationMultiplier")));
     } else {
+      // This is legacy support for a misspelling of the config option. 
+      GET_CONFIG_VALUE(cf, sect, ACE_TEXT("DCPSChunkAssociationMutltiplier"), this->association_chunk_multiplier_, size_t)
       GET_CONFIG_VALUE(cf, sect, ACE_TEXT("DCPSChunkAssociationMultiplier"), this->association_chunk_multiplier_, size_t)
     }
```
lij-oci, line 1670: Would Line 1669 overwrite Line 1668?

ClaytonCalabrese: If the user for some reason used both the old incorrect spelling and the new correct spelling, the new value would overwrite the old value. 

If the user used only the old spelling, it would not be overwritten because get_config_value would be a no-op for the correct spelling.

If the user only used the correct spelling, it would be a no-op for the old spelling and get the value from the correct spelling.

lij-oci (approved):

iguessthislldo (approved):

PR #2724, opened 2021-06-08
Added a Dockerfile as a starting point for OpenDDS development

mitza-oci:
Building and testing alternate configurations, especially complex ones
like Safety Profile, may be done using Docker containers.  This
Dockerfile sets up a baseline development environment using
Ubuntu 20.04 but doesn't make any decisions about how OpenDDS
is configured, which optional dependencies are used, etc.

### DIFF
--- a/tools/scripts/Dockerfile
+++ b/tools/scripts/Dockerfile
@@ -1,0 +1,48 @@
+# Using Docker for OpenDDS development
+#
+# Images created from this Dockerfile can be used as to launch containers
+# as build platforms for OpenDDS.  The difference between this image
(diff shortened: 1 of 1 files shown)


tools/scripts/Dockerfile
```diff
@@ -0,0 +1,48 @@
+# Using Docker for OpenDDS development
+#
+# Images created from this Dockerfile can be used as to launch containers
```
jrw972, line 3:
```suggestion
# Images created from this Dockerfile can be used to launch containers
```

jrw972 (changes requested):

PR #2728, opened 2021-06-10
rtps_udp: make data_delivered callbacks in order

mitza-oci:
based on sequence numbers since WriteDataContainer uses this to keep the
list of durable data in order

### DIFF
--- a/dds/DCPS/transport/framework/TransportQueueElement.h
+++ b/dds/DCPS/transport/framework/TransportQueueElement.h
@@ -163,7 +163,8 @@ public:
   struct OrderBySequenceNumber {
     bool operator()(const TransportQueueElement* lhs, const TransportQueueElement* rhs) const
     {
-      return lhs->sequence() < rhs->sequence();
+      const SequenceNumber seq_l = lhs->sequence(), seq_r = rhs->sequence();
+      return seq_l < seq_r || (seq_l == seq_r && lhs < rhs);
     }
   };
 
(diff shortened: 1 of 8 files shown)


jrw972 (approved):

dds/DCPS/transport/framework/TransportQueueElement.h
```diff
@@ -160,6 +160,13 @@ class OpenDDS_Dcps_Export TransportQueueElement : public PoolAllocationBase {
 
   virtual bool is_retained_replaced() const { return false; }
 
+  struct OrderBySequenceNumber {
+    bool operator()(const TransportQueueElement* lhs, const TransportQueueElement* rhs) const
+    {
+      return lhs->sequence() < rhs->sequence();
```
simpsont-oci, line 166: Add a fallback to pointer comparison when sequence numbers are equal (assuming this is possible?)
```suggestion
      return lhs->sequence() < rhs->sequence() || (lhs->sequence() == rhs->sequence() && lhs < rhs);
```

jrw972: I don't believe that is possible.  `to_deliver` is populated from `elems_not_acked_` which is a map from sequence number to TQE.  Thus, it looks like we are already assuming a one-to-one correspondence with sequence number.  If anything, we should add asserts when we insert into `elems_not_acked_` if we believe this is a problem.  Other transports may be different but they aren't using this code yet.

However, I'm okay with the extra check.

simpsont-oci (approved):

jrw972 (approved):

PR #2729, opened 2021-06-10
Delay TransportQueue memcpy Until Required

simpsont-oci:
Eventually, we should probably do the same optimization for the TransportQueueElements above (calling duplicate instead of clone_mb), but it requires more extensive changes to the transport framework.


jrw972 (approved):

simpsont-oci:
I'll investigate the FooTest5 multicast error. I believe I saw this on my other PR and resolved it, but can't remember the details.

simpsont-oci:
Looking at the remaining test failures, I don't think they are related to the changes in this PR. The multicast test failure is due to multicast group join errors, which is an ongoing problem, and I haven't been able to reproduce the modeling test SEGV, which is probably an issue with the test itself. The other failures seem typical.

jrw972 (approved):

PR #2732, opened 2021-06-11
Fix BestEffort Multicast, Restore Test Coverage

simpsont-oci:
At some point we seem to have gotten rid of test coverage for best effort (reliable=0 in the transport config) multicast. This restores test coverage and get it working again enough to process with performance testing.

### DIFF
--- a/tests/DCPS/Messenger/run_test.pl
+++ b/tests/DCPS/Messenger/run_test.pl
@@ -37,9 +37,17 @@ if ($test->flag('udp')) {
     $pub_opts .= " -DCPSConfigFile pub_udp.ini";
     $sub_opts .= " -DCPSConfigFile sub_udp.ini";
 }
+elsif ($test->flag('udp_free')) {
+    $pub_opts .= " -DCPSConfigFile pub_udp_free.ini";
+    $sub_opts .= " -DCPSConfigFile sub_udp_free.ini";
+}
 elsif ($test->flag('multicast')) {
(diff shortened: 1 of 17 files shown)


tests/DCPS/Messenger/run_test.pl
```diff
@@ -37,10 +37,18 @@
     $pub_opts .= " -DCPSConfigFile pub_udp.ini";
     $sub_opts .= " -DCPSConfigFile sub_udp.ini";
 }
+elsif ($test->flag('udp_free')) {
```
mitza-oci, line 40: What is this one for?

simpsont-oci: Ah, strictly speaking this doesn't need to be here. I added a test case to help me debug why "./run_test.pl udp" wasn't working on my machine but "./run_test.pl default_udp" was. After invesitaging, it turned out that the problem was that Ubuntu only maps "localhost" to 127.0.0.1 and not ::1 (which is "ip6-localhost") and I had ipv6 enabled. Once I added "localhost ::1" to my /etc/hosts file, the normal udp test worked as expected. I added the NetworkAddress test for it in a different PR, but kept this test here. I don't know that it adds much and I didn't add it to dcps_tests.lst, so if you want me to remove it I'm fine with that.

mitza-oci: It's fine to keep it, maybe just add a comment here or in the ini files.

mitza-oci (approved):

PR #2733, opened 2021-06-11
Add NetworkAddress Test For localhost

simpsont-oci:


tests/unit-tests/dds/DCPS/transport/framework/NetworkAddress.cpp
```diff
@@ -101,6 +101,22 @@ TEST(network_address_test, choose_single_coherent_address_ipv6)
...
+TEST(network_address_test, choose_single_coherent_address_localhost)
+{
+  //ScopedDebugLevels sdl(6); // Uncomment for greater debug levels
+
+  ACE_INET_Addr addr1 = choose_single_coherent_address("localhost:5200", false);
+  ACE_INET_Addr addr2 = choose_single_coherent_address("localhost:5200", true);
+  EXPECT_NE(addr1, ACE_INET_Addr());
+  EXPECT_NE(addr2, ACE_INET_Addr());
+  EXPECT_EQ(addr1, addr2);
+  EXPECT_EQ(addr2, addr1);
```
mitza-oci, line 113: Are both needed?

simpsont-oci: Probably not, but that's the pattern in the rest of the file. I suppose it makes more sense to check that sort of thing in the unit test for ACE_Inet_Addr, but I think when I first wrote these tests I was losing faith in how well written ACE_Inet_Addr was.

mitza-oci: Symmetry of `==` should be a given, but maybe we shouldn't assume too much.

mitza-oci (approved):

PR #2737, opened 2021-06-15
Support Shared / Dynamic Libraries For Google Test

simpsont-oci:


configure
```diff
@@ -1352,6 +1352,10 @@ if (exists $opts{'gtest'} || $tests) {
         "default install dir '$sys_dir', please pass a correct version of Google Test to --gtest\nStopped";
     }
   }
+  # Check for existence of googletest shared / dynamic libraries
+  if (-d File::Spec->catdir($opts{'gtest'}, 'bin')) {
+    push_path(\%targetEnv, $opts{'gtest'} . $slash . 'bin');
```
mitza-oci, line 1357: Should this be for Windows only?  Does the equivalent for `LD_LIBRARY_PATH` need to be added for Linux?

Or, if that gets too complicated... could we just go with archive/static libs for gtest?

simpsont-oci: It already defaults to static, and if we build it as part of configure it's using the default / static. So this is really for when people pass in --gtest on the command line or if, for some reason, tests/googletest already contains the non-static variety. I just checked, and for Linux it looks like the .so's go into "lib" not "bin" which I guess is what I would have expected. I suppose, in that case, I need to add a special check for the .so file name to support shared libraries on Linux?

mitza-oci: It should probably be similar to Qt5 with something like `  push_libpath(\%targetEnv, $qt_path);` but with `$qt_path` replaced by the relevant gtest directory.  `push_libpath()` knows about the `PATH` vs. `LD_LIBRARY_PATH` distinction (plus the one that macOS uses).

configure
```diff
@@ -1339,23 +1344,31 @@ if (exists $opts{'gtest'} || $tests) {
...
     if (-f File::Spec->catfile($sm_src, $gtest_sanity)) {
       # Check for existing build
-      $build_gtest = ! -d File::Spec->catdir($sm_dir, 'build/install');
+      $build_gtest = ! -d File::Spec->catdir($sm_dir, 'build' . $slash . 'install');
       $opts{'gtest'} = $sm_dir;
       # No need to set env. variable with submodule.
+      # Check for existence of googletest shared / dynamic libraries
+      if ($is_windows && -d File::Spec->catdir($opts{'gtest'}, 'build' . $slash . 'install' . $slash . 'bin')) {
+        push_path(\%targetEnv, $opts{'gtest'} . $slash . 'build' . $slash . 'install' . $slash . 'bin');
+      } elsif (!$is_windows && -f File::Spec->catfile($opts{'gtest'}, 'build' . $slash . 'install' . $slash . 'lib' . $slash . 'libgtest.so')) {
```
iguessthislldo, line 1353: This can't help you right now, but [I have a function that joins path elements in my other PR](https://github.com/objectcomputing/OpenDDS/blob/24693f9a4de65ef7b54fc41463f1bca7808a40d5/configure#L1685).

mitza-oci: You can also get away with `/` everywhere since Perl (or libc?) on Windows is fine with it.

simpsont-oci: It's just a shame that the path stored in `opts{'gtest'}` is slightly different when we're planning on building it ourselves, so it requires the `'build/install'` prefix and the logic here can't be combined with the other two places this check needs to occur. I looked at changing that assumption, but it would require "unwinding" it when we do the actual build and it felt more complex than necessary for the change I was trying to make.

iguessthislldo: Actually [`catdir` and `catfile` are basically doing what my function does](https://perldoc.perl.org/File::Spec#catdir). So you could do this here and the other places:

```suggestion
      } elsif (!$is_windows && -f File::Spec->catfile($opts{'gtest'}, 'build', 'install', 'lib', 'libgtest.so')) {
```

Or we could just use `/` like Adam suggested.

To simplify the checks would it help create a variable separate from `$opts{'gtest'}` to store the install location? Like this simplified code:

```perl
my $gtest_root;
if (-f File::Spec->catfile($sm_src, $gtest_sanity)) {
  $gtest_root = "$sm_dir/build/install";
}
elsif (-f File::Spec->catfile($sys_dir, $gtest_sanity)) {
  $gtest_root = $sys_dir;
}
# ...

# Check for existence of googletest shared / dynamic libraries
if ($is_windows && -d "$gtest_root/bin") {
  # ...
}
elif (!$is_windows && -f "$gtest_root/lib/libgtest.so") {
  # ...
}
```

iguessthislldo (approved):

PR #2740, opened 2021-06-17
java: installation of jar files

mchalain:
This patch install the jar files into INSTALL_LIBDIR directory instead the `lib` directory.
This allows to be uniform with the rest of the installation of OpenDDS libraries.
And it removes an error when the installation uses the DESTDIR variable and with the `lib` directory doesn't exist. This is particularly true for the packaging of OpenDDS. The rpm packaging renames the opendds_dcps.jar file to $(DESTIDR)$(INSTALL_PREFIX)/lib, and it is impossible to copy the other files inside the directory.

### DIFF
--- a/java/dds/dcps_java.mpc
+++ b/java/dds/dcps_java.mpc
@@ -126,8 +126,8 @@ project: idl2jni, javah, dcpslib, optional_jni_check, dcps_java_optional, dcps_t
 
   verbatim(gnuace, postinstall) {
 "	cp $(DDS_ROOT)/lib/OpenDDS_DCPS.jar $(DESTDIR)$(INSTALL_PREFIX)/$(INSTALL_LIB)"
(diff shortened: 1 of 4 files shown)


mitza-oci:
This looks good, running through regression testing now.  Is there something that could be added to the GitHub Actions configs to have it run a `make install` with a different value for `INSTALL_LIB`?

iguessthislldo (changes requested):
There might be more places where we should do this, but this should be fine for now as long as it works for your use case.

iguessthislldo:
> Is there something that could be added to the GitHub Actions configs to have it run a `make install` with a different value for `INSTALL_LIB`?

Yea, as far as I can tell we should be able to define it like we define [`DESTDIR`](https://github.com/objectcomputing/OpenDDS/blob/d7f423514b3d58f1fb74f5a89cdb0062ee722c0c/.github/workflows/build_and_test.yml#L304)

mchalain:
Yes 
# make DESTDIR=$PWD/package INSTALL_PREFIX=/usr INSTALL_LIB=/lib/x86_64 install

The prefix has to be set with configure first, but the libraries' directory may be changed only during the installation.
Here the command to install on Debian.

iguessthislldo (approved):

PR #2744, opened 2021-06-17
Don't use get_host_name, use get_host_addr as alternative

jwillemsen:
On Windows with a broken DNS setup the usage of `get_host_name()` could take seconds to return. Use `get_host_addr()` as alternative which is fast and tells more as it gives the concrete network interface used when multiple NICs are installed. 

We see it happen often on Windows that DNS is not setup correctly, when we stay away of using hostnames OpenDDS does work without problems when DNS is broken but the logging of the hostname is problematic.

### DIFF
--- a/dds/DCPS/transport/framework/NetworkAddress.cpp
+++ b/dds/DCPS/transport/framework/NetworkAddress.cpp
@@ -514,8 +514,8 @@ ACE_INET_Addr tie_breaker(const T& addrs, const String& name)
       }
     }
   }
-  VDBG((LM_DEBUG, "(%P|%t) choose_single_coherent_address(list) - Choosing Address %C:%d (%C)\n",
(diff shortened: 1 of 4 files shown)


simpsont-oci (approved):
In general, I'm fine with the short-term change to avoid get_host_name() for performance reasons, but the loss of these details in the logging is unfortunate, since it has been helpful in debugging some of the issues we were having with getaddrinfo() in the GitHub Actions virtual runner environments. Perhaps we need to look to see if we can't make get_host_name() more performant in poorly configured DNS scenarios or provide our own variant.

jwillemsen:
I don't think we can create our own variant, at the moment the get_host_name() takes long from the OS it is hard to workaround that. 

I see that the tcp transport does advertise with the fqdn, in how far would it be an option to add some configuration setting to let the tcp transport advertise using a ip address, similar to the ORBDottedDecimalAddress which TAO has?

PR #2749, opened 2021-06-18
XTypes: fixed XCDR2 encoding used for TypeObject

mitza-oci:


tests/DCPS/UnitTests/TypeObject.cpp
```diff
@@ -0,0 +1,115 @@
...
+    return false;
+  }
+  return strm.skip(end_of_stru - strm.pos());
+}
+
+template<typename T, size_t N>
+void check_encode(const T& object, const Encoding& enc, const unsigned char (&expected)[N])
+{
+  size_t size = 0;
+  serialized_size(enc, size, object);
```
iguessthislldo, line 56:
```suggestion
  const size_t size = serialized_size(enc, object);
```

jrw972 (approved):

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -1959,25 +1988,35 @@ bool operator>>(Serializer& strm, XTypes::Sequence<T>& seq)
       return false;
     }
   }
-  return true;
+  return (total_size == 0) ? true : strm.skip(end_of_seq - strm.pos());
```
iguessthislldo, line 1991: Small simplification if you want to take it.

```suggestion
  return (total_size == 0) || strm.skip(end_of_seq - strm.pos());
```

iguessthislldo (approved):

dds/DCPS/XTypes/TypeObject.h
sonndinh, line 1991: I'm not totally sure if this function handles the cases when the received data has a DHEADER but we skip the DHEADER, and when the received data doesn't have a DHEADER but we still read DHEADER. The two other cases (received data has DHEADER & we read DHEADER, and received data doesn't have DHEADER & we skip it) are handled correctly though.

dds/DCPS/Service_Participant.h
```diff
@@ -434,6 +434,11 @@ class OpenDDS_Dcps_Export Service_Participant {
   DDS::Duration_t bit_autopurge_disposed_samples_delay() const;
   void bit_autopurge_disposed_samples_delay(const DDS::Duration_t& duration);
 
+  enum TypeObjectEncoding { Encoding_Normal, Encoding_WriteOldFormat, Encoding_ReadOldFormat };
```
sonndinh, line 437: `Encoding_ReadOldFormat` seems not used.

dds/DCPS/RTPS/RtpsRpc.idl
```diff
@@ -38,14 +38,14 @@ enum RemoteExceptionCode_t
 
 @final
 struct RequestHeader {
-  SampleIdentity request_id;
-  InstanceName instance_name;
+  SampleIdentity requestId;
+  InstanceName _instanceName;
```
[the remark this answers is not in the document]
mitza-oci: It's not a typo, IDL doesn't allow identifiers that only differ by case.  The leading underscore is treated like an escape, it doesn't appear in the generated code.

I think it didn't exist at the time I wrote the code, but there may be a spec issue for this now.  I'll look for it.

dds/DCPS/Service_Participant.h
mitza-oci: It is indirectly, since there is a state that is ` != Service_Participant::Encoding_Normal` and also `!= Service_Participant::Encoding_WriteOldFormat`

dds/DCPS/XTypes/TypeObject.h
mitza-oci: It turns out that deserializing with this "skip sequence DHEADER" mode wasn't needed.  I'll take it out and simplify this function.

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -1949,6 +1963,19 @@ bool operator<<(Serializer& strm, const XTypes::Sequence<T>& seq)
...
+  }
+  // special case for compatibility with older versions that encoded this
+  // sequence incorrectly - if the DHeader was read as a 0, it's an empty
+  // sequence although it should have been encoded as DHeader (4) + Length (0)
+  if (total_size == 0) {
+    seq.length(0);
+    return true;
+  }
+
+  const size_t end_of_seq = strm.pos() + total_size;
```
sonndinh, line 1978: A couple of questions I have with this function.
(1) When user provides `-DCPSTypeObjectEncoding ReadOldFormat` option, does that mean this function should skip reading DHEADER, because from my understanding, the old format is the incorrect one that doesn't have DHEADER? If so, this function still reads DHEADER regardless of the encoding type given to `-DCPSTypeObjectEncoding`.

(2) In case the received data doesn't have DHEADER and the sequence has a length greater than 0, it seems to me it's possible that the function could incorrectly construct `seq` and still return true. This's because `total_size` is actually the length of the sequence, and `length` is read from the first 4 bytes of the first element of the sequence, and so on. So seems like the function could read wrong data into `seq` but still returns true?

mitza-oci: (1) `ReadOldFormat` means that we need to be compatible with both peers that are correct and peers that use the wrong encoding.  In this mode (and actually in all modes since there's no drawback) it will allow a 0 in this position in the stream.

(2) Fortunately for us, this case does not happen.  The version that has this bug always sends the `sequence<TypeIdentfierWithSize> dependent_typeids;` with length 0, and that's the only sequence-of-struct in `TypeInformation`.

sonndinh (approved):

PR #2673, opened 2021-05-18
tests/DCPS/UnitTests does not conform to guidelines

jrw972:
Problem
-------

* Unit tests under /tests/DCPS/UnitTests do not
  /docs/internal/unit_tests.rst.

* /tests/DCPS/UnitTests contains an integration test.

* /tests/DCPS/UnitTests contains a stress test.

Solution
--------

* Move unit tests to /tests/unit-tests.

* Move stress test to /tests/stress-tests.

* Move integratin tests to /tests/DCPS/BitDataReader.

* Provide a generic run_test.pl for launching unit and stress tests.

PR #2752, opened 2021-06-19
Avoid ACE_Message_Block Allocation During Marshalling of RTPS Parameter

simpsont-oci:


simpsont-oci:
Not entirely sure about the tests/DCPS/XTypes/run_test.pl failure on test_m10_o1d0_sec but the rest looks reassuring. Going to mark this as ready.

dds/DCPS/Serializer.h
```diff
@@ -669,25 +669,31 @@ class OpenDDS_Dcps_Export Serializer {
 
   void set_construction_status(ConstructionStatus cs);
 
+  struct OpenDDS_Dcps_Export SavePoint
+  {
```
iguessthislldo, line 673:
```suggestion
  struct OpenDDS_Dcps_Export SavePoint {
```

dds/DCPS/Serializer.h
```diff
@@ -669,25 +669,31 @@ class OpenDDS_Dcps_Export Serializer {
...
+    explicit SavePoint(Serializer& ser);
+
+    void restore(Serializer& ser) const;
+
+    const size_t pos_;
+    ACE_Message_Block* current_;
```
iguessthislldo, line 679:
```suggestion
    ACE_Message_Block* const current_;
```

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/Serializer.h
```diff
@@ -669,25 +669,30 @@ class OpenDDS_Dcps_Export Serializer {
...
+  struct OpenDDS_Dcps_Export SavePoint {
+    explicit SavePoint(Serializer& ser);
+
+    void restore(Serializer& ser) const;
+
+    const size_t pos_;
+    ACE_Message_Block* const current_;
+    char* const rd_ptr_;
+    char* const wr_ptr_;
+  };
```
mitza-oci, line 681: While this does preserve the functionality that was in `peek()` before this PR, I don't think this is sufficient to save and restore the Serializer's state.  The alignment support is not considered here (like `align_rshift_`).

I think we should use a different design for generalized `peek`.  Construct a new Serializer object using a duplicated message block.  These are both lightweight operations -- just updating reference counts and other scalars.

Before this PR, `peek()` is only used for the XCDR2 `LC > 4` case in `Serializer::read_parameter_id`.  In that specific case I don't know if we need a new Serializer object or if we should just have a peek_uint32 as a fundamental operation on Serializer.

mitza-oci (changes requested):

dds/DCPS/Serializer.h
simpsont-oci: Gotcha. I saw that this was only used in that one place. Do you want me to nuke the template for now and just make the a uint32 version? Or leave the template and add a comment / make a Jira ticket? Or try to fix the template and also add the uint32 version?

simpsont-oci: I mean, could the SavePoint just save/restore `align_rshift_` as well and that fixes the problem?

simpsont-oci: Looking at Serializer, it might also want to grab `align_wshift_`, `good_bit_`, `encoding_`, `swap_bytes_` ... anything else?

simpsont-oci: and `construction_status_` ... I think that's it?

mitza-oci: `duplicate()` itself has an allocation so if that needs to be avoided, the `message_block_allocator_` would be needed.

I think the align_rshift_ approach is fine.  There is a unit test for Serializer which should be extended for this, including the corner cases of advancing the message block chain to a block that has a different alignment.

dds/DCPS/Serializer.h
```diff
@@ -669,25 +673,51 @@ class OpenDDS_Dcps_Export Serializer {
...
+    const size_t wblock_;
+  };
+
+  template <typename T>
+  bool peek_helper(ACE_Message_Block* const block, T& t)
+  {
+    bool result = false;
+    char* const rd_ptr = block->rd_ptr();
+    if (block->cont()) {
+      result = peek_helper(block->cont(), t);
```
mitza-oci, line 696: This makes the complexity of the peek `O(n)` for n message blocks.
Since we expect this to be mostly used for a peek of a basic type, should those have a specialization/overload so that they only need to look `max_padding + n_bytes` ahead?

mitza-oci (changes requested):

tests/DCPS/UnitTests/Serializer.cpp
```diff
@@ -341,6 +343,186 @@ TEST(serializer_test, Serializer__swap_bytes_endianness)
   EXPECT_EQ(ser.endianness(), ENDIAN_NONNATIVE);
 }
 
+TEST(serializer_test, Serializer__align_context_basic_reference)
```
mitza-oci, line 346:
```suggestion
TEST(serializer_test, Serializer_align_context_basic_reference)
```
(and others)
Identifiers containing a double-underscore are reserved.

simpsont-oci: Should the other tests above also not have double-underscore?

mitza-oci: Nothing in the entire repo "should" have it.  But for this PR we can limit the scope to files or functions you're already changing.

mitza-oci:
Merge conflict from #2673

dds/DCPS/Serializer.cpp
```diff
@@ -381,6 +381,26 @@ Serializer::ScopedAlignmentContext::restore(Serializer& ser) const
...
+bool
+Serializer::peek(ACE_CDR::ULong& t)
+{
+  // save
+  const size_t rpos = rpos_;
+  const unsigned char align_rshift = align_rshift_;
+  ACE_Message_Block* const current = current_;
+
+  // read
+  if (!peek_helper(current_, 2 * sizeof (ACE_CDR::ULong), t)) {
```
mitza-oci, line 393: `uint32_cdr_size` (constant in Serializer.h) instead of sizeof

dds/DCPS/Serializer.h
```diff
@@ -688,14 +688,15 @@ class OpenDDS_Dcps_Export Serializer {
...
-  bool peek_helper(ACE_Message_Block* const block, T& t)
+  bool peek_helper(ACE_Message_Block* const block, size_t bytes, T& t)
   {
     bool result = false;
     char* const rd_ptr = block->rd_ptr();
-    if (block->cont()) {
-      result = peek_helper(block->cont(), t);
-    } else {
+    const size_t length = block->length();
+    if (!block->cont() || (block != 0 && bytes <= length)) {
```
mitza-oci, line 696: The `block != 0` expression doesn't look right because `block` was just dereferenced.

mitza-oci (approved):

PR #2758, opened 2021-06-21
Use empty to check whether a std::string has content or not

jwillemsen:


dds/DCPS/security/SSL/Certificate.cpp
```diff
@@ -481,7 +481,7 @@ X509* Certificate::x509_from_pem(const std::string& path,
 
   BIO* filebuf = BIO_new_file(path.c_str(), "r");
   if (filebuf) {
-    if (password != "") {
+    if (!password.empty ()) {
```
iguessthislldo, line 484:
```suggestion
    if (!password.empty()) {
```

dds/DCPS/transport/tcp/TcpInst.cpp
```diff
@@ -41,7 +41,7 @@ OpenDDS::DCPS::TcpInst::load(ACE_Configuration_Heap& cf,
   std::string local_address;
   GET_CONFIG_STRING_VALUE(cf, trans_sect, ACE_TEXT("local_address"), local_address);
 
-  if (local_address != "") {
+  if (!local_address.empty ()) {
```
iguessthislldo, line 44:
```suggestion
  if (!local_address.empty()) {
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #2773, opened 2021-06-27
Attempt Coverity Issue Fix For Serializer

simpsont-oci:


dds/DCPS/Serializer.cpp
```diff
@@ -328,6 +328,7 @@ Serializer::Serializer(ACE_Message_Block* chain, const Encoding& encoding)
   , align_rshift_(0)
   , align_wshift_(0)
   , pos_(0)
+  , construction_status_(ConstructionStatusUndefined)
```
iguessthislldo, line 331:
```C++
  , construction_status_(ConstructionStatusSuccessful)
  , align_rshift_(0)
  , align_wshift_(0)
  , pos_(0)
```

Correct order. Also I don't think we need a new enum member for this.

simpsont-oci: Wasn't sure what the default should be, since some places seemed to explicitly set success and others check against success. This PR was more of a reminder to myself to ask folks about this tomorrow. :)

dds/DCPS/Serializer.h
```diff
@@ -670,6 +670,7 @@ class OpenDDS_Dcps_Export Serializer {
   bool write_delimiter(size_t size);
 
   enum ConstructionStatus {
+    ConstructionStatusUndefined,
     ConstructionSuccessful,
     ElementConstructionFailure,
     BoundConstructionFailure
```
mitza-oci, line 676: @iguessthislldo @ClaytonCalabrese is there an enumerator from the existing 3 options that should be used as a default value, or should that be a new "Undefined" state?

iguessthislldo: I don't think there should be an "undefined" status unless we change the code that checks the status to error when it encounters that status. Otherwise from my understanding `ConstructionSuccessful` is basically the default state.

simpsont-oci: There are lots of places that seem to set the construction status to ConstructionSuccessful... are those strictly unnecessary? I'm more inclined to put an assert in get_construction_status() to check that the state being returned isn't undefined (since that should never happen, right?).

simpsont-oci: Or are there times where we'd temporarily mark the status as failed but then later may revert to successful?

iguessthislldo: > Or are there times where we'd temporarily mark the status as failed but then later may revert to successful?

Yes. From what I remember it's like an exception. It can be handled in a way that doesn't cause serialization to fail depending on how the user has annotated their IDL.

iguessthislldo (approved):

ClaytonCalabrese (approved):

PR #2775, opened 2021-06-28
Spdp: initialize is_application_participant_

mitza-oci:


simpsont-oci (approved):

mitza-oci:
This change is also in #2776.

PR #2776, opened 2021-06-28
RtpsRelay application participant purge is not robust

jrw972:
Problem
-------

If an RtpsRelay is restarted without changing network connectivity,
then a user has no incentive to reset the address and purge the old
application participant.

Solution
--------

When a new application participant is detected, remove all of the old
application participants.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -289,6 +289,7 @@ Spdp::Spdp(DDS::DomainId_t domain,
   , domain_(domain)
   , guid_(guid)
   , participant_discovered_at_(MonotonicTimePoint::now().to_monotonic_time())
+  , is_application_participant_(false)
   , tport_(DCPS::make_rch<SpdpTransport>(rchandle_from(this)))
   , initialized_flag_(false)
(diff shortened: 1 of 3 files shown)


mitza-oci (approved):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2628,14 +2648,29 @@ Spdp::SpdpTransport::write_i(WriteFlags flags)
...
+  }
+
+  for (DiscoveredParticipantIter pos = participants_.begin(), limit = participants_.end(); pos != limit;) {
+    if (pos != iter && pos->second.pdata_.participantProxy.opendds_rtps_relay_application_participant) {
+      if (DCPS::DCPS_debug_level) {
+        ACE_DEBUG((LM_DEBUG,
+                   ACE_TEXT("(%P|%t) Spdp::update_rtps_relay_application_participant - removing previous RtpsRelay application participant %C\n"),
+                   DCPS::LogGuid(pos->first).c_str()));
+      }
+      DiscoveredParticipantIter iter = pos++;
```
simpsont-oci, line 2670: It looks like you're shadowing `iter` here... maybe use a new name?

PR #2777, opened 2021-06-28
SampleLost test: induce a sample lost event

### LINKED ISSUE #2095: SampleLost test doesn't get any sample lost
The tests/DCPS/SampleLost doesn't see any samples lost happening, this looks incorrect, in some of my testing I would also expect to get a sample_lost but didn't got it, so tried this unit test, which also seems to not get a sample lost

mitza-oci:
Replaces #2096
May fix #2095

### DIFF
--- a/tests/DCPS/SampleLost/DataReaderListener.cpp
+++ b/tests/DCPS/SampleLost/DataReaderListener.cpp
@@ -10,7 +10,7 @@
 using namespace Messenger;
 
 DataReaderListenerImpl::DataReaderListenerImpl()
-  : num_reads_(0),
+  : num_data_available_(0),
     num_samples_lost_ (0),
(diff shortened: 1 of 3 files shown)


jwillemsen (approved):

tests/DCPS/SampleLost/DataReaderListener.cpp
```diff
@@ -24,37 +24,7 @@ DataReaderListenerImpl::~DataReaderListenerImpl ()
 void
 DataReaderListenerImpl::on_data_available(DDS::DataReader_ptr reader)
```
simpsont-oci, line 25:
```suggestion
DataReaderListenerImpl::on_data_available(DDS::DataReader_ptr /*reader/*)
```

PR #2779, opened 2021-06-29
Use ACE_OS::chdir instead of native windows call

jwillemsen:
* dds/DCPS/FileSystemStorage.cpp:

### DIFF
--- a/dds/DCPS/FileSystemStorage.cpp
+++ b/dds/DCPS/FileSystemStorage.cpp
@@ -127,7 +127,7 @@ int dds_chdir(const ACE_TCHAR* path)
     throw std::runtime_error("GetShortPathNameW failed.");
   }
 
-  ACE_OSCALL_RETURN(::_wchdir(spath), int, -1);
+  return ACE_OS::chdir(spath);
 }
 
 bool is_dir(const ACE_TCHAR* path)


mitza-oci:
I need to do some more research to see why we have it this way.  It is using the special `\\?\` prefix on Windows to avoid path length limits.  That may not have been compatible with the way ACE has it.  ACE also seems to be doing a wide-to-narrow conversion when wchar is enabled.

jwillemsen:
At the moment wchar is enabled there is wchar_t version which does exactly the same as OpenDDS used, see https://github.com/DOCGroup/ACE_TAO/blob/df4927c9dd70cd32aab0ea2e35a21d726e827c4e/ACE/ace/OS_NS_unistd.inl#L144

mitza-oci:
> At the moment wchar is enabled there is wchar_t version which does exactly the same as OpenDDS used, see https://github.com/DOCGroup/ACE_TAO/blob/df4927c9dd70cd32aab0ea2e35a21d726e827c4e/ACE/ace/OS_NS_unistd.inl#L144

Sorry, my previous comment wasn't correct.  It's when wchar is *not* enabled.

mitza-oci:
Replaced by #2791

PR #2784, opened 2021-07-01
Avoid Allocation of ACE_Message_Block in Rtps Parameter Deserialization

simpsont-oci:
Makes use of the ScopedAllocationContext from previous PR, but also adds "min read" to it in order to force reading of full parameter size when deserialized objects don't take full size (e.g. vendor id).

### DIFF
--- a/dds/DCPS/Serializer.cpp
+++ b/dds/DCPS/Serializer.cpp
@@ -364,11 +364,12 @@ Serializer::~Serializer()
 {
 }
 
-Serializer::ScopedAlignmentContext::ScopedAlignmentContext(Serializer& ser)
+Serializer::ScopedAlignmentContext::ScopedAlignmentContext(Serializer& ser, size_t min_read)
   : ser_(ser)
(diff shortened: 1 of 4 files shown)


ClaytonCalabrese (approved):

dds/DCPS/Serializer.cpp
```diff
@@ -364,11 +364,12 @@ Serializer::~Serializer()
 {
 }
 
-Serializer::ScopedAlignmentContext::ScopedAlignmentContext(Serializer& ser)
+Serializer::ScopedAlignmentContext::ScopedAlignmentContext(Serializer& ser, size_t min_read)
```
iguessthislldo, line 367: What about a unit test where min_read is tested?

simpsont-oci: Added. I didn't realize I needed min_read until I was having issues parsing our version parameter, which says it's 4 bytes but only reads two octets. So I didn't have a unit test for it at first, but good call, it should be there.

PR #2791, opened 2021-07-01
Removed usage of ACE_OSCALL_RETURN, that macro expands to just a return

jwillemsen:
* dds/DCPS/FileSystemStorage.cpp:


jwillemsen:
Alternative to #2779

mitza-oci (approved):

PR #2794, opened 2021-07-01
OpenDDS lacks intentional unit test coverage

jrw972:
Problem
-------

Intentional unit test coverages means that 1. for every unit, which is
a group of source files with the same prefix, there is a corresponding
set of unit tests and 2. the coverage report for the files belonging
to that unit are generated by executing the corresponding unit test.
The goal is that each line of code is covered by a unit that was
intending to execute the line of code.

Solution
--------

Adjust the names of the unit tests and a bash script to generate the
report.  The script can be converted to perl if we need support
non-bash platforms.


configure
```diff
@@ -179,6 +179,7 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'prefix=s', 'Installation prefix (none)',
     'install-origin-relative!', 'Install with RPATH relative to $ORIGIN (no)',
     'workspace=s', 'Custom MPC workspace file to copy and use (Use a builtin one)',
+    'coverage!', 'Enable code coverage (no)',
```
mitza-oci, line 182: Does this need to be its own option or could it just be `--features=<something>`?

tools/scripts/unit_test_coverage.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
iguessthislldo, line 1:
```suggestion
#!/bin/bash

set -e
```

configure
jrw972: We could do it that way.  What are the long-term maintenance tradeoffs between `--coverage` and `--features=dds_non_coverage=0`?

mitza-oci: I don't expect this to be used all that much outside of the maintainers/contributors -- I'd rather keep the configure script simpler and not have to know about every single way the configuration can be tweaked through MPC (or gnuace or config.h...).

mitza-oci (approved):

tests/unit-tests/dds/DCPS/security/SSL/DiffieHellman.cpp
```diff
@@ -14,10 +14,10 @@ using namespace OpenDDS::Security::SSL;
 using OpenDDS::DCPS::move;
 
 
-class DiffieHellmanTest : public ::testing::Test
+class dds_DCPS_security_SSL_DiffieHellman : public ::testing::Test
```
iguessthislldo, line 17: Couldn't put things like this in anonymous namespaces instead of renaming them?

iguessthislldo (changes requested):
I will also look into the script when or if I have time.

mitza-oci (approved):

docs/internal/unit_tests.rst
```diff
@@ -128,43 +128,31 @@ The use of a mock assumes that an interface exists for the stand-in.
...
   TEST(TestModule, TestSubmodule)
   {
   }
 
-Each of these tests contain evaluators.
+All tests in a unit test source file must have the same TestModule which is name of the unit under test with underscores, e.g., ``dds_DCPS_security_AccessControlBuiltInImpl``.
+This naming convention is required for intentional unit test coverage.
+The TestSubmodule can be any identifier, however, it should typical describe the class, function, or scenario being tested.
+
+Each tests contains evaluators.
```
iguessthislldo, line 155:
```suggestion
Each test contains evaluators.
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #2795, opened 2021-07-01
LogAddr

lij-oci:


dds/DCPS/LogAddr.h
```diff
@@ -0,0 +1,33 @@
...
+
+class OpenDDS_Dcps_Export LogAddr
+{
+public:
+  enum Fmt {IP, IP_PORT, HOST, HOST_PORT, IP_PORT_HOST};
+  explicit LogAddr(const ACE_INET_Addr& addr, Fmt fmt = IP_PORT);
+  const std::string& str() const { return addr_; }
+  const char* c_str() const { return addr_.c_str(); }
+private:
+  std::string addr_;
```
jrw972, line 25:
```suggestion
  OPENDDS_STRING addr_;
```

jrw972: I hate safety profile too.

dds/DCPS/LogAddr.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include <sstream>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+LogAddr::LogAddr(const ACE_INET_Addr& addr, Fmt fmt)
+{
+  std::stringstream s;
```
jrw972, line 14: Safety profile will probably choke on this.

jrw972 (changes requested):

jrw972:
The only issue I see is safety profile and you won't see it until it hits the scoreboard.
Basically, I would replace the `std::string` with an `ACE_TCHAR[OpenDDS::DCPS::AddrToStringSize]` and `ACE_OS::snprintf` into it (or use the functions on ACE_INET_Addr).
Everything else looks good and I think it makes the code look a little cleaner.

dds/DCPS/LogAddr.h
iguessthislldo:
```suggestion
  String addr_;
```

dds/DCPS/LogAddr.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <ace/INET_Addr.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+class OpenDDS_Dcps_Export LogAddr
+{
```
iguessthislldo, line 17:
```suggestion
class OpenDDS_Dcps_Export LogAddr {
```

dds/DCPS/LogAddr.h
```diff
@@ -0,0 +1,32 @@
...
+public:
+  enum Fmt {IP, IP_PORT, HOST, HOST_PORT, IP_PORT_HOST};
```
iguessthislldo, line 19: I'm a little nervous about using things like `IP` and `HOST` because they might collide with macros. How about using capitalized camel case like we do [here](https://github.com/objectcomputing/OpenDDS/blob/cd813f4e19416c05acec937b3de784663fb1e0ad/dds/DCPS/Serializer.h#L693-L695)?

iguessthislldo (changes requested):

dds/DCPS/LogAddr.h
lij-oci: Good catch, since these symbols are so common.

lij-oci:
Timothy Simpson made this comment in chat last Fri 3:01 PM:
"I would also recommend the versions of ACE_INET_Addr::get_host_name() and ACE_INET_Addr::get_host_addr() that take a parameter, since the version that doesn't use a parameter will inherently have race conditions with other threads."

I think this is very important. (It may have revealed some of the safety profile problems in the Thrasher test.) Changes are made accordingly in LogAddr.

dds/DCPS/LogAddr.cpp
```diff
@@ -0,0 +1,55 @@
+#include "DCPS/DdsDcps_pch.h"
+
+#include "LogAddr.h"
+#include "Definitions.h"
```
iguessthislldo, line 4:
```suggestion
#include "Definitions.h"

```

iguessthislldo (changes requested):
Other than this small thing it looks alright to me.

iguessthislldo (approved):

PR #2807, opened 2021-07-06
rtps_udp: address caching for locators / bundling

simpsont-oci:
Problem: We spend quite a bit of time calculating the lists of addresses needed to send each bundled message, but in reality these lists don't change all that often.

Solution: Cache them... both at the locator level and at the bundling level.

### DIFF
--- a/dds/DCPS/AddressCache.h
+++ b/dds/DCPS/AddressCache.h
@@ -118,10 +118,12 @@ public:
   void store(const Key& key, const AddrSet& addrs, const MonotonicTimePoint& expires = MonotonicTimePoint::max_value)
   {
     ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
-    std::pair<typename MapType::iterator, bool> ins = map_.insert(typename MapType::value_type(key, make_rch<AddressCacheEntry>(addrs, expires)));
-    if (!ins.second) {
-      ins.first->second->addrs_ = addrs;
-      ins.first->second->expires_ = expires;
+    RcHandle<AddressCacheEntry>& rch = map_[key];
+    if (rch) {
+      rch->addrs_ = addrs;
+      rch->expires_ = expires;
+    } else {
+      rch = make_rch<AddressCacheEntry>(addrs, expires);
     }
   }
 
--- a/dds/DCPS/AddressCache.h
+++ b/dds/DCPS/AddressCache.h
@@ -40,8 +40,7 @@ template <typename Key>
 class AddressCache {
 public:
 
-  typedef Key KeyType;
-  typedef OPENDDS_MAP(KeyType, AddressCacheEntry) MapType;
+  typedef OPENDDS_MAP(Key, AddressCacheEntry) MapType;
 
   AddressCache() {}
   virtual ~AddressCache() {}
--- a/dds/DCPS/AddressCache.h
+++ b/dds/DCPS/AddressCache.h
@@ -18,6 +18,7 @@
 #include "PoolAllocator.h"
 #include "TimeTypes.h"
 #include "GuidUtils.h"
+#include "RcObject.h"
 
 #include "ace/INET_Addr.h"
 
@@ -28,7 +29,8 @@ namespace DCPS {
 
 typedef OPENDDS_SET(ACE_INET_Addr) AddrSet;
 
-struct AddressCacheEntry {
+struct AddressCacheEntry : public virtual RcObject {
+
   AddressCacheEntry() : addrs_(), expires_(MonotonicTimePoint::max_value) {}
   AddressCacheEntry(const AddrSet& addrs, const MonotonicTimePoint& expires) : addrs_(addrs), expires_(expires) {}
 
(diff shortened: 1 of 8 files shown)


tests/unit-tests/dds/DCPS/AddressCache.cpp
```diff
@@ -0,0 +1,87 @@
+#include <dds/DCPS/AddressCache.h>
+
+#include <gtest/gtest.h>
+
+#include <cstring>
+
+using namespace OpenDDS::DCPS;
+
+struct TestKey
+{
```
iguessthislldo, line 10:
```suggestion
struct TestKey {
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,47 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
 typedef RcHandle<RtpsUdpTransport> RtpsUdpTransport_rch;
 typedef RcHandle<TransportClient> TransportClient_rch;
 
+struct LocatorCacheKey {
+  LocatorCacheKey(const RepoId& remote, const RepoId& local, bool prefer_unicast) : remote_(remote), local_(local), prefer_unicast_(prefer_unicast) {}
```
iguessthislldo, line 69:
```suggestion
  LocatorCacheKey(const RepoId& remote, const RepoId& local, bool prefer_unicast) 
    : remote_(remote)
    , local_(local)
    , prefer_unicast_(prefer_unicast)
  {
  }
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,47 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
...
+  bool operator<(const LocatorCacheKey& rhs) const {
```
iguessthislldo, line 70:
```suggestion
  bool operator<(const LocatorCacheKey& rhs) const
  {
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,47 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
...
+    return std::memcmp(this, &rhs, sizeof (LocatorCacheKey)) < 0;
```
iguessthislldo, line 71: This assumes that the struct is the same as if it were packed. Is there a cleaner way you could do this?

iguessthislldo (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
simpsont-oci: Hmm, I suppose I am assuming it's packed here, so I ought to make that explicit.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,47 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
...
+  const RepoId remote_;
+  const RepoId local_;
+  const bool prefer_unicast_;
+};
+typedef AddressCache<LocatorCacheKey> LocatorCache;
+
+struct BundlingCacheKey {
+  BundlingCacheKey(const RepoId& dst_guid, const RepoId& from_guid, const RepoIdSet& to_guids) : dst_guid_(dst_guid), from_guid_(from_guid), to_guids_(to_guids) {}
+  bool operator<(const BundlingCacheKey& rhs) const {
+    int r1 = std::memcmp(&dst_guid_, &rhs.dst_guid_, sizeof (RepoId));
```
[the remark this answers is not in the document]
simpsont-oci: For this one in particular, I was actually thinking I'd combine the memcmp for the first two guids, since the behavior should be the same. But your comment about packing below reminds me I need to make that explicit.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2339,41 +2363,62 @@ namespace {
 void
 RtpsUdpDataLink::build_meta_submessage_map(MetaSubmessageVecVecVec& meta_submessages, AddrDestMetaSubmessageMap& adr_map)
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, locators_lock_);
-  AddrSet addrs;
+  size_t cache_hits = 0;
+  size_t cache_misses = 0;
+  size_t addrset_min_size = -1;
```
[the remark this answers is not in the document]
iguessthislldo: Sorry I copied that from some where else and left a stray `)` at the end.

dds/DCPS/AddressCache.h
```diff
@@ -0,0 +1,143 @@
...
+class AddressCache {
+public:
+
+  typedef OPENDDS_MAP(Key, AddressCacheEntry) MapType;
+
+  AddressCache() {}
+  virtual ~AddressCache() {}
+
+  struct ScopedAccess
+  {
```
mitza-oci, line 49:
```suggestion
  struct ScopedAccess {
```

dds/DCPS/AddressCache.h
```diff
@@ -0,0 +1,143 @@
...
+    ScopedAccess(const ScopedAccess&);
+    ScopedAccess& operator=(const ScopedAccess&);
+  };
+
+  bool load(const Key& key, AddrSet& addrs) const
+  {
+    ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+    typename MapType::const_iterator pos = map_.find(key);
+    if (pos != map_.end()) {
+      if (MonotonicTimePoint::now() < pos->second.expires_) {
```
mitza-oci, line 80: Looks like this could be an `&&`? may be more readable that way

dds/DCPS/AddressCache.h
```diff
@@ -0,0 +1,143 @@
...
+    return false;
+  }
+
+  void store(const Key& key, const AddrSet& addrs, const MonotonicTimePoint& expires = MonotonicTimePoint::max_value)
+  {
+    ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+    std::pair<typename MapType::iterator, bool> ins = map_.insert(typename MapType::value_type(key, AddressCacheEntry(addrs, expires)));
+    if (!ins.second) {
+      ins.first->second.addrs_ = addrs;
+      ins.first->second.expires_ = expires;
```
mitza-oci, line 97: Is this equivalent to just using `operator[]`?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2339,41 +2363,62 @@ namespace {
...
         if (std::memcmp(&(it->dst_guid_.guidPrefix), &GUIDPREFIX_UNKNOWN, sizeof(GuidPrefix_t)) != 0) {
           adr_map[addrs][make_unknown_guid(it->dst_guid_.guidPrefix)].push_back(it);
         } else {
           adr_map[addrs][GUID_UNKNOWN].push_back(it);
         }
-        addrs.clear();
       }
     }
   }
+  VDBG((LM_DEBUG, "(%P|%t) RtpsUdpDataLink::build_meta_submessage_map() - Bundling Cache Stats: hits = %d, misses = %d, min = %d, max = %d\n", cache_hits, cache_misses, addrset_min_size, addrset_max_size));
```
mitza-oci, line 2421: use a different format specifier for inserting size_t

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
mitza-oci: Can it delegate to `GUID_t`'s `operator<`?  The compiler should inline.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,60 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
...
+    , to_guids_(to_guids)
+  {
+  }
+  bool operator<(const BundlingCacheKey& rhs) const
+  {
+    int r1 = std::memcmp(&dst_guid_, &rhs.dst_guid_, sizeof (RepoId));
+    if (r1 < 0) {
+      return true;
+    }
+    else if (r1 == 0) {
```
mitza-oci, line 99:
```suggestion
    } else if (r1 == 0) {
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -64,6 +65,60 @@ typedef RcHandle<RtpsUdpInst> RtpsUdpInst_rch;
...
+      int r2 = std::memcmp(&from_guid_, &rhs.from_guid_, sizeof (RepoId));
+      if (r2 < 0) {
+        return true;
+      }
+      else if (r2 == 0) {
```
mitza-oci, line 104:
```suggestion
      } else if (r2 == 0) {
```

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
simpsont-oci: I'm inclined to expand it to a single memcmp to avoid multiple function calls...

dds/DCPS/AddressCache.h
simpsont-oci: Hmm, yeah, I think I was doing it this way because store originally returned a bool letting you know if it was a new entry... but now the API is void so it doesn't matter.

mitza-oci (approved):

PR #2808, opened 2021-07-08
Fix failing unit tests on static builds

pmesnier:
Issue 1: SecuityConfigTest fails with an assertion 
Solution has a straight forward component and a surprise. The straight forward part was to explicitly include a header related to the ACE service object that will be "loaded" when needed. This is necessary because the ACE Service Configuration framework abstracts away those details and so the linker doesn't see the need to include the appropriate object files in the final executable.

The surprise was that the loader class for this service object was using a locally defined static instance of a std::string initialized with a literal, but when that string was referenced during loader initialization which is also done during static initialization phase, the std::string value was empty or null. 

Issue 2: NetworkAddressTest 
On the github acion host the tests involved with comparing IPV6 addresses literals are failing.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -3971,7 +3971,7 @@ jobs:
         cd OpenDDS
         . setenv.sh
         cd tests/DCPS/Messenger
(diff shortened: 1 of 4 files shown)


simpsont-oci (approved):

mitza-oci:
This one is still failing 
https://github.com/objectcomputing/OpenDDS/pull/2808/checks?check_run_id=3021362035

PR #2731, opened 2021-06-11
Topics returned from find_topic should be distinct Entities

mitza-oci:

PR #2809, opened 2021-07-08
Add another find_topic/delete_topic test and check the instance handles

jwillemsen:
Our AXCIOMA find_topic/delete_topic test still fails after  #2731, added some more test code, this seems to work as it should

### DIFF
--- a/tests/DCPS/FindTopic/findtopic.cpp
+++ b/tests/DCPS/FindTopic/findtopic.cpp
@@ -51,6 +51,24 @@ int sub_program(DDS::DomainParticipant* participant)
     ACE_ERROR_RETURN((LM_ERROR, "%N:%l: main() ERROR: create_datawriter 2 failed!\n"), EXIT_FAILURE);
   }
 
+  // Do now a find for topic "Movie Discussion List", that should return a new topic which we directly
+  // delete again which should work because this new topic is not used.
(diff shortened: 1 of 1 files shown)


mitza-oci:
I think I see an issue, but it won't cause the FindTopic test itself to fail (since it's abstracted away from the RTPS Discovery implementation).

jwillemsen:
I think I also need to update some of our code now that for one instance handle we get two different topic pointers.

mitza-oci:
#2810 may fix this - it includes the same changes to the test

jwillemsen:
Got further with AXCIOMA, had to update the administration we have to use the C++ topic pointer instead of the instance handle for the map we use to store additional data related to the C++ topic. 

After doing that I found that our AXCIOMA unit test with RTI Connext DDS failed, after analyzing what happens it looks the RTI implementation is not according to the specification. They return the same C++ pointer with create_topic/find_topic and the first delete_topic call you do always succeeds, they don't check whether the topic returned by find_topic is used by any other DDS entity (they can't because the pointers returned by create_topic/find_topic are the same), it just returns ok. Only when you do a second delete_topic there is a real check whether the topic is still used by any other DDS entity.

Created a support case with RTI to get their feedback on this.

mitza-oci:
Closing since these changes are in #2810

PR #2811, opened 2021-07-08
XTypes: DynamicType and CompleteTypeObject to DynamicType

ClaytonCalabrese:
This PR will add DynamicType and the associated classes to OpenDDS.  It will also add the conversion function for a CompleteTypeObject to a DynamicType.


dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -436,6 +436,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+      case TI_STRING16_LARGE:
+        dt->descriptor_->kind = TK_STRING16;
+        dt->descriptor_->name = "wstring";
+        dt->descriptor_->bound.length(1);
+        dt->descriptor_->bound[0] = ti.string_ldefn().bound;
+        type_identifier_to_dynamic(dt->descriptor_->element_type, TypeIdentifier(TK_CHAR16), dt_map);
+        break;
+      case TI_PLAIN_SEQUENCE_SMALL:
+        dt->descriptor_->kind = TK_SEQUENCE;
+        dt->descriptor_->name = "plain sequence";
```
ClaytonCalabrese, line 800: The TypeIdentifier only cases require us to decide on a name for each.  For now a bunch of these are just names I came up with on the fly.  Are there specific names for these cases that we want for each TI?

iguessthislldo: Is this 7.5.2.4.10?

> The fully qualified name of the type described by this descriptor. To be consistent, this name
must be a valid identifier for the given type kind, as defined elsewhere in this document.

I don't know what it means exactly by "as defined elsewhere in this document". It might mean IDL or it might mean the names like "Boolean" given in Table 3 in the "Type System" section of the spec.

iguessthislldo: Although "fully qualified name" seems to imply that this shouldn't just be `"sequence"`, but `"sequence<ELEMENT TYPE>"`.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -436,6 +436,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+    dt->descriptor_->kind = TK_ANNOTATION;
+    dt->descriptor_->name = cto.annotation_type.header.annotation_name;
+    dt->descriptor_->bound.length(0);
+    for (ulong i = 0; i < cto.annotation_type.member_seq.length(); ++i) {
+      DynamicTypeMember_rch dtm(new DynamicTypeMember, OpenDDS::DCPS::keep_count());
+      MemberDescriptor* md(new MemberDescriptor);
+      dtm->descriptor_ = md;
+      complete_annotation_member_to_member_descriptor(dtm->descriptor_, cto.annotation_type.member_seq[i], dt_map);
+      dtm->descriptor_->index = i;
+      dtm->descriptor_->id = i; //Clayton: I'm not entirely sure what this should be for annotation
```
iguessthislldo, line 580: Members of annotations don't really have ids like members of structs and unions, but the definition for this particular property 7.5.2.7.5 specifies it's for an aggregated types else it should be set to `MEMBER_ID_INVALID`. However the spec seems kinda split on if annotations are aggregated: 7.3.2.5 vs Table 8. Maybe this deserves a spec issue to clarify?

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+          return false;
+        }
+      }
+    }
+  } else {
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("DynamicTypeMembersByName: Member count is different\n")));
+    return false;
+  }
+  return true;
```
sonndinh, line 63:
```suggestion
  if (lhs.size() == rhs.size()) {
    for (DynamicTypeMembersByName::const_iterator it = lhs.begin(); it != lhs.end(); ++it) {
      if(rhs.find(it->first) != rhs.end()) {
        if (!test_equality(*lhs.find(it->first)->second->descriptor_, *rhs.find(it->first)->second->descriptor_, dt_ptr_pair)) {
          ACE_DEBUG((LM_DEBUG, ACE_TEXT("DynamicTypeMembersByName: Members are not the same\n")));
          return false;
        }
      } else {
        return false;
      }
    }
    return true;
  } else {
    ACE_DEBUG((LM_DEBUG, ACE_TEXT("DynamicTypeMembersByName: Member count is different\n")));
    return false;
  }
```
Similar changes for the other `test_equality` function.

sonndinh (changes requested):
I haven't finished yet but will look at the remaining files tomorrow.

dds/DCPS/XTypes/DynamicType.h
```diff
@@ -0,0 +1,49 @@
...
+  DDS::ReturnCode_t get_all_members_by_name(DynamicTypeMembersByName& member);
+  DDS::ReturnCode_t get_member(DynamicTypeMember_rch& member, const MemberId& id);
+  DDS::ReturnCode_t get_all_members(DynamicTypeMembersById& member);
+  unsigned long get_member_count();
+  DDS::ReturnCode_t get_member_by_index(DynamicTypeMember_rch& member, unsigned long index);
+  bool equals(const DynamicType& other);
+  DynamicTypeMembersByName member_by_name;
+  DynamicTypeMembersById member_by_id;
+  DynamicTypeMembersByIndex member_by_index;
+  TypeDescriptor* descriptor_;
```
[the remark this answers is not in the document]
ClaytonCalabrese: We are ignoring annotation for our implementation according to our outline.

Making TypeDescriptor not a pointer requires the complete implementation rather than a declaration which causes circular include issues.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,89 @@
...
+  //check pair seen
+  DynamicTypePtrPair this_pair = std::make_pair(&lhs, &rhs);
+  DynamicTypePtrPairSeen::iterator have_seen = dt_ptr_pair.find(this_pair);
+  if (have_seen == dt_ptr_pair.end()) {
+    dt_ptr_pair.insert(this_pair);
+    return
+      test_equality(*lhs.descriptor_, *rhs.descriptor_, dt_ptr_pair) &&
+      test_equality(lhs.member_by_name, rhs.member_by_name, dt_ptr_pair) &&
+      test_equality(lhs.member_by_id, rhs.member_by_id, dt_ptr_pair);
+  }
```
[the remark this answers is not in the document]
ClaytonCalabrese: DynamicTypePtrPairSeen tracks all comparisons we have already done when we are checking equality to make sure we don't recurse infinitely in strongly connected components and enums.

Enums will recurse infinitely because each member of the enum has the parents as its "type" meaning you would just go forever checking member->type.

SCC will recurse infinitely because A leads to B which leads to A.

By checking if we have already seen an equality comparison of two dynamic types, we stop the loop and return true, because if we already haven't returned false, then the last time we saw those two dynamic types, they were equal.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,89 @@
...
+  return true;
+}
+
+bool test_equality(const DynamicType_rch& lhs, const DynamicType_rch& rhs, DynamicTypePtrPairSeen& dt_ptr_pair)
+{
+  if (lhs.in() == rhs.in()) {
+    return true;
```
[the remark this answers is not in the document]
ClaytonCalabrese: Yes, and I think this section of code will return true if that is the case.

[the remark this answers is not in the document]
ClaytonCalabrese: It might on calls besides the initial when the equality check is looping.

dds/DCPS/XTypes/DynamicType.h
```diff
@@ -0,0 +1,49 @@
...
+};
+
+typedef std::pair<const DynamicType*,const DynamicType*> DynamicTypePtrPair;
+typedef OPENDDS_SET(DynamicTypePtrPair) DynamicTypePtrPairSeen;
+typedef OpenDDS::DCPS::RcHandle<DynamicType> DynamicType_rch;
+
+OpenDDS_Dcps_Export bool test_equality_i(const DynamicType& lhs, const DynamicType& rhs);
+OpenDDS_Dcps_Export bool test_equality(const DynamicType& lhs, const DynamicType& rhs, DynamicTypePtrPairSeen& dt_ptr_pair);
+OpenDDS_Dcps_Export bool test_equality(const DynamicType_rch& lhs, const DynamicType_rch& rhs, DynamicTypePtrPairSeen& dt_ptr_pair);
```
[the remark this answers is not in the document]
ClaytonCalabrese: I think the only one that needs to be an export is the _i for the test and whatever outside use we have.  The others don't need an exports.

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,89 @@
...
+{
+  descriptor = *descriptor_;
+  return DDS::RETCODE_OK;
+}
+
+bool DynamicTypeMember::equals(const DynamicTypeMember& other)
+{
+  DynamicTypePtrPairSeen dt_ptr_pair;
+  return test_equality(*this, other, dt_ptr_pair);
+}
```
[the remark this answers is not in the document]
ClaytonCalabrese: I don't think the way the spec defines DynamicTypeMember and equals are compatible.  It wants to know if the two members have the same containing DynamicType, which DynamicTypeMember has no knowledge of.  It also wants comparison of thier annotation which we are skipping.  I can add a comment saying that our equals is not spec consistent due to shortcuts?

dds/DCPS/XTypes/DynamicType.h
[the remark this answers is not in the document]
sonndinh: Yeah, the design doc is ignoring `AnnotationDescriptor` but I think we need to implement it. It captures the annotations applied to types or members. If we ignore it, `DynamicType` will not have the complete information of the type, and equality checking can return an incorrect result.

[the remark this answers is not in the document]
ClaytonCalabrese: What annotations are you concerned about? 

We have tryconstruct, extensibility, nested, key, optional, must understand, shared, and default label as flags.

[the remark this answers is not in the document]
sonndinh: Annotations like `@topic`, `@autoid`, `@hashid` haven't been captured. From 7.5.2.8.1, 7.5.2.8.4, 7.5.2.6.1, and 7.5.2.6.3 in the spec, I think all annotations applied to a type and its members need to be compared to determine equality. If we implement `AnnotationDescriptor` we have a general class that represents any annotation, so we don't have to enumerate any annotations currently supported or might be supported in the future. We can discuss further with the team.

dds/DCPS/XTypes/DynamicType.h
```diff
@@ -0,0 +1,47 @@
...
+namespace OpenDDS {
+namespace XTypes {
+
+class TypeDescriptor;
+
+class OpenDDS_Dcps_Export DynamicType : public OpenDDS::DCPS::RcObject
+{
+public:
+  void get_descriptor(TypeDescriptor& descriptor) const;
+  OPENDDS_STRING get_name();
```
iguessthislldo, line 20: Can you replace instances of `OPENDDS_STRING` with `DCPS::String`?

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+        dt->descriptor_->bound.length(0);
+        break;
+      case TK_INT8:
+        dt->descriptor_->kind = TK_INT8;
+        dt->descriptor_->name = "int";
+        dt->descriptor_->bound.length(0);
+        break;
+      case TK_UINT8:
+        dt->descriptor_->kind = TK_UINT8;
+        dt->descriptor_->name = "uint";
```
iguessthislldo, line 753:
```suggestion
        dt->descriptor_->name = "int8";
        dt->descriptor_->bound.length(0);
        break;
      case TK_UINT8:
        dt->descriptor_->kind = TK_UINT8;
        dt->descriptor_->name = "uint8";
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+  if (ti_found != dt_map.end()) {
+    dt = ti_found->second;
+    return;
+  }
+  DynamicType_rch dt_instantiation(new DynamicType, OpenDDS::DCPS::keep_count());
+  dt = dt_instantiation;
+  dt_map.insert(dt_map.end(), std::make_pair(ti , dt));
+  TypeDescriptor* td(new TypeDescriptor);
+  dt->descriptor_ = td;
+    switch (ti.kind()) {
```
sonndinh, line 688:
```suggestion

    switch (ti.kind()) {
```
This `switch` block's indentation is off by 2 spaces.

dds/DCPS/XTypes/TypeLookupService.cpp
mitza-oci: The spec is less than clear here... does the clause "To be consistent," mean this is not even required?  (Probably not.)

Typically "fully qualified" means to include all enclosing module names.

For names that are built-in to the type system like "Boolean" (as opposed to names that directly come from the user's IDL), it seems like we should be using the more generic names that Fred pointed out in Table 3.  The dynamic types API may also be used (in some future version) for types that aren't defined in IDL.

dds/DCPS/XTypes/TypeLookupService.cpp
mitza-oci: Annotation is not aggregated based on Figure 4

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+      md->try_construct_kind = USE_DEFAULT;
+    } else {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("Invalid TryConstruct Kind in complete_struct_member_to_member_descriptor\n")));
+    }
+  }
+  md->is_key = (cm.common.member_flags & (1 << 5));
+  md->is_optional = (cm.common.member_flags & (1 << 3));
+  md->is_must_understand = (cm.common.member_flags & (1 << 4));
+  md->is_shared = (cm.common.member_flags & (1 << 2));
+  md->is_default_label = 0;
```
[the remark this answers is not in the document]
ClaytonCalabrese: Index is set a level above.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+    } else {
+      md->try_construct_kind = DISCARD;
+    }
+  } else {
+    if (cm.common.member_flags & (1 << 1)) {
+      md->try_construct_kind = USE_DEFAULT;
+    } else {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("Invalid TryConstruct Kind in complete_union_member_to_member_descriptor\n")));
+    }
+  }
```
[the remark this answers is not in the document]
ClaytonCalabrese: Unfortunately, cm in each is a different type so I don't think they can be put into a single function.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
   }
 }
 
+void TypeLookupService::insert_dynamic_member(DynamicType_rch& dt, const DynamicTypeMember_rch& dtm)
+{
+  dt->member_by_index.insert(dt->member_by_index.end(), dtm);
+  dt->member_by_id.insert(dt->member_by_id.end(), std::make_pair(dtm->descriptor_->id , dtm));
+  dt->member_by_name.insert(dt->member_by_name.end(), std::make_pair(dtm->descriptor_->name , dtm));
```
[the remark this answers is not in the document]
ClaytonCalabrese: Right now the code will loop through the members in the order they appear and add them to the vector in that order, which just so happens to be the index order. This is why the code which sets the index appears a level above where the DynamicTypeMembers are handled.

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -38,6 +42,8 @@ class OpenDDS_Dcps_Export TypeLookupService : public virtual DCPS::RcObject {
   void update_type_identifier_map(const TypeIdentifierPairSeq& tid_pairs);
   bool complete_to_minimal_type_object(const TypeObject& cto, TypeObject& mto) const;
 
+  void complete_to_dynamic_i(DynamicType_rch& dt, const CompleteTypeObject&);
```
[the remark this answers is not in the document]
ClaytonCalabrese: Would it be acceptable to just remove the _i from the exported version or do you feel it needs a better way to identify it from the other functions of the same name?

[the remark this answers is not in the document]
sonndinh: For this specific class, we could just swap the names of the `complete_to_dynamic` and `complete_to_dynamic_i` methods. In general, I think `*_i` functions are private to the corresponding classes or containing scopes that provide detailed implementation, so they probably are not meant to be exported.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,119 @@
...
+bool test_equality_i(const DynamicType& lhs, const DynamicType& rhs)
+{
+//7.5.2.8.4 Operation: equals
+//Two types shall be considered equal if and only if all of their respective properties, as identified
+//in Table 54 above, are equal.
+
+//Note: We are comparing the TypeDescriptor even though the spec seems to say not to
+  DynamicTypePtrPairSeen dt_ptr_pair;
+  return test_equality(lhs, rhs, dt_ptr_pair);
+}
```
sonndinh, line 88: This is another example of `*_i` functions. `test_equality_i` is currently exported, but `test_equality` is the one that does all the work. So we should swap these two function names so that `test_equality` is exported and it calls `test_equality_i`, which does the actual work.

dds/DCPS/XTypes/TypeLookupService.cpp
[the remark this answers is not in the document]
sonndinh: The function can have a `MemberFlag` in/out argument which could be from either `CompleteStructMember` or `CompleteUnionMember`.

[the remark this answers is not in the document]
mitza-oci: Or... templates?

dds/DCPS/XTypes/TypeLookupService.cpp
ClaytonCalabrese: Table 51 says annotations are aggregated though...

dds/DCPS/XTypes/DynamicType.h
```diff
@@ -0,0 +1,50 @@
...
+
+  DynamicTypeMembersByName member_by_name;
+  DynamicTypeMembersById member_by_id;
+  DynamicTypeMembersByIndex member_by_index;
+  TypeDescriptor* descriptor_;
+};
+
+typedef OpenDDS::DCPS::RcHandle<DynamicType> DynamicType_rch;
+
+OpenDDS_Dcps_Export bool test_equality(const DynamicType& lhs, const DynamicType& rhs);
```
mitza-oci, line 41: Is this different from using the `equals` method?  Should it be the `operator==` for this type?

dds/DCPS/XTypes/DynamicType.h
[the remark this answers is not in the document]
mitza-oci: Use a smart pointer of some kind

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+void DynamicType::get_descriptor(TypeDescriptor& descriptor) const
+{
+  descriptor = *descriptor_;
+}
+
+DCPS::String DynamicType::get_name() const
+{
+  return this->descriptor_->name;
```
mitza-oci, line 26:
```suggestion
  return descriptor_->name;
```

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+TypeKind DynamicType::get_kind() const
+{
+  return this->descriptor_->kind;
+}
+
+DDS::ReturnCode_t DynamicType::get_member_by_name(DynamicTypeMember_rch& member, const DCPS::String& name)
+{
+  //We have a map of members in member_by_name which takes a string and gives 0 or 1 type DTMs
+  //this should use the string as the key and put the returned DTM into member.
```
mitza-oci, line 37:
```suggestion
```

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+void DynamicType::get_all_members(DynamicTypeMembersById& member)
+{
+  member = member_by_id;
+}
+
+ACE_CDR::ULong DynamicType::get_member_count()
+{
+  return member_by_name.size();
```
mitza-oci, line 70: see warnings from GitHub Actions builds

dds/DCPS/XTypes/DynamicType.h
```diff
@@ -0,0 +1,50 @@
...
+class OpenDDS_Dcps_Export DynamicType : public OpenDDS::DCPS::RcObject
+{
+public:
+  DynamicType();
+  ~DynamicType();
+
+  void get_descriptor(TypeDescriptor& descriptor) const;
+  DCPS::String get_name() const;
+  TypeKind get_kind() const;
+  DDS::ReturnCode_t get_member_by_name(DynamicTypeMember_rch& member, const DCPS::String& name);
```
mitza-oci, line 25: Any method that can be made in to a `const` member function probably should be one.  The only reason I could think of to not do so would be to strictly follow the mapping from the Annex C IDL, but we're not doing that either.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,128 @@
...
+TypeKind DynamicType::get_kind() const
+{
+  return this->descriptor_->kind;
+}
+
+DDS::ReturnCode_t DynamicType::get_member_by_name(DynamicTypeMember_rch& member, const DCPS::String& name)
+{
+  //We have a map of members in member_by_name which takes a string and gives 0 or 1 type DTMs
+  //this should use the string as the key and put the returned DTM into member.
+  DynamicTypeMembersByName::const_iterator pos = member_by_name.find(name);
```
mitza-oci, line 38:
```suggestion
  const DynamicTypeMembersByName::const_iterator pos = member_by_name.find(name);
```

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,128 @@
...
+  return test_equality(*this, other);
+}
+
+bool test_equality(const DynamicType& lhs, const DynamicType& rhs)
+{
+//7.5.2.8.4 Operation: equals
+//Two types shall be considered equal if and only if all of their respective properties, as identified
+//in Table 54 above, are equal.
+
+//Note: We are comparing the TypeDescriptor even though the spec seems to say not to
```
mitza-oci, line 94: Indent looks off.
Add more detail here, if it's deviating from the spec then we should have a reason for that

dds/DCPS/XTypes/MemberDescriptor.h
```diff
@@ -0,0 +1,60 @@
...
+  bool is_optional;
+  bool is_must_understand;
+  bool is_shared;
+  bool is_default_label;
+ };
+
+inline bool operator==(const UnionCaseLabelSeq& lhs, const UnionCaseLabelSeq& rhs)
+{
+  if (lhs.length() == rhs.length()) {
+    for (ACE_CDR::ULong i = 0 ; i < lhs.length() ; ++i) {
```
mitza-oci, line 43:
```suggestion
    for (ACE_CDR::ULong i = 0; i < lhs.length(); ++i) {
```

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -826,6 +826,7 @@ namespace XTypes {
 
   // ID of a type member
   typedef ACE_CDR::ULong MemberId;
+  const ACE_CDR::ULong MEMBER_ID_INVALID = UINT32_MAX;
```
mitza-oci, line 829: Does this need to be checked in the IDL compiler to make sure it isn't actually used as an ID?

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -432,6 +432,426 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+  dt->member_by_index.push_back(dtm);
+  if (dtm->descriptor_->id != MEMBER_ID_INVALID) {
+    dt->member_by_id.insert(std::make_pair(dtm->descriptor_->id , dtm));
+  }
+  dt->member_by_name.insert(std::make_pair(dtm->descriptor_->name , dtm));
+}
+
+void TypeLookupService::handle_tryconstruct_flags(MemberDescriptor*& md, MemberFlag mf)
+{
+  if (mf & (1 << 0)) {
```
mitza-oci, line 446: Instead of having expressions like `(1 << 0)` in the implementation, either use a named constant or just the plain integer literal that you'd get from evaluating that expression.

dds/DCPS/XTypes/TypeLookupService.cpp
mitza-oci: I thought annotation was out of scope?  We may need to have a design discussion outside of the PR review to figure out where the line gets drawn.

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+bool DynamicTypeMember::equals(const DynamicTypeMember& other)
+{
+//7.5.2.6.3 Operation: equals
+//Two members shall be considered equal if and only if they belong to the same type and all of
+//their respective properties, as identified in Table 52 above, are equal.
+
+//This spec implementation is not currently possible so I have done what I consider to be a deep compare
+//8/3/2021 Clayton Calabrese
```
mitza-oci, line 32:
```suggestion
```
This is why git exists

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,97 @@
...
+bool test_equality_i(const DynamicTypeMember& lhs, const DynamicTypeMember& rhs, DynamicTypePtrPairSeen& dt_ptr_pair)
+{
+  return test_equality_i(*lhs.descriptor_, *rhs.descriptor_, dt_ptr_pair);
+}
+
+bool test_equality_i(const DynamicTypeMembersByName& lhs, const DynamicTypeMembersByName& rhs, DynamicTypePtrPairSeen& dt_ptr_pair)
+{
+  if (lhs.size() == rhs.size()) {
+    for (DynamicTypeMembersByName::const_iterator it = lhs.begin(); it != lhs.end(); ++it) {
+      if(rhs.find(it->first) != rhs.end()) {
```
mitza-oci, line 56:
```suggestion
      if (rhs.find(it->first) != rhs.end()) {
```

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,97 @@
...
+        if (!test_equality_i(*lhs.find(it->first)->second->descriptor_, *rhs.find(it->first)->second->descriptor_, dt_ptr_pair)) {
+          ACE_DEBUG((LM_DEBUG, ACE_TEXT("test_equality_i: DynamicTypeMembersByName - Members are not the same\n")));
```
mitza-oci, line 58: Does this need to be logged?

tests/DCPS/UnitTests/CompleteToDynamicType/CompleteToDynamicType.cpp
```diff
@@ -0,0 +1,608 @@
...
+#include <dds/DCPS/XTypes/TypeLookupService.h>
+#include <dds/DCPS/XTypes/TypeObject.h>
+#include <dds/DCPS/XTypes/DynamicType.h>
+#include <dds/DCPS/XTypes/DynamicTypeMember.h>
+
+#include <gtest/gtest.h>
+
+using namespace OpenDDS;
+
+XTypes::TypeLookupService_rch tls(new XTypes::TypeLookupService, OpenDDS::DCPS::keep_count());
```
mitza-oci, line 21: use the `make_rch` function

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicType.cpp
ClaytonCalabrese: DynamicTypeMember
> 7.5.2.6.3 Operation: equals
> Two members shall be considered equal if and only if they belong to the same type and all of
> their respective properties, as identified in Table 52 above, are equal.

DynamicType

> 7.5.2.8.4 Operation: equals
> Two types shall be considered equal if and only if all of their respective properties, as identified
> in Table 54 above, are equal.

This is the way the spec defines "equals" for DynamicTypes and DynamicTypeMembers.

The way this is defined will lead to infinite loops which never resolve anything.  If the equality of two DynamicTypes is dependant on the equality of two DynamicTypeMembers, and the DynamicTypeMembers' equality is dependant on the equality of the two DynamicTypes, we have a very simple unending loop.

Additionally, even if this didn't force an infinite loop, a simple thought experiment of two different DynamicTypeMembers belonging to the same DynamicType would be "equal" according to the spec's definition of "equals", because the "properties" section does not include descriptors.

I made the assumption that the only reasonable way to determining "equals" would be to check the descriptors in a recursive way until we reach the very base of each type, which is what I use for my test.

We could either not implement equals, or we can overwrite the spec defined equals, and make a spec issue for clarification.

simpsont-oci: Without a deep dive into the spec, it seems like there are two issues here.

- One issue is that it seems like the spec is missing some properties in tables 52 and 54. Either that or they mean more than the table's "properties" when they say properties... possibly properties + results of certain operations.
- The other issue is that "the same type" in their definition can't mean equality through the equals method, as you point out. So what does it mean? Same name? Same descriptor? I don't have an answer here, since I'm not sure I have a full grasp of all the various "properties" to choose from.

dds/DCPS/XTypes/TypeObject.h
ClaytonCalabrese: > 7.2.2.4.4.4.4 Member IDs
> Each member of an aggregated type is uniquely identified within that type by an integer
> “member ID.” Member IDs are unsigned and have a range that can be represented in 28 bits:
> from zero to 268,435,455 (0x0FFFFFFF). (The full range of a 32-bit unsigned integer is not used
> in order to allow binary Data Representations the freedom to embed a small amount of meta-data
> into a single 32-bit field if they so desire.)

> The remaining part of the member ID range—from 0 to 268,402,687 (0x0FFFBFFF)—is
> available for use by application-defined types compliant with this specification.

According to the spec, UINT32_MAX will never be naturally used as a MemberID.

The only possible collision would occur with "binary Data Representations", but I have no clue what the spec is talking about with that.

dds/DCPS/XTypes/TypeLookupService.cpp
mitza-oci: Consensus seems to be to treat them as if they were `autoid(SEQUENTIAL)` and file a spec issue.

dds/DCPS/XTypes/TypeObject.h
iguessthislldo: > The only possible collision would occur with "binary Data Representations", but I have no clue what the spec is talking about with that.

It's talking about XCDR2 where there is the must understand flag and the length code in the 4 most significant bits of the parameter id, with the rest of the 28 bits being for the member id.

> Does this need to be checked in the IDL compiler to make sure it isn't actually used as an ID?

It doesn't look like we do:

https://github.com/objectcomputing/OpenDDS/blob/7a16ebb3052cb05473a5899612576a35799876c9/dds/idl/be_global.cpp#L988-L1014

`hash_member_name_to_id` won't break 28 bits, so I think the only thing to worry about at the moment would be if a user did something like `@id(0xffffffff)`.

To be safe we should probably add checks for this to `compute_id` and `write_parameter_id`.

dds/DCPS/XTypes/DynamicType.cpp
sonndinh: > The way this is defined will lead to infinite loops which never resolve anything. If the equality of two DynamicTypes is dependant on the equality of two DynamicTypeMembers, and the DynamicTypeMembers' equality is dependant on the equality of the two DynamicTypes, we have a very simple unending loop.

Ignoring the implementation details, I think the definition of "equals" for `DynamicTypeMember` in the spec makes sense. I interpret 2 `DynamicTypeMember` objects as equal if they refer to the same member of the same type. So it makes sense that checking the equality of 2 `DynamicTypeMember` objects involving checking that they represent members from the same type.

For implementation, maybe we can check that 2 `DynamicTypeMember` objects belong to the same type by keeping the `TypeIdentifier` of the containing type within each `DynamicTypeMember` object. If 2 such objects have the same containing type's `TypeIdentifier` then they belong to the same type. With that, we may solve the infinite loop issue.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -519,6 +519,7 @@ void TypeLookupService::complete_to_dynamic_i(DynamicType_rch& dt,
     for (ACE_CDR::ULong i = 0; i < dt->get_descriptor()->base_type->get_member_count(); ++i) {
       DynamicTypeMember_rch dtm;
       dt->get_descriptor()->base_type->get_member_by_index(dtm, i);
+      dtm->get_parent()= dt;
```
iguessthislldo, line 522:
```suggestion
      dtm->get_parent() = dt;
```

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+void DynamicType::get_all_members(DynamicTypeMembersById& member)
+{
+  member = member_by_id;
+}
+
+ACE_CDR::ULong DynamicType::get_member_count() const
+{
+  return CORBA::ULong(member_by_index.size());
```
mitza-oci, line 68: Use a consistent typedef for this type.  `ACE_CDR::ULong` has the advantage of not being tied to any specific IDL mapping spec.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+bool DynamicType::test_equality_i(const DynamicType& rhs, DynamicTypePtrPairSeen& dt_ptr_pair) const
+{
+  //check pair seen
+  DynamicTypePtrPair this_pair = std::make_pair(this, &rhs);
+  DynamicTypePtrPairSeen::const_iterator have_seen = dt_ptr_pair.find(this_pair);
+  if (have_seen == dt_ptr_pair.end()) {
+    dt_ptr_pair.insert(this_pair);
+    return
+      OpenDDS::XTypes::test_equality_i(*this->descriptor_.in(), *rhs.descriptor_.in(), dt_ptr_pair) &&
```
mitza-oci, line 111:
```suggestion
      test_equality_i(*descriptor_, *rhs.descriptor_, dt_ptr_pair) &&
```

ClaytonCalabrese: test_equality_i exists both in the class namespace as well as in OpenDDS::XTypes.  These calls specifically need the versions from the OpenDDS::XTypes namespace, which they won't find without the scoping.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -454,6 +473,393 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
   }
 }
 
+void TypeLookupService::complete_struct_member_to_member_descriptor(MemberDescriptor_rch& md,
```
mitza-oci, line 476: Could this return a `MemberDescriptor_rch`?

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -454,6 +473,427 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+    break;
+  case TK_BITSET:
+    td.kind = TK_BITSET;
+    td.name = cto.bitset_type.header.detail.type_name;
+    td.bound.length(0);
+    break;
+  // Collection TKs
+  case TK_SEQUENCE:
+    td.kind = TK_SEQUENCE;
+    if(cto.sequence_type.header.detail.present == true) {
```
iguessthislldo, line 657:
```suggestion
    if (cto.sequence_type.header.detail.present) {
```

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -454,6 +473,427 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+  case TK_ARRAY: {
+    td.kind = TK_ARRAY;
+    td.name = cto.array_type.header.detail.type_name;
+    td.bound = cto.array_type.header.common.bound_seq;
+    td.element_type = type_identifier_to_dynamic(cto.array_type.element.common.type, dt_map);
+  }
+  break;
+  case TK_MAP:
+    td.kind = TK_MAP;
+    if (cto.map_type.header.detail.present == true) {
```
iguessthislldo, line 675:
```suggestion
    if (cto.map_type.header.detail.present) {
```

tests/DCPS/UnitTests/CompleteToDynamicType/run_test.pl
```diff
@@ -0,0 +1,26 @@
...
+# -*- perl -*-
+use Sys::Hostname;
+
+use Env (DDS_ROOT);
+use lib "$DDS_ROOT/bin";
+use Env (ACE_ROOT);
+use lib "$ACE_ROOT/bin";
+use PerlDDS::Run_Test;
+use FileHandle;
+use Cwd;
```
iguessthislldo, line 14:
```suggestion
```

Not used

tests/DCPS/UnitTests/CompleteToDynamicType/run_test.pl
```diff
@@ -0,0 +1,26 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
+    & eval 'exec perl -S $0 $argv:q'
+    if 0;
+
+# -*- perl -*-
+use Sys::Hostname;
+
```
iguessthislldo, line 7:
```suggestion
```

Not used

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+namespace OpenDDS {
+namespace XTypes {
+
+DynamicType::DynamicType()
+{}
+
+DynamicType::~DynamicType()
+{}
```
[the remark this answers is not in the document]
ClaytonCalabrese: They are needed for Windows or it gets confused and tries to do some stuff when compiling headers that depend on types that are not fully defined.

For example, Windows was trying to inline the default destructor if we don't have it explicitly in the .cpp.

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -40,6 +44,9 @@ class OpenDDS_Dcps_Export TypeLookupService : public virtual DCPS::RcObject {
   void update_type_identifier_map(const TypeIdentifierPairSeq& tid_pairs);
   bool complete_to_minimal_type_object(const TypeObject& cto, TypeObject& mto) const;
 
+  typedef OPENDDS_MAP(TypeIdentifier, DynamicType_rch) DynamicTypeMap;
+  DynamicType_rch complete_to_dynamic(const CompleteTypeObject&, DynamicTypeMap& dt_map) const;
```
sonndinh, line 48: Can we remove the parameter `dt_map` in `complete_to_dynamic()` and instead have a private member of type `DynamicTypeMap` for caching the computed mappings instead? That way callers won't have to worry about passing a `dt_map` argument, and it can be completely internal to the implementation.

Another advantage is that multiple calls to `complete_to_dynamic()` may share some common mappings, and if so a later call doesn't have to recompute a common mapping again.

sonndinh (changes requested):

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -454,6 +473,425 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+    td.element_type = type_identifier_to_dynamic(*ti.map_ldefn().element_identifier, dt_map);
+    td.key_element_type = type_identifier_to_dynamic(*ti.map_ldefn().element_identifier, dt_map);
+    dt->set_descriptor(td);
+    break;
+  case TI_STRONGLY_CONNECTED_COMPONENT:
+    complete_to_dynamic_i(dt, get_type_object_i(ti).complete, dt_map);
+    break;
+  case EK_COMPLETE:
+    complete_to_dynamic_i(dt, get_type_object_i(ti).complete, dt_map);
+    break;
```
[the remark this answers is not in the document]
ClaytonCalabrese: I break that infinite loop by using the `DynamicTypeMap` which checks to see if I have already encountered the DynamicType. Recent changes made it so I had to move the creation of `DynamicTypeMap` to outside where I call into `complete_to_dynamic` so that the weak_rch don't destroy themselves early. Any time you are using `complete_to_dynamic` you will have to do the same.

In the spec issue document I think I already created an entry for the infinite looping problem.

dds/DCPS/XTypes/TypeLookupService.h
ClaytonCalabrese: Recent changes made it so I had to move the creation of DynamicTypeMap to outside where I call into complete_to_dynamic so that the weak_rch don't destroy themselves early. It was a compromise that had to be made so we don't have memory leaks, but also don't have failures due to early deletion where nothing held the rch.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -454,6 +473,425 @@ bool TypeLookupService::complete_to_minimal_type_object(const TypeObject& cto, T
...
+    td.name = cto.alias_type.header.detail.type_name;
+    td.bound.length(0);
+    td.base_type = type_identifier_to_dynamic(cto.alias_type.body.common.related_type, dt_map);
+    // The spec says that Alias DynamicTypes should have DynamicTypeMembers, but that leads to redundancy
+    break;
+  // Enumerated TKs
+  case TK_ENUM:
+    td.kind = TK_ENUM;
+    td.name = cto.enumerated_type.header.detail.type_name;
+    td.bound.length(0);
```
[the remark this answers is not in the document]
ClaytonCalabrese: That makes sense to me. I'll add a quick note to the spec issues document.

dds/DCPS/XTypes/TypeLookupService.cpp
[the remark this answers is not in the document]
ClaytonCalabrese: Edit: I reread what I put for the spec issue and that writeup I did about infinite recursion is about checking equality, not about what is discussed here.

We could discuss with team tomorrow about whether this is more of an implementation choice vs whether it is a spec issue.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -29,7 +46,7 @@ TypeLookupService::~TypeLookupService()
 void TypeLookupService::get_type_objects(const TypeIdentifierSeq& type_ids,
                                          TypeIdentifierTypeObjectPairSeq& types) const
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  //ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
```
iguessthislldo, line 49: Do you still need this commented out?

iguessthislldo (changes requested):

dds/DCPS/XTypes/TypeDescriptor.cpp
```diff
@@ -0,0 +1,43 @@
+#include <DCPS/DdsDcps_pch.h>
+
+#include "TypeDescriptor.h"
+
```
sonndinh, line 4:
```suggestion
```

dds/DCPS/XTypes/MemberDescriptor.cpp
```diff
@@ -0,0 +1,62 @@
+#include <DCPS/DdsDcps_pch.h>
+
+#include "MemberDescriptor.h"
+
```
sonndinh, line 4:
```suggestion
```

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,114 @@
...
+  //their respective properties, as identified in Table 52 above, are equal.
+
+  //In addition to what the spec says to compare, we will be comparing the MemberDescriptors of both
+  //DynamicTypeMembers. The spec seems to assume this is the case, despite not listing the MemberDescriptor
+  //as a property in table 52. If this were not the case, any two members within a type would be considered
+  //equal, regardless of whether they are actually the same member.
+  DynamicTypePtrPairSeen dt_ptr_pair;
+  return
+    test_equality_i(parent_.lock(), other.parent_.lock(), dt_ptr_pair) &&
+    OpenDDS::XTypes::test_equality_i(descriptor_, other.descriptor_, dt_ptr_pair);
```
sonndinh, line 60:
```suggestion
    test_equality_i(descriptor_, other.descriptor_, dt_ptr_pair);
```

sonndinh (commented):
A few minor comments and then I think it's ready to go.

dds/DCPS/XTypes/DynamicTypeMember.cpp
```diff
@@ -0,0 +1,114 @@
+#include <DCPS/DdsDcps_pch.h>
+
+#include "DynamicTypeMember.h"
+
```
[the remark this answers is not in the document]
ClaytonCalabrese: https://github.com/objectcomputing/OpenDDS/blob/master/docs/guidelines.md#includes

I think what I have is correct based off our docs.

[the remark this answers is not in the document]
sonndinh: Yeah, sorry I misremembered it. Please ignore this and similar comments.

dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -0,0 +1,146 @@
...
+{
+  //check pair seen
+  DynamicTypePtrPair this_pair = std::make_pair(this, &rhs);
+  DynamicTypePtrPairSeen::const_iterator have_seen = dt_ptr_pair.find(this_pair);
+  if (have_seen == dt_ptr_pair.end()) {
+    dt_ptr_pair.insert(this_pair);
+    return
+      OpenDDS::XTypes::test_equality_i(descriptor_, rhs.descriptor_, dt_ptr_pair) &&
+      OpenDDS::XTypes::test_equality_i(member_by_name, rhs.member_by_name, dt_ptr_pair) &&
+      OpenDDS::XTypes::test_equality_i(member_by_id, rhs.member_by_id, dt_ptr_pair);
```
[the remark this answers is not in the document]
ClaytonCalabrese: These unfortunately need to stay because for some reason the compiler can't find them without the `OpenDDS::XTypes::` even thought it is already in that namespace.

PR #2812, opened 2021-07-08
RtpsUdpDataLink leaks remote writers on failed associations

jrw972:
Problem
-------

An association for a local reader that times out is removed from the
`pending_` set of the TransportClient and not added to the
`data_link_index_`.  A subsequent call to `disassociate` when
discovery times out is effectively a no-op.  For RTPS, this means that
the reader has a pre-association writer and will send pre-association
acknacks forever since the writer may never associate and never be
removed.

Solution
--------

- Add a parameter to `stop_accepting_or_connecting` that indicates
  association failure.  RTPS will release its reservations when the
  association fails.

- Fix a bug in `RtpsUdpDataLink::disassociated` where the parameters
  to `release_reservations_i` were reversed.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -704,7 +704,7 @@ void
 RtpsUdpDataLink::disassociated(const RepoId& local_id,
                                const RepoId& remote_id)
 {
-  release_reservations_i(local_id, remote_id);
+  release_reservations_i(remote_id, local_id);
 
   ACE_GUARD(ACE_Thread_Mutex, g, locators_lock_);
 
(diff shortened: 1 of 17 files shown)


mitza-oci (approved):
After "lint" issues are fixed it should be good to go

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -704,7 +704,7 @@ void
 RtpsUdpDataLink::disassociated(const RepoId& local_id,
                                const RepoId& remote_id)
 {
-  release_reservations_i(local_id, remote_id);
+  release_reservations_i(remote_id, local_id);
```
simpsont-oci, line 707: wow... how long has this been wrong?

PR #2821, opened 2021-07-12
Test current behavior of calling create_topic with the same topic nam…

jwillemsen:
…e twice

    * tests/DCPS/FindTopic/findtopic.cpp:

### DIFF
--- a/tests/DCPS/FindTopic/findtopic.cpp
+++ b/tests/DCPS/FindTopic/findtopic.cpp
@@ -92,6 +92,38 @@ int sub_program(DDS::DomainParticipant* participant)
   return EXIT_SUCCESS;
 }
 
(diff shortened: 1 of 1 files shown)


mitza-oci:
Is the result of the test what was expected based on the spec?

jwillemsen:
The spec is not clear what should be done, the test now validates the behavior that is currently implemented and that we can call delete_topic twice. RTI does throw an exception when you do a second create_topic.

mitza-oci:
> 
> 
> The spec is not clear what should be done, the test now validates the behavior that is currently implemented and that we can call delete_topic twice. RTI does throw an exception when you do a second create_topic.

OK, so we can go ahead and merge it?

PR #2832, opened 2021-07-13
Relay statistics

jrw972:


dds/DCPS/transport/framework/TransportDefs.h
```diff
@@ -249,6 +249,26 @@ static const ConnectionInfoFlags CONNINFO_UNICAST = (1 << 0);
 static const ConnectionInfoFlags CONNINFO_MULTICAST = (1 << 1);
 static const ConnectionInfoFlags CONNINFO_ALL = static_cast<ConnectionInfoFlags>(-1);
 
+struct RelayMessageCounts {
```
simpsont-oci, line 252: It feels a bit messy to have RelayMessageCounts get passed down into the transport where stun_send and stun_recv are ignored. Maybe RelayMessageCounts (w/ stun_send, stun_recv) inherits from Basic(Rtps)MessageCounts (w/ (rtps_)send & recv)? And then a BasicMessageCount reference could get passed down into the transport with a `get_and_reset_message_counts` method??

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -293,6 +294,10 @@ RtpsUdpSendStrategy::send_single_i(const iovec iov[], int n,
 {
   OPENDDS_ASSERT(addr != ACE_INET_Addr());
 
+  if (addr == link_->transport().config().rtps_relay_address()) {
+    ++link_->transport().relay_message_counts_.rtps_send;
```
simpsont-oci, line 298: Should this only get incremented on a success? Should there be a "failed" count somewhere?

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -427,6 +427,13 @@ RtpsUdpTransport::update_locators(const RepoId& remote,
   }
 }
 
+void
+RtpsUdpTransport::get_and_reset_relay_message_counts(RelayMessageCounts& counts)
+{
+  counts = relay_message_counts_;
```
simpsont-oci, line 433: I know it's not on master yet, but I'm pretty sure this is going to eventually cause trouble with ThreadSanitizer tests without some kind of synchronization between threads, assuming whatever thread is querying the stats is not the same as the write thread and/or the transport reactor thread. That said, I don't see anything in this PR actually making the call into get_and_reset (presumably happens in the relay?) so maybe not an issue yet.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -720,6 +727,10 @@ RtpsUdpTransport::IceEndpoint::choose_send_socket(const ACE_INET_Addr& destinati
 void
 RtpsUdpTransport::IceEndpoint::send(const ACE_INET_Addr& destination, const STUN::Message& message)
 {
+  if (destination == transport.config().rtps_relay_address()) {
+    ++transport.relay_message_counts_.stun_send;
```
simpsont-oci, line 731: same question about success & failure here... and for rtps_send increments.

dds/DCPS/transport/framework/TransportDefs.h
jrw972: Ideally, this wouldn't be in the framework at all.
However, the way the dependencies work and the design of the transport framework forced me to put them here.
Removing the stun members to a derived class obviously means that there would be some form of casting in the transport.  I think we wait to see if fixing the framework makes some of these unnecessary and/or another transport to make use of the feature.

mitza-oci (approved):

PR #2846, opened 2021-07-16
GitHub Actions: build Doxygen documentation

mitza-oci:


.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,104 @@
...
+      - '!etc/**'
+      - '!docs/**'
+      - '!hooks/**'
+      - '!tools/scripts/**'
+      - '!**.md'
+      - '!**.rst'
+      - '!**/.gitignore'
+      - '!**/.lint_config'
+      - '!**/README*'
+  pull_request:
```
simpsont-oci, line 29: Do want to build Doxygen for pull requests? I could see it being handy to check for errors if we ever get the Doxygen build clean, but we're pretty far from that right now I think...

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,104 @@
+name: Doxygen
+
+on:
+  push:
+    branches:
+      - master
+      - branch-DDS-3.*
+      - gh_wf_*
```
simpsont-oci, line 8: Not sure we want to build doxygen for gh_wf_ branches...

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,104 @@
...
+        export DDS_ROOT=`pwd`/OpenDDS
+        mkdir -p Doxygen
+        OpenDDS/tools/scripts/generate_combined_doxygen.pl Doxygen
+    - name: create Doxygen artifact
+      if: steps.cache-artifact.outputs.cache-hit != 'true'
+      shell: bash
+      run: |
+        tar cJf Doxygen.tar.xz Doxygen
+    - name: upload Doxygen output
+      uses: actions/upload-artifact@v2
```
simpsont-oci, line 101: I'm pretty sure this is going to create a zip with a tar.xz inside it, which is a bit redundant. If you just want a zip of the Doxygen directory, I think you can just give the directory name and it'll zip the whole thing, similar to what we do for the autobuild_output. I used the tar.xz for the cached artifacts to try to cut down as much as possible on artifact size, since I knew some would be very large.

PR #2847, opened 2021-07-16
Make SEDP's passive connect duration configurable

mitza-oci:


dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -34,6 +36,13 @@ namespace {
 #endif
     return fallback;
   }
+
+  OpenDDS::DCPS::TimeDuration convert_msec(unsigned long msec)
```
simpsont-oci, line 40: You can use TimeDuration::from_msec (static method) instead of this.

simpsont-oci (changes requested):

simpsont-oci (approved):

simpsont-oci:
SafetyProfile build issues?

mitza-oci:
> 
> 
> SafetyProfile build issues?

Looks like it's not just safety profile.  I'll come back to this later.

PR #2853, opened 2021-07-19
prevent sign extension of char to int for isprint

mitza-oci:


dds/idl/dds_generator.h
```diff
@@ -661,7 +661,7 @@ std::ostream& char_helper(std::ostream& o, CharType value)
   case '\a':
     return o << "\\a";
   }
-  if (isprint(value)) {
+  if (isprint(static_cast<unsigned char>(value))) {
```
jwillemsen, line 664: Why not use ACE_OS::ace_isprint?

mitza-oci: I didn't write the code, so I'll let @iguessthislldo answer specifically.  I'm just trying to get the builds fixed.
There is a `C++ Standard` section in guidelines.md that is relevant.

iguessthislldo: This needs to work for `wchar` too, so maybe it needs to use `char_value`, which I defined right about this funcion?

jwillemsen: is_print result depends on the locale, different locale could result in different output, is that intended?

iguessthislldo: Also what ever we do, it should not truncate the `wchar_t` value like this looks like it would.

mitza-oci: `wchar_t` has its own stdlib function (`iswprint`?)

iguessthislldo: Yes, but we can't print a directly print a `wchar_t` to the byte-char based `ostream`. We have to basically see if it's ASCII first, else it falls back to the hex value.

iguessthislldo: > is_print result depends on the locale, different locale could result in different output, is that intended

Sorta. I'm not really at ease with how text encodings are handled (or more accurately not) in IDL and DDS in general, but I figured it was good enough for now. I was just trying to improve it so it's not just printing out the char as is or the wchar casted to a char, but easier to read than doing something like always using a hex escape.

iguessthislldo (approved):

PR #2855, opened 2021-07-19
Performance Improvements From Profiling Fragmentation Tests

simpsont-oci:
Primary Improvements:
- [X] Avoid full copy of TransportQueueElement in TransportSendBuffer (SingleSendBuffer) by allowing calls to ACE_Message_Block::duplicate().
- [x] Avoid many unnecessary dynamic_casts inside TransportCustomizedElement::original_send_element() by caching result (which should not change over time).

### DIFF
--- a/dds/DCPS/transport/framework/TransportCustomizedElement.h
+++ b/dds/DCPS/transport/framework/TransportCustomizedElement.h
@@ -36,6 +36,8 @@ public:
   SequenceNumber sequence() const;
   void set_sequence(const SequenceNumber& value);
 
+  virtual ACE_Message_Block* duplicate_msg() const;
+
   virtual const ACE_Message_Block* msg() const;
   void set_msg(Message_Block_Ptr m);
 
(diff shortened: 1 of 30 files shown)


dds/DCPS/transport/framework/TransportRetainedElement.inl
```diff
@@ -9,49 +9,66 @@
...
+    const RepoId&                      pubId
+) : TransportQueueElement(1),
+    publication_id_(pubId),
+    mb_allocator_(0),
+    db_allocator_(0),
+    is_duplicate_(true)
+{
+  if (message != 0) {
+    msg_.reset(message->duplicate());
+  }
```
mitza-oci, line 24: I'd prefer using the initializer list, but I'm OK with it as-is

dds/DCPS/transport/framework/TransportCustomizedElement.h
```diff
@@ -55,8 +57,11 @@ class OpenDDS_Dcps_Export TransportCustomizedElement
 
   virtual ~TransportCustomizedElement();
 
+  static const TransportSendElement* find_original_send_element(TransportQueueElement* orig);
+
 private:
   TransportQueueElement* orig_;
+  const TransportSendElement* ose_;
```
mitza-oci, line 64:
```suggestion
  const TransportSendElement* original_send_element_;
```

mitza-oci (commented):
It would be good to have unit test coverage for the CopyChain algorithm

mitza-oci (approved):

PR #2862, opened 2021-07-22
Explicit Int Support And Test For Java / JNI

simpsont-oci:


java/idl2jni/codegen/im_jni.cpp
```diff
@@ -1649,8 +1659,10 @@ ostream &operator<< (ostream &o, AST_Expression::AST_ExprValue *ev)
   case AST_Expression::EV_wchar:
     o << "L\'" << ev->u.wcval << '\'';
     break;
+  case AST_Expression::EV_int8:
```
iguessthislldo, line 1662: I think combining `uint8` and `octet` are okay, but the `int8` values won't come out right because the other two are unsigned. This is what I'm doing in `dds_generator.h`: https://github.com/objectcomputing/OpenDDS/blob/ef88d4e8561b8e3bbad42b953118949a2322f6b2/dds/idl/dds_generator.h#L684-L685

java/idl2jni/codegen/im_java.cpp
```diff
@@ -217,6 +217,8 @@ std::string param_type(AST_Type *t, AST_Argument::Direction dir)
     case AST_PredefinedType::PT_boolean:
     case AST_PredefinedType::PT_char:
     case AST_PredefinedType::PT_wchar:
+    case AST_PredefinedType::PT_int8:
+    case AST_PredefinedType::PT_uint8:
```
iguessthislldo, line 221: I think I forgot about this earlier, but you will need macro checks around `PT_int8` and `PT_uint8`. Like here https://github.com/objectcomputing/OpenDDS/blob/ef88d4e8561b8e3bbad42b953118949a2322f6b2/dds/idl/langmap_generator.cpp#L138-L141

This is where you include `dds/DCPS/Definitions.h`.

iguessthislldo (changes requested):

java/idl2jni/codegen/im_java.cpp
```diff
@@ -376,6 +380,8 @@ ostream &operator<< (ostream &o, AST_Expression::AST_ExprValue *ev)
   case AST_Expression::EV_wchar:
     o << ev->u.wcval;
     break;
+  case AST_Expression::EV_int8:
+  case AST_Expression::EV_uint8:
   case AST_Expression::EV_octet:
     o << static_cast<int>(ev->u.oval);
```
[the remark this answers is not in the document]
simpsont-oci: I don't see any warnings during java compilation, and we're not actually executing anything here. I'll add a conversion method, but I'm a little worried we'll be modifying how existing octets gets translated.

java/idl2jni/codegen/im_java.cpp
```diff
@@ -376,8 +391,12 @@ ostream &operator<< (ostream &o, AST_Expression::AST_ExprValue *ev)
   case AST_Expression::EV_wchar:
     o << ev->u.wcval;
     break;
+  case AST_Expression::EV_int8:
+    o << static_cast<short>(ev->u.int8val);
+    break;
+  case AST_Expression::EV_uint8:
```
iguessthislldo, line 397: These will need macro checks

java/idl2jni/codegen/im_java.cpp
```diff
@@ -452,6 +471,8 @@ bool idl_mapping_java::gen_const(UTL_ScopedName *name, bool nestedInInteface,
   case AST_Expression::EV_wchar:
     type_str = "char";
     break;
+  case AST_Expression::EV_int8:
+  case AST_Expression::EV_uint8:
```
iguessthislldo, line 475: Also these

java/idl2jni/codegen/im_jni.cpp
```diff
@@ -1649,8 +1669,12 @@ ostream &operator<< (ostream &o, AST_Expression::AST_ExprValue *ev)
...
+    o << static_cast<short>(ev->u.int8val);
+    break;
+  case AST_Expression::EV_uint8:
```
iguessthislldo, line 1675: And these

iguessthislldo (changes requested):

iguessthislldo (changes requested):
So I tried compiling this because I wanted to confirm the Java constants were correct in the jar. The jar file was empty and in addition to that `idl2jni` was outputting this:

```
ERROR: Bad discriminant type (shouldn't happen here)
ERROR: Bad discriminant type (shouldn't happen here)
```

java/idl2jni/codegen/im_java.cpp
```diff
@@ -781,7 +814,7 @@ void writeUnionDefaultValue(ostream &os, AST_Expression::ExprType udisc_type,
     os << dv.u.ulonglong_val;
     break;
   default:
-    cerr << "ERROR: Bad discriminant type (shouldn't happen here)\n";
+    cerr << "ERROR - " << __FILE__ << ":" << __LINE__ << " - Bad discriminant type '" << udisc_type << "' (shouldn't happen here)\n";
```
iguessthislldo, line 817: Can you have it do a `BE_arbort()`` here and in the other errors? It really shouldn't keep going in these cases.

iguessthislldo (changes requested):
I was able to build the jar, disassemble it, and confirm the values were correct. Besides not exiting after an error I'd say this is almost ready to merge.

iguessthislldo (approved):

PR #2863, opened 2021-07-22
Disable fail-fast strategy for matrix builds

simpsont-oci:
In the event that one of the check test result jobs matrix jobs fails (invalid input, job failure, etc), the other jobs should not get canceled.

### DIFF
--- a/.github/workflows/check_test_results.yml
+++ b/.github/workflows/check_test_results.yml
@@ -52,7 +52,7 @@ jobs:
     needs: get_info
     runs-on: ubuntu-20.04
     strategy:
-      fast-fail: false
+      fail-fast: false
       matrix:
         info: ${{ fromJSON(needs.get_info.outputs.info) }}
     if: ${{ github.event.workflow_run.conclusion != 'cancelled' }}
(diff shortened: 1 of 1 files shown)


iguessthislldo (approved):

iguessthislldo:
I wish this wasn't the default. I've had to disable it in all the multi-job workflows I've created.

mitza-oci:
> 
> 
> I wish this wasn't the default. I've had to disable it in all the multi-job workflows I've created.

Is this a problem in DOCGroup/ACE_TAO?

iguessthislldo:
> > I wish this wasn't the default. I've had to disable it in all the multi-job workflows I've created.
> 
> Is this a problem in DOCGroup/ACE_TAO?

Yes and I keep wanting to change everytime I see it like here https://github.com/DOCGroup/ACE_TAO/pull/1627#issuecomment-884679260, but I never get around to it.

iguessthislldo:
The reason I didn't change it there is I wasn't sure if it was a general vcpkg failure, so I figured I'd leave it for the time being.

.github/workflows/check_test_results.yml
```diff
@@ -52,6 +52,7 @@ jobs:
     needs: get_info
     runs-on: ubuntu-20.04
     strategy:
+      fast-fail: false
```
simpsont-oci, line 55:
```suggestion
      fail-fast: false
```

simpsont-oci:
https://github.com/DOCGroup/ACE_TAO/pull/1632

PR #2870, opened 2021-07-23
Use bitwise alignment calculation

simpsont-oci:
For any alignment of a power of two (the only kind we support), this calculation avoids modulus calculation and branch prediction.

### DIFF
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -18,12 +18,7 @@ namespace DCPS {
 ACE_INLINE
 void align(size_t& value, size_t by)
 {
-  // TODO(iguessthislldo): If this is always alignment by a power of two, it
-  // might benefit from a bitwise version.
-  const size_t offset_by = value % by;
-  if (offset_by) {
-    value += by - offset_by;
-  }
+  value = (value + by - 1) & ~(by - 1);
 }
 
 ACE_INLINE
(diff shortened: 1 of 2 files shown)


iguessthislldo (approved):
I was pretty sure we only did power of two, just was trying to avoid premature optimization when doing the dealing with all the XTypes stuff. We may also want to assert it if we want to be sure, but I'll leave that up to you to decide.

simpsont-oci:
Normally, I would be in favor of the assertion, but when I was profiling I made a no-inline build and was surprised at the amount of time spent in this code. The profile results before and after this change were noticeable, and I assume it's due to the lack of modulus or branching. If we added the assert back in, debug builds would regain the branch prediction issue... and so I vote no for the assert here.

Perhaps a different way to prevent misuse would be to keep the old align method but make a new `align_pow2` method implemented this way, and move everything over to use the new method. But then again, will we ever need / support non-power-of-2 alignments? I think it's unlikely?

iguessthislldo:
> The profile results before and after this change were noticeable, and I assume it's due to the lack of modulus or branching.

I would expect branch prediction to be much worse than modulus, but yeah that's the reason I made the comment. I figured we did it so much optimizing it would probably be worth it.

> But then again, will we ever need / support non-power-of-2 alignments? I think it's unlikely?

Yea, I don't think we will. I was more of thinking if it got accidentally set something that's not a power of 2. I think it'll be okay as is though.

PR #2872, opened 2021-07-23
Avoid copying RepoIdSet values in match_endpoints

simpsont-oci:


simpsont-oci:
Hmm, the Lifecycle test failures seem problematic...

simpsont-oci:
I think this will need to be revisited at a later time, after we've dispatched some of the BIT reverse locking.

PR #2876, opened 2021-07-24
XTypes: Integrate complete TypeObject with builtin topics

sonndinh:


ClaytonCalabrese (approved):
In general the PR looks good. If we want to delay the locking problem in the XTypes test I am okay with that.

The w19 failures on CI are weird but I don't see how they could be because of this PR.

sonndinh:
> In general the PR looks good. If we want to delay the locking problem in the XTypes test I am okay with that.
> 
> The w19 failures on CI are weird but I don't see how they could be because of this PR.

Many of the tests' timeout errors have been fixed after I removed the lock acquisition statement in `Sedp::TypeLookupReplyReader::get_continuation_point()`, including the timeout errors of the XTypes test, as it creates a deadlock which caused the timeout issue.

dds/DCPS/TypeSupportImpl.cpp
```diff
@@ -49,35 +49,46 @@ TypeSupportImpl::get_type_name()
...
+    ti_with_deps.typeid_with_size.type_id = ti;
+    const XTypes::TypeObject& to = pos->second;
+    const size_t sz = serialized_size(XTypes::get_typeobject_encoding(), to);
+    ti_with_deps.typeid_with_size.typeobject_serialized_size = static_cast<unsigned>(sz);
   }
 
-  type_info.minimal.dependent_typeid_count = 0;
-  type_info.complete.typeid_with_size.typeobject_serialized_size = 0;
-  type_info.complete.dependent_typeid_count = 0;
+  ti_with_deps.dependent_typeid_count = -1;
```
mitza-oci, line 78: Since this `-1` is a special value defined by the spec, we should use a named constant for it.

dds/DCPS/DiscoveryBase.h
```diff
@@ -1247,40 +1271,56 @@ namespace OpenDDS {
...
-        md.writer = writer;
-        md.reader = reader;
+        // If the type object is not in cache, send RPC request
         md.time_added_to_map = MonotonicTimePoint::now();
 
+        // NOTE(sonndinh): Is it possible for a discovered endpoint to include only the "complete"
+        // part in its TypeInformation? If it's possible, then we may need to handle that case, i.e.,
+        // request only the remote complete TypeObject (if it's not already in the cache).
+        // The following code assumes when the "minimal" part is not included in the discovered
+        // endpoint's TypeInformation, then the "complete" part also is not included.
```
mitza-oci, line 1281: Do we expect to resolve before merging this PR?

dds/DCPS/DiscoveryBase.h
```diff
@@ -1247,40 +1271,56 @@ namespace OpenDDS {
...
             }
           } else if (!reader_local && writer_local) {
-            if (type_lookup_service_ && !type_lookup_service_->type_object_in_cache(
-                reader_type_info->minimal.typeid_with_size.type_id)) {
+            const bool need_minimal_tobjs = type_lookup_service_ &&
+              !type_lookup_service_->type_object_in_cache(reader_type_info->minimal.typeid_with_size.type_id);
+            const bool need_complete_tobjs = type_lookup_service_ && use_xtypes_complete_ &&
+              reader_type_info->complete.typeid_with_size.type_id.kind() != XTypes::TK_NONE &&
+              !type_lookup_service_->type_object_in_cache(reader_type_info->complete.typeid_with_size.type_id);
+
```
mitza-oci, line 1311: Since this is the same for remote-reader as the code above for remote-writer, could it be refactored to a common method?

dds/DCPS/DiscoveryBase.h
```diff
@@ -1163,9 +1176,19 @@ namespace OpenDDS {
...
 
       struct MatchingData {
-        RepoId writer;
-        RepoId reader;
-        SequenceNumber rpc_sequence_number;
+        // Sequence number of the first request for remote minimal types.
+        SequenceNumber rpc_seqnum_minimal;
+
+        // Whether all minimal types are obtained.
+        bool got_minimal;
```
mitza-oci, line 1183: Prior to this change, `MatchingData` didn't have any data members that needed initialization beyond the defaults.  Now that there are the two `bool` members, there should be an explicit default constructor and not just the compiler generated one.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4302,15 +4325,33 @@ bool Sedp::TypeLookupReplyReader::process_get_types_reply(const XTypes::TypeLook
...
-        "received reply with no data\n"));
+                 "received reply with no data\n"));
     }
     return false;
   }
 
   sedp_.type_lookup_service_->add_type_objects_to_cache(reply._cxx_return.getType.result.types);
+
+  if (reply._cxx_return.getType.result.complete_to_minimal.length() != 0) {
+    if (DCPS::DCPS_debug_level) {
```
mitza-oci, line 4348: Should this be conditional on a higher debug level?

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -4620,6 +4661,15 @@ Sedp::DiscoveryReader::data_received_i(const DCPS::ReceivedDataSample& sample,
       wdata.ice_agent_info_ = pos->second;
     }
 #endif
+
+    // Cache non-empty TypeInformation
```
mitza-oci, line 4665:
```suggestion
```

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
sonndinh: Is debug level 2 (logs once per DDS entity) appropriate for this?

dds/DCPS/DiscoveryBase.h
sonndinh: No, I think it's not necessary to have this resolve in this PR. This is more of a note on whether our current assumption, that the minimal member is always present in `TypeInformation`, is compatible with the spec. If we decide that we need to make some changes to this part, I'd like to do it in a separate PR.

dds/DCPS/TypeSupportImpl.cpp
```diff
@@ -49,35 +49,46 @@ TypeSupportImpl::get_type_name()
...
+}
+
+void TypeSupportImpl::to_type_info(XTypes::TypeInformation& type_info) const
+{
+  to_type_info_i(type_info.minimal, getMinimalTypeIdentifier(), getMinimalTypeMap());
+
+  // Properly populate the complete member if complete TypeObjects are generated.
```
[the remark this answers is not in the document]
sonndinh: Yes, the use case makes sense. So if we move the Complete/Minimal configuration to Service_Participant and an application has multiple domain participants, each belongs to a domain, all of these domain participants would have the same Complete/Minimal config, right? If that is the case, it would reduce the flexibility we're having which allows each domain to have its own XTypes config.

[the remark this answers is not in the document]
mitza-oci: That's a good point.  We could consider having it configured in the `OpenDDS::DCPS::Discovery` object.  But this could also come in a future iteration when someone needs to have different domains with different configurations.

dds/DCPS/RTPS/Sedp.cpp
mitza-oci: Maybe 3 or 4?  It could be more than once per entity since it would be based on the associations with remote entities.  @iguessthislldo may have some thoughts on this

dds/DCPS/DiscoveryBase.h
ClaytonCalabrese: If our goal for this task is to only have self compatibility up and running I don't think it needs to be resolved, but this note will be helpful for if/when we worry about interoperability in the future.

mitza-oci: If the remote peer provides only a Complete TypeIdentifier in discovery, we could request the corresponding Complete TypeObject via RPC and then run the complete-to-minimal algorithm on it.

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -712,14 +712,14 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
         } else if (name == "UseXTypes") {
           const OPENDDS_STRING& value = it->second;
           int smInt;
-          if (!DCPS::convertToInteger(value, smInt)) {
+          if (!DCPS::convertToInteger(value, smInt) || smInt < 0 || smInt > 2) {
```
mitza-oci, line 715: For configuration parsing, how about allowing 0 or 1 (for backwards compatibility) or a string value for the enumerator?

mitza-oci (changes requested):

dds/DCPS/RTPS/RtpsDiscovery.cpp
sonndinh: This allows `UseXTypes` to be 0, 1, 2 for "non-XTypes", "Minimal", and "Complete" respectively. What else am I missing?

mitza-oci: I meant allowing a config file that has the string in it, `UseXTypes=Complete`

mitza-oci (approved):

PR #2879, opened 2021-07-28
Support INSTALL_LIB in CMake Module

iguessthislldo:
Follow up to https://github.com/objectcomputing/OpenDDS/pull/2740

Support `INSTALL_LIB` being used with `make install` to change the name of the `lib` directory. To make it possible for `find_package` to find the OpenDDS CMake module in this case, the installation destination for the CMake module was changed from `lib` to `share`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ OpenDDS 3.20.0 is currently in development, so this list might change.
     - To help facilitate this, a new function called `opendds_get_library_dependencies` has been added.
   - To help install generated files, `OPENDDS_TARGET_SOURCES` now adds lists of the files that where passed in and generated that are part of the `PUBLIC` and `INTERFACE` scopes as properties on the target (#3315)
   - Added an `OPENDDS_DEFAULT_SCOPE` option that allows changing the default scope of `OPENDDS_TARGET_SOURCES` (#3315)
+  - Support `INSTALL_LIB` being used with `make install` to change the name of the `lib` directory (#2879)
+    - NOTE: This required changing the CMake module installation destination from `lib` to `share`.
   - See `docs/cmake.md` for details on all of these new features
 
 ### Fixes:
--- a/dds/DdsDcps.mpc
+++ b/dds/DdsDcps.mpc
@@ -81,14 +81,22 @@ project(OpenDDS_Dcps): core, coverage_optional, \
     ../rules.dds.GNU
   }
 
+  verbatim (gnuace, top, 1) {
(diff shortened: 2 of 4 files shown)


mitza-oci:
Looks like this error is relevant

https://github.com/objectcomputing/OpenDDS/runs/5430674525?check_suite_focus=true#step:8:54

NEWS.md
```diff
@@ -9,6 +9,8 @@ OpenDDS 3.20.0 is currently in development, so this list might change.
     - To help facilitate this, a new function called `opendds_get_library_dependencies` has been added.
   - To help install generated files, `OPENDDS_TARGET_SOURCES` now adds lists of the files that where passed in and generated that are part of the `PUBLIC` and `INTERFACE` scopes as properties on the target (#3315)
   - Added an `OPENDDS_DEFAULT_SCOPE` option that allows changing the default scope of `OPENDDS_TARGET_SOURCES` (#3315)
+  - Support `INSTALL_LIB` being used with `make install` to change the name of the `lib` directory (#2879)
+    - NOTE: This required changing the CMake module installation destination from `lib` to `share`.
```
mitza-oci, line 13: I thought we were following a convention or recommendation by using `lib/cmake/` for this.  Was that not correct?

What is the impact on users (if any)? Seems like the markdown doc about using CMake may need to change for this.

dds/DdsDcps.mpc
```diff
@@ -81,14 +81,22 @@ project(OpenDDS_Dcps): core, coverage_optional, \
     ../rules.dds.GNU
   }
 
+  verbatim (gnuace, top, 1) {
+    cmake_dir_install_dest=$(DESTDIR)$(INSTALL_PREFIX)/share/cmake/OpenDDS
```
mitza-oci, line 85: Instead of a separate variable this could use GNU Make's function for dirname.

At this point I wonder if this should be its own script.

NEWS.md
iguessthislldo: They're both correct as both `lib` and `share` are searched, this is would just be easier to deal with if we don't know where `lib` is going to be. There is no impact on users unless they're assuming the exact location of the files in an install for some reason. If they're using something like `CMAKE_PREFIX_PATH`, then it will still work.

mitza-oci (approved):
Ready to merge once the CI error in [build_u18_o1d0v1_xer0_j12](https://github.com/objectcomputing/OpenDDS/runs/5430674525?check_suite_focus=true#logs) is addressed

PR #2880, opened 2021-07-28
RtpsRelay improvements

jrw972:


dds/DCPS/transport/framework/TransportReassembly.cpp
```diff
@@ -403,10 +402,10 @@ void TransportReassembly::check_expirations(const MonotonicTimePoint& now)
                      "purge expired leaving %B fragments\n", fragments_.size()));
         }
       } else {
-        expiration_queue_.insert(std::make_pair(iter->second.expiration_, pos->second));
+        expiration_queue_.push_back(std::make_pair(iter->second.expiration_, iter->first));
```
simpsont-oci, line 405: Since push_back doesn't sort like insert, I think we need to be be careful with our while loop's check for `expiration_queue_.front().first <= now`. If the expiration is only ever set to "now + configured_expiration" then I think it's safe, but if we did anything fancier with the expirations (e.g. if "initial" expiration > "update" expiration) they could potentially get inserted out of order and the loop's end condition would be incorrect.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -281,6 +282,11 @@ RtpsUdpSendStrategy::send_single_i(const iovec iov[], int n,
 {
   OPENDDS_ASSERT(addr != ACE_INET_Addr());
 
+  if (addr == link_->transport().config().rtps_relay_address()) {
+    ++link_->transport().relay_rtps_send_count_;
+    link_->transport().report_relay();
```
simpsont-oci, line 287: This seems like a pretty low level place to add a reporting method... should this maybe get called by a periodic task?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2774,6 +2779,26 @@ Spdp::SpdpTransport::send(WriteFlags flags, const ACE_INET_Addr& local_address)
...
+    ++relay_rtps_send_count_;
+    report_relay();
+  }
+}
+
+void
+Spdp::SpdpTransport::report_relay()
+{
+  const DCPS::MonotonicTimePoint now = DCPS::MonotonicTimePoint::now();
+  if (DCPS::DCPS_debug_level >= 4 && now > last_relay_report_ + DCPS::TimeDuration(300)) {
```
simpsont-oci, line 2791: Every 5 minutes? Could this get called as part of a periodic tasks rather than having every call to send_i potentially need to make these checks?

dds/DCPS/transport/framework/TransportReassembly.cpp
jrw972: Viewed another way, we would just be holding on to fragments longer than anticipated.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
jrw972: Probably.  This was quick and dirty when I added it.  I'll remove it from this PR.

simpsont-oci (approved):

PR #2884, opened 2021-07-29
RtpsRelay expiration can be more efficient

jrw972:
Problem
-------

The RtpsRelay has two sets of things that expire:  the IP:port
combinations of clients and GUIDs that are pending (have not completed
discovery).  Both of these involved a multimap that goes from time to
the thing to be expired.  Eliminating the overhead of maintaing the
maps will increase the perfomance of the RtpsRelay.

Solution
--------

Replace the expiration multimaps with a list.  This may create an
inversion when tracking IP:ports where something that should expire
sooner is behind something that will expire later in the list.  In the
worst case, a IP:port might be removed after two expiration durations
which is acceptable.

### DIFF
--- a/tools/dds/rtpsrelaylib/Utility.h
+++ b/tools/dds/rtpsrelaylib/Utility.h
@@ -24,12 +24,41 @@ inline std::string guid_to_string(const OpenDDS::DCPS::GUID_t& a_guid)
   return ss.str();
 }
 
+enum Port {
+  SPDP,
+  SEDP,
+  DATA
+};
+
+struct AddrPort {
+  ACE_INET_Addr addr;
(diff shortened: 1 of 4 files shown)


mitza-oci (changes requested):
Tests are failing because this part is dereferencing an "end" iterator (pos):
```
  if (send_to_application_participant) {
    GuidAddrSet::GuidAddrSetMap::iterator pos = proxy.find(config_.application_participant_guid());
    venqueue_message(application_participant_addr_, *pos->second.select_stats_reporter(port()), msg, now);
    ++sent;
  }
```

mitza-oci (approved):

PR #2886, opened 2021-08-01
Fix for deadlocks brought out by the GroupPresentation test.

pmesnier:
This PR focuses on two potential deadlock scenarios, both in the subscriber impl code, the function to get a list of current data readers and a function to notify datareaders. The problem is that multiple threads are passing through these functions and must mediate the sharing of resources using locks. However there is a lock on the SubscriberImpl object that is grabbed a the start of these functions, then during processing a second lock on the DataReaderImpl object is repeatedly taken while evaluating individual data readers. This triggers a deadlock when a thread is holding the DataReaderImpl instance lock and waiting for the SubscriberImpl lock, but another thread is holding the SI lock and wants the particular DataReaderImpl instance lock currently held by the first thread.
This solution is to make a local copy of the data reader collection to be searched then release the SI lock before iterating the data reader collection. This avoids the deadlock condition.

### DIFF
--- a/dds/DCPS/SubscriberImpl.cpp
+++ b/dds/DCPS/SubscriberImpl.cpp
@@ -592,7 +592,7 @@ SubscriberImpl::notify_datareaders()
   }
 
 #ifndef OPENDDS_NO_MULTI_TOPIC
-  MultiTopicReaderMap localmtr;
+  MultitopicReaderMap localmtr;
   {
     ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
                     guard,
--- a/dds/DCPS/SubscriberImpl.cpp
+++ b/dds/DCPS/SubscriberImpl.cpp
@@ -1037,16 +1037,18 @@ SubscriberImpl::coherent_change_received (RepoId&         publisher_id,
                                           DataReaderImpl* reader,
                                           Coherent_State& group_state)
 {
-  ACE_GUARD(ACE_Recursive_Thread_Mutex,
-            guard,
-            this->si_lock_);
-
+  DataReaderSet localdrs;
+  {
+    ACE_GUARD(ACE_Recursive_Thread_Mutex,
+              guard,
+              this->si_lock_);
+     localdrs = datareader_set_;
+  }
   // Verify if all readers complete the coherent changes. The result
   // is either COMPLETED or REJECTED.
   group_state = COMPLETED;
-  DataReaderSet::const_iterator endIter = datareader_set_.end();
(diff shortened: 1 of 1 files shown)


ClaytonCalabrese:
Has anyone seen the tests/DCPS/LivelinessTimeout/run_test.pl rtps_disc failure on test_m10_o1d0_sec before, or could that have been introduced by this PR?

dds/DCPS/SubscriberImpl.cpp
```diff
@@ -569,15 +571,17 @@ SubscriberImpl::get_datareaders(
...
-  for (it = datareader_map_.begin(); it != datareader_map_.end(); ++it) {
-    if (it->second->have_sample_states(DDS::NOT_READ_SAMPLE_STATE)) {
+  {
+    ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
+                    guard,
+                    this->si_lock_,
+                    DDS::RETCODE_ERROR);
+    localreadermap = datareader_map_;
+  }
+  for (it = localreadermap.begin(); it != localreadermap.end(); ++it) {
```
mitza-oci, line 583: Does the `it` object need to outlive the for loop?  I realize it was that way before this PR, but it would be good to reduce its scope to the smallest necessary.

dds/DCPS/SubscriberImpl.cpp
```diff
@@ -569,15 +571,17 @@ SubscriberImpl::get_datareaders(
...
+    if ( it->second->have_sample_states(DDS::NOT_READ_SAMPLE_STATE)) {
```
mitza-oci, line 584:
```suggestion
    if (it->second->have_sample_states(DDS::NOT_READ_SAMPLE_STATE)) {
```

simpsont-oci (approved):
I'd be curious what ThreadSanitizer reports for the GroupPresentation tests, but that may need to come later...

dds/DCPS/SubscriberImpl.cpp
pmesnier: My first inclination was to minimize changes, but I agree that the iterator is not needed again outside the loop.

dds/DCPS/SubscriberImpl.cpp
pmesnier: done

mitza-oci:
These test failures in GitHub actions seem to be new to this PR:
```
tests/security/attributes/run_test.pl Partitions_Reordered
tools/modeling/tests/MultiTopic/run_test.pl
```

PR #2887, opened 2021-08-02
RtpsRelay discovery logging and statistics don't handle updates

jrw972:
Problem
-------

It is expected that RtpsRelay clients will determine and update their
server-reflexive address.  These updates will appear as new samples in
the BITs read by the RtpsRelay.  The RtpsRelay will log these updates
as new participants, publications, and subscriptions which is
misleading.

Solution
--------

Distinguish between new participants, publications, and subscriptions
and updates to existing ones.

### DIFF
--- a/tools/rtpsrelay/GuidPartitionTable.cpp
+++ b/tools/rtpsrelay/GuidPartitionTable.cpp
@@ -1,9 +1,9 @@
 #include "GuidPartitionTable.h"
 namespace RtpsRelay {
 
-void GuidPartitionTable::insert(const OpenDDS::DCPS::GUID_t& guid, const DDS::StringSeq& partitions)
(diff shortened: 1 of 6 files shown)


tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -14,6 +14,7 @@ void GuidPartitionTable::insert(const OpenDDS::DCPS::GUID_t& guid, const DDS::St
     parts.insert("");
   }
 
+  const Result result = guid_to_partitions_.count(guid) == 0 ? ADDED : UPDATED;
   const auto& x = guid_to_partitions_[guid];
```
mitza-oci, line 18: combine these to avoid needing to repeat the search (if UPDATED)

mitza-oci (changes requested):

mitza-oci (approved):

PR #2888, opened 2021-08-02
ThreadSanitizer Error / Warning Cleanup, Round 1

simpsont-oci:
Running ThreadSanitizer against `tests/DCPS/Messenger/run_test.pl rtps_disc` and `tests/DCPS/EntityLifecycleStress/run_test.pl rtps_disc` revealed a number of data races and mutex lock order inversions (potential deadlocks).

These are some of the 'safer' / 'easier' changes that resulted in clean runs, but aren't enough on their own to fix all the reported issues or to allow a theoretical `tsan` build to catch these issues earlier.

In particular, the timer RCH changes in RtpsUdpDatalink are actually heap-use-after-free issues, resulting from the fact that the reactor believes it's holding a reference to the timer's RcEventHandler base class, when in reality these timers were simply deleted without regard for their reference count, since they were simple members of RtpsWriter/RtpsReader/RtpsUdpDatalink.

If the PR experiences too many errors, it may make sense to split this PR into several smaller ones for debugging.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -2056,6 +2056,7 @@ Spdp::remove_discovered_participant_i(DiscoveredParticipantIter& iter)
 void
 Spdp::init_bit(const DDS::Subscriber_var& bit_subscriber)
 {
+  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
   bit_subscriber_ = bit_subscriber;
 
   // This is here to make sure thread status gets a valid BIT Subscriber
@@ -2070,8 +2071,12 @@ public:
 void
 Spdp::fini_bit()
 {
-  bit_subscriber_ = 0;
-  DCPS::ReactorTask_rch reactor_task = sedp_->reactor_task();
+  DCPS::ReactorTask_rch reactor_task;
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, lock_);
+    bit_subscriber_ = 0;
+    reactor_task = sedp_->reactor_task();
+  }
   if (!reactor_task->is_shut_down()) {
     DCPS::ReactorInterceptor::CommandPtr command = reactor_task->interceptor()->execute_or_enqueue(new Noop());
     command->wait();
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -2056,8 +2056,10 @@ Spdp::remove_discovered_participant_i(DiscoveredParticipantIter& iter)
 void
 Spdp::init_bit(const DDS::Subscriber_var& bit_subscriber)
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
-  bit_subscriber_ = bit_subscriber;
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, lock_);
+    bit_subscriber_ = bit_subscriber;
+  }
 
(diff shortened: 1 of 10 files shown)


dds/DCPS/transport/framework/TransportClient.h
```diff
@@ -135,6 +135,11 @@ class OpenDDS_Dcps_Export TransportClient
   bool associated_with(const GUID_t& remote) const;
   bool pending_association_with(const GUID_t& remote) const;
 
+  RepoId repo_id() const {
```
mitza-oci, line 138:
```suggestion
  RepoId repo_id() const
  {
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -570,7 +570,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
       , id_(id)
       , stopping_(false)
       , nackfrag_count_(0)
-      , preassociation_task_(link->reactor_task_->interceptor(), *this, &RtpsReader::send_preassociation_acknacks)
+      , preassociation_task_(new RtpsReader::Sporadic(link->reactor_task_->interceptor(), *this, &RtpsReader::send_preassociation_acknacks), keep_count())
```
mitza-oci, line 573: Use the `make_rch` helper (like `std::make_shared`)

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
simpsont-oci: I originally had it that way, but unfortunately it wouldn't compile. I believe it was a `const` issue for the constructor... the macro assumes the input is const, but the constructor requires it to be non-const.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2056,6 +2056,7 @@ Spdp::remove_discovered_participant_i(DiscoveredParticipantIter& iter)
 void
 Spdp::init_bit(const DDS::Subscriber_var& bit_subscriber)
 {
+  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
```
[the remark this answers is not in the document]
simpsont-oci: Hmm, the part this was fixing was for multi-threaded access to bit_subscriber_. I can limit the scope to just that part.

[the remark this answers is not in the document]
simpsont-oci: Although it looks like it's just enabling timers... so probably fine? If it's called with the reactor thread used for the timers, then they won't start until we exit anyway. If it's called with a different thread, they will need to acquire the same lock?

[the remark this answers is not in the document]
mitza-oci: I haven't looked into it in depth.  Maybe @jrw972 could help answering these questions (if needed).

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
mitza-oci: That's right, `make_rch` doesn't have all the const/non-const combinations covered since it would be cumbersome.
You can change the `PmfSporadicTask` constructor to use const.  I tried it in branch `rch-experiment` of my fork and it compiled.

pmesnier (approved):
Looks good to me.

mitza-oci (approved):

PR #2895, opened 2021-08-04
RtpsRelay is inefficient

jrw972:
Problem
-------

The RtpsRelay uses std::set and std::map in various places where
std::unordered_set and std::unordered_map may provide superior
performance.

The lookup of guid to partitions iterates over all readers/writers
with the given prefix requiring a log(N) find and P O(1) operations to
iterate.

Solution
--------

Replace ordered sets and maps with unordered ones where possible.  A
hash function for ACE_INET_Addr is need to convert the remaining
cases.

Add a cache for the guid to partitions lookup.

### DIFF
--- a/tools/dds/rtpsrelaylib/Name.h
+++ b/tools/dds/rtpsrelaylib/Name.h
@@ -3,7 +3,7 @@
 
 #include "export.h"
 
-#include <set>
+#include <unordered_set>
 #include <string>
 #include <vector>
 
@@ -21,7 +21,7 @@ public:
 
   explicit Atom(Kind kind) : kind_(kind), character_(0) {}
   explicit Atom(char c) : kind_(CHARACTER), character_(c) {}
-  Atom(bool negated, const std::set<char>& characters)
+  Atom(bool negated, const std::unordered_set<char>& characters)
     : kind_(negated ? NEGATED_CHARACTER_CLASS : CHARACTER_CLASS)
     , character_(0)
     , characters_(characters) {}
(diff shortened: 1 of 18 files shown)


tools/dds/rtpsrelaylib/Name.h
```diff
@@ -109,7 +105,7 @@ class OpenDDS_RtpsRelayLib_Export Name {
   const_iterator begin() const { return atoms_.begin(); }
   const_iterator end() const { return atoms_.end(); }
 
-  bool operator<(const Name& other) const { return atoms_ < other.atoms_; }
+  //bool operator<(const Name& other) const { return atoms_ < other.atoms_; }
```
mitza-oci, line 108: Remove or uncomment

tools/dds/rtpsrelaylib/Utility.h
```diff
@@ -112,14 +112,14 @@ inline bool operator<(const Duration_t& x, const Duration_t& y)
   return x.nanosec() < y.nanosec();
 }
 
-inline OpenDDS::DCPS::RepoId guid_to_repoid(const GUID_t& a_guid)
+inline OpenDDS::DCPS::GUID_t guid_to_repoid(const GUID_t& a_guid)
```
mitza-oci, line 115: consider renaming both this function and the following one

tools/dds/rtpsrelaylib/Utility.h
```diff
@@ -150,7 +150,7 @@ struct GuidHash {
       (std::hash<::CORBA::Octet>{}(guid.entityId.entityKind) << 0);
   }
 };
-typedef std::unordered_set<OpenDDS::DCPS::RepoId, GuidHash> GuidSet;
+typedef std::unordered_set<OpenDDS::DCPS::GUID_t, GuidHash> GuidSet;
```
mitza-oci, line 153: Not necessarily in-scope here... but how about using `using` for these?

mitza-oci (approved):

PR #2896, opened 2021-08-05
Fix and Refactor XML Security File Parsing

iguessthislldo:
- Fix the following issues with the XML files:
  - The custom date/time parsing for the grant validity time range interpreted the time without timezone as local system time, not UTC, and did not parse time zone offsets correctly. Replaced custom code with functionality provided by Xerces if available, else fallback to new custom code.
  - If there are multiple domain rules in governance, then it use only uses the settings from the first domain rule for the rest.
  - Allow for using "1" and "0" in addition to "true" and "false" for XML boolean values in compliance in the XML schema spec.
  - Added missing support for `id_range`s with no `max` in the security XML domain id sets. As part of that tweaked and made use of `OrderedRanges` from `DisjointSequence`.
- Make some the XML parsing more strict and improve error reporting.
- Try to deduplicate code within governance and between governance and permissions.
- Added the `access_error` security debug category at level 1. Moved `access_warn` to level 2.
- Added the `DebugRestore` helper class in `debug.h` to save and restore debug state.

### DIFF
--- a/dds/DCPS/security/AccessControl/Governance.cpp
+++ b/dds/DCPS/security/AccessControl/Governance.cpp
@@ -18,6 +18,7 @@ namespace Security {
 using XML::XStr;
 using namespace XmlUtils;
 using namespace DDS::Security;
+using OpenDDS::DCPS::security_debug;
 
 Governance::TopicAccessRule::TopicAccessRule()
 {
@@ -36,9 +37,11 @@ namespace {
     const ACE_TCHAR* name, bool& value)
   {
     if (!parse_bool(node, value)) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
-        "\"%s\" value, \"%C\", in \"%C\" is not a valid boolean value\n",
-        name, to_string(node).c_str(), doc.filename().c_str()));
+      if (security_debug.access_error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+          "\"%s\" value, \"%C\", in \"%C\" is not a valid boolean value\n",
+          name, to_string(node).c_str(), doc.filename().c_str()));
+      }
       return false;
     }
 
@@ -50,9 +53,11 @@ namespace {
   {
     const xercesc::DOMNodeList* const nodes = parent->getElementsByTagName(XStr(name));
     if (nodes->getLength() != 1) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
-        "expected 1 boolean value \"%s\" in parent element in \"%C\", found %B\n",
-        name, doc.filename().c_str(), nodes->getLength()));
+      if (security_debug.access_error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+          "expected 1 boolean value \"%s\" in parent element in \"%C\", found %B\n",
+          name, doc.filename().c_str(), nodes->getLength()));
+      }
       return false;
     }
 
@@ -65,9 +70,11 @@ namespace {
   {
     const xercesc::DOMNodeList* const nodes = parent->getElementsByTagName(XStr(name));
     if (nodes->getLength() != 1) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
-        "expected 1 proctection kind value named \"%s\" in parent element in \"%C\", found %B\n",
-        name, doc.filename().c_str(), nodes->getLength()));
+      if (security_debug.access_error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+          "expected 1 proctection kind value named \"%s\" in parent element in \"%C\", found %B\n",
+          name, doc.filename().c_str(), nodes->getLength()));
+      }
       return false;
     }
 
@@ -88,8 +95,11 @@ namespace {
       attributes |= enc_attr;
       attributes |= oa_attr;
     } else {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: invalid %s, \"%C\", in \"%s\"\n",
-        name, value.c_str(), doc.filename().c_str()));
+      if (security_debug.access_error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+          "invalid %s, \"%C\", in \"%s\"\n",
+          name, value.c_str(), doc.filename().c_str()));
+      }
       return false;
     }
 
@@ -103,7 +113,10 @@ int Governance::load(const SSL::SignedDocument& doc)
   doc.get_original_minus_smime(xml);
   ParserPtr parser(get_parser(doc.filename(), xml));
   if (!parser) {
-    ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: get_parser failed\n"));
+    if (security_debug.access_error) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+        "get_parser failed\n"));
+    }
     return -1;
   }
 
@@ -123,9 +136,11 @@ int Governance::load(const SSL::SignedDocument& doc)
       const XStr dn_tag = ruleNode->getNodeName();
       if (ACE_TEXT("domains") == dn_tag) {
         if (!parse_domain_id_set(ruleNode, domain_rule.domains)) {
-          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
-            "failed to process domain ids in \"%C\"\n",
-            doc.filename().c_str()));
+          if (security_debug.access_error) {
+            ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: {access_error} Governance::load: "
+              "failed to process domain ids in \"%C\"\n",
+              doc.filename().c_str()));
+          }
           return -1;
         }
       }
--- a/dds/DCPS/security/AccessControl/Governance.cpp
+++ b/dds/DCPS/security/AccessControl/Governance.cpp
@@ -5,21 +5,19 @@
 
 #include "Governance.h"
 
-#include "xercesc/parsers/XercesDOMParser.hpp"
-#include "xercesc/dom/DOM.hpp"
-#include "xercesc/sax/HandlerBase.hpp"
-#include "xercesc/framework/MemBufInputSource.hpp"
-#include "xercesc/util/PlatformUtils.hpp"
+#include "XmlUtils.h"
 
-#include "ace/OS_NS_strings.h"
-#include "ace/XML_Utils/XercesString.h"
-
-#include <stdexcept>
+#include <ace/OS_NS_strings.h>
+#include <ace/XML_Utils/XercesString.h>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace Security {
+
+using XML::XStr;
+using namespace XmlUtils;
+using namespace DDS::Security;
 
 Governance::TopicAccessRule::TopicAccessRule()
 {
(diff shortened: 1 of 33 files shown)


jwillemsen:
Why not use [XSC](https://github.com/DOCGroup/XSC) as is done for all QoS settings?

dds/DCPS/security/AccessControl/XmlUtils.h
```diff
@@ -0,0 +1,95 @@
...
+ParserPtr get_parser(const std::string& xml, const std::string& filename);
+
+OpenDDS_Security_Export
+std::string to_string(const XMLCh* in);
+
+OpenDDS_Security_Export
+bool to_bool(const XMLCh* in, bool& value);
+
+OpenDDS_Security_Export
+bool to_time(const XMLCh* in, time_t& value);
```
mitza-oci, line 43: Since this pattern differs from `to_string`, should we use a different naming scheme like `parse_bool`/`parse_time`?  This may help the caller keep track of the difference between the return type and the out parameter (especially for bool).

dds/DCPS/security/AccessControl/XmlUtils.h
```diff
@@ -0,0 +1,95 @@
...
+
+inline std::string to_string(const xercesc::XMLException& ex)
+{
+  std::string msg = to_string(ex.getMessage());
+  const std::string filename = ex.getSrcFile();
+  if (filename.size()) {
+    msg = filename + ":" + DCPS::to_dds_string(ex.getSrcLine());
```
mitza-oci, line 50: This overwrites the `getMessage` content.  Should it append instead?

dds/DCPS/security/SSL/SignedDocument.cpp
```diff
@@ -142,19 +166,19 @@ class verify_signature_impl
...
-    if (NULL == (store_ = X509_STORE_new())) {
+    if (!store_) {
       OPENDDS_SSL_LOG_ERR("X509_STORE_new failed");
     }
-    if (NULL == (store_ctx_ = X509_STORE_CTX_new())) {
+    if (!store_ctx_) {
       OPENDDS_SSL_LOG_ERR("X509_STORE_CTX_new failed");
     }
-    if (NULL == (reader_ = BIO_new(BIO_s_mem()))) {
+    if (reader_) {
```
mitza-oci, line 181:
```suggestion
    if (!reader_) {
```

dds/DCPS/security/AccessControl/Governance.cpp
```diff
@@ -33,250 +31,185 @@ Governance::Governance()
...
-  xercesc::XMLPlatformUtils::Initialize();
-  DCPS::unique_ptr<xercesc::XercesDOMParser> parser(new xercesc::XercesDOMParser());
+namespace {
+  bool get_bool_tag_value(const SSL::SignedDocument& doc, const xercesc::DOMNode* node,
+    const ACE_TCHAR* name, bool& value)
+  {
+    if (!to_bool(node, value)) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
+        "\"%s\" value, \"%C\", in \"%C\" is not a valid boolean value\n",
+        name, to_string(node).c_str(), doc.filename()));
```
mitza-oci, line 41:
```suggestion
        name, to_string(node).c_str(), doc.filename().c_str()));
```

dds/DCPS/security/AccessControl/Governance.cpp
```diff
@@ -33,250 +31,185 @@ Governance::Governance()
...
-                                        cleaned.size(),
-                                        gMemBufId);
+  bool get_bool_tag(const SSL::SignedDocument& doc, const xercesc::DOMElement* parent,
+    const ACE_TCHAR* name, bool& value)
+  {
+    const xercesc::DOMNodeList* const nodes = parent->getElementsByTagName(XStr(name));
+    if (nodes->getLength() != 1) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
+        "expected 1 boolean value \"%s\" in parent element in \"%C\", found %b\n",
+        name, doc.filename(), nodes->getLength()));
```
mitza-oci, line 55:
```suggestion
        name, doc.filename().c_str(), nodes->getLength()));
```

dds/DCPS/security/AccessControl/Governance.cpp
```diff
@@ -33,250 +31,185 @@ Governance::Governance()
...
-    return -1;
+  bool get_protection_kind(const SSL::SignedDocument& doc, const xercesc::DOMElement* parent,
+    const ACE_TCHAR* name, bool& is_protected, CORBA::ULong& attributes,
+    CORBA::ULong enc_attr, CORBA::ULong oa_attr)
+  {
+    const xercesc::DOMNodeList* const nodes = parent->getElementsByTagName(XStr(name));
+    if (nodes->getLength() != 1) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: "
+        "expected 1 proctection kind value named \"%s\" in parent element in \"%C\", found %b\n",
+        name, doc.filename(), nodes->getLength()));
```
mitza-oci, line 70:
```suggestion
        name, doc.filename().c_str(), nodes->getLength()));
```

dds/DCPS/security/AccessControl/Governance.cpp
```diff
@@ -33,250 +31,185 @@ Governance::Governance()
...
+    } else if (ACE_OS::strcasecmp(value.c_str(), "SIGN_WITH_ORIGIN_AUTHENTICATION") == 0) {
+      is_protected = true;
+      attributes |= oa_attr;
+    } else if (ACE_OS::strcasecmp(value.c_str(), "ENCRYPT_WITH_ORIGIN_AUTHENTICATION") == 0) {
+      is_protected = true;
+      attributes |= enc_attr;
+      attributes |= oa_attr;
+    } else {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Governance::load: invalid %s, \"%C\", in \"%s\"\n",
+        name, value.c_str(), doc.filename()));
```
mitza-oci, line 92:
```suggestion
        name, value.c_str(), doc.filename().c_str()));
```

mitza-oci (changes requested):

mitza-oci:
The Jenkins GHPRB build failed with this error.  Is it different between Xerces-C++ 3.1 and 3.2?
```
OpenDDS/dds/DCPS/security/AccessControl/XmlUtils.cpp:105:17: error: 'class xercesc_3_1::XMLDateTime' has no member named 'getEpoch'
```

iguessthislldo:
> Why not use [XSC](https://github.com/DOCGroup/XSC) as is done for all QoS settings?

I don't think the QoS XML handler actually works with all the possible QoS, but regardless I don't know as I wasn't involved much in the original security work. It might be because it might not have been worth the effort for such a small amount of schema.

mitza-oci (approved):

mitza-oci:
There are still compile errors on Jenkins:
```
DCPS/security/AccessControl/XmlUtils.cpp:132:24: error: 'class xercesc_3_1::XMLDateTime' has no member named 'getYear'
```

iguessthislldo:
> There are still compile errors on Jenkins:
> 
> ```
> DCPS/security/AccessControl/XmlUtils.cpp:132:24: error: 'class xercesc_3_1::XMLDateTime' has no member named 'getYear'
> ```

Yeah I saw. I didn't see that they added those methods at the same time as `getEpoch`. I'm trying to figure out what to do now. I've been trying to look through the 3.1 source code for how they provide that information, but it's looking more and more like they just didn't do anything beyond validating that the string is in the right format.

I don't think it's probably worth it to require 3.2 just for this so I'm thinking a fixed version of the custom parse code should be brought back, at least for when Xerces is 3.1.

After this I think it would be good policy to specify what our minimum required version for our dependencies are in `dependencies.md`, so that we can track that information in a central place.

iguessthislldo:
Out of curiosity I looks at XSC and it doesn't look like it handles datetime at all, though it has code for interpreting booleans (though not `TRUE`/`FALSE`), which Xerces doesn't have.

iguessthislldo:
Also if we have custom code parse code for datetime, at least we now have a way to test it.

dds/DCPS/security/AccessControl/XmlUtils.cpp
```diff
@@ -112,48 +114,268 @@ bool parse_bool(const XMLCh* in, bool& value)
...
+  bool parse_time_string(
+    const std::string& whole, size_t& pos, std::string& value, size_t width = std::string::npos)
+  {
+    const bool to_end = width == std::string::npos;
+    if (pos >= whole.size() || (!to_end && pos + width > whole.size())) {
+      return false;
+    }
+    try {
+      value = whole.substr(pos, width);
+    } catch (const std::out_of_range& e) {
```
mitza-oci, line 129: It doesn't seem possible to get an `out_of_range` here since `pos` is checked above.

dds/DCPS/security/AccessControl/XmlUtils.cpp
```diff
@@ -112,48 +114,268 @@ bool parse_bool(const XMLCh* in, bool& value)
...
+      return false;
+    }
+    if (!to_end) {
+      pos += width;
+    }
+    return true;
+  }
+
+  bool parse_time_field(
+    const std::string& whole, size_t& pos, int& value, size_t width = 2, int offset = 0)
```
mitza-oci, line 139: `offset` isn't adding much value, consider letting the caller handle it since it's not required for parsing

dds/DCPS/security/AccessControl/XmlUtils.cpp
```diff
@@ -112,48 +114,268 @@ bool parse_bool(const XMLCh* in, bool& value)
...
+
+    int tz_min;
+    if (!parse_time_field(str, pos, tz_min)) {
+      return false;
+    }
+
+    /*
+     * We will have to do the reverse of the sign to convert the time timezone
+     * to the UTC. For example 00:00:00+00:00 (Midnight UTC) is 06:00:00-06:00
+     * (6 AM CST).
```
mitza-oci, line 329: CST is indeed UTC-06:00, 6 hours behind UTC.  So midnight in UTC would be 6 PM in CST.

mitza-oci (changes requested):

dds/DCPS/security/AccessControl/XmlUtils.cpp
iguessthislldo: Yea, I added the check afterwards to get it behave how I wanted. I will go ahead and remove it.

dds/DCPS/security/AccessControl/XmlUtils.cpp
iguessthislldo: Thanks, I mixed that up.

mitza-oci (approved):

tests/security/permissions/permissions_test_participant_01.xml
```diff
@@ -10,6 +10,7 @@
       </validity>
       <allow_rule>
         <domains>
+          <min>0</min>
```
mitza-oci, line 13:
```suggestion
          <id>0</id>
```
there are more of these, and the signed docs need to get updated

iguessthislldo: I think I left this in here by accident when I was trying to get validation working, which I tested by having something wrong. I'll change it to the right think which allows all domains.

jwillemsen:
New warnings, see https://buildlogs.remedy.nl/rhel75_opendds_versioned/

```
[Details] /home/build/jenkins/workspace/rhel75_opendds_versioned/BUILD/DOC_ROOT/DDS/dds/DCPS/security/AccessControl/XmlUtils.cpp: In function ‘bool OPENDDS_3_18_0::OpenDDS::Security::XmlUtils::parse_time(const XMLCh*, time_t&)’:
[Details] /home/build/jenkins/workspace/rhel75_opendds_versioned/BUILD/DOC_ROOT/DDS/dds/DCPS/security/AccessControl/XmlUtils.cpp:374:60: warning: ‘tz_char’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```

PR #2903, opened 2021-08-11
more deadlock risk mitigation when using Topic presentation QOS

pmesnier:
This PR adds a couple more uses of local copies of otherwise locked data to avoid conflicts across function calls. This PR also aaddresses some peculiar code specific to the GroupPresentation test case.

### DIFF
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -253,9 +253,10 @@ namespace OpenDDS {
 
     const Observer_rch observer = get_observer(Observer::e_SAMPLE_READ);
 
-    const typename InstanceMap::iterator the_end = instance_map_.end();
-    for (typename InstanceMap::iterator it = instance_map_.begin(); it != the_end; ++it) {
-      const SubscriptionInstance_rch ptr = get_handle_instance(it->second);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+        iter != instances_.end();
+        ++iter) {
+      SubscriptionInstance_rch ptr = iter->second;
 
       bool most_recent_generation = false;
 
@@ -311,10 +312,10 @@ namespace OpenDDS {
 
     const Observer_rch observer = get_observer(Observer::e_SAMPLE_TAKEN);
 
-    const typename InstanceMap::iterator the_end = instance_map_.end();
-    for (typename InstanceMap::iterator it = instance_map_.begin(); it != the_end; ++it) {
-      DDS::InstanceHandle_t handle = it->second;
-      OpenDDS::DCPS::SubscriptionInstance_rch ptr = get_handle_instance(handle);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+         iter != instances_.end();
+         ++iter) {
+      SubscriptionInstance_rch ptr = iter->second;
 
       bool most_recent_generation = false;
 
@@ -1386,10 +1387,11 @@ DDS::ReturnCode_t take_i(MessageSequenceType& received_data,
   if (!group_coherent_ordered) {
 #endif
 
-    for (typename InstanceMap::iterator it = instance_map_.begin(), the_end = instance_map_.end(); it != the_end; ++it) {
-
-      const DDS::InstanceHandle_t handle = it->second;
-      const SubscriptionInstance_rch inst = get_handle_instance(handle);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+         iter != instances_.end();
+         ++iter) {
+      DDS::InstanceHandle_t handle = iter->first;
+      SubscriptionInstance_rch inst = iter->second;
 
       if (inst->instance_state_->match(view_states, instance_states)) {
         size_t i(0);
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -256,7 +256,7 @@ namespace OpenDDS {
     for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
         iter != instances_.end();
         ++iter) {
-      SubscriptionInstance_rch ptr = iter->second;
+      const SubscriptionInstance_rch ptr = iter->second;
 
       bool most_recent_generation = false;
 
@@ -1390,8 +1390,8 @@ DDS::ReturnCode_t take_i(MessageSequenceType& received_data,
     for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
          iter != instances_.end();
          ++iter) {
-      DDS::InstanceHandle_t handle = iter->first;
-      SubscriptionInstance_rch inst = iter->second;
+      const DDS::InstanceHandle_t handle = iter->first;
+      const SubscriptionInstance_rch inst = iter->second;
 
       if (inst->instance_state_->match(view_states, instance_states)) {
         size_t i(0);
--- a/tests/DCPS/GroupPresentation/SubscriberListener.cpp
+++ b/tests/DCPS/GroupPresentation/SubscriberListener.cpp
@@ -29,19 +29,10 @@ SubscriberListenerImpl::~SubscriberListenerImpl()
 }
 
 void
-SubscriberListenerImpl::on_data_on_readers(
-  DDS::Subscriber_ptr subs)
+SubscriberListenerImpl::on_data_on_readers(DDS::Subscriber_ptr subs)
 {
-  ::DDS::ReturnCode_t ret = subs->begin_access ();
-  if (ret != ::DDS::RETCODE_OK) {
-    ACE_ERROR((LM_ERROR,
-                  ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
-                  ACE_TEXT(" ERROR: begin_access failed!\n")));
-    ACE_OS::exit(-1);
-  }
-
   DDS::SubscriberQos qos;
-  ret = subs->get_qos (qos);
+  ::DDS::ReturnCode_t ret = subs->get_qos (qos);
   if (ret != ::DDS::RETCODE_OK) {
     ACE_ERROR((LM_ERROR,
                   ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
@@ -65,7 +56,13 @@ SubscriberListenerImpl::on_data_on_readers(
   CORBA::ULong len = readers->length ();
 
   if (qos.presentation.access_scope == ::DDS::GROUP_PRESENTATION_QOS) {
-
+    ret = subs->begin_access ();
+    if (ret != ::DDS::RETCODE_OK) {
+      ACE_ERROR((LM_ERROR,
+                 ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
+      ACE_TEXT(" ERROR: begin_access failed!\n")));
+      ACE_OS::exit(-1);
+    }
     ACE_DEBUG((LM_DEBUG,
                 ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers() ")
                 ACE_TEXT("GROUP_PRESENTATION_QOS get_datareaders returned %d readers!\n"),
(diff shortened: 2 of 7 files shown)


tests/DCPS/GroupPresentation/SubscriberListener.h
```diff
@@ -67,6 +67,7 @@ class SubscriberListenerImpl
 
   DDS::Subscriber_var subscriber_;
   bool  verify_result_;
+  ACE_Recursive_Thread_Mutex listener_lock_;
```
simpsont-oci, line 70: Does this really need to be recursive? I only see it being locked in on_data_on_readers(), which I wouldn't think calls back to itself?

tests/DCPS/GroupPresentation/Writer.cpp
```diff
@@ -143,7 +143,7 @@ Writer::svc()
   }
 
   finished_instances_ ++;
-
+  ACE_OS::sleep (1);
```
simpsont-oci, line 146: Is this still required? I think several of our other tests wait until subscriber match count is zero, which might be a better fit for the intended behavior here... "wait until the subscriber has gotten all our updates"

tests/DCPS/GroupPresentation/SubscriberListener.cpp
```diff
@@ -127,66 +127,52 @@ SubscriberListenerImpl::on_data_on_readers(
...
     subs->notify_datareaders ();
   }
-
   ret = subs->end_access ();
   if (ret != ::DDS::RETCODE_OK) {
     ACE_ERROR((LM_ERROR,
-                  ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
-                  ACE_TEXT(" ERROR: end_access failed!\n")));
+               ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
+    ACE_TEXT(" ERROR: end_access failed!\n")));
```
simpsont-oci, line 173: I think this line should line up with the one above, since it's still part of the ACE_ERROR message.

tests/DCPS/GroupPresentation/SubscriberListener.cpp
```diff
@@ -127,66 +127,52 @@ SubscriberListenerImpl::on_data_on_readers(
...
 
-      if (ret != DDS::RETCODE_OK) {
+      CORBA::ULong num_samples = si.length();
+      if (num_samples > 1 && (msg.length() != num_samples || num_samples < num_messages * 2)) {
         ACE_ERROR((LM_ERROR,
-                    ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
-                    ACE_TEXT(" ERROR: read failed!\n")));
-        ACE_OS::exit(-1);
+                   ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
+        ACE_TEXT(" ERROR: MessageSeq %d SampleInfoSeq %d != %d\n"),
```
simpsont-oci, line 156: indentation

tests/DCPS/GroupPresentation/SubscriberListener.cpp
```diff
@@ -127,66 +127,52 @@ SubscriberListenerImpl::on_data_on_readers(
...
+          ACE_ERROR((LM_ERROR,ACE_TEXT("%N:%l:%t: data reader[%d] has no data\n"),readers[i]->get_instance_handle()));
+          continue;
+        }
         ACE_ERROR((LM_ERROR,
-                    ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
-                    ACE_TEXT(" ERROR: MessageSeq %d SampleInfoSeq %d !\n"),
-                    msg.length(), si.length()));
-        this->verify_result_ = false;
+                   ACE_TEXT("%N:%l:%t: SubscriberListenerImpl::on_data_on_readers()")
+          ACE_TEXT(" ERROR: reader[%d] failed with %C\n"),
```
simpsont-oci, line 147: indentation

tests/DCPS/GroupPresentation/SubscriberListener.cpp
```diff
@@ -29,17 +31,16 @@ SubscriberListenerImpl::~SubscriberListenerImpl()
...
-  DDS::Subscriber_ptr subs)
+SubscriberListenerImpl::on_data_on_readers(DDS::Subscriber_ptr subs)
 {
   ::DDS::ReturnCode_t ret = subs->begin_access ();
   if (ret != ::DDS::RETCODE_OK) {
     ACE_ERROR((LM_ERROR,
-                  ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
-                  ACE_TEXT(" ERROR: begin_access failed!\n")));
+               ACE_TEXT("%N:%l: SubscriberListenerImpl::on_data_on_readers()")
+    ACE_TEXT(" ERROR: begin_access failed!\n")));
```
simpsont-oci, line 40: indentation

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1386,10 +1387,11 @@ DDS::ReturnCode_t take_i(MessageSequenceType& received_data,
...
 #endif
 
-    for (typename InstanceMap::iterator it = instance_map_.begin(), the_end = instance_map_.end(); it != the_end; ++it) {
-
-      const DDS::InstanceHandle_t handle = it->second;
-      const SubscriptionInstance_rch inst = get_handle_instance(handle);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+         iter != instances_.end();
+         ++iter) {
+      DDS::InstanceHandle_t handle = iter->first;
```
simpsont-oci, line 1393:
```suggestion
      const DDS::InstanceHandle_t handle = iter->first;
```

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1386,10 +1387,11 @@ DDS::ReturnCode_t take_i(MessageSequenceType& received_data,
...
+      SubscriptionInstance_rch inst = iter->second;
```
simpsont-oci, line 1394:
```suggestion
      const SubscriptionInstance_rch inst = iter->second;
```

tests/DCPS/GroupPresentation/SubscriberListener.cpp
```diff
@@ -127,66 +127,52 @@ SubscriberListenerImpl::on_data_on_readers(
...
-         || si[0].instance_state == DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE)
-
-      if (msg.length() != num_messages || si.length() != num_messages) {
+      ret = message_dr->take(msg, si, DDS::LENGTH_UNLIMITED,
+                             ::DDS::NOT_READ_SAMPLE_STATE,
+                             ::DDS::ANY_VIEW_STATE,
+                             ::DDS::ANY_INSTANCE_STATE);
+      if (ret != DDS::RETCODE_OK) {
+        if (ret == DDS::RETCODE_NO_DATA) {
+          ACE_ERROR((LM_ERROR,ACE_TEXT("%N:%l:%t: data reader[%d] has no data\n"),readers[i]->get_instance_handle()));
```
simpsont-oci, line 142:
```suggestion
          ACE_ERROR((LM_ERROR, ACE_TEXT("%N:%l:%t: data reader[%d] has no data\n"), readers[i]->get_instance_handle()));
```

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -253,9 +253,10 @@ namespace OpenDDS {
 
     const Observer_rch observer = get_observer(Observer::e_SAMPLE_READ);
 
-    const typename InstanceMap::iterator the_end = instance_map_.end();
-    for (typename InstanceMap::iterator it = instance_map_.begin(); it != the_end; ++it) {
-      const SubscriptionInstance_rch ptr = get_handle_instance(it->second);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
+        iter != instances_.end();
+        ++iter) {
+      SubscriptionInstance_rch ptr = iter->second;
```
simpsont-oci, line 259:
```suggestion
      const SubscriptionInstance_rch ptr = iter->second;
```

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -311,10 +312,10 @@ namespace OpenDDS {
 
     const Observer_rch observer = get_observer(Observer::e_SAMPLE_TAKEN);
 
-    const typename InstanceMap::iterator the_end = instance_map_.end();
-    for (typename InstanceMap::iterator it = instance_map_.begin(); it != the_end; ++it) {
-      DDS::InstanceHandle_t handle = it->second;
-      OpenDDS::DCPS::SubscriptionInstance_rch ptr = get_handle_instance(handle);
+    for (SubscriptionInstanceMapType::iterator iter = instances_.begin();
```
simpsont-oci, line 315: @mitza-oci what's the difference between instances_ and instance_map_? It looks like we're avoiding get_handle_instance() altogether this way (and the weakptr lock / lookup), but I can't tell if that's a good thing or a bad thing. I think you touched the code for these handles most recently, right?

simpsont-oci (changes requested):
A partial review... overall looks OK. I'm a little worried that we're not holding both mutexes in some places where we used to be, but I recognize that might have been unnecessary (as well as part of the deadlock issue when the lock order was inverted elsewhere).

dds/DCPS/DataReaderImpl_T.h
mitza-oci: `instances_` is `InstanceHandle_t` -> `SubscriptionInstance_rch` and comes from the non-templated base class
`instance_map_` is `MessageType` -> `InstanceHandle_t` and must be in the template since it uses `MessageType`

tests/DCPS/GroupPresentation/SubscriberListener.h
pmesnier: Probably not, but I was following a pattern, since pretty much all the other locks are recursive, I just went with it.

tests/DCPS/GroupPresentation/Writer.cpp
pmesnier: This is a deeper sleep. This is in the service object's thread function, and is giving the subscriber a chance to handle the end_coherence message before discarding  the association with the writer. In the subscriber app, there are multiple threads handling I/O events at the CORBA level. In the DataReaderImpl, there are two functions that serve as entry points from the infrastructure that come into conflict and motivated me to add the sleep. First is data_received() which serves as a message dispatcher. It tries very early on to inform the system that a given writer is still active. The second function that is accessed by the infrastructure is remove_associations(). The problem is that there are two infrastructure threads, and without the sleep, sometimes a thread obtains the end_coherency message but is swapped out by the OS before it can enter data_received(). Meanwhile the second thread detects a connection closure or whatever happens at the end of the writer's service function and it calls remove_associations() and cleans up the writer info before the first thread can process its message. When that first thread does get a chance to run, it has no way to dispatch the message as the necessary associations have been cleaned up. This sleep makes it so the end_coherency message is always processed before the associations are purged.

tests/DCPS/GroupPresentation/SubscriberListener.cpp
pmesnier: Yeah, that's my editor being "helpful."

dds/DCPS/DataReaderImpl_T.h
pmesnier: One of the problems I saw was that traversing the instance_map_ and then calling get_handle_instance() led to inconsistent ordering when  taking collections of samples. For instance the subscriberimpl::get_datareaders() function iterates over the instances_ map, and may evaluate instances with handles 7 and 9, but then the datareaderimpl::take() starts out with instance handle 8.

pmesnier:
> A partial review... overall looks OK. I'm a little worried that we're not holding both mutexes in some places where we used to be, but I recognize that might have been unnecessary (as well as part of the deadlock issue when the lock order was inverted elsewhere).

Yes, lock inversion was a major source of deadlocks. What would happen is that, for example, a thread would be updating a liveliness state and would first grab the instance lock on the datareaderimpl, then would enter the subscriberimpl via some stack and would then want to take the si_lock. But the si_lock is currently held by some other thread that wants to get a list of datareaders from the subscriberimpl and that would eventually visit the datareaderimpl that the first thread is accessing, and so when it tries to acquire the instance_lock, you now have a deadlock between these two threads, and if there are any more threads in the process, such as the main thread, that attempts to use either of these contended locks, well then the pile up just gets worse. 

So my remedy is to make stack based copies of the affected collections while holding the affective lock then releasing that lock and iterating over the local copy of the list. I believe this is safe because in the two cases where I do this, the elements of the local copy are RCH's so they should be good to go for the duration of the function at hand. However if this assumption is invalid, then some other strategy will be necessary to avoid deadlocks and also avoid resource poaching. That is why I added a lock to restrict access to subscriberListener::on_data_on_readers, to ensure that it had all the data is was expecting.

dds/DCPS/DataReaderImpl_T.h
pmesnier: done

tests/DCPS/GroupPresentation/Writer.cpp
pmesnier: I have removed this sleep. The need for it no longer exists on the subscriber side.

simpsont-oci (approved):

PR #2907, opened 2021-08-12
Liveliness Test: Automatic Reader receives incorrect liveliness change count

ClaytonCalabrese:
Problem: The Automatic reader was in rare cases getting interrupted while it was still handling the message that was received.  If the writer was shut down before the read/take was completed, an incorrect liveliness changed value would occur.

Solution: I introduced a way of tracking where the message has been read by the automatic data reader listener, and block the deletion of writer/publisher until the message has been handled.

### DIFF
--- a/tests/DCPS/LivelinessTest/LivelinessTest.cpp
+++ b/tests/DCPS/LivelinessTest/LivelinessTest.cpp
@@ -283,9 +283,8 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
     dcs->wait_for("driver", LOCAL_MANUAL_DATAREADER, "LIVELINESS_LOST_" + OpenDDS::DCPS::to_dds_string(i));
   }
 
-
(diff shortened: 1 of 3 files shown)


tests/DCPS/LivelinessTest/LivelinessTest.cpp
```diff
@@ -283,6 +283,11 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
     dcs->wait_for("driver", LOCAL_MANUAL_DATAREADER, "LIVELINESS_LOST_" + OpenDDS::DCPS::to_dds_string(i));
   }
 
+
+  while (automatic_drl_servant->samples_handled() < 1) {
+    ACE_OS::sleep(0.25);
```
simpsont-oci, line 288: build warnings are saying this is getting cast to zero on some platforms

simpsont-oci (changes requested):

PR #2909, opened 2021-08-13
Changed ACE_ERROR into a VDBG to not get this message on the console

jwillemsen:
* dds/DCPS/transport/framework/TransportSendStrategy.cpp:

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendStrategy.cpp
+++ b/dds/DCPS/transport/framework/TransportSendStrategy.cpp
@@ -1483,12 +1483,12 @@ TransportSendStrategy::direct_send(bool do_relink)
     } else if ((outcome == OUTCOME_PEER_LOST) ||
                (outcome == OUTCOME_SEND_ERROR)) {
       if (outcome == OUTCOME_SEND_ERROR) {
-        if (Transport_debug_level > 0) {
-          VDBG((LM_WARNING,
-                "(%P|%t) WARNING: Problem detected in "
-                "send buffer management: %p.\n",
-                "send_bytes"));
+        VDBG((LM_WARNING,
+              "(%P|%t) WARNING: Problem detected in "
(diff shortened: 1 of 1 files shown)


dds/DCPS/transport/framework/TransportSendStrategy.cpp
```diff
@@ -1483,10 +1483,10 @@ TransportSendStrategy::direct_send(bool do_relink)
     } else if ((outcome == OUTCOME_PEER_LOST) ||
                (outcome == OUTCOME_SEND_ERROR)) {
       if (outcome == OUTCOME_SEND_ERROR) {
-        ACE_ERROR((LM_WARNING,
-                   ACE_TEXT("(%P|%t) WARNING: Problem detected in ")
-                   ACE_TEXT("send buffer management: %p.\n"),
-                   ACE_TEXT("send_bytes")));
+        VDBG((LM_WARNING,
```
mitza-oci, line 1486: Should this be logged at a lower transport debug level?

mitza-oci (changes requested):

dds/DCPS/transport/framework/TransportSendStrategy.cpp
jwillemsen: Because this was an unconditional logging I decided to use VDBG just a few lines below, https://github.com/objectcomputing/OpenDDS/blob/500af9520332c16d7682299b42caf1d80282b797/dds/DCPS/transport/framework/TransportSendStrategy.cpp#L1495. I can make it dependent on a level, which one do you prefer?

mitza-oci: Since it was unconditional before, let's use level 1 here.

mitza-oci (approved):

PR #2911, opened 2021-08-16
Allow fallback for unanswered heartbeats

simpsont-oci:
Problem: In periods of high congestion, heartbeat traffic can become excessive.

Solution: Allow fallback for heartbeat periods when heartbeats go unanswered.


mitza-oci (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1469,12 +1469,13 @@ RtpsUdpDataLink::RtpsWriter::send_heartbeats(const MonotonicTimePoint& /*now*/)
   gather_heartbeats_i(meta_submessages);
 
   if (!preassociation_readers_.empty() || !lagging_readers_.empty()) {
-    heartbeat_->schedule(link->config().heartbeat_period_);
+    heartbeat_->schedule(fallback_.get());
```
jrw972, line 1472: I think you want to `advance` in the if and add
`else {
  fallback_.reset()`
}`

That puts it under the protection of `g` as well which is probably what you want.

jrw972 (changes requested):

PR #2912, opened 2021-08-16
Fix XML-related Compile Issues

iguessthislldo:
Follow up to #2896

### DIFF
--- a/dds/DCPS/security/AccessControl/Governance.cpp
+++ b/dds/DCPS/security/AccessControl/Governance.cpp
@@ -111,7 +111,7 @@ int Governance::load(const SSL::SignedDocument& doc)
 {
   std::string xml;
   doc.get_original_minus_smime(xml);
-  ParserPtr parser(get_parser(doc.filename(), xml));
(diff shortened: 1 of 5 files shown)


mitza-oci (approved):

dds/DCPS/security/AccessControl/Governance.cpp
```diff
@@ -111,7 +111,7 @@ int Governance::load(const SSL::SignedDocument& doc)
 {
   std::string xml;
   doc.get_original_minus_smime(xml);
-  ParserPtr parser(get_parser(doc.filename(), xml));
+  ParserPtr parser(move(get_parser(doc.filename(), xml)));
```
simpsont-oci, line 114: compiler warnings for moving a temporary... I guess clang wants you to make a named variable to move here?

iguessthislldo: I'm kinda fed up with `unqiue_ptr` right now, I just went with pass-by-reference.

simpsont-oci (approved):

PR #2914, opened 2021-08-17
Filter pre-discovery RTPS samples for non-durable readers

simpsont-oci:
Problem: non-durable RTPS can currently receive historical samples when writer cache (heartbeat window) is slow to update (e.g. when readers do not shut down cleanly).

Solution: non-durable RTPS readers should filter samples based on source timestamps (or received timestamps, if source timestamps are not available) in order to prevent delivery of samples generated prior to time of discovery / association.

### DIFF
--- a/dds/DCPS/DataSampleHeader.h
+++ b/dds/DCPS/DataSampleHeader.h
@@ -14,6 +14,8 @@
 #include "SequenceNumber.h"
 #include "RepoIdTypes.h"
 #include "Message_Block_Ptr.h"
+#include "TimeTypes.h"
+
 #include <iosfwd>
 
 #if !defined (ACE_LACKS_PRAGMA_ONCE)
@@ -260,9 +262,9 @@ struct OpenDDS_Dcps_Export DataSampleHeader : public PoolAllocationBase {
     }
   }
 
-  ACE_Time_Value get_source_timestamp() const
+  SystemTimePoint get_source_timestamp() const
   {
-    return ACE_Time_Value(source_timestamp_sec_, source_timestamp_nanosec_ / 1000);
+    return SystemTimePoint(ACE_Time_Value(source_timestamp_sec_, source_timestamp_nanosec_ / 1000));
   }
 
 private:
--- a/dds/DCPS/DataSampleHeader.h
+++ b/dds/DCPS/DataSampleHeader.h
@@ -260,6 +260,11 @@ struct OpenDDS_Dcps_Export DataSampleHeader : public PoolAllocationBase {
     }
   }
 
+  ACE_Time_Value get_source_timestamp() const
+  {
+    return ACE_Time_Value(source_timestamp_sec_, source_timestamp_nanosec_ / 1000);
+  }
+
 private:
   /// Keep track of the amount of data read from a buffer.
   size_t serialized_size_;
--- a/dds/DCPS/transport/framework/TransportClient.h
+++ b/dds/DCPS/transport/framework/TransportClient.h
@@ -143,7 +143,6 @@ public:
 
   SystemTimePoint creation_time() const
   {
-    ACE_Guard<ACE_Thread_Mutex> guard(lock_);
     return creation_time_;
   }
 
--- a/dds/DCPS/transport/framework/TransportClient.h
+++ b/dds/DCPS/transport/framework/TransportClient.h
@@ -141,6 +141,11 @@ public:
     return repo_id_;
   }
 
+  SystemTimePoint creation_time() const {
+    ACE_Guard<ACE_Thread_Mutex> guard(lock_);
+    return creation_time_;
+  }
+
 private:
 
   // Implemented by derived classes (DataReaderImpl/DataWriterImpl)
@@ -329,6 +334,7 @@ private:
   Reverse_Lock_t reverse_lock_;
 
   RepoId repo_id_;
+  const SystemTimePoint creation_time_;
 };
 
 typedef RcHandle<TransportClient> TransportClient_rch;
--- a/dds/DCPS/transport/framework/TransportClient.h
+++ b/dds/DCPS/transport/framework/TransportClient.h
@@ -141,7 +141,8 @@ public:
     return repo_id_;
   }
 
-  SystemTimePoint creation_time() const {
+  SystemTimePoint creation_time() const
+  {
     ACE_Guard<ACE_Thread_Mutex> guard(lock_);
     return creation_time_;
   }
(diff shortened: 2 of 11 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4323,7 +4323,10 @@ RtpsUdpDataLink::RtpsReader::deliver_held_data(const RepoId& src)
     const SequenceNumber ca = wi->second->recvd_.cumulative_ack();
     const WriterInfo::HeldMap::iterator end = wi->second->held_.upper_bound(ca);
     for (WriterInfo::HeldMap::iterator it = wi->second->held_.begin(); it != end; /*increment in loop body*/) {
-      to_deliver.push_back(it->second);
+      const DCPS::SystemTimePoint source_time(ACE_Time_Value(it->second.header_.source_timestamp_sec_, it->second.header_.source_timestamp_nanosec_ / 1000));
```
iguessthislldo, line 4326: This seems like something that should be a getter in `DataSampleHeader`. This along with a setter that would be used `deliver_sample_i`.

jrw972:
I assume we are talking about reliable writers and readers.
Yes.  If a reader dies, the writer may start accumulating and offering that to the reader.
The alternative is to set a starting point for that reader.
The problem is that precludes the use of non-directed heartbeats.
IMHO, the RTPS spec didn't think about this case and didn't provide enough in the protocol to address it.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1230,6 +1233,10 @@ RtpsUdpDataLink::RtpsWriter::customize_queue_element_helper(
     return 0;
   }
 
+  if (source_time == SystemTimePoint::zero_value) {
+    source_time == SystemTimePoint::now();
```
mitza-oci, line 1237:
```suggestion
    source_time = SystemTimePoint::now();
```

dds/DCPS/transport/framework/TransportClient.h
```diff
@@ -141,6 +141,11 @@ class OpenDDS_Dcps_Export TransportClient
     return repo_id_;
   }
 
+  SystemTimePoint creation_time() const {
```
mitza-oci, line 144:
```suggestion
  SystemTimePoint creation_time() const
  {
```

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -345,6 +345,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
     ACE_CDR::Long required_acknack_count_;
     OPENDDS_MAP(SequenceNumber, TransportQueueElement*) durable_data_;
     MonotonicTimePoint durable_timestamp_;
+    SystemTimePoint discovery_time_;
```
jrw972, line 348:
```suggestion
    const SystemTimePoint discovery_time_;
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -551,6 +553,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
     OPENDDS_MAP(SequenceNumber, RTPS::FragmentNumber_t) frags_;
     CORBA::Long heartbeat_recvd_count_, hb_frag_recvd_count_;
     const ACE_CDR::ULong participant_flags_;
+    SystemTimePoint discovery_time_;
```
jrw972, line 556:
```suggestion
    const SystemTimePoint discovery_time_;
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -631,12 +637,14 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
...
     WeakRcHandle<RtpsUdpDataLink> link_;
     const RepoId id_;
+    const bool durable_;
     WriterInfoMap remote_writers_;
     WriterInfoSet preassociation_writers_;
     bool stopping_;
     CORBA::Long nackfrag_count_;
     typedef PmfSporadicTask<RtpsReader> Sporadic;
     RcHandle<Sporadic> preassociation_task_;
+    SystemTimePoint creation_time_;
```
jrw972, line 647:
```suggestion
    const SystemTimePoint creation_time_;
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4323,7 +4331,10 @@ RtpsUdpDataLink::RtpsReader::deliver_held_data(const RepoId& src)
     const SequenceNumber ca = wi->second->recvd_.cumulative_ack();
     const WriterInfo::HeldMap::iterator end = wi->second->held_.upper_bound(ca);
     for (WriterInfo::HeldMap::iterator it = wi->second->held_.begin(); it != end; /*increment in loop body*/) {
-      to_deliver.push_back(it->second);
+      const DCPS::SystemTimePoint source_time(it->second.header_.get_source_timestamp());
```
jrw972, line 4334: Am I correct in that you are going filter on both sides?

jrw972 (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Yes, current plan is writer side filters based on association with remote readers, and reader side filters based on reader creation time (which should hopefully only come into play for interop).

dds/DCPS/DataSampleHeader.h
```diff
@@ -260,6 +260,11 @@ struct OpenDDS_Dcps_Export DataSampleHeader : public PoolAllocationBase {
     }
   }
 
+  ACE_Time_Value get_source_timestamp() const
```
iguessthislldo, line 263: Return a `SystemTimePoint`?

simpsont-oci: I thought it might be more flexible to have it return ACE_Time_Value in case the user needs a MonotonicTimePoint rather than SystemTimePoint? But maybe that's an unlikely use case...

iguessthislldo: No this should always be system time:

> `SystemTimePoint` should be used when dealing with the DDS API and timestamps on incoming and outgoing messages.

Also if there's a situation in OpenDDS code where a `ACE_Time_Value` can be either be system or monotonic then something is wrong with the design because it could break the kind of guarantees that the time types are trying to ensure. That's why I also say that we should only deal with `ACE_Time_Value` directly when interacting with ACE.

simpsont-oci: Gotcha, ok, I'll change it to SystemTimePoint.

iguessthislldo (approved):

PR #2917, opened 2021-08-18
No deserialization on size 0.

lij-oci:
check size == 0 before further deserialization

### DIFF
--- a/dds/idl/marshal_generator.cpp
+++ b/dds/idl/marshal_generator.cpp
@@ -3637,7 +3637,7 @@ namespace {
       extraction.endArgs();
       be_global->impl_ <<
         "  ACE_CDR::UShort disc, size;\n"
-        "  if (!(strm >> disc) || !(strm >> size)) {\n"
+        "  if (!(strm >> disc) || !(strm >> size) || size == 0) {\n"
         "    return false;\n"
         "  }\n"
         "  if (disc == OpenDDS::RTPS::PID_SENTINEL) {\n"


dds/idl/marshal_generator.cpp
```diff
@@ -3637,7 +3637,7 @@ namespace {
       extraction.endArgs();
       be_global->impl_ <<
         "  ACE_CDR::UShort disc, size;\n"
-        "  if (!(strm >> disc) || !(strm >> size)) {\n"
+        "  if (!(strm >> disc) || !(strm >> size) || size == 0) {\n"
```
mitza-oci, line 3640: Parameters of size 0 are legal, so this can return true.

mitza-oci (changes requested):

mitza-oci (approved):

mitza-oci:
Please add a unit test

dds/idl/marshal_generator.cpp
```diff
@@ -3645,6 +3645,10 @@ namespace {
         "    return true;\n"
         "  }\n"
         "  const Serializer::ScopedAlignmentContext sac(strm, size);\n"
+        "  if (size == 0) {\n"
+        "    uni._d(disc);\n"
+        "    return true;\n"
+        "  }\n"
```
mitza-oci, line 3651: Move this block up one line -- if this one returns `true` there is no need to construct a `ScopedAlignmentContext`

mitza-oci (changes requested):

dds/idl/marshal_generator.cpp
lij-oci: Right, I thought the same way. 

Two things made me doubt myself were:
1. The object sac was not explicitly used after creation:
const Serializer::ScopedAlignmentContext sac(strm, size);

2. The constructor takes a non-const reference of the first argument:
Serializer::ScopedAlignmentContext::ScopedAlignmentContext(Serializer& ser, size_t min_read)
  : ser_(ser)
  , max_align_(ser.encoding().max_align())
  , start_rpos_(ser.rpos())
  , rblock_(max_align_ ? (ptrdiff_t(ser.current_->rd_ptr()) - ser.align_rshift_) % max_align_ : 0)
  , min_read_(min_read)
  , start_wpos_(ser.wpos())
  , wblock_(max_align_ ? (ptrdiff_t(ser.current_->wr_ptr()) - ser.align_wshift_) % max_align_ : 0)
{
  ser_.reset_alignment();
}

The constructor does modify the Serializer parameter by calling reset_alignment();
So I was afraid that this might be needed to set the newly deserialized object right.

mitza-oci (approved):

PR #2918, opened 2021-08-19
Fix Wireshark Dissector with GLib 2.68, Perl Refactoring

iguessthislldo:
This is the GLib included with Ubuntu 21.04.

Continues work on problem discovered as part of https://github.com/objectcomputing/OpenDDS/pull/2916.

This GLib PR https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1715 added code with C++ linkage to GLib headers which we were wrapping with `extern "C"`. Our's and Wireshark's usage of `extern "C"` apparently didn't line up with what GLib developers thought we should be doing.

Tested with Wireshark 1.2.18, 1.6.16, and 3.4.4. Might be broken in other versions...


tests/cmake/README.md
```diff
@@ -5,6 +5,17 @@ CMake interoperability with the MPC-Generated OpenDDS codebase.
 
 ## Building and Running All the CMake Tests
 
+If you have a new enough CMake, this can be down with:
```
mitza-oci, line 8:
```suggestion
If you have a new enough CMake, this can be done with:
```
```suggestion
If you have a new enough CMake, this can be down with:
```

mitza-oci (changes requested):

tests/dissector/run_test.pl
```diff
@@ -0,0 +1,187 @@
...
+use lib "$ENV{ACE_ROOT}/bin";
+use lib "$ENV{DDS_ROOT}/bin";
+use PerlDDS::Run_Test;
+
+my $no_install = 0;
+my $help = 0;
+my $tshark_cmd = $ENV{TSHARK} // 'tshark';
+my $xdg_config_home = $ENV{XDG_CONFIG_HOME} // "$ENV{HOME}/.config";
+my $pcap = "$FindBin::Bin/messenger.pcap";
+$ENV{OPENDDS_DISSECTORS} = "$FindBin::Bin";
```
mitza-oci, line 26:
```suggestion
$ENV{OPENDDS_DISSECTORS} = $FindBin::Bin;
```

?

tests/dissector/run_test.pl
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/perl
```
mitza-oci, line 1: Should this use `env perl`?

mitza-oci (approved):

PR #2921, opened 2021-08-19
Lock held when writing relay participant topics

jrw972:
Problem
-------

Under heavy load, the handler thread of the RtpsRelay seems to block
when a participant is discovered or a participant times out.  This
shows up as network receive errors.  The timeout case indicates that a
problem exists regardless of extra messaging required for discovery.
The common link in both scenarios is that GuidPartition table is
updated which may lead to samples being written by the relay
participant.

Upon inspection, the lock protecting the GuidPartition table is held
while the samples are written.  Thus, the application participants
discovery thread acquires the lock, updates the GuidPartition table,
and then writes to the relay participant topics.  The handler is
unable to access the GuidPartition table during this transaction.

Solution
--------

Refactor the GuidPartition table to split the lock into one that
protects the table and one use for serializing writes.


mitza-oci (approved):

jrw972:
Merge to green-dev.

PR #2923, opened 2021-08-19
Improve RTPS messaging behaviors under IPv6

simpsont-oci:
Problem: When IPv6 support is enabled, the behavior for updating last_recv locators is wrong, and duplicate messages are sent out to each multicast address (v4 & v6), even when we know all discovered participants support & prefer IPv6. This causes more traffic than necessary (by twice as much!)

Solution: Use the last_recv_addr_ in our locator info to restrict outbound addresses to the preferred type, when applicable.


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4443,7 +4452,28 @@ RtpsUdpDataLink::accumulate_addresses(const RepoId& local, const RepoId& remote,
...
+              normal_addrs.insert(*it);
+            }
+          }
+        } else {
+          for (AddrSet::const_iterator it = mc_addrs.begin(); it != mc_addrs.end(); ++it) {
+            if (it->get_type() == AF_INET) {
+              normal_addrs.insert(*it);
+            }
+          }
+        }
```
iguessthislldo, line 4470:
```suggestion
        for (AddrSet::const_iterator it = mc_addrs.begin(); it != mc_addrs.end(); ++it) {
          if (it->get_type() == pos->second.last_recv_addr_.get_type()) {
            normal_addrs.insert(*it);
          }
        }
```

iguessthislldo:
Will this cause a problem if the remote goes from having IPv6 addresses to just having valid IPv4 addresses?

simpsont-oci:
> Will this cause a problem if the remote goes from having IPv6 addresses to just having valid IPv4 addresses?

Probably, yes, until the last_recv_addr expires and the address caching gets triggered again.

iguessthislldo:
Is the relay smoke failure in `test_u20_p1_asan_sec` related to this?

iguessthislldo (approved):
If it's not, I'm just going to

PR #2924, opened 2021-08-19
RtpsRelay PendingTimeout is unnecessary

jrw972:
Problem
-------

Determining the `-PendingTimeout` for the RtpsRelay is tedious.

Solution
--------

Compute the pending timeout by computing the exponentially weighted
moving (.99/.01) average discovery time of each client and then
multiply by a safety factor (10).  The initial expected discovery time
is the lifespan divided by the safety factor.

### DIFF
--- a/tools/rtpsrelay/RelayHandler.h
+++ b/tools/rtpsrelay/RelayHandler.h
@@ -88,7 +88,7 @@ public:
               RelayStatisticsReporter& relay_stats_reporter)
     : config_(config)
     , relay_stats_reporter_(relay_stats_reporter)
-    , expected_discovery_time_(config.lifespan() / 10)
+    , expected_discovery_time_(config.lifespan() * 0.1)
   {}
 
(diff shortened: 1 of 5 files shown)


tools/rtpsrelay/RelayHandler.h
```diff
@@ -83,6 +88,7 @@ class GuidAddrSet {
               RelayStatisticsReporter& relay_stats_reporter)
     : config_(config)
     , relay_stats_reporter_(relay_stats_reporter)
+    , expected_discovery_time_(config.lifespan() / 10)
```
iguessthislldo, line 91: The errors for this look like an oversight on my part. There are multiplication operators for `TimeDuration`, but not ones for division. So you could add them or do:

```suggestion
    , expected_discovery_time_(config.lifespan() * 0.1)
```

mitza-oci (approved):

jrw972:
I'm going to removing the pending queue.

PR #2928, opened 2021-08-21
RtpsRelay buffer size is not configurable

jrw972:
Problem
-------

The send and receive buffers for the RtpsRelay are not configurable.

Solution
--------

Expose an option and set the receive buffer size.


jrw972:
Merge to green-dev.

mitza-oci (approved):

PR #2930, opened 2021-08-21
Writers of Fragmented Data Samples Have Issues Reconnecting

simpsont-oci:
Problem: Writers of fragmented data have issues reconnecting after all readers disconnect (and pre_stop_helper is called) because "late" fragments fool the writer into thinking they have access to the full range of fragments for a given sequence number. When the reconnect occurs, the writer then issues an incorrect heartbeat range, and a reliable reader is never able to recover all the fragments required to deliver sample data.

Solution: advance a "minimum" SN for the send buffer when sequence numbers are acked, preventing futher inserts from being used with that sequence number. This blocks the "late" fragments and prevents the inaccurate heartbeat.

Note: A few other issues were identified and cleaned up in this PR as well, noticeably corruption of NackFrag submessages through reuse of a MetaSubmessage object, parameter shadowing for a few loop iterators, and faulty resends of un-requested fragment data when the correct fragment data wasn't available.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -893,6 +893,8 @@ RtpsUdpDataLink::RtpsWriter::pre_stop_helper(TqeVector& to_drop, bool true_stop)
     }
   }
 
+  send_buff_->pre_clear();
+
   g2.release();
   g.release();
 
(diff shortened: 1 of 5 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1670,6 +1672,8 @@ RtpsUdpDataLink::RtpsReader::process_data_i(const RTPS::DataSubmessage& data,
     return false;
   }
 
+  const GuidConverter conv(src);
```
mitza-oci, line 1675: Is this used anywhere?

simpsont-oci: Hmm, looks like a debugging remnant. Good catch!

mitza-oci (approved):

PR #2935, opened 2021-08-23
RtpsDiscovery buffer sizes are not configurable

jrw972:
Problem
-------

The send and receive buffer sizes for RtpsDiscovery (the SPDP socket
and SEDP socket) are not configurable.  Certain applications like the
application participant in the RtpsRelay may need to configure these
buffer sizes.

Solution
--------

Add configuration options for setting the buffer sizes.


jrw972:
Merge to green-dev

mitza-oci (approved):

PR #2936, opened 2021-08-23
RtpsRelay does not log discovery time

jrw972:
Problem
-------

The time required for the relay to discover a participant is an
important datum for measuring the performance of the RtpsRelay.
Currently, this datum is not logged.

Solution
--------

Measure the time from first generic SPDP announcement to the time the
participant was discovered.


tools/rtpsrelay/ParticipantListener.cpp
```diff
@@ -51,10 +53,11 @@ void ParticipantListener::on_data_available(DDS::DataReader_ptr reader)
         const auto p = guids_.insert(repoid);
         if (p.second) {
           if (config_.log_discovery()) {
-            ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) INFO: ParticipantListener::on_data_available add local participant %C %C\n"), guid_to_string(repoid).c_str(), OpenDDS::DCPS::to_json(data).c_str()));
+            const auto d = now - guid_addr_set_.get_first_spdp(repoid);
+            ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) INFO: ParticipantListener::on_data_available add local participant %C %C discovery time %d.%d\n"), guid_to_string(repoid).c_str(), OpenDDS::DCPS::to_json(data).c_str(), d.value().sec(), d.value().usec()));
```
mitza-oci, line 57: Check log formatting characters for these types.  `sec()` is `time_t` and `usec()` is `suseconds_t`

iguessthislldo: Looks like it's `%:`?

jrw972: I don't see a formated for suseconds_t.

mitza-oci: Convert the argument to a type (`int`) that we know will match what the varargs macro is looking for based on the format (`%d`).

jrw972:
Merge to green-dev.

mitza-oci (approved):

PR #2943, opened 2021-08-25
Add support for manually triggering gh_pages

jwillemsen:


jwillemsen:
OpenDDS.org is not updated after the release

mitza-oci:
> OpenDDS.org is not updated after the release

I think that was a problem with the release script or checklist (which still needs to be addressed) @iguessthislldo your thoughts?

iguessthislldo:
I'm not sure what happened here. I did have that mishap where it was using my fork instead of `objectcomputing/OpenDDS`, but I thought I reran all the steps with the corrected remote...

PR #2944, opened 2021-08-25
verify sequence length in deserialization

lij-oci:


dds/idl/marshal_generator.cpp
```diff
@@ -1006,6 +1006,10 @@ namespace {
         be_global->impl_ <<
           "  CORBA::ULong length;\n"
           << streamAndCheck(">> length");
+        be_global->impl_ <<
+          "  if (length > strm.length()) {\n"
```
iguessthislldo, line 1010: `length` and `strm.length()` measure different things. `length` is the element count in the sequence and `strm.length()` is the number of bytes left in the stream.

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
mitza-oci: That's by design -- it's a check (like the one we already have in XTypes) to make sure a very large number of elements, more than are possible to have in the stream, aren't allocated

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
```diff
@@ -50,11 +52,59 @@ void extract_Parameter_user_data(const std::string& user_data, const Encoding en
   }
 }
 
+// ---------- ---------- ---------- ---------- ----------
+template <typename Seq, typename Element>
+bool insert_sequence(Serializer& s, const std::vector<Element> v)
```
mitza-oci, line 57:
```suggestion
bool insert_sequence(Serializer& s, const std::vector<Element>& v)
```

mitza-oci (changes requested):

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
```diff
@@ -64,3 +114,31 @@ TEST(RtpsCoreTypeSupportImpl, ExtractParameter)
   extract_Parameter_user_data("");
   extract_Parameter_user_data("extract_Parameter_user_data");
 }
+
+// ---------- ---------- ---------- ---------- ----------
+TEST(RtpsCoreTypeSupportImpl, InsertFilterResult_t)
+{
+  std::vector<CORBA::Long> v;
```
mitza-oci, line 121: I think having `vector` involved is unnecessary complication.  Sequences are just like vectors, so just use the sequence itself instead of first using a vector and then copying each element into a corresponding sequence.

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
```diff
@@ -50,11 +52,59 @@ void extract_Parameter_user_data(const std::string& user_data, const Encoding en
...
+
+  Seq seq;
+  return s >> seq;
+}
+
+// ---------- ---------- ---------- ---------- ----------
+void extract_FilterResult_t(const std::vector<CORBA::Long> v)
+{
+  const CORBA::ULong length = static_cast<CORBA::ULong>(v.size());
+  ACE_Message_Block mb(length * 32 + 128);
```
mitza-oci, line 96: Where does this expression come from?  Why is it 32 and 128?

mitza-oci (changes requested):

dds/idl/marshal_generator.cpp
iguessthislldo: Then there should be a comment, and maybe an error message, saying what's going on.

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
lij-oci: I was trying to use vector as a generic way to pass elements to a generic sequence creation function. Since std::vector does not need explicit setting length as sequence does, we can easily test any length of sequences (including 0) using vector as input source for insertion, or as comparison for exertion. They are all changed to const references as suggested.

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
lij-oci: They are both used roughly to provide enough space in this test, instead of using the same sized block such as 1024. '32' is for each element (may be increased if bigger elements are used). '123' is an extra space for the sequence, including the zero length cases.

iguessthislldo: 32 bytes is 8 times the sizes of a 4 byte ULong. It could use `serialized_size` on the sequence to get the exact size it requires.

lij-oci: You are both right. Since this is not a generic sequence extracting function, there is no need to provide too much extra space for each element.

dds/idl/marshal_generator.cpp
```diff
@@ -1006,8 +1006,14 @@ namespace {
         be_global->impl_ <<
           "  CORBA::ULong length;\n"
           << streamAndCheck(">> length");
+        //Since the element size may differ between C++ representation and CDR stream,
+        //checking 'if (sizeof(T) * length > strm.length())' may not always work.
```
iguessthislldo, line 1010:
```suggestion
```

It's a bad idea to suggest CDR and C++ sizes can be interchangeable at all.

dds/idl/marshal_generator.cpp
```diff
@@ -1006,8 +1006,14 @@ namespace {
...
+        //The check here is to prevent very large sequences from being allocated.
```
iguessthislldo, line 1011:
```suggestion
        // The check here is to prevent very large sequences from being allocated.
```

tests/unit-tests/dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp
```diff
@@ -50,11 +52,59 @@ void extract_Parameter_user_data(const std::string& user_data, const Encoding en
...
+
+  Seq seq;
+  return s >> seq;
+}
+
+// ---------- ---------- ---------- ---------- ----------
+void extract_FilterResult_t(const std::vector<CORBA::Long>& v)
+{
+  const CORBA::ULong length = static_cast<CORBA::ULong>(v.size());
+  ACE_Message_Block mb(length * sizeof(CORBA::Long) + 128);
```
iguessthislldo, line 96: You shouldn't use `sizeof` for any CDR buffer size, because `sizeof` is for C++ objects, not CDR ones. Please use `serialized_size`.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #2950, opened 2021-08-27
Added DomainParticipantImpl::await_handle

pmesnier:
it is because of a race between the main thread and a service thread. This commit addresses that case.
It does this by adding a method to the Domain Participant servant impl to wait until it is positively signaled that a desired instance handle has been assigned for a given GUID. The new method is called await_handle() and is similar to the method lookup_handle().

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.h
+++ b/dds/DCPS/DomainParticipantImpl.h
@@ -38,7 +38,6 @@
 #include <ace/Null_Mutex.h>
 #include <ace/Thread_Mutex.h>
 #include <ace/Recursive_Thread_Mutex.h>
-#include <ace/Condition_Thread_Mutex.h>
 
 #if !defined (ACE_LACKS_PRAGMA_ONCE)
 #pragma once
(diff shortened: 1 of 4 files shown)


mitza-oci:
If we do end up going with a "wait" here, use a configurable timeout just like the `pending_timeout()` in `Service_Participant`.

dds/DCPS/DomainParticipantImpl.h
```diff
@@ -309,6 +309,15 @@ class OpenDDS_Dcps_Export DomainParticipantImpl
    */
   DDS::InstanceHandle_t lookup_handle(const GUID_t& id) const;
 
+  /**
+   * Similar to lookup_handle in that it will return a previously mapped handle,
+   * but will coorindate with assign_handle when a desired handle has not yet
+   * been mapped, but is expected to be. The optional max_wait argument can be
+   * supplied to limit the time spent waiting for a handle. It the wait times out
```
ClaytonCalabrese, line 316:
```suggestion
   * supplied to limit the time spent waiting for a handle. If the wait times out,
```

ClaytonCalabrese (approved):
Extremely minor spelling/grammar fix but otherwise looks good.

PR #2956, opened 2021-08-28
Only using doc-group3.

vermaete:
There is a rather big letter soup in and under DDS.
Having two ACE/TAO implemenations is a kind of freedom.
But having for the DOC group the vesion 6.x and 7.x is, IMHO, a bit too much.

What I understood from Johnny is that --doc-group3 is, or could be,
the most used option. [1]

This patch is update the README that wasn't mentioning the ACE 7.0 / TAO 3.0.
And removes the two switches in the configure file for the doc-group.

Feel free to apply or not for a next release.

[1] fd6813e#commitcomment-55427969


mitza-oci:
We (Object Computing, Inc.) are running 18 scoreboard builds using ACE 6.x / TAO 2.x versions.  We will eventually move those over to ACE 7.x / TAO 3.x, but we're not ready to do so at this time.

My impression is that most users of OpenDDS aren't too concerned with which ACE version is being used, as long as it works.  The current situation is that users have a few options to select from (if they do in fact care).

vermaete:
@mitza-oci I do understand that this is a process that should be care full handled.  You may close my request

Just to give a bit of background for my PR.
I do notice a slow change in mindset where in the past `it's working, don't change it` was the baseline in the stage-gate meetings.  Now it becomes more and more  'the latest is the greatest`.  And this not only for the toplevel (e.g. the application) but also for all layers below (e.g. opendds, ace, tao, ...).

br

Issue #2265, opened 2021-01-12
Add support for a customer error handler for the XML QoS files


jwillemsen:
At the moment the loading/parsing of the QoS XML files fails we get just a DDS::RETCODE_ERROR back, all detailed information why is logged by OpenDDS internally but we can't access that error information to provide back to the user.

Internally the QoS Xml support uses xercesc which provides the ErrorHandler class (see https://xerces.apache.org/xerces-c/apiDocs-3/classErrorHandler.html). Instead of exposing the xercesc errorhandler to the user code we would like to see that OpenDDS exposes its own Error Handler class the user can implement and and provide to OpenDDS so that user code stays independent of xercesc.

iguessthislldo:
I've had a similar desire for something like this for [PyOpenDDS](https://github.com/oci-labs/pyopendds/issues/14) for a while. This is something that might be possible after we revise logging in OpenDDS if we plan for it.

PR #2961, opened 2021-09-01
Extend QoS XML handlers to allow the user to pass its own XML_Error_Handler

jwillemsen:
Extend QoS XML handlers to allow the user to pass its own XML_Error_Handler. Also use the XML Helper as member instead of using a hidden global static object which is not needed and could cause cleanup issues.


jwillemsen:
Fixes https://github.com/objectcomputing/OpenDDS/issues/2265

mitza-oci (approved):

PR #2962, opened 2021-09-01
Fixed generation of unreachable code in operator<<

doug1234:


dds/idl/marshal_generator.cpp
```diff
@@ -1428,12 +1431,14 @@ namespace {
...
           indent << "}\n";
       }
-      be_global->impl_ <<
-        "  if (discard_flag) {\n"
-        "    strm.set_construction_status(Serializer::ElementConstructionFailure);\n"
-        "    return false;\n"
-        "  }\n"
-        "  return true;\n";
+	  if (!primitive) {
+          be_global->impl_ <<
```
iguessthislldo, line 1435: This can be simplified by merging this whole statement with the one that ends with `indent << "}\n";` above since it's the in the `else` block of `if (primitive)`.

dds/idl/marshal_generator.cpp
```diff
@@ -1339,9 +1339,12 @@ namespace {
       extraction.addArg("strm", "Serializer&");
       extraction.addArg("arr", wrapper.wrapped_type_name());
       extraction.endArgs();
-      be_global->impl_ <<
-        "  bool discard_flag = false;\n"
-        "  const Encoding& encoding = strm.encoding();\n"
+	  
+	  if (!primitive) {
```
iguessthislldo, line 1343: We don't accept tabs or trailing whitespace on lines. [Here's a list of the whitespace issues](https://github.com/objectcomputing/OpenDDS/pull/2962/checks?check_run_id=3489990433).

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
doug1234: Good suggestion.  Should be fixed now.

dds/idl/marshal_generator.cpp
doug1234: Tabs and trailing white space removed.

dds/idl/marshal_generator.cpp
```diff
@@ -1426,14 +1429,13 @@ namespace {
         }
         indent.erase(0, 2);
         be_global->impl_ <<
-          indent << "}\n";
+          indent << "}\n"
+            "  if (discard_flag) {\n"
+            "    strm.set_construction_status(Serializer::ElementConstructionFailure);\n"
+            "    return false;\n"
+            "  }\n"
+            "  return true;\n";
```
iguessthislldo, line 1437: My suggestion was a mistake in that it didn't go far enough and it's caused some pretty nasty errors in the tests that ran on this PR.

I didn't see/forgot about [this here](https://github.com/objectcomputing/OpenDDS/blob/b03436f9969b5a3430bdbaf2b41d6eaf6c8fad75/dds/idl/marshal_generator.cpp#L1368). It makes the code in the scope loop over each array element, so the change moved what used to be outside the loop  to be inside it, for example:

```C++
  for (CORBA::ULong i0 = 0; i0 < 2; ++i0) {
    // Do deserialization on elements
  } // End of Loop
  if (discard_flag) {
    return false;
  }
  return true;
```

became

```C++
  for (CORBA::ULong i0 = 0; i0 < 2; ++i0) {
    // Do deserialization on elements
  if (discard_flag) {
    return false;
  }
  return true;
  } // End of Loop
```

So I think statement needs to be separated again and the existing code from `indent << "}\n"` up to:

```C++
        string indent = "  ";
        NestedForLoops nfl("CORBA::ULong", "i", arr, indent);
```

needs put in it's own scope. So it would look like:

```C++
      } else { // Enum, String, Struct, Array, Sequence, Union
        {
          string indent = "  ";
          NestedForLoops nfl("CORBA::ULong", "i", arr, indent);
          // ...
          be_global->impl_ <<
            indent << "}\n";
        }
        be_global->impl_ <<
          "  if (discard_flag) {\n"
          "    strm.set_construction_status(Serializer::ElementConstructionFailure);\n"
          "    return false;\n"
          "  }\n"
          "  return true;\n";
      }
```

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
doug1234: I see what you are saying.  I'll take a look into it.

ClaytonCalabrese (changes requested):
I think this and the `marshal_generator::generate_dheader_code(` whitespace changes are all that I see left to do.

Thanks for your contribution towards OpenDDS!

dds/idl/marshal_generator.cpp
```diff
@@ -1339,11 +1339,14 @@ namespace {
...
-        "  bool discard_flag = false;\n"
+
+      if (!primitive) {
+          be_global->impl_ << "  bool discard_flag = false;\n";
+       }
+       be_global->impl_ <<
         "  const Encoding& encoding = strm.encoding();\n"
         "  ACE_UNUSED_ARG(encoding);\n";
-      marshal_generator::generate_dheader_code(
+       marshal_generator::generate_dheader_code(
```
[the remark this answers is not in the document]
iguessthislldo: I think this is the last bit left to fix.

ClaytonCalabrese:
Fred's comment up above still needs to be addressed. Should just be deleting an extra space.

mitza-oci:
> 
> 
> Fred's comment up above still needs to be addressed. Should just be deleting an extra space.

I added it

ClaytonCalabrese (approved):

doug1234:
Thanks for the help with the last couple errant spaces!  Do you guys have a format program you use for any future contributions?

iguessthislldo:
No, unfortunately we don't, but it's something that I've played around with before. The closest thing we have to that at the moment is [`tools/scripts/lint.pl`](https://github.com/objectcomputing/OpenDDS/blob/master/tools/scripts/lint.pl), but it's just regex based script that finds things like the trailing spaces and tabs.

iguessthislldo:
Also thanks for submitting the fix and putting up with my mistake!

PR #2971, opened 2021-09-04
Updated xercesc source download link

jwillemsen:


iguessthislldo:
This was changed to a more generic link as part of the fixes for all the broken link in #2972:

https://github.com/objectcomputing/OpenDDS/blob/59f55973ad2ee476af29df54a281440fd8dffdc6/quickstart/GettingStartedPi.html#L126-L128

PR #2972, opened 2021-09-06
Make Website Use Highest Versioned Release

### LINKED ISSUE #2970: Last release on OpenDDS.org should be 3.18, not 3.16.1
On OpenDDS.org the last release is listed as 3.16.1, but that should be 3.18, 3.16.1 seems to be a bugfix release which should be in the list with previous releases but the high lighted release should be 3.18 to my idea

iguessthislldo:
Fixes #2970

The releases will now be ordered by their version and will also hide older patch releases, so for example 3.16.1 will be shown, but 3.16.0 won't. The current list would look like:

Screenshot from 2021-09-05 15-47-31

Also:
- Make Favicon Square, so it Doesn't get Squished
- Make Jekyll Template Parsing More Strict
- Add a README which says how to build the site locally
- Regenerate Gemfile.lock because the existing versions couldn't be built on Ubuntu 21.04.
- Replace Some Hardcoded URLs with Jekyll variables.


ClaytonCalabrese (approved):

iguessthislldo:
@mitza-oci An better alternative to this is to have the release script sort the list of releases and give it to Jekyll as a JSON file in `_data`. This would give us the same fine control as having javascript code do it without the limitations listed [here](https://github.com/objectcomputing/OpenDDS/blob/59f55973ad2ee476af29df54a281440fd8dffdc6/_includes/latest_releases.html#L6-L13) (which includes not having javascript). Doing this for every release would have the added benefit of updating the website release list for micro releases, which we still haven't fixed in the release script.

PR #2973, opened 2021-09-07
Make error logs dependent on DCPS::DCPS_debug_level

jwillemsen:
* dds/DCPS/RTPS/Spdp.cpp:

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -946,9 +946,11 @@ Spdp::data_received(const DataSubmessage& data,
 #endif
 
   if (!ParameterListConverter::from_param_list(plist, pdata)) {
-    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: Spdp::data_received - ")
-      ACE_TEXT("failed to convert from ParameterList to ")
-      ACE_TEXT("SPDPdiscoveredParticipantData\n")));
+    if (DCPS::DCPS_debug_level > 0) {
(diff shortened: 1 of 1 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -4202,10 +4272,11 @@ void Spdp::SpdpTransport::thread_status_task(const DCPS::MonotonicTimePoint& /*n
     StatusMap running;
     StatusMap removed;
     if (!local_thread_status_manager_.sync_with_parent(
-          *global_thread_status_manager_, running, removed) &&
-        DCPS::DCPS_debug_level) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: Spdp::SpdpTransport::thread_status_task: "
-        "syncing with global thread status manager failed\n"));
+          *global_thread_status_manager_, running, removed)) {
+      if (DCPS::DCPS_debug_level > 0) {
```
mitza-oci, line 4276: Why is this changing?  The change doesn't seem to have any effect.

mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
jwillemsen: The return was done in the past only when debug level was set

mitza-oci: Right, I didn't see that detail before.

mitza-oci (approved):

dds/DCPS/RTPS/Spdp.cpp
jwillemsen: This was tricky code, took me also some time to see the initial issue when checking the debug log level in this file

PR #2975, opened 2021-09-07
Make the `auto_run_tests.pl` Names of CMake Tests GHA Friendly

iguessthislldo:
Currently `ctest-to-auto-run-tests.py` generates names to mimic `auto_run_tests.pl` names based on the `CMakeLists.txt` directory relative to `tests/cmake` to differentiate it from non-CMake tests. This path is then appended with the test command. The current GitHub Actions test infrastructure doesn't like this scheme when these tests fail for two reasons:

1. The DevGuide Messenger tests have `..` in the path/name:
    - `tests/cmake/../../DevGuideExamples/DCPS/Messenger/run_test.pl`
2. In the case of the non-DevGuide Messengers, the path to the `run_test.pl` doesn't exist in the repository because it's copied from the sources in `tests/DCPS` to where the test is being built. For the annotations to work the file actually has to exist in the git repository. For example:
    - `tests/cmake/Messenger/Messenger_1/run_test.pl`

So instead of using the path to `run_test.pl`, use the path to the `CMakeLists.txt` file followed by the `run_test.pl` command. This will always be different from any possible non-CMake test and should always exist in the git repository. The previously shown `auto_run_tests.pl` names are now:

1. `DevGuideExamples/DCPS/Messenger/CMakeLists.txt run_test.pl`
2. `tests/cmake/Messenger/Messenger_1/CMakeLists.txt run_test.pl`

Also delete `tests/cmake/run_ci_tests.pl` because it shouldn't be used anymore.


iguessthislldo:
Success:

Screenshot from 2021-09-07 19-36-01

Screenshot from 2021-09-07 19-35-15

ClaytonCalabrese (approved):

PR #2981, opened 2021-09-08
fix SpdpTransport::handle_input coverity warning

lij-oci:


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2853,6 +2853,12 @@ Spdp::SpdpTransport::choose_recv_socket(ACE_HANDLE h) const
   return unicast_socket_;
 }
 
+bool valid_size(const ACE_INET_Addr& a)
+{
+  const int max_size = 16;
```
mitza-oci, line 2858: Why is this 16?

mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
lij-oci: I tried separate max (4 bytes for IPv4, 16 bytes for IPv6) based on ACE_HAS_IPV6 and get_type(). Somehow it didn't work. Simply using 16 bytes here worked. I tested this with the RtpsDiscovery test.

mitza-oci: There are two separate things to consider here:
1. We don't want to have a "magic number" in the code without any explanation for why this is the magic number
2. This magic number won't work IPv6 is larger, and it's bound to be platform dependent so hard to know if a single number could ever work for all platforms.

Take a look at what ACE is doing in INET_Addr, specifically in methods like `ACE_INET_Addr::set_address`.  The `get_size()` value is computed based on `sizeof` operators.  The same approach is needed here.

simpsont-oci (commented):
Looking at the Coverity issue for this, it seems like a better approach would be to put a sanity check for the size in ACE itself, rather than having Spdp double-check the sizes coming out of ACE.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2979,7 +2985,12 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
         DCPS::RcHandle<Spdp> outer = outer_.lock();
 
         if (outer) {
-          outer->data_received(data, plist, remote);
+          if (valid_size(remote)) {
```
mitza-oci, line 2988: Move this up as far as it can go -- in other words, why not check `remote`'s validity right after that object is written to?

mitza-oci:
> 
> 
> Looking at the Coverity issue for this, it seems like a better approach would be to put a sanity check for the size in ACE itself, rather than having Spdp double-check the sizes coming out of ACE.

Ideally, yes.  Practically, let's keep the check in OpenDDS.

dds/DCPS/RTPS/Spdp.cpp
lij-oci: 1. ACE_INET_Addr::set_address() sets addr_size_ to 4 bytes for ip4 and 16 bytes for ip6.

2. ACE_Addr::get_size() does not use sizeof operator:
ACE_Addr::get_size (void) const
{
  return this->addr_size_;
}

3. Therefore, a checking function like this should work across platforms:
bool valid_size(const ACE_INET_Addr& a)
{
  const int in6_size = 16;
  return a.get_size() <= in6_size;
}

4. The function that uses sizeof operator is get_addr_size():
ACE_INLINE int
ACE_INET_Addr::get_addr_size (void) const
{
  ACE_TRACE ("ACE_INET_Addr::get_addr_size");
#if defined (ACE_HAS_IPV6)
  if (this->get_type () == PF_INET)
    return sizeof this->inet_addr_.in4_;
  else
    return sizeof this->inet_addr_.in6_;
#else
  return sizeof this->inet_addr_.in4_;
#endif /* ACE_HAS_IPV6 */
}

5. I agree with Tim's comment that a sanity check should be in ACE object, and I would prefer the good c++ practice, i.e. constructor puts an object in valid state. I also found a lot of design issues in ACE_Addr and its favorite child ACE_INET_Addr. We have 4 options (from ideal to practical): 1) Replace ACE with std; 2) Improve ACE; 3) Wrap ACE to remove the dependency on ACE from the majority of our code; and 4) Make patches like this or the LogAddr I did before.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2853,6 +2853,12 @@ Spdp::SpdpTransport::choose_recv_socket(ACE_HANDLE h) const
   return unicast_socket_;
 }
 
+bool valid_size(const ACE_INET_Addr& a)
+{
+  const int ipv6_size = 16;
```
mitza-oci, line 2858: This doesn't work for IPv6.  On one Linux system I tried, `sizeof(sockaddr_in6)` was 28.  So like I was saying before, this needs to be based on `sizeof` so it agrees with the system's headers.

mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
lij-oci: Ok, we have no control over where or how the protected member variable addr_size_ is set, since there are too many functions, such as ACE_Addr::set_size, ACE_Addr::base_set, ACE_INET_Addr::set_address, ACE_INET_Addr::set(). Also any derived class can directly modify addr_size_ (talking about design issues).

ACE_INET_Addr::set() does use the sizeof operator to set addr_size_ which is totally different meaning of addr_size_ as in ACE_INET_Addr::set_address().

ACE_INET_Addr::set()
{
//...
this->base_set (AF_INET6, sizeof this->inet_addr_.in4_);
//...
this->inet_addr_.in6_.sin6_len = sizeof this->inet_addr_.in6_;
//...
this->set_size (sizeof this->inet_addr_.in6_);
//...
}

Based on the above, valid_size is modified to compare addr_size_ with the appropriate sizeof operator.

mitza-oci (approved):

PR #2988, opened 2021-09-09
Fix for occasional timeout in MultiTopic test.

pmesnier:
Ensure the query plan for the desired topic is fully initialized upon receiving data related to it.

### DIFF
--- a/dds/DCPS/MultiTopicDataReaderBase.cpp
+++ b/dds/DCPS/MultiTopicDataReaderBase.cpp
@@ -136,18 +136,20 @@ void MultiTopicDataReaderBase::init(const DDS::DataReaderQos& dr_qos,
     }
 
 
-    DDS::DataReader_var incoming =
-      parent->create_datareader(t, DATAREADER_QOS_USE_TOPIC_QOS,
-                                listener_.get(), ALL_STATUS_MASK);
-    if (!incoming.in()) {
+    QueryPlan& qp = query_plans_[selection[i]];
+    {
+      ACE_WRITE_GUARD(ACE_RW_Thread_Mutex, write_guard, qp.lock_);
+      qp.data_reader_ =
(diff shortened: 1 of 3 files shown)


dds/DCPS/MultiTopicDataReaderBase.cpp
```diff
@@ -136,18 +136,20 @@ void MultiTopicDataReaderBase::init(const DDS::DataReaderQos& dr_qos,
...
-                                listener_.get(), ALL_STATUS_MASK);
-    if (!incoming.in()) {
+    QueryPlan& qp = query_plans_[selection[i]];
+    {
+      ACE_WRITE_GUARD(ACE_RW_Thread_Mutex, write_guard, qp_lock_);
+      qp.data_reader_ =
+        parent->create_datareader(t, DATAREADER_QOS_USE_TOPIC_QOS,
+                                  listener_.get(), ALL_STATUS_MASK);
+    }
+    if (!qp.data_reader_.in()) {
```
mitza-oci, line 146: If this fails it leaves a new entry in `query_plans_`.  Does that need to get cleaned up?

mitza-oci (changes requested):

dds/DCPS/MultiTopicDataReaderBase.cpp
pmesnier: That's a good question. the code handling the nil data_reader_ is old. Is this the only place where query_plans_ elements are created? Could query_plans be created elsewhere or elsewhen and the data_reader_ initialized later?

pmesnier: After  some analysis and review, I conclude that removing the entry here is no benefit as there are a few places where query_plans_ is directly indexed with the brackets operator meaning that if this topic is looked up again, a new entry will be created at that time.

mitza-oci (approved):

PR #2989, opened 2021-09-10
RtpsRelay does not log subject name

jrw972:
Problem
-------

When using security, the subject name of a participant can provide
useful information when debugging connectivity issues.

Solution
--------

Extract the subject name and log it.

### DIFF
--- a/dds/DCPS/security/AuthenticationBuiltInImpl.h
+++ b/dds/DCPS/security/AuthenticationBuiltInImpl.h
@@ -36,14 +36,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace Security {
 
-const std::string Identity_Status_Token_Class_Id("DDS:Auth:PKI-DH:1.0");
-const std::string Auth_Peer_Cred_Token_Class_Id("DDS:Auth:PKI-DH:1.0");
+const char Identity_Status_Token_Class_Id[] = "DDS:Auth:PKI-DH:1.0";
+const char Auth_Peer_Cred_Token_Class_Id[] = "DDS:Auth:PKI-DH:1.0";
 
-const std::string dds_cert_sn("dds.cert.sn");
-const std::string dds_cert_algo("dds.cert.algo");
+const char dds_cert_sn[] = "dds.cert.sn";
+const char dds_cert_algo[] = "dds.cert.algo";
 
-const std::string dds_ca_sn("dds.ca.sn");
-const std::string dds_ca_algo("dds.ca.algo");
+const char dds_ca_sn[] = "dds.ca.sn";
+const char dds_ca_algo[] = "dds.ca.algo";
 
 /**
 * @class AuthenticationBuiltInImpl
--- a/dds/DCPS/security/AuthenticationBuiltInImpl.h
+++ b/dds/DCPS/security/AuthenticationBuiltInImpl.h
@@ -36,6 +36,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace Security {
 
+const std::string Identity_Status_Token_Class_Id("DDS:Auth:PKI-DH:1.0");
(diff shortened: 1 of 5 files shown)


dds/DCPS/security/AuthenticationBuiltInImpl.h
```diff
@@ -36,6 +36,14 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
 namespace Security {
 
+const std::string Identity_Status_Token_Class_Id("DDS:Auth:PKI-DH:1.0");
+const std::string Auth_Peer_Cred_Token_Class_Id("DDS:Auth:PKI-DH:1.0");
+
+const std::string dds_cert_sn("dds.cert.sn");
+const std::string dds_cert_algo("dds.cert.algo");
+
+const std::string dds_ca_sn("dds.ca.sn");
+const std::string dds_ca_algo("dds.ca.algo");
```
mitza-oci, line 46: For all of these.... no need for `std::string` namespace-scoped constants, especially want to keep them out of headers.
```
const char Identity_Status_Token_Class_Id[] = "DDS:Auth:PKI-DH:1.0";
```

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -907,6 +908,87 @@ SpdpHandler::SpdpHandler(const Config& config,
...
+          return "";
+        }
+
+        if (writerId != OpenDDS::DCPS::ENTITYID_SPDP_BUILTIN_PARTICIPANT_WRITER) {
+          // Not our message: this could be the same multicast group used
+          // for SEDP and other traffic.
+          break;
+        }
+
+        if (submessage_header.flags & (OpenDDS::RTPS::FLAG_D)) {
```
mitza-oci, line 945: At this point the sequence number has been parsed, so next up is the inlineQos.  If `FLAG_Q` is true, we expect an inlineQos element -- but there's one more wrinkle.  The `octetsToInlineQos` has to be checked (even if no `FLAG_Q`) in case the sender decided to put some additional optional fields in there (a new minor version may add them).

See if other places that parse this have some code that could be borrowed or added to MessageParser.

mitza-oci (changes requested):

tools/rtpsrelay/RelayHandler.cpp
mitza-oci: Specifically see `bool operator>>(Serializer& strm,  ::OpenDDS::RTPS::DataSubmessage& stru)` in `RtpsCoreTypeSupportImpl.cpp`

```
  if (!strm.skip(stru.octetsToInlineQos - 16)) {
    return false;
  }
  if ((stru.smHeader.flags & 2) && !(strm >> stru.inlineQos)) {
    return false;
  }
```

mitza-oci (approved):

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -907,6 +908,98 @@ SpdpHandler::SpdpHandler(const Config& config,
...
+          }
+          message_parser >> options;
+          // bit 8 in encap is on if it's PL_CDR_LE
+          message_parser.serializer().swap_bytes(((encap & 0x100) >> 8) != ACE_CDR_BYTE_ORDER);
+          if (!(message_parser >> plist)) {
+            ACE_ERROR((LM_ERROR,
+                       ACE_TEXT("(%P|%t) ERROR: extract_common_name() - ")
+                       ACE_TEXT("failed to deserialize data payload for SPDP\n")));
+            return "";
+          }
```
iguessthislldo, line 975: This can be replaced by deserializing an `EncapsualtionHeader` and getting the encoding from that.

iguessthislldo (changes requested):

tools/rtpsrelay/RelayHandler.cpp
mitza-oci: Let's handle that in a separate PR.

PR #2991, opened 2021-09-11
XTypes: Implement DynamicData

sonndinh:


dds/DCPS/XTypes/DynamicData.h
```diff
@@ -0,0 +1,217 @@
...
+  DDS::ReturnCode_t get_string_values(StringSeq& value, MemberId id) const;
+  DDS::ReturnCode_t set_string_values(MemberId id, const StringSeq& value);
+
+  typedef Sequence<DCPS::WString> WStringSeq;
+  DDS::ReturnCode_t get_wstring_values(WStringSeq& value, MemberId id) const;
+  DDS::ReturnCode_t set_wstring_values(MemberId id, const WStringSeq& value);
+
+  // Skip the whole part of the data stream that corresponds to this type.
+  // This is called by a containing type when it wants to skip a member which
+  // is an object of this type.
```
iguessthislldo, line 169:
```suggestion
  /// Skip the whole part of the data stream that corresponds to this type.
  /// This is called by a containing type when it wants to skip a member which
  /// is an object of this type.
```

Too make this Doxygen comments. Please do the rest with other comments like this.

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+  if (it != offset_lookup_table_.cend()) {
+    size_t offset = it->second;
+    std::ostringstream err_msg;
+    err_msg << "Failed to skip to the member with ID " << id;
+    if (!skip("get_int32_value", err_msg.str().c_str(), offset)) {
+      return DDS::RETCODE_ERROR;
+    }
+
+    if (!(strm_ >> value)) {
+      if (DCPS_debug_level >= 10) {
```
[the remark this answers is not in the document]
sonndinh: The guideline in debug.h says it should be 10 if the logs may happen more than once per sample. It seems to be the case for this class because it operates on a single sample and can generate multiple logs.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+{
+}
+
+DDS::ReturnCode_t DynamicData::get_int32_value(ACE_CDR::Long& value, MemberId id) const
+{
+  const OPENDDS_MAP<MemberId, size_t>::const_iterator it = offset_lookup_table_.find(id);
+  if (it != offset_lookup_table_.cend()) {
+    size_t offset = it->second;
+    std::ostringstream err_msg;
+    err_msg << "Failed to skip to the member with ID " << id;
```
jwillemsen, line 67: This looks pretty costly, use a ostreamstream to construct an error string to pass to skip, where it is only used when there is an error and debug level >= 10

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+      }
+      return DDS::RETCODE_ERROR;
+    }
+
+    reset_rpos();
+    return DDS::RETCODE_OK;
+  }
+
+  DynamicTypeMember_rch member;
+  DDS::ReturnCode_t retcode = type_->get_member(member, id);
```
jwillemsen, line 85: const?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+  if (retcode != DDS::RETCODE_OK) {
+    if (DCPS_debug_level >= 10) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) DynamicData::get_int32_value -")
+                 ACE_TEXT(" Failed to get DynamicTypeMember with ID %d\n"), id));
```
jwillemsen, line 89: Log retcode as string?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+    return retcode;
+  }
+
+  MemberDescriptor member_desc;
+  member->get_descriptor(member_desc);
+
+  if (member_desc.type->get_kind() != TK_INT32) {
+    if (DCPS_debug_level >= 10) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) DynamicData::get_int32_value -")
+                 ACE_TEXT(" Member with ID %d is not an int32\n"), id));
```
jwillemsen, line 100: Maybe log which kind it is, would help logging. Maybe even log kind as a readable string

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,1068 @@
...
+    return true;
+  } else {
+    return skip_collection_member(TK_MAP);
+  }
+}
+
+bool DynamicData::skip_collection_member(TypeKind kind)
+{
+  const char* kind_str = 0;
+  if (kind == TK_SEQUENCE) {
```
jwillemsen, line 798: case? Maybe use a helper function for the string translation just like we have for retcode?

dds/DCPS/XTypes/DynamicData.cpp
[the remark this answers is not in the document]
iguessthislldo: The list in debug.h isn't a complete guide because it doesn't mention errors at all. I think it assumes errors are unconditionally logged as that is the case in a lot of the older code and it puts warnings at 1. So I don't think that list really applies here because errors should be higher priority than normal debug logging. For example I have [errors at 1 to 3 in the security logging scheme](https://github.com/objectcomputing/OpenDDS/blob/1fd77cf02fd6f3ed9f1fb3c873455619816d4e75/dds/DCPS/debug.cpp#L107-L109). Thinking about it again, it probably doesn't have to be at level 1 if there is another log point down the line at 1 to report ultimately that an error happened. I don't have a concrete number to tell you, but theoretically if we wanted to put a normal debug statement in this function, then it'd have to have the same priority as an error, which doesn't seem right to me.

Also I'm not sure how this happens more than once per sample. If there an error in deserialization we won't come back here, right? So if this wasn't an actually error, I would say should be at debug level 8.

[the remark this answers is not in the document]
sonndinh: > The list in debug.h isn't a complete guide because it doesn't mention errors at all. I think it assumes errors are unconditionally logged as that is the case in a lot of the older code and it puts warnings at 1. So I don't think that list really applies here because errors should be higher priority than normal debug logging. For example I have [errors at 1 to 3 in the security logging scheme](https://github.com/objectcomputing/OpenDDS/blob/1fd77cf02fd6f3ed9f1fb3c873455619816d4e75/dds/DCPS/debug.cpp#L107-L109). Thinking about it again, it probably doesn't have to be at level 1 if there is another log point down the line at 1 to report ultimately that an error happened. I don't have a concrete number to tell you, but theoretically if we wanted to put a normal debug statement in this function, then it'd have to have the same priority as an error, which doesn't seem right to me.
> 
Yeah, this makes sense. I can change this to log unconditionally since I think this is an error and should be logged regardless of the debug level.

> Also I'm not sure how this happens more than once per sample. If there an error in deserialization we won't come back here, right? So if this wasn't an actually error, I would say should be at debug level 8.
>
So a user can call the same method more than once even though there is an error in the first call. It doesn't make much sense if they do so, but we don't disallow that.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -58,135 +58,27 @@ DynamicData DynamicData::clone() const
...
-      if (DCPS_debug_level >= 10) {
-        ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) DynamicData::get_int32_value -")
-                   ACE_TEXT(" Failed to read the int32 member with ID %d\n"), id));
-      }
-      return DDS::RETCODE_ERROR;
-    }
-  } else {
-    // TODO: Support encodings other than XCDR2?
-    return DDS::RETCODE_UNSUPPORTED;
+  bool good = strm_ >> value;
```
jwillemsen, line 70:
``` suggestion
  const bool good = strm_ >> value;
```

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -58,135 +58,27 @@ DynamicData DynamicData::clone() const
...
+  if (!good) {
+    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) DynamicData::get_int32_value -")
```
jwillemsen, line 72: Check debug level

dds/DCPS/XTypes/DynamicData.cpp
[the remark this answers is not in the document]
iguessthislldo: > Yeah, this makes sense. I can change this to log unconditionally since I think this is an error and should be logged regardless of the debug level.

I meant that because we don't always unconditionally log errors anymore, the list in debug.h isn't completely representative of what our current practice is. If this error status is going to returned to the user, then it definitely shouldn't be logged unconditionally.

ClaytonCalabrese:
```
DDS::ReturnCode_t get_member(
 inout DynamicTypeMember member,
 in MemberId id);
```

Looks like I should have made MemberId a const when I was making DynamicType.  As a result I think all(?) of the MemberId in DynamicData.h/.cpp would turn into consts, which makes sense considering these shouldn't be changing the ID.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -577,135 +739,150 @@ DDS::ReturnCode_t DynamicData::get_single_value(ValueType& value, MemberId id,
...
+          ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) DynamicData::get_char_common -")
                      ACE_TEXT(" Failed to read wstring with ID %d\n"), id));
         }
         good = false;
         break;
       }
       ACE_CDR::ULong index;
-      if (!get_index_from_id(id, index, str.length())) {
+      const size_t str_len = ACE_OS::strlen(str);
+      if (!get_index_from_id(id, index, str_len)) {
```
ClaytonCalabrese, line 828: CI on W19 builds warns about possible loss of value when converting str_len to a ULong.

ClaytonCalabrese (changes requested):

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -0,0 +1,277 @@
...
+  DynamicData();
+
+  /// This creates a duplicated ACE_Message_Block chain from the provided chain.
+  /// The duplicated chain is released when the object is destroyed. Caller is
+  /// responsible for the release of the input message block chain.
+  DynamicData(ACE_Message_Block* chain,
+              const DCPS::Encoding& encoding,
+              const DynamicType_rch& type);
+  DynamicData(DCPS::Serializer& ser, const DynamicType_rch& type);
+  DynamicData& operator=(const DynamicData& other);
```
mitza-oci, line 48: Since it now has copy assignment it should also have a copy constructor.

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -0,0 +1,277 @@
...
+
+  ~DynamicData();
+
+  DDS::ReturnCode_t get_descriptor(MemberDescriptor& value, MemberId id) const;
+  DDS::ReturnCode_t set_descriptor(MemberId id, const MemberDescriptor& value);
+
+  MemberId get_member_id_by_name(DCPS::String name) const;
+  MemberId get_member_id_at_index(ACE_CDR::ULong index);
+  ACE_CDR::ULong get_item_count();
```
mitza-oci, line 57: These two both look like they should be const members.

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -0,0 +1,277 @@
...
+
+  DynamicData clone() const;
+
+  DDS::ReturnCode_t get_int32_value(ACE_CDR::Long& value, MemberId id);
```
mitza-oci, line 61: Should this whole group (both `get_*_value` and `get_*_values`) be const members?

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -252,18 +232,41 @@ class OpenDDS_Dcps_Export DynamicData {
   /// Skip a member which is an union.
   bool skip_union_member(const DynamicType_rch& type);
 
-  // These methods can be moved to DynamicType-related classes.
+  // TODO: This method can be moved to DynamicType-related classes.
```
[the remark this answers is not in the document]
sonndinh: Right, we can have a separate PR to move it later.

dds/DCPS/XTypes/DynamicData.h
sonndinh: Unfortunately these functions need to update the associated Serializer object, which is a member of the class. So I can't make them const members.

dds/DCPS/XTypes/DynamicData.h
sonndinh: For the same reason as the get_*_value and get_*_values functions, these two examine the underlying data to determine the number of items and compute the id. They modify the Serializer data member and can't be const.

dds/DCPS/XTypes/DynamicData.h
mitza-oci: I don't think we need to do anything for it in this PR, but it may be good to revisit later.  `ACE_Message_Block::duplicate()` is const and Serializers can each be local to the method that needs them.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -0,0 +1,2169 @@
...
+}
+
+DynamicData::~DynamicData()
+{
+  ACE_Message_Block::release(chain_);
+}
+
+void DynamicData::copy(const DynamicData& other)
+{
+  chain_ = other.chain_->duplicate();
```
mitza-oci, line 81: Since `chain_` may be non-null when entering this method, this needs an `ACE_Message_Block::release()` first.

dds/DCPS/XTypes/DynamicData.h
sonndinh: Right. We can remove the `strm_` data member and the code would be cleaner. I guess I was just avoiding passing the Serializer object from `get_*_value` or `get_*_values` to the other methods that they use. I'll revisit this.

tests/DCPS/UnitTests/DynamicData/run_test.pl
```diff
@@ -0,0 +1,21 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
```
ClaytonCalabrese, line 1: CI detected memory leaks in the test

ClaytonCalabrese (changes requested):

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+  SingleValueStruct expected;
+  expected.my_enum = E_UINT8; expected.int_32 = 10; expected.uint_32 = 11;
+  expected.int_8 = 5; expected.uint_8 = 6; expected.int_16 = 0x1111;
+  expected.uint_16 = 0x2222; expected.int_64 = 0x7fffffffffffffff;
+  expected.uint_64 = 0xffffffffffffffff; expected.float_32 = 1.0f;
+  expected.float_64 = 1.0; set_float128_value(expected.float_128);
+  expected.char_8 = 'a'; expected.char_16 = 0x0061; expected.byte = 0xff;
+  expected._cxx_bool = true; expected.nested_struct.l = 12;
+  expected.str = "abc"; expected.wstr = L"abc";
```
iguessthislldo, line 79: These should be on separate lines.

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
```diff
@@ -0,0 +1,905 @@
...
+
+  ACE_Message_Block msg(1024);
+  msg.copy((const char*)single_value_struct, sizeof(single_value_struct));
+  XTypes::DynamicData data(&msg, xcdr2, dt);
+
+  // For reading value indirectly through a nested DynamicData object.
+  XTypes::MemberId random_id = 111;
```
iguessthislldo, line 86: Is there something special about 111?

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
```diff
@@ -0,0 +1,905 @@
...
+  CORBA::wstring_free(wstr);
+
+  wstr = 0;
+  ret = data.get_complex_value(nested_dd, 18);
+  EXPECT_EQ(ret, DDS::RETCODE_OK);
+  ret = nested_dd.get_wstring_value(wstr, random_id);
+  EXPECT_EQ(ret, DDS::RETCODE_OK);
+  EXPECT_STREQ(expected.wstr.in(), wstr);
+  CORBA::wstring_free(wstr);
+}
```
iguessthislldo, line 302: Is there a test where members are read out of order and/or skipping members?

iguessthislldo (changes requested):

dds/DCPS/Serializer.h
```diff
@@ -762,6 +767,14 @@ class OpenDDS_Dcps_Export Serializer {
 
   bool peek(ACE_CDR::ULong& t);
 
+  struct RdState {
+    unsigned char align_rshift;
+    size_t rpos;
+  };
+
+  RdState rdstate() const;
+  void rdstate(const RdState& state);
```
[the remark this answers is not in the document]
sonndinh: It could be but that would provide more information than what `DynamicData` needs. There are other data in `Serializer` like the pointer to its message block, write state whereas `DynamicData` only needs the two member variables for the read state.

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
sonndinh: No, it's just an arbitrary ID that is used to test `DynamicData` that has data for non-constructed types.

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
sonndinh: That doesn't really matter because each interface `get_*_value` or `get_*_values` has a clean, new `Serializer` object, and starts from the beginning of the data stream and skips all members in front of the requested member.

dds/DCPS/Serializer.h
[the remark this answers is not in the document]
iguessthislldo: Maybe, but what about when we want to add writing support to `DynamicData`? That would require more of the state that already provided by `Serializer`. What if someone wants to add a member to `Serializer`, but doesn't know to add it to `RdState` as well?

tests/DCPS/UnitTests/DynamicData/DynamicData.cpp
iguessthislldo: I'm not saying it's not going to work now, I'm saying it might break one day. (I might break it for all I know) It wouldn't have to be much. Something like repeat one of the previous reads at the end and add a comment that it's a test of an out of order read.

sonndinh: Okay, I can add that. I was ignoring the tests that I'm pretty sure will work given the current implementation. But you're right that a different implementation might be broken by an out of order test.

dds/DCPS/Serializer.h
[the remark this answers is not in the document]
sonndinh: The current design is specific to the original goal of `DynamicData`, that is to support reading only. You have a good point that `Serializer` might add more members to its internal reading state. Should a comment in `RdState` struct saying that it must contain all reading-related variables of `Serializer` be okay in this case?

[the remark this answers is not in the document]
iguessthislldo: Yes, I'm fine with that. For the sake of finishing this PR we can put off any major changes for now.

iguessthislldo (approved):

mitza-oci (approved):

ClaytonCalabrese (approved):

PR #2992, opened 2021-09-12
Fix LargeSample RTPS Test To Actually Use RTPS

simpsont-oci:


iguessthislldo:
Looks like I was the one to change the domain id to 113, although I don't know the reason I did it. However unless we're doing something special that really requires a `domain` section, I think we should use [`DCPSDefaultDiscovery` like in `rtps_disc_sec.ini`](https://github.com/objectcomputing/OpenDDS/blob/6bba41be97742386a0c8d36282addb6906794f07/tests/DCPS/LargeSample/rtps_disc_sec.ini#L2) so that it's doesn't require that the correct domain id is in the ini file

mitza-oci (approved):

PR #2996, opened 2021-09-13
RtpsRelay pending queue delays discovery

jrw972:
Problem
-------
An RtpsRelay client is moved into pending status when the RtpsRelay
receives any message and conditions allow the client to be admitted.
However, discovery will not start until an SPDP message is received
and a message on the SEDP port is received.  This could delay
discovery by the SPDP resend period.

Solution
--------
Record every new client but only admit them when SPDP is received and
an SEDP address is available.


jrw972:
Merge to green-dev.

mitza-oci (approved):

PR #2997, opened 2021-09-13
RtpsUdpDataLink doesn't gap for new reliable reader

jrw972:
Problem
-------

The following sequence was observed:

1. A reliable reader dies unexpectedly with unacknowledged data.

2. This freezes the heartbeat first number.

3. A subsequent reliable reader was offered the heartbeat first and
   had to nack/gap the entire sequence.

This is effectively the same problem as #2954 which does not address
the case where the reader is initialized with a non-directed
heartbeat.

Solution
--------

Send a "catch up" gap to readers that are nacking sequence numbers
below their starting sequence number.


jrw972:
Merge to green-dev.

mitza-oci (approved):

PR #2998, opened 2021-09-13
RtpsWriter send buffer leaks with persistent pre-association reader

jrw972:
Problem
-------

The send buffer in an RtpsWriter is never cleaned so long as there is
a pre-association reader.

Solution
--------

Track the sequence numbers promised to preassociation readers so they
are retained when association completes but allow the cleanup to occur.


jrw972:
Merge to green-dev.

mitza-oci (approved):

PR #3000, opened 2021-09-14
check spdp address with locators

lij-oci:


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -980,7 +980,7 @@ Spdp::data_received(const DataSubmessage& data,
     return;
   }
 
-  if (!ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList)) {
+  if (!ip_in_locator_list(DCPS::LogAddr::ip(from), pdata.participantProxy.metatrafficUnicastLocatorList)) {
     ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Spdp::data_received - IP not in locator list: %C\n"), DCPS::LogAddr(from).c_str()));
```
iguessthislldo, line 984: This needs to be conditional on debug level.

iguessthislldo (changes requested):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,6 +980,13 @@ Spdp::data_received(const DataSubmessage& data,
     return;
   }
 
+  if (!ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList)) {
```
jrw972, line 983: You need to handle the case that from is the IP of the relay.

jrw972: Similarly, we need to handle ICE (now that I've though about it).

jrw972 (changes requested):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,16 +1014,30 @@ Spdp::data_received(const DataSubmessage& data,
...
+  if (!from_relay && !ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList)) {
+    if (DCPS::DCPS_debug_level) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Spdp::data_received - IP not in locator list: %C\n"), DCPS::LogAddr(from).c_str()));
+    }
+    return;
+  }
 
 #ifdef OPENDDS_SECURITY
   if (!is_security_enabled()) {
+    if (!ip_in_AgentInfo(from, plist)) {
```
jrw972, line 1027: This condition needs to be combined with the previous condition.
Thus, you should have one check if security is enabled that includes the relay, locator list, and agent info and another check for no security that only includes the relay and locator list.

jrw972 (changes requested):

dds/DCPS/RTPS/Spdp.cpp
lij-oci: The current code seems to do just that. Let's see where it's not what we want.
Three checks: 1. from_relay, 2. in_locator_list, 3. in_AgentInfo (if OPENDDS_SECURITY)

Steps:

1. check from_relay
1.1 Y, check in_AgentInfo (if OPENDDS_SECURITY)
1.2 N, check in_locator_list

2. check in_locator_list
2.1 Y, check in_AgentInfo (if OPENDDS_SECURITY)
2.2 N, return

3. check in_AgentInfo (if OPENDDS_SECURITY)
3.1 Y, process_participant_ice
3.2 N, return

4. if not yet returned, handle_participant_data 

Do you want to check in_AgentInfo first?

jrw972 (approved):

mitza-oci:
Seems like just about every test is failing in the safety profile builds.  I tried debugging one.  It fails the new SPDP locators check because the `from` address was an actual IP address assigned to the host but the metatraffic locator list only had `127.0.0.1`.

lij-oci:
Does this mean that there is a problem in the checking logic, or that we should add both the local ip and 127.0.0.1 in the test unicast locator list?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -613,6 +614,54 @@ Spdp::get_ice_endpoint_if_added()
   return tport_->ice_endpoint_added_ ? tport_->get_ice_endpoint() : 0;
 }
 
+bool ip_in_locator_list(const ACE_INET_Addr& from, const DCPS::LocatorSeq& locators)
+{
+  bool locator_found = false;
```
mitza-oci, line 619:
```suggestion
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -613,6 +614,54 @@ Spdp::get_ice_endpoint_if_added()
...
+  for (CORBA::ULong i = 0; i < locators.length(); ++i) {
+    ACE_INET_Addr addr;
+    if (locator_to_address(addr, locators[i], true) == 0) {
+      if (from.is_ip_equal(addr)) {
+        locator_found = true;
+        break;
```
mitza-oci, line 625:
```suggestion
        return true;
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -613,6 +614,54 @@ Spdp::get_ice_endpoint_if_added()
...
+      }
+    }
+  }
+  return locator_found;
```
mitza-oci, line 629:
```suggestion
  return false;
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,16 +1014,30 @@ Spdp::data_received(const DataSubmessage& data,
...
+  if (!from_relay && !ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList) && !ip_in_AgentInfo(from, plist)) {
+    if (DCPS::DCPS_debug_level) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Spdp::data_received - dropped IP: %C\n"), DCPS::LogAddr(from).c_str()));
+    }
+    return;
+  }
   if (!is_security_enabled()) {
     process_participant_ice(plist, pdata, guid);
   }
+#else
```
mitza-oci, line 1028:
```suggestion
#elif !defined OPENDDS_SAFETY_PROFILE
```

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -28,6 +28,14 @@
...
 #include <exception>
 #include <iostream>
+#include <string>
+
+#if defined _WIN64 || defined _WIN32
+#include <winsock.h>
+#else
+#include <arpa/inet.h>
+#include <unistd.h>
+#endif
```
mitza-oci, line 38: This block should be removed.  Determine which platform functions/macros you need and call the equivalent from ACE - is it just `ntohl`?

mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
lij-oci: Should we also put #ifndef OPENDDS_SAFETY_PROFILE in the #ifdef OPENDDS_SECURITY branch?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,16 +1012,32 @@ Spdp::data_received(const DataSubmessage& data,
...
   }
 
-  handle_participant_data(
-    (data.inlineQos.length() && disposed(data.inlineQos)) ?
-      DCPS::DISPOSE_INSTANCE : DCPS::SAMPLE_DATA,
-    pdata, to_opendds_seqnum(data.writerSN), from, false);
-
+  const bool from_relay = from == config_->spdp_rtps_relay_address();
 #ifdef OPENDDS_SECURITY
+#ifndef OPENDDS_SAFETY_PROFILE in the #ifdef OPENDDS_SECURITY
```
mitza-oci, line 1017: I don't know what happened in the last commit but this line isn't right.  Please change it back to what it was before.

mitza-oci (changes requested):

mitza-oci:
tests/transport/spdp is failing on multiple builds in GitHub Actions

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -613,6 +614,65 @@ Spdp::get_ice_endpoint_if_added()
   return tport_->ice_endpoint_added_ ? tport_->get_ice_endpoint() : 0;
 }
 
+bool is_ip_equal(const ACE_INET_Addr& a, const DCPS::Locator_t& locator)
+{
+  const ACE_UINT32 a_ip = ntohl(a.get_ip_address());
```
simpsont-oci, line 619: This approach won't work for when `a` is a real IPv6 address, since IPv6 addresses are 128 bits (not 32).

What you will probably want to do instead is use ACE_INET_Addr::get_addr() and cast the result (void *) to a  sockaddr* check for family / address pointer that way. You will specifically need to handle the case when one address is IPv4 and the other is IPv4-mapped-IPv6 (different kinds) but the last four bytes of the IPv6 address match the IPv4 address.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,16 +1025,34 @@ Spdp::data_received(const DataSubmessage& data,
...
+#elif !defined OPENDDS_SAFETY_PROFILE
+  if (!from_relay && !ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList)) {
+    if (DCPS::DCPS_debug_level) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Spdp::data_received - IP not in locator list: %C\n"), DCPS::LogAddr(from).c_str()));
+    }
+    return;
+  }
 #endif
+
+  handle_participant_data(
```
simpsont-oci, line 1052: Did you mean to move the call to handle_participant_data() to after process_participant_ice()? That change in order doesn't seem related to checking the spdp received addresses against locators.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -251,12 +254,29 @@ struct ReactorTask : ACE_Task_Base {
   }
 };
 
+void get_local_ip(OpenDDS::DCPS::OctetArray16 address)
+{
+  const String hostname = get_fully_qualified_hostname();
+  const ACE_INET_Addr addr = choose_single_coherent_address(hostname + ":54321", false);
+  std::memset(address, 0, 12);
+  *(ACE_UINT32*)(&address[12]) = ntohl(addr.get_ip_address());
```
simpsont-oci, line 262: The address returned can be IPv6, so you can't rely on get_ip_address(). See other comment.

tests/transport/spdp/spdp_transport.cpp
```diff
@@ -330,47 +348,42 @@ bool run_test()
...
 #endif
     },
-    { // Duration_t (leaseDuration)
-      static_cast<CORBA::Long>((rd.resend_period() * 10).value().sec()),
-      0 // we are not supporting fractional seconds in the lease duration
-    },
-    { 0, 0},
-    0
+    // Duration_t (leaseDuration - we are not supporting fractional seconds)
+    {static_cast<CORBA::Long>((rd.resend_period() * 10).value().sec()), 0},
```
simpsont-oci, line 385: I don't think this change improves anything. The comment about fractional seconds is now in a less helpful position.

simpsont-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
lij-oci: ACE_INET_Addr::get_ip_address() handles the type issue and byte order, and returns a 4-byte ip address:
 
// Return the 4-byte IP address, converting it into host byte order.
ACE_UINT32
ACE_INET_Addr::get_ip_address (void) const
{
  ACE_TRACE ("ACE_INET_Addr::get_ip_address");
#if defined (ACE_HAS_IPV6)
  if (this->get_type () == AF_INET6)
    {
      if (IN6_IS_ADDR_V4MAPPED (&this->inet_addr_.in6_.sin6_addr) ||
          IN6_IS_ADDR_V4COMPAT (&this->inet_addr_.in6_.sin6_addr)    )
        {
          ACE_UINT32 addr;
          // Return the last 32 bits of the address
          char *thisaddrptr = (char*)this->ip_addr_pointer ();
          thisaddrptr += 128/8 - 32/8;
          ACE_OS::memcpy (&addr, thisaddrptr, sizeof (addr));
          return ACE_NTOHL (addr);
        }

      ACELIB_ERROR ((LM_ERROR,
                  ACE_TEXT ("ACE_INET_Addr::get_ip_address: address is a IPv6 address not IPv4\n")));
      errno = EAFNOSUPPORT;
      return 0;
    }
#endif /* ACE_HAS_IPV6 */
  return ACE_NTOHL (ACE_UINT32 (this->inet_addr_.in4_.sin_addr.s_addr));
}

dds/DCPS/RTPS/Spdp.cpp
lij-oci: Justin added this:

Acceptance Criteria

Implement the metatraffic locator check described in the parent issue (378 - see “possible solutions #1”) with appropriate logging.

Here is the sketch for ICE.

We may be tempted to move this block 

#ifdef OPENDDS_SECURITY
  if (!is_security_enabled()) {
    process_participant_ice(plist, pdata, guid);
  }
#endif

before the call to handle_participant_data.  However, we can (and should) assume that we get an SPDP message through the relay or some other legitimate means first.

2.  The definition of process_participant_ice refers to members named sedp_info_ and spdp_info_.  These members are an ICE::AgentInfo (dds/DCPS/Ice.h).  This contains a list of candidates and each candidate is an address.  Extend the  ip_in_locator_list to check these candidates.

dds/DCPS/RTPS/Spdp.cpp
lij-oci: removed call to get_ip_address

tests/transport/spdp/spdp_transport.cpp
lij-oci: removed call to get_ip_address

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -965,16 +1041,30 @@ Spdp::data_received(const DataSubmessage& data,
...
-  handle_participant_data(
-    (data.inlineQos.length() && disposed(data.inlineQos)) ?
-      DCPS::DISPOSE_INSTANCE : DCPS::SAMPLE_DATA,
-    pdata, to_opendds_seqnum(data.writerSN), from, false);
-
+  const bool from_relay = from == config_->spdp_rtps_relay_address();
 #ifdef OPENDDS_SECURITY
+  if (!from_relay && !ip_in_locator_list(from, pdata.participantProxy.metatrafficUnicastLocatorList) && !ip_in_AgentInfo(from, plist)) {
+    if (DCPS::DCPS_debug_level) {
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Spdp::data_received - dropped IP: %C\n"), DCPS::LogAddr(from).c_str()));
```
simpsont-oci, line 1048: @jrw972 Is this worth a warning, or is debug OK?

jrw972: I would convert it to a warning.

simpsont-oci (approved):

Issue #3003, opened 2021-09-14
_ace_bin_hints/_tao_bin_hints Mix-up in OpenDDSConfig.cmake


cpb-dynetics:
In cmake/OpenDDSConfig.cmake, the `*_bin_hints` variables for ACE and TAO are mixed-up, i.e. the ACE directory is given to `_tao_bin_hints` and the TAO directory is given to `_ace_bin_hints`.

Normally in a local install this might not be a problem as everything is thrown into the ACE_wrappers directory. However when `OPENDDS_USE_PREFIX_PATH` is enabled (after installing OpenDDS using `./configure --prefix=/some/path`) it errors as it cannot find the path to `tao_idl`.

The bug is located on lines 80-81 of OpenDDSConfig.cmake (as of latest commit e0f9aa0 on Apr 27):
```
set(_tao_bin_hints ${ACE_BIN_DIR})
set(_ace_bin_hints ${TAO_BIN_DIR})
```
should be
```
set(_ace_bin_hints ${ACE_BIN_DIR})
set(_tao_bin_hints ${TAO_BIN_DIR})
```

jwillemsen:
Please open a PR with the proposed fix (one PR for each issue), that is the easiest to review and merge

cpb-dynetics:
Sure thing, will do so shortly.

mitza-oci:
Is the use case different from what's tested in GitHub Actions?  See job `build_u18_o1d0v1_xer0_j12` step `make install build cmake`

cpb-dynetics:
Apologies for messing up the issue submission. I will be closing this one in favor of #3006 with the active PR.

PR #3006, opened 2021-09-14
_ace_bin_hints/_tao_bin_hints Mix-up in OpenDDSConfig.cmake

cpb-dynetics:
Resolves [issue #3003](https://github.com/objectcomputing/OpenDDS/issues/3003)

### DIFF
--- a/cmake/OpenDDSConfig.cmake
+++ b/cmake/OpenDDSConfig.cmake
@@ -77,8 +77,8 @@ endif()
 include(${CMAKE_CURRENT_LIST_DIR}/init.cmake)
 
 set(_dds_bin_hints ${OPENDDS_BIN_DIR})
-set(_tao_bin_hints ${ACE_BIN_DIR})
-set(_ace_bin_hints ${TAO_BIN_DIR})
+set(_ace_bin_hints ${ACE_BIN_DIR})
+set(_tao_bin_hints ${TAO_BIN_DIR})
 
 find_program(PERL perl)
 


cpb-dynetics:
@mitza-oci to answer your question: While my use case is somewhat different the method seems to be similar, if not the same. Could it be possible that the TAO_IDL is already found by the point the test arrives at the `make install build cmake` step?

I rebuilt my CMake-based project from scratch (including clearing all OpenDDS-related environment variables) and the error that I get is `Missing required dependencies TAO_IDL`. I am leveraging your guys' CMake module (`/lib/cmake/OpenDDS/OpenDDSConfig.cmake`) to fill in the blanks when rebuilding, and it builds successfully after I apply this fix.

Sorry, I'm pushing end of my work day right now so I will continue looking into your testing tomorrow to be sure.

iguessthislldo (approved):

ClaytonCalabrese (approved):
This seems to have been broken for years, good catch.

mitza-oci:
> 
> 
> This seems to have been broken for years, good catch.

Yes, it's a typo.  But what impact would it have?

TAO doesn't have its own "bin" directory so I'm not sure what this really means.  When would `${ACE_BIN_DIR}` and `${TAO_BIN_DIR}` differ?

cpb-dynetics:
> When would `${ACE_BIN_DIR}` and `${TAO_BIN_DIR}` differ?

Because of the `tao_idl` symlink in `<prefix>/share/ace/bin`, sure it might not make a difference. However in `<prefix>/lib/cmake/OpenDDS/init.cmake`, if `OPENDDS_USE_PREFIX_PATH` is ON/true, `ACE_BIN_DIR` and `TAO_BIN_DIR` are technically set to different directories (even though in a normal situation it should resolve to the same place).

So if they're meant to be separate variables, then fixing this typo is necessary. Though if they're not meant to differ, should there be two variables?

Issue #3004, opened 2021-09-14
Wrong ACE_BIN_DIR in init.cmake when using OPENDDS_USE_PREFIX_PATH


cpb-dynetics:
In cmake/init.cmake, the `ACE_BIN_DIR` variable is incorrectly set if `OPENDDS_USE_PREFIX_PATH` is on  (after installing OpenDDS using `./configure --prefix=/some/path`). This causes `ACE_BIN_DIR` to be "/some/path/share/ace/bin" instead of "/some/path/bin", so CMake errors as it cannot find the ACE_GPERF dependency.

The bug is located in the ACE_ROOT initialization block on lines 80-96 of init.cmake (as of latest commit d481345 on May 4):
```
if (NOT DEFINED ACE_ROOT)
  if(OPENDDS_USE_PREFIX_PATH)
    set(ACE_ROOT "${_OPENDDS_RELATIVE_PREFIX_ROOT}/share/ace")
    set(ACE_INCLUDE_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/include")
    set(ACE_LIB_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/lib")

  elseif(OPENDDS_ACE)
    set(ACE_ROOT ${OPENDDS_ACE})
    set(ACE_INCLUDE_DIR "${ACE_ROOT}")
    set(ACE_LIB_DIR "${ACE_ROOT}/lib")

  else()
    _OPENDDS_RETURN_ERR("Failed to locate ACE_ROOT")
  endif()

  set(ACE_BIN_DIR "${ACE_ROOT}/bin")                         # <-- HERE
endif()
```
should be
```
if (NOT DEFINED ACE_ROOT)
  if(OPENDDS_USE_PREFIX_PATH)
    set(ACE_ROOT "${_OPENDDS_RELATIVE_PREFIX_ROOT}/share/ace")
    set(ACE_INCLUDE_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/include")
    set(ACE_BIN_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/bin")  # <-- HERE
    set(ACE_LIB_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/lib")

  elseif(OPENDDS_ACE)
    set(ACE_ROOT ${OPENDDS_ACE})
    set(ACE_INCLUDE_DIR "${ACE_ROOT}")
[...]

PR #3007, opened 2021-09-14
Wrong ACE_BIN_DIR in init.cmake when using OPENDDS_USE_PREFIX_PATH

cpb-dynetics:
Resolves [issue #3004](https://github.com/objectcomputing/OpenDDS/issues/3004)

### DIFF
--- a/cmake/init.cmake
+++ b/cmake/init.cmake
@@ -81,18 +81,18 @@ if (NOT DEFINED ACE_ROOT)
   if(OPENDDS_USE_PREFIX_PATH)
     set(ACE_ROOT "${_OPENDDS_RELATIVE_PREFIX_ROOT}/share/ace")
     set(ACE_INCLUDE_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/include")
+    set(ACE_BIN_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/bin")
     set(ACE_LIB_DIR "${_OPENDDS_RELATIVE_PREFIX_ROOT}/lib")
 
   elseif(OPENDDS_ACE)
     set(ACE_ROOT ${OPENDDS_ACE})
     set(ACE_INCLUDE_DIR "${ACE_ROOT}")
+    set(ACE_BIN_DIR "${ACE_ROOT}/bin")
     set(ACE_LIB_DIR "${ACE_ROOT}/lib")
 
   else()
     _OPENDDS_RETURN_ERR("Failed to locate ACE_ROOT")
   endif()
-
-  set(ACE_BIN_DIR "${ACE_ROOT}/bin")
 endif()
 
 if (NOT DEFINED TAO_ROOT)


mitza-oci:
I'm not sure the way it is now is incorrect.  What's the error message you're seeing?

For some context, please see the `### Application Development with an Installed OpenDDS` section of `INSTALL.md`.

iguessthislldo:
This won't work as is because we're using `ACE_BIN_DIR` to find `generate_export_file.pl`, which is only in `PREFIX/share/bin`:

https://github.com/objectcomputing/OpenDDS/blob/1fd77cf02fd6f3ed9f1fb3c873455619816d4e75/cmake/api_macros.cmake#L79

Besides that you mentioned in the issue that it couldn't find `ace_gperf`. There is supposed to be a symlink at `PREFIX/share/bin/ace_gpref` that links to `PREFIX/bin/ace_gperf`. Does that symlink exist in your install prefix?

cpb-dynetics:
I see the bit about the three shell scripts now (`dds-devel.sh`, `ace-devel.sh`, and `tao-devel.sh`). Must've skimmed right over it, my bad.

After going through all of your guys' CMake files -- especially in this context: init.cmake -- I was under the impression that the OpenDDS install (from `make install`) should enter the conditionals where `OPENDDS_USE_PREFIX_PATH == true`, and subsequently use `_OPENDDS_RELATIVE_PREFIX_ROOT` when determining all of the `*_ROOT` env vars. Sourcing those three scripts solves this issue, but I still believe that this variable needs to be fixed in init.cmake.

To give some context: my use case is that I'm trying to facilitate the use of OpenDDS in my project by letting CMake figure out the absolute path regardless of whose computer it's running on. For example, if I ask a teammate to pull my repo and attempt to use one of my tools, they should be able to do so without issue through a standard CMake project (i.e. `find_package(OpenDDS)`). I would also rather not add any unnecessary command-line arguments or settings to my project if I see that a fix is possible elsewhere, i.e. at another source.

By relying on config.cmake setting `OPENDDS_USE_PREFIX_PATH` to ON, and adding my project's dependencies directory to `CMAKE_PREFIX_PATH`, I can leverage the relevant bit in init.cmake to set those paths for me based on where the teammate put my project.

iguessthislldo:
> For some context, please see the ### Application Development with an Installed OpenDDS section of INSTALL.md.

> I see the bit about the three shell scripts now (dds-devel.sh, ace-devel.sh, and tao-devel.sh). Must've skimmed right over it, my bad.

Those are files are more important for MPC-built applications than CMake-built ones because of `OPENDDS_USE_PREFIX_PATH`.

>  I was under the impression that the OpenDDS install (from make install) should enter the conditionals where OPENDDS_USE_PREFIX_PATH == true, and subsequently use _OPENDDS_RELATIVE_PREFIX_ROOT when determining all of the *_ROOT env vars.

That is correct. Theoretically you shouldn't have to set any environment variables to use the OpenDDS CMake module.

> To give some context: my use case is that I'm trying to facilitate the use of OpenDDS in my project by letting CMake figure out the absolute path regardless of whose computer it's running on.

I don't see a problem with this use case from just from what you've written so far, but maybe we're missing something that's actually causing the issue, like the location of `ace_gperf` I mentioned above.

mitza-oci:
Let my clarify my comment from yesterday.  As a user of the CMake module, you should follow `Example Using Installed OpenDDS` in `docs/cmake.md`.  If you want to change the CMake module, or understand why it does things internally the way it does, the info in `INSTALL.md` gives you some background on how the "Installed OpenDDS" works regardless of which build system you use.

cpb-dynetics:
> Besides that you mentioned in the issue that it couldn't find ace_gperf. There is supposed to be a symlink at PREFIX/share/ace/bin/ace_gpref that links to PREFIX/bin/ace_gperf. Does that symlink exist in your install prefix?
It does exist. Just tried rebuilding my project from scratch without my "fix" and... it worked. I must've been messing around yesterday with some CMake on my end that broke it somehow, or maybe just didn't copy the install directory in its entirety.
Didn't mean to waste your guys' time here! Thought I was onto something.

PR #3012, opened 2021-09-16
Liveliness Test: Remove Counting

ClaytonCalabrese:
When the LivelinessTest was written 14 years ago, it used a count to keep track of liveliness. 

This lead to a problem with a remote reader where depending on whether the Transport, or the DCPSInfoRepo cleaned up first, we would get a different liveliness count.  In reality, it doesn't matter which order these cleanups happen in, so long as they both happen.

We already ensure this by checking the `verify_last_liveliness_status()` of each reader to make sure everything is clean upon the exiting of the test. As such, I removed the liveliness count because it was reporting problems that were not actually problems, indicating it is a bad metric for the test to be using when determining success.

### DIFF
--- a/tests/DCPS/LivelinessTest/DataReaderListener.cpp
+++ b/tests/DCPS/LivelinessTest/DataReaderListener.cpp
@@ -98,8 +98,6 @@ void DataReaderListenerImpl::on_liveliness_changed(
       dcs_->post(actor_, "LIVELINESS_LOST_" + OpenDDS::DCPS::to_dds_string(liveliness_gained_count_));
     }
 
-    ++liveliness_changed_count_;
-
     if (liveliness_changed_count_ > 1) {
       if (last_status_.alive_count == 0 && last_status_.not_alive_count == 0) {
         ACE_ERROR ((LM_ERROR,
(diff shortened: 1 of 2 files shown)


tests/DCPS/LivelinessTest/DataReaderListener.cpp
```diff
@@ -135,8 +133,8 @@ void DataReaderListenerImpl::on_liveliness_changed(
     last_status_ = status;
     ACE_DEBUG((LM_DEBUG,
       "(%P|%t) %T DataReaderListenerImpl::on_liveliness_changed %d\n"
-      "alive_count %d not_alive_count %d alive_count_change %d not_alive_count_change %d\n",
-      liveliness_changed_count_, status.alive_count, status.not_alive_count,
+      "%C alive_count %d not_alive_count %d alive_count_change %d not_alive_count_change %d\n",
+      liveliness_changed_count_, actor_.c_str(), status.alive_count, status.not_alive_count,
```
simpsont-oci, line 137: If we're not incrementing `liveliness_changed_count_` anymore, we don't need to print the value here anymore.

mitza-oci (approved):

PR #3028, opened 2021-09-24
Extend RawDataSample with the header so that we can deserialize a XCDR2 stream

### LINKED ISSUE #3025: RawDataSample can't be deserialized in user code
When using the Recorder we try to deserialize the received RawDataSample (see for example https://github.com/objectcomputing/OpenDDS/blob/2db4ae425bb6e5a55724755b9cdad02ce62b7480/tests/DCPS/RecorderReplayer/Relay.cpp#L85) to log it in a custom recorder application. This works when using for example shared memory transport, but not when using rtps_udp. It looks not enough information is available to demarshal the RawDataSample using a Serializer in user code which mimics https://github.com/objectcomputing/OpenDDS/blob/2db4ae425bb6e5a55724755b9cdad02ce62b7480/dds/DCPS/DataReaderImpl_T.h#L1087

jwillemsen:
At the moment `rtps_udp` is used as transport the data is encoded using XCDR2 which we can only deserialize when we have some more information available as part of the `RawDataSample`. Instead of adding even more fields we now pass the `DataSampleHeader` with all available information. 

Added a new unit test do demonstrate how to deserialize the `RawDataSample` in user code, in the near future we can hopefully use DynamicData from DDSXTypes for this.

Fixed #3025.


iguessthislldo (changes requested):

mitza-oci:
In the current design (in master), RawDataSample provides isolation from the internal DataSampleHeader.  It seems like the only field needed for the test which is not already in RawDataSample is cdr_encapsulation_.

jwillemsen:
> In the current design (in master), RawDataSample provides isolation from the internal DataSampleHeader. It seems like the only field needed for the test which is not already in RawDataSample is cdr_encapsulation_.

I am not 100% certain that we have enough fields for all user code, instead of adding field by field I decided to expose the DataSampleHeader completely giving the Recorder access to all possible data.

jwillemsen:
And when DDSTypes Dynamic is ready on master I would assume we migrate to that, but this way we only have 1 change to get any user code working before that is ready.

mitza-oci (approved):

PR #3040, opened 2021-09-28
CDR Encapsulation option for non-rtps_udp transports

ClaytonCalabrese:
This PR adds the capability to put cdr_encapsulation for transports besides rtps_udp.  Unlike rtps_udp, this will be disabled by default to maintain current functionality for users.


dds/DCPS/transport/framework/TransportClient.cpp
```diff
@@ -159,6 +159,7 @@ TransportClient::enable_transport_using_config(bool reliable, bool durable,
 #endif
 
         cdr_encapsulation_ |= inst->requires_cdr_encapsulation();
+        ACE_DEBUG((LM_DEBUG, "CDR_ENCAPSULATION: %d\n", cdr_encapsulation_));
```
jwillemsen, line 162: Check debug level?

tests/DCPS/XTypes/run_test.pl
```diff
@@ -165,7 +165,25 @@
     "NoXTypesNoMatchTypeNames" => {reader_type => "AppendableStructNoXTypes", writer_type => "AppendableStructNoXTypes",
                                    r_reg_type => "ReaderTypeName", w_reg_type => "WriterTypeName",
                                    expect_to_fail => 1, topic => "NoXTypesNoMatchTypeNames_nn", key_val => 1,
-                                   r_ini => "rtps_disc_no_xtypes.ini", w_ini => "rtps_disc_no_xtypes.ini"}
+                                   r_ini => "rtps_disc_no_xtypes.ini", w_ini => "rtps_disc_no_xtypes.ini"},
+    "Tcp_AppendableMatch"         => {reader_type => "AppendableStruct", writer_type => "AdditionalPostfixFieldStruct",
+                                  expect_to_fail => 0, topic => "AppendableStructT", key_val => 4,
+                                  r_ini => "tcp.ini", w_ini => "tcp.ini"},
```
iguessthislldo, line 171: I should have said something about this before when the test was first added, but can you make your new entries something like:

```Perl
    "Tcp_AppendableMatch" => {
      reader_type => "AppendableStruct", writer_type => "AdditionalPostfixFieldStruct",
      expect_to_fail => 0, topic => "AppendableStructT", key_val => 4,
      r_ini => "tcp.ini", w_ini => "tcp.ini",
    },
```

That way they're easier to copy, expand, and rename without worrying about indent. You don't have to apply this to the old hash entries if you don't want to.

tests/DCPS/XTypes/run_test.pl
```diff
@@ -165,7 +165,25 @@
...
+    "Tcp_MutableStructNoMatchName"=> {reader_type => "MutableStruct", writer_type => "ModifiedNameMutableStruct",
+                                  expect_to_fail => 1, topic => "MutableStructT_NoMatchName", key_val => 9,
+                                  r_ini => "tcp.ini", w_ini => "tcp.ini"},
+    "Tcp_MutableUnion"            => {reader_type => "MutableUnion", writer_type => "ModifiedMutableUnion",
+                                  expect_to_fail => 0, topic => "MutableUnionT", key_val => 10,
+                                  r_ini => "tcp.ini", w_ini => "tcp.ini"},
+    "Tcp_MutableUnionNoMatchDisc" => {reader_type => "MutableUnion", writer_type => "ModifiedDiscMutableUnion",
+                                  expect_to_fail => 1, topic => "MutableUnionT_NoMatchDisc", key_val => 11,
+                                  r_ini => "tcp.ini", w_ini => "tcp.ini"}
```
iguessthislldo, line 186: What ever you do, add a comma after the final item because Perl allows that unlike C++03. That way new entries don't have to add a comma like you did.

dds/DCPS/transport/tcp/TcpInst.cpp
```diff
@@ -68,6 +68,9 @@ OpenDDS::DCPS::TcpInst::load(ACE_Configuration_Heap& cf,
   GET_CONFIG_VALUE(cf, trans_sect, ACE_TEXT("active_conn_timeout_period"),
                    this->active_conn_timeout_period_, int)
 
+  GET_CONFIG_VALUE(cf, trans_sect, ACE_TEXT("requires_cdr_encapsulation"),
```
iguessthislldo, line 71: As a function name I think this was fine at the time, but I'm not sure about using it as the name of the config option. This is because none of the transport except RTPS really require encapsulation from our perspective. But I guess the user is requiring to get XCDR, then maybe it is a good name? You don't have to do anything about this, I'm mainly just thinking out loud.

iguessthislldo (changes requested):

mitza-oci:
*Requires* CDR Encapsulation is an inherent property of the transport implementation.  It isn't meant to be configured -- it's a fact about the transport library (`rtps_udp` requires the encapsulation header, `tcp` doesn't) that the framework level needs to know.

`cdr_encapsulation_` itself is an attribute of `TransportClient`.  We can give users options of configuring this, just like `swap_bytes_`.  Following that pattern, the `TransportConfig` is the class that deals with the API or ini-file configuration.  

This will avoid the duplication across different transport types and is a more direct configuration switch for controlling what the DataWriter needs to do, instead of indirectly determining it through a logical-OR across all configured transportimpls.

dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -280,6 +280,17 @@ TransportRegistry::load_transport_configuration(const OPENDDS_STRING& file_name,
                                   value.c_str(), config_id.c_str()),
                                  -1);
               }
+            } else if (name == "cdr_encapsulation") {
+              if ((value == "1") || (value == "true")) {
+                config->cdr_encapsulation_ = true;
+                ACE_DEBUG((LM_DEBUG, "HELLO WORLD\n\n"));
```
iguessthislldo, line 286:
```suggestion
```

iguessthislldo (changes requested):
Other than this debug statement I think it looks good to go assuming it passes CI. I wish it didn't require a separate INI section to do it, but there's a lot I wish was different about how the INI files worked, so I guess this is just another one.

iguessthislldo (approved):

PR #3050, opened 2021-10-01
Fix SEGV in TransportSendStrategy for Empty Packet Chain

simpsont-oci:
This should fix the SEGV for an empty packet chain in TransportSendStrategy.

It remains confusing to me, though, why this was necessary, since header_.length_ and pkt_chain_ seem to be "reset" at the same time when they are modified.

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendStrategy.inl
+++ b/dds/DCPS/transport/framework/TransportSendStrategy.inl
@@ -31,7 +31,6 @@ ThreadSynch* TransportSendStrategy::synch() const
 ACE_INLINE
 void TransportSendStrategy::set_header_source(ACE_INT64 source)
 {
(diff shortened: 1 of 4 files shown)


dds/DCPS/transport/framework/TransportSendStrategy.inl
```diff
@@ -28,6 +28,13 @@ ThreadSynch* TransportSendStrategy::synch() const
   return synch_.get();
 }
 
+ACE_INLINE
+void TransportSendStrategy::set_header_source(ACE_INT64 source)
+{
+  GuardType guard(lock_);
```
simpsont-oci, line 34:
```suggestion
```

mitza-oci (approved):

PR #3054, opened 2021-10-04
Extend RawDataSample with the header so that we can deserialize a XCD…

jwillemsen:
…R2 stream, merged to master as #3028

    * dds/DCPS/RawDataSample.cpp:
    * dds/DCPS/RawDataSample.h:
    * dds/DCPS/RecorderImpl.cpp:
    * tests/DCPS/RecorderReplayer/Relay.cpp:

### DIFF
--- a/dds/DCPS/RawDataSample.cpp
+++ b/dds/DCPS/RawDataSample.cpp
@@ -1,6 +1,4 @@
 /*
- *
- *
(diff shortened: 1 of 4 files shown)


jwillemsen:
OpenSSL on MacOSX is different issue, see #3055

PR #3057, opened 2021-10-04
Added special-case handling for PID_PROPERTY_LIST

mitza-oci:
PID_PROPERTY_LIST is strange because the RTPS and DDS Security
specs each define it differently.  It needs to be deserialized in its
own buffer so that the the DDS::PropertyQosPolicy can detect if it
was using the RTPS definition or the DDS Security definition.


dds/idl/marshal_generator.cpp
```diff
@@ -3661,6 +3661,24 @@ namespace {
         "    uni._d(disc);\n"
         "    return true;\n"
         "  }\n"
+        "  if (disc == RTPS::PID_PROPERTY_LIST) {\n"
+        "    // support special case deserialization of DDS::PropertyQosPolicy\n"
+        "    ACE_Message_Block param(size);\n"
```
simpsont-oci, line 3666: It would be nice if we could avoid the allocation here. Is this a potential use case for the new peek / peek_helper functions on Serializer?

dds/idl/marshal_generator.cpp
```diff
@@ -3661,6 +3661,24 @@ namespace {
         "    uni._d(disc);\n"
         "    return true;\n"
         "  }\n"
+        "  if (disc == RTPS::PID_PROPERTY_LIST) {\n"
```
simpsont-oci, line 3664: Does this belong after the ScopedAlignmentContext, or is it OK here?

PR #3098, opened 2021-10-20
news post for training classes

kroegerk:
@mitza-oci


kroegerk (commented):
@mitza-oci , I think this is ready for publication now.

PR #3121, opened 2021-10-29
RemoveAssociationSweeper is unnecessary

jrw972:
Problem
-------

The RemoveAssociationSweeper class was designed to allow the
"reader (to) process more information that may have queued up from the
writer."  This queueing of data could occur in the transport or the
DataReader.  If the author was referring the DataReader, then the
RemoveAssociationSweeper is superfluous because the DCPS half of the
DataReader is already providing the reader process an opportunity to
process samples.  If the author was referring to the transport, then
the transport is at fault for not promptly deliverying the samples to
the DataReader.

Solution
--------

Remove the RemoveAssociationSweeper.


mitza-oci (approved):
@simpsont-oci if you approve, do we want to merge this before or after #3253 ?

simpsont-oci:
> @simpsont-oci if you approve, do we want to merge this before or after #3253 ?

You can do this one first and I'll handle any merge issues.

simpsont-oci (approved):

PR #3122, opened 2021-10-30
Static/Security with Relay and Attributes Tests

iguessthislldo:
Try to add the correct the links and includes in these tests so they work with static.


.github/workflows/build_and_test.yml
```diff
@@ -4685,6 +4685,124 @@ jobs:
         cat "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"
         grep -q 'Failures: 0' "$GITHUB_WORKSPACE/OpenDDS/${{ github.job }}_autobuild_workspace/latest.txt"
 
+  security_u18_stat_qt_ws_sec:
```
mitza-oci, line 4688: We may want to split this into two jobs so that when a red `x` appears in a pull request the author knows if it's a compile issue (in the test code) or a test failure.  Similar to how the safety profile builds are currently merged but will eventually be split.

mitza-oci (approved):

PR #3125, opened 2021-11-01
XTypes Dynamic: Integration Test

ClaytonCalabrese:
This is the integration test for DnyamicTypes.  For now the files changed will be cluttered until DynamicData is merged it.


dds/DCPS/RecorderImpl.cpp
```diff
@@ -228,6 +228,8 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
 
   // we only support SAMPLE_DATA messages
   if (sample.header_.message_id_ == SAMPLE_DATA && listener_.in()) {
+    // Skip the encapsulation header
+    sample.sample_->rd_ptr(4);
```
iguessthislldo, line 232: Are we passing if the encoding some where else? If not we shouldn't be skipping this.

mitza-oci: Use `sample.header_.cdr_encapsulation_`

dds/DCPS/RecorderImpl.cpp
```diff
@@ -434,6 +438,7 @@ RecorderImpl::add_association(const RepoId&            yourId,
   }
   XTypes::TypeObject cto = tls->get_type_object(type_info.complete.typeid_with_size.type_id);
   XTypes::DynamicType_rch dt = tls->complete_to_dynamic(cto.complete, writer.writerId);
+  ACE_DEBUG((LM_DEBUG, "DynamicType added: %C\n", LogGuid(writer.writerId).c_str()));
```
ClaytonCalabrese, line 441: Put behind debug level

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -1003,6 +1027,29 @@ DDS::ReturnCode_t DynamicData::get_wstring_value(ACE_CDR::WChar*& value, MemberI
   return get_single_value<TK_STRING16>(value, id);
 }
 
+DDS::ReturnCode_t DynamicData::get_discriminator_value(DynamicData& value) {
```
sonndinh, line 1030: I don't think we need a separate method for reading discriminator as it can be read using the existing interfaces. Something like the one Adam suggested in chat would work, but I think it can be simpler if we use a sentinel ID value, like MEMBER_ID_INVALID, to indicate the discriminator. For example, if the discriminator type is `unsigned short`, its value can be read using `union_data.get_uint16_value(disc_value, MEMBER_ID_INVALID)` where `union_data` is a `DynamicData` object of the union. This would require a bit of change to `get_value_from_union` in `DynamicData` but not a new method.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -267,6 +284,13 @@ ACE_CDR::ULong DynamicData::get_item_count()
     return type_->get_member_count();
   case TK_UNION:
     {
+      const ExtensibilityKind ek = descriptor_.extensibility_kind;
+      if (ek == APPENDABLE || ek == MUTABLE) {
+        size_t size;
+        if (!strm_.read_delimiter(size)) {
+          return 0;
+        }
+      }
```
sonndinh, line 293: This is correct. Thanks for adding this.

dds/DCPS/XTypes/DynamicData.cpp
mitza-oci: How about using a value that's above MEMBER_ID_MAX but is not MEMBER_ID_INVALID?
As in the code I posted before, this shouldn't be a detail the user needs to know.  It would be returned by `get_member_id_by_name`.

sonndinh: I think this is one of the uses of MEMBER_ID_INVALID as stated in the spec, where the ID is missing or irrelevant. Even though the spec doesn't say explicitly, if we let users call `get_*_value` with MEMBER_ID_INVALID, `DynamicData` can decide that the user wants to read the dicsriminator without needing to delegate to `DynamicType` and related classes. Else if we choose to delegate to `DynamicType`-related classes, they would need to change too because right now discriminator is not a member in the member list in `DynamicType` for union.

mitza-oci: I'm not sure `DynamicType` would need to change in order for `DynamicData` to treat this differently.

I suggest not using MEMBER_ID_INVALID since that would make it seem like `get_member_id_by_name` failed.

ClaytonCalabrese: `get_member_id_by-name` isn't being used in this case and it would fail regardless because the discriminator is not a DynamicTypeMember of the DynamicType of the union.

dds/DCPS/Recorder.cpp
```diff
@@ -21,6 +21,21 @@ Recorder::~Recorder()
 {
 }
 
+XTypes::DynamicData Recorder::get_dynamic_data(const RawDataSample& sample) {
+
+  Encoding enc(Encoding::KIND_XCDR2, sample.header_.byte_order_ ? ENDIAN_LITTLE : ENDIAN_BIG);
```
iguessthislldo, line 26: The `Encoding` object should be read from the encapsulation header to make sure the encoding is correct.

dds/DCPS/Recorder.cpp
```diff
@@ -21,6 +21,21 @@ Recorder::~Recorder()
 {
 }
 
+XTypes::DynamicData Recorder::get_dynamic_data(const RawDataSample& sample) {
+
```
iguessthislldo, line 25:
```suggestion
XTypes::DynamicData Recorder::get_dynamic_data(const RawDataSample& sample)
{
```

dds/DCPS/Recorder.h
```diff
@@ -13,9 +13,9 @@
 #include "RcHandle_T.h"
 #include "RawDataSample.h"
 #include "dds/DdsDcpsInfrastructureC.h"
+#include "GuidConverter.h"
 
-#include "XTypes/TypeObject.h"
-#include "XTypes/DynamicType.h"
+#include "XTypes/DynamicData.h"
```
iguessthislldo, line 18:
```
#include "XTypes/DynamicData.h"

```

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2237,349 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+  OpenDDS::DCPS::String type_name;
+  OpenDDS::XTypes::DynamicData temp_dd;
+  OpenDDS::XTypes::DynamicType_rch temp_dt;
+  switch (dt->get_kind()) {
+  case OpenDDS::XTypes::TK_INT8: {
+    ACE_CDR::Int8 my_int8;
+    if (dd.get_int8_value(my_int8, 0) != DDS::RETCODE_OK) {
+      ACE_ERROR_RETURN((LM_ERROR, "(%P|%t) Error: print_dynamic_data - failed to get_int8_value\n"), false);
+    }
+    type_string += " = " + std::to_string(my_int8) + "\n";
```
iguessthislldo, line 2252: `to_string` is C++11.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2237,349 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+  // case OpenDDS::XTypes::TK_BITSET:
+  case OpenDDS::XTypes::TK_ALIAS: {
+    if (!print_dynamic_data(dd, dt->get_descriptor().base_type, type_string, indent)) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) print_dynamic_data: failed to read alias\n"));
+    }
+    break;
+  }
+  case OpenDDS::XTypes::TK_SEQUENCE: {
+    OpenDDS::DCPS::String temp_indent = indent;
+    indent += "  ";
```
iguessthislldo, line 2394: The `Printable` object can manage indents. It also allows what the indent is to be changed.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2237,349 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+{
+  OpenDDS::DCPS::String member_name;
+  OpenDDS::DCPS::String type_name;
+  OpenDDS::XTypes::DynamicData temp_dd;
+  OpenDDS::XTypes::DynamicType_rch temp_dt;
+  switch (dt->get_kind()) {
+  case OpenDDS::XTypes::TK_INT8: {
+    ACE_CDR::Int8 my_int8;
+    if (dd.get_int8_value(my_int8, 0) != DDS::RETCODE_OK) {
+      ACE_ERROR_RETURN((LM_ERROR, "(%P|%t) Error: print_dynamic_data - failed to get_int8_value\n"), false);
```
iguessthislldo, line 2250: Please merge in `xtypes_dynamic` to get `log_level` and bring the logging in line with the new logging guidelines https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#usage-in-opendds.

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -304,6 +309,8 @@ class OpenDDS_Dcps_Export DynamicData {
   ACE_CDR::ULong item_count_;
 };
 
+bool OpenDDS_Dcps_Export print_dynamic_data(OpenDDS::XTypes::DynamicData dd, OpenDDS::XTypes::DynamicType_rch dt, OpenDDS::DCPS::String& type_string, OpenDDS::DCPS::String indent);
```
iguessthislldo, line 312:
```suggestion
bool OpenDDS_Dcps_Export print_dynamic_data(
  DynamicData dd, DynamicType_rch dt, DCPS::String& type_string, DCPS::String indent);
```

dds/DCPS/XTypes/TypeObject.h
```diff
@@ -828,6 +828,7 @@ namespace XTypes {
   // ID of a type member
   typedef ACE_CDR::ULong MemberId;
   const ACE_CDR::ULong MEMBER_ID_INVALID = UINT32_MAX;
+  const ACE_CDR::ULong DISCRIMINATOR_ID = 4294967294U;
```
iguessthislldo, line 831:
```suggestion
  const ACE_CDR::ULong DISCRIMINATOR_ID = UINT32_MAX - 1;
```

Also maybe add a note that this one is implementation specific.

tests/DCPS/XTypes/DynamicTypes/Pub/XTypesDynamicPub.cpp
```diff
@@ -0,0 +1,165 @@
...
+using namespace DDS;
+/// parse the command line arguments
+int parse_args (int argc, ACE_TCHAR *argv[])
+{
+  u_long mask =  ACE_LOG_MSG->priority_mask(ACE_Log_Msg::PROCESS);
+  ACE_LOG_MSG->priority_mask(mask | LM_TRACE | LM_DEBUG, ACE_Log_Msg::PROCESS);
+  ACE_Arg_Shifter arg_shifter (argc, argv);
+  return 0;
+}
+void stru_narrow_write(DataWriter_var dw) {
```
iguessthislldo, line 33:
```suggestion
}

void stru_narrow_write(DataWriter_var dw)
{
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,181 @@
...
+  ACE_Thread_Semaphore sem_;
+};
+
+
+int run_test(int argc, ACE_TCHAR *argv[]){
+  int ret_val = 0;
+  try {
+    OpenDDS::DCPS::String topic_name = argv[1];
+    OpenDDS::DCPS::String type_name = argv[2];
+    DDS::DomainId_t domainid = ACE_OS::atoi(argv[3]);
```
iguessthislldo, line 82: Please check that there are that many arguments and don't check the arguments until after `TheParticipantFactoryWithArgs`. `TheParticipantFactoryWithArgs` pulls out the `-DCPS*` options.

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,181 @@
...
+      // }
+      // ret_val = recorder_listener->ret_val_;
+      // service->delete_recorder(recorder);
+    }
+    // participant->delete_contained_entities();
+    // dpf->delete_participant(participant);
+    // TheServiceParticipant->shutdown();
+  } catch (const CORBA::Exception& e) {
+    e._tao_print_exception("Exception caught in main():");
+    return -1;
```
iguessthislldo, line 164: Please don't return negative numbers from `main`. They can't be properly checked from a shell.

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
```diff
@@ -0,0 +1,10 @@
+project(Dynamic_Data_Monitor): dcps_test, dcps_rtps_udp, msvc_bigobj {
+  dcps_ts_flags += -Gxtypes-complete
+  exename   = dynamic_data_monitor
```
iguessthislldo, line 3:
```suggestion
  exename   = dynamic-data-monitor
```

Underscores are unusual in command names.

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
```diff
@@ -0,0 +1,10 @@
...
+
+  includes += ..
```
iguessthislldo, line 5: I don't think this is needed.

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicData.cpp
ClaytonCalabrese: Is this change necessary? It seems like unnecessary complexity for this case.

iguessthislldo: This is exactly what the `Printer` was made for is all I'm trying to say. We don't have to do anything about it now, but when we rewrite this to use `ValueReader`/`ValueWriter` we should do it make by making a `PrinterValueWriter`.

dds/DCPS/Recorder.h
```diff
@@ -104,6 +105,7 @@ class OpenDDS_Dcps_Export Recorder
    */
   virtual RecorderListener_rch get_listener() = 0;
   DynamicTypeByPubId dt_map_;
+  OpenDDS::XTypes::DynamicData get_dynamic_data(const RawDataSample& sample);
```
mitza-oci, line 108:
```suggestion
  XTypes::DynamicData get_dynamic_data(const RawDataSample& sample);
```

dds/DCPS/Recorder.h
```diff
@@ -12,10 +12,11 @@
...
 #include "RcHandle_T.h"
 #include "RawDataSample.h"
-#include "dds/DdsDcpsInfrastructureC.h"
+#include "GuidConverter.h"
 
-#include "XTypes/TypeObject.h"
-#include "XTypes/DynamicType.h"
+#include "XTypes/DynamicData.h"
+
+#include <dds/DdsDcpsInfrastructureC.h>
```
mitza-oci, line 19: I guess something in the `#include`s is causing the error in build_u18_d0i0v1_sec_js0_FM-08, but I don't know what the root cause is.

Try changing line 125 to just use `bool`.  There is no need to use a typedef for `bool` when the type is always `bool`.  Maybe there's some other problem in CI.

Also try making the minimal change that you need in these `#includes` in the header, moving other `#include`s to the cpp.

dds/DCPS/RecorderImpl.cpp
```diff
@@ -228,13 +228,45 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
...
     RawDataSample rawSample(sample.header_,
                             static_cast<MessageId> (sample.header_.message_id_),
                             sample.header_.source_timestamp_sec_,
                             sample.header_.source_timestamp_nanosec_,
                             sample.header_.publication_id_,
                             sample.header_.byte_order_,
-                            sample.sample_.get());
+                            sample.sample_.get(),
+                            kind
+                            );
```
mitza-oci, line 269:
```suggestion
                            kind);
```

dds/DCPS/RecorderImpl.cpp
```diff
@@ -228,13 +228,45 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
 
   // we only support SAMPLE_DATA messages
   if (sample.header_.message_id_ == SAMPLE_DATA && listener_.in()) {
+    Encoding::Kind kind = Encoding::KIND_UNALIGNED_CDR;
+    if (sample.header_.cdr_encapsulation_) {
+      Encoding enc(Encoding::KIND_XCDR2, sample.header_.byte_order_ ? ENDIAN_LITTLE : ENDIAN_BIG);
```
mitza-oci, line 233:
```suggestion
      const Encoding enc(Encoding::KIND_XCDR2, sample.header_.byte_order_ ? ENDIAN_LITTLE : ENDIAN_BIG);
```

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,460 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
   }
 }
 
+bool print_dynamic_data(OpenDDS::XTypes::DynamicData dd, OpenDDS::XTypes::DynamicType_rch dt, OpenDDS::DCPS::String& type_string, OpenDDS::DCPS::String indent)
```
mitza-oci, line 2242: Remove all redundant namespace qualification (whole PR).

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -101,6 +104,8 @@ class OpenDDS_Dcps_Export DynamicData {
   DDS::ReturnCode_t get_string_values(CORBA::StringSeq& value, MemberId id);
   DDS::ReturnCode_t get_wstring_values(CORBA::WStringSeq& value, MemberId id);
 
+  DynamicType_rch get_type() { return type_; }
```
mitza-oci, line 107: This is just plain `type()` in the spec, right?  It's an IDL attribute.

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -304,6 +309,10 @@ class OpenDDS_Dcps_Export DynamicData {
   ACE_CDR::ULong item_count_;
 };
 
+bool OpenDDS_Dcps_Export print_dynamic_data(
+  OpenDDS::XTypes::DynamicData dd, OpenDDS::XTypes::DynamicType_rch dt,
+  OpenDDS::DCPS::String& type_string, OpenDDS::DCPS::String indent);
+
```
mitza-oci, line 315: If each DynamicData knows its type, is the second parameter needed?
Does DynamicData need to be passed by `const &`?
Should this be part of the library or part of the application?

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -11,26 +11,29 @@
...
+class DoubleSeq;
+class LongDoubleSeq;
+class CharSeq;
+class WCharSeq;
+class OctetSeq;
+class BooleanSeq;
+class StringSeq;
+class WStringSeq;
+}
+
```
mitza-oci, line 36: OK, this seems like it's the namespace conflict when versioned namespaces are enabled.  So perhaps my other comment isn't relevant (but I'm going to leave it in the review just in case).

If the forward declaration of these `*Seq` types is needed, they don't go within `OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL` because they belong to TAO.  So they would need the corresponding TAO macro.  However, I don't think that's the way to go.  Probably just include the headers you need for the sequence types, like `tao/FloatSeqC.h` for the sequence of floats, and so on for all built-in types.

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicData.h
ClaytonCalabrese: 1. No it is not, I will remove it.
2. No, that would cause errors when calling the other DynamicData functions like `get_*_value`.
3. I put it in the library so I could have both the app and the test both use the same code.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+        return false;
+      }
+    }
+    indent = temp_indent;
+    break;
+  }
+  case OpenDDS::XTypes::TK_STRUCTURE: {
+    OpenDDS::DCPS::String temp_indent = indent;
+    indent += "  ";
+    type_string += "struct " + dd.type()->get_name() + " {\n";
```
iguessthislldo, line 2506: This could be up to debate, but I'm not sure we should mimic IDL necessarily.

tests/DCPS/XTypes/DynamicTypes/Pub/XTypesDynamicPub.cpp
```diff
@@ -0,0 +1,163 @@
...
+  } else if (!ACE_OS::strcmp(type_name, ACE_TEXT("nested"))) {
+    ts_var = new Dynamic::outer_structTypeSupportImpl;
+  } else if (!ACE_OS::strcmp(type_name, ACE_TEXT("union")) ||
+             !ACE_OS::strcmp(type_name, ACE_TEXT("union_default"))) {
+    ts_var = new Dynamic::my_unionTypeSupportImpl;
+  }
+  ts_var->register_type(dp, ACE_TEXT_ALWAYS_CHAR(type_name));
+
+  Topic_var topic =
+      dp->create_topic ("recorder_topic", ACE_TEXT_ALWAYS_CHAR(type_name), TOPIC_QOS_DEFAULT,
```
iguessthislldo, line 121:
```suggestion
      dp->create_topic("recorder_topic", ACE_TEXT_ALWAYS_CHAR(type_name), TOPIC_QOS_DEFAULT,
```

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,239 @@
+/*
+ *
+ *
```
iguessthislldo, line 3:
```suggestion
```

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,239 @@
...
+    ACE_Time_Value timeout = ACE_OS::gettimeofday() + tv;
+    return sem_.acquire(timeout);
+  }
+  int ret_val_;
+private:
+  ACE_Thread_Semaphore sem_;
+};
+
+
+int run_test(int argc, ACE_TCHAR *argv[]){
```
iguessthislldo, line 142:
```suggestion
int run_test(int argc, ACE_TCHAR* argv[])
{
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include <dds/DCPS/transport/framework/TransportRegistry.h>
+#include <dds/DCPS/transport/framework/TransportConfig.h>
+#include <dds/DCPS/transport/framework/TransportInst.h>
+
```
iguessthislldo, line 11:
```suggestion
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <dds/DCPS/XTypes/TypeObject.h>
+#include <dds/DCPS/XTypes/DynamicData.h>
+#include "dds/DCPS/StaticIncludes.h"
+#if defined ACE_AS_STATIC_LIBS && !defined OPENDDS_SAFETY_PROFILE
+#include <dds/DCPS/transport/udp/Udp.h>
+#include <dds/DCPS/transport/multicast/Multicast.h>
+#include <dds/DCPS/RTPS/RtpsDiscovery.h>
+#include <dds/DCPS/transport/rtps_udp/RtpsUdp.h>
+#include <dds/DCPS/transport/shmem/Shmem.h>
+#endif
```
iguessthislldo, line 28: Do we need all this if this only supports pure RTPS?

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+  while (arg_shifter.is_anything_left())
+  {
+    // options:
+    //  -samples num_samples       defaults to infinite
+    //  -time num_seconds          defaults to infinite
+
+    const ACE_TCHAR *currentArg = 0;
+
+    if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("-samples"))) != 0) {
```
iguessthislldo, line 54:
```suggestion
    if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("--samples"))) != 0) {
```

Using double dashes gives use more flexibility later.

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+      num_samples = ACE_OS::atoi (currentArg);
+      arg_shifter.consume_arg();
+    } else if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("-time"))) != 0) {
```
iguessthislldo, line 57:
```suggestion
    } else if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("--time"))) != 0) {
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+  {
+    return sample_count_;
+  }
+private:
+  ACE_Thread_Semaphore sem_;
+  long sample_count_;
+};
+
+
+int run_test(int argc, ACE_TCHAR *argv[]){
```
iguessthislldo, line 122:
```suggestion
int run_test(int argc, ACE_TCHAR* argv[])
{
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+  while (arg_shifter.is_anything_left())
+  {
+    // options:
+    //  -samples num_samples       defaults to infinite
+    //  -time num_seconds          defaults to infinite
+
+    const ACE_TCHAR *currentArg = 0;
+
+    if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("-samples"))) != 0) {
+      num_samples = ACE_OS::atoi (currentArg);
```
iguessthislldo, line 55:
```suggestion
      num_samples = ACE_OS::atoi(currentArg);
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+      arg_shifter.consume_arg();
+    } else if ((currentArg = arg_shifter.get_the_parameter(ACE_TEXT("-time"))) != 0) {
+      num_seconds = ACE_OS::atoi (currentArg);
```
iguessthislldo, line 58:
```suggestion
      num_seconds = ACE_OS::atoi(currentArg);
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+      if (!recorder.in()) {
+        if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main(): create_recorder failed!\n"));
+        }
+        return 1;
+      }
+
+      ACE_Time_Value tv;
+      ACE_Time_Value start_tv = tv.now();
```
iguessthislldo, line 199: Use `SystemTimePoint` instead to avoid this?

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,235 @@
...
+    if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Error) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main(): failed to properly analyze sample!\n"));
+    }
+    return 1;
+  }
+  return ret_val;
+}
+
+int
+ACE_TMAIN(int argc, ACE_TCHAR *argv[])
```
iguessthislldo, line 230:
```suggestion
int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
```

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
```diff
@@ -0,0 +1,8 @@
+project(Dynamic_Data_Monitor): dcps_test, dcps_rtps_udp, msvc_bigobj {
```
iguessthislldo, line 1:
```suggestion
project(Dynamic_Data_Monitor): dcps_rtps_udp, msvc_bigobj {
```

It's not strictly a test anymore. Also does it need `msvc_bigobj`?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
   }
 }
 
+bool print_dynamic_data(DynamicData dd, DCPS::String& type_string, DCPS::String indent)
+{
+  OpenDDS::DCPS::String member_name;
```
iguessthislldo, line 2244:
```suggestion
  DCPS::String member_name;
```

And for the rest of the function.

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include <dds/DCPS/XTypes/TypeObject.h>
+#include <dds/DCPS/XTypes/DynamicData.h>
+#include "dds/DCPS/StaticIncludes.h"
+#if defined ACE_AS_STATIC_LIBS && !defined OPENDDS_SAFETY_PROFILE
+#include <dds/DCPS/transport/udp/Udp.h>
+#include <dds/DCPS/transport/multicast/Multicast.h>
+#include <dds/DCPS/RTPS/RtpsDiscovery.h>
+#include <dds/DCPS/transport/rtps_udp/RtpsUdp.h>
+#include <dds/DCPS/transport/shmem/Shmem.h>
+#endif
```
iguessthislldo, line 22: Limit to the includes that are needed?

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+#include <dds/DCPS/transport/framework/TransportRegistry.h>
+#include <dds/DCPS/transport/framework/TransportConfig.h>
+#include <dds/DCPS/transport/framework/TransportInst.h>
+
+#include <dds/DCPS/Recorder.h>
```
iguessthislldo, line 28:
```suggestion
#include <dds/DCPS/transport/framework/TransportRegistry.h>
#include <dds/DCPS/transport/framework/TransportConfig.h>
#include <dds/DCPS/transport/framework/TransportInst.h>
#include <dds/DCPS/Recorder.h>
```

iguessthislldo (changes requested):

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
```diff
@@ -0,0 +1,8 @@
...
+  dcps_ts_flags += -Gxtypes-complete
```
iguessthislldo, line 2:
```suggestion
```

iguessthislldo (changes requested):

dds/DCPS/RecorderImpl.cpp
```diff
@@ -228,13 +228,44 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
...
+          kind = Encoding::KIND_XCDR2;
+          break;
+        default:
+          if (log_level >= LogLevel::Notice) {
+            ACE_ERROR((LM_NOTICE, ACE_TEXT("(%P|%t) NOTICE: RecorderImpl::data_received: ")
+              ACE_TEXT("Unsupported Encoding\n")));
+            }
+          break;
+        }
+      }
```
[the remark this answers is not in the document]
ClaytonCalabrese: We would need to pass the expected encoding, which was causing an error when not matching.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+  switch (dd.type()->get_kind()) {
+  case OpenDDS::XTypes::TK_INT8: {
+    ACE_CDR::Int8 my_int8;
+    if (dd.get_int8_value(my_int8, 0) != DDS::RETCODE_OK) {
+      if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: print_dynamic_data: failed to get_int8_value\n"));
+      }
+      return false;
+    }
+    type_string += " = " + OpenDDS::DCPS::to_dds_string(my_int8) + "\n";
```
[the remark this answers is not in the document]
ClaytonCalabrese: Just tried this. It causes problems in nested Structs/ Unions.

Would print like:

```
struct ::Dynamic::outer_struct {
  ::Dynamic::inner_struct **=** isstruct ::Dynamic::inner_struct {
    Int32 l = 5
  };
};
```

Bolded part is the problem.  As you mentioned in a different conversation, we probably want to discuss what we want the printed type to look like anyways, so maybe that would clear this up.

dds/DCPS/XTypes/DynamicData.h
ClaytonCalabrese: There is a circular include problem due to ServiceParticipant.h including Recorder.h which includes DynamicData.h which includes the sequence include files which each include ServiceParticipant.h. This is why I forward declared these here.

Do you have a suggestion on how to solve this issue?

mitza-oci: Note that my comment yesterday was about `tao/FloatSeqC.h`, not an OpenDDS header.

dds/DCPS/XTypes/DynamicData.h
mitza-oci: 2. then pass by reference - we don't want a copy

dds/DCPS/XTypes/DynamicData.cpp
mitza-oci: Is there an example from the sample dissector that would be helpful here?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+      ACE_CDR::Int8 my_int8;
+      if (dd.get_int8_value(my_int8, XTypes::DISCRIMINATOR_ID) != DDS::RETCODE_OK) {
+        if (DCPS::log_level >= DCPS::LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: print_dynamic_data: failed to get_int8_value\n"));
+        }
+        return false;
+      }
+      type_string += " = " + DCPS::to_dds_string(my_int8) + "\n";
+      break;
+    }
```
sonndinh, line 2544: Can we factor out the code in these cases to share with the code in the outer `switch`? Looks like they are similar enough.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+      if (dd.get_uint16_value(my_ushort, XTypes::DISCRIMINATOR_ID) != DDS::RETCODE_OK) {
+        if (DCPS::log_level >= DCPS::LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: print_dynamic_data: failed to get_uint16_value\n"));
+        }
+        return false;
+      }
+      type_string += " = " + DCPS::to_dds_string(my_ushort) + "\n";
+      break;
+    }
+    case XTypes::TK_ENUM:
```
sonndinh, line 2578: `bit_bound` of enums should be checked to read the correct integer type.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+      DCPS::String decoy_string;
+      dd.type()->get_all_members(dtmbi);
+      for (XTypes::DynamicTypeMembersById::iterator iter = dtmbi.begin(); iter != dtmbi.end(); ++iter) {
+        dd.type()->get_member(temp_dtm, iter->first);
+        if (dd.get_complex_value(dd, iter->first) == DDS::RETCODE_OK) {
+          if (print_dynamic_data(dd, decoy_string, indent)) {
+            member_name = temp_dtm->get_descriptor().name;
+            type_name = temp_dtm->get_descriptor().get_type()->get_descriptor().name;
+            type_string += indent + type_name + " " + member_name;
+            print_dynamic_data(dd, type_string, indent);
```
sonndinh, line 2680: This prints the selected member the second time, is it intended?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+            break;
```
sonndinh, line 2681: Looks like this `break` can be moved to outside of the current `if` block.

dds/DCPS/Recorder.cpp
```diff
@@ -21,6 +21,23 @@ Recorder::~Recorder()
 {
 }
 
+XTypes::DynamicData Recorder::get_dynamic_data(const RawDataSample& sample)
+{
+  Encoding enc(sample.encoding_kind_, sample.header_.byte_order_ ? ENDIAN_LITTLE : ENDIAN_BIG);
+  const DynamicTypeByPubId::const_iterator dt_found = dt_map_.find(sample.publication_id_);
+  if (dt_found == dt_map_.end()) {
+    if (log_level >= LogLevel::Error) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: on_sample_data_received: "
```
sonndinh, line 30:
```suggestion
      ACE_ERROR((LM_ERROR, "(%P|%t) on_sample_data_received: "
```

dds/DCPS/Recorder.cpp
```diff
@@ -21,6 +21,23 @@ Recorder::~Recorder()
...
+        "failed to find GUID: %C in DynamicTypeByPubId.\n", LogGuid(sample.publication_id_).c_str()));
+    }
+    return XTypes::DynamicData();
+  } else {
+    XTypes::DynamicType_rch dt = dt_found->second;
+    XTypes::DynamicData dd(sample.sample_.get(), enc, dt);
+    return dd;
```
sonndinh, line 37:
```suggestion
    return XTypes::DynamicData(sample.sample_.get(), enc, dt);
```

sonndinh (changes requested):

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2239,458 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+{
+  DCPS::String member_name;
+  DCPS::String type_name;
+  XTypes::DynamicData temp_dd;
+  switch (dd.type()->get_kind()) {
+  case XTypes::TK_INT8: {
+    ACE_CDR::Int8 my_int8;
+    if (dd.get_int8_value(my_int8, 0) != DDS::RETCODE_OK) {
+      if (DCPS::log_level >= DCPS::LogLevel::Error) {
+        ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: print_dynamic_data: failed to get_int8_value\n"));
```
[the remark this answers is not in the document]
ClaytonCalabrese: The docs seems to indicate that the `ERROR:` should be in the message.

https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#id2

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -11,26 +11,29 @@
...
-#include <dds/CorbaSeq/ULongLongSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/FloatSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/DoubleSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/LongDoubleSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/CharSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/WCharSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/OctetSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/BooleanSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/StringSeqTypeSupportImpl.h>
-#include <dds/CorbaSeq/WStringSeqTypeSupportImpl.h>
```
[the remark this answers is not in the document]
ClaytonCalabrese: There was a circular include problem due to ServiceParticipant.h including Recorder.h which includes DynamicData.h which includes the sequence include files which each include ServiceParticipant.h. I had to move these to the .cpp to avoid this.  Adam gave me a replacement to the forwad declarations of including `/tao/*SeqC.h`.

dds/DCPS/XTypes/DynamicData.cpp
ClaytonCalabrese: Yes it is intended. The first call to `print_dynamic_data` is passed `decoy_string` which is a dummy to catch/suppress the errors that result when the non-selected branches are printed.  Once the selected branch is found, the if will be entered and `print_dynamic_data` will be called with `type_string`.

dds/DCPS/XTypes/DynamicData.cpp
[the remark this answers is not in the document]
iguessthislldo: Yes, that is correct. I don't think I got it to it in my review, but these should probably be "notice" messages since there's an error return value.

dds/DCPS/XTypes/DynamicData.cpp
iguessthislldo: I suppose. This is a screenshot of what that looks like from my article: 

Another example of non-JSON text representation would be the `Printer` though, it's basically the same. Here I added the printer to DevGuideMessenger along with a nested struct because I didn't remember how it handled that.

```
    from: "Comic Book Guy"
    subject: "Review"
    subject_id: 107
    text: "Worst. Movie. Ever."
    count: 8
    n:
        value: 2147483647
```

This is probably the first time this code has ran since I worked on it so I'm glad to see it still works. I don't have an issue with printing out the type name ([`Printer` has a todo for this](https://github.com/objectcomputing/OpenDDS/blob/05c82664fc481a2d9cc620d3eb4f00f9911c14ff/dds/DCPS/Printer.h#L41)) or things like `struct`. I do thing we could omit the braces though as the indent should be enough.

dds/DCPS/RecorderImpl.cpp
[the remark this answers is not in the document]
iguessthislldo: You mean the extensibility? Yes, it could be modified to not expect a specific extensibility so we don't have to duplicate the code here.

dds/DCPS/XTypes/DynamicData.cpp
sonndinh: `dd.get_complex_value(dd, iter->first)` returns `RETCODE_OK` if `iter->first` is the ID of the selected branch and returns `RETCODE_ERROR` in other cases. So the code inside the outer `if` block only executes for the selected branch.

mitza-oci (approved):

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+using namespace OpenDDS::DCPS;
+
+long num_samples = 0;
+long num_seconds = 0;
+// parse the command line arguments
+int parse_args (int& argc, ACE_TCHAR *argv[])
+{
+  u_long mask =  ACE_LOG_MSG->priority_mask(ACE_Log_Msg::PROCESS);
+  ACE_LOG_MSG->priority_mask(mask | LM_TRACE | LM_DEBUG, ACE_Log_Msg::PROCESS);
+  ACE_Arg_Shifter arg_shifter (argc, argv);
```
iguessthislldo, line 33:
```suggestion
  ACE_Arg_Shifter arg_shifter(argc, argv);
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+  while (arg_shifter.is_anything_left())
+  {
```
iguessthislldo, line 36:
```suggestion
  while (arg_shifter.is_anything_left()) {
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+    // options:
+    //  -samples num_samples       defaults to infinite
+    //  -time num_seconds          defaults to infinite
+
+    const ACE_TCHAR *currentArg = 0;
```
iguessthislldo, line 41:
```suggestion
    const ACE_TCHAR* currentArg = 0;
```

dds/DCPS/RecorderImpl.cpp
```diff
@@ -228,13 +228,24 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
...
+    Encoding::Kind kind = Encoding::KIND_UNALIGNED_CDR;
+    if (sample.header_.cdr_encapsulation_) {
+      Encoding enc;
+      Serializer ser(sample.sample_.get(), enc);
+      EncapsulationHeader encap;
+      if (ser >> encap) {
+        encap.to_encoding(enc, FINAL, false);
+        kind = enc.kind();
+      }
+    }
```
iguessthislldo, line 240: Could we have a setting in the `Recorder` to skip this? That would be a simple way leave it open to be manually parsed by the user.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -9,12 +9,100 @@
...
+  case '\a':
+    return o << "\\a";
+  }
+  const unsigned cvalue = char_value(value);
+  if (cvalue <= UCHAR_MAX && isprint(cvalue)) {
+    return o << static_cast<char>(value);
+  }
+  return hex_value(o << "\\x", cvalue, sizeof(CharType) == 1 ? 1 : 2);
+}
+}
```
iguessthislldo, line 104: Move this into a common header that's shared with `dds_generator.h` to avoid code duplication.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2302,416 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+      if (DCPS::log_level >= DCPS::LogLevel::Notice) {
+        ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: print_dynamic_data: failed to get_wstring_value\n"));
+      }
+      return false;
+    }
+    std::stringstream os;
+    for (size_t i = 0; my_wstring[i] != 0; ++i) {
+      char_helper<ACE_CDR::WChar>(os, my_wstring[i]);
+    }
+    type_string += " = L\"" + os.str() + "\"\n";
```
iguessthislldo, line 2571: Turn this into a `string_helper` template function and also apply it to narrow strings.

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+    std::cout << my_type;
+    ++sample_count_;
+    if (num_samples > 0 && sample_count_ >= num_samples) {
+      gc_->set_trigger_value(true);
+    }
+  }
+
+  virtual void on_recorder_matched(Recorder*,
+                                   const ::DDS::SubscriptionMatchedStatus& status )
```
iguessthislldo, line 85:
```suggestion
                                   const ::DDS::SubscriptionMatchedStatus& status)
```

iguessthislldo (changes requested):

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  int wait(const ACE_Time_Value & tv)
+  {
+    ACE_Time_Value timeout = ACE_OS::gettimeofday() + tv;
+    return sem_.acquire(timeout);
+  }
+  int ret_val_;
+private:
```
sonndinh, line 139:
```suggestion
  }

  int ret_val_;

private:
```

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,246 @@
...
+      if (!topic) {
+        if (log_level >= LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main(): create_topic failed!\n"));
+        }
+        return 1;
+      }
+
+      ACE_Time_Value wait_time(60, 0);
+
+      RcHandle<TestRecorderListener> recorder_listener = make_rch<TestRecorderListener> (gc);
```
sonndinh, line 191:
```suggestion
      RcHandle<TestRecorderListener> recorder_listener = make_rch<TestRecorderListener>(gc);
```

tests/DCPS/XTypes/DynamicTypes/run_test.pl
```diff
@@ -0,0 +1,75 @@
...
+use PerlDDS::Run_Test;
+use Getopt::Long;
+use strict;
+
+my $test = new PerlDDS::TestFramework();
+$test->enable_console_logging();
+
+my $status = 0;
+
+sub run_test {
```
sonndinh, line 16: Maybe use a for loop in this function that loops through the different types.

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+  return 0;
+}
+
+class TestRecorderListener : public RecorderListener
+{
+public:
+  explicit TestRecorderListener(DDS::GuardCondition_var gc)
+    : sem_(0),
+      sample_count_(0),
+      gc_(gc)
```
sonndinh, line 63:
```suggestion
    : sem_(0)
    , sample_count_(0)
    , gc_(gc)
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,232 @@
...
+                                       DEFAULT_STATUS_MASK);
+
+      if (!topic) {
+        if (log_level >= LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main(): create_topic failed!\n"));
+        }
+        return 1;
+      }
+
+      RcHandle<TestRecorderListener> recorder_listener = make_rch<TestRecorderListener> (gc);
```
sonndinh, line 172:
```suggestion
      RcHandle<TestRecorderListener> recorder_listener = make_rch<TestRecorderListener>(gc);
```

sonndinh (changes requested):
The Dynamic API integration test is currently put under the subdirectory tests/DCPS/XTypes/DynamicTypes, but I think we may want to move it to a separate subdirectory outside of tests/DCPS/XTypes. I haven't seen the test share any code or interact with the existing XTypes test, so it's better to have its own folder.

tests/DCPS/XTypes/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include <ace/Semaphore.h>
+#include <ace/Thread_Semaphore.h>
+
+#include <iostream>
+#include <sstream>
+
+using namespace OpenDDS::DCPS;
+bool sample_read = false;
+class TestRecorderListener : public RecorderListener
+{
```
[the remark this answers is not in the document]
ClaytonCalabrese: Shouldn't a class have the `{` on a new line?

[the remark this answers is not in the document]
sonndinh: No, classes have `{` on the same line. It's in the guideline.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2302,416 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
   }
 }
 
+bool print_integral_value(DynamicData& dd, DCPS::String& type_string, DynamicType_rch dt) {
```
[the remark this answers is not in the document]
iguessthislldo: This hasn't changed.

tests/DCPS/XTypes/.gitignore
```diff
@@ -25,3 +25,4 @@ Sub/XTypesSubTypeSupportImpl.cpp
 Sub/XTypesSubTypeSupportImpl.h
 Sub/XTypesSubTypeSupportS.cpp
 Sub/XTypesSubTypeSupportS.h
+
```
iguessthislldo, line 28:
```suggestion
```

dds/idl/value_helper.h
```diff
@@ -0,0 +1,111 @@
...
+
+#ifndef value_helper_H
+#define value_helper_H
+
+#include <ostream>
+#include <iostream>
+#include <iomanip>
+
+#include <dds/DCPS/RestoreOutputStreamState.h>
+#include <dds/DCPS/Definitions.h>
```
iguessthislldo, line 14: Swap these include groups around.

dds/idl/value_helper.h
```diff
@@ -0,0 +1,111 @@
+/*
```
iguessthislldo, line 1: This file should probably be in `dds/DCPS` because there are several headers in `dds/DCPS` that we include in `dds/idl`, but I don't think there's any that do the opposite of that.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2235,416 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
...
+      }
+    }
+    indent = temp_indent;
+    break;
+  }
+  case TK_ARRAY: {
+    DCPS::String temp_indent = indent;
+    indent += "  ";
+    LBound bound = dd.type()->get_descriptor().bound[0];
+    type_string += "[" + DCPS::to_dds_string(bound) + "] =\n";
```
iguessthislldo, line 2544: I'd prefer not to have braces at all in the output, but if we are going to have them in structs and unions I think we should be consistent and have them in arrays and sequences.

tests/DCPS/DynamicTypes/Recorder/XTypesDynamicRecorder.cpp
```diff
@@ -0,0 +1,251 @@
...
+    , gc_(gc)
+  {
+  }
+
+  virtual void on_sample_data_received(Recorder* rec,
+                                       const RawDataSample& sample)
+  {
+    OpenDDS::XTypes::DynamicData dd = rec->get_dynamic_data(sample);
+    String my_type = "";
+    String indent = "";
```
iguessthislldo, line 39:
```suggestion
    String my_type;
    String indent;
```

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -0,0 +1,230 @@
...
+    , gc_(gc)
+  {
+  }
+
+  virtual void on_sample_data_received(Recorder* rec,
+                                       const RawDataSample& sample)
+  {
+    OpenDDS::XTypes::DynamicData dd = rec->get_dynamic_data(sample);
+    String my_type = "";
+    String indent = "";
```
iguessthislldo, line 70:
```suggestion
    String my_type;
    String indent;
```

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicData.cpp
ClaytonCalabrese: I'll remove them from structs and unions.

dds/DCPS/ValueHelper.h
```diff
@@ -0,0 +1,111 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_VALUE_HELPER_H
+#define OPENDDS_DCPS_VALUE_HELPER_H
+
+#include <dds/DCPS/RestoreOutputStreamState.h>
+#include <dds/DCPS/Definitions.h>
```
iguessthislldo, line 10:
```suggestion
#include "RestoreOutputStreamState.h"
#include "Definitions.h"
```

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -2193,6 +2235,414 @@ const char* DynamicData::typekind_to_string(TypeKind tk) const
   }
 }
 
+bool print_integral_value(DynamicData& dd, DCPS::String& type_string, DynamicType_rch dt) {
```
iguessthislldo, line 2238:
```suggestion
bool print_integral_value(DynamicData& dd, DCPS::String& type_string, DynamicType_rch dt)
{
```

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/Serializer.h
```diff
@@ -69,7 +69,8 @@ String endianness_to_string(Endianness endianness);
 enum Extensibility {
   FINAL,
   APPENDABLE,
-  MUTABLE
+  MUTABLE,
+  ANY_EXTENSIBILITY
```
sonndinh, line 73: For naming, something like `EXTENSIBILITY_ANY` seems preferred, but it's not a big deal.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -78,7 +99,6 @@ DynamicData::~DynamicData()
 
 void DynamicData::copy(const DynamicData& other)
 {
-  ACE_Message_Block::release(chain_);
```
sonndinh, line 81: This needs to keep because we want to release the old message chain before storing a new chain into `chain_`.

tests/DCPS/DynamicTypes/run_test.pl
```diff
@@ -0,0 +1,39 @@
...
+
+  $test->process("reader_$test_name_param", './Recorder/xtypes_dynamic_recorder', join(' ', @reader_args));
+  $test->start_process("reader_$test_name_param");
+
+  $test->process("writer_$test_name_param", './Pub/xtypes_dynamic_pub', join(' ', @writer_args));
+  $test->start_process("writer_$test_name_param");
+
+  $status |= $test->wait_kill("reader_$test_name_param", 15);
+  $status |= $test->wait_kill("writer_$test_name_param", 15);
+  }
```
sonndinh, line 30: Add indentation for the body of this for-loop.

sonndinh (changes requested):

sonndinh (approved):

PR #3130, opened 2021-11-03
new_length can be const

jwillemsen:
* dds/idl/marshal_generator.cpp:


jwillemsen:
Can't be const in all cases

PR #3131, opened 2021-11-03
Make members of SerializedSizeBound const

jwillemsen:
* dds/DCPS/TypeSupportImpl.h:


mitza-oci:
Moving to draft since it's not compiling successfully on multiple builds

Issue #2946, opened 2021-08-26
Demarshal problem when using @appendable and types evolve


jwillemsen:
I have an application v1 that uses the following IDL type

```
module DDS_Native
{
  typedef string<128> color_type;
  @nested(FALSE)
  @appendable
  struct ShapeType {
    @key color_type color;
    long x;
    long y;
    long shapesize;
  };
}; // module DDS_Native
```

Now this application evolves into v2 which uses uses the following IDL definition which is the same type but with an additional field

```
module DDS_Native
{
  typedef string<128> color_type;
  @nested(FALSE)
  @appendable
  struct ShapeType {
    @key color_type color;
    long x;
    long y;
    long shapesize;
    long angle;
  };
}; // module DDS_Native
```

When application v2 publishes a sample v1 does receive the data, the `angle` is just dropped. But when v1 publishes a sample than v2 receives the following error:
```
(17330|17339) ERROR DDS_Native::ShapeTypeDataReaderImpl::dds_demarshal deserialization failed, dropping sample.
```

[openddslog.txt](https://github.com/objectcomputing/OpenDDS/files/7060593/openddslog.txt)

jwillemsen:
This issue was not fixed, see attached log, when having v1 published v2 can't demarshal it, even with the fixes of #3040 

[extensibility_20211003.log](https://github.com/objectcomputing/OpenDDS/files/7272949/extensibility_20211003.log)

jwillemsen:
[extensibliy_transportlogging_20211003.log](https://github.com/objectcomputing/OpenDDS/files/7272954/extensibliy_transportlogging_20211003.log)

jwillemsen:
Full log

[extensibliy_transportlogging_20211003.log](https://github.com/objectcomputing/OpenDDS/files/7272961/extensibliy_transportlogging_20211003.log)

jwillemsen:
Looks we still get unaligned CDR, anything special we should do in our settings?

```
(8948|8948) DataReaderImpl::setup_deserialization: Setup successfully with the following data representations: Unaligned CDR
```

mitza-oci:
Is the new configuration option from #3040 enabled?

jwillemsen:
> Is the new configuration option from #3040 enabled?

[...]

PR #3135, opened 2021-11-03
CDR Encapsulation option for non-rtps_udp transports: Follow up fixes

ClaytonCalabrese:
The previous PR for this option was here https://github.com/objectcomputing/OpenDDS/pull/3040

However, following discussion of the issue here:  https://github.com/objectcomputing/OpenDDS/issues/2946 , it was determined additional work was needed.

Rather than using a config option to determine encapsulation, it is now determined by the default of the transport (false for anything besides RTPS_UDP),

If the user has explicitly (not by default) selected XCDR/XCDR2 as their Data Representation QoS policy, it is enabled for that transport, regardless of the default. 

As a result of these changes, I made a couple updates to the XTypes test to ensure the `--tcp` option runs properly.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -1278,6 +1278,11 @@ DataReaderImpl::enable()
       return DDS::RETCODE_ERROR;
     }
 
+    const DDS::ReturnCode_t setup_deserialization_result = setup_deserialization();
+    if (setup_deserialization_result != DDS::RETCODE_OK) {
+      return setup_deserialization_result;
+    }
+
     const TransportLocatorSeq& trans_conf_info = connection_info();
 
     CORBA::String_var filterClassName = "";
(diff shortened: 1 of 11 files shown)


dds/DCPS/DataWriterImpl_T.h
```diff
@@ -327,6 +327,13 @@ DataWriterImpl_T
   {
     const DDS::DataRepresentationIdSeq repIds =
       get_effective_data_rep_qos(qos_.representation.value, false);
+    if (qos_.representation.value.length() > 0 &&
+       (qos_.representation.value[0] == DDS::XCDR_DATA_REPRESENTATION ||
+        qos_.representation.value[0] == DDS::XCDR2_DATA_REPRESENTATION))
```
iguessthislldo, line 332: I know we don't support it, but `XML_DATA_REPRESENTATION` also uses encapsulation. That could also be added, but it would be simpler if we could assume that if `representation.value` isn't empty then we need encapsulation.

iguessthislldo (changes requested):

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -327,6 +327,11 @@ DataWriterImpl_T
   {
     const DDS::DataRepresentationIdSeq repIds =
       get_effective_data_rep_qos(qos_.representation.value, false);
+    if (qos_.representation.value.length() > 0)
+    {
```
iguessthislldo, line 331:
```suggestion
    if (qos_.representation.value.length() > 0) {
```

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3384,57 +3379,6 @@ DataReaderImpl::get_ice_endpoint()
   return TransportClient::get_ice_endpoint();
 }
 
-DDS::ReturnCode_t DataReaderImpl::setup_deserialization()
```
iguessthislldo, line 3387: I think I see what you mean by part of this being already being done in match. However the checks being done in `dds_demarshal` and `lookup_instance` are more direct as it doesn't rely on what QoS is saying. So it's not going to get triggered under normal conditions, but I think it would be useful as a final internal check. For example if there was a bug where QoS and the encoding didn't line up anymore, this could help reveal that. Even if that wasn't the case or we wanted to get rid of those checks anyways there are things being done here that I'm not certain are being done anywhere else. I used the ["Setup successfully with the following data representation" message in the issue that prompted these changes](https://github.com/objectcomputing/OpenDDS/issues/2946#issuecomment-906529994).

So I don't think we should remove any of this unless it's actually causing an issue.

iguessthislldo (changes requested):

dds/DCPS/DataReaderImpl.cpp
iguessthislldo: Well actually the encoding comes from the QoS, so that's a bad example unless something was really wrong, like one side turned the QoS into the wrong encoding. I guess the more possible scenario would if something was wrong with matching.

ClaytonCalabrese: Okay. I readded the removed functions. It still checks `cdr_encapsulation()` but this is used to determine whether the default for a reader should be unaligned cdr (blank QoS, non rtps_udp transport). The reason all this was removed in the first place was it was forcing unaligned, regardless of the QoS, this solution should be more elegant.

iguessthislldo (approved):
I think the code looks alright to me, but there's a failure in the XTypes test.

PR #3142, opened 2021-11-05
Check for invalid member IDs

sonndinh:


dds/idl/be_global.cpp
```diff
@@ -1009,6 +1009,9 @@ OpenDDS::XTypes::MemberId BE_GlobalData::compute_id(AST_Field* field, AutoidKind
     mid = member_id++;
   }
 
+  if (mid > OpenDDS::DCPS::Serializer::MEMBER_ID_MAX) {
+    be_util::misc_error_and_abort("Member id exceeds the maximum allowed value");
```
mitza-oci, line 1013: Could this have more context for the user?  Which type and member is the problem?

mitza-oci (approved):

PR #3149, opened 2021-11-08
SporadicTask optimizations are ineffective, buggy, and complicated

jrw972:
Problem
-------

There are two optimizations for SporadicTask:

1. The first optimization attempts to only submit a schedule command
   if not scheduled or the scheduled dispatch time is nearer than the
   desired dispatch time.

   This logic is controlled by variables that are only set by the
   reactor thread.  Thus, a non-reactor thread attempting to schedule
   could still create a large number of ScheduleCommands before the
   reactor thread gets around to processing the first one.

   The optmization that examines the next dispatch time changes the
   semantics of SporadicTask in that it is implied that the user can
   schedule a dispatch nearer in time without canceling.  With the
   introduction of code that modulates a SporadicTask, this created a
   bug where calls to cancel the timer were dropped and/or not
   written.  By not canceling the timer, the `scheduled_` flag was
   never reset meaning that scheduling for an earlier dispatch never
   occurred.

2. The second optmization attempts to reduce the number of calls into
   the reactor itself.  This took the form of hooks around the
   lifecycle of a ReactorInterceptor::Command.  For SporadicTask, the
   goal of the implementation was to skip commands that were
   effectively no-ops.

Solution
--------

The new design of SporadicTask is based on the idea that there is a
desired state provided by the user (scheduled or not scheduled) and
there is an actual state which is the result of interacting with the
reactor.  When scheduling or canceling, the same
ReactorInterceptor::Command is submitted to the reactor.  When
executed, the command attempts to make the actual state the desired
state if they differ.  To avoid dynamic allocation, the command is
allocated once and only put on the ReactorInterceptors queue if it is
not on the queue already.

Notes
-----

The pattern for SporadicTask should be adopted by PeriodicTask,
MultiTask, and other uses of the ReactorInterceptor.

### DIFF
--- a/dds/DCPS/ReactorInterceptor.h
+++ b/dds/DCPS/ReactorInterceptor.h
@@ -31,17 +31,31 @@ public:
   class Command
   : public RcObject {
   public:
-    Command() : executed_(false), condition_(mutex_), reactor_(0) {}
+    Command() : executed_(false), on_queue_(false), condition_(mutex_), reactor_(0) {}
     virtual ~Command() { }
-    virtual bool should_execute() const { return true; }
-    virtual void will_execute() {}
-    virtual void execute() = 0;
-    virtual void queue_flushed() {}
 
-    void reset()
+    bool reset()
+    {
+      ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, mutex_, false);
+      executed_ = false;
+      bool retval = on_queue_;
+      on_queue_ = true;
+      return retval;
+    }
+
+    virtual void execute() = 0;
+
+    void dequeue()
     {
       ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
-      executed_ = false;
+      on_queue_ = false;
+    }
+
+    void executed()
+    {
+      ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
+      executed_ = true;
+      condition_.notify_all();
     }
 
     void wait() const
@@ -50,13 +64,6 @@ public:
       while (!executed_) {
         condition_.wait();
       }
-    }
-
-    void executed()
-    {
-      ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
-      executed_ = true;
-      condition_.notify_all();
     }
 
   protected:
@@ -68,6 +75,7 @@ public:
     void set_reactor(ACE_Reactor* reactor) { reactor_ = reactor; }
 
     bool executed_;
+    bool on_queue_;
     mutable ACE_Thread_Mutex mutex_;
     mutable ConditionVariable<ACE_Thread_Mutex> condition_;
     ACE_Reactor* reactor_;
@@ -89,7 +97,7 @@ public:
 
   bool should_execute_immediately();
 
-  CommandPtr execute_or_enqueue(Command* c)
+  CommandPtr execute_or_enqueue(CommandPtr c)
   {
     OPENDDS_ASSERT(c);
     const bool immediate = should_execute_immediately();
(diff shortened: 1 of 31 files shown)


dds/DCPS/ReactorInterceptor.h
```diff
@@ -31,25 +31,24 @@ class OpenDDS_Dcps_Export ReactorInterceptor : public RcEventHandler {
...
-    virtual void execute() = 0;
-    virtual void queue_flushed() {}
 
-    void reset()
+    bool reset()
     {
-      ACE_GUARD(ACE_Thread_Mutex, guard, mutex_);
+      ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, mutex_, false);
       executed_ = false;
+      bool retval = on_queue_;
```
mitza-oci, line 41:
```suggestion
      const bool retval = on_queue_;
```

dds/DCPS/SporadicTask.h
```diff
@@ -20,27 +20,32 @@ class SporadicTask : public RcEventHandler {
 public:
   explicit SporadicTask(RcHandle<ReactorInterceptor> interceptor)
     : interceptor_(interceptor)
-    , scheduled_(false)
-    , last_command_(SPORADIC_TASK_NONE)
+    , desired_scheduled_(false)
+    , actual_scheduled_(false)
   {
+    sporadic_command_ = make_rch<SporadicCommand>(this);
```
mitza-oci, line 26: move to initializer list?

dds/DCPS/SporadicTask.h
```diff
@@ -78,113 +83,72 @@ class SporadicTask : public RcEventHandler {
...
     }
     execute(now);
     return 0;
   }
 
-  void schedule_i(const TimeDuration& delay)
+  void execute_i()
   {
+    // Bump up our refcount so the Reactor doesn't delete us.
+    RcHandle<SporadicTask> x = rchandle_from(this);
```
mitza-oci, line 123: Isn't deriving from `RcEventHandler` (which this class does) supposed to avoid this problem?  In order to run code in this method something else should already have a reference.  

Is this only for the case where a `cancel_timer` removes it from the reactor and that was the last reference?  If so, it would be nice for the reactor to take care of this, but maybe not...

dds/DCPS/SporadicTask.h
```diff
@@ -78,113 +83,72 @@ class SporadicTask : public RcEventHandler {
...
-    }
-
-    SporadicTask* const sporadic_task_;
-    const TimeDuration delay_;
-  };
-
-  struct CancelCommand : public ReactorInterceptor::Command {
-    CancelCommand(SporadicTask* hb)
+  struct SporadicCommand : public ReactorInterceptor::Command {
+    SporadicCommand(SporadicTask* hb)
```
mitza-oci, line 87:
```suggestion
    explicit SporadicCommand(SporadicTask* hb)
```

mitza-oci (changes requested):

dds/DCPS/SporadicTask.h
mitza-oci: So there's a stack like this: `run_reactor_event_loop` -> `our_handler::handle_timeout` -> `cancel_timer`.  Wouldn't (or shouldn't) the `run_reactor_event_loop` level have its own reference?

jrw972: Yes.  The problematic stack is `run_reactor_event_loop` -> `not_our_handler::handle_timeout` -> `cancel_timer`.  Since it is running in our reactor, it executes immediately which may drop the reference count.

mitza-oci (approved):

PR #3150, opened 2021-11-09
GitHub Actions: use a newer vcpkg

mitza-oci:
vcpkg installs Qt5 and its dependencies which wireshark needs
older vcpkg uses a bad download URL for PCRE

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4995,7 +4995,7 @@ jobs:
       if: steps.cache-artifact.outputs.cache-hit != 'true'
       uses: lukka/run-vcpkg@v7
       with:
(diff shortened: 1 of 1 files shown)


ClaytonCalabrese:
Do we want to change this for all of the vcpkgGitCommitId or just this one?

mitza-oci:
> 
> 
> Do we want to change this for all of the vcpkgGitCommitId or just this one?

If it works for wireshark we could try changing it for the others.

ClaytonCalabrese (approved):

jwillemsen:
When this works, can you also make a PR to upgrade vcpkg for ACE/TAO/MPC? That way we build everything with the same version

mitza-oci:
pcre2 is still broken in vcpkg, waiting for https://github.com/microsoft/vcpkg/pull/21248 to be merged

PR #3159, opened 2021-11-12
Add Return Code to `ServiceParticipant::shutdown()`

iguessthislldo:
`ServiceParticipant::shutdown` now returns an error code, such as if there are still participants.

Also if being destroyed without an explicit shutdown, `ServiceParticipant` will also attempt to delete any remaining participants before attempting to call `shutdown`.


mitza-oci (approved):
Based on the CI errors in BuiltInTopicTest, does the InfoRepo code need to be updated for this?

mitza-oci (approved):

dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -1052,10 +1075,13 @@ DomainParticipantImpl::delete_contained_entities()
   if (disc)
     disc->fini_bit(this);
 
+  /*
```
mitza-oci, line 1078: remove or add reason why it's commented-out

mitza-oci (changes requested):

dds/DCPS/DomainParticipantImpl.cpp
iguessthislldo: It was a test that I never cleaned up, but I'm uncommenting it out because I think it should probably go back to how it was for now.

mitza-oci (approved):

PR #3170, opened 2021-11-17
Made the allocators in TransportReceiveStrategy configurable

mitza-oci:


iguessthislldo (approved):
Besides the warnings, it looks alright to me.

mitza-oci:
> 
> 
> Besides the warnings, it looks alright to me.

Those are odd.  I didn't realize `-Wextra` was in effect.
I guess the enums should be changed to normal constants.

jrw972 (approved):

PR #3180, opened 2021-11-23
Fix V8 generator invalid write issue for wide strings

simpsont-oci:
Problem: V8 conversion code for wide strings allocate temporary buffers that don't include enough space for the terminating null character, causing writes into invalid memory.

Solution: Increase temporary buffer allocation by 1.


jwillemsen:
Something to check in a test?

simpsont-oci:
node-opendds has the tests for the v8 generator, it will get some updates shortly

mitza-oci (approved):

PR #3182, opened 2021-11-25
Update to mention x.0.5

jwillemsen:


iguessthislldo:
These are phrased as minimum versions, so don't think we should automatically update them when a new version comes out unless the new version is required.

mitza-oci:
> These are phrased as minimum versions, so don't think we should automatically update them when a new version comes out unless the new version is required.

It's a little more complicated than that.  And maybe this needs re-wording.
The original idea was to document, at the time of an OpenDDS release, which ACE_TAO versions it was tested with.

mitza-oci:
Incorporated in https://github.com/objectcomputing/OpenDDS/pull/3217

PR #3189, opened 2021-11-30
XTypes Dynamic: Fix Warnings, Cleanup Code, Improve `dynamic-data-monitor`

iguessthislldo:
- Fix warnings from `use_xtypes_` and `ANY_EXTENSIBILITY`.
- Add namespaces to `ValueHelper.h`.
- Make logging in `RecorderImpl.cpp` align to guidelines, most significantly three unconditional debug logging points.
- Check for invalid arguments in the DynamicTypes test publisher.
- Several improvements to `dynamic-data-monitor`:
  - Cleaned up unused code.
  - Setup RTPS programmatically, so an ini file isn't always required.
  - Put `dynamic-data-monitor` in `$DDS_ROOT/bin`.
  - Improved argument parsing:
    - Replaced `ACE_Arg_Shifter` with a more systematic approach with stricter checks loosely based on argument parsing in the bench 2 programs.
    - Added a `--help` message.
    - Added a `--writer-count` option to display the number of writers when they associate and disassociate.

### DIFF
--- a/dds/DCPS/RecorderImpl.cpp
+++ b/dds/DCPS/RecorderImpl.cpp
@@ -233,8 +233,7 @@ void RecorderImpl::data_received(const ReceivedDataSample& sample)
       Encoding enc;
       Serializer ser(sample.sample_.get(), enc);
       EncapsulationHeader encap;
-      if (ser >> encap) {
-        encap.to_encoding(enc, EXTENSIBILITY_ANY);
+      if (ser >> encap && encap.to_any_encoding(enc)) {
         kind = enc.kind();
       }
     }
--- a/dds/DCPS/RecorderImpl.cpp
+++ b/dds/DCPS/RecorderImpl.cpp
@@ -1,11 +1,9 @@
 /*
- *
- *
  * Distributed under the OpenDDS License.
  * See: http://www.opendds.org/license.html
  */
 
-#include "DCPS/DdsDcps_pch.h" //Only the _pch include should start with DCPS/
+#include <DCPS/DdsDcps_pch.h> // Only the _pch include should start with DCPS/
 
 #include "RecorderImpl.h"
 
@@ -114,11 +112,11 @@ RecorderImpl::cleanup()
   if (!disco->remove_subscription(this->domain_id_,
                                   participant_servant_->get_id(),
                                   this->subscription_id_)) {
-    ACE_ERROR_RETURN((LM_ERROR,
-                      ACE_TEXT("(%P|%t) ERROR: ")
-                      ACE_TEXT("RecorderImpl::cleanup: ")
-                      ACE_TEXT(" could not remove subscription from discovery.\n")),
-                     DDS::RETCODE_ERROR);
+    if (log_level >= LogLevel::Error) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: RecorderImpl::cleanup: "
+        "could not remove subscription from discovery\n"));
+    }
+    return DDS::RETCODE_ERROR;
   }
 
   // Call remove association before unregistering the datareader from the transport,
@@ -154,11 +152,8 @@ void RecorderImpl::init(
   DomainParticipantImpl*     participant,
   DDS::SubscriberQos         subqos)
 {
-  //
-  if (DCPS_debug_level >= 1) {
-
-    ACE_DEBUG((LM_DEBUG,
-               ACE_TEXT("(%P|%t) RecorderImpl::init\n")));
+  if (DCPS_debug_level >= 2) {
+    ACE_DEBUG((LM_DEBUG, "(%P|%t) RecorderImpl::init\n"));
   }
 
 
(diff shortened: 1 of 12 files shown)


dds/DCPS/StaticDiscovery.cpp
```diff
@@ -87,7 +87,6 @@ StaticEndpointManager::StaticEndpointManager(const RepoId& participant_id,
 #endif
   , max_type_lookup_service_reply_period_(0)
   , type_lookup_service_sequence_number_(0)
-  , use_xtypes_(true)
   , use_xtypes_complete_(false)
```
mitza-oci, line 90:
```suggestion
```

dds/DCPS/StaticDiscovery.h
```diff
@@ -513,7 +513,6 @@ class StaticEndpointManager
   RcHandle<StaticEndpointManagerSporadic> type_lookup_reply_deadline_processor_;
   TimeDuration max_type_lookup_service_reply_period_;
   SequenceNumber type_lookup_service_sequence_number_;
-  const bool use_xtypes_;
   bool use_xtypes_complete_;
```
mitza-oci, line 516:
```suggestion
```

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
```diff
@@ -1,6 +1,7 @@
-project(Dynamic_Data_Monitor): dcps_rtps_udp {
-  requires += no_opendds_safety_profile
+project(Dynamic_Data_Monitor): dcps_rtps_udp, install {
   exename = dynamic-data-monitor
+  exeout = $(DDS_ROOT)/bin
```
mitza-oci, line 3:
```suggestion
  exename = dynamic_data_monitor
  exeout = $(DDS_ROOT)/bin
```
Nothing else uses dashes

mitza-oci (changes requested):

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
iguessthislldo: This was [deliberate change I had Clayton make](https://github.com/objectcomputing/OpenDDS/pull/3125#discussion_r748770003).

I don't think we should use dashes (or capital letters) in new command names because it's unusual. For example on my system: `ls -1 /usr/bin | grep '_' | wc -l` gives 146, with the large majority of them either having `x86_64` in their name or are `dh_*` commands used in building debian packages. `ls -1 /usr/bin | grep '-' | wc -l` gives 702.

Also assuming standard PC-based keyboard, one less button to press when typing the name.

dds/DCPS/StaticDiscovery.cpp
iguessthislldo: Unlike `use_xtypes_`, this is tied to the rest of the leftover XTypes stuff in `StaticDiscovery`, should I pull it all out?

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
mitza-oci: It's not a change I'd like us to take on right now.

dds/DCPS/StaticDiscovery.cpp
mitza-oci: Yes, unless that proves to be difficult

tools/dynamicdatamonitor/DynamicDataMonitor.mpc
iguessthislldo: I don't think there's any negative with it when the existing commands are not consistent (underscores vs camel case vs nothing). Also unlike most of the other commands this is a command that users are more like to be directly invoked on the command like so I think it's even more critical to get the name right when we're introducing it. `opendds_idl` and `dcpinfo_dump` are the only commands that have underscores. `opendds_idl` isn't normally invoked manually and `dcpsinfo_dump` is a minor utility as far as I'm aware.

mitza-oci: I think we're looking at different data here.  As far as I can tell, executables that are installed in bin and use a separator use underscore.

dds/DCPS/StaticDiscovery.cpp
ClaytonCalabrese: Yeah, all that stuff was from me not being sure whether we were going to try and do anything with static discovery and XTypes. We decided we weren't so pull it out if possible.  There's some weirdness with functions needing to be in both discoveries so it might not be the easiest thing to do.

tools/dynamicdatamonitor/DynamicDataMonitor.cpp
```diff
@@ -69,7 +220,7 @@ class TestRecorderListener : public RecorderListener {
     String my_type;
     String indent;
     if (!OpenDDS::XTypes::print_dynamic_data(dd, my_type, indent)) {
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: TestRecorderListener::on_sample_data_received: "
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: RecorderListenerImpl::on_sample_data_received: "
         "Failed to read dynamic data\n"));
```
ClaytonCalabrese, line 224:
```suggestion
      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: RecorderListenerImpl::on_sample_data_received: "
                 "Failed to read dynamic data\n"));
```

ClaytonCalabrese (changes requested):

dds/DCPS/RecorderImpl.cpp
```diff
@@ -1006,9 +998,10 @@ RecorderImpl::enable()
                               type_info);
 
     if (this->subscription_id_ == OpenDDS::DCPS::GUID_UNKNOWN) {
-      ACE_ERROR((LM_ERROR,
-                 ACE_TEXT("(%P|%t) ERROR: RecorderImpl::enable, ")
-                 ACE_TEXT("add_subscription returned invalid id.\n")));
+      if (log_level >= LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: RecorderImpl::enable: "
+          "add_subscription returned invalid id\n"));
+      }
```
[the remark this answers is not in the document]
mitza-oci: Either one is fine according to dev_guidelines.rst (see "Continuation lines")

mitza-oci (approved):

PR #3190, opened 2021-11-30
RtpsRelay statistics contain total count for publishers

jrw972:
Problem
-------

The relay statistics for the RtpsRelay report the total publisher
count for various topics.  This is not useful for measuring cluster
connectivity.

Solution
--------

Report the current count.


iguessthislldo (approved):
Yes, I realized that they were a bit more complicated recently when looking at the status use for #3189.

PR #3191, opened 2021-11-30
Update use of receive address in RtpsUdpDataLink::accumulate_addresses

mitza-oci:
The receive address may not be a valid locator since datagrams can be
sent from different addresses (other than those in the locator list).

This change makes it so that the receive address, without port number,
is used to find a matching locator.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -4598,13 +4598,15 @@ bool RtpsUdpDataLink::RemoteInfo::insert_recv_addr(AddrSet& aset) const
   if (last_recv_addr_ == ACE_INET_Addr()) {
     return false;
   }
-  for (AddrSet::iterator it = unicast_addrs_.begin(); it != unicast_addrs_.end(); ++it) {
-    if (it->get_type() == last_recv_addr_.get_type() &&
-        it->get_addr_size() == last_recv_addr_.get_addr_size() &&
-        std::memcmp(it->get_addr(), last_recv_addr_.get_addr(), it->get_addr_size()) == 0) {
-      aset.insert(last_recv_addr_);
-      return true;
-    }
+  ACE_INET_Addr recv_no_port(last_recv_addr_);
+  recv_no_port.set_port_number(0);
+  const AddrSet::iterator it = unicast_addrs_.lower_bound(recv_no_port);
+  if (it != unicast_addrs_.end() &&
+      it->get_type() == last_recv_addr_.get_type() &&
+      it->get_addr_size() == last_recv_addr_.get_addr_size() &&
+      std::memcmp(it->get_addr(), last_recv_addr_.get_addr(), it->get_addr_size()) == 0) {
+    aset.insert(last_recv_addr_);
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4593,11 +4593,26 @@ RtpsUdpDataLink::get_addresses_i(const RepoId& local) const {
   return retval;
 }
 
+bool RtpsUdpDataLink::RemoteInfo::insert_recv_addr(AddrSet& aset) const
+{
+  if (last_recv_addr_ == ACE_INET_Addr()) {
+    return false;
+  }
+  for (AddrSet::iterator it = unicast_addrs_.begin(); it != unicast_addrs_.end(); ++it) {
```
simpsont-oci, line 4601: Could we use lower_bound (maybe with port = 0?) to avoid the linear search?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4593,11 +4593,26 @@ RtpsUdpDataLink::get_addresses_i(const RepoId& local) const {
...
+    if (it->get_type() == last_recv_addr_.get_type() &&
```
simpsont-oci, line 4602: Thinking about this a bit more, this probably also needs to check the relay address as well as what's in the locators.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4624,8 +4639,7 @@ RtpsUdpDataLink::accumulate_addresses(const RepoId& local, const RepoId& remote,
 
   const RemoteInfoMap::const_iterator pos = locators_.find(remote);
   if (pos != locators_.end()) {
-    if (prefer_unicast && pos->second.last_recv_addr_ != NO_ADDR) {
-      normal_addrs.insert(pos->second.last_recv_addr_);
+    if (prefer_unicast && pos->second.insert_recv_addr(normal_addrs)) {
```
simpsont-oci, line 4642: I think this is probably OK, but it's worth noting that this is a change from the previous behavior, where the normal_addrs_expires and valid_last_recv_addr were always updated, regardless of the source address.

simpsont-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
mitza-oci: It made sense to me to make it conditional, but if it needs to change I'm OK with that too.  The NO_ADDR case still needs the conditional.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: Nevermind, you're right... last_recv_addr_ shouldn't ever be set to the relay address based on the logic at the top of update_last_recv_addr, so no need to check for it.

simpsont-oci (approved):

PR #3197, opened 2021-12-02
thread utilization BIT data

pmesnier:
Add thread utilization percentage as a part of the builtin topic data for internal threads

### DIFF
--- a/tools/rtpsrelay/RelayThreadMonitor.cpp
+++ b/tools/rtpsrelay/RelayThreadMonitor.cpp
@@ -2,6 +2,7 @@
 
 #include <dds/DCPS/ReactorTask.h>
 #include <ace/Thread.h>
+#include <ace/OS_NS_strings.h>
 
 #include <stdexcept>
 
@@ -15,6 +16,10 @@ RelayThreadMonitor::RelayThreadMonitor(TimeDuration perd, size_t depth)
   ThreadMonitor::installed_monitor_ = this;
 }
 
+RelayThreadMonitor::~RelayThreadMonitor() noexcept
+{
+  for (auto r : reporters_) delete r;
+}
 void RelayThreadMonitor::preset(ThreadStatusManager *tsm, const char *alias)
 {
   pending_reg_[alias] = tsm;
@@ -24,7 +29,7 @@ void RelayThreadMonitor::update(ThreadMonitor::UpdateMode mode,
                                 const char* alias)
 {
   MonotonicTimePoint tnow = MonotonicTimePoint::now();
-  struct Sample s = { mode, tnow };
+  Sample s = { mode, tnow };
   ACE_thread_t key = ACE_OS::thr_self();
   try {
     ThreadDescriptor_rch& td = descs_.at(key);
@@ -41,7 +46,7 @@ void RelayThreadMonitor::update(ThreadMonitor::UpdateMode mode,
   }
 }
 
-double RelayThreadMonitor::get_busy_pct(const char* key) const
+double RelayThreadMonitor::get_utilization(const char* key) const
 {
   try {
     return busy_map_.at(key);
@@ -50,8 +55,7 @@ double RelayThreadMonitor::get_busy_pct(const char* key) const
   }
 }
 
-RelayThreadMonitor::ThreadDescriptor::ThreadDescriptor(const char* alias,
-                                                     MonotonicTimePoint tnow)
+ThreadDescriptor::ThreadDescriptor(const char* alias, MonotonicTimePoint tnow)
 : alias_(alias)
 , tsm_(nullptr)
 , last_(tnow)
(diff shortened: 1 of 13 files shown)


dds/DdsDcpsCore.idl
```diff
@@ -617,6 +617,8 @@ module OpenDDS {
     struct InternalThreadBuiltinTopicData {
       BUILT_IN_TOPIC_KEY string thread_id;
       DDS::OctetArray16 participant_guid;
+      double utilization;
+      DDS::Time_t timestamp;
```
iguessthislldo, line 621:
```suggestion
```

There actually already was a timestamp field in this struct before, but we decided to remove it because `SampleInfo::source_timestamp` is a way DDS already provides to get timestamp of a sample. https://github.com/objectcomputing/OpenDDS/pull/2247/commits/ac92e2dd1586b6babc2158bcfe9fea9b1a835631#diff-f32cd34225dfd7aaabc590e1494701202f10cd801600b0f37c9da8ff16113359

If you look at the `store_synthetic_data`, the timestamp is getting passed in, so it's the same value.

tools/rtpsrelay/RelayThreadMonitor.h
```diff
@@ -88,13 +42,60 @@ class RelayThreadMonitor : public OpenDDS::DCPS::ThreadMonitor {
...
+    typedef ACE_SYNCH_MUTEX LockType;
+    typedef ACE_Guard<LockType> GuardType;
+    typedef OpenDDS::DCPS::ConditionVariable<LockType> ConditionVariableType;
+
+    LockType lock_;
+    ConditionVariableType condition_;
+    OpenDDS::DCPS::ThreadMonitor& owner_;
+  };
+
+  /**
```
iguessthislldo, line 73:
```suggestion
/**
```

tools/rtpsrelay/RelayThreadMonitor.h
```diff
@@ -79,6 +32,7 @@ class RelayThreadMonitor : public OpenDDS::DCPS::ThreadMonitor {
   typedef std::deque<struct Sample> LoadSamples;
   typedef std::deque<struct LoadSummary> LoadHistory;
   typedef std::map<const char*, double> BusyMap;
+  typedef std::map<std::string, OpenDDS::DCPS::ThreadStatusManager *> registrants;
```
iguessthislldo, line 35:
```suggestion
  typedef std::map<std::string, OpenDDS::DCPS::ThreadStatusManager*> registrants;
```

iguessthislldo (changes requested):

dds/DdsDcpsCore.idl
pmesnier: Yes, I am aware of the timestamp you reference,  But isn't that the time the message was created/sent" My timestamp is the time the utilization value was measured.  it seems like these BIT updates happen at their own frequency which may be somewhat lower than the thread utilization freequency. So it may be the case that I need to extend this struct further to include a sequence fo double/time_t pairs. OTOH if  if a BIT sample is emitted as a result of any thread state update meaning there will never be more than one data point possible then I suppose this timestamp is redundant

iguessthislldo: As part of the changes I linked to above I changed `store_synthetic_data` to optionally take the time stamp to use instead of doing the time at the moment like it normally does. For the thread BIT, it's the same value as what's in the `Thread::timestamp`, which for the thread status update is when the thread is declaring it's alive.

https://github.com/objectcomputing/OpenDDS/blob/fb0d13e99e0e148fde34473db22d7b8d545635c4/dds/DCPS/RTPS/Spdp.cpp#L4461-L4464

As for what the requirements of this are, yeah they might warrant something else than what was needed for thread status updates. The thread status was designed just to deliver the last update time. I don't know what it would look like or if it would actually be ideal, but you could aggregate multiple utilization into one. You should also be able to adjust the `thread_status_interval` since if need be.

pmesnier: got it. Thanks for the explanation.

tools/rtpsrelay/RelayThreadMonitor.cpp
```diff
@@ -112,51 +118,104 @@ void RelayThreadMonitor::summarize()
       *bucket += udiff;
       td->last_ = tnow;
     }
+
+    double pbusy = 0.0;
+    double uspan = ls.accum_idle_ + ls.accum_busy_;
```
iguessthislldo, line 123: Visual studio doesn't like the implicit cast from  here.

tools/rtpsrelay/RelayThreadMonitor.cpp
```diff
@@ -112,51 +118,104 @@ void RelayThreadMonitor::summarize()
...
+
+void AceLogReporter::report_thread(ThreadDescriptor_rch td) const
+{
+  if (td->summaries_.empty()) {
+    ACE_DEBUG((LM_INFO, "     TLM thread \"%s\" busy:  n/a    idle: n/a", td->alias_.c_str()));
+    return;
+  }
+  const LoadSummary& ls = td->summaries_.back();
+  double pbusy = 0.0;
+  double uspan = ls.accum_idle_ + ls.accum_busy_;
```
iguessthislldo, line 172: Same warning here.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3206, opened 2021-12-06
Unable to compile due to variable name conflicts with some Windows API headers

bluenlive:
In Visual Studio (test with 2019, 2022) the names of the variables are confused and cannot be compiled. (`interface`, `small`)


mitza-oci:
To avoid accidental breaks in the future, add a test case that conditional (`_MSC_VER`) includes the problematic headers.

bluenlive:
> To avoid accidental breaks in the future, add a test case that conditionally (`_MSC_VER`) includes the problematic headers.

Okay. Thank you.

jrw972 (approved):

PR #3210, opened 2021-12-06
`lint.pl` Fixes

iguessthislldo:
- Replaced all whitespace with a dot (`·`) in marked error output when color is enabled. This dot is used in other software to explicitly mark whitespace. This should make it less confusing if output hides this whitespace.
- Fixed ability to check files outside an OpenDDS repo. Remove unneeded `--ext-path` option.
- Check if directories are ignored by git to avoid unnecessary checks on the contents.
- Removed possibly confusing usage of both "color" and "colour".
- Rename `--files-only` to `--files-only-ouput` to be consistent with `--simple-output`. `--files-only` should still work though.
- Spelling and other minor fixes.

### DIFF
--- a/tools/scripts/lint.pl
+++ b/tools/scripts/lint.pl
@@ -11,7 +11,7 @@ use File::Find qw/find/;
 use Encode qw/decode encode FB_CROAK/;
 use Cwd qw/abs_path/;
 use Getopt::Long qw/GetOptions/;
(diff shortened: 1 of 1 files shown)


iguessthislldo:
Looks like it's okay on GitHub:

Screenshot from 2021-12-06 17-46-14

simpsont-oci (approved):

PR #3211, opened 2021-12-07
XTypes dynamic: move headers and to_string for safety profile

mitza-oci:


dds/DCPS/SafetyProfileStreams.cpp
```diff
@@ -96,6 +96,24 @@ String to_dds_string(unsigned long to_convert, bool as_hex)
...
+  for (size_t i = 0; i < length; ++i) {
+    const int used = ACE_OS::snprintf(buf, bytes_per_elt + 1,
+                                      i < length - 1 ? "%d, " : "%d", array[i]);
+    if (used < 1) {
+      return "";
+    }
+    buf += used;
+  }
+
+  ret.resize(ret.find('\0'));
```
jrw972, line 113: `ret` can be resized to the loop-sum of `used`.

jrw972 (changes requested):

PR #3214, opened 2021-12-08
Performance Improvements For AddressCache

simpsont-oci:
Improvements based on profiling from centipede test environment. The bundling cache was having pretty significant slowdown copying keys when keys had large numbers of "to" guids.

Primary improvements are:
 - AddressCache id_map_ (reverse map) now holds a vector of RcHandles to keys, keeping key-copying to a minimum when adding keys to the reverse map
 - Precalculate hash value for bundling keys to avoid repeated lengthy calls to the hash function

### DIFF
--- a/dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
+++ b/dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
@@ -64,7 +64,7 @@ struct OpenDDS_Rtps_Udp_Export BundlingCacheKey : public RcObject {
 
   bool operator<(const BundlingCacheKey& rhs) const
   {
-    int r = std::memcmp(&dst_guid_, &(rhs.dst_guid_), 2 * sizeof (GUID_t));
+    int r = std::memcmp(&dst_guid_, &rhs.dst_guid_, 2 * sizeof (GUID_t));
     if (r < 0) {
       return true;
     } else if (r == 0) {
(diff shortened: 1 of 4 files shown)


dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
```diff
@@ -26,25 +26,45 @@ namespace DCPS {
...
+#if defined ACE_HAS_CPP11
+    , hash_(val.hash_)
+#endif
+  {
   }
 
   bool operator<(const BundlingCacheKey& rhs) const
   {
-    int r = std::memcmp(this, &rhs, 2 * sizeof (GUID_t));
+    int r = std::memcmp(&dst_guid_, &(rhs.dst_guid_), 2 * sizeof (GUID_t));
```
mitza-oci, line 67:
```suggestion
    int r = std::memcmp(&dst_guid_, &rhs.dst_guid_, 2 * sizeof (GUID_t));
```

mitza-oci (changes requested):
Important: 
warning on line 47 in dds/DCPS/transport/rtps_udp/LocatorCacheKey.h

Not so important:
Don't add extra `()` around expressions that aren't changing precedence or preventing confusion when reading (perhaps subjective)
Seemingly redundant namespace qualification - not sure if required for disambiguation

mitza-oci (approved):

PR #3217, opened 2021-12-08
NEWS and README updates

mitza-oci:


NEWS.md
```diff
@@ -4,13 +4,33 @@
 OpenDDS 3.19.0 is currently in development, so this list might change.
 
 ### Additions:
-- TODO: Add your features here
+- XTypes: Dynamic Language Binding (#2976)
```
iguessthislldo, line 7: I fell like since this is our headliner we should give some more information. Something like:

```suggestion
- XTypes: Dynamic Language Binding (#2976)
  - Read support only for the time being.
  - New `inspect` tool that prints data being written to a topic in text form.
  - See section 16.7 of the Developer's guide for details.
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3223, opened 2021-12-10
Attempt To Resolve AddressCache Key Coverity Issues

simpsont-oci:


iguessthislldo:
Is there a way to do this without making assumptions about the memory layout?

simpsont-oci:
No, which is why they're all using pragma pack of 1.

mitza-oci (approved):

PR #3233, opened 2021-12-14
Data Reader redesign for Encapsulation and Encoding

ClaytonCalabrese:
This PR refactors how readers handle encapsulation.  Readers should depend on the writer telling them whether CDR encapsulation is used.  Writers and Reader continue to default to unaligned for non rtps_udp transports, but it is done differently than before.  Unaligned has been added as an OpenDDS custom representation qos value and can now be set manually.


dds/idl/annotations.h
```diff
@@ -345,6 +345,7 @@ namespace OpenDDS {
     bool xcdr1;
     bool xcdr2;
     bool xml;
+    bool unaligned;
```
iguessthislldo, line 348: `UNALIGNED_CDR` also needs to be added to `@OpenDDS::data_representation` or else this doesn't do much: https://github.com/objectcomputing/OpenDDS/blob/fe59c08bc4316281a4b7805d679327435ff7ac8a/dds/idl/annotations.cpp#L374

https://github.com/objectcomputing/OpenDDS/blob/fe59c08bc4316281a4b7805d679327435ff7ac8a/dds/idl/annotations.cpp#L419-L421

tests/DCPS/XTypes/Subscriber.cpp
```diff
@@ -357,81 +362,91 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+    if (!failed) {
+      failed = !check_inconsistent_topic_status(topic);
+    }
 
-  if (!failed) {
-    failed = !check_inconsistent_topic_status(topic);
-  }
+    if (!failed) {
+      failed = !check_inconsistent_topic_status(topic);
+    }
```
iguessthislldo, line 397: Is this supposed to be done twice?

iguessthislldo (changes requested):

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -464,21 +477,35 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
+}
+
+DDS::DataRepresentationIdSeq get_reader_effective_data_rep_qos(const DDS::DataRepresentationIdSeq& qos)
+{
+  if (qos.length() == 0) {
+    DDS::DataRepresentationIdSeq ids;
+    ids.length(3);
+    ids[0] = DDS::XCDR_DATA_REPRESENTATION;
+    ids[1] = DDS::XCDR2_DATA_REPRESENTATION;
+    ids[2] =  DDS::UNALIGNED_CDR_DATA_REPRESENTATION;
```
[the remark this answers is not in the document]
ClaytonCalabrese: Doesn't it need to be in the DDS namespace to be defined alongside the others or are you saying to define it seperately?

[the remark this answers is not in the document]
mitza-oci: It doesn't need to be in any specific namespace, these are just integer constants (not an enum).  It should not be in the DDS namespace because it's not in the spec.

iguessthislldo (approved):

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -321,12 +321,12 @@ DataWriterImpl_T
...
+  }
 
   DDS::ReturnCode_t setup_serialization()
   {
     const DDS::DataRepresentationIdSeq repIds =
-      get_effective_data_rep_qos(qos_.representation.value, false);
+      get_writer_effective_data_rep_qos(qos_.representation.value, cdr_encapsulation());
     if (qos_.representation.value.length() > 0) {
       // If the QoS explicitly sets XCDR, XCDR2, or XML, force encapsulation
       cdr_encapsulation(true);
```
mitza-oci, line 332: This looks like it needs to be updated to account for the new UNALIGNED_CDR_DATA_REPRESENTATION

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -340,12 +340,19 @@ DataWriterImpl_T
...
+              "%CDataWriterImpl::setup_serialization: "
+              "Encountered unsupported combination of XCDR1 encoding and mutable extensibility\n",
               TraitsType::type_name()));
           }
           return DDS::RETCODE_ERROR;
+        } else if (encoding_kind == Encoding::KIND_UNALIGNED_CDR) {
+          if (::OpenDDS::DCPS::DCPS_debug_level) {
+            ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: "
+              "%CDataWriterImpl::setup_serialization: "
+              "Unaligned CDR is not allowed in rtps_udp transport\n",
```
mitza-oci, line 353:
```suggestion
              "Unaligned CDR is not supported by transport types that require encapsulation\n",
```

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1284,7 +1290,7 @@ bool to_param_list(const DCPS::DiscoveredReaderData& reader_data,
     add_param(param_list, param);
   }
 
-  add_DataRepresentationQos(param_list, reader_data.ddsSubscriptionData.representation.value, true);
+  add_DataRepresentationQos(param_list, reader_data.ddsSubscriptionData.representation.value, true, true);
```
mitza-oci, line 1293: How do we know the last parameter can always be `true`?
Another way to see the same issue - can the last parameter be removed?  It seems to only matter when the sequence is empty.  This function is called when transforming info about one of our local DataReaders to ParameterList format.  The caller could make sure that the sequence already accounts for the defaults -- before calling this we can know what we want to appear in the ParameterList in discovery, which may be an empty list (if we don't want anything to be encoded in the ParameterList) or not.

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -252,10 +252,21 @@ compatibleQOS(OpenDDS::DCPS::IncompatibleQosStatus* writerStatus,
     increment_incompatibility_count(readerStatus,
                                     OpenDDS::TRANSPORTTYPE_QOS_POLICY_ID);
   }
-
+  bool encapsulated_only = false;
+  bool exit_loop = false;
+  for (CORBA::ULong i = 0; i < pubTLS.length() && !exit_loop; ++i) {
+    for (CORBA::ULong j = 0; j < subTLS.length() && !exit_loop; ++j) {
+      if (0 == std::strcmp(pubTLS[i].transport_type, subTLS[j].transport_type)) {
+        if (0 == std::strcmp(pubTLS[i].transport_type, "rtps_udp")) {
```
mitza-oci, line 260: This should not have a dependency on the transport directly in the DCPS_Utils.  Since the `DataRepresentationIdSeq` for each side is known, is that not sufficient to determine compatibility?  At this point we're not dealing with the "implied defaults" of an empty sequence, we should know the actual contents of `DataRepresentationIdSeq`.

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -464,21 +477,35 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
+  return qos;
+}
+
+DDS::DataRepresentationIdSeq get_reader_effective_data_rep_qos(const DDS::DataRepresentationIdSeq& qos)
+{
+  if (qos.length() == 0) {
+    DDS::DataRepresentationIdSeq ids;
+    ids.length(3);
+    ids[0] = DDS::XCDR_DATA_REPRESENTATION;
+    ids[1] = DDS::XCDR2_DATA_REPRESENTATION;
```
mitza-oci, line 507: Prior to this change, XCDR2 came first in the sequence.  I'm not sure it matters but it seems like we should preserve that unless there's a reason not to.

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -7189,8 +7189,20 @@ void Sedp::match_continue(const GUID_t& writer, const GUID_t& reader)
       reader_type_info->minimal.typeid_with_size.type_id;
     if (writer_type_id.kind() != XTypes::TK_NONE && reader_type_id.kind() != XTypes::TK_NONE) {
       if (!writer_local || !reader_local) {
+        bool encapsulated_only = false;
+        bool exit_loop = false;
+        for (CORBA::ULong i = 0; i < wTls->length() && !exit_loop; ++i) {
+          for (CORBA::ULong j = 0; j < rTls->length() && !exit_loop; ++j) {
+            if (0 == std::strcmp((*wTls)[i].transport_type, (*rTls)[j].transport_type)) {
+              if (0 == std::strcmp((*wTls)[i].transport_type, "rtps_udp")) {
```
mitza-oci, line 7197: Similar to DCPS_Utils note above.

java/dds/DDS/.gitignore
```diff
@@ -548,6 +548,7 @@
 /XCDR2_DATA_REPRESENTATION.java
 /XCDR_DATA_REPRESENTATION.java
 /XML_DATA_REPRESENTATION.java
+/UNALIGNED_DATA_REPRESENTATION.java
```
mitza-oci, line 551: This needs to change for the new namespace

tests/DCPS/DataRepresentation/run_test.pl
```diff
@@ -10,6 +10,6 @@
 my $test = new PerlDDS::TestFramework();
 $test->enable_console_logging();
 $test->setup_discovery();
-$test->process('test', 'DataRepresentation');
+$test->process('test', 'DataRepresentation', '-DCPSConfigFile rtps_disc.ini');
```
mitza-oci, line 13: Does it need to be tested with multiple configurations?

mitza-oci (changes requested):

mitza-oci:
> Currently a WIP.

Please update the description.

Adding some more context which may help clarify my review comments:
One of the key points of this design is that the meaning of an empty `DataRepresentationIdSeq` is context-dependent (as it was before, but more so now).  Instead of needing to make the decision about what an empty sequence means at various points throughout the system (like `compatibleQOS`), it seems like it should be possible to "resolve" or "expand" an empty sequence at the four places it may occur: local DataWriter, local DataReader, discovered (remote) DataWriter, discovered (remote) DataReader.  Of course some of these are already done.

dds/DCPS/DataWriterImpl_T.h
ClaytonCalabrese: Users cannot explicitly set UNALIGNED_CDR_DATA_REPRESENTATION because it is not considered a valid QoS policy in `Qos_Helper::valid`

mitza-oci: That needs to be changed too

dds/idl/annotations.h
ClaytonCalabrese: The goal is to keep Unaligned as the interpreted default for transports besides rtps_udp.  This is to maintain compatibility with earlier OpenDDS versions.  I don't think we want the users explicitly setting unaligned in idl. The long term goal is removal of Unaligned, and we would need to consider a lot more cases if users could set it.

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -1472,6 +1463,18 @@ DataWriterImpl::enable()
     return DDS::RETCODE_ERROR;
   }
 
+  // Must be done after transport enabled.
+  set_writer_effective_data_rep_qos(qos_.representation.value, cdr_encapsulation());
+  if (!topic_servant_->check_data_representation(qos_.representation.value, true)) {
+    if (DCPS_debug_level) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: DataWriterImpl::enable: ")
```
jwillemsen, line 1470: Maybe log the values that are there?

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3416,34 +3416,37 @@ DataReaderImpl::get_ice_endpoint()
...
+        xcdr1_mutable = true;
+      } else if (encoding_kind == Encoding::KIND_UNALIGNED_CDR && cdr_encapsulation()) {
+        illegal_unaligned = true;
+      } else {
+        decoding_modes_.insert(encoding_kind);
       }
+    } else if (DCPS_debug_level) {
+      ACE_DEBUG((LM_WARNING, "(%P|%t) WARNING: "
+                 "DataReaderImpl::setup_deserialization: "
+                 "Encountered unsupported or unknown data representation: %u\n",
```
jwillemsen, line 3434: Maybe use a helper to print the value as readable string?

mitza-oci (approved):

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -321,38 +321,44 @@ DataWriterImpl_T
...
-  };
+  }
 
   DDS::ReturnCode_t setup_serialization()
   {
-    const DDS::DataRepresentationIdSeq repIds =
-      get_effective_data_rep_qos(qos_.representation.value, false);
-    if (qos_.representation.value.length() > 0) {
+    if (qos_.representation.value.length() > 0 &&
+        qos_.representation.value[0]!= OpenDDS::DCPS::UNALIGNED_CDR_DATA_REPRESENTATION) {
```
mitza-oci, line 329:
```suggestion
        qos_.representation.value[0] != OpenDDS::DCPS::UNALIGNED_CDR_DATA_REPRESENTATION) {
```

dds/DCPS/ReplayerImpl.h
```diff
@@ -18,6 +18,7 @@
 #include "CoherentChangeControl.h"
 #include "GuidUtils.h"
 #include "unique_ptr.h"
+#include "DCPS_Utils.h"
```
iguessthislldo, line 21: This should be in the cpp file since that's were it's being used.

dds/DCPS/RecorderImpl.h
```diff
@@ -17,6 +17,7 @@
 #include "EntityImpl.h"
 #include "TopicImpl.h"
 #include "OwnershipManager.h"
+#include "DCPS_Utils.h"
```
iguessthislldo, line 20: This should also be in the cpp file.

iguessthislldo (changes requested):

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -464,24 +463,73 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
+    repr_string = "XCDR2_DATA_REPRESENTATION";
+    break;
+  case OpenDDS::DCPS::UNALIGNED_CDR_DATA_REPRESENTATION:
+    repr_string = "UNALIGNED_CDR_DATA_REPRESENTATION";
+    break;
+#ifndef OPENDDS_SAFETY_PROFILE
+  default:
+    ss << repr;
+    repr_string = ss.str();
+#endif
```
mitza-oci, line 498: This doesn't need to be conditional.  Use the helper functions in `SafetyProfileStreams.h` to convert an int to a string portably.

dds/DCPS/DCPS_Utils.cpp
```diff
@@ -464,24 +463,73 @@ bool repr_to_encoding_kind(DDS::DataRepresentationId_t repr, Encoding::Kind& kin
...
+  }
+  return repr_string;
+}
+
+DCPS::String repr_seq_to_string(const DDS::DataRepresentationIdSeq& id_seq, bool is_data_writer)
+{
+  DCPS::String repr_string;
+  ACE_CDR::ULong length = is_data_writer ? 1 : id_seq.length();
```
mitza-oci, line 506:
```suggestion
  const ACE_CDR::ULong length = is_data_writer ? 1 : id_seq.length();
```

mitza-oci (changes requested):

java/dds/OpenDDS/.gitignore
```diff
@@ -2,3 +2,4 @@
 /MAX_USER_DOMAINID.java
 /TRANSPORTTYPE_QOS_POLICY_ID.java
 /TRANSPORTTYPE_QOS_POLICY_NAME.java
+/UNALIGNED_DATA_REPRESENTATION.java
```
mitza-oci, line 5: This doesn't match the IDL constant.  Not sure why GitHub Actions isn't flagging it as a problem.

iguessthislldo (approved):

mitza-oci (approved):

PR #3235, opened 2021-12-15
Move NetworkAddress out of unit-tests

ClaytonCalabrese:
The NetworkAddress test broke the rules we had defined to unit tests because it depended on the network.  As such, it needed to be moved out of unit-tests.


simpsont-oci:
Rather than moving the whole test, part of which is valid unit tests and should remain in unit tests, just pull out the tests which run DNS queries into a separate test file.

simpsont-oci (approved):

PR #3237, opened 2021-12-15
RtpsRelay topics are not partitioned

jrw972:
Problem
-------

When building infrastructure around the RtpsRelay, it is useful to
subscribe to individual relay instances.  One way is to use content
filtering, however, this is not supported on the publisher side for
RTPS and therefore will lead to many wasted messages.

Solution
--------

Partition the relay domain by the id of each instance.  Processes
interested in a specific relay instance can subscribe to that partition.

### DIFF
--- a/tools/rtpsrelay/RtpsRelay.cpp
+++ b/tools/rtpsrelay/RtpsRelay.cpp
@@ -441,7 +441,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
   DDS::SubscriberQos subscriber_qos;
   relay_participant->get_default_subscriber_qos(subscriber_qos);
   subscriber_qos.partition.name.length(1);
-  subscriber_qos.partition.name[0] = "*"; // Subscriber to all partitions.
+  subscriber_qos.partition.name[0] = "*"; // Subscribe to all partitions.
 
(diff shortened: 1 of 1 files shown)


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -425,15 +425,25 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
   if (!relay_publisher) {
     ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR: failed to create Relay publisher\n")));
     return EXIT_FAILURE;
   }
 
-  DDS::Subscriber_var relay_subscriber = relay_participant->create_subscriber(SUBSCRIBER_QOS_DEFAULT, nullptr,
+  DDS::SubscriberQos subscriber_qos;
+  relay_participant->get_default_subscriber_qos(subscriber_qos);
+  subscriber_qos.partition.name.length(1);
+  subscriber_qos.partition.name[0] = "*"; // Subscriber to all partitions.
```
mitza-oci, line 444:
```suggestion
  subscriber_qos.partition.name[0] = "*"; // Subscribe to all partitions.
```

sonndinh (approved):

mitza-oci (approved):

PR #3239, opened 2021-12-16
Improve XTypes test

sonndinh:
This makes the following changes to the test:
- Move the code for publisher and subscriber to a single directory so that it only needs to generate and run makefiles only once.
- Factor out some common IDL types of publisher and subscriber.
- Not sending an ack message and stop when the previous steps fail.


tests/DCPS/XTypes/XTypes.mpc
```diff
@@ -0,0 +1,31 @@
+project(Subscriber): dcps_test, dcps_rtps_udp, msvc_bigobj, dcps_transports_for_test, opendds_optional_security {
```
mitza-oci, line 1: In general, keeping each project in its own directory is the recommended "best practice."  It is not a strict requirement.

However in this case there are now two projects that can generate the same files (from Common.idl).  To do this correctly with parallel builds, add an `after` to one of the projects so that they can't run concurrently.

The project names must be changed since they need to be unique across the entire workspace.

mitza-oci (changes requested):

mitza-oci (approved):

tests/DCPS/XTypes/XTypes.mpc
```diff
@@ -0,0 +1,32 @@
+project(*Subscriber): dcps_test, dcps_rtps_udp, msvc_bigobj, dcps_transports_for_test, opendds_optional_security {
+
+  exename = subscriber
+
+  includes += .
+
```
iguessthislldo, line 6:
```suggestion
```

This and the other `includes += .` are not needed if everything's in the same directory.

iguessthislldo (changes requested):

iguessthislldo (approved):

ClaytonCalabrese (changes requested):
Something is causing failures in the mac runs of the CI.

sonndinh:
> Something is causing failures in the mac runs of the CI.

Looks like the `subscriber` process wasn't built for `test_m10_i0_j_FM-1f`. I don't know why. I pushed a minor change and see if it builds this time.

PR #3247, opened 2021-12-20
Improve Fragment NACKing Behavior

simpsont-oci:
Problem: Fragment NACKing is non-optimal for a number of reasons:
- Heartbeats are sent for the sample before the final fragment has been sent, causing pre-emptive NACK_FRAGs
- Only one NACK_FRAG range is stored per sequence number, causing a delay in resends when multiple NACK_FRAGs have been sent between heartbeats
- NACK_FRAGs are being generated for fully "received" samples (i.e. unneeded)

Solution: Don't do those things.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -3705,11 +3705,6 @@ RtpsUdpDataLink::RtpsWriter::gather_nack_replies_i(MetaSubmessageVec& meta_subme
   }
 
   {
-    for (FragmentInfo::const_iterator pos = consolidated_fragment_requests.begin(),
-           limit = consolidated_fragment_requests.end(); pos != limit; ++pos) {
-      consolidated_requests.erase(pos->first);
-    }
-
     // Send the consolidated requests.
     const OPENDDS_VECTOR(SequenceRange) ranges = consolidated_requests.present_sequence_ranges();
     for (OPENDDS_VECTOR(SequenceRange)::const_iterator pos = ranges.begin(), limit = ranges.end();
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -1103,7 +1103,7 @@ RtpsUdpDataLink::MultiSendBuffer::insert(SequenceNumber /*transport_seq*/,
     const RtpsCustomizedElement* const rce =
       dynamic_cast<const RtpsCustomizedElement*>(tqe);
     if (rce) {
-      send_buff->insert_fragment(seq, rce->last_fragment(), q, chain);
+      send_buff->insert_fragment(seq, rce->last_fragment(), rce->is_last_fragment(), q, chain);
     } else if (Transport_debug_level) {
       ACE_ERROR((LM_ERROR, "(%P|%t) RtpsUdpDataLink::MultiSendBuffer::insert()"
         " - ERROR: couldn't get fragment number for pub_id %C seq %q\n",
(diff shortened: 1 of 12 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -1047,19 +1050,46 @@ RtpsUdpReceiveStrategy::has_fragments(const SequenceRange& range,
...
-        RTPS::FragmentNumberSet& missing_frags = frag_info->back().second;
-        missing_frags.numBits = 0; // make sure this is a valid number before passing to get_gaps
-        missing_frags.bitmap.length(8); // start at max length
-        missing_frags.bitmapBase.value =
-          reassembly_.get_gaps(sn, pub_id, missing_frags.bitmap.get_buffer(),
-                               8, missing_frags.numBits);
-        // reduce length in case get_gaps() didn't need all that room
-        missing_frags.bitmap.length((missing_frags.numBits + 31) / 32);
+        if (total_frags > 256) {
+          const CORBA::Long empty_buffer[8] = { 0, 0, 0, 0, 0, 0, 0, 0 };
```
mitza-oci, line 1057:
```suggestion
          static const CORBA::Long empty_buffer[8] = { 0, 0, 0, 0, 0, 0, 0, 0 };
```

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -1047,19 +1050,46 @@ RtpsUdpReceiveStrategy::has_fragments(const SequenceRange& range,
...
+          OPENDDS_VECTOR(CORBA::Long) buffer(total_frags + 31/ 32, 0);
```
mitza-oci, line 1058:
```suggestion
          OPENDDS_VECTOR(CORBA::Long) buffer(total_frags + 31 / 32, 0);
```

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -3698,6 +3705,11 @@ RtpsUdpDataLink::RtpsWriter::gather_nack_replies_i(MetaSubmessageVec& meta_subme
   }
 
   {
+    for (FragmentInfo::const_iterator pos = consolidated_fragment_requests.begin(),
+           limit = consolidated_fragment_requests.end(); pos != limit; ++pos) {
+      consolidated_requests.erase(pos->first);
+    }
+
```
mitza-oci, line 3712: Could this be removed?  May not be needed if the NackFrags are only generated when the sequence number is not nacked.

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -346,12 +346,15 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink : public DataLink, public virtual
 
   // RTPS reliability support for local writers:
 
+  typedef OPENDDS_MAP(CORBA::ULong, RTPS::FragmentNumberSet) RequestedFragMap;
```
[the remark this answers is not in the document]
simpsont-oci: I don't like the idea of making the key a struct... we'd have to implement a custom comparison operation, since we can't define FragmentNumber_t::operator<(). How about if we add a typedef of ULong called FragmentNumberValue and make that the key type here? Should help with readability, but won't change the code that uses it?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
simpsont-oci: I'll give it a try and test it locally

mitza-oci (approved):

PR #3257, opened 2022-01-04
RtpsRelay admission control

pmesnier:
Dominant changes in this PR:
-Move ThreadStatusManager to its own source file.
-Add function to compute whether the designated  thread is saturated
-Add a field for the saturated flag to the InternalThreadStatus Built in topics.
-Expand the processing of thread usage measurements to include high/low watermarks to determine when the thread is saturated. 
-Add a condition to pause threads that are ready to send to the relay. This is done by adding a call prior to sending SPDP messages.
-Streamlined the configuration to avoid having redundant config options.


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -4457,6 +4458,7 @@ void Spdp::SpdpTransport::thread_status_task(const DCPS::MonotonicTimePoint& /*n
         assign(data.participant_guid, guid);
         data.thread_id = i->first.c_str();
         data.utilization = 0.0;
+        data.saturated = false;
```
jrw972, line 4461: Why is `saturated` necessary?  Can't the user determine their own limits?

dds/DCPS/Service_Participant.cpp
```diff
@@ -163,6 +164,9 @@ static bool got_log_verbose = false;
 static bool got_default_address = false;
 static bool got_bidir_giop = false;
 static bool got_thread_status_interval = false;
+static bool got_relay_thread_key = false;
+static bool got_relay_thread_hwm = false;
```
jrw972, line 168: This should not appear in OpenDDS proper.  These are solely for the benefit of the rtps relay so it should be handled entirely in the relay.

jrw972 (changes requested):

dds/DCPS/RTPS/Spdp.cpp
pmesnier: I added saturated to allow the relay to declare that its processing load is exceeding its load limit. Otherwise we are imposing on all the connecting apps to provide the configuration necessary to determine that the relay is busy and should therefore take a break before sending anything more messages.

dds/DCPS/Service_Participant.cpp
pmesnier: While I can refactor computation of thread load and saturation into relay-specific code, I think the relay_thread_key attribute needs to stay because that identifies the thread who's saturation status determines if message senders are held back for a while.

dds/DCPS/RTPS/Spdp.cpp
jrw972: The only thing that needs to be available for the relay to do its jobs in the utilization of the discovery thread for the application participant.  The thread utilization of the rtps relay handler is not a concern at this time.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -4423,56 +4415,44 @@ Spdp::SpdpTransport::process_lease_expirations(const DCPS::MonotonicTimePoint& n
...
 
-void Spdp::SpdpTransport::thread_status_task(const DCPS::MonotonicTimePoint& /*now*/)
+void Spdp::SpdpTransport::thread_status_task(const DCPS::MonotonicTimePoint& now)
 {
 #ifndef DDS_HAS_MINIMUM_BIT
   DCPS::RcHandle<Spdp> outer = outer_.lock();
   if (!outer) return;
 
-  if (DCPS::DCPS_debug_level > 4) {
+  if (true || DCPS::DCPS_debug_level > 4) {
```
mitza-oci, line 4424:
```suggestion
  if (DCPS::DCPS_debug_level > 4) {
```

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -1067,7 +1068,12 @@ WriteDataContainer::obtain_buffer(DataSampleElement*& element,
         }
 
         waiting_on_release_ = true;
-        switch (condition_.wait_until(timeout)) {
+        CvStatus status;
+        {
+          ThreadStatusManager::Sleeper sleeper(thread_status_manager);
+          status = condition_.wait_until(timeout);
+        }
```
mitza-oci, line 1075: Since the type of `condition_` is a class in this library, would it be cleaner to integrate directly with the ThreadStatusManager?
Maybe as `condition_.wait_until(timeout, thread_status_manager)` which would let the return value be used directly in the `switch` like it was before.

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -873,12 +877,27 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+
+  if (thread_status_manager.update_thread_status()) {
+    if (relay_thread_monitor->start() == -1) {
+      ACE_ERROR((LM_ERROR, ACE_TEXT("(%P:%t) ERROR: failed to activate Thread Load Monitor\n")));
+      return EXIT_FAILURE;
+    }
+
+    OpenDDS::DCPS::ThreadStatusManager::Start s(thread_status_manager, "RtpsRelay Main");
+
+    for (;;) {
```
mitza-oci, line 891: Should there be some way to `break;` this loop?

tools/rtpsrelay/RelayThreadMonitor.h
```diff
@@ -1,128 +1,58 @@
...
-  ThreadDescriptor(const char *alias, OpenDDS::DCPS::MonotonicTimePoint tnow);
-  mutable ACE_Thread_Mutex queue_lock_;
-  std::string alias_;
-  OpenDDS::DCPS::ThreadStatusManager *tsm_;
-  LoadSamples samples_;
-  std::stack<OpenDDS::DCPS::ThreadMonitor::UpdateMode> nested_;
-  OpenDDS::DCPS::MonotonicTimePoint last_;
-  LoadHistory summaries_;
-};
+  RelayThreadMonitor(const Config& config)
```
mitza-oci, line 16:
```suggestion
  explicit RelayThreadMonitor(const Config& config)
```

mitza-oci (changes requested):

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -0,0 +1,223 @@
...
+ */
+
+#include <DCPS/DdsDcps_pch.h> // Only the _pch include should start with DCPS/
+
+#include "ThreadStatusManager.h"
+#include "Service_Participant.h"
+
+#include <ace/Arg_Shifter.h>
+#include <ace/Configuration_Import_Export.h>
+#include <ace/Argv_Type_Converter.h>
```
iguessthislldo, line 15: I don't think these are needed?

dds/DCPS/Service_Participant.cpp
```diff
@@ -1557,6 +1557,14 @@ Service_Participant::load_configuration(
                      -1);
   }
 
+  if (status != 0) {
+    ACE_ERROR_RETURN((LM_ERROR,
+    ACE_TEXT("(%P|%t) ERROR: Service_Participant::load_configuration ")
+    ACE_TEXT("load_common_configuration () returned %d\n"),
+    status),
+    -1);
```
iguessthislldo, line 1565: Indent is missing and this should probably be a conditional notice log with a separate return.

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -873,12 +877,27 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+      ACE_Time_Value t = thread_status_manager.thread_status_interval().value();
```
iguessthislldo, line 892:
```suggestion
      const ACE_Time_Value t = thread_status_manager.thread_status_interval().value();
```

tools/rtpsrelay/RtpsRelay.cpp
iguessthislldo: What about checking if `run_reactor_event_loop` returned zero. If I'm reading the code right it looks like the reactor has been stopped if that's the case.

iguessthislldo (changes requested):

tools/rtpsrelay/RtpsRelay.cpp
jrw972: `run_reactor_event_loop` modifies the `ACE_Time_Value`.

dds/DCPS/SafetyProfileStreams.h
```diff
@@ -29,6 +29,7 @@ OpenDDS_Dcps_Export String to_dds_string(long long to_convert);
 OpenDDS_Dcps_Export String to_dds_string(unsigned long long to_convert, bool as_hex = false);
 OpenDDS_Dcps_Export String to_dds_string(unsigned long to_convert, bool as_hex = false);
 OpenDDS_Dcps_Export String to_dds_string(const unsigned char* array, size_t length);
+OpenDDS_Dcps_Export String to_dds_string(const String& to_convert);
```
mitza-oci, line 32: Why is this function needed?  Add a comment if we're keeping it?

jrw972: The thread id can be a string or integer.  In either case, calling `to_dds_string` generically converts it to a string.

mitza-oci: That makes sense, but I think it should be `inline` and add a comment.

iguessthislldo (approved):

mitza-oci:
CI error due to new feature:

```
GitHub Actions / test_u20_p1_sec

tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl reverse

2022-01-15 20:22:18.573@LM_ERROR@(17518|17521) ERROR: RelayThreadMonitor::one_data_available failed to read No data
test FAILED.
auto_run_tests.pl: Error: "tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl reverse" returned with status 255

```

PR #3258, opened 2022-01-05
Fix LivelinessChangedStatus Count Values In DataReaderImpl

simpsont-oci:
Problem: There's a race condition for resetting LivelinessChangedStatus *_count_change values in DataReaderImpl (as well as a data race for the internal object when releasing the lock).

Solution: Use the lock to copy status & reset values, pass the copy.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -2865,12 +2865,12 @@ void DataReaderImpl::notify_liveliness_change()
   = listener_for(DDS::LIVELINESS_CHANGED_STATUS);
 
   if (!CORBA::is_nil(listener.in())) {
-    listener->on_liveliness_changed(this, liveliness_changed_status_);
-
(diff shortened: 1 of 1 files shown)


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -2865,12 +2865,12 @@ void DataReaderImpl::notify_liveliness_change()
   = listener_for(DDS::LIVELINESS_CHANGED_STATUS);
 
   if (!CORBA::is_nil(listener.in())) {
-    listener->on_liveliness_changed(this, liveliness_changed_status_);
-
-    ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, sample_lock_);
-
+    ACE_Guard<ACE_Recursive_Thread_Mutex> guard(sample_lock_);
+    DDS::LivelinessChangedStatus status = liveliness_changed_status_;
```
mitza-oci, line 2869:
```suggestion
    const DDS::LivelinessChangedStatus status = liveliness_changed_status_;
```

mitza-oci (approved):

PR #3259, opened 2022-01-05
Fix Windows Iterator Assertion In RemoveAssociationSweeper

simpsont-oci:
Problem: RemoveAssociationSweeper tries to pre-calculate the last iterator, but on Windows this causes an iterator assertion when the set is empty (despite the fact it's never used).

Solution: Don't even proceed to making the calculation if the set is empty.


mitza-oci (approved):
I don't mind merging this as-is.  Two comments though:
- I'll be pedantic as usual: Windows doesn't know or care about C++ iterators, they're tied to the compiler/stdlib not OS
- Should we look at getting #3121 from @jrw972 merged?

PR #3260, opened 2022-01-06
RapidJson generator is redundant

jrw972:
Problem
-------

The RapidJson generator code in the IDL compiler is redundant with the
JsonValueReader and JsonValueWriter.

Solution
--------

Remove the RapidJson generator and replace with JsonValueReader and
JsonValueWriter.  This required some changes:

* union discriminators in bench are now "$discriminator"
* JsonValueReader and JsonValueWriter now treat characters as strings
* JsonValueReader accepts an integer for a enum


mitza-oci:
Does the doxygen comment on `JsonValueWriter` need to be updated?

performance-tests/bench/bench_exe.mpb
```diff
@@ -1,4 +1,4 @@
-project: dcps_rtps_udp, dcps_rapidjson {
+project: dcps_rtps_udp {
```
mitza-oci, line 1: Does this project, or some related project, need to inherit from `rapidjson`?

tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.json
```diff
@@ -0,0 +1,81 @@
...
+    "l": -2147483632,
+    "ull": 9223372036854775810,
+    "ll": -9223372036854775806,
+    "f": 0.0014100000262260437,
+    "d": -0.00000141,
+    "ld": 1e+34,
+    "b": false,
+    "c": "\r",
+    "str": "The most JSON of rapids",
+    "wstr": "Τηισ ισ α τεστ οφ τηε εμεργενψυ βροαδψαστ συστεμ. פךקשדק לקקפ טםור ישמגד ןמדןגק איק הקיןבךק שא שךך אןצקד"
```
simpsont-oci, line 22: I think we mentioned removing UTF-8 from our json files recently... is now the time to try to switch the test strings over to escaped unicode points?

performance-tests/bench/example/config/worker/bollea_aggregator.json
```diff
@@ -142,10 +142,10 @@
       "writers": [ "datawriter_01" ],
       "params": [
         { "name": "data_buffer_bytes",
-          "value": { "_d": "PVK_ULL", "ull_prop": 2372 }
+          "value": { "$discriminator": "PVK_ULL", "ull_prop": 2372 }
```
simpsont-oci, line 145: We will also need to update the JSON configs in nightly for the performance dashboard. Ideally, we'd check in those changes just before merging this to master so the dashboard run pulled the right format JSON.

performance-tests/bench/bench_exe.mpb
simpsont-oci: Yeah, how does this work? Does the core library / ValueReader now require RapidJSON or do we need to specify it here?

dds/DCPS/JsonValueWriter.h
```diff
@@ -87,219 +81,233 @@ class JsonValueWriter : public ValueWriter {
...
+  ACE_CDR::WChar s[2] = { value, 0 };
+
+  rapidjson::GenericStringStream<rapidjson::UTF16<> > source(s);
+  rapidjson::GenericStringBuffer<rapidjson::UTF8<> > target;
+
+  if (!rapidjson::Transcoder<rapidjson::UTF16<>, rapidjson::UTF8<> >::Transcode(source, target)) {
+    return;
+  }
+
+  writer_.String(target.GetString(), target.GetLength());
```
simpsont-oci, line 285: conversion warnings, in case you missed it... here and line 291 and line 306.

dds/DCPS/JsonValueWriter.h
```diff
@@ -308,26 +316,38 @@ void JsonValueWriter<Buffer>::write_enum(const char* name,
 template<typename T>
 std::string to_json(const T& sample)
 {
-  JsonValueWriter<> jvw;
+  rapidjson::StringBuffer buffer;
+  rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+  JsonValueWriter<rapidjson::Writer<rapidjson::StringBuffer>> jvw(writer);
```
simpsont-oci, line 321: Issue with older compilers and ``>>`` for nested templates... here and line 340.

simpsont-oci (changes requested):

tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.json
jrw972: What was the motivation for that?

simpsont-oci: The security issue with unicode-encoded source files. I believe there are escape sequences which can move the cursor, allowing for "rewriting" the string. @mitza-oci was the one who brought it up, originally.

simpsont-oci: So the JSON might be fine, but the place where it's encoded in C++ wouldn't be, for example.

simpsont-oci (approved):
I think we'll also want to remove Bench_Idl_Json in order to resolve link issues with Bench_Idl. Other than that, looks good.

tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.json
simpsont-oci: This can probably wait...

PR #3277, opened 2022-01-20
MultiTopic Test Fix

ClaytonCalabrese:
The MultiTopic test is experiencing a few failures on the OpenDDS Scoreboard. Among these is one where an improperly joined sample in being inserted using `store_synthetic_data`. This happens if the order of received topics is done in such a way that all keys are known, but not all topics are received. A a topic causing cross join would then need to be received before the other remaining topic, leading to the failure.

Currently still need to resolve how to deal with the `ACE_GUARD` because it is attempting an empty `return` if uncommented.

### DIFF
--- a/dds/DCPS/MultiTopicDataReader_T.cpp
+++ b/dds/DCPS/MultiTopicDataReader_T.cpp
@@ -245,7 +245,7 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::process_joins(
           }
           DDS::ReturnCode_t ret = join(join_result, starting[i], keys,
             other_key.ptr_, other_dr, other_meta);
-          if (ret == DDS::RETCODE_NO_DATA) {
+          if (ret != DDS::RETCODE_OK) {
             return ret;
           }
         }
@@ -253,7 +253,7 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::process_joins(
         if (!join_result.empty() && !seen.count(other_topic)) {
           // recurse
           DDS::ReturnCode_t ret = process_joins(partialResults, join_result, withJoin, other_qp);
-          if (ret == DDS::RETCODE_NO_DATA) {
+          if (ret != DDS::RETCODE_OK) {
             return ret;
           }
         }
@@ -294,7 +294,7 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::cross_join(
       for (typename SampleVec::iterator i = iterPR->second.begin();
         i != iterPR->second.end(); ++i) {
         DDS::ReturnCode_t ret = join(resulting, *i, no_keys, NULL, qp.data_reader_, other_meta);
-        if (ret == DDS::RETCODE_NO_DATA) {
+        if (ret != DDS::RETCODE_OK) {
           return ret;
         }
       }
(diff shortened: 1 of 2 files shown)


iguessthislldo:
> Currently still need to resolve how to deal with the ACE_GUARD because it is attempting an empty return if uncommented.

`ACE_GUARD_RETURN` is the form of `ACE_GUARD` with a return value if the mutex lock attempt fails. There is a `ACE_READ_GUARD_RETURN` as well, but I think @mitza-oci has commented before that the read/write specific mutexes don't apply to any platform we use. As a side note, for a mutex failure for `DDS::ReturnCode_t` I personally use `DDS::RETCODE_OUT_OF_RESOURCES`, but I'm not sure it matters too much and a `DDS::RETCODE_ERROR` would probably be okay too.

mitza-oci:
> > Currently still need to resolve how to deal with the ACE_GUARD because it is attempting an empty return if uncommented.
> 
> `ACE_GUARD_RETURN` is the form of `ACE_GUARD` with a return value if the mutex lock attempt fails. 

Yes, you can use this macro (see `Global_Macros.h`), or you can use the ACE objects directly without macros.

>There is a `ACE_READ_GUARD_RETURN` as well, but I think @mitza-oci has commented before that the read/write specific mutexes don't apply to any platform we use. As a side note, for a mutex failure for `DDS::ReturnCode_t` I personally use `DDS::RETCODE_OUT_OF_RESOURCES`, but I'm not sure it matters too much and a `DDS::RETCODE_ERROR` would probably be okay too.

Readers/writer locks are a whole different design dimension.  If you're not already using them I'd go with just simple mutexes.  RW locks are an optimization of that, which can be handled separately. I don't think I said that they don't apply -- I think all platforms do support them in some way.  But there is overhead to managing the more complex synchronization object so it's not going to be faster in all cases.  Probably depends on how many concurrent readers would be expected.

dds/DCPS/MultiTopicDataReader_T.cpp
```diff
@@ -283,18 +293,26 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::cross_join(
       SampleVec resulting;
       for (typename SampleVec::iterator i = iterPR->second.begin();
         i != iterPR->second.end(); ++i) {
-        join(resulting, *i, no_keys, NULL, qp.data_reader_, other_meta);
+        DDS::ReturnCode_t ret = join(resulting, *i, no_keys, NULL, qp.data_reader_, other_meta);
```
iguessthislldo, line 296:
```suggestion
        const DDS::ReturnCode_t ret = join(resulting, *i, no_keys, 0, qp.data_reader_, other_meta);
```

And the rest of the return codes.

dds/DCPS/MultiTopicDataReader_T.cpp
```diff
@@ -240,13 +243,19 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::process_joins(
             other_meta.assign(other_key.ptr_, keys[j].c_str(),
               &starting[i], keys[j].c_str(), resulting_meta);
           }
-          join(join_result, starting[i], keys,
+          DDS::ReturnCode_t ret = join(join_result, starting[i], keys,
```
iguessthislldo, line 246:
```suggestion
          const DDS::ReturnCode_t ret = join(join_result, starting[i], keys,
```

dds/DCPS/MultiTopicDataReader_T.h
```diff
@@ -128,22 +128,22 @@ class MultiTopicDataReader_T
                      const MetaStruct& meta);
 
   // Process all joins (recursively) in the QueryPlan 'qp'.
-  void process_joins(OPENDDS_MAP(TopicSet, SampleVec)& partialResults,
+  DDS::ReturnCode_t process_joins(OPENDDS_MAP(TopicSet, SampleVec)& partialResults,
                      SampleVec starting, const TopicSet& seen,
```
iguessthislldo, line 132: Can you fix the indent here and below?

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3278, opened 2022-01-20
Various CI Fixes

iguessthislldo:
- `auto_run_tests.pl`:
  - Two small parts of #2452: Automatically add  `-Config RTPS` and `-Config GH_ACTIONS` if running on GitHub Actions
  - Automatically add an OS `-Config` value according to `$^O`.
  - Try to fix issue with the CTest missing build config on Windows.
- `build_and_test.yml`:
  - Add script to lint workflow that tries to determine when GHA C++ compiler problem matchers are needed.
  - Added, removed, and moved problem matchers entries based on feedback from the script.
  - Remove `-Config` options no longer needed because they are automatically added.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  DDS_ROOT: ${{github.workspace}}
+  DDS_ROOT: ${{github.workspace}}/OpenDDS
   DOC_ROOT: ${{github.workspace}}/ACE_TAO
   ACE_ROOT: ${{github.workspace}}/ACE_TAO/ACE
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,6 +35,6 @@ jobs:
         install-modules: |
           YAML
     - name: Run lint.pl
-      run: cd OpenDDS ; perl $DDS_ROOT/tools/scripts/lint.pl --color --ace
+      run: cd $DDS_ROOT ; perl tools/scripts/lint.pl --color --ace
     - name: Run lint_build_and_test.pl
       run: perl $DDS_ROOT/.github/workflows/lint_build_and_test.pl
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,6 +21,8 @@ jobs:
     steps:
     - name: checkout OpenDDS
       uses: actions/checkout@v2
+      with:
+        path: ${{ env.DDS_ROOT }}
     - name: checkout ACE_TAO
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -11,22 +11,30 @@ concurrency:
...
+      uses: shogo82148/actions-setup-perl@v1
+      with:
+        install-modules: |
+          YAML
     - name: Run lint.pl
-      run: |
-        perl $DDS_ROOT/tools/scripts/lint.pl --color --ace
+      run: cd $DDS_ROOT ; perl tools/scripts/lint.pl --color --ace
+    - name: Run lint_build_and_test.pl
+      run: perl $DDS_ROOT/.github/workflows/lint_build_and_test.pl
```
mitza-oci, line 40: Should this script live in `tools/scripts`?  Would a developer run it locally?

.github/workflows/lint_build_and_test.pl
```diff
@@ -0,0 +1,113 @@
...
+for my $job_name (keys(%{$gha->{jobs}})) {
+  my $job = $gha->{jobs}->{$job_name};
+  # print("#### Job $job_name\n");
+  my $expected_matcher;
+  my @expected_build_commands = ();
+  if ($job->{'runs-on'} =~ /^windows/) {
+    $expected_matcher = 'msvc-problem-matcher';
+    push(@expected_build_commands, qr/(?<!\w)msbuild/);
+  }
+  elsif ($job->{'runs-on'} =~ /^ubuntu/ || $job->{'runs-on'} =~ /^macos/) {
```
mitza-oci, line 25:
```suggestion
  elsif ($job->{'runs-on'} =~ /^ubuntu|macos/) {
```

.github/workflows/lint_build_and_test.pl
```diff
@@ -0,0 +1,113 @@
...
+        $build_command_expected = 1;
+      }
+
+      my $step_has_build_command = 0;
+      for my $expected_build_command (@expected_build_commands) {
+        if ($step->{run} =~ /($expected_build_command)/) {
+          $job_has_build_command = 1;
+          $step_has_build_command = 1;
+          die("$what unexpected build command: based on run string: $1")
+            if (!$build_command_expected);
```
mitza-oci, line 91:
```suggestion
            unless $build_command_expected;
```
Not sure there's a preference... but this seems like it's cleaner

mitza-oci (changes requested):

.github/workflows/lint.yml
iguessthislldo: I see `tools/scripts` as mostly scripts that apply to multiple places or all of the repo, so I'm hesitant to move it away from the only file it's checking. A developer might want to run it, but they'd have to remember and it might be easier to remember if the script is in the same directory as the file they're editing.

mitza-oci (approved):

PR #3282, opened 2022-01-20
[GitHub Actions] Append Printed List of Failed Tests

simpsont-oci:


jwillemsen:
Is the autobuild repo not a better place, that way other projects can also use it

simpsont-oci:
It's a good thought, but unfortunately this script is pretty hand-tailored to just spitting out the test results in a way that looks OK for our pull requests. It's not generic enough to even handle the "full" output, let alone anything fancy like using autobuild for both building & testing.

mitza-oci (approved):

PR #3284, opened 2022-01-20
updated rapidjson submodule

mitza-oci:


simpsont-oci:
Looks like there are a bunch of 32-bit float unit test failures?

mitza-oci:
> Looks like there are a bunch of 32-bit float unit test failures?

Could be that it's making an assumption about rounding or other floating point conversion problem

simpsont-oci (approved):

PR #3285, opened 2022-01-21
Attempt to fix StatusCondition waiting timeout in XTypes test

sonndinh:


ClaytonCalabrese:
The wait_helper and wait_pub/sub_matched functions seem like something that is going to be reused in a bunch of tests. Perhaps it would be better to pull these out and make them generic? I don't think it is necessary for this PR but something that we may need in the future.

PR #3290, opened 2022-01-22
Reduce TransportLocatorSeq Copying in SEDP

simpsont-oci:
Problem: We're currently copying the Reader and WriterAssociation objects around quite a bit, as well as the serialized type info during associations.

Solution: Make ReaderAssociationRecord and WriterAssociationRecord inherit from RcObject, only create them when necessary, and avoid extra copying when possible by assigning / serializing directly into the record objects (via const casts).

### DIFF
--- a/dds/DCPS/RTPS/Sedp.cpp
+++ b/dds/DCPS/RTPS/Sedp.cpp
@@ -7111,7 +7111,9 @@ void Sedp::match_continue(const GUID_t& writer, const GUID_t& reader)
   const DDS::SubscriberQos* subQos = 0;
   DCPS::TransportLocatorSeq* rTls = 0;
   ACE_CDR::ULong rTransportContext = 0;
+#ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
   const DCPS::ContentFilterProperty_t* cfProp = 0;
+#endif
   XTypes::TypeInformation* reader_type_info = 0;
   DCPS::MonotonicTime_t reader_participant_discovered_at;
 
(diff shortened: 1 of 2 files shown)


dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -7304,27 +7304,36 @@ void Sedp::match_continue(const GUID_t& writer, const GUID_t& reader)
...
 
     // Copy reader and writer association data prior to releasing lock
-    DDS::OctetSeq octet_seq_type_info_reader;
-    XTypes::serialize_type_info(*reader_type_info, octet_seq_type_info_reader);
-    const DCPS::ReaderAssociation ra = {
-      *rTls, rTransportContext, reader, *subQos, *drQos,
+    WriterAssociationRecord_rch war;
+    if (call_writer) {
+      war = make_rch<WriterAssociationRecord>(dwr, writer, DCPS::ReaderAssociation());
+      DCPS::ReaderAssociation& ra = const_cast<DCPS::ReaderAssociation&>(war->reader_association_);
```
mitza-oci, line 7310: Would it be cleaner just to change this to be a nonconst member?
Or pass it to the constructor (unless that's too slow).

dds/DCPS/RTPS/Sedp.cpp
```diff
@@ -7304,27 +7304,36 @@ void Sedp::match_continue(const GUID_t& writer, const GUID_t& reader)
...
+      ra.readerId = reader;
+      ra.subQos = *subQos;
+      ra.readerQos = *drQos;
 #ifndef OPENDDS_NO_CONTENT_FILTERED_TOPIC
-      cfProp->filterClassName, cfProp->filterExpression,
-#else
-      "", "",
+      ra.filterClassName = cfProp->filterClassName;
+      ra.filterExpression = cfProp->filterExpression;
+      ra.exprParams = cfProp->expressionParameters;
```
mitza-oci, line 7319: See warnings annotated on line 7114 due to conditional compilation

mitza-oci (changes requested):

dds/DCPS/RTPS/Sedp.cpp
simpsont-oci: I didn't want to go the constructor route because the TransportLocatorSeq is a member of ReaderAssociation (a taoidl-generated type) and unless taoidl's correctly generating move-constructors it would probably wind up copying the sequence twice.

I think having the const is nice for the users of the ReaderAssociationRecord, so I left that in place even though it makes the creation side of things a bit uglier.

mitza-oci (approved):

PR #3293, opened 2022-01-24
Governance files for security attributes test are large

jrw972:
Problem
-------

The governance files for the security attributes test take up a
significant amount of disk space.

Solution
--------

Dynamically generate the governance file.


tests/security/attributes/run_test.pl
```diff
@@ -59,47 +59,58 @@
...
 }
 
+sub gov_gen {
+    my $aup = shift;
+    my $ejac = shift;
+    my $dpk = shift;
+    my $lpk = shift;
+    my $rpk = shift;
+
+    system("./gov_gen -aup $aup -ejac $ejac -dpk $dpk -lpk $lpk -rpk $rpk -o governance.xml");
```
mitza-oci: Use existing Perl support libraries to spawn processes

tests/security/attributes/run_test.pl
```diff
@@ -59,47 +59,58 @@
...
+    system("openssl smime -sign -in governance.xml -text -out governance.xml.p7s -signer $DDS_ROOT/tests/security/certs/permissions/permissions_ca_cert.pem -inkey $DDS_ROOT/tests/security/certs/permissions/permissions_ca_private_key.pem");
```
mitza-oci: I'm not sure we can guarantee that `openssl` is in the PATH on all builds.  What if we had `gov_gen` take care of this step by linking to the OpenSSL libs.

mitza-oci (changes requested):

tests/security/attributes/run_test.pl
jrw972: Can you be more specific and/or point to an example?

mitza-oci:
```
    $test->process($name, $exe);
    $test->start_process($name);
```
I'm not sure if `TestFramework` is best suited to this, but it's one option.  Another would be using `PerlDDS/Process.pm` which inherits from the corresponding class in ACE.

mitza-oci (approved):

mitza-oci:
Looks like there's a few instances of the attributes failing in GitHub Actions.  May be easiest to see the results here https://github.com/objectcomputing/OpenDDS/actions/runs/1752812019

simpsont-oci (approved):

PR #3296, opened 2022-01-25
Release message block chains after each get_*_value(s) call in DynamicData

sonndinh:


dds/DCPS/XTypes/DynamicData.h
```diff
@@ -154,7 +154,30 @@ class OpenDDS_Dcps_Export DynamicData {
 
   bool check_xcdr1_mutable(const DynamicType_rch& dt);
 
+  typedef OPENDDS_VECTOR(ACE_Message_Block*) IntermediateChains;
+  const IntermediateChains& get_intermediate_chains() { return chains_to_release; }
```
mitza-oci, line 158: Could these be private?

mitza-oci, line 158:
```suggestion
  typedef OPENDDS_VECTOR(ACE_Message_Block*) IntermediateChains;
  const IntermediateChains& get_intermediate_chains() const { return chains_to_release; }
```

dds/DCPS/XTypes/DynamicData.h
```diff
@@ -154,7 +154,30 @@ class OpenDDS_Dcps_Export DynamicData {
...
+
 private:
+
+  class ScopedChainManager {
+  public:
+    ScopedChainManager(DynamicData& dd)
```
mitza-oci, line 164:
```suggestion
    explicit ScopedChainManager(DynamicData& dd)
```

mitza-oci (changes requested):

mitza-oci (approved):
Does this use case work?  I think it does, but just want to verify.  If it's not already in tests we may want to add it.
```
DynamicData d2;
{
  DynamicData d1(init_for_overall_data_sample);
  d1.get_complex_value(d2, id_nested);
}
// d2 is still valid
```

sonndinh:
> 

Yes, it should work because `d2` is independent of `d1`. I'll add a test case for that.

PR #3309, opened 2022-02-02
ReactorTask catches

jrw972:
Problem
-------

ReactorTask catches exceptions, logs them, and throws them again.  The
stack may be lost so a stack trace does not show the source of the
exception.

Solution
--------

Don't catch the exception.


jrw972:
For reference #3090

mitza-oci (approved):

PR #3315, opened 2022-02-04
Improvements to Installation of CMake Targets

iguessthislldo:
- CMake Install-related:
    -  Fix issue when using `install(EXPORT)` on a target used with `OPENDDS_TARGET_SOURCES`.
    - Added CMake test where the target is installed.
    - OpenDDS, ACE, and TAO libraries can now be installed using [`install(IMPORTED_RUNTIME_ARTIFACTS)`](https://cmake.org/cmake/help/latest/command/install.html#imported-runtime-artifacts) if using CMake 3.21 or later.
       - Added `opendds_get_library_dependencies` to help use this.
    - `OPENDDS_TARGET_SOURCES` now sets properties on the target which lists the files that where passed and generated that are part of the `PUBLIC` and `INTERFACE` scopes. This is meant to help setting up those files to be installed.
    - Add `OPENDDS_DEFAULT_SCOPE` option that allows changing the default scope of `OPENDDS_TARGET_SOURCES`.
- Other CMake Changes:
    - Add explicit `PUBLIC` to IDL libraries in tests that are used by executables, except `generated_global`.
    - Added some groundwork for allowing a non-flat hierarchy of IDL files.
    - Add test for and possible fix for https://github.com/objectcomputing/OpenDDS/issues/1297


cmake/dds_idl_sources.cmake
```diff
@@ -245,6 +270,9 @@ function(opendds_target_idl_sources target)
           -I${DDS_ROOT}
           -I${idl_file_dir} # The type-support IDL will include the primary IDL file
           ${_arg_TAO_IDL_FLAGS}
+          -o "${output_dir}"
+          -oA "${output_dir}"
+          -oS "${output_dir}"
```
mitza-oci, line 275: doesn't seem like these are necessary, `-o` should cover it

cmake/tao_idl_sources.cmake
```diff
@@ -46,101 +46,120 @@ if (NOT TAO_HAS_OPTIMIZE_COLLOCATED_INVOCATIONS)
   list(APPEND TAO_CORBA_IDL_FLAGS -Sp -Sd)
 endif()
 
-function(tao_idl_command name)
-  set(multiValueArgs IDL_FLAGS IDL_FILES WORKING_DIRECTORY)
+function(opendds_get_generated_output_dir target output_dir_var)
+  # TODO base output_dir_var on target
```
mitza-oci, line 50: TODO in this PR?

configure
```diff
@@ -2485,35 +2485,49 @@ sub write_cmake_file {
...
 
     my %disabled_by_default = map {$_, 0} qw(
+      ipv6
       security
+      static
     );
 
     my %ignore = map {$_, 0} qw(
+      ace
+      ace-github-latest
```
mitza-oci, line 2508: I'm not that familiar with how this is used... but could we just assume that a key is an `ignore` key if it's not in one of the other hashes?

docs/cmake.md
```diff
@@ -158,76 +181,86 @@ The following assumes Visual Studio 2017 using 64-bit architecture (adjust the
...
 
 ```bash
-DDS_WORKSPACE=$(pwd)
-cd OpenDDS-src
-./configure --prefix=$DDS_WORKSPACE/opendds-install
-make
+OPENDDS_PREFIX="$PWD/opendds-install"
+DDS_ROOT="$PWD/OpenDDS"
+ACE_ROOT="$DDS_ROOT/ACE_wrappers"
+export PERL5LIB="$DDS_ROOT/bin:$ACE_ROOT/bin"
```
mitza-oci, line 204: Is this needed for the build or could we move it down to right before (or with) the `run_test.pl` line?

docs/cmake.md
```diff
@@ -384,3 +423,102 @@ one or more of the following:
...
+`opendds_idl` or static libraries.
+
+If OpenDDS and ACE/TAO is built with `clang`, the shared libraries might be
+missing an `SONAME` entry. Is is
+[an issue with ACE/TAO](https://github.com/DOCGroup/ACE_TAO/issues/1790).
+If trying to use `install(IMPORTED_RUNTIME_ARTIFACTS)` in this case, it causes
+the dynamic linker to ignore the libraries and report that they could not be
+found. One workaround is to add `SOFLAGS+=-Wl,-h,$(SONAME)` to
+`$ACE_ROOT/include/makeinclude/platform_macros.GNU` before building. This can
+be done mannually after running the configure script or by passing
```
mitza-oci, line 494:
```suggestion
be done manually after running the configure script or by passing
```

run the resulting doc through a spellcheck

tests/cmake/install/install_test_framework.pl
```diff
@@ -0,0 +1,49 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
```
mitza-oci, line 1: Is this generally useful (as in, outside of `tests/cmake`)?

tests/cmake/install/library/install_library.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+  const Vector vel_change = multipy(object.acc, time);
+  const Vector new_vel = add(object.vel, vel_change);
+  object.pos = add(object.pos, add(multipy(object.vel, time), multipy(vel_change, time / 2)));
+  object.vel = new_vel;
+}
+
+void print(const MovingObject& object, const char* name)
+{
+  printf("%s is at %lf %lf\n", name, object.pos.x, object.pos.y);
```
mitza-oci, line 32:
```suggestion
  printf("%s is at %f %f\n", name, object.pos.x, object.pos.y);
```

mitza-oci (changes requested):

cmake/OpenDDSConfig.cmake
```diff
@@ -302,53 +306,71 @@ macro(opendds_vs_force_static)
...
+endif()
 
 include(SelectLibraryConfigurations)
 
 foreach(_lib ${_all_libs})
   string(TOUPPER ${_lib} _LIB_VAR)
   select_library_configurations(${_LIB_VAR})
+
+  # These paths might be symlinks and IMPORTED_RUNTIME_ARTIFACTS seems to not
+  # work correctly with symlinks, so resolve them now.
```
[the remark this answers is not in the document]
iguessthislldo: I didn't think about that, but maybe I should looks for an issue or make if there isn't one. The issue was it copied the symlinks instead of the actual files.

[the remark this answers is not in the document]
iguessthislldo: I have created the issue: https://gitlab.kitware.com/cmake/cmake/-/issues/23249 and will note it in the comment.

tests/cmake/install/install_test_framework.pl
iguessthislldo: I'm not entirely sure how useful it is. I made this out of paranoia because I wanted to remove as many references to the OpenDDS source tree as possible. Some one might find it useful though, so I can move it to `tools/scripts/`.

mitza-oci (approved):

PR #3318, opened 2022-02-04
Splitting GHA jobs for safety profile builds

sonndinh:


ClaytonCalabrese:
I think I've seen that error from tar before if what I'm trying to zip is too big.

simpsont-oci:
> I think I've seen that error from tar before if what I'm trying to zip is too big.

Just now taking a look at this. It looks like the output for git clean is not what's expected (of the form "would remove X" for the first line ("would not remove the current working directory"). Pretty sure this is happening because your start directory (build/target) is different than the other jobs, since it does a form of cross-compiling for safety profile. You can probably either ignore that first line or you can try to build he artifact from one step further up. Since we've never tried to cache the results for a cross compile, to the best of my knowledge, this may take a bit of fiddling to get right...

sonndinh:
> > I think I've seen that error from tar before if what I'm trying to zip is too big.
> 
> Just now taking a look at this. It looks like the output for git clean is not what's expected (of the form "would remove X" for the first line ("would not remove the current working directory"). Pretty sure this is happening because your start directory (build/target) is different than the other jobs, since it does a form of cross-compiling for safety profile. You can probably either ignore that first line or you can try to build he artifact from one step further up. Since we've never tried to cache the results for a cross compile, to the best of my knowledge, this may take a bit of fiddling to get right...

Yes, the first line of the git clean output is different than that of the other builds, so tar failed. I'm trying to add a bash script to one of the steps to skip any line that has more than 3 tokens. But I'm having an issue with how GHA environment handles bash variables. Seems like the unexpected output of git clean only occurs on the first line. If that is the case, I think I can simplify the script a bit.

mitza-oci:
> > > I think I've seen that error from tar before if what I'm trying to zip is too big.
> > 
> > 
> > Just now taking a look at this. It looks like the output for git clean is not what's expected (of the form "would remove X" for the first line ("would not remove the current working directory"). Pretty sure this is happening because your start directory (build/target) is different than the other jobs, since it does a form of cross-compiling for safety profile. You can probably either ignore that first line or you can try to build he artifact from one step further up. Since we've never tried to cache the results for a cross compile, to the best of my knowledge, this may take a bit of fiddling to get right...
> 
> Yes, the first line of the git clean output is different than that of the other builds, so tar failed. I'm trying to add a bash script to one of the steps to skip any line that has more than 3 tokens. But I'm having an issue with how GHA environment handles bash variables. Seems like the unexpected output of git clean only occurs on the first line. If that is the case, I think I can simplify the script a bit.

You may not need `git clean` for this.  The outputs of the build process are in `build/host` and `build/target`.  Those can be archived.  You could probably have it skip archiving symlinks, or just go ahead and keep them since they should be small.

sonndinh:
> > > > I think I've seen that error from tar before if what I'm trying to zip is too big.
> > > 
> > > 
> > > Just now taking a look at this. It looks like the output for git clean is not what's expected (of the form "would remove X" for the first line ("would not remove the current working directory"). Pretty sure this is happening because your start directory (build/target) is different than the other jobs, since it does a form of cross-compiling for safety profile. You can probably either ignore that first line or you can try to build he artifact from one step further up. Since we've never tried to cache the results for a cross compile, to the best of my knowledge, this may take a bit of fiddling to get right...
> > 
> > 
> > Yes, the first line of the git clean output is different than that of the other builds, so tar failed. I'm trying to add a bash script to one of the steps to skip any line that has more than 3 tokens. But I'm having an issue with how GHA environment handles bash variables. Seems like the unexpected output of git clean only occurs on the first line. If that is the case, I think I can simplify the script a bit.
> 
> You may not need `git clean` for this. The outputs of the build process are in `build/host` and `build/target`. Those can be archived. You could probably have it skip archiving symlinks, or just go ahead and keep them since they should be small.

You're right. That simplifies it a lot.

ClaytonCalabrese (approved):

mitza-oci:
These kind of test failures https://github.com/objectcomputing/OpenDDS/runs/5120217620?check_suite_focus=true#step:14:11 seem to be due to a mismatch between which tests are compiled and which ones it tries to run.  I think once those are resolved we'll be ready to merge.

sonndinh:
> These kind of test failures https://github.com/objectcomputing/OpenDDS/runs/5120217620?check_suite_focus=true#step:14:11 seem to be due to a mismatch between which tests are compiled and which ones it tries to run. I think once those are resolved we'll be ready to merge.

Right. There is a difference between the configure steps of the ACE_TAO build job and the OpenDDS build job that causes those tests not compiled. I'm fixing it.

PR #3321, opened 2022-02-07
Internal DDS Proof of Concept (POC)

jrw972:
There are a number of cases in the OpenDDS code base where one component produces data that is consumed by other components.  This is often accomplished by registering a listener.  In some cases, the listener needs to receive already produced data.  Invoking listeners presents a problem with respect to concurrency since the listener cannot know the locks that may be held by the caller.

DDS/DCPS is a solution to the problems outlined above.  Conceptually, we can define a topic, writers can write samples that are stored in readers, and readers can read/take samples and/or use a listener.  This PR is an exploration of these possibilities applied to the NetworkConfig subsystem.  The writer is a NetworkConfigMonitor.  The readers are the SPDP transport and the RTPS transports.  The readers use a listener to respond to changes in network interfaces.


mitza-oci (changes requested):

dds/DCPS/InternalDataReader.h
```diff
@@ -0,0 +1,256 @@
...
+      }
+    }
+
+    const Listener_rch listener = listener_.lock();
+    if (schedule && listener) {
+      listener->schedule(rchandle_from(this));
+    }
+  }
+
+    void register_instance(InternalEntity_wrch publication_handle, const T& sample)
```
mitza-oci, line 92:
```suggestion
  void register_instance(InternalEntity_wrch publication_handle, const T& sample)
```

mitza-oci (approved):

PR #3328, opened 2022-02-09
configure: Ensure `--prefix` is an Absolute Path

iguessthislldo:
Prompted by a note in this
https://github.com/objectcomputing/OpenDDS/blob/c62aa01c8d97e41f498a3e4b6765531f3a59451e/docs/cmake.md#example-using-installed-opendds-unix-only

### DIFF
--- a/configure
+++ b/configure
@@ -463,6 +463,10 @@ if (exists($opts{'workspace'})) {
   $opts{'workspace'} = Cwd::realpath($opts{'workspace'});
 }
 
+if (exists($opts{'prefix'}) && !File::Spec->file_name_is_absolute($opts{'prefix'})) {
+  die("ERROR: --prefix argument $opts{'prefix'} is not an absolute path, stopped");
+}
+
 # Set initial libpath
 addCurLibPathRef(\%hostEnv);
 addCurLibPathRef(\%targetEnv);


mitza-oci:
Just curious... what happens if you use a relative path?  We're probably not expecting it in the makefiles, so each project-level makefile may install to a different directory.

mitza-oci (approved):

iguessthislldo:
I didn't think to try it, but I just did and yes, that's exactly what it does:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	dds/DCPS/InfoRepoDiscovery/totally-not-absolute/
	dds/DCPS/RTPS/totally-not-absolute/
	dds/DCPS/transport/multicast/totally-not-absolute/
	dds/DCPS/transport/rtps_udp/totally-not-absolute/
	dds/DCPS/transport/shmem/totally-not-absolute/
	dds/DCPS/transport/tcp/totally-not-absolute/
	dds/DCPS/transport/udp/totally-not-absolute/
	dds/FACE/totally-not-absolute/
	dds/InfoRepo/totally-not-absolute/
	dds/idl/totally-not-absolute/
	dds/monitor/totally-not-absolute/
	dds/totally-not-absolute/
	tools/dds/rtpsrelaylib/totally-not-absolute/
	tools/inspect/totally-not-absolute/
	tools/modeling/codegen/model/totally-not-absolute/
```

PR #3334, opened 2022-02-11
Network Address Changes

simpsont-oci:
Problem: ACE_INET_Addr is too heavyweight for efficient use as keys within map and set containers. In particular, operator< winds up taking far too long, especially when the internal vector of resolved addresses is being used, so data structures like the AddressCache suffer in performance when a simple memcmp would be preferable.

Solution: Introduce a lighter-weight NetworkAddress class which preserves much of the behavior of ACE_INET_Addr but with fast / simple comparisions. Refactor existing transport/framework/NetworkAddress to "NetworkResource" (as a string, it is more of a named resource than an actual address anyway), and update most comparison-sensitive usages to take advantage of NetworkAddress. Add unit test, and update how RtpsUdpDataLink's handling of "last_recv_addr" based on new class features (is_uniquelocal / is_linklocal / etc).

Note: We will eventually want to update more of the NetworkResource code to use NetworkAddress, as well as migrating other usages of ACE_INET_Addr to use NetworkAddress as comparable functionality is added.


mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -348,7 +349,7 @@ RtpsUdpSendStrategy::send_single_i(const iovec iov[], int n,
     errno = err;
   } else {
     if (link_->transport().config().count_messages()) {
-      const InternalMessageCountKey key(addr, MCK_RTPS, addr == link_->transport().config().rtps_relay_address());
+      const InternalMessageCountKey key(addr, MCK_RTPS, addr == NetworkAddress(link_->transport().config().rtps_relay_address()));
```
jrw972, line 352: Should `rtps_relay_address` and others just be a NetworkAddress?

dds/DCPS/NetworkAddress.h
```diff
@@ -0,0 +1,87 @@
...
+  bool is_loopback() const;
+
+  bool is_private() const; // IPv4 only
+
+  bool is_uniquelocal() const; // IPv6 only
+  bool is_linklocal() const; // IPv6 only
+  bool is_sitelocal() const; // IPv6 only
+
+private:
+  union ip46
```
jrw972, line 71: Was it impossible to declare and ACE_INET_Addr and get a pointer to its internal data structure?  Then, `to_addr` can just return the internal ACE_INET_Addr but the memcmp function will still be more performant.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -1612,7 +1612,68 @@ RtpsUdpDataLink::RtpsWriter::add_gap_submsg_i(RTPS::SubmessageSeq& msg,
   msg[idx].gap_sm(gap);
 }
 
-void RtpsUdpDataLink::update_last_recv_addr(const RepoId& src, const ACE_INET_Addr& addr)
+namespace
+{
+bool allow_addr_update(const NetworkAddress& current, const NetworkAddress& incoming)
```
jrw972, line 1617: This seems general enough that it belongs  in `NetworkAddress.h`.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2065,8 +2118,10 @@ RtpsUdpDataLink::RtpsReader::process_heartbeat_i(const RTPS::HeartBeatSubmessage
 
   // Only valid heartbeats (see spec) will be "fully" applied to writer info
   if (!(hb_first < 1 || hb_last < 0 || hb_last < hb_first.previous())) {
+    ACE_CDR::ULong cumulative_bits_added = 0;
     if (writer->recvd_.empty() && (directed || !writer->sends_directed_hb())) {
       OPENDDS_ASSERT(preassociation_writers_.count(writer));
+      gather_ack_nacks_i(writer, link, false, meta_submessages, cumulative_bits_added);
```
jrw972, line 2124: Why is this here?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -4610,18 +4664,19 @@ RtpsUdpDataLink::get_addresses_i(const RepoId& local) const {
...
-  const AddrSet::const_iterator it = unicast_addrs_.lower_bound(recv_no_port);
-  if (it != unicast_addrs_.end() &&
-      it->get_type() == last_recv_addr_.get_type() &&
-      it->get_addr_size() == last_recv_addr_.get_addr_size() &&
-      std::memcmp(it->get_addr(), last_recv_addr_.get_addr(), it->get_addr_size()) == 0) {
-    aset.insert(last_recv_addr_);
-    return true;
+  const uint16_t last_addr_type = last_recv_addr_.get_type();
+  NetworkAddress limit;
+  limit.set_type(last_recv_addr_.get_type());
```
jrw972, line 4672:
```suggestion
  limit.set_type(last_addr_type);
```

jrw972 (changes requested):

dds/DCPS/LogAddr.cpp
```diff
@@ -46,6 +46,24 @@ LogAddr::LogAddr(const ACE_INET_Addr& addr, Option opt)
   }
 }
 
+LogAddr::LogAddr(const NetworkAddress& sa, Option opt)
+{
+  ACE_INET_Addr addr = sa.to_addr();
+  if (opt == IpPort) {
```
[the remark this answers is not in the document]
simpsont-oci: Copied from the method above. I'll make it a switch and avoid the duplication.

dds/DCPS/NetworkAddress.cpp
```diff
@@ -0,0 +1,263 @@
...
+#endif
+  }
+  return false;
+}
+
+bool NetworkAddress::is_loopback() const
+{
+  if (inet_addr_.in4_.sin_family == AF_INET) {
+    // RFC 3330 defines loopback as any address with 127.x.x.x
+    return (ACE_HTONL(inet_addr_.in4_.sin_addr.s_addr) & 0xFF000000) == (INADDR_LOOPBACK & 0XFF000000);
```
sonndinh, line 210: Can the right-hand side just be `0XFF000000`?

dds/DCPS/NetworkAddress.h
simpsont-oci: I think the problem with this approach is that ACE_INET_Addr doesn't take pains to zero any padding for its internal data structure, so direct memcmp there is risky. So far, constructing ACE_INET_Addr based on this internal data structure seems pretty inexpensive, but if we're willing to eat the memory cost for storing an internal copy we could remove that construction cost as well.

dds/DCPS/NetworkAddress.cpp
```diff
@@ -0,0 +1,263 @@
...
+bool NetworkAddress::addr_bytes_equal(const NetworkAddress& rhs) const
+{
+  if (inet_addr_.in4_.sin_family == AF_INET) {
+    return memcmp(&inet_addr_.in4_.sin_addr, &rhs.inet_addr_.in4_.sin_addr, sizeof (inet_addr_.in4_.sin_addr)) == 0;
+#if defined (ACE_HAS_IPV6)
+  } else if (inet_addr_.in6_.sin6_family == AF_INET6) {
+    return memcmp(&inet_addr_.in6_.sin6_addr, &rhs.inet_addr_.in6_.sin6_addr, sizeof (inet_addr_.in6_.sin6_addr)) == 0;
+#endif
+  }
+  return 0;
```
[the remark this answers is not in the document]
simpsont-oci: I'd rather not put any logging in here if we can help it. Leave the logging to helper classes like LogAddr. Exceptions or asserts, maybe, but I'd prefer to add those to constructors / set methods if we're going to do that, so we know we can "trust" what's in existing NetworkAddress classes to at least be somewhat coherent.

dds/DCPS/NetworkAddress.cpp
simpsont-oci: `0x7F000000`, yes.

iguessthislldo (changes requested):

dds/DCPS/NetworkAddress.h
```diff
@@ -0,0 +1,87 @@
...
+
+  bool operator<(const NetworkAddress& rhs) const;
+
+  bool addr_bytes_equal(const NetworkAddress& rhs) const;
+
+  ACE_INET_Addr to_addr() const;
+  void to_addr(ACE_INET_Addr&) const;
+
+  int16_t get_type() const;
+  void set_type(int16_t type);
```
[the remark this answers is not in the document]
simpsont-oci: I think ACE_INET_Addr just uses `u_short` but I don't even know where that comes from. Looking into it... it looks like int16_t and u_short are defined in the same ACE header... `os_include/os_stdint.h`. I don't have a strong opinion here, other than desiring to avoid the identifier CORBA if possible.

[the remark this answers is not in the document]
mitza-oci: A few options to consider:
- this is a public API of a wrapper type, just use plain `int` -- it's not a protocol "on the wire" type -- and convert internally
- use whatever ACE defines in its os_include (but this is a bit messy since it wants to make POSIX and nonstandard unixy types interface types on all platforms)
- use ACE's CDR_Base types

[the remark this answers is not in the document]
simpsont-oci: CDR_Base types it is...

mitza-oci (approved):

dds/DCPS/NetworkAddress.h
jrw972: Right.  I'm assuming that it is possible to get a pointer to the guts of the ACE_INET_Addr.  If that's not possible, then my suggestion won't work and we are left with either constructing the ACE_INET_Addr on the fly or declaring one and eating the storage cost.  If it does, then we can still zero-pad and use memcmp.

jrw972 (approved):

PR #3335, opened 2022-02-11
DataReaderImpl_T allocator is not thread safe

jrw972:
Problem
-------

Allocations from `data_allocator_` in DataReaderImpl_T are
synchronized on the `sample_lock_`.  However, not all deallocations
are not synchronized.  For example, taking samples will place the
samples in a sequence and destroying the sequence may cause a
deallocation.

Solution
--------

Use a mutex in the relevant part of the allocator.

### DIFF
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -86,7 +86,7 @@ namespace OpenDDS {
       ACE_New_Allocator* allocator_;
     };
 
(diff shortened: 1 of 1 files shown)


mitza-oci (approved):

xufang-github:
After this issue was resolved, DataReader became especially time-consuming, and the efficiency of the entire DDS dropped by four times.

xufang-github:
How to solve the above problem?

jrw972:
Can you create a new issue with logs, metrics, etc. to illustrate the issue?

PR #3346, opened 2022-02-17
Improve DataReader Lookup Speeds For Large Numbers of Instances

simpsont-oci:
Problem: For large numbers of instances and instances, several lookups within DataReaderImpl_T are non-performant, opting to loop over the entire list of instances looking for appropriate samples, and then potentially looping over many samples to find matching samples.

Solution: Maintain separate lookup maps based on instance, view, and sample states, and keep them up to date as these values get updated.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -3146,18 +3146,17 @@ void DataReaderImpl::get_ordered_data(GroupRakeData& data,
       localsubs.insert(iter->second);
     }
   }
+
   ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, sample_lock_);
 
   for (SubscriptionInstanceSet::iterator iter = localsubs.begin(); iter != localsubs.end(); ++iter) {
-    if ((*iter)->instance_state_->match(view_states, instance_states)) {
+    const SubscriptionInstance_rch inst = *iter;
+    if (inst->instance_state_->match(view_states, instance_states)) {
       size_t i(0);
-      ReceivedDataElement* item = (*iter)->rcvd_samples_.get_next_match(sample_states, 0);
-      while (item) {
-        if (!item->coherent_change_) {
-          data.insert_sample(item, &(*iter)->rcvd_samples_, *iter, ++i);
-          group_coherent_ordered_data_.insert_sample(item, &(*iter)->rcvd_samples_, *iter, ++i);
-        }
-        ReceivedDataElement* item = (*iter)->rcvd_samples_.get_next_match(sample_states, item);
+      for (ReceivedDataElement* item = inst->rcvd_samples_.get_next_match(sample_states, 0);
+           item; item = inst->rcvd_samples_.get_next_match(sample_states, item)) {
+          data.insert_sample(item, &inst->rcvd_samples_, *iter, ++i);
+          group_coherent_ordered_data_.insert_sample(item, &inst->rcvd_samples_, *iter, ++i);
       }
     }
   }
@@ -3576,6 +3575,49 @@ void DataReaderImpl::OnDataAvailable::execute()
   if (set_subscriber_status_) {
     subscriber->set_status_changed_flag(::DDS::DATA_ON_READERS_STATUS, false);
   }
+}
+void DataReaderImpl::initialize_lookup_maps()
+{
+  // These all start at 1 (0 mask is bogus) and include the full mask (any)
+  for (CORBA::ULong is = 1; is <= MAX_SAMPLE_STATE_MASK; ++is) {
+    for (CORBA::ULong iv = 1; iv <= MAX_VIEW_STATE_MASK; ++iv) {
+      for (CORBA::ULong ii = 1; ii <= MAX_INSTANCE_STATE_MASK; ++ii) {
+        combined_state_lookup_[to_combined_states(is, iv, ii)] = HandleSet();
+      }
+    }
+  }
+  // catch-all for "bogus" lookups
+  combined_state_lookup_[0] = HandleSet();
+}
+
+void DataReaderImpl::update_lookup_maps(const typename SubscriptionInstanceMapType::iterator& input)
+{
+  for (LookupMap::iterator it = combined_state_lookup_.begin(); it != combined_state_lookup_.end(); ++it) {
+    if (it->first == 0) continue;
+    CORBA::ULong sample_states, view_states, instance_states;
+    split_combined_states(it->first, sample_states, view_states, instance_states);
+    if (input->second->matches(sample_states, view_states, instance_states)) {
+      it->second.insert(input->first);
+    } else {
+      it->second.erase(input->first);
+    }
+  }
+}
+
+void DataReaderImpl::remove_from_lookup_maps(DDS::InstanceHandle_t handle)
+{
+  for (LookupMap::iterator it = combined_state_lookup_.begin(), the_end = combined_state_lookup_.end(); it != the_end; ++it) {
+    if (it->first == 0) continue;
+    it->second.erase(handle);
+  }
+}
+
+const DataReaderImpl::HandleSet& DataReaderImpl::lookup_matching_instances(CORBA::ULong sample_states, CORBA::ULong view_states, CORBA::ULong instance_states) const
+{
+  const CORBA::ULong combined_states = to_combined_states(sample_states, view_states, instance_states);
+  LookupMap::const_iterator ci = combined_state_lookup_.find(combined_states);
+  OPENDDS_ASSERT(ci != combined_state_lookup_.end());
+  return ci->second;
 }
 
 } // namespace DCPS
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -2058,6 +2058,11 @@ DataReaderImpl::release_instance(DDS::InstanceHandle_t handle)
   }
 }
 
+void
+DataReaderImpl::state_updated(DDS::InstanceHandle_t handle)
+{
+  state_updated_i(handle);
+}
 
 OpenDDS::DCPS::WriterStats::WriterStats(
     int amount,
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -3590,7 +3590,7 @@ void DataReaderImpl::initialize_lookup_maps()
   combined_state_lookup_[0] = HandleSet();
 }
 
-void DataReaderImpl::update_lookup_maps(const typename SubscriptionInstanceMapType::iterator& input)
+void DataReaderImpl::update_lookup_maps(const SubscriptionInstanceMapType::iterator& input)
 {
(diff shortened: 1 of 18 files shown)


dds/DCPS/DataReaderImpl_T.h
```diff
@@ -54,6 +54,9 @@ namespace OpenDDS {
 
     typedef OPENDDS_MAP_CMP_T(MessageType, DDS::InstanceHandle_t,
                               typename TraitsType::LessThanType) InstanceMap;
+    typedef OPENDDS_MAP(DDS::InstanceHandle_t, typename InstanceMap::iterator) ReverseInstanceMap;
```
mitza-oci, line 57: Does this need `_MAP_T`?

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1085,6 +1077,73 @@ namespace OpenDDS {
 
 protected:
 
+  // Update max flag if the spec ever changes
+  static const CORBA::ULong MAX_VIEW_STATE_FLAG = DDS::NOT_NEW_VIEW_STATE;
+  static const CORBA::ULong MAX_VIEW_STATE_MASK = (MAX_VIEW_STATE_FLAG << 1) - 1;
+
+  // Update max flag if the spec ever changes
+  static const CORBA::ULong MAX_INSTANCE_STATE_FLAG = DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE;
+  static const CORBA::ULong MAX_INSTANCE_STATE_MASK = (MAX_INSTANCE_STATE_FLAG << 1) - 1;
```
mitza-oci, line 1086: Consider moving these, along with HandleSet and LookupMap, to DataReaderImpl (they don't need to be in the template).

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1085,6 +1077,73 @@ namespace OpenDDS {
...
+    }
+  }
+
+  void update_lookup_maps(const typename SubscriptionInstanceMapType::iterator& input)
+  {
+    for (LookupMap::iterator it = view_state_lookup_.begin(), the_end = view_state_lookup_.end(); it != the_end; ++it) {
+      if (input->second->instance_state_->match(it->first, DDS::ANY_INSTANCE_STATE)) {
+        it->second.insert(input->first);
+      }
+      else {
```
mitza-oci, line 1105:
```suggestion
      } else {
```

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -1085,6 +1077,73 @@ namespace OpenDDS {
...
+        it->second.erase(input->first);
+      }
+    }
+
+    for (LookupMap::iterator it = instance_state_lookup_.begin(), the_end = instance_state_lookup_.end(); it != the_end; ++it) {
+      if (input->second->instance_state_->match(DDS::ANY_VIEW_STATE, it->first)) {
+        it->second.insert(input->first);
+      }
+      else {
```
mitza-oci, line 1114:
```suggestion
      } else {
```

dds/DCPS/ReceivedDataElementList.inl
```diff
@@ -39,31 +50,66 @@ OpenDDS::DCPS::ReceivedDataElementList::add(ReceivedDataElement *data_sample)
 }
 
 ACE_INLINE
-OpenDDS::DCPS::ReceivedDataElement *
-OpenDDS::DCPS::ReceivedDataElementList::remove_head()
+void
+OpenDDS::DCPS::ReceivedDataElementList::add_by_timestamp(ReceivedDataElement *data_sample)
```
mitza-oci, line 54: This is probably too large for the `*.inl`

dds/DCPS/ReceivedDataElementList.cpp
```diff
@@ -131,8 +147,138 @@ OpenDDS::DCPS::ReceivedDataElementList::remove(
...
+{
+  return sample_states_ & sample_states;
+}
+
+OpenDDS::DCPS::ReceivedDataElement*
+OpenDDS::DCPS::ReceivedDataElementList::get_next_match(CORBA::ULong sample_states, ReceivedDataElement* prev)
+{
+  OPENDDS_ASSERT(sanity_check(prev));
+  if (prev == tail_) {
+    return NULL;
```
mitza-oci:
```suggestion
    return 0;
```
a few more of these below

dds/DCPS/ReceivedDataElementList.cpp
```diff
@@ -131,8 +147,138 @@ OpenDDS::DCPS::ReceivedDataElementList::remove(
...
+  }
+  ReceivedDataElement* item;
+  if (prev == NULL) {
+    item = head_;
+  } else {
+    item = prev->next_data_sample_;
+  }
```
mitza-oci:
```suggestion
  ReceivedDataElement* item = prev ? prev->next_data_sample_ : head_;
```

dds/DCPS/DataReaderImpl.h
```diff
@@ -434,6 +436,9 @@ class OpenDDS_Dcps_Export DataReaderImpl
   /// Release the instance with the handle.
   void release_instance(DDS::InstanceHandle_t handle);
 
+  // Take appropriate actions upon learning instance or view state has been updated
+  void state_updated(DDS::InstanceHandle_t handle);
```
jrw972, line 440: Given the current way DataReaderImpl and associates is implemented, this is the most straightforward way to implement this functionality.

However, it also points out why the design/implementation is bad.  A thread starting in DataReaderImpl or DataReaderImpl_T accesses SubscriptionInstance (which lacks encapsulation) which dispatches to InstanceState which dispatches back to DataReaderImpl.  Hence, we have indirect recursion.  This results in a number of undesirable things.  First, it bloats the public interface of the DataReader.  Alternatively, one must add friends which is similarly bad.  Second, it makes the code difficult to understand, e.g., if the recursion happens while iterating over a data structure, will the iterators become invalid?  Third, it makes the code difficult to understand with respect to concurrency. This leads to comments like "Why is this lock recursive?"

jrw972 (approved):

dds/DCPS/DataReaderImpl.h
```diff
@@ -594,6 +599,55 @@ class OpenDDS_Dcps_Export DataReaderImpl
...
+
+  // Update max flag if the spec ever changes
+  static const CORBA::ULong MAX_INSTANCE_STATE_FLAG = DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE;
+  static const CORBA::ULong MAX_INSTANCE_STATE_MASK = (MAX_INSTANCE_STATE_FLAG << 1) - 1;
+  static const CORBA::ULong MAX_INSTANCE_STATE_BITS = 3u;
+
+  // These may need to be updated if the spec ever changes
+  static const CORBA::ULong COMBINED_VIEW_STATE_SHIFT = MAX_INSTANCE_STATE_BITS;
+  static const CORBA::ULong COMBINED_SAMPLE_STATE_SHIFT = COMBINED_VIEW_STATE_SHIFT + MAX_VIEW_STATE_BITS;
+  static const CORBA::ULong MAX_COMBINED_STATE_MASK = (MAX_SAMPLE_STATE_FLAG << (COMBINED_SAMPLE_STATE_SHIFT + MAX_SAMPLE_STATE_BITS)) - 1;
```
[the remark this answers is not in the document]
simpsont-oci: Good catch. Since it's using `MAX_SAMPLE_STATE_FLAG` it doesn't need to factor in `MAX_SAMPLE_STATE_BITS`, so it's too large. But we don't end up using this constant anyway (I originally thought we'd need it) so I'll just remove it.

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3632,6 +3576,49 @@ void DataReaderImpl::OnDataAvailable::execute()
...
+void DataReaderImpl::initialize_lookup_maps()
+{
+  // These all start at 1 (0 mask is bogus) and include the full mask (any)
+  for (CORBA::ULong is = 1; is <= MAX_SAMPLE_STATE_MASK; ++is) {
+    for (CORBA::ULong iv = 1; iv <= MAX_VIEW_STATE_MASK; ++iv) {
+      for (CORBA::ULong ii = 1; ii <= MAX_INSTANCE_STATE_MASK; ++ii) {
+        combined_state_lookup_[to_combined_states(is, iv, ii)] = HandleSet();
+      }
+    }
+  }
```
jrw972, line 3588:
```suggestion
  for (CORBA::ULong is = 1; is <= MAX_SAMPLE_STATE_MASK; ++is) {
    for (CORBA::ULong iv = 1; iv <= MAX_VIEW_STATE_MASK; iv <<= 1) {
      for (CORBA::ULong ii = 1; ii <= MAX_INSTANCE_STATE_MASK; <<= 1) {
        combined_state_lookup_[to_combined_states(is, iv, ii)] = HandleSet();
      }
    }
  }
```

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3632,6 +3576,49 @@ void DataReaderImpl::OnDataAvailable::execute()
...
+  for (LookupMap::iterator it = combined_state_lookup_.begin(); it != combined_state_lookup_.end(); ++it) {
+    if (it->first == 0) continue;
+    CORBA::ULong sample_states, view_states, instance_states;
+    split_combined_states(it->first, sample_states, view_states, instance_states);
+    if (input->second->matches(sample_states, view_states, instance_states)) {
+      it->second.insert(input->first);
+    } else {
+      it->second.erase(input->first);
+    }
+  }
```
jrw972, line 3604: While this is "constant time", it should be not depend on the length of `combined_state_lookup_`.

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -738,23 +685,21 @@ namespace OpenDDS {
     const bool filter_has_non_key_fields =
       evaluator.has_non_key_fields(getMetaStruct<MessageType>());
 
-    for (SubscriptionInstanceMapType::iterator iter = instances_.begin(), end = instances_.end(); iter != end; ++iter) {
-      SubscriptionInstance& inst = *iter->second;
+    const HandleSet& matches = lookup_matching_instances(sample_states, view_states, instance_states);
+    for (HandleSet::const_iterator it = matches.begin(); it != matches.end(); ++it) {
```
jrw972, line 689: Can the state of the instance change in this loop?  If so, it may invalidate the iterators.

jrw972 (changes requested):

dds/DCPS/DataReaderImpl.cpp
simpsont-oci: I don't think this will work. My understanding is that the combined lookup states need to cover any valid mask combinations a user can throw at us. We can disregard anything with 0 sample state, 0 view state, or 0 instance state, since those are invalid requests, but that leaves 63 (3 * 3 * 7) lists + 1 empty list (for invalid / bogus requests, currently mapped to combined state of '0') that get maintained in this map.

jrw972: Yeah.  This is bogus.

jrw972 (approved):

PR #3349, opened 2022-02-21
Fix safety profile TimeBasedFilter test.

jonesc-oci:
The first sample is received after a delay in
safety profile builds. Discard the first sample
to avoid timing issues that cause test failures.
Update log messages.

### DIFF
--- a/tests/DCPS/TimeBasedFilter/main.cpp
+++ b/tests/DCPS/TimeBasedFilter/main.cpp
@@ -72,7 +72,9 @@ public:
   , condition_(lock_)
   , num_samples_(0)
   , expected_num_samples_(expected_num_samples)
+#ifdef OPENDDS_SAFETY_PROFILE
   , first_sample_(true)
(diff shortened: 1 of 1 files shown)


tests/DCPS/TimeBasedFilter/main.cpp
```diff
@@ -576,6 +586,19 @@ ACE_TMAIN(int argc, ACE_TCHAR** argv)
 
     SampleMap send_map;
 
+    // receiving the first safety profile sample takes longer than
+    // subsequent samples, so write a throwaway
```
sonndinh, line 590: Just for my own understanding, why does the first sample take longer than the others?

jonesc-oci: I checked with Adam and think it is due to RTPS discovery which is only used in safety profile.

sonndinh (approved):

PR #3354, opened 2022-02-23
Add Sanitizers to configure Script and CMake

iguessthislldo:


configure
```diff
@@ -2147,9 +2210,9 @@ sub write_environment {
   print $SE "$specific{'comment'} OpenDDS configure script: $0 $args_\n";
   print $MK "$specific{'comment'} OpenDDS configure script: $0 $args_\n" if $MK;
 
-  for my $key (keys %buildEnv) {
+  for my $key (sort(keys(%buildEnv))) {
     if ($key ne 'build' && $key ne 'mpctype') {
-      print $SE "$specific{'export'} $key=$buildEnv{$key}\n";
+      print $SE "$specific{'export'} $key=\"$buildEnv{$key}\"\n";
```
mitza-oci, line 2215: Is this quoting going to work in all cases?  I don't remember if `%buildEnv` can come from the user's command-line options.  If so, could it already have a quote? Does the makefile below `$MK` also need quoting?

configure
```diff
@@ -191,6 +191,9 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'inline!', 'Inlining (yes)',
     'static!', 'Static libraries (no)',
     'ipv6!', 'IPv6 support (no)',
+    'asan!', 'Address Sanitizer (no)',
+    'tsan!', 'Thread Sanitizer (no)',
+    'ubsan!', 'Undefined Behavior Sanitizer (no)',
```
mitza-oci, line 196: Should we make these a single option `--sanitize=foo`?

configure
iguessthislldo: > Is this quoting going to work in all cases? 

Maybe not, as this might be causing the missing Perl on the Windows builds.

>I don't remember if `%buildEnv` can come from the user's command-line options. If so, could it already have a quote?

I don't think they should already be quoted, but I probably have to look into that more.

> Does the makefile below` $MK` also need quoting?

Makefiles apparently doesn't need quoting:

```shell
$ echo -n "var=Hello goodbye\n\$(info \$(var))\n" | make -f -
Hello goodbye
make: *** No targets.  Stop.
```

configure
iguessthislldo: If the sanitizers were all mutably exclusive I would say yes, but ubsan can be safely combined with asan as [encouraged when using libfuzzer](https://llvm.org/docs/LibFuzzer.html#fuzzer-usage)

mitza-oci: The options processor supports multi-value.  I don't remember if it's `--sanitize=a --sanitize=b` or `--sanitize=a,b` or either of those.

iguessthislldo: Yes, I can do that. I think it's the first one that's builtin, but we could make work like the second one if we really wanted.

configure
iguessthislldo: > I don't remember if %buildEnv can come from the user's command-line options. If so, could it already have a quote?

Okay I played around with this a little. In batch script I don't think the behavior will break anything if I do this correctly using `set "NAME=VALUE"` as I just pushed. It seems to ignore all quotes expect for the last one if the name starts with a quote.

```batch
>set foo="bar
>echo %foo%
"bar
>set "foo="bar"
>echo %foo%
"bar

>set foo="bar""
>echo %foo%
"bar""
>set "foo="bar"""
>echo %foo%
"bar""
```

I think spaces already work, but quoting it allows it to contain special characters like `&`.

```cmd
>set foo= bar 
>echo {%foo%}
{ bar }

>set foo=bar&
>echo "%foo%"
"bar"
>set "foo=bar&"
>echo "%foo%"
"bar&"
```

Of course shell script is different. In the commit I just I pushed it's removing removing surrounding double quotes and escaping any other double quotes.

etc/ubsan-suppr.txt
```diff
@@ -0,0 +1,5 @@
+# Undefined behavior sanitizer suppressions
+# See here for how this works:
+#   https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
+
+pointer-overflow:be_visitor_operation_upcall_command_ss::gen_upcall
```
mitza-oci, line 5: Does this mean that `tao_idl` has undefined behavior? Do we need a new Issue in that repo?

iguessthislldo: Technically yes:

https://github.com/DOCGroup/ACE_TAO/blob/466542bd7d6d5b23783163fe420c2ea986734177/TAO/TAO_IDL/be/be_visitor_operation/upcall_command_ss.cpp#L223-L225

If `strstr` returns null, then it's doing pointer arithmetic with null, which for some reason is under the ubsan category called `pointer-overflow`. I didn't think it was worth fixing, at least not yet as it seems harmless because it's only doing a comparison. Also there's probably a lot more stuff to fix at runtime so I think we can defer fixing this until we decide to properly tackle undefined behavior.

PR #3373, opened 2022-03-01
Fix run-out-of-space error on some GitHub Actions jobs

sonndinh:


ClaytonCalabrese:
Looks like the w22 build timed out. Any idea what might have happened?

sonndinh:
> Looks like the w22 build timed out. Any idea what might have happened?

I don't have an idea yet. We may try reducing the number of tests run in this job.

ClaytonCalabrese:
Do we want to try re-enabling the cmake builds or were they what was causing the problem?

mitza-oci:
There are some test failures in `test_w22_x86_i0_sec` -- should that be disabled for now?

sonndinh:
> Do we want to try re-enabling the cmake builds or were they what was causing the problem?

Yes, we'll have a follow-up task to fix the CMake & messenger tests that failed. Adam and I chatted a bit about this and we're good with getting this in first so we can get rid of the disk space limitation issue.

sonndinh:
> There are some test failures in `test_w22_x86_i0_sec` -- should that be disabled for now?

Yes, I'm good with disabling it for now. I wanted to let it run one more time and some tests failed in a similar way to the ones from CMake and messenger tests. Once we fix these test fails, it'd be good to have this test job for more test coverage on CI for Windows. Currently, all other Windows build configurations only run a small subset of tests (CMake, messenger, compiler, unit).

PR #3374, opened 2022-03-02
Fix #3268 opendds_idl Segfault

### LINKED ISSUE #3268: 'opendds_idl' crashes with C++11 when sequence contains typedef to primitive type
Invoking opendds_idl with parameters "-Lc++11 DDS_.idl" does not produce any output due to crash.

Reason is in file "dds/idl/marshal_generator.cpp" line 2590:

          AST_PredefinedType* const pt = dynamic_cast<AST_PredefinedType*>(base);
          if (pt->pt() == AST_PredefinedType::PT_octet) {
            octetSeqOnly = field->local_name()->get_string();
          }

dynamic_cast<> fails with null_ptr.
[DDS_.idl.txt](https://github.com/objectcomputing/OpenDDS/files/7862838/DDS_.idl.txt)

iguessthislldo:
Fixes https://github.com/objectcomputing/OpenDDS/issues/3268. The check for a struct with just a octet sequence member added to support ROS didn't resolve the sequence base type. Fixed this and added a compiler test for this.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -509,7 +509,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ${{ github.job }}.tar.xz
-        key: c05_${{ github.job }}_ace6tao2_${{ env.ACE_COMMIT }}_${{ env.COMPILER_VERSION }}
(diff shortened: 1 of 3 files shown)


sonndinh (commented):
There is a compile error in the CI. Looks like it's from the anonymous sequence member.

jrw972 (approved):

PR #3375, opened 2022-03-03
Last Received Address Hinting Using Discovery Info

simpsont-oci:
Problem: For newly associated RTPS readers and writers, excess traffic may be generated until inbound traffic is received to determine a viable "last received" address. For readers and writers with many locators, the amount of redundant traffic can be significant.

Solution: When possible, use addresses / locators from discovery to "hint" at the last received address on association and cut down the amount of initially-generated traffic.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
@@ -10,6 +10,8 @@
 #include "RtpsUdpSendStrategy.h"
 #include "RtpsUdpReceiveStrategy.h"
 
+#include <dds/OpenddsDcpsExtTypeSupportImpl.h>
+
 #include <dds/DCPS/AssociationData.h>
 #include <dds/DCPS/BuiltInTopicUtils.h>
 #include <dds/DCPS/LogAddr.h>
@@ -17,7 +19,6 @@
 #include <dds/DCPS/transport/framework/TransportClient.h>
 #include <dds/DCPS/transport/framework/TransportExceptions.h>
 #include <dds/DCPS/RTPS/BaseMessageUtils.h>
-
 #include <dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.h>
 
 #include <ace/CDR_Base.h>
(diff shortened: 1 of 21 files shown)


jrw972 (commented):
Discovery (SPDP/SEDP in this case) is providing a list of locators and an address where discovery packets are being received.  I was expecting a loop that scans the list of locators for a preferred locator which is the one that is most similar to the receive address.  This preferred locator is what should be used to initialize the last receive address.  Can you point me to this section of code?

Would you consider using the "Internal DDS" features to solve this problem?  Presumably, there is a "Last Recv Address" topic.  It is written by the RtpsUdpDataLink (and SpdpTransport).  I think it would probably be read by the RtpsUdpDataLink (or RtpsUdpTransport).  The "Last Recv Address" topic could be read when the locators are unpacked to find the preferred locator.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -305,11 +308,27 @@ RtpsUdpTransport::use_datalink(const RepoId& local_id,
...
+  if (discovery_locator.length()) {
+    AddrSet disco_uc_addrs, disco_mc_addrs;
+    bool disco_requires_inline_qos;
+    unsigned int disco_blob_bytes_read;
+    get_connection_addrs(discovery_locator, &disco_uc_addrs, &disco_mc_addrs, &disco_requires_inline_qos, &disco_blob_bytes_read);
+
+    for (AddrSet::const_iterator it = disco_uc_addrs.begin(); it != disco_uc_addrs.end(); ++it) {
+      for (AddrSet::const_iterator it2 = uc_addrs.begin(); it2 != uc_addrs.end(); ++it2) {
+        if (it->addr_bytes_equal(*it2) && DCPS::is_more_local(disco_addr_hint, *it2)) {
+          disco_addr_hint = *it2;
```
simpsont-oci, line 321: @jrw972 This is the loop. Currently, the discovery_locator blob should only have one address in it, but I left it open to accept multiple in case we ever wanted to pick from multiple.

jrw972: Thanks.

jrw972 (approved):

PR #3385, opened 2022-03-07
Dump pool and hostname in ShmemInst::dump_to_str

jwillemsen:
* dds/DCPS/transport/shmem/ShmemInst.cpp:


jwillemsen:
Removed also some redundant newlines in the output

PR #3386, opened 2022-03-07
Make it possible to override the host_name for the shared memory tran…

jwillemsen:
…sport

    * dds/DCPS/transport/shmem/ShmemInst.cpp:


mitza-oci (approved):
Looks like this needs to be resolved before merging:
```
test_u20_p1_asan_sec

tests/cmake/install/CMakeLists.txt run_test.pl clean-env
/home/runner/work/OpenDDS/OpenDDS/OpenDDS/tests/cmake/build/install/the-install-prefix/bin/opendds_install_test_user: error while loading shared libraries: libTAO_PortableServer.so: cannot open shared object file: No such file or directory
```

jwillemsen:
@mitza-oci Do you add this new flag to the DevGuide?

PR #3390, opened 2022-03-08
Defer ReactorInterceptor Execution If Queue Is Non-Empty

simpsont-oci:
Problem: ReactorInterceptor::execute_or_enqueue may currently wind up executing "unplanned" code if commands already exist in the command queue. As a result, any code that may even have a remote chance of calling execute_or_enqueue needs to release all held mutexes, or else the "unplanned" code may double-lock a held mutex.

Solution: Since this isn't really practical, one simple solution is to dis-allow "immediate" execution by the reactor thread when commands already exist in the command queue. This prevents running "unknown" code in-situ, and so hopefully the calling context knows enough about what will get called in order to release all mutexes which might cause deadlock.

### DIFF
--- a/dds/DCPS/ReactorInterceptor.cpp
+++ b/dds/DCPS/ReactorInterceptor.cpp
@@ -21,7 +21,6 @@ namespace DCPS {
 
 ReactorInterceptor::Command::Command()
   : executed_(false)
-  , on_queue_(false)
   , condition_(mutex_)
   , reactor_(0)
 {}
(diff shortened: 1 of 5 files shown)


dds/DCPS/ReactorInterceptor.cpp
```diff
@@ -49,8 +49,25 @@ ReactorInterceptor::~ReactorInterceptor()
 
 bool ReactorInterceptor::should_execute_immediately()
 {
-  return ACE_OS::thr_equal(owner_, ACE_Thread::self()) ||
-    reactor_is_shut_down();
+  bool is_safe_to_execute = false;
```
jrw972, line 52: Go further.  Manually in-line all of the functions in `execute_or_enqueue` and then simplify the logic.

jrw972 (changes requested):

jrw972 (approved):

PR #3392, opened 2022-03-08
Fix Parsing, Check Paths in `show_build_config.pl`

iguessthislldo:
- `show_build_config.pl` parses the setenv files to get ACE_ROOT, but I added quotes to setenv values in https://github.com/objectcomputing/OpenDDS/pull/3354 This kept this from getting the correct ACE_ROOT, so it wouldn't show any files from ACE. This fixes it by adding additional quoted regex.
- Check to see if ACE_ROOT and the files exist so the script doesn't silently accept the previously described situation.
- Make it so the script no longer has to always be run from the root of the source directory.


tools/scripts/show_build_config.pl
```diff
@@ -36,23 +58,25 @@ sub show_config {
...
   show_config('setenv.cmd');
 }
-elsif (-r 'build/host/setenv.sh') {
+elsif (-r "$this_opendds_root/build/host/setenv.sh") {
   print '>' x 16 . ' Host Build ' . '<' x 16 . "\n";
-  show_config('build/host/setenv.sh');
+  show_config("$this_opendds_root/build/host/setenv.sh");
   print '>' x 16 . ' Target Build ' . '<' x 16 . "\n";
-  show_config('build/target/setenv.sh');
+  show_config("$this_opendds_root/build/build/target/setenv.sh");
```
mitza-oci, line 75:
```suggestion
  show_config("$this_opendds_root/build/target/setenv.sh");
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #3396, opened 2022-03-11
GitHub Actions cleanup: m10 tests and wireshark build

mitza-oci:


simpsont-oci (approved):

.github/workflows/build_and_test.yml
```diff
@@ -8279,7 +8279,7 @@ jobs:
       uses: lukka/run-vcpkg@v7
       with:
         vcpkgDirectory: '${{ github.workspace }}/vcpkg-qt'
-        vcpkgGitCommitId: 70033dbb31527fb3e69654731f540f59c87787f9
+        vcpkgGitCommitId: b86c0c35b88e2bf3557ff49dc831689c2f085090
```
simpsont-oci, line 8282: Should we keep this consistent between jobs that depend on wireshark_w19-release? Or maybe just everywhere?

mitza-oci: There should be an empty set intersection between the packages installed in `....\vcpkg` and those in `....\vcpkg-qt`.

PR #3401, opened 2022-03-14
GitHub Actions: default.features needed in ACE_TAO_u18_stat_qt_ws_sec

mitza-oci:
OpenDDS's configure when given an already-configured ACE+TAO will not
edit the default.features.  We need default.features to have the expected
features in order to generate makefiles for specific tests later on.

This means that the generated ACE_TAO artifact is biased for how we're
going to use it in OpenDDS, but it's only used once so this should be fine.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4587,7 +4587,6 @@ jobs:
       shell: bash
       run: |
         cd OpenDDS/ACE_TAO
-        perl -ni.bak -e "print unless /opendds/" ACE/bin/MakeProjectCreator/config/default.features # remove opendds features from here, let individual build configure scripts handle those
         find . -iname "*\.o" | xargs rm
         tar cvf ../../${{ github.job }}.tar ACE/ace/config.h
         git clean -xdfn | cut -d ' ' -f 3- | xargs tar uvf ../../${{ github.job }}.tar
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -4587,7 +4587,6 @@ jobs:
       shell: bash
       run: |
         cd OpenDDS/ACE_TAO
-        perl -ni.bak -e "print unless /opendds/" ACE/bin/MakeProjectCreator/config/default.features # remove opendds features from here, let individual build configure scripts handle those
```
simpsont-oci, line 4590: It seems like it would make more sense to tweak the downloaded ACE_TAO artifact's `default.features` to include the desired opendds features before running configure, since theoretically more than one build job could use this ACE_TAO artifact.

Since it's just one build, I agree it's not a big deal at the moment, but you'll need to make sure you've added the features to the configure line for the ACE_TAO call to configure (perhaps they are already there) and then bump the cache key index value.

sonndinh: Many ACE_TAO jobs are using this line to remove features with `opendds` in their names. So we may want to check that each OpenDDS build job has all the intended features which can be missing in `default.features`.

simpsont-oci (approved):

sonndinh (approved):

PR #3407, opened 2022-03-16
Use openssl 3 for one Linux build

jrw972:


.github/workflows/build_and_test.yml
```diff
@@ -3822,6 +3855,8 @@ jobs:
     - name: extract OpenDDS artifact
       shell: bash
       run: |
+        SSL_DIR=$PWD/openssl_install
+        export SSL_LIBDIR=$SSL_DIR/lib64
```
iguessthislldo, line 3859: I don't think this does anything.

```suggestion
```

iguessthislldo (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -3632,7 +3632,18 @@ jobs:
 
     runs-on: ubuntu-20.04
 
+    needs: [openssl3_u20]
```
mitza-oci, line 3635:
```suggestion
    needs: openssl3_u20
```

.github/workflows/build_and_test.yml
```diff
@@ -11256,3 +11291,40 @@ jobs:
...
+        key: c01_${{ github.job }}_${{ env.OPENSSL_COMMIT }}
+    - name: openssl 3.0.2
+      if: steps.cache-artifact.outputs.cache-hit != 'true'
+      run: |
+        PREFIX=$PWD/openssl_install
+        echo PREFIX=$PREFIX
+        mkdir $PREFIX
+        cd openssl
+        ./Configure no-tests --prefix=$PREFIX
+        make install_sw
```
mitza-oci, line 11319: Would a `-j` option help speed this up?

.github/workflows/build_and_test.yml
jrw972: SSL_LIBDIR needs to be defined and exported when `configure` runs because it must be defined for mpc.

jrw972: Didn't realize I duplicated it.

mitza-oci:
`lint_build_and_test.pl` is failing with a false positive @iguessthislldo (also, the annotation for this error seems to be attaching it to the lint script and not the yaml)

.github/workflows/build_and_test.yml
mitza-oci: Was this the cause of...
```
mv: cannot stat 'apps/lib/libapps-lib-app_libctx.d.tmp': No such file or directory
make[1]: *** [Makefile:3857: apps/lib/libapps-lib-app_libctx.o] Error 1
```
https://github.com/objectcomputing/OpenDDS/runs/5975598293?check_suite_focus=true

Maybe the makefiles from OpenSSL don't work with parallel builds

iguessthislldo:
It's not a false positive. I'm actually not sure how it wasn't failing before. Also yea, that script doesn't produce annotations.

.github/workflows/build_and_test.yml
```diff
@@ -11256,3 +11289,40 @@ jobs:
...
+        git clone --branch openssl-3.0.2 --depth 1 https://github.com/openssl/openssl.git
+        cd openssl
+        echo "OPENSSL_COMMIT=$(git rev-parse HEAD)" >> $GITHUB_ENV
+    - name: Cache Artifact
+      id: cache-artifact
+      uses: actions/cache@v2
+      with:
+        path: ${{ github.job }}.tar.xz
+        key: c02_${{ github.job }}_${{ env.OPENSSL_COMMIT }}
+    - name: openssl 3.0.2
```
iguessthislldo, line 11309: To make the script happy:

```suggestion
    - name: Build openssl 3.0.2
```

iguessthislldo (changes requested):

.github/workflows/build_and_test.yml
jrw972: @iguessthislldo What does the current lint failure mean?

iguessthislldo: Ah, I didn't think about it wanting a problem matching. The script a bit wonky here because of that and without `-j` it's matching `make install`, which doesn't count as a build command as far it's concerned. Now that I think about it, the script probably should just get an exception for OpenSSL like it [already has for Google Test and ACE/TAO](https://github.com/objectcomputing/OpenDDS/blob/9c284b4c8c8ae5a10d14f5824306cad12dcb4040/.github/workflows/lint_build_and_test.pl#L61-L67). Locally I removed the problem matcher and added:

```
      elsif ($step->{name} =~ /build openssl/i) {
        $avoids_problem_matcher = 1;
      }
```

to the script and the script ran successfully.

iguessthislldo (approved):

PR #3408, opened 2022-03-16
Set ACE_TEST_LOG_STUCK_STACKS=1 for GitHub Actions

jonesc-oci:


.github/workflows/build_and_test.yml
```diff
@@ -271,6 +271,7 @@ jobs:
       shell: bash
       run: |
         cd OpenDDS
+        echo "export ACE_TEST_LOG_STUCK_STACKS=1" >> setenv.sh
```
mitza-oci, line 274: I think you'd get the same results without quotes, but I'm OK with it.

mitza-oci (approved):

PR #3415, opened 2022-03-22
Fix 2 Anon. `sequence<string>` in the same `struct`

iguessthislldo:
Defining two anonymous `sequence<string>` members in the same struct causes a redefinition compile error in the type support. This happens because the anonymous type code relies on the pointer to the `AST_Decl` of the base type to tell if the type has already been used. The problem with that is apparently there can be multiple objects for string types.

This replaces the pointer with the classification of the container type and the classification and name of the base type.

### DIFF
--- a/dds/idl/field_info.cpp
+++ b/dds/idl/field_info.cpp
@@ -8,28 +8,41 @@
 #include "field_info.h"
 
 #include <sstream>
+#include <cstring>
 
 using namespace AstTypeClassification;
 
-FieldInfo::EleLen::EleLen(const FieldInfo& af)
-  : cls_(af.cls_)
-  , len_(af.n_elems_)
+FieldInfo::EleLen::EleLen(AST_Type* type)
+  : cls_(classify(type))
+  , len_(container_element_limit(type))
 {
-  AST_Type* const base = resolveActualType(af.as_base_);
-  base_cls_ = classify(base);
+  AST_Type* const base = resolveActualType(container_base_type(type));
+  const Classification base_cls = classify(base);
+  base_cls_ = base_cls;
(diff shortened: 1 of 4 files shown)


dds/idl/field_info.cpp
```diff
@@ -8,18 +8,41 @@
...
 bool FieldInfo::EleLen::operator<(const EleLen& o) const
 {
-  return ele_ < o.ele_ || (ele_ == o.ele_ && len_ < o.len_);
+  if (cls_ != o.cls_) {
+    return cls_ < o.cls_;
+  }
+  if (len_ != o.len_) {
+    return len_ < o.len_;
+  }
+  if (bool(base_container_.get()) != bool(o.base_container_.get())) {
```
mitza-oci, line 36: I'm not sure if ACE's smart pointer template is the same way, but with the standard library one couldn't we leave out `.get()` and just have the explicit conversion to bool?

dds/idl/field_info.cpp
```diff
@@ -8,18 +8,41 @@
...
+    base_container_ = Container(new EleLen(base));
+  }
 }
 
 bool FieldInfo::EleLen::operator<(const EleLen& o) const
 {
-  return ele_ < o.ele_ || (ele_ == o.ele_ && len_ < o.len_);
+  if (cls_ != o.cls_) {
+    return cls_ < o.cls_;
+  }
```
mitza-oci, line 32: Is it sufficient to compare based on the classification of the element type instead of the actual element type?

mitza-oci (approved):

ClaytonCalabrese (approved):

PR #3416, opened 2022-03-22
Return signal number in PerlDDS

sonndinh:


bin/PerlDDS/Process.pm
```diff
@@ -63,9 +63,9 @@ sub WaitKill ($)
 {
   my $self = shift;
   my $timeout = shift;
-  my $ret_value = $self->SUPER::WaitKill($timeout);
+  my @ret_values = $self->SUPER::WaitKill($timeout);
   $self->return_coverage_process();
-  return $ret_value;
+  return @ret_values;
```
iguessthislldo: This changes the result for the existing calls to this function as forcing a array into a scalar takes the last element. This is the issue I was talking about yesterday. For example:

```perl
sub WaitKill_scalar {
  return 1;
}

sub WaitKill_array {
  return 1, 0;
}

my $s = WaitKill_scalar();
print($s, "\n"); # 1
my $a = WaitKill_array();
print($a, "\n"); # 0
```

What I suggested was to do something like this:

```perl
sub WaitKill {
  my $self = shift;
  my $timeout = shift;
  my $opts = shift // {};
  my $signal_ref = $opts->{signal_ref};

  if ($signal_ref) {
    ${$signal_ref} = 10;
  }
}

my $signal;
WaitKill(undef, 20, {signal_ref => \$signal});
print($signal, "\n");

```

sonndinh: Yes, I realized that and used `wantarray` function to decide if the caller needs an array or a scalar.

iguessthislldo: Okay, looks like you did that shortly after or maybe even right before I commented.

bin/PerlDDS/Run_Test.pm
```diff
@@ -798,14 +805,24 @@ sub stop_process {
...
+                                    $timed_wait,
+                                    $name,
+                                    $self->{test_verbose});
+    $kill_status = $ret_array[0];
+  } else {
+    $kill_status = PerlDDS::wait_kill($self->{processes}->{process}->{$name}->{process},
+                                      $timed_wait,
+                                      $name,
+                                      $self->{test_verbose});
+  }
```
iguessthislldo, line 821: I'm not sure about this now that I see it. It seems like it would be cleaner to go with the hash ref method I showed in the other comment. That way we wouldn't have to double everything up.

iguessthislldo (changes requested):

bin/PerlDDS/Run_Test.pm
```diff
@@ -92,25 +92,36 @@ sub formatted_time {
...
   my $time_str = formatted_time();
   if ($result != 0) {
-      my $ext = ($verbose ? "" : "(started waiting for termination at $start_time)");
-      print STDERR "$time_str: ERROR: $desc returned $result $ext\n";
-      $ret_status = 1;
+    my $ext = ($verbose ? "(started waiting for termination at $start_time)" : "");
+    if ($result == -1) {
+      print STDERR "$time_str: ERROR: $name timedout $ext\n";
+    } elsif ($result == 255) {
+      print STDERR "$time_str: ERROR: $name terminated with a signal $ext\n";
```
iguessthislldo, line 116: It'd be great if we could print the signal. It seems to be possible to get the name from the number too: https://stackoverflow.com/questions/29862178/name-of-signal-number-2

iguessthislldo (changes requested):

bin/PerlDDS/Run_Test.pm
sonndinh: The signal name should already be printed by PerlACE's`check_return_value` function that is called by this function. Here, we're just kind of summarizing the result with a timestamp.

iguessthislldo (approved):

PR #3417, opened 2022-03-22
Fix Potential WaitSet Concurrent Wait Issues, Add Unit Test

simpsont-oci:
Problem: WaitSet was using two distinct approaches to atomic operations, which allowed for a potential race condition when multiple threads tried to call WaitSet::wait() on the same WaitSet object (which is intended to be prohibited by the spec).

Solution: Remove use of std::atomic / ACE_Atomic_Op and place `waiting_` under the protection of the object's mutex. In addition, add a real unit test for WaitSet and include testing of the new functionality as part of the unit test.

### DIFF
--- a/dds/DCPS/WaitSet.cpp
+++ b/dds/DCPS/WaitSet.cpp
@@ -114,10 +114,6 @@ ReturnCode_t WaitSet::wait(ConditionSeq& active_conditions,
 {
   using namespace OpenDDS::DCPS;
 
-  if (waiting_ != false) {
-    return RETCODE_PRECONDITION_NOT_MET;
-  }
-
   if (!non_negative_duration(timeout)) {
     return DDS::RETCODE_BAD_PARAMETER;
   }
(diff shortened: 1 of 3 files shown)


dds/DCPS/WaitSet.cpp
```diff
@@ -130,6 +126,11 @@ ReturnCode_t WaitSet::wait(ConditionSeq& active_conditions,
 
   ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, g, lock_,
                    RETCODE_OUT_OF_RESOURCES);
+
+  if (waiting_ != false) {
```
iguessthislldo, line 130:
```suggestion
  if (waiting_) {
```

iguessthislldo (changes requested):

tests/unit-tests/dds/DCPS/WaitSet.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+#ifdef ACE_HAS_CPP11
+TEST(dds_DCPS_WaitSet, WaitForever)
+{
+  DDS::WaitSet_var ws = new DDS::WaitSet;
+
+  DDS::GuardCondition_var gc = new DDS::GuardCondition;
+  EXPECT_EQ(ws->attach_condition(gc), DDS::RETCODE_OK);
+
+  std::thread thread([&](){
```
ClaytonCalabrese, line 121: Is std::thread going to cause issues on certain builds?

ClaytonCalabrese: Never mind, I'm dumb and you already blocked properly for CPP11.

sonndinh (approved):

iguessthislldo (approved):

PR #3424, opened 2022-03-25
Modernize Messenger Test

ClaytonCalabrese:
Previously, the messenger test was blocking on the publisher until all messages had been acked using `wait_for_acknowledgement`. This redesign was intended to remove this and change to have the subscriber going away be the signal for the publisher to go away. Additionally, unreliable readers were expecting to receive all 40 messages or else they would report a failure, which does not fit with the definition of unreliable transports. As such, it has been updated to exit upon receiving 1 of the 40 messages, and exit afterwards. This will prove that the transport works and communication is possible, which is the goal of the test.


tests/Utils/TestUtils.mpc
```diff
@@ -1,3 +1,9 @@
-project: dcps_test {
+project: dcps_test, install {
```
iguessthislldo, line 1: I don't think we should install anything. We probably shouldn't be installing things in `tests` at all and if we really did, it would probably better belong in `dds` or something else that gets installed. Also `tests/Utils` isn't a path we should install things to as it doesn't follow Unix directory conventions (like those in the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), though we don't follow that directly).

I couldn't think of the name at the time, but I was thinking of [`MPC/config/test_root.mpb`](https://github.com/objectcomputing/OpenDDS/blob/master/MPC/config/test_root.mpb) earlier in the meeting. When we go to build tests using an installed OpenDDS, `$TEST_ROOT` can be set to reference things from the source tree. Part of the problem seems to be that the test utils don't have a mpb file where this would be setup centrally. See [`MPC/config/dcps_test_framework.mpb`](https://github.com/objectcomputing/OpenDDS/blob/master/MPC/config/dcps_test_framework.mpb) for an example of this.

tests/DCPS/Messenger/DataReaderListener.cpp
```diff
@@ -91,81 +102,100 @@ void DataReaderListenerImpl::on_data_available(DDS::DataReader_ptr reader)
...
+          ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) DEBUG: DataReaderListenerImpl::on_data_available:"
+                    " instance is unregistered\n")));
+        }
       } else {
-        ACE_ERROR((LM_ERROR,
-                   ACE_TEXT("%N:%l: on_data_available()")
-                   ACE_TEXT(" ERROR: unknown instance state: %d\n"),
-                   si.instance_state));
+        if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Error) {
+          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: DataReaderListenerImpl::on_data_available:"
```
iguessthislldo, line 116:
```suggestion
          ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: DataReaderListenerImpl::on_data_available: "
```

iguessthislldo (changes requested):

tests/Utils/TestUtils.mpc
mitza-oci: I'd go further than "I don't think we should install..." -- we can't install these files, at least not without major changes which would seem to be out of scope for this PR.

Many existing tests use StatusMatching.h and build without errors against an installed tree.  What is different in those tests?

If you need to, you can generate makefiles with a `-base` command line argument.

iguessthislldo: In the tick install build the mpc command is `<command name="generate_workspace" options="-type gnuace DDS.mwc -features no_cxx11=0,no_opendds_security=0,no_rapidjson=0 -value_project libout=. -value_project exeout=. -value_project includes+=/tao_builds/jenkins/workspace/dds_doc_ace6tao2_tick_linux_gcc_d1o0/inst"/>`

iguessthislldo: So it appears it's using `-value_project includes+=/tao_builds/jenkins/workspace/dds_doc_ace6tao2_tick_linux_gcc_d1o0/inst"` to force the include on everything. I don't think that's what we should do though, the test utils need a proper mpb.

mitza-oci: This is by design.  When testing make install it allows the tests to include the headers they need.

iguessthislldo: Yes, I can see that, but it's a brute forcing it when there's an existing finer-grained mechanism that only hands out the include to projects that need it. Brute forcing also could hide issues with using headers that are missing in the install because the headers would always be in the include path. I suggest we make a test utils mpb just for messenger for now, set `$TEST_ROOT` in the CI make install test, and make a task for the future to convert the rest of the test suite and scoreboard to follow suite.

mitza-oci: The headers that are installed cannot also be found locally since they are removed from the "inst" build tree.

java/tests/messenger/subscriber/DataReaderListenerImpl.java
```diff
@@ -29,6 +33,16 @@ private void initialize_counts() {
         }
     }
 
+    public void set_expected_count(int expected)
+    {
```
mitza-oci, line 37:
```suggestion
    public void set_expected_count(int expected) {
```

java/tests/messenger/subscriber/DataReaderListenerImpl.java
```diff
@@ -29,6 +33,16 @@ private void initialize_counts() {
...
+        expected_count = expected;
+    }
+
+    public void set_guard_condition(GuardCondition guard_cond)
+    {
```
mitza-oci, line 42:
```suggestion
    public void set_guard_condition(GuardCondition guard_cond) {
```

java/tests/messenger/subscriber/DataReaderListenerImpl.java
```diff
@@ -143,7 +161,7 @@ public void report_validity() {
             if (val == false)
                 ++missed_counts;
         }
-        if (missed_counts > 0) {
+        if (missed_counts > 40 - expected_count) {
```
mitza-oci, line 164:
```suggestion
        if (missed_counts > N_EXPECTED - expected_count) {
```

mitza-oci (changes requested):

tests/Utils/TestUtils.mpc
mitza-oci: There is a separately compiled TestUtils library, but StatusMatching.h (while formally belonging to TestUtils) can be used on its own as header-only.  The separately compiled TestUtils is only used in two tests: Instances and PersistentInfoRepo.

mitza-oci (approved):

iguessthislldo (approved):

.github/workflows/build_and_test.yml
```diff
@@ -562,7 +562,9 @@ jobs:
         perl -ni.bak -e "print unless /opendds/" ACE/bin/MakeProjectCreator/config/default.features # remove opendds features from here, let individual build configure scripts handle those
         cd $GITHUB_WORKSPACE/OpenDDS
         find . -iname "*\.o" | xargs rm
-        rm build/target/setenv.sh
+        cd build/target
+        rm -rf !("ACE_TAO")
+        cd $GITHUB_WORKSPACE/OpenDDS
         git clean -xdfn | cut -d ' ' -f 3- | xargs tar cvf ../${{ github.job }}.tar
```
sonndinh, line 568: I agree with what we suggested in the meeting -- we probably just need to run tar with `build/target/ACE_TAO` and `build/host/ACE_TAO` and remove the `git clean` line to create the archive.

mitza-oci: Since there are a lot of symlinks in the build directories, you could try using `find` to filter them out:
```
find build/host/ACE_TAO build/target/ACE_TAO -type f | xargs tar .....
```

PR #3432, opened 2022-03-28
PropertySeq: check for invalid length

mitza-oci:
Similar to PR #3405 but in a different RTPS special-case encoding


tests/unit-tests/dds/DdsDcpsCoreTypeSupportImpl.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <dds/DdsDcpsCoreTypeSupportImpl.h>
+
+#include <dds/DCPS/Serializer.h>
+
+#include <gtest/gtest.h>
+
+using namespace OpenDDS::DCPS;
+
+
+TEST(DdsDcpsCoreTypeSupportImpl, InvalidPropertySeq)
```
jrw972, line 16: Use canonical name so the test gets picked up by the coverage report.

jrw972 (changes requested):

tests/unit-tests/dds/DdsDcpsCoreTypeSupportImpl.cpp
jrw972, line 16:
```suggestion
TEST(dds_DdsDcpsCoreTypeSupportImpl, InvalidPropertySeq)
```

jrw972 (changes requested):

jrw972 (approved):

PR #3434, opened 2022-03-29
Lock SPDP During Shutdown Checks To Avoid NULL Transport SEGV in process_participant_ice()

simpsont-oci:
Problem: Shutdown is clearing the `tport_` member of spdp in the middle of process_participant_ice() because we're not guaranteeing mutual exclusion during / after checking shutdown flags.

Solution: Don't do that. Lock to check, and recheck when we release / reacquire the lock.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -1080,7 +1080,10 @@ Spdp::data_received(const DataSubmessage& data,
   const bool relay_in_use = (config_->rtps_relay_only() || config_->use_rtps_relay());
   const bool from_relay = relay_in_use && (from == config_->spdp_rtps_relay_address());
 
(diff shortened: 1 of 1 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -1063,7 +1069,11 @@ Spdp::data_received(const DataSubmessage& data,
     }
     return;
   }
-  if (!is_security_enabled()) {
+
+  const bool security_enabled = is_security_enabled();
+  guard.release();
```
mitza-oci, line 1074: This line is conditional on `OPENDDS_SECURITY`.  Was that intended?

simpsont-oci: Nope... will take a look.

jrw972 (approved):

simpsont-oci:
If we didn't temporarily release the lock, which would require some design changes, we could remove a number of the checks needed for when we reacquire.

PR #3440, opened 2022-03-31
More Thrasher Cleanup, CI Coverage Expansion

simpsont-oci:
This PR expands some of the test coverage for Thrasher to more builds and makes the test able to run (with different assumptions) under the no ownership profile builds. It also makes the test resilient against association timeout errors which are problematic when association times start to exceed 30 seconds (expected when the VM is overloaded).

### DIFF
--- a/tests/DCPS/Thrasher/Subscriber.h
+++ b/tests/DCPS/Thrasher/Subscriber.h
@@ -14,7 +14,6 @@ public:
 private:
   void cleanup();
   const DDS::DomainId_t domainId_;
-  const size_t n_pub_threads_;
   const size_t expected_samples_;
   const bool durable_;
   DDS::DomainParticipantFactory_var dpf_;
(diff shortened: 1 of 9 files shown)


tests/DCPS/Thrasher/Subscriber.h
```diff
@@ -8,14 +8,14 @@
...
-  Subscriber(const DDS::DomainId_t domainId, const std::size_t n_pub_threads, const std::size_t expected_samples, const bool durable);
+  Subscriber(DDS::DomainId_t domainId, size_t n_pub_threads, size_t expected_samples, bool durable);
   ~Subscriber();
   int wait_and_check_received() const;
 private:
   void cleanup();
   const DDS::DomainId_t domainId_;
-  const std::size_t n_pub_threads_;
-  const std::size_t expected_samples_;
+  const size_t n_pub_threads_;
```
mitza-oci, line 17:
```suggestion
```
Warnings from CI builds telling us it's not used

mitza-oci (approved):

PR #3457, opened 2022-04-13
GitHub Actions: Re-enable MacOS testing

ClaytonCalabrese:
This PR re-enables macos10 testing

It had been previously disabled due to timeouts: #3396 

This PR will also attempt to re-enable macos11 CI.


docs/internal/github_actions.rst
```diff
@@ -139,7 +140,7 @@ Test runs which only contain CMake tests are prefixed by ``cmake_``.
 
 .lst files contain a list of tests with configuration options that will turn tests on or off.
 The *test_* jobs pass in :ghfile:`tests/dcps_tests.lst`.
-Static and Release builds instead use :ghfile:`tests/static_ci_tests.lst`.
+MacOS, Windows 22, Static, and Release builds instead use :ghfile:`tests/static_ci_tests.lst`.
```
mitza-oci, line 143: Since `static_ci_tests.lst` is not only used for static builds, should it be renamed?

ClaytonCalabrese: `core_ci_tests.lst` sound okay?

mitza-oci: Yes

ClaytonCalabrese:
It's weird because on my local runs with just the m11 and m10, it passed all tests fine. 

https://github.com/ClaytonCalabrese/OpenDDS/actions/runs/2196905135

It failed this way twice in a row now on this PR.

simpsont-oci:
> It's weird because on my local runs with just the m11 and m10, it passed all tests fine.
> 
> https://github.com/ClaytonCalabrese/OpenDDS/actions/runs/2196905135
> 
> It failed this way twice in a row now on this PR.

Looking at the most recent failed build and the one on your fork, it looks like the build_m11_sec artifacts are very different sizes... 50ish MB vs 80ish MB. Any thoughts on why that might be happening? Maybe it's just not building everything?

simpsont-oci:
> > It's weird because on my local runs with just the m11 and m10, it passed all tests fine.
> > https://github.com/ClaytonCalabrese/OpenDDS/actions/runs/2196905135
> > It failed this way twice in a row now on this PR.
> 
> Looking at the most recent failed build and the one on your fork, it looks like the build_m11_sec artifacts are very different sizes... 60ish MB vs 80ish MB. Any thoughts on why that might be happening? Maybe it's just not building everything?

Looking into this more, the unpacked size seems about the same, just a big difference in compression I guess.

ClaytonCalabrese:
It now fails the same way on my local branch. I think we should leave it commented and I'll create a ticket for follow up work.

simpsont-oci (changes requested):
Looks good, but let's clean up the mac build warnings

.github/workflows/build_and_test.yml
```diff
@@ -10291,444 +10291,937 @@ jobs:
...
+  #     shell: bash
+  #     run: |
+  #       cd OpenDDS
+  #       . setenv.sh
+  #       tools/scripts/show_build_config.pl
+  #   - name: disable ntp
+  #     shell: bash
+  #     run: |
+  #       sudo systemsetup -setusingnetworktime off
+  #       sudo rm -rf /etc/ntp.conf
```
sonndinh, line 10503: What happens if NTP is enabled?

.github/workflows/build_and_test.yml
```diff
@@ -10291,444 +10291,937 @@ jobs:
...
+        </configuration>
+        <command name="log" options="on"/>
+        <command name="print_os_version"/>
+        <command name="print_env_vars"/>
+        <command name="print_ace_config"/>
+        <command name="print_make_version"/>
+        <command name="check_compiler" options="gcc"/>
+        <command name="print_perl_version"/>
+        <command name="print_autobuild_config"/>
+        <command name="auto_run_tests" options="script_path=$ACE_ROOT/bin dir=$GITHUB_WORKSPACE/OpenDDS -l tests/core_ci_tests.lst -Config STATIC_MESSENGER -Config STATIC_COMPILER -Config STATIC_COMPILER2 -Config STATIC_UNIT_TESTS -Config CXX11 -Config RAPIDJSON -Config NO_SHMEM -Config DDS_NO_ORBSVCS -Config NO_BUILT_IN_TOPICS"/>
```
sonndinh, line 10840: I think this doesn't run the TAO Hello test and modeling test built from the build job.

.github/workflows/build_and_test.yml
```diff
@@ -10861,6 +11355,33 @@ jobs:
...
+        cd OpenDDS
+        . setenv.sh
+        mwc.pl -type gnuace -workers 4 $TAO_ROOT/tests/Hello
+    - name: build TAO Hello test
+      shell: bash
+      run: |
+        cd OpenDDS
+        . setenv.sh
+        make -j4 dir=$TAO_ROOT/tests/Hello
+    - name: set up Modeling tests
```
sonndinh, line 11370: The TAO Hello test and modeling tests are also built here but it seems to me they are not run in any test job.

.github/workflows/build_and_test.yml
```diff
@@ -10291,444 +10291,937 @@ jobs:
...
+        <command name="log" options="on"/>
+        <command name="print_os_version"/>
+        <command name="print_env_vars"/>
+        <command name="print_ace_config"/>
+        <command name="print_cmake_version"/>
+        <command name="print_make_version"/>
+        <command name="check_compiler" options="gcc"/>
+        <command name="print_perl_version"/>
+        <command name="print_autobuild_config"/>
+        <command name="auto_run_tests" options="script_path=$ACE_ROOT/bin dir=$GITHUB_WORKSPACE/OpenDDS -l tests/core_ci_tests.lst -Config STATIC_MESSENGER -Config STATIC_COMPILER -Config STATIC_COMPILER2 -Config STATIC_UNIT_TESTS -Config XERCES3 -Config OPENDDS_SECURITY -Config CXX11 -Config RAPIDJSON -Config NO_SHMEM -Config DDS_NO_ORBSVCS -Config GH_ACTIONS_M10"/>
```
sonndinh, line 11098: Should we consider running a broader set of tests than the ones from `core_ci_tests.lst` for this and other test jobs. I think the build jobs have built a much broader set of tests from `dcps_tests.lst` too.

.github/workflows/build_and_test.yml
ClaytonCalabrese: I'm not sure. I brought that change over from m10 builds because I thought it might also be a problem on m11.
https://github.com/objectcomputing/OpenDDS/pull/2904 is the PR in which it was added to m10.

.github/workflows/build_and_test.yml
ClaytonCalabrese: When I attempted running  dcps_tests.lst, it seemed like it was headed for a 6 hour timeout on at least one of the builds:
https://github.com/ClaytonCalabrese/OpenDDS/runs/6082702187?check_suite_focus=true.  Looking at it again, I suppose it is possible it just takes forever to run on m11 because the other m11 build finished after almost 3 hours: https://github.com/ClaytonCalabrese/OpenDDS/runs/6082631385?check_suite_focus=true.  I'll try again with dcps_tests.lst.

.github/workflows/build_and_test.yml
simpsont-oci: Some of the unit tests were failing on macos builds due to ACE timers not firing at the right times. There had been a number of GitHub issues surrounding time & timers in the the macos GitHub virtual environments, and this was one suggested change. It didn't wind up having much impact on our tests, so we can probably remove it now.

ClaytonCalabrese: Ok, will do.

.github/workflows/build_and_test.yml
ClaytonCalabrese: Do we ever run the TAO Hello test? Should I just remove it everywhere?

sonndinh: I haven't seen it in the non-macOS jobs. I agree if we're not running it, it should be removed.

.github/workflows/build_and_test.yml
ClaytonCalabrese: Looks like it still times out: 

https://github.com/ClaytonCalabrese/OpenDDS/runs/6134686486?check_suite_focus=true
https://github.com/ClaytonCalabrese/OpenDDS/runs/6134922703?check_suite_focus=true

simpsont-oci: Ideally we would figure out what tests were causing the timeout and add a Config option for GH_MACOS to prevent those tests from running in GitHub Actions, but I think we should view that as a followup task and not something to accomplish in this PR. The core suite is much better than what we had previously, which was nothing.

.github/workflows/build_and_test.yml
simpsont-oci: I'm OK with this. We might need a couple followup tasks to flesh out and fix MacOS testing in CI. This task is just to get _something_ back up and running.

.github/workflows/build_and_test.yml
simpsont-oci: @mitza-oci does checking the ability to compile the TAO Hello test add enough value to keep?

.github/workflows/build_and_test.yml
sonndinh: I agree. We can fix that and run more tests in a follow-up task.

.github/workflows/build_and_test.yml
```diff
@@ -10825,244 +11431,188 @@ jobs:
...
+        <command name="log" options="on"/>
+        <command name="print_os_version"/>
+        <command name="print_env_vars"/>
+        <command name="print_ace_config"/>
+        <command name="print_cmake_version"/>
+        <command name="print_make_version"/>
+        <command name="check_compiler" options="gcc"/>
+        <command name="print_perl_version"/>
+        <command name="print_autobuild_config"/>
+        <command name="auto_run_tests" options="script_path=tests dir=$GITHUB_WORKSPACE/OpenDDS --no-dcps -Config CXX11 -Config RAPIDJSON -Config NO_SHMEM -Config DDS_NO_ORBSVCS -Config NO_BUILT_IN_TOPICS -Config NO_UNIT_TESTS -Config GH_ACTIONS_M10 --modeling"/>
```
sonndinh, line 11593: I think this job can be combined with the cmake test job below, something like adding `--modeling` to the `<command name="auto_run_tests" ... />` line in the cmake test job.

sonndinh (approved):

PR #3458, opened 2022-04-13
Test & Fixes For Durable WaitForAck

simpsont-oci:
Problem: Previous `wait_for_acknowledgments` behavior was poorly defined, since any acknowledgment would be applied universally across all readers.

Solution: In general, wait_for_acknowledgments should delay until every currently associated reader has acknowledged the current sequence number at the time of the call to wait_for_acknowledgments.

### DIFF
--- a/dds/DCPS/WriteDataContainer.cpp
+++ b/dds/DCPS/WriteDataContainer.cpp
@@ -172,10 +172,15 @@ WriteDataContainer::~WriteDataContainer()
 }
 
 void
-WriteDataContainer::add_reader_acks(const RepoId& reader)
+WriteDataContainer::add_reader_acks(const RepoId& reader, const SequenceNumber& base)
 {
   ACE_Guard<ACE_Thread_Mutex> guard(wfa_lock_);
   acked_sequences_[reader].reset();
+  if (base == SequenceNumber::SEQUENCENUMBER_UNKNOWN()) {
+    acked_sequences_[reader].insert(SequenceNumber::ZERO());
+  } else {
+    acked_sequences_[reader].insert(SequenceRange(SequenceNumber(), base));
+  }
 }
 
 void
@@ -1554,7 +1559,7 @@ WriteDataContainer::wait_ack_of_seq(const MonotonicTimePoint& deadline,
 }
 
 bool
-WriteDataContainer::sequence_acknowledged(SequenceNumber sequence)
+WriteDataContainer::sequence_acknowledged(const SequenceNumber& sequence)
 {
   ACE_Guard<ACE_SYNCH_MUTEX> guard(wfa_lock_);
   return sequence_acknowledged_i(sequence);
--- a/dds/DCPS/WriteDataContainer.cpp
+++ b/dds/DCPS/WriteDataContainer.cpp
@@ -183,8 +183,8 @@ WriteDataContainer::remove_reader_acks(const RepoId& reader)
 {
   ACE_Guard<ACE_Thread_Mutex> guard(wfa_lock_);
 
-  SequenceNumber prev_cum_ack = get_cumulative_ack();
-  AckedSequenceMap::iterator it = acked_sequences_.find(reader);
+  const SequenceNumber prev_cum_ack = get_cumulative_ack();
+  const AckedSequenceMap::iterator it = acked_sequences_.find(reader);
   if (it != acked_sequences_.end()) {
     acked_sequences_.erase(it);
     if (prev_cum_ack != get_cumulative_ack()) {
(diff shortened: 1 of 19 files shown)


dds/DCPS/WriteDataContainer.h
```diff
@@ -407,11 +407,24 @@ class OpenDDS_Dcps_Export WriteDataContainer : public RcObject {
...
+  typedef OPENDDS_UNORDERED_MAP(RepoId, DisjointSequence) AckedSequenceMap;
+#else
+  typedef OPENDDS_MAP_CMP(RepoId, DisjointSequence, GUID_tKeyLessThan) AckedSequenceMap;
+#endif
+  AckedSequenceMap acked_sequences_;
+
+  SequenceNumber get_cumulative_ack();
+  SequenceNumber get_last_ack();
+  void update_acked(const SequenceNumber& seq, const RepoId& id = GUID_UNKNOWN);
+  bool sequence_acknowledged_i(const SequenceNumber sequence);
```
mitza-oci, line 427: Pass by reference or remove top-level const from the declaration

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -170,6 +171,89 @@ WriteDataContainer::~WriteDataContainer()
...
+  acked_sequences_[reader].reset();
+}
+
+void
+WriteDataContainer::remove_reader_acks(const RepoId& reader)
+{
+  ACE_Guard<ACE_Thread_Mutex> guard(wfa_lock_);
+
+  SequenceNumber prev_cum_ack = get_cumulative_ack();
+  AckedSequenceMap::iterator it = acked_sequences_.find(reader);
```
mitza-oci, line 187:
```suggestion
  const SequenceNumber prev_cum_ack = get_cumulative_ack();
  const AckedSequenceMap::iterator it = acked_sequences_.find(reader);
```

dds/DCPS/WriteDataContainer.cpp
```diff
@@ -228,6 +312,7 @@ WriteDataContainer::reenqueue_all(const RepoId& reader_id,
                    lock_,
                    DDS::RETCODE_ERROR);
 
+
```
mitza-oci, line 315:
```suggestion
```

tests/DCPS/WaitForAckRace/publisher.cpp
```diff
@@ -0,0 +1,256 @@
...
+  sigset_t mask, prev;
+  ACE_OS::sigemptyset(&mask);
+  ACE_OS::sigaddset(&mask, SIGPIPE);
+  ACE_OS::sigprocmask(SIG_BLOCK, &mask, &prev);
+
+  int total_readers = 1;
+
+  bool large_samples = false;
+  bool sub_delayed_exit = false;
+  for (int i = 0; i < argc; ++i) {
```
mitza-oci, line 77:
```suggestion
  for (int i = 1; i < argc; ++i) {
```

tests/DCPS/WaitForAckRace/publisher.cpp
```diff
@@ -0,0 +1,256 @@
...
+  try
+  {
+    DDS::DomainParticipantFactory_var dpf =
+      TheParticipantFactoryWithArgs(argc, argv);
+
+    DDS::DomainParticipant_var participant =
+      dpf->create_participant(31,
+                              PARTICIPANT_QOS_DEFAULT,
+                              DDS::DomainParticipantListener::_nil(),
+                              ::OpenDDS::DCPS::DEFAULT_STATUS_MASK);
```
mitza-oci, line 101:
```suggestion
                              0,
                              0);
```

It would be good find an example to work off of that doesn't have so much unnecessary cruft in it.  All the `.in()` and `is_nil`, `_nil()`, etc...
Otherwise the old patterns of usage just keep getting propagated

mitza-oci (changes requested):

jrw972 (changes requested):
The RtpsUdpDataLink does a similar thing with the leader/lagger concepts.  I think if you adopted a design where you were tracking in the same way, it would be more efficient.

mitza-oci:
> The RtpsUdpDataLink does a similar thing with the leader/lagger concepts. I think if you adopted a design where you were tracking in the same way, it would be more efficient.

Could this be refactored into a common component that tracks multiple DisjointSequences (by GUID) and knows the overall cumulative ack, etc?

simpsont-oci:
> > The RtpsUdpDataLink does a similar thing with the leader/lagger concepts. I think if you adopted a design where you were tracking in the same way, it would be more efficient.
> 
> Could this be refactored into a common component that tracks multiple DisjointSequences (by GUID) and knows the overall cumulative ack, etc?

I'm fine with that approach, but I think we should address it in a subsequent issue / PR so there will be scope for adding a unit test for the new component.

simpsont-oci:
Otherwise, I think I'm content with where this PR is at. The occasional Thrasher test failure (previous build attempts) might be resolved by #3440 (which also expands coverage) and the modeling tests will probably just need some refactoring to make the Sync logic make better assumptions.

jrw972 (approved):

mitza-oci (approved):

PR #3459, opened 2022-04-14
key information not passed to `begin_struct_member`

jrw972:
Problem
-------

ValueWriter implementations may treat members that are keys
differently.  As a concrete example, consider a ValueWriter that
prepares a SQL INSERT statement for a received sample.  Most likely,
this will be an UPSERT which often requires a list of keys and a list
of non-keys.

Solution
--------

Change `begin_struct_member` to take an XTypes::MemberDescriptor which
contains the member name and a bool for key status.  This PR only sets
these two members.

### DIFF
--- a/dds/DCPS/ValueWriter.h
+++ b/dds/DCPS/ValueWriter.h
@@ -7,6 +7,7 @@
 #define OPENDDS_DCPS_VALUE_WRITER_H
 
 #include "Definitions.h"
+#include "XTypes/MemberDescriptor.h"
 
 #include <dds/Versioned_Namespace.h>
 #include <FACE/Fixed.h>
@@ -41,7 +42,7 @@ public:
 
   virtual void begin_struct() {}
   virtual void end_struct() {}
-  virtual void begin_struct_member(const char* /*name*/) {}
+  virtual void begin_struct_member(const XTypes::MemberDescriptor& /*descriptor*/) {}
   virtual void end_struct_member() {}
 
   virtual void begin_union() {}
--- a/dds/DCPS/XTypes/MemberDescriptor.h
+++ b/dds/DCPS/XTypes/MemberDescriptor.h
@@ -30,6 +30,17 @@ class OpenDDS_Dcps_Export MemberDescriptor {
 public:
   MemberDescriptor();
   ~MemberDescriptor();
+  MemberDescriptor(const char* a_name, bool a_is_key)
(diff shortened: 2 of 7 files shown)


ClaytonCalabrese:
The unit tests need some namespace/include fixes

ClaytonCalabrese:
The unit tests need some namespace/include fixes

dds/DCPS/XTypes/MemberDescriptor.h
```diff
@@ -30,6 +30,10 @@ class OpenDDS_Dcps_Export MemberDescriptor {
 public:
   MemberDescriptor();
   ~MemberDescriptor();
+  MemberDescriptor(const char* a_name, bool a_is_key)
+    : name(a_name)
+    , is_key(a_is_key)
```
mitza-oci, line 35: Initialize all other non-static data members that need initialization

mitza-oci (changes requested):

dds/DCPS/ValueWriter.h
```diff
@@ -9,6 +9,7 @@
 #include "Definitions.h"
 
 #include <dds/Versioned_Namespace.h>
+#include "XTypes/MemberDescriptor.h"
 #include <FACE/Fixed.h>
```
iguessthislldo, line 13:
```suggestion
#include "Definitions.h"
#include "XTypes/MemberDescriptor.h"

#include <dds/Versioned_Namespace.h>
#include <FACE/Fixed.h>
```

iguessthislldo (changes requested):

dds/DCPS/ValueWriter.h
ClaytonCalabrese: Aren't `Versioned_Namespace.h`, `XTypes/MemberDescriptor.h`, and `FACE/Fixed.h` all in the "Headers from external OpenDDS-based libraries."?

mitza-oci (approved):

ClaytonCalabrese (approved):

iguessthislldo (approved):

dds/DCPS/ValueWriter.h
iguessthislldo: That's probably worded poorly, but by "external OpenDDS-based libraries" I meant things like Bench and other non-core libraries that use the core libraries. All of these headers are part of the core libraries.

PR #3460, opened 2022-04-14
NopDataReaderListener and NopDataWriterListener

iguessthislldo:
These are listener classes that already derive from `LocalObject` with NOP default implementations to simplify the task of implementing listeners.


iguessthislldo:
We could have opendds_idl generate these automatically on `interface`s we annotation with something like `@OpenDDS::internal::nop_impl("NopDataReaderListener")`.

PR #3465, opened 2022-04-19
Fix TSAN Issue In ConditionImpl.cpp

simpsont-oci:
Problem: Code was accessing protected data structure without mutex.

Solution: Don't do that. Just use the local copy.

### DIFF
--- a/dds/DCPS/ConditionImpl.cpp
+++ b/dds/DCPS/ConditionImpl.cpp
@@ -31,8 +31,7 @@ void ConditionImpl::signal_all()
   if (DCPS_debug_level > 9) {
     ACE_DEBUG((LM_DEBUG,
                ACE_TEXT("(%P|%t) ConditionImpl::signal_all() - ")
-               ACE_TEXT("number of sets: %d, locally: %d.\n"),
-               this->waitsets_.size(),
+               ACE_TEXT("number of sets: %d\n"),
                local_ws.size()));
   }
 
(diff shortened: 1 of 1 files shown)


dds/DCPS/ConditionImpl.cpp
```diff
@@ -31,8 +31,7 @@ void ConditionImpl::signal_all()
   if (DCPS_debug_level > 9) {
     ACE_DEBUG((LM_DEBUG,
                ACE_TEXT("(%P|%t) ConditionImpl::signal_all() - ")
-               ACE_TEXT("number of sets: %d, locally: %d.\n"),
-               this->waitsets_.size(),
+               ACE_TEXT("number of sets: %d\n"),
```
mitza-oci, line 34:
```suggestion
               ACE_TEXT("number of sets: %B\n"),
```

mitza-oci (changes requested):

dds/DCPS/ConditionImpl.cpp
mitza-oci: https://github.com/DOCGroup/ACE_TAO/blob/9a54764694aaf7c0830d916e85f3de8e90a252f9/ACE/ace/Log_Msg.h#L525

The value passed into the varargs is 64-bit (on platforms with 64-bit size_t, which should be pretty much everywhere), so when the vararg is consumed from the stack it should also consume a 64-bit value.

mitza-oci (approved):

PR #3466, opened 2022-04-19
Fix Coverity Issue In WaitForAckRace Test

simpsont-oci:
Problem: Coverity doesn't like that WaitForAckRace's publisher is using an atoi from a command line variable as a loop boundary.

Solution: Add additional bounds checking so Coverity can sleep at night knowing no fewer than 1 and no more than 99 readers can be specified.


mitza-oci:
> Problem: Coverity doesn't like that WaitForAckRace's publisher is using an atoi from a command line variable as a loop boundary.
> 
> Solution: Add additional bounds checking so Coverity can sleep at night knowing no fewer than 1 and no more than 99 readers can be specified.

I was planning on this being a false positive.  The command line arguments passed to tests are trusted, this is not a security boundary.

mitza-oci (approved):

PR #3467, opened 2022-04-20
Fix Another Potential Ownership SEGV Issue

simpsont-oci:
Problem: Captured another core file from SEGV in ownership code.

Solution: Pretty sure this function just needs to hold the mutex to prevent other threads from corrupting / invalidating map access.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -2648,6 +2648,7 @@ DataReaderImpl::ownership_filter_instance(const SubscriptionInstance_rch& instan
 #ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
   if (this->is_exclusive_ownership_) {
 
+    ACE_Guard<ACE_RW_Thread_Mutex> write_guard(writers_lock_);
     WriterMapType::iterator iter = writers_.find(pubid);
 
     if (iter == writers_.end()) {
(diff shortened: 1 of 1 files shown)


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -2648,6 +2648,7 @@ DataReaderImpl::ownership_filter_instance(const SubscriptionInstance_rch& instan
 #ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
   if (this->is_exclusive_ownership_) {
 
+    ACE_Guard<ACE_RW_Thread_Mutex> write_guard(writers_lock_);
```
mitza-oci, line 2651: Other uses of this lock use ACE_READ_GUARD / ACE_WRITE_GUARD.  I think we should stick with that for consistency and clarity when using the RW flavor of mutex.

simpsont-oci, line 2651:
```suggestion
    ACE_WRITE_GUARD(ACE_RW_Thread_Mutex, write_guard, writers_lock_);
```

jrw972 (approved):

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -2648,6 +2648,7 @@ DataReaderImpl::ownership_filter_instance(const SubscriptionInstance_rch& instan
 #ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
   if (this->is_exclusive_ownership_) {
 
+    ACE_WRITE_GUARD(ACE_RW_Thread_Mutex, write_guard, writers_lock_);
```
simpsont-oci, line 2651:
```suggestion
    ACE_WRITE_GUARD_RETURN(ACE_RW_Thread_Mutex, write_guard, writers_lock_, true);
```

PR #3470, opened 2022-04-21
Issue 3606 changes replacing GuidConverter to LogGuid

iamharvir:
Me & JoanneOrozco applied the changes to replace every GuidConverter instance with LogGuid as specified in issue 3606. In total we replaced 352 instances.


iguessthislldo (changes requested):
There are quite a bit of build issues that GitHub Actions found below. Please fix them. Most, if not all, should be listed on [this page](https://github.com/objectcomputing/OpenDDS/pull/3470/files) along with the code.

Also please remove these files it looks like were added accidentally: `tests/unit-tests/Run_Test.pm` and `tools/monitor/MonitorDataStorage.odt`.

simpsont-oci:
> There are quite a bit of build issues that GitHub Actions found below. Please fix them. Most, if not all, should be listed on [this page](https://github.com/objectcomputing/OpenDDS/pull/3470/files) along with the code.

Note that `LogGuid` is not a 1-to-1 replacement for `GuidConverter`. It's a handy inline replacement for `OPENDDS_STRING(DCPS::GuidConverter(X))`. If one of the logic methods on `GuidConverter` is being called (such as `isUserDomainEntity()`) it may not be possible to do the replacement.

Also, in places where a variable name is being declared (e.g. `GuidConverter writer_converter(id);`, either the entire declaration ought to be removed (so `LogGuid` can be used inline within the debug statement) or the variable name ought to be updated to reflect the fact that it's no longer a `GuidConverter` (e.g. `LogGuid writer_lg(id);`)

PR #3471, opened 2022-04-21
Print stack trace in PerlDDS

sonndinh:


bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
   my $process = $self->{processes}->{process}->{$name}->{process};
 
-  return PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+  my $result = PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+
+  # If opts has indicated that user wants to get core dump/signal information
+  # then find the core dump file if any and print its stack trace.
+  if (defined $opts && defined $opts->{dump_ref} && ${$opts->{dump_ref}}) {
+      # Get the core file pattern
+      my $core_pattern_file = "/proc/sys/kernel/core_pattern";
```
simpsont-oci, line 473: This is fancier than what I'd imagined (I imagined we'd either enforce a set pattern or put it into an environment variable). I like this better, though, assuming it works as expected. :)

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
+      if (!(-e $core_pattern_file)) {
```
simpsont-oci, line 474: I assume this just fails on Windows (which is expected / fine)? What about MacOS? Does it use this file as well, same syntax?

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
 
   my $process = $self->{processes}->{process}->{$name}->{process};
 
-  return PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+  my $result = PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+
+  # If opts has indicated that user wants to get core dump/signal information
+  # then find the core dump file if any and print its stack trace.
+  if (defined $opts && defined $opts->{dump_ref} && ${$opts->{dump_ref}}) {
```
simpsont-oci, line 471: So is this something we'd have to enable in each test? Is there a way to enable it globally (for example, with an environment variable for everything running on GitHub Actions?)

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
+      # Get the core file pattern
+      my $core_pattern_file = "/proc/sys/kernel/core_pattern";
+      if (!(-e $core_pattern_file)) {
+          printf STDOUT "WARNING: Core file pattern $core_pattern_file not exist\n";
```
simpsont-oci, line 475:
```suggestion
          printf STDOUT "WARNING: Core file pattern $core_pattern_file does not exist\n";
```

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
 
   my $process = $self->{processes}->{process}->{$name}->{process};
 
-  return PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+  my $result = PerlDDS::wait_kill($process, $wait_time, $name, $verbose, $opts);
+
+  # If opts has indicated that user wants to get core dump/signal information
+  # then find the core dump file if any and print its stack trace.
+  if (defined $opts && defined $opts->{dump_ref} && ${$opts->{dump_ref}}) {
+      # Get the core file pattern
```
simpsont-oci, line 472: Is the indent right for this block (and children)? It looks like it's 2 spaces elsewhere in this file.

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
+      }
+
+      if (!(-e $core_file_path)) {
+          printf STDOUT "WARNING: Core file $core_file_path not exist\n";
+          return $result;
+      }
+
+      # Print stack trace.
+      my $stack_trace;
+      if (system("gdb --version") != -1) {
```
simpsont-oci, line 577: If we have both, but we built with clang, should we prefer to use lldb? I'm not sure there's an easy way to check for this so maybe it's fine to just try gdb first.

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
+          } else {
+              printf STDOUT "WARNING: Could not determine a core file with timestamp\n";
+              return $result;
+          }
+      } else {
+          $core_file_path = $path . "/" . $pattern;
+      }
+
+      if (!(-e $core_file_path)) {
+          printf STDOUT "WARNING: Core file $core_file_path not exist\n";
```
simpsont-oci, line 571:
```suggestion
          printf STDOUT "WARNING: Core file $core_file_path does not exist\n";
```

bin/PerlDDS/Run_Test.pm
```diff
@@ -464,7 +464,132 @@ sub wait_kill {
...
+      }
+
+      my $timestamp_idx = index($pattern, "%t");
+      my $core_file_path;
+      if ($timestamp_idx != -1) {
+          my $prefix = substr($pattern, 0, $timestamp_idx);
+          my $suffix_len = length($pattern) - $timestamp_idx - 2;
+          my $suffix = substr($pattern, $timestamp_idx + 2, $suffix_len);
+
+          # Get the core file that has latest timestamp.
```
simpsont-oci, line 543:
```suggestion
          # Get the core file with latest timestamp.
```

bin/PerlDDS/Run_Test.pm
simpsont-oci, line 471: Also, does this need to check for SEGV or other error condition? I may be misunderstanding the Perl, but it looks like this will always run (even on success?).

simpsont-oci (changes requested):

bin/PerlDDS/Run_Test.pm
sonndinh: We can make it independent of what users set in `$opts`. I think we always want to print out the stack trace when a core dump is generated. Is there a case where we want to turn it off?

bin/PerlDDS/Run_Test.pm
sonndinh: Yeah, the indentation is not right. I'll fix it.

bin/PerlDDS/Run_Test.pm
sonndinh: So if there is a core dump, generated by SEGV or some other signal, the body of the `if` block will run. But as in the previous comment, we can make it independent of what user sets.

simpsont-oci: I'm wondering what will happen if we have a successful run but there happen to be old core files laying around and the pattern doesn't exclude them... for example if the pattern is just "core" or "core.%e" etc.

bin/PerlDDS/Run_Test.pm
sonndinh: The mechanism on macOS seems different; I think it doesn't have a core file pattern. I'll need to find out more about macOS and maybe Windows.

bin/PerlDDS/Run_Test.pm
sonndinh: `${$opts->{dump_ref}}` indicates if the process has failed and had a core dump. So it won't try finding a core dump if the process succeeded.

jwillemsen:
Is this something that also be worth for the ACE perl test framework, that way anyone using the ACE test framework could leverage this.

bin/PerlDDS/Run_Test.pm
```diff
@@ -455,6 +455,132 @@ sub new {
...
+    return;
+  }
+
+  # Print stack trace.
+  my $stack_trace;
+  if (system("gdb --version") != -1) {
+    $stack_trace = `gdb $exec_path -c $core_file_path -ex bt -ex quit`;
+  } elsif (system("lldb --version") != -1) {
+    printf STDOUT "WARNING: Failed printing stack trace with gdb. Trying lldb...\n";
+    $stack_trace = `lldb $exec_path -c $core_file_path -o bt -o quit`;
```
iguessthislldo, line 572: This looks like it will always use gdb as long as it's there. I think this is okay as a default, but it's preferable to use binary tools from the same toolchain as the compiler. So Linux builds using clang we should probably use lldb, even though these systems are also going to have gdb. There are also cases where the names could be different, like I have `lldb-13` on my personal system, but not a regular `lldb`.

So what I'm thinking is this should probably be configurable, I guess as environment variables, so that we could force lldb on Linux builds using clang or set a specific lldb or gdb to use.

bin/PerlDDS/Run_Test.pm
```diff
@@ -455,6 +455,132 @@ sub new {
...
+
+  my $timestamp_idx = index($pattern, "%t");
+  my $core_file_path;
+  if ($timestamp_idx != -1) {
+    my $prefix = substr($pattern, 0, $timestamp_idx);
+    my $suffix_len = length($pattern) - $timestamp_idx - 2;
+    my $suffix = substr($pattern, $timestamp_idx + 2, $suffix_len);
+
+    # Get the core file with latest timestamp.
+    opendir(my $dh, $path);
```
iguessthislldo, line 535: Probably should check if this succeeded and print why it failed:

```suggestion
    if (!opendir(my $dh, $path)) {
      print STDERR "WARNING: print_stacktrace: Couldn't opendir $path: $!\n";
    }
```

iguessthislldo (changes requested):

sonndinh:
> 

Yes, I'm working on a draft PR to ACE (https://github.com/DOCGroup/ACE_TAO/pull/1827).

bin/PerlDDS/Run_Test.pm
sonndinh: That makes sense. I made changes to the PR to ACE. Users can now set the debugger they want with an environment variable.

bin/PerlDDS/Run_Test.pm
sonndinh: I added support for macOS. It's in the PR to the ACE repo. The GitHub Actions workflow file will need updates to make it able to generate core files. I'll work on this later.

bin/PerlDDS/Run_Test.pm
sonndinh: Users can set the debugger they want now. Please see the PR to ACE.

bin/PerlDDS/Run_Test.pm
simpsont-oci: Should we make a follow-up Jira task to fix GitHub Actions for rmacOS?

sonndinh: Yes, we can do that. I'm still not able to make core files generated for macOS on GitHub Actions even though the same commands worked locally. I think it's better to fix this later, so it doesn't hold up the rest of the PR.

simpsont-oci (approved):

iguessthislldo (approved):

PR #3472, opened 2022-04-21
Fix Shutdown Leak In DCPSInfoRepo Under Load

simpsont-oci:
Problem: Under load, DCPSInfoRepo leaks some buffers from BIT writes.

Solution: prevent reinsertion of dropped / delivered buffers during shutdown & after disposal to avoid leaks

### DIFF
--- a/dds/DCPS/transport/tcp/TcpConnection.cpp
+++ b/dds/DCPS/transport/tcp/TcpConnection.cpp
@@ -635,7 +635,6 @@ OpenDDS::DCPS::TcpConnection::active_reconnect_i()
 void
 OpenDDS::DCPS::TcpConnection::notify_connection_lost()
 {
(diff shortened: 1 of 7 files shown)


dds/DCPS/transport/tcp/TcpConnection.cpp
```diff
@@ -618,29 +619,30 @@ OpenDDS::DCPS::TcpConnection::active_reconnect_i()
...
     this->handle_stop_reconnecting();
   }
 }
 
 void
 OpenDDS::DCPS::TcpConnection::notify_connection_lost()
 {
-  if (this->link_) {
-    this->link_->notify(DataLink::LOST);
+  ACE_DEBUG((LM_DEBUG, "(%P|%t) TcpConnection::notify_connection_lost()\n"));
```
simpsont-oci, line 638:
```suggestion
```

Issue #477, opened 2017-03-14
Qos_Helper should provide some logging why a QoS is invalid


jwillemsen:
It would help the user when the Qos_Helper would provide some logging why a QoS is invalid

jwillemsen:
Currently trying to convert some of our tests to OpenDDS using the QoS XML support. We have logging to dump the QoS settings at the moment it fails and now have a test where OpenDDS rejects our QoS, at the moment there would be logging in the QoS validation it would be much easier to find the conflicting setting.

```
[...]

PR #3475, opened 2022-04-22
Add logging for invalid Qos 

tmayoff:
I just copied what was done previously on #494 to each of these: Topic, Publisher, Subscriber, DataReader and DomainParticipantFactory. This should also close this, #477


dds/DCPS/Qos_Helper.inl
```diff
@@ -927,22 +927,87 @@ bool Qos_Helper::valid(const DDS::DomainParticipantQos& qos)
...
 bool Qos_Helper::valid(const DDS::TopicQos& qos)
 {
-  return
-    valid(qos.topic_data)
-    && valid(qos.durability)
+  if (!valid(qos.topic_data))
+    ACE_ERROR_RETURN ((LM_ERROR,
+        ACE_TEXT("(%P|%t) ERROR: ")
+        ACE_TEXT("Qos_Helper::valid::TopicQos, ")
+        ACE_TEXT("invalid topic_data qos.\n")), false);
```
ClaytonCalabrese, line 934:
```suggestion
  if (!valid(qos.topic_data)) {
    if (log_level >= LogLevel::Error) {
      ACE_ERROR((LM_ERROR,
        "(%P|%t) ERROR: Qos_Helper::valid: invalid topic_data qos.\n"));
    }
    return false;
  }
```

ACE_TEXT isn't needed.
ACE_ERROR_RETURN is something we are trying to avoid using because it hides return statements behind functions.
The log_level code helps with avoiding debug message spam.  You will need to add #include "debug.h" for this code.
And then a few formatting changes.

@iguessthislldo is invalid QoS a Notice or an Error?

ClaytonCalabrese (changes requested):

dds/DCPS/Qos_Helper.inl
tmayoff: Should I apply these changes to the whole file or just what I originally changed in the PR (and by that I mean the other logging that was added a while ago)

iguessthislldo: It should probably be a notice because it's reported back to the user through the error status and it's not fatal to the program or causes OpenDDS to function incorrectly, though the part about misconfiguration in the guidelines makes this a little muddled. [These are the guidelines we are referencing btw](https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#logging).

You don't have to apply these changes to the whole file if you don't want to. I hope to automate it someday and do it all at once, but in the mean time it's been something we've been trying to do incrementally for the whole code base for the most part.

dds/DCPS/Qos_Helper.inl
```diff
@@ -1047,46 +1112,163 @@ bool Qos_Helper::valid(const DDS::DataWriterQos& qos)
...
+        ACE_TEXT("Qos_Helper::valid::DataReaderQos, ")
+        ACE_TEXT("invalid ownership qos.\n")), false);
+
+  if (!valid(qos.representation))
+    ACE_ERROR_RETURN ((LM_ERROR,
+        ACE_TEXT("(%P|%t) ERROR: ")
+        ACE_TEXT("Qos_Helper::valid::DataReaderQos, ")
+        ACE_TEXT("invalid representation qos.\n")), false);
+
+    return true;
```
iguessthislldo, line 1229:
```suggestion
  return true;
```

dds/DCPS/Qos_Helper.inl
```diff
@@ -927,22 +927,87 @@ bool Qos_Helper::valid(const DDS::DomainParticipantQos& qos)
...
+        ACE_TEXT("Qos_Helper::valid::TopicQos, ")
+        ACE_TEXT("invalid ownership qos.\n")), false);
+
+  if (!valid(qos.representation))
+    ACE_ERROR_RETURN ((LM_ERROR,
+        ACE_TEXT("(%P|%t) ERROR: ")
+        ACE_TEXT("Qos_Helper::valid::TopicQos, ")
+        ACE_TEXT("invalid representation qos.\n")), false);
+
+    return true;
```
iguessthislldo, line 1010:
```suggestion
  return true;
```

iguessthislldo (changes requested):

dds/DCPS/Qos_Helper.inl
tmayoff: I'll change them to the be notices, and I've already started to change the whole file as it wasn't a huge deal. Also as for the formatting issues, is this the script to auto format it correctly, https://github.com/objectcomputing/OpenDDS/blob/master/tools/scripts/style/style.sh.

iguessthislldo: Okay, thanks. That script it hasn't been maintained in a while, so I can't really recommend using it as I'm not sure it follows our current practices. [`tools/scripts/lint.pl`](https://github.com/objectcomputing/OpenDDS/blob/master/tools/scripts/lint.pl) is maintained and runs as part of the PR checks (in the "lint" job). It can fix things like trailing spaces on lines if `--try-fix` is passed, but it's mostly just meant to report issues.

mitza-oci:
> Also in the original issue https://github.com/jwillemsen, was concerned about adding too much to the inline functions, is this still a problem?

Yes, functions that are sufficiently large should be in the `*.cpp` file to avoid increasing build times and code size with `inline=1`.

tmayoff:
For that is it enough to just copy past the functions into the `.cpp` and presumably remove the ACE_INLINE from the beginning of every function?

mitza-oci:
> For that is it enough to just copy past the functions into the `.cpp` and presumably remove the ACE_INLINE from the beginning of every function?

That's right.  Also look at the `#include`s at the top of the inl file and move any that are no longer needed there to the cpp.

simpsont-oci:
Have you generated dependencies? They're not generated by default, but necessary for rebuilding with make. Try running `make -j4 depend` and then rerunning your `make` command.

dds/DCPS/Qos_Helper.cpp
```diff
@@ -6,9 +6,477 @@
...
+      ACE_ERROR((LM_NOTICE,
+        "(%P|%t) ERROR: Qos_Helper::valid::DomainParticipantQos, invalid entity_factory qos.\n"));
+    }
+    return false;
+  }
+
+  if (!valid(qos.property)) {
+    if (log_level >= LogLevel::Notice) {
+      ACE_ERROR((LM_NOTICE,
+        "(%P|%t) ERROR: Qos_Helper::valid::DomainParticipantQos, invalid property qos.\n"));
```
ClaytonCalabrese, line 39:
```suggestion
        "(%P|%t) NOTICE: Qos_Helper::valid::DomainParticipantQos, invalid property qos.\n"));
```

dds/DCPS/Qos_Helper.cpp
```diff
@@ -6,9 +6,477 @@
...
 
+#include "Qos_Helper.h"
 #if !defined (__ACE_INLINE__)
 #include "Qos_Helper.inl"
 #endif /* __ACE_INLINE__ */
+
+#include "debug.h"
+#include "Service_Participant.h"
+
+bool Qos_Helper::valid(const DDS::DomainParticipantQos& qos)
```
iguessthislldo, line 18: Will need to put the declarations in the `OpenDDS::DCPS` namespace. This will need to include the version namespace macro. Here is an example:

https://github.com/objectcomputing/OpenDDS/blob/eba3661e705994915f7ca8486d2182509029d9b4/dds/DCPS/DCPS_Utils.cpp#L24-L27

https://github.com/objectcomputing/OpenDDS/blob/eba3661e705994915f7ca8486d2182509029d9b4/dds/DCPS/DCPS_Utils.cpp#L527-L530

iguessthislldo (changes requested):

tmayoff:
> Have you generated dependencies? They're not generated by default, but necessary for rebuilding with make. Try running `make -j4 depend` and then rerunning your `make` command.

Still not compiling everything (or anything it seems), I even copied the configure and make commands from the CI to make sure I'm getting them right

simpsont-oci:
Hmm, not sure. One more common gotcha is that rerunning configure often doesn't do what you might expect (it skips steps if it sees things are already partially configured). If you're familiar with MPC you can modify features and rerun mwc, but otherwise you may want to start from a "clean" repo (git clean -xdf (make sure you back up any untracked files you want to save)) and rerun configure / make.

tmayoff:
This might need a new issue, but I just tried again and only thing in the bin folder is opendds_idl, this is on Fedora 36 (where I write this PR), but I tried it on Debian 11 (what I use for work) and it worked fine.

ClaytonCalabrese (approved):

iguessthislldo (approved):

iguessthislldo:
Thanks for the contribution!

PR #3481, opened 2022-04-28
Messenger Test: Java and UDP Fixes

ClaytonCalabrese:
The previous PR for Java Messenger test fixes was enabling `DCPSPendingTimeout` within a conditional meaning it wasn't being used on the scoreboard. This lead to the errors still happening. This PR move it outside of the conditional.

Additionally, UDP on the normal messenger test failed at least once, due to missing every message. Prior to the rewrite, there were some microsleeps to spread out the writing of messages in an unreliable Data Writer. This PR adds that sleep back to the test.


mitza-oci (approved):

ClaytonCalabrese:
I'm guessing either no-bit or wchar is somehow breaking test_u18_w1_j_FM-2f. Once the build finishes I'll take a look at the full  logs and see if they tell us anything.

PR #3485, opened 2022-05-02
TSAN Fix and Minor Cleanup for Ownership Test DataReaderListener

simpsont-oci:
Problem: Different threads can call the Ownership DataReaderListener's on_data_available method, causing a TSAN complaint.

Solution: Add a mutex to the DataReaderListener to avoid TSAN complaints

### DIFF
--- a/tests/DCPS/Ownership/DataReaderListener.cpp
+++ b/tests/DCPS/Ownership/DataReaderListener.cpp
@@ -22,12 +22,12 @@ const int num_messages_per_writer = 20;
 
 DataReaderListenerImpl::DataReaderListenerImpl(const char* reader_id)
   : num_reads_(0),
-    reader_id_ (reader_id),
-    verify_result_ (true),
-    result_verify_complete_ (false)
+    reader_id_(reader_id),
+    verify_result_(true),
+    result_verify_complete_(false)
 {
-  this->current_strength_[0] = 0;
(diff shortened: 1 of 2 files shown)


tests/DCPS/Ownership/DataReaderListener.cpp
```diff
@@ -101,22 +101,25 @@ void DataReaderListenerImpl::on_requested_deadline_missed(
...
           ACE_TEXT(" handle %d total_count_change %d\n"),
-          this->reader_id_, status.last_instance_handle, status.total_count_change));
+          reader_id_, status.last_instance_handle, status.total_count_change));
 }
 
 void DataReaderListenerImpl::on_requested_incompatible_qos(
   DDS::DataReader_ptr,
   const DDS::RequestedIncompatibleQosStatus &)
 {
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
```
mitza-oci, line 114: Do this and the other no-op callbacks (on_subscription_matched, etc.) need to lock?  I guess it doesn't hurt

simpsont-oci: I included them in case we add code here eventually, but no technically these don't need locks. at the moment.

mitza-oci (approved):

PR #3487, opened 2022-05-02
Always Deliver Held Historic Samples

simpsont-oci:
Problem: When using wait_for_acknowledgments to determine delivery of samples, publishing applications which exit immediately may cause a race condition in associated subscribing applications between the delivery of held historic samples in WriterInfo::historic_samples_ and the disassociation caused by the exiting publisher.

Solution: In order to honor the assumption that all data "acknowledged" by the transport is delivered to the subscribing application, the DCPS layer needs to ensure that held historic samples are also guaranteed to be delivered to the subscribing application.

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -741,7 +741,7 @@ DataReaderImpl::signal_liveliness(const RepoId& remote_participant)
 
   ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, this->sample_lock_);
 
-  typedef std::pair<RepoId, WriterInfo_rch > RepoWriterPair;
+  typedef std::pair<RepoId, WriterInfo_rch> RepoWriterPair;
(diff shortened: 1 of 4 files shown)


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -738,7 +741,7 @@ DataReaderImpl::signal_liveliness(const RepoId& remote_participant)
 
   ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, this->sample_lock_);
 
-  typedef std::pair<RepoId, RcHandle<WriterInfo> > RepoWriterPair;
+  typedef std::pair<RepoId, WriterInfo_rch > RepoWriterPair;
```
simpsont-oci, line 744:
```suggestion
  typedef std::pair<RepoId, WriterInfo_rch> RepoWriterPair;
```

mitza-oci (approved):

PR #3493, opened 2022-05-06
Fix `UseXTypes` Deadlock and Added `inspect` Example

iguessthislldo:
- `UseXTypes` RTPS Discovery Option:
  - OpenDDS would deadlock if trying to use the string values `no`, `minimal`, or `complete` because the underlying string setter was trying to use the enum setter. Changed to avoid calling the other setter.
  - Reaction to invalid `UseXTypes` values would depend on if the value was a string or integer. Changed it to also get feedback from string setter.
- `inspect`:
  - Tweaked README and added an example.
  - Added a `--version` option and tweaked `--help` output.


tools/inspect/Inspect.cpp
```diff
@@ -120,10 +120,10 @@ bool writer_count = false;
 
 void print_usage(bool for_error = false)
 {
-  std::ostream& os = for_error ? std::cout : std::cerr;
+  std::ostream& os = for_error ? std::cerr : std::cout;
```
mitza-oci, line 123: I'm not sure what was intended originally. @ClaytonCalabrese ?

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -791,18 +791,23 @@ RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
...
+          bool valid;
           if (value.size() == 1) {
             int smInt;
-            if (!DCPS::convertToInteger(value, smInt) || smInt < 0 || smInt > 2) {
-              ACE_ERROR_RETURN((LM_ERROR,
-                                ACE_TEXT("(%P|%t) RtpsDiscovery::Config::discovery_config ")
-                                ACE_TEXT("Invalid entry (%C) for UseXTypes in ")
-                                ACE_TEXT("[rtps_discovery/%C] section.\n"),
-                                value.c_str(), rtps_name.c_str()), -1);
+            valid = DCPS::convertToInteger(value, smInt) && smInt >= 0 && smInt <= 2;
```
mitza-oci, line 797: Consider combining the `size() == 1` check above into this validity check

tools/inspect/Inspect.cpp
iguessthislldo: I wrote this part as part of #3189. Usage can be part of help, which is not an error, or part of a error for invalid arguments. I noticed I mixed this up.

mitza-oci (approved):

tools/inspect/README.rst
```diff
@@ -0,0 +1,107 @@
...
+This should result in ``inspect`` printing messages like these after a short period of time::
+
+    struct ::Messenger::Message
+      String8 from = "Comic Book Guy"
+      String8 subject = "Review"
+      Int32 subject_id = 99
+      String8 text = "Worst. Movie. Ever."
+      Int32 count = 0
+
+If the delays are not added to publisher or the timing of the added delays need to be increased, then it could just prints something like::
```
sonndinh, line 104:
```suggestion
If the delays are not added to publisher or the timing of the added delays need to be increased, then it could just print something like::
```

sonndinh (approved):

PR #3496, opened 2022-05-06
OpenDDS lacks a way to print IDL defined types (in a default build)

jrw972:
Problem
-------

Occassionally, it is useful to print out an IDL defined type.
Currently, OpenDDS only supports this when built with rapidjson.

Solution
--------

Define PrinterValueWriter.


dds/DCPS/PrinterValueWriter.h
```diff
@@ -0,0 +1,290 @@
...
+void PrinterValueWriter::write_string(const ACE_CDR::Char* value, size_t length)
+{
+  stream_ << '"' << std::string(value, length) << '"';
+}
+
+void PrinterValueWriter::write_wstring(const ACE_CDR::WChar* /*value*/, size_t /*length*/)
+{
+  // FUTURE
+  stream_ << "wide string";
+}
```
iguessthislldo, line 277: There are existing functions in [`ValueHelper.h`](https://github.com/objectcomputing/OpenDDS/blob/master/dds/DCPS/ValueHelper.h) for printing a representation all the character based types. These are the same functions `opendds_idl` and `print_dynamic_data`/`inspect` are using.

dds/idl/marshal_generator.cpp
```diff
@@ -3206,57 +3206,6 @@ bool marshal_generator::gen_struct(AST_Structure* node,
...
-        expr += " \"";
-      }
-      expr += ";\n"
-        "  }\n";
-      if (is_string_type) {
-        expr +=
-          "  strm.os() << '\"';\n";
-      }
-      expr += indent + streamCommon(
-        indent, field_name, fields[i]->field_type(), "<< stru", false, intro, cxx, true);
```
iguessthislldo, line 3246: There's also some printer support that's running through `streamCommon` to help it distinguish between outputting to `Serializer` and `ostream`. I think it's the last `true`. I don't know if we want to cover that in this PR.

iguessthislldo (changes requested):

dds/DCPS/PrinterValueWriter.h
```diff
@@ -0,0 +1,290 @@
...
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+/// Convert values to a readable format.
+class PrinterValueWriter : public ValueWriter {
+public:
+  static unsigned int indent;
```
mitza-oci, line 32: Is this supposed to be `const`? If not, how does it handle concurrency/reentrancy?

jrw972: No, it's a global default so it should only be changed once at start up.

mitza-oci: How about using `Service_Participant` to keep track of it like other `[common]` configurable values?

iguessthislldo (approved):

tests/unit-tests/dds/DCPS/PrinterValueWriter.cpp
```diff
@@ -0,0 +1,447 @@
...
+  pvw.begin_element(0);
+  pvw.write_int16(5);
+  pvw.end_element();
+  pvw.begin_element(1);
+  pvw.write_int16(6);
+  pvw.end_element();
+  pvw.end_array();
+  pvw.end_struct_member();
+  pvw.end_struct();
+  EXPECT_STREQ(pvw.str().c_str(), "    a: \n        [0]: 5\n        [1]: 6");
```
iguessthislldo, line 222: If you don't want to change this to avoid the trailing whitespace, this should make the lint script stop complaining about it.

```suggestion
  pvw.end_struct();
  // lint.pl ignores whitespace_before_newline_in_string on next line
  EXPECT_STREQ(pvw.str().c_str(), "    a: \n        [0]: 5\n        [1]: 6");
```

dds/DCPS/PrinterValueWriter.h
```diff
@@ -0,0 +1,283 @@
...
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+/// Convert values to a readable format.
+class PrinterValueWriter : public ValueWriter {
+public:
+  PrinterValueWriter(unsigned int a_indent = TheServiceParticipant->printer_value_writer_indent())
```
mitza-oci, line 33:
```suggestion
  explicit PrinterValueWriter(unsigned int a_indent = TheServiceParticipant->printer_value_writer_indent())
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #3498, opened 2022-05-10
Avoid non-id version of canceling reactor timers where possible

simpsont-oci:
Problem: The non-id version of ACE_Timer_Heap_T has linear performance over all registered timers. For small numbers of timers, it's not a big deal, but generally good to avoid when possible by using the id version, which does an O(log(n)) lookup.

Solution: Store timer id when it makes sense to do so for faster cancels.

### DIFF
--- a/dds/DCPS/SporadicTask.h
+++ b/dds/DCPS/SporadicTask.h
@@ -25,7 +25,7 @@ public:
     : time_source_(time_source)
     , interceptor_(interceptor)
     , desired_scheduled_(false)
-    , actual_scheduled_(false)
+    , timer_id_(0)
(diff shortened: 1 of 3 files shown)


dds/DCPS/SporadicTask.h
```diff
@@ -25,7 +25,7 @@ class SporadicTask : public RcEventHandler {
     : time_source_(time_source)
     , interceptor_(interceptor)
     , desired_scheduled_(false)
-    , actual_scheduled_(false)
+    , timer_id_(0)
```
jrw972, line 28: Do timer ids start at 1?

simpsont-oci: Nope... I was thinking of the timers in EventDispatcher.

PR #3501, opened 2022-05-11
Update extracting hostname to handle more general cases

sonndinh:
Hostname is extracted from an address string passed to `choose_single_coherent_address` using `substr` function. The second argument passed to `substr` seems to be the end index of the substring but it should be the number of characters in the substring. The current code works as long as the open bracket is at index 0 (`openb` is 0), but doesn't work in other cases, like `"    [::1234:abcd]:8080"`. This updates the calls to `substr` to use character count as its second argument.


sonndinh:
> Can we update the NetworkAddress unit test to cover this?

Sure, let me take a look.

PR #3511, opened 2022-05-16
RtpsUdpDataLink::RtpsWriter's readers_expecting_data_ Must Also Consider Fragment Requests

simpsont-oci:


mitza-oci (approved):

jrw972 (commented):
This is already fixed in https://github.com/objectcomputing/OpenDDS/pull/3508

simpsont-oci:
> This is already fixed in #3508

Ah, OK. Well, it's why my rtps_send_queue PR had so many fragmentation failures. I'm guessing it's also behind the bench fan_frag failures.

PR #3516, opened 2022-05-20
POC: Incorporate EventDispatcher Into TransportImpl / RtpsUdpDataLink

simpsont-oci:
Problem: Most of our scheduled and dispatched jobs / tasks wind up running on one of our transport threads where they potentially delay reading of data, with the sole exception of the datalink cleanup job which runs on a (mostly unused) task queue thread.

Solution: Replace the unused task queue with a more general EventDispatcher and offload many (but not all) of the scheduled and dispatched jobs onto the new EventDispatcher.

Notes:
- The EventDispatcher API is generic, but an initial non-reactor / non-proactor implementation has been provided (ServiceEventDispatcher, based on DispatchService)
- The old ReactorInterceptor classes (JobQueue / SporadicTask etc) have been left untouched, potentially to be removed at a future date if/when ReactorTask is ever removed.


mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.h
```diff
@@ -448,6 +448,8 @@ class OpenDDS_Rtps_Export Spdp
   DiscoveredParticipantMap participants_;
   RtpsDiscovery* disco_;
   DCPS::RcHandle<RtpsDiscoveryConfig> config_;
+  const double quick_resend_ratio_;
+  const DCPS::TimeDuration min_resend_delay_;
```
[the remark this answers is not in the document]
simpsont-oci: Right. this could be a separate PR if needed, but I noticed high mutex contention for one of the RtpsDiscoveryConfigs during the aggressive Thrasher tests so I tried to reduce load by having Spdp copy / cache these values. Currently, nothing is updating these once they are configured, but every Spdp/Sedp transport associated with the config will repeatedly try to go grab these values.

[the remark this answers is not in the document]
jrw972: I don't think there is a reason to change these two variables.

dds/DCPS/DispatchService.h
```diff
@@ -0,0 +1,105 @@
...
+  static const bool DS_ERROR = false;
+
+  typedef void*(*FunPtr)(void*);
+  typedef std::pair<FunPtr, void*> FunArgPair;
+  typedef OPENDDS_DEQUE(FunArgPair) EventQueue;
+
+  DispatchService(size_t count = 1);
+  virtual ~DispatchService();
+
+  void shutdown(bool immediate = false, EventQueue* const pending = 0);
```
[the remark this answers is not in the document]
simpsont-oci: I want a const pointer to a non-const EventQueue.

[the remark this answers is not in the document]
simpsont-oci: I guess for the function call, it would get copied so whether it's a const pointer doesn't matter?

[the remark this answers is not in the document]
mitza-oci: Right, top-level const is not part of the method signature.  You can add it to the definition in the .cpp file if you'd like, that's the same as declaring a local const which can help avoid mistakes where the author assigns to something that shouldn't be.

dds/DCPS/ThreadPool.h
```diff
@@ -0,0 +1,55 @@
...
+class OpenDDS_Dcps_Export ThreadPool
+{
+public:
+
+  typedef ACE_THR_FUNC_RETURN(*FunPtr)(void*);
+
+  ThreadPool(size_t count, FunPtr fun, void* arg = 0);
+  virtual ~ThreadPool();
+
+  static ACE_THR_FUNC_RETURN run(void* arg);
```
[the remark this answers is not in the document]
simpsont-oci: This seems much simpler / cleaner to me, since it doesn't pull in ACE_Service_Object / ACE_Event_Handler / ACE_Shared_Object. As it stands, this could pretty easily be migrated (or #ifdef'd) to C++11 only if we switch to std::thread.

sonndinh (commented):
I haven't been able to look at all the changes yet. But overall I think it would be helpful to have some documentation for the new classes and their methods in `dds/DCPS`, like the `EventDispatcher` class has, to give an idea of what each is doing.

dds/DCPS/Barrier.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Barrier provides thread-coordination across an expected number of threads
+ *
+ * Threads which call Barrier::wait() will all wait until the expected number
+ * of threads have called Barrier::wait(), at which point they will all be woken
+ * up and the next "phase" of waiting will begin (again, being released when the
+ * expected number of waiting threads is reached
+ */
+class OpenDDS_Dcps_Export Barrier {
```
jwillemsen, line 29: Why not use ACE_Barrier?

simpsont-oci: I didn't even think to check if ACE had a barrier class. It doesn't look too heavy-weight and the API is mostly the same, so I'll give it a shot, thanks.

dds/DCPS/DispatchService.cpp
```diff
@@ -0,0 +1,207 @@
...
+void DispatchService::shutdown(bool immediate, EventQueue* const pending)
+{
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  allow_dispatch_ = false;
+  stop_when_empty_ = true;
+  running_ = running_ && !immediate; // && with existing state in case shutdown has already been called
+  cv_.notify_all();
+
+  if (pool_.contains(ACE_Thread::self())) {
+    ACE_DEBUG((LM_DEBUG, "(%P|%t) Warning :: DispatchService::shutdown: Contained Thread Attempting To Call Shutdown."));
```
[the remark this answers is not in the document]
simpsont-oci: DispatchService is not aware of system debug levels, but this should probably follow the dev guidelines and be a full error message, since it's unrecoverable / unreportable.

dds/DCPS/PeriodicEvent.cpp
```diff
@@ -0,0 +1,87 @@
...
+    EventDispatcher_rch dispatcher = dispatcher_.lock();
+    if (dispatcher) {
+      const MonotonicTimePoint expiration = now + period;
+      long id = dispatcher->schedule(rchandle_from(this), expiration);
+      if (id > 0) {
+        period_ = period;
+        expiration_ = expiration;
+        timer_id_ = id;
+        strict_timing_ = strict_timing;
+      }
```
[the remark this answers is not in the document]
simpsont-oci: This should match the PeriodicTask behavior now, though based on @jwillemsen 's other comment we may want to make this dependent on debug levels?

dds/DCPS/EventDispatcher.h
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * EventBase is an abstract base class for events
+ *
+ * EventBase is an abstract base class for events dispatched or scheduled by
+ * an EventDispatcher. Derived classes are required to implement handle_event
+ * which will be called when the event is dispatched by the event dispatcher
+ */
+struct OpenDDS_Dcps_Export EventBase : public virtual RcObject
+{
```
mitza-oci, line 30:
```suggestion
struct OpenDDS_Dcps_Export EventBase : virtual RcObject {
```

dds/DCPS/transport/framework/TransportInst.cpp
```diff
@@ -218,6 +218,13 @@ OpenDDS::DCPS::TransportInst::reactor_task()
   return temp ? temp->reactor_task() : OpenDDS::DCPS::ReactorTask_rch();
 }
 
+OpenDDS::DCPS::EventDispatcher_rch
+OpenDDS::DCPS::TransportInst::event_dispatcher()
+{
+  const OpenDDS::DCPS::TransportImpl_rch temp = impl();
+  return temp ? temp->event_dispatcher() : OpenDDS::DCPS::EventDispatcher_rch();
```
mitza-oci, line 225:
```suggestion
  const TransportImpl_rch temp = impl();
  return temp ? temp->event_dispatcher() : EventDispatcher_rch();
```

mitza-oci (changes requested):

simpsont-oci:
Pretty sure the remaining bench failures (echo / echo_frag, maybe others?) will be resolved by #3586.

dds/DCPS/DispatchService.cpp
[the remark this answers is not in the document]
jrw972: According to the guidlines, this should be conditioned on `log_level >= Error`.

jrw972 (approved):

dds/DCPS/SporadicEvent.h
```diff
@@ -0,0 +1,83 @@
...
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+/**
+ * SporadicEvent is an event which can be scheduled multiple times but will only dispatch once
+ *
+ * SporadicEvent serves as a helpful utility class for allowing repeated calls
```
mitza-oci, line 21:
```suggestion
 * SporadicEvent allows repeated calls
```

dds/DCPS/ThreadPool.cpp
```diff
@@ -0,0 +1,75 @@
...
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+ThreadPool::ThreadPool(size_t count, FunPtr fun, void* arg)
+ : barrier_(count + 1)
+ , fun_(fun)
+ , arg_(arg)
+ , ids_(count, 0)
```
mitza-oci, line 23: warnings here

dds/DCPS/ThreadPool.h
```diff
@@ -0,0 +1,73 @@
...
+ * ThreadPool creates several threads at construction and attempts to join them
+ * at destruction. Users of ThreadPool are responsible for making sure the
+ * running threads are in a joinable state before the destruction of ThreadPool
+ */
+class OpenDDS_Dcps_Export ThreadPool
+{
+public:
+
+  /// A typedef for the starting point of the ThreadPool
+  typedef ACE_THR_FUNC_RETURN(*FunPtr)(void*);
```
mitza-oci, line 35:
```suggestion
  typedef ACE_THR_FUNC_RETURN (*FunPtr)(void*);
```

dds/DCPS/ThreadPool.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  {
+    ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+    for (size_t i = 0; i < count; ++i) {
+      ACE_Thread::spawn(run, this, THR_NEW_LWP | THR_JOINABLE, 0, &(ids_[i]));
```
mitza-oci, line 28: Is this the same as `spawn_n`?

dds/DCPS/ThreadPool.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+ThreadPool::~ThreadPool()
+{
+  join_all();
+}
+
+ACE_THR_FUNC_RETURN ThreadPool::run(void* arg)
+{
+  ThreadPool& pool = *(static_cast<ThreadPool*>(arg));
```
mitza-oci, line 43:
```suggestion
  ThreadPool& pool = *static_cast<ThreadPool*>(arg);
```

dds/DCPS/DispatchService.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace OpenDDS {
+namespace DCPS {
+
+template <typename T>
+void* fun_ptr_proxy(void* arg)
+{
+  T& ref = *static_cast<T*>(arg);
+  ref();
+  return 0;
```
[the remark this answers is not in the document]
simpsont-oci: It needs to match the function signature of ACE's thread start functions.

dds/DCPS/ThreadPool.cpp
simpsont-oci: I think I had it that way at first but changed it for some reason, perhaps because the flags weren't defaulted or because I wasn't getting the handles the way I expected.

dds/DCPS/DispatchService.h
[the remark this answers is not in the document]
mitza-oci: This would be helpful to have as a comment in the code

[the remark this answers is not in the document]
simpsont-oci: Ah, sorry, I misunderstood the compile error I saw. It needs to match the signature for DispatchService's dispatch and schedule, which we're free to change to match. Updated.

dds/DCPS/SporadicEvent.h
```diff
@@ -0,0 +1,81 @@
...
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+/**
+ * SporadicEvent is an event which can be scheduled multiple times but will only dispatch once
+ *
+ * SporadicEvent allows repeated calls to schedule (when the application is
+ * unsure if the event has already been scheduled, but cumulative calls to
```
mitza-oci, line 22:
```suggestion
 * unsure if the event has already been scheduled), but cumulative calls to
```

mitza-oci (approved):

dds/DCPS/DispatchService.h
```diff
@@ -0,0 +1,152 @@
...
+   * Cancel a scheduled event by function pointer
+   * @param fun the function pointer of the event/s to cancel
+   * @param arg the argument passed at time of scheduling
+   * @return the number of timers successfully canceled (potentially several)
+   */
+  size_t cancel(FunPtr fun, void* arg = 0);
+
+  /// Helper function to cancel arbitrary function objects (see fun_ptr_proxy)
+  template <typename T>
+  size_t cancel(T& ref) {
```
iguessthislldo, line 119:
```suggestion
  size_t cancel(T& ref)
  {
```

dds/DCPS/EventDispatcher.cpp
```diff
@@ -0,0 +1,57 @@
...
+  } catch (...) {
+    handle_error();
+  }
+  /// In order to avoid extra allocations for containers of RcHandle<EventBase>
+  /// implementations of EventDispatcher will increase the reference count of
+  /// all queued EventBase objects, and decrement them when they are dispatched
+  /// or canceled.
+  /// Since the EventDispatcher does not directly hold copies of the events,
+  /// decrementing the reference count after a normal dispatch currently needs
+  /// to happen here in the call to EventBase::operator()() after handle_event.
```
iguessthislldo, line 42: Doxygen comments do nothing here.

tests/unit-tests/dds/DCPS/DispatchService.cpp
```diff
@@ -0,0 +1,422 @@
...
+  void operator()() { if (increment_call_count() % 2) { dispatcher_.dispatch(*this); } }
+
+  OpenDDS::DCPS::DispatchService& dispatcher_;
+};
+
+struct RecursiveTestObjTwo : public TestObjBase
+{
+  RecursiveTestObjTwo(OpenDDS::DCPS::DispatchService& dispatcher, size_t dispatch_scale) : dispatcher_(dispatcher), dispatch_scale_(dispatch_scale) {}
+
+  void operator()() { increment_call_count(); const size_t scale = dispatch_scale_.value(); for (size_t i = 0; i < scale; ++i) { dispatcher_.dispatch(*this); } }
```
iguessthislldo, line 70: Make multiple lines?

tests/unit-tests/dds/DCPS/DispatchService.cpp
```diff
@@ -0,0 +1,422 @@
...
+ */
+
+#include <dds/DCPS/DispatchService.h>
+
+#include <ace/OS_NS_unistd.h>
+
+#include <gtest/gtest.h>
+
+namespace
+{
```
iguessthislldo, line 15:
```suggestion
namespace {
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3524, opened 2022-05-25
Consistency In Use Of Virtual Inheritance

simpsont-oci:
Problem: There are occasional TSAN warnings reported surrounding races between destructors and virtual function calls, as well as double destruction. While these might signal an issue with RcObject or RcHandle, it's also possible they are related to inconsistent use of virtual inheritance surrounding reference counted objects. There are, for example, several places where we do not use virtual inheritance for ACE_Event_Handler, RcObject, RcEventHandler, etc.

Solution: Be more consistent in use of virtual inheritance for these classes.

Note: Originally, this PR encompassed ReactorInterceptor as well, but the use of a protected constructor made consistent use of virtual inheritance difficult. Specifically, under virtual inheritance, grand-child classes (such as RequestedDeadlineWatchdog (via Watchdog)), indirectly complain about not having access to ReactorInterceptor's protected constructor (complaining instead about a lack of a default constructor, which is intentional). Consequently, use of ReactorInterceptor was made consistently non-virtual.

### DIFF
--- a/dds/DCPS/RTPS/ICE/AgentImpl.h
+++ b/dds/DCPS/RTPS/ICE/AgentImpl.h
@@ -35,9 +35,9 @@ typedef std::vector<FoundationType> FoundationList;
 
 class AgentImpl
   : public virtual Agent
-  , public virtual DCPS::ReactorInterceptor
   , public virtual DCPS::ShutdownListener
   , public virtual DCPS::InternalDataReaderListener<DCPS::NetworkInterfaceAddress>
+  , public DCPS::ReactorInterceptor
 {
 public:
   AgentImpl();
--- a/dds/DCPS/RequestedDeadlineWatchdog.h
+++ b/dds/DCPS/RequestedDeadlineWatchdog.h
@@ -36,7 +36,7 @@ class DataReaderImpl;
  * @c on_requested_deadline_missed() listener callback when the
  * configured finite deadline period expires.
  */
-class RequestedDeadlineWatchdog : public virtual OpenDDS::DCPS::Watchdog {
+class RequestedDeadlineWatchdog : public virtual Watchdog {
 public:
 
   typedef ACE_Recursive_Thread_Mutex  lock_type;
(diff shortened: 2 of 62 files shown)


dds/DCPS/RTPS/ICE/AgentImpl.h
```diff
@@ -35,9 +35,9 @@ typedef std::vector<FoundationType> FoundationList;
 
 class AgentImpl
   : public virtual Agent
-  , public virtual DCPS::ReactorInterceptor
   , public virtual DCPS::ShutdownListener
   , public virtual DCPS::InternalDataReaderListener<DCPS::NetworkInterfaceAddress>
+  , public DCPS::ReactorInterceptor
```
mitza-oci, line 40: Is the order important?  @jrw972

dds/DCPS/RequestedDeadlineWatchdog.h
```diff
@@ -36,18 +36,18 @@ class DataReaderImpl;
  * @c on_requested_deadline_missed() listener callback when the
  * configured finite deadline period expires.
  */
-class RequestedDeadlineWatchdog : public Watchdog {
+class RequestedDeadlineWatchdog : public virtual OpenDDS::DCPS::Watchdog {
```
mitza-oci, line 39:
```suggestion
class RequestedDeadlineWatchdog : public virtual Watchdog {
```

dds/DCPS/RTPS/ICE/AgentImpl.h
simpsont-oci: It was complaining about the initialization order between InternalDataReaderListener and Reactor Interceptor after I made this change. Evidently virtual inheritance gets initialized in a different order?

mitza-oci: Yes, it's on this page https://en.cppreference.com/w/cpp/language/constructor scroll down to `Initialization order`

mitza-oci (approved):

iguessthislldo (approved):

jrw972 (approved):

PR #3527, opened 2022-05-26
Update logic for choosing address and hostname in NetworkResource

sonndinh:
A couple of functions in `NetworkResource` are updated to check for hostnames that resolve to bogus IP addresses.
- `get_fully_qualified_hostname`: ignore hostnames that resolve to an IP address that is not found from the list of IP interface addresses.
- `choose_single_coherent_address` (string version): cache addresses returned from `getaddrinfo` for Windows. Ignore the port numbers of the cached addresses. The port numbers are set accordingly before returning a chosen address to the caller.

In addition, it enables Windows 2022 GitHub Actions test jobs as the above changes should fix the IP choosing issue on Windows 2022.

### DIFF
--- a/dds/DCPS/NetworkResource.cpp
+++ b/dds/DCPS/NetworkResource.cpp
@@ -69,7 +69,7 @@ namespace DCPS {
 bool verify_hostname(String hostname, ACE_INET_Addr* addr_array, size_t addr_count,
                      bool prefer_loopback, bool allow_ipv4_fallback)
 {
-  const ACE_INET_Addr addr = choose_single_coherent_address(hostname, prefer_feedback, allow_ipv4_fallback);
+  const ACE_INET_Addr addr = choose_single_coherent_address(hostname, prefer_loopback, allow_ipv4_fallback);
   for (size_t i = 0; i < addr_count; ++i) {
     if (addr == addr_array[i]) {
       return true;
--- a/dds/DCPS/NetworkResource.cpp
+++ b/dds/DCPS/NetworkResource.cpp
@@ -66,7 +66,7 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
-bool verify_hostname(String hostname, ACE_INET_Addr* addr_array, size_t addr_count,
+bool verify_hostname(const String& hostname, ACE_INET_Addr* addr_array, size_t addr_count,
                      bool prefer_loopback, bool allow_ipv4_fallback)
 {
   const ACE_INET_Addr addr = choose_single_coherent_address(hostname, prefer_loopback, allow_ipv4_fallback);
--- a/dds/DCPS/NetworkResource.cpp
+++ b/dds/DCPS/NetworkResource.cpp
@@ -901,6 +901,7 @@ ACE_INET_Addr choose_single_coherent_address(const String& address, bool prefer_
   // "some_domain.com:1234". We would expect to have address 10.1.0.71:1234 returned. However,
   // address 10.1.0.71:0 can be returned if it is still in the cache when the second call is made.
 
+
   // Maybe a better way to maintain the address cache is to have it only store IP addresses and
   // leave out the port number (by setting it to 0). This is because the choose_single_coherent_address
   // vector version only cares about IP part. That way we can reduce the number of entries in the cache
--- a/dds/DCPS/NetworkResource.cpp
+++ b/dds/DCPS/NetworkResource.cpp
@@ -870,7 +870,7 @@ ACE_INET_Addr choose_single_coherent_address(const String& address, bool prefer_
 
     ACE_INET_Addr temp;
     temp.set_addr(&addr, sizeof addr);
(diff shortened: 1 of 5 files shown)


dds/DCPS/NetworkResource.cpp
```diff
@@ -774,6 +797,15 @@ ACE_INET_Addr choose_single_coherent_address(const String& address, bool prefer_
   }
 #endif /* ACE_WIN32 */
 
+  size_t addr_count;
```
simpsont-oci, line 800: I don't think you can embed the check for whether the resolved address belongs to a local interface within choose_single_coherent_address, because both the local and remote processes (potentially on a different host) will call choose_single_coherent_address. I think the check should get pulled out into a new helper function that takes the address list as a parameter (since get_fully_qualified_hostname already has the list) and an address to check against the list... and then get_fully_qualifed_hostname can use it to reject candidates from being the FQDN as well as to evaluate the potential of nonFQDN candidates.

dds/DCPS/NetworkResource.cpp
```diff
@@ -794,18 +826,27 @@ ACE_INET_Addr choose_single_coherent_address(const String& address, bool prefer_
...
 #endif /* ACE_WIN32 */
+      }
+    }
   }
 
 #ifdef ACE_WIN32
   g.release();
 #endif /* ACE_WIN32 */
 
+  delete[] addr_array;
```
simpsont-oci, line 849: I don't think we need to do this here (see other comment), but if we did I'd prefer to see something like ArrayGuard above to handle the deletion. I suppose the same goes for freeaddrinfo. Scoped protection object is better in case we get an exception or if we add an early return for some reason in the future.

dds/DCPS/NetworkResource.cpp
```diff
@@ -90,55 +90,59 @@ String get_fully_qualified_hostname(ACE_INET_Addr* addr)
...
-            VDBG_LVL((LM_DEBUG, "(%P|%t) found fqdn %C from %C\n", hostname, LogAddr(addr_array[i]).c_str()), 2);
-            selected_address = addr_array[i];
-            fullname = hostname;
-            if (addr) {
-              *addr = selected_address;
+          VDBG_LVL((LM_DEBUG, "(%P|%t) get_fully_qualified_hostname: Considering fqdn %C\n", hostname), 4);
+          // Find the first FQDN that resolves to a valid IP interface address.
+          // If there is no such FQDN, consider the list of non-FQDNs.
+          if (ACE_OS::strchr(hostname, '.') != 0) {
+            if (!addr_array[i].is_loopback() &&
```
simpsont-oci, line 124: What happens to the candidate name if it passes the first check (if `strchr()` returns non-zero) but fails the second (it's a loopback or choose_single_coherent_address returns `ACE_INET_Addr()`)? Before, we did some checks to see if we needed to add it to non-FQDN, but now it looks like we ignore it completely?

simpsont-oci (changes requested):

dds/DCPS/NetworkResource.cpp
sonndinh: The reason I wanted to check the resolved addresses against the list of IP interfaces inside `choose_single_coherent_address` is that I think there can be cases where a hostname resolves to some valid IP and some incorrect IP, and `choose_single_coherent_address` may return an incorrect IP address instead of a correct one. In that case, we will ignore this hostname if we just rely on a single address returned by `choose_single_coherent_address` while actually, we can still use it with a correct IP mapped to it. So I wanted to do a more comprehensive check that checks all resolved addresses against the IP interfaces list. And `choose_single_coherent_address` seemed to be a good place to do that check. But I didn't know `choose_single_coherent_address` is also called by a remote peer to find the address of a local peer. So this won't work.

sonndinh: From the remote peer's perspective, there isn't a way for it to know if an address returned from `choose_single_coherent_address` is a correct one. So I think we probably should just ignore the hostname with polluted IP addresses altogether.

sonndinh:
This PR can be reviewed now. There are some additional logs and comments that I will clean up when the CI is done. But the main logic changes are ready for review.

dds/DCPS/NetworkResource.h
```diff
@@ -63,6 +63,10 @@ struct OpenDDS_Dcps_Export NetworkResource {
   String addr_;
 };
 
+// Make sure that choose_single_coherent_address picks one of the IP interface addresses.
+bool verify_hostname(String hostname, ACE_INET_Addr* addrs, size_t addr_count,
```
mitza-oci, line 67:
```suggestion
bool verify_hostname(const String& hostname, ACE_INET_Addr* addrs, size_t addr_count,
```

mitza-oci (changes requested):

mitza-oci (approved):

simpsont-oci (approved):
Looks good. Can you update the PR description with a brief summary of changes?

sonndinh:
> Looks good. Can you update the PR description with a brief summary of changes?

Sure, I updated it.

PR #3528, opened 2022-05-27
Required acknack flag not reset

jrw972:
Problem
-------

The required acknack count flag is not reset.  The logic for updating
the required acknack count can be simplified.

Solution
--------

* Reset the flag when adding the submessage.
* Any submessage with the flag should update the count.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -2705,19 +2705,17 @@ RtpsUdpDataLink::queue_submessages(MetaSubmessageVec& in, double scale)
 }
 
 void
-RtpsUdpDataLink::RtpsWriter::update_required_acknack_count(const RepoId& id, CORBA::Long previous, CORBA::Long current)
+RtpsUdpDataLink::RtpsWriter::update_required_acknack_count(const RepoId& id, CORBA::Long current)
 {
   ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
   ReaderInfoMap::iterator ri = remote_readers_.find(id);
   if (ri != remote_readers_.end()) {
-    if (ri->second->required_acknack_count_ == previous) {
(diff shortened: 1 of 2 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2803,16 +2801,14 @@ RtpsUdpDataLink::bundle_and_send_submessages(MetaSubmessageVec& meta_submessages
...
-                if (res.sm_.heartbeat_sm().count.value != map_pair.new_) {
-                  update_required_acknack_count(res.src_guid_, res.dst_guid_, res.sm_.heartbeat_sm().count.value, map_pair.new_);
-                }
-                res.sm_.heartbeat_sm().smHeader.flags &= ~RTPS::OPENDDS_FLAG_R;
-              }
             }
             map_pair.is_new_assigned_ = true;
           }
           res.sm_.heartbeat_sm().count.value = map_pair.new_;
+          if (res.sm_.heartbeat_sm().smHeader.flags & RTPS::OPENDDS_FLAG_R) {
```
simpsont-oci, line 2808: We should only call `update_required_acknack_count` once when the count mapping assignment happens (after map_pair.new_ is assigned) for directed heartbeats. Where it's at now, the function will get called multiple times for the same heartbeat submessage, if it exists in multiple bundles. And it doesn't make sense to call it for undirected heartbeats, right (which it will also currently do).

simpsont-oci (approved):
My only remaining hesitancy is removing the check against previous count, since that would protect you against updating the count incorrectly if another heartbeat came through (e.g. a HB of 5 gets remapped to 3 followed by another HB with count 7 that doesn't get remapped... the 7 might get replaced with a 3). Practically though, I don't know how frequently this would happen... the later update would need to be excluded from the send queue for some reason and if everything is running on the reactor thread I don't think that can happen.

It's a shame we don't have more tests set up for for the reflective heartbeats.

PR #3542, opened 2022-06-06
Remove addr_guids from BundlingCacheKey for Performance

simpsont-oci:
Remove addr_guids from BundlingCacheKey to improve bundling cache lookup performance and work around for Static Discovery. This allows for AddressCache Key simplification and removes the need for keys to inherit from RcObject. Additionally, this PR combines the three separate bundling vector structures (submessages, sizes, addresses) into a single structure to reduce total number of allocations and improve code clarity.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
+++ b/dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
@@ -50,17 +50,6 @@ struct OpenDDS_Rtps_Udp_Export BundlingCacheKey {
     return std::memcmp(static_cast<const void*>(this), static_cast<const void*>(&rhs), sizeof (BundlingCacheKey)) == 0;
   }
 
-  /*
-  BundlingCacheKey& operator=(const BundlingCacheKey& rhs)
-  {
-    if (this != &rhs) {
-      const_cast<GUID_t&>(src_guid_) = rhs.src_guid_;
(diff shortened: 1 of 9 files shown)


dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
```diff
@@ -27,120 +27,57 @@ namespace DCPS {
...
-    return BundlingCacheKeyCore::operator==(rhs);
+    return std::memcmp(static_cast<const void*>(this), static_cast<const void*>(&rhs), sizeof (BundlingCacheKey)) == 0;
   }
 
   BundlingCacheKey& operator=(const BundlingCacheKey& rhs)
   {
     if (this != &rhs) {
-      BundlingCacheKeyCore::operator=(rhs);
+      const_cast<GUID_t&>(src_guid_) = rhs.src_guid_;
+      const_cast<GUID_t&>(dst_guid_) = rhs.dst_guid_;
```
mitza-oci, line 57: If the class supports assignment then these aren't really const.  I guess I'd rather see `const` removed from the data members.

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/BundlingCacheKey.h
simpsont-oci: removing operator= for now

mitza-oci (approved):

PR #3549, opened 2022-06-08
Shmem Association Race Fix

ClaytonCalabrese:
Shmem transport has an association race where the writer's side can write messages and call `sema_post` before the reader's side is associated enough to find the messages. The reader's side will take from `sema_wait`, and be unable to read anything, but due to the design of `sema_wait`, the semaphore will be decremented anyways. 

This boils down to the reader not truly being ready to receive at the time in which `publication_matched_status` is seen by the writer.

In order to avoid this, we decided to use the design from TCP which has the reader side send a message to the writer side to signify that the communication channel is ready.


dds/DCPS/transport/shmem/ShmemTransport.cpp
```diff
@@ -260,7 +281,9 @@ ShmemTransport::ReadTask::svc()
     if (stopped_.value()) {
       return 0;
     }
-    outer_->read_from_links();
+    while (!(outer_->read_from_links() || stopped_.value())) {
+      ACE_OS::sleep(1);
```
iguessthislldo, line 285: I've kept this sleep in so far, but I'm not sure if it's still necessary or not.

dds/DCPS/transport/shmem/ShmemTransport.cpp
```diff
@@ -286,7 +309,7 @@ ShmemTransport::ReadTask::signal_semaphore()
   ACE_OS::sema_post(&semaphore_);
 }
 
-void
+bool
```
mitza-oci, line 308: Is the return value used?

mitza-oci (changes requested):

dds/DCPS/transport/shmem/ShmemTransport.cpp
iguessthislldo: It used in the inner loop in of `ReadTask::svc`.

mitza-oci: It is... but I'm not sure this is necessary.  

In the original design, `read_from_links` happens once per successful wait on the semaphore (since it's a counting semaphore, it knows how many "resources" are available).  Would that work now that the rest of the transport framework integration is done?

If the answer to the above ends up being "no," then could `read_from_links` just do all the work it needs to do before returning?  I guess the downside there is potentially starving-out the `stopped_` signal, but that could be passed by reference if needed.

One reason I brought this up is that if we keep it in we'll need to review each path through `read_from_links` to see if it's returning the correct value.

dds/DCPS/transport/shmem/ShmemReceiveStrategy.cpp
```diff
@@ -75,6 +75,7 @@ ShmemReceiveStrategy::read()
   // If we get this far, current_data_ points to the first SHMEM_DATA_IN_USE.
   // handle_dds_input() will call our receive_bytes() to get the data.
   handle_dds_input(ACE_INVALID_HANDLE);
+  return;
```
mitza-oci, line 78: seems to be unnecessary

dds/DCPS/transport/shmem/ShmemDataLink.cpp
```diff
@@ -89,12 +97,117 @@ ShmemDataLink::open(const std::string& peer_address)
...
+}
+
+void ShmemDataLink::resend_association_msgs(const MonotonicTimePoint&)
+{
+  VDBG((LM_DEBUG, "(%P|%t) ShmemDataLink::resend_association_msgs\n"));
+
+  ACE_GUARD(ACE_Recursive_Thread_Mutex, g, mutex_);
+  for (AssocResends::iterator i = assoc_resends_.begin(); i != assoc_resends_.end();) {
+    send_association_msg(i->first.local, i->first.remote);
+    i->second--;
```
mitza-oci, line 174:
```suggestion
    --i->second;
```
or combine with next line (the conditional for the `if`)

dds/DCPS/transport/shmem/ShmemDataLink.cpp
```diff
@@ -24,27 +27,34 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
+
   peer_address_ = peer_address;
   const ACE_TString name = ACE_TEXT_CHAR_TO_TCHAR(peer_address.c_str());
-  ShmemAllocator::MEMORY_POOL_OPTIONS alloc_opts;
+  ShmemAllocator::MEMORY_POOL_OPTIONS* alloc_opts_ptr = 0;
 
 #ifdef OPENDDS_SHMEM_WINDOWS
-  const bool use_opts = true;
+  ShmemAllocator::MEMORY_POOL_OPTIONS alloc_opts;
+  alloc_opts_ptr = &alloc_opts;
```
mitza-oci, line 57: I had a comment about this somewhere else (maybe in this same PR), but I think this is more complex than necessary.  We know at compile time if the `opts` will be used or not.

mitza-oci (changes requested):

mitza-oci (approved):

PR #3554, opened 2022-06-09
Updated development guidelines

mitza-oci:
- mostly based on what we learned in #3544
- and a  few other clarifications

### DIFF
--- a/docs/internal/dev_guidelines.rst
+++ b/docs/internal/dev_guidelines.rst
@@ -50,9 +50,10 @@ Dependencies
 * MPC is the build system, used to configure the build and generate platform specific build files (Makefiles, VS solution files, etc.).
 * ACE is a library used for cross-platform compatibility, especially networking and event loops.
   It is used both directly and through TAO.
-* TAO is a C++ CORBA implementation built on ACE used extensively in the traditional OpenDDS operating mode which uses the DCPSInfoRepo.
-  TAO types are also used in the End User DDS API.
-  The TAO IDL compiler is used internally and by the end user to allow OpenDDS to use user defined IDL types as topic data.
+* TAO is a C++ CORBA implementation built on ACE.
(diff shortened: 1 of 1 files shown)


docs/internal/dev_guidelines.rst
```diff
@@ -50,9 +50,10 @@ Dependencies
 * MPC is the build system, used to configure the build and generate platform specific build files (Makefiles, VS solution files, etc.).
 * ACE is a library used for cross-platform compatibility, especially networking and event loops.
   It is used both directly and through TAO.
-* TAO is a C++ CORBA implementation built on ACE used extensively in the traditional OpenDDS operating mode which uses the DCPSInfoRepo.
-  TAO types are also used in the End User DDS API.
-  The TAO IDL compiler is used internally and by the end user to allow OpenDDS to use user defined IDL types as topic data.
+* TAO is a C++ CORBA implementation built on ACE.
+  * It's used to communicate with DCPSInfoRepo, which is one option for Discovery.
+  * TAO's data types and support for the OMG IDL-to-C++ mapping are also used in the End User DDS API.
+  * The TAO IDL compiler is used internally and by the end user to allow OpenDDS to use user-defined IDL types as topic data.
```
iguessthislldo, line 56:
```suggestion

  * It's used to communicate with DCPSInfoRepo, which is one option for Discovery.
  * TAO's data types and support for the OMG IDL-to-C++ mapping are also used in the End User DDS API.
  * The TAO IDL compiler is used internally and by the end user to allow OpenDDS to use user-defined IDL types as topic data.

```

For some reason RST won't recognize it as a nested list unless it's separated from the parent list. See GitHub's "rich diff" of this to see what it does otherwise.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3551, opened 2022-06-08
add TAB minutes from 3-7-22 meeting

kroegerk:
@mitza-oci

PR #3556, opened 2022-06-09
Fix Jekyll Template Issue from #3551

iguessthislldo:
Fix Jeykell template that produces a malformed `<ul>` element when there
are minutes entries in the "Minutes Archive" Box for multiple years.

Also added instructions on how to run `htmlproofer.sh` locally because
they were not there before and I had re-figure out how to run it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ _site
 vendor
 .vscode/
 /Gemfile.lock
+/.bundle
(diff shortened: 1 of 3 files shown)


mitza-oci (approved):

kroegerk (commented):
Thank you for figuring that out, @iguessthislldo !

PR #3557, opened 2022-06-09
Instance security

jrw972:


dds/DCPS/XTypes/DynamicType.cpp
```diff
@@ -7,85 +7,120 @@
 #include <dds/DdsDcpsInfrastructureC.h>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+DDS::DynamicTypeMembersById_ptr TAO::Objref_Traits<DDS::DynamicTypeMembersById>::nil()
+{
+  // TODO: Help.
```
jrw972, line 13: Since we don't have support for `map`s in IDL, I defined DynamicTypeMembersById and DynamicTypeMembersByName as interfaces.

It seems like nil should be easy to implement, but how about `release below`?

dds/DCPS/XTypes/MemberDescriptor.cpp
```diff
@@ -8,52 +8,241 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
-  if (strong_type.in() == 0) {
-    ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) MemberDescriptor::get_type(): grabbing of lock failed\n")));
-  }
-  return strong_type;
+  DynamicTypePtrPairSeen dt_ptr_pair;
+  return test_equality_i(this, other, dt_ptr_pair);
 }
 
-bool MemberDescriptor::equals(const MemberDescriptor& other) const
+CORBA::ValueBase* MemberDescriptorImpl::_copy_value()
```
jrw972, line 31: This comes from ValueType.  I need some guidance on how to implement these.

dds/DCPS/XTypes/MemberDescriptor.cpp
```diff
@@ -8,52 +8,241 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
+}
+
+::CORBA::Boolean MemberDescriptorImpl::is_consistent()
+{
+  // FUTURE
+  ACE_ERROR((LM_ERROR, "(%P|%t) ERRROR: MemberDescriptorImpl::is_consistent is not implemented\n"));
+  return false;
+}
+
+::CORBA::Boolean MemberDescriptorImpl::_tao_marshal__DDS_MemberDescriptor(TAO_OutputCDR &, TAO_ChunkInfo &) const
```
jrw972, line 217: Since these should never be marshalled, I don't think we need to do anything.  Maybe print an error.

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -553,22 +589,50 @@ ACE_CDR::ULong DynamicData::get_item_count()
...
+{
+  return DDS::RETCODE_UNSUPPORTED;
+}
+
+DDS::ReturnCode_t DynamicDataImpl::clear_value(DDS::MemberId /*id*/)
+{
+  return DDS::RETCODE_UNSUPPORTED;
+}
+
+DDS::DynamicData_ptr DynamicDataImpl::loan_value(DDS::MemberId /*id*/)
```
jrw972, line 613: Do we need the loaning support?

dds/DCPS/XTypes/DynamicData.cpp
```diff
@@ -1367,7 +1464,8 @@ bool DynamicData::read_values(SequenceType& value, TypeKind elem_tk)
   case TK_STRING8:
   case TK_STRING16:
     {
-      return strm_ >> value;
+      // TODO: Help.
```
jrw972, line 1467: The `DdsDynamicData.idl` file defines various sequence types as part of the Dynamic Data Interface.  The serializer doesn't understand these types.  Thus, we need to decide if we typedef them to the ones it does understand, switch to the new ones, etc.

dds/DCPS/XTypes/DynamicDataHolder.h
```diff
@@ -0,0 +1,807 @@
...
+      return DDS::RETCODE_ERROR;
+    }
+
+    DDS::DynamicType_var type = get_base_type(md->type());
+    if (type->get_kind() != TK_STRING8) {
+      return DDS::RETCODE_ERROR;
+    }
+
+    const DCPS::Value v = meta_struct_.getValue(&value_, id);
+    // TODO: Allocate?  Transfer?
```
jrw972, line 538: I only need string for my purposes.

mitza-oci: It's an `inout string` parameter. As the implementer, you get a buffer that the caller allocated (it may not be NULL). You can decide to leave it as-is, change the contents of the buffer, or free it and allocate a new one in its place.  In all cases, on return the caller is responsible for the buffer that it points to.  String_var has support for this.

dds/DCPS/XTypes/DynamicType.cpp
jrw972: These types are going to get used extensively when processing a DynamicData so the map feature is useful.

dds/DCPS/XTypes/DynamicData.cpp
jrw972: No.  But they will be deserialized.

To be specific, if I uncomment these lines, the compiler complains that the operation is not defined for the DDS::*Seq type.  It is defined for the sequence types we are used to dealing with.  Thus, it seems I can either typedef DDS::*Seq to the type we already have or replace/rename the existing types to DDS::*Seq.

mitza-oci: OK, I think I'm starting to understand what's going on here. The XTypes Annex C IDL defines its own series of `typedef sequence<X> YSeq;` which overlap with, but are not completely the same as, other ones like `OctetSeq` and `StringSeq` in the core spec's IDL.

The comments in DdsDcps.mpc describe the different code generation strategies.  One way to handle this would be to refactor this IDL file and pull out any types that need `opendds_idl` to generate XCDR serialization support.

Another strategy would be to treat these new types like `ByteSeq` as containers that have a C++ mapping but do not have XCDR serializers generated by opendds_idl.  This is the same as `XTypes::Sequence<T>` in `TypeObject.h`.  The only real trick is keeping track of when DHeader is used in XCDR2. Aside from that they're very simple.

dds/DCPS/XTypes/DynamicType.cpp
mitza-oci: See https://github.com/DOCGroup/ACE_TAO/pull/1842 for maps

If they're changed to interfaces the, API for functions that pass/return them will be significantly different.

dds/DCPS/DataWriterImpl_T.h
```diff
@@ -235,6 +264,21 @@ DataWriterImpl_T
...
+#ifndef OPENDDS_NO_CONTENT_SUBSCRIPTION_PROFILE
+    XTypes::DynamicDataHolder<MessageType> ddh(dynamic_type_, getMetaStruct<MessageType>(), instance_data);
+    DDS::Security::SecurityException ex;
+
+    if (!security_config_->get_access_control()->check_local_datawriter_dispose_instance(participant_permissions_handle_, this, &ddh, ex)) {
+      ACE_ERROR((LM_ERROR,
+                 "(%P|%t) ERROR: DataWriterImpl_T::dispose_instance_w_timestamp: unable to dispose instance SecurityException[%d.%d]: %C\n",
+                 ex.code, ex.minor_code, ex.message.in()));
+      // TODO: What should this be?
+      return DDS::RETCODE_ERROR;
```
iguessthislldo, line 277: `DDS::Security::NOT_ALLOWED_BY_SECURITY`?

dds/DCPS/ValueWriter.h
```diff
@@ -21,6 +21,28 @@
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
+#ifdef OPENDDS_SAFETY_PROFILE
+namespace DDS {
+
+class MemberDescriptor {
+public:
+
+  MemberDescriptorImpl(const char* a_name, bool a_is_key)
```
mitza-oci, line 30:
```suggestion
  MemberDescriptor(const char* a_name, bool a_is_key)
```

mitza-oci (changes requested):
This is a partial review of e2abe88c

tests/security/CheckInstance/run_test.pl
```diff
@@ -0,0 +1,22 @@
+eval '(exit $?0)' && eval 'exec perl -S $0 ${1+"$@"}'
+    & eval 'exec perl -S $0 $argv:q'
+    if 0;
+
+use lib "$ENV{ACE_ROOT}/bin";
+use lib "$ENV{DDS_ROOT}/bin";
+use PerlDDS::Run_Test;
+use strict;
+
+PerlDDS::add_lib_path('$ENV{DDS_ROOT}/tess/DCPS/ConsolidatedMessengerIdl');
```
mitza-oci, line 10:
```suggestion
PerlDDS::add_lib_path("$ENV{DDS_ROOT}/tests/DCPS/ConsolidatedMessengerIdl");
```

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -114,35 +117,30 @@ void DynamicData::copy(const DynamicData& other)
...
   type_ = other.type_;
-  descriptor_ = other.descriptor_;
   item_count_ = other.item_count_;
 }
 
-DDS::ReturnCode_t DynamicData::get_descriptor(MemberDescriptor& value, MemberId id) const
+DDS::ReturnCode_t DynamicDataImpl::get_descriptor(DDS::MemberDescriptor*& value, MemberId id)
 {
-  DynamicTypeMember_rch dtm;
+  const DDS::DynamicType_var base_type = DDS::DynamicType::_duplicate(get_base_type(type_));
```
mitza-oci, line 125: This kind of reference counting doesn't follow the normal patterns.  For functions not defined in IDL, there are two options:
- Follow the same parameter passing / return type rules as if it was an IDL operation. In this case the caller should increment the reference count, not the callee.
- Use _vars directly for parameters and return types and let the _var constructors/assignment/etc take care of reference counting

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -601,23 +666,30 @@ bool DynamicData::read_value(ValueType& value, TypeKind tk)
...
+                                            TypeKind enum_or_bitmask, LBound lower, LBound upper)
 {
-  MemberDescriptor md;
+  DDS::MemberDescriptor_var md;
   if (get_from_struct_common_checks(md, id, MemberTypeKind)) {
     return skip_to_struct_member(md, id) && read_value(value, MemberTypeKind);
-  } else if (get_from_struct_common_checks(md, id, enum_or_bitmask)) {
-    const DynamicType_rch member_type = md.type.lock();
+  }
+  // Reset md.
```
mitza-oci, line 683:
```suggestion

```

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -626,122 +698,143 @@ bool DynamicData::get_value_from_struct(MemberType& value, MemberId id,
...
+    if (it->second->get_descriptor(md) != DDS::RETCODE_OK) {
+      return 0;
+    }
+    const DDS::UnionCaseLabelSeq& labels = md->label();
     for (ACE_CDR::ULong i = 0; i < labels.length(); ++i) {
       if (label == labels[i]) {
-        out_md = md;
-        return true;
+        CORBA::add_ref(md);
+        return md;
```
mitza-oci, line 740:
```suggestion
        return md._retn();
```

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -1122,26 +1267,31 @@ DDS::ReturnCode_t DynamicData::get_char_common(CharT& value, MemberId id)
...
 
 #ifdef DDS_HAS_WCHAR
-DDS::ReturnCode_t DynamicData::get_char16_value(ACE_CDR::WChar& value, MemberId id)
+DDS::ReturnCode_t DynamicDataImpl::get_char16_value(ACE_CDR::WChar& value, MemberId id)
 {
+  #ifdef DDS_HAS_WCHAR
   return get_char_common<TK_CHAR16, TK_STRING16, ACE_InputCDR::to_wchar>(value, id);
+  #else
+  return DDS::RETCODE_UNSUPPORTED;
+  #endif
```
mitza-oci, line 1282:
```suggestion
#ifdef DDS_HAS_WCHAR
  return get_char_common<TK_CHAR16, TK_STRING16, ACE_InputCDR::to_wchar>(value, id);
#else
  return DDS::RETCODE_UNSUPPORTED;
#endif
```

dds/DCPS/XTypes/DynamicDataHolder.h
```diff
@@ -0,0 +1,812 @@
...
+#include <dds/DdsDynamicDataC.h>
+#include <dds/DCPS/FilterEvaluator.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace XTypes {
+
+template <typename T>
+class DynamicDataHolder : public DDS::DynamicData {
```
mitza-oci, line 21: Is there a more descriptive name for this? 
AdaptedDynamicData
DynamicAccess
DynamicView

mitza-oci (changes requested):

dds/DCPS/XTypes/MemberDescriptorImpl.h
```diff
@@ -0,0 +1,107 @@
...
+  MemberDescriptor(const char* a_name, bool a_is_key)
+    : name_(a_name)
+    , is_key_(a_is_key)
+  { }
+
+  const char* name() const { return name_; }
+  bool is_key() const { return is_key_; }
+
+private:
+  const char* name_;
```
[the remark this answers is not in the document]
jrw972: They should always be string literals.  I'll comment to that effect.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -85,27 +88,27 @@ DynamicData::DynamicData(DCPS::Serializer& ser, const DynamicType_rch& type)
   strm_.rdstate(align_state_);
 }
 
-DynamicData::DynamicData(const DynamicData& other)
+DynamicDataImpl::DynamicDataImpl(const DynamicDataImpl& other)
```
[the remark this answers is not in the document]
mitza-oci: Is this still an issue?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -601,23 +666,33 @@ bool DynamicData::read_value(ValueType& value, TypeKind tk)
...
+bool DynamicDataImpl::get_value_from_struct(MemberType& value, MemberId id,
+                                            TypeKind enum_or_bitmask, LBound lower, LBound upper)
 {
-  MemberDescriptor md;
-  if (get_from_struct_common_checks(md, id, MemberTypeKind)) {
-    return skip_to_struct_member(md, id) && read_value(value, MemberTypeKind);
-  } else if (get_from_struct_common_checks(md, id, enum_or_bitmask)) {
-    const DynamicType_rch member_type = md.type.lock();
+  {
+    // Scoped because reusing md creates a memory leak.
```
mitza-oci, line 680: Does this mean there's another bug lurking?

dds/DCPS/XTypes/DynamicDataImpl.cpp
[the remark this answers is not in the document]
mitza-oci: I think I see what's going on here.  Either inherit from `DCPS::LocalObject` or do something functionally equivalent wrt the use of TAO's `CORBA::LocalObject`.

mitza-oci (approved):

PR #3559, opened 2022-06-10
Fix TSAN Issues Surrounding InstanceState::registered_

simpsont-oci:
See: https://github.com/objectcomputing/OpenDDS/runs/6820767076?check_suite_focus=true

### DIFF
--- a/dds/DCPS/InstanceState.cpp
+++ b/dds/DCPS/InstanceState.cpp
@@ -282,6 +282,7 @@ bool InstanceState::is_exclusive() const
 
 bool InstanceState::registered()
 {
+  ACE_Guard<ACE_Recursive_Thread_Mutex> guard(lock_);
   const bool ret = registered_;
   registered_ = true;
   return ret;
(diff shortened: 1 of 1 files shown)


dds/DCPS/InstanceState.cpp
```diff
@@ -282,18 +282,21 @@ bool InstanceState::is_exclusive() const
...
 void InstanceState::registered(bool flag)
 {
+  ACE_Guard<ACE_Recursive_Thread_Mutex> guard(lock_);
   registered_ = flag;
 }
 
 void InstanceState::reset_ownership(DDS::InstanceHandle_t instance)
 {
+  ACE_Guard<ACE_Recursive_Thread_Mutex> guard(lock_);
   set_owner(GUID_UNKNOWN);
```
mitza-oci, line 300: `set_owner` has its own guard

simpsont-oci: yep

PR #3563, opened 2022-06-15
StaticDiscovery mpmrmw generates too much load

jrw972:
Problem
-------

Tests on Windows buildfarm VMs show that the StaticDiscovery mpmrmw
configuration may be generating too much load.

Solution
--------

Use a smaller configuration.


mitza-oci (approved):
Once it passes CI we'll be ready to merge

PR #3576, opened 2022-06-27
Remove Watchdog

jrw972:


dds/DCPS/DataReaderImpl_T.h
```diff
@@ -994,10 +995,28 @@ namespace OpenDDS {
         const DDS::Duration_t zero = { DDS::DURATION_ZERO_SEC, DDS::DURATION_ZERO_NSEC };
         if (qos_.time_based_filter.minimum_separation != zero) {
           if (qos.time_based_filter.minimum_separation != zero) {
-            const TimeDuration new_interval(qos.time_based_filter.minimum_separation);
-            filter_delayed_handler_->reset_interval(new_interval);
+            const MonotonicTimePoint now = MonotonicTimePoint::now();
+            const TimeDuration interval(qos_.time_based_filter.minimum_separation);
+            FilterDelayedSampleQueue queue_;
```
simpsont-oci, line 1000:
```suggestion
            FilterDelayedSampleQueue queue;
```

dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3570,6 +3540,185 @@ const DataReaderImpl::HandleSet& DataReaderImpl::lookup_matching_instances(CORBA
...
+    }
+  }
+}
+
+void DataReaderImpl::deadline_task(const MonotonicTimePoint& now)
+{
+  ThreadStatusManager::Event ev(TheServiceParticipant->get_thread_status_manager());
+
+  ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, sample_lock_);
+  for (DeadlineQueue::iterator pos = deadline_queue_.begin(), limit = deadline_queue_.end(); pos != limit && pos->first <= now;) {
```
simpsont-oci, line 3710: For consistency, might want to use lower bound here?

dds/DCPS/DataReaderImpl_T.h
```diff
@@ -2149,244 +2168,183 @@ DDS::ReturnCode_t check_inputs(const char* method_name,
...
+  // Make a copy because finish_store_instance_data will release the sample lock.
+  typedef OPENDDS_VECTOR(DDS::InstanceHandle_t) Handles;
+  Handles handles;
 
-      // use the handle to erase, since the sample lock was released
-      map_.erase(handle);
-    }
+  ACE_GUARD(ACE_Recursive_Thread_Mutex, guard, sample_lock_);
+
+  for (FilterDelayedSampleQueue::iterator pos = filter_delayed_sample_queue_.begin(), limit = filter_delayed_sample_queue_.end(); pos != limit && pos->first <= now;) {
```
simpsont-oci, line 2254: lower bound?

dds/DCPS/DataReaderImpl.cpp
jrw972: For `limit`?  I would prefer to not to do a separate lookup.  It doesn't make sense for `pos` because the front of the queue needs to be processed.

simpsont-oci: Yeah, for limit. It's O(logn) timepoint comparisons to find the stopping point (via a single lower_bound lookup) rather than O(n) for comparing every iterator's timepoint against `now`. It's not a big deal, performance-wise, but above you're using both lower and upper bound to determine your start / stop positions so I thought it would make sense for consistency to do it here as well.

jrw972: I'm assuming that the actual number of elements processed is small, i.e., less than `log N`, so that it would still be more efficient to compare instead of doing the lookup.

dds/DCPS/DataReaderImpl.h
```diff
@@ -973,7 +971,24 @@ class OpenDDS_Dcps_Export DataReaderImpl
   CORBA::Long last_deadline_missed_total_count_;
   /// Watchdog responsible for reporting missed offered
   /// deadlines.
-  RcHandle<RequestedDeadlineWatchdog> watchdog_;
+  TimeDuration deadline_period_;
+  typedef OPENDDS_MULTIMAP(MonotonicTimePoint, SubscriptionInstance_rch) DeadlineQueue;
+  DeadlineQueue deadline_queue_;
+  bool deadline_queue_enabled_;
+  typedef DCPS::PmfSporadicTask<DataReaderImpl> DRISporadicTask;
```
mitza-oci, line 978:
```suggestion
  typedef PmfSporadicTask<DataReaderImpl> DRISporadicTask;
```

And the same on lines 981-989

mitza-oci (changes requested):

dds/DCPS/DataReaderImpl.cpp
simpsont-oci: Gotcha. Hadn't considered that we're usually waiting on the first deadline, meaning we often / usually only need to consider the first element in the map.

simpsont-oci (approved):

mitza-oci (approved):

PR #3592, opened 2022-07-06
RtpsUdpSendStrategy sporadically reports failure to encode submessage

jrw972:
Problem
-------

The RtpsUdpSendStrategy sporadically reports an error when encoding a
submessage.  The submessage kind is a heartbeat so it must be a
writer.  Review of the code shows that it is probably the volatile
writer.

Solution
--------

Improve the logging to confirm it is the volatile writer and narrow
down the root cause.  The two possible causes are:

1. Sending to `GUID_UNKNOWN` which is a bug in the RtpsUdpDataLink.

2. No crypto handle for the remote datareader.  This could be caused
   by a race condition where the heartbeat (presumably caused by
   association) is sent before the handle is in place.

Also, improve the atomicity in the Crypto plugin.

### DIFF
--- a/dds/DCPS/security/CryptoBuiltInImpl.cpp
+++ b/dds/DCPS/security/CryptoBuiltInImpl.cpp
@@ -1386,7 +1386,6 @@ bool CryptoBuiltInImpl::encode_submessage(
   NativeCryptoHandle sender_handle,
   SecurityException& ex)
 {
-  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
   const KeyTable_t::const_iterator iter = keys_.find(sender_handle);
   if (iter == keys_.end()) {
     encoded_rtps_submessage = plain_rtps_submessage;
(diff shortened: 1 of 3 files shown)


dds/DCPS/security/CryptoBuiltInImpl.cpp
```diff
@@ -1552,6 +1549,7 @@ bool CryptoBuiltInImpl::encode_datareader_submessage(
   }
 
   NativeCryptoHandle encode_handle = sending_datareader_crypto;
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
   if (receiving_datawriter_crypto_list.length() == 1) {
     ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
```
mitza-oci, line 1554:
```suggestion
  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
  if (receiving_datawriter_crypto_list.length() == 1) {
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #3593, opened 2022-07-07
Fix deadlock involving SubscriberImpl lock

sonndinh:


dds/DCPS/SubscriberImpl.cpp
```diff
@@ -752,7 +746,7 @@ SubscriberImpl::begin_access()
   {
     ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex,
```
simpsont-oci, line 747: Given that `begin_access` and `end_access` are user-facing functions, I would expect them to lock `si_lock_` at the beginning of the function (before touching `enabled_`) but then releasing it before iterating over `to_call`. That seems like a pre-existing bug to me, which this PR probably ought to fix. And then we can limit our use of `dr_set_lock_` to the immediate context of where we touch `datareader_set_` (after we check `access_depth_`). The same pattern of changes should also be be made to `end_access`.

simpsont-oci (approved):

PR #3596, opened 2022-07-08
Implement More Aggressive ASAN Options

simpsont-oci:
We're currently not taking advantage of all the checks that AddressSanitizer can make. This PR updates our AddressSanitizer builds to do additional checks including stack use after return and initialization order checking, as well as a few fixes for issues discovered.

One of the issues discovered exists in ACE static member initialization, so this PR will also need [that fix](https://github.com/DOCGroup/ACE_TAO/pull/1879) in order to successfully pass all tests.

### DIFF
--- a/configure
+++ b/configure
@@ -1014,8 +1014,9 @@ my %all_sanitizers = (
     fsanitize => 'address',
     env => {
       LSAN_OPTIONS => "suppressions=$FindBin::RealBin/etc/asan-suppr.txt",
-      ASAN_OPTIONS => 'detect_leaks=1:fast_unwind_on_malloc=0',
+      ASAN_OPTIONS => 'detect_leaks=1:fast_unwind_on_malloc=0:strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1',
     },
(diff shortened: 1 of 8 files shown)


mitza-oci (approved):

configure
```diff
@@ -1014,8 +1014,9 @@ my %all_sanitizers = (
     fsanitize => 'address',
     env => {
       LSAN_OPTIONS => "suppressions=$FindBin::RealBin/etc/asan-suppr.txt",
-      ASAN_OPTIONS => 'detect_leaks=1:fast_unwind_on_malloc=0',
+      ASAN_OPTIONS => 'detect_leaks=1:fast_unwind_on_malloc=0:strict_string_checks=1:detect_stack_use_after_return=1::check_initialization_order=1:strict_init_order=1',
```
mitza-oci, line 1017:
```suggestion
      ASAN_OPTIONS => 'detect_leaks=1:fast_unwind_on_malloc=0:strict_string_checks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1',
```
I assume `::` is not significant for asan

simpsont-oci: I don't think so, but easy enough to fix.

PR #3601, opened 2022-07-11
Reduce TransportReceiveStrategy Memory Usage For RtpsUdp

simpsont-oci:
Problem: Our TransportReceiveStrategy allocates quite a bit of memory for RTPS (both initially and relative to each new packet). Especially for UDP-based protocols, there is no need for more than 64k bytes total of receive buffer (currently we allocate 16x that amount). Additionally for RTPS, many of the incoming packets will not need to be held because they contain non-data submessages (HEARTBEAT, ACKNACK, GAP, etc), so additional temporary allocations could be avoided by inspecting submessage headers.

Solution: Use a single 64K receive buffer for RTPS, check submessages headers for DATA and DATA_FRAG, and only allocate (correctly-sized) temporary buffers for these submessages which maybe held by the datalink.


dds/DCPS/transport/framework/ReceivedDataSample.h
```diff
@@ -32,7 +32,7 @@ namespace DCPS {
  */
 class OpenDDS_Dcps_Export ReceivedDataSample {
 public:
-  explicit ReceivedDataSample(ACE_Message_Block* payload);
+  ReceivedDataSample(ACE_Message_Block* payload, bool take_ownership = true);
```
mitza-oci, line 35:
```suggestion
  explicit ReceivedDataSample(ACE_Message_Block* payload, bool take_ownership = true);
```

dds/DCPS/transport/framework/ReceivedDataSample.inl
```diff
@@ -41,6 +43,9 @@ ACE_INLINE
 ReceivedDataSample::~ReceivedDataSample()
 {
   DBG_ENTRY_LVL("ReceivedDataSample", "~ReceivedDataSample", 6);
+  if (!owns_sample_) {
+    sample_.release();
```
mitza-oci, line 47: This was confusing when I first read it because `unique_ptr::release` and `ACE_Message_Block::release` are different things with the same name.  Certainly not your issue but maybe it could use a comment.

dds/DCPS/transport/framework/TransportReceiveStrategy_T.h
```diff
@@ -82,7 +75,8 @@ class TransportReceiveStrategy
   DSH& received_sample_header();
 
 protected:
-  explicit TransportReceiveStrategy(const TransportInst& config);
+  TransportReceiveStrategy(const TransportInst& config,
```
mitza-oci, line 78:
```suggestion
  explicit TransportReceiveStrategy(const TransportInst& config,
```

mitza-oci (commented):
partial review

mitza-oci (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -57,7 +57,193 @@ RtpsUdpReceiveStrategy::handle_input(ACE_HANDLE fd)
 {
   ThreadStatusManager::Event ev(TheServiceParticipant->get_thread_status_manager());
 
-  return handle_simple_dds_input(fd);
+  for (size_t index = 0; index < receive_buffers_.size(); ++index) {
```
jrw972, line 60: Do we need the loop since the size will be one?

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -57,7 +57,193 @@ RtpsUdpReceiveStrategy::handle_input(ACE_HANDLE fd)
...
+          ACE_Time_Value::zero,               // Default
+          ACE_Time_Value::max_time,           // Default
+          &db_allocator_,                     // Our data block cache
+          &mb_allocator_                      // Our message block cache
+        ),
+        -1);
+    }
+  }
+
+  ACE_Message_Block* cur_rb = receive_buffers_[buffer_index_];
```
jrw972, line 82: `buffer_index_` won't change and is always 0.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -57,7 +57,193 @@ RtpsUdpReceiveStrategy::handle_input(ACE_HANDLE fd)
...
+
+      // For the reassembly algorithm, the 'last_fragment_' header bit only
+      // applies to the first DataSampleHeader in the TransportHeader
+      receive_transport_header_.last_fragment(false);
+    }
+  }
+
+  finish_message();
+
+  // Attempt to quickly switch to unreferenced buffer for next read
```
jrw972, line 212: I think this can be removed.

jrw972 (changes requested):
Continue to simplify `handle_input`.

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -57,7 +57,186 @@ RtpsUdpReceiveStrategy::handle_input(ACE_HANDLE fd)
...
+
+  if (!pdu_remaining_) {
+    receive_transport_header_.length_ = static_cast<ACE_UINT32>(bytes_remaining);
+  }
+
+  receive_transport_header_ = *cur_rb;
+  if (!receive_transport_header_.valid()) {
+    cur_rb->reset();
+    if (DCPS_debug_level > 0) {
+      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: TransportHeader invalid.\n")));
```
sonndinh, line 135:
```suggestion
      ACE_DEBUG((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: RtpsUdpReceiveStrategy::handle_input: TransportHeader invalid.\n")));
```

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -57,7 +57,186 @@ RtpsUdpReceiveStrategy::handle_input(ACE_HANDLE fd)
...
+      // For the reassembly algorithm, the 'last_fragment_' header bit only
+      // applies to the first DataSampleHeader in the TransportHeader
+      receive_transport_header_.last_fragment(false);
+    }
+  }
+
+  finish_message();
+
+  // If newly selected buffer index still has a reference count, we'll need to allocate a new one for the read
+  if (receive_buffers_[INDEX]->data_block()->reference_count() > 1) {
```
jrw972, line 214: Is this possible with the new logic?  If it shouldn't happen, we may want to print an error/warning/notice.

jrw972 (approved):

PR #3602, opened 2022-07-11
ThreadPool Updates To Accommodate SafetyProfile

simpsont-oci:
Problem: Safety Profile Does Not Allow For pthread_join (ACE_OS::thr_join returns -1 (Not Supported)).

Solution: Switch from Barrier to Condition + Count. Add additional counter for exited threads and use condition for both startup and shutdown (when join is not available) of thread pool.

### DIFF
--- a/dds/DCPS/ThreadPool.h
+++ b/dds/DCPS/ThreadPool.h
@@ -10,9 +10,9 @@
 
 #include "dcps_export.h"
 
+#include "ConditionVariable.h"
 #include "PoolAllocator.h"
 
-#include <ace/Condition_Thread_Mutex.h>
 #include <ace/Thread.h>
 
 #if ! defined ACE_HAS_THREAD || (! defined ACE_HAS_STHREADS && defined ACE_HAS_PTHREADS && defined ACE_LACKS_PTHREAD_JOIN)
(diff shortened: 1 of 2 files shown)


dds/DCPS/ThreadPool.h
```diff
@@ -10,11 +10,25 @@
 
 #include "dcps_export.h"
 
+#include "ConditionVariable.h"
 #include "PoolAllocator.h"
 
-#include <ace/Barrier.h>
 #include <ace/Thread.h>
 
+#if defined ACE_HAS_THREADS
```
mitza-oci, line 18: Couldn't this be simplified?  We can assume the generic HAS_THREADS is there (or else nothing would work) and ignore the STHREADS for Solaris.  So OPENDDS_NO_THREAD_JOIN is only true if ACE has pthreads and lacks pthread join.

mitza-oci (changes requested):

dds/DCPS/ThreadPool.h
simpsont-oci: I tried a simpler version before and got it wrong so I just duplicated what I was seeing in OS_NS_Thread.cpp for when it was returning `NOTSUP`. I'll try just checking for pthreads and lacks join.

mitza-oci (approved):

PR #3603, opened 2022-07-11
map support

### LINKED ISSUE #3236: "map" support for IDL
The IDL documentation supports the use of "map" that corresponds to C++'s std::map object with the following language

`typedef map<T,V> M1;`

It appears that either TAO or OpenDDS (or both) don't support this type. Are there plans to implement this typing? And if not, what is the preferred solution when you want to create an IDL that has a property which is a list of key-value pairs? Do you suggest defining a struct for the key-value pairs and then using a sequence?

tmayoff:
This PR adds support for the map type in OpenDDS, this requires https://github.com/DOCGroup/ACE_TAO/pull/1842, to work and it closes #3236.

I managed to get a fair amount into this before running into issues which is why I created the PR to get some feedback.


dds/idl/dds_generator.h
```diff
@@ -409,6 +409,8 @@ namespace AstTypeClassification {
         ((dynamic_cast<AST_String*>(type)->max_size()->ev()->u.ulval == 0)
         ? 0 : CL_BOUNDED) |
         ((type->node_type() == AST_Decl::NT_wstring) ? CL_WIDE : 0);
+    case AST_Decl::NT_map:
```
iguessthislldo, line 412: To stop the errors, you can test if TAO has maps by updating this:

https://github.com/objectcomputing/OpenDDS/blob/51e2ed25da43f5a52ac1189756edffdc2e049482/dds/DCPS/Definitions.h#L69-L75

to match the changes you made to `idl_features.h` in TAO, then you should be able to use:

```suggestion
#if OPENDDS_HAS_MAP
    case AST_Decl::NT_map:
```
For IDL you can use the TAO_IDL macros directly like in explicit ints test:

https://github.com/objectcomputing/OpenDDS/blob/51e2ed25da43f5a52ac1189756edffdc2e049482/tests/DCPS/Compiler/explicit_ints/test.idl#L1-L3

jwillemsen:
Looks this PR is necessary to get OpenDDS to compile again with ACE/TAO master, see https://scoreboard.opendds.org/dds.html

tmayoff:
The map tests pass on my machine (not sure if they're being run in the CI), however, [this one](https://github.com/OpenDDS/OpenDDS/pull/3603/files#diff-885ad75ef2f2072bbc2c354eb33ca657cd5ddd04ef6431e793208502d5c48b53) only works over rtps which I'm not sure why

iguessthislldo:
> Looks this PR is necessary to get OpenDDS to compile again with ACE/TAO master, see https://scoreboard.opendds.org/dds.html

This should be fixed by https://github.com/DOCGroup/ACE_TAO/pull/2100

tmayoff:
I think this is ready for review, I believe the CI is failing because of the cached versions of ACE & TAO, but other than that (and the other issue I mentioned before), it should be ready

jwillemsen:
Most CI uses the latest ACE6TAO2 release, you should adapt your test to also compile when using this older release which lacks map support, some ifdefs should do the trick, making the test just a noop test when there is no map support

dds/DCPS/Definitions.h
```diff
@@ -89,6 +89,11 @@
 #  define OPENDDS_GCC_HAS_DIAG_PUSHPOP 0
 #endif
 
+#if defined TAO_HAS_IDL_FEATURES && TAO_HAS_IDL_FEATURES
+#  include <tao/idl_features.h>
+#define OPENDDS_HAS_MAP TAO_IDL_HAS_MAP
```
iguessthislldo, line 94:
```suggestion
#  define OPENDDS_HAS_MAP TAO_IDL_HAS_MAP
```

dds/idl/dds_visitor.cpp
```diff
@@ -476,6 +479,14 @@ dds_visitor::visit_sequence(AST_Sequence*)
...
 }
 
+int
+dds_visitor::visit_map(AST_Map*)
+{
+  //maps always appear as typedefs, see visit_typedef ()
+  return 0;
+}
+
+
```
iguessthislldo, line 489:
```suggestion
#if OPENDDS_HAS_MAP
int
dds_visitor::visit_map(AST_Map*)
{
  // maps always appear as typedefs, see visit_typedef ()
  return 0;
}
#endif

```

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,366 @@ namespace {
...
+    UTL_ScopedName* tdname, AST_Map* map, bool nested_key_only, const FieldInfo* anonymous = 0)
+  {
+    // be_global->add_include("dds/DCPS/Util.h");
+    // be_global->add_include("dds/DCPS/Serializer.h");
+    if (anonymous) {
+      map = dynamic_cast<AST_Map*>(anonymous->type_);
+    }
+    const std::string named_as = anonymous ? anonymous->scoped_type_ : scoped(tdname);
+    RefWrapper base_wrapper(map, named_as, "map");
+    base_wrapper.nested_key_only_ = nested_key_only;
```
iguessthislldo, line 923: `RefWrapper` needs to be finalized for it can properly decide what it needs to return. Otherwise it can cause an assert like it's doing now in a least on of the CI jobs.

```suggestion
    RefWrapper base_wrapper(map, named_as, "map");
    base_wrapper.nested_key_only_ = nested_key_only;
    base_wrapper.done();
```

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,366 @@ namespace {
...
+    NamespaceGuard ng(!anonymous);
+
+    // if (!anonymous) {
+    //   for (size_t i = 0; i < array_count(special_sequences); ++i) {
+    //     if (special_sequences[i].check(base_wrapper.type_name_)) {
+    //       special_sequences[i].gen(base_wrapper.type_name_);
+    //       return;
+    //     }
+    //   }
+    // }
```
iguessthislldo, line 934: You don't have to worry about this with maps.

```suggestion
```

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,366 @@ namespace {
...
+
+    AST_Type* key = resolveActualType(map->key_type());
+    AST_Type* val = resolveActualType(map->value_type());
+    // TryConstructFailAction try_construct = be_global->sequence_element_try_construct(map);
```
iguessthislldo, line 938: You will need to be able to get try_construct annotation values, I think from both the key and value types.

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,366 @@ namespace {
...
+      //   if (be_global->language_mapping() != BE_GlobalData::LANGMAP_FACE_CXX &&
+      //       be_global->language_mapping() != BE_GlobalData::LANGMAP_SP_CXX) {
+      //     const std::string hdr = "dds/CorbaSeq/" + nameOfSeqHeader(key) + "SeqTypeSupportImpl.h";
+      //     be_global->conditional_include(hdr.c_str(), BE_GlobalData::STREAM_CPP,
+      //                                    "#ifndef OPENDDS_SAFETY_PROFILE");
+      //   }
+      // } else {
+      // be_global->add_referenced(key->file_name().c_str());
+      // }
+    }
```
iguessthislldo, line 957: I don't think this would apply to maps.

```suggestion
```

iguessthislldo (changes requested):
This is a partial review.

tmayoff:
@iguessthislldo, now that the cmake build stuff has been merged back into main, I'd like to use it for the for testing, how can I add the map tests to the cmake build?

iguessthislldo:
> how can I add the map tests to the cmake build

The goal of the recent CMake changes was mainly just to get the main OpenDDS library and utilities built. Converting tests was put off for later. However there's nothing stopping you from adding the tests if you want to and I won't ask you to remove them because they would have to be added eventually.

I was planning on trying to better organize this when working on the tests in earnest, but for right now they can go here:

https://github.com/OpenDDS/OpenDDS/blob/7c1c3c48978479b3bc445d8b44daaf1879b9a5f5/CMakeLists.txt#L140-L150

As you can see there's not many, but add you can add the test directories here.

You can base the tests' `CMakeLists.txt` off of [`DevGuideExamples/DCPS/Messenger/CMakeLists.txt`](https://github.com/OpenDDS/OpenDDS/blob/master/DevGuideExamples/DCPS/Messenger/CMakeLists.txt), but there's probably a few things you'd need to change. Remove the part starting with "Make sure the MPC-generated headers are gone", as I don't think we need to have that in all the tests. Change the project names to something like `opendds_test_maps` and `opendds_test_compiler_maps`. The testing parts will also probably have to change. There's probably other things, but disregarding the DevGuide Messenger I've never converted a normal test to CMake yet, so you're breaking new ground.

>  also can I specify the version of ACE_TAO used with cmake?

I'm assuming you mean locally and not on GitHub Actions. By default it checks out a fixed version, [7.1.1](https://github.com/OpenDDS/OpenDDS/blob/7c1c3c48978479b3bc445d8b44daaf1879b9a5f5/cmake/get_ace_tao.cmake#L8) right now, so I guess it wouldn't have maps support because I don't think we've had ACE/TAO releases yet. To change that you can either build ACE/TAO separately or use an existing build and pass ACE using [`-DOPENDDS_ACE=<path to ACE>`](https://opendds.readthedocs.io/en/master/devguide/building/cmake.html#var-OPENDDS_ACE). It should be able to pick up on how to use the ACE. You can also pass the source using [`-DOPENDDS_ACE_TAO_SRC=<path to ACE_TAO>`](https://opendds.readthedocs.io/en/master/devguide/building/index.html#var-OPENDDS_ACE_TAO_SRC) and it will attempt to build it for you. Thinking about it, it might good to add something like `--ace-github-latest`, but maybe `-DOPENDDS_ACE_TAO_COMMIT=<branch or commit>`, so it's more flexible.

dds/idl/marshal_generator.cpp
tmayoff: What exactly are try_construct annotations?

mitza-oci: `@try_construct` from the XTypes spec

tmayoff:
Yeah, I started looking into it although haven't gotten anywhere with it yet. 

I feel like this PR might be coming to an end and wanted to get started with some more reviews.

dds/idl/value_reader_generator.cpp
```diff
@@ -103,6 +103,52 @@ namespace {
...
+  {
+    const bool use_cxx11 = be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+    const std::string indent(level * 2, ' ');
+    const std::string kidx = "k" + idx;
+    be_global->impl_ <<
+      indent << "if (!value_reader.begin_map()) return false;\n" <<
+      indent << "for (" << (use_cxx11 ? "size_t " : "unsigned int ") << idx << " = 0; "
+        "value_reader.pairs_remaining(); ++" << idx << ") {\n";
+
+    std::string key_name = scoped(map->key_type()->name());
```
mitza-oci, line 118:
```suggestion
    const std::string key_name = scoped(map->key_type()->name());
```
make anything that can be const const

dds/idl/typeobject_generator.cpp
```diff
@@ -1353,6 +1362,83 @@ typeobject_generator::generate_array_type_identifier(AST_Type* type, bool force_
...
+
+  const OpenDDS::XTypes::TypeIdentifier minimal_key_ti = get_minimal_type_identifier(n->key_type());
+  const OpenDDS::XTypes::TypeIdentifier minimal_val_ti = get_minimal_type_identifier(n->value_type());
+  const OpenDDS::XTypes::TypeIdentifier complete_key_ti = get_complete_type_identifier(n->key_type());
+  const OpenDDS::XTypes::TypeIdentifier complete_val_ti = get_complete_type_identifier(n->value_type());
+
+  if (be_global->is_plain(type) && !force_type_object) {
+    const OpenDDS::XTypes::EquivalenceKind key_minimal_ek =
+      OpenDDS::XTypes::is_fully_descriptive(minimal_key_ti) ? OpenDDS::XTypes::EK_BOTH : OpenDDS::XTypes::EK_MINIMAL;
+    const OpenDDS::XTypes::EquivalenceKind val_minimal_ek =
```
mitza-oci, line 1398: warnings here for an unused local

dds/idl/typeobject_generator.cpp
```diff
@@ -1353,6 +1362,83 @@ typeobject_generator::generate_array_type_identifier(AST_Type* type, bool force_
   }
 }
 
+#if OPENDDS_HAS_MAP
+void
+typeobject_generator::generate_map_type_identifier(AST_Type* type, bool force_type_object)
+{
+  AST_Map* const n = dynamic_cast<AST_Map*>(type);
+
+  // I'm not sure if this is setup for maps
```
mitza-oci, line 1371: Remove this?  Also see TODO on 1381

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,422 @@ namespace {
...
+      } else {
+        elem_type_name = use_cxx11 ? "std::string" : "CORBA::String_var";
+      }
+    }
+
+   std::string tempvar = "tempvar";
+   // be_global->impl_ <<
+   //    indent << "if (encoding.xcdr_version() == Encoding::XCDR_VERSION_2) {\n" <<
+   //    indent << "  strm.skip(end_of_map - strm.rpos());\n" <<
+   //    indent << "} else {\n";
```
mitza-oci, line 931: Remove or resolve?

dds/idl/marshal_generator.cpp
```diff
@@ -3187,6 +3614,8 @@ marshal_generator::gen_field_getValueFromSerialized(AST_Structure* node, const s
           "        throw std::runtime_error(\"Field '" + idl_name + "' could not be skipped\");\n"
           "      }\n"
           "    }\n";
+    } else if (fld_cls & CL_MAP) {
+      // TODO(tyler) fill this out
```
mitza-oci, line 3618: TODO here

mitza-oci (changes requested):
This is a partial review - I didn't get to all the details in marshal_generator.cpp and I didn't review tests.  

I marked some places with TODO comments in the review since that may help to get visibility and we can track them in the review.

dds/idl/value_writer_generator.cpp
```diff
@@ -157,6 +184,13 @@ namespace {
       array_helper(expression, array, 0, idx, level);
       return;
     }
+#if OPENDDS_HAS_IDL_MAP
+    if (c & CL_MAP) {
+      AST_Map * const map = dynamic_cast<AST_Map*>(actual);
```
mitza-oci, line 189:
```suggestion
      AST_Map* const map = dynamic_cast<AST_Map*>(actual);
```

tests/DCPS/Compiler/maps/maps.mpc
```diff
@@ -0,0 +1,12 @@
+project: dcps_test, opendds_cxx11, googletest, msvc_bigobj {
+  exename = maps
+  includes += .
+  macros += CPP11_MAPPING=1
```
mitza-oci, line 4: This may not be needed if it's always using the C++1 language mapping

tests/DCPS/Compiler/maps/test.idl
```diff
@@ -0,0 +1,34 @@
...
+    struct TestStruct {
+        int32 id;
+        string msg;
+    };
+
+    enum TEST_ENUM {TEST1, TEST2, TEST3};
+
+    typedef sequence<TestStruct> TestStructSeq;
+    typedef map<int32, TestStruct> TestStructMap;
+    // typedef map<int32, string> TestIntStringMap; //TODO Not compiling
```
mitza-oci, line 13: Does this need to resolved before the PR is merged?

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,422 @@ namespace {
...
+
+      be_global->impl_ <<
+        "  for (auto it = " << value_access <<  ".begin(); it != " << value_access << ".end(); ++it) {\n";
+
+      if (key_cls & CL_INTERFACE || val_cls & CL_INTERFACE) {
+        be_global->impl_ <<
+          "  return false; // map with either a key or value of objrefs is not marshaled\n";
+      } else if (key_cls & CL_UNKNOWN || val_cls & CL_UNKNOWN) {
+        be_global->impl_ <<
+          "  return false; // map with either key or value of unknown/unsupported type\n";
```
mitza-oci, line 1159: It would probably be helpful for users to indicate if the key or the value type is the problem -- unless this is handled somewhere else?

dds/idl/marshal_generator.cpp
```diff
@@ -909,6 +909,422 @@ namespace {
...
+          val_generated = true;
+        }
+
+        // Complex Types
+        if (!key_generated || !val_generated) {
+          be_global->impl_ <<
+            "  for (auto it = " << value_access << ".begin(); it != " << value_access << ".end(); ++it) {\n";
+          const std::string indent = "    ";
+
+          if (!key_generated) { // Key
```
mitza-oci, line 1062: Would it work to refactor the "Key" and "Value" sections to be calls to a common function?

mitza-oci (changes requested):

tmayoff:
> > Yeah, I started looking into it although haven't gotten anywhere with it yet.
> 
> Looks like that's a tao_idl issue so it may not directly impact this PR.
I did a git bisect to see where it started, and this was the first bad commit: [`a52b816` (#3603)](https://github.com/OpenDDS/OpenDDS/pull/3603/commits/a52b81655c027af38b5721435fe85bbcf4931627), which is a merge of master into this branch

mitza-oci:
> > > Yeah, I started looking into it although haven't gotten anywhere with it yet.
> > 
> > 
> > Looks like that's a tao_idl issue so it may not directly impact this PR.
> 
> I did a git bisect to see where it started, and this was the first bad commit: [`a52b816` (#3603)](https://github.com/OpenDDS/OpenDDS/pull/3603/commits/a52b81655c027af38b5721435fe85bbcf4931627), which is a merge of master into this branch

Is that the first time the CI ran with a new commit in ACE_TAO?

I'm not sure if GitHub keeps this much history of the actions / workflows to validate this.

tmayoff:
This is the asan CI run that first failed, https://github.com/OpenDDS/OpenDDS/actions/runs/6723902243/job/18275112894?pr=3603

tests/DCPS/Compiler/maps/test.idl
tmayoff: The TODO may have been an artifact from something else works fine now

PR #3607, opened 2022-07-12
Added unit tests for RcHandle and RcObject

mitza-oci:


jrw972 (approved):

mitza-oci:
Looks like there are some leaks in the new tests, looking into it now...

PR #3611, opened 2022-07-14
Add Initial UndefinedBehaviorSanitizer (UBSAN) Jobs

simpsont-oci:
Unit tests only, for now. Will incorporate (prior?) fixes once I see what's actually failing in the unit tests. The only default check from 'undefined' group that has been disabled is the check for invalid enum values, since that will take some thought and design changes for OpenDDS's code generation (and may not be worth the benefits).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -3919,6 +3919,7 @@ jobs:
     - name: configure OpenDDS
       run: |
         cd OpenDDS
+        sed -i 's/DevGuideExamples/tests\/unit-tests/' DDS_TAOv2.mwc
         ./configure \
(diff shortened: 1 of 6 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -3790,6 +3793,295 @@ jobs:
...
+    - name: checkout MPC
+      uses: actions/checkout@v3
+      with:
+        repository: DOCGroup/MPC
+        path: MPC
+    - name: checkout autobuild
+      uses: actions/checkout@v3
+      with:
+        repository: DOCGroup/autobuild
+        ref: update_prettify_for_ubsan
```
jrw972, line 3986: Change to master.

jrw972 (changes requested):

jrw972 (approved):

PR #3624, opened 2022-07-20
Coverage script executes rm without an argument

jrw972:
Problem
-------

The coverage script executes `rm` in a way that it will not have an
argument.

Solution
--------

Check the input to `rm`.

### DIFF
--- a/tools/scripts/unit_test_coverage.sh
+++ b/tools/scripts/unit_test_coverage.sh
@@ -10,13 +10,17 @@
 # single unit, the script can be used to geneerate the coverage for
 # that unit in isolation.
 
-set -e
-
 start=$(date +'%s')
 
 args=($@)
 
 cd "${DDS_ROOT}"
+
+function cleanup {
+    rm -f baseline current files_to_keep files_to_remove
+}
+
+trap cleanup EXIT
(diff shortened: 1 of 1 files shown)


tools/scripts/unit_test_coverage.sh
```diff
@@ -97,7 +97,11 @@ do
     # Delete gcda files that did not change.
     find . -name "*.gcda" | xargs md5sum | sort -k2,2 > current
     diff baseline current | grep '^<' | cut -d' ' -f4 > files_to_keep
-    find . -name "*.gcda" | grep -v -F -f files_to_keep | xargs rm
+    find . -name "*.gcda" | grep -v -F -f files_to_keep > files_to_remove
+    if [ -s files_to_remove ]
+    then
+        cat files_to_remove | xargs rm
+    fi
```
simpsont-oci, line 104: remove `files_to_remove` once you're done with it?

mitza-oci: When I ran the script it left these files:
baseline
current
files_to_keep
files_to_remove

mitza-oci (approved):

PR #3630, opened 2022-07-21
Split Out Service_Participant Protection Of shut_down_ Flag To Avoid Deadlock

simpsont-oci:
Problem: See https://github.com/objectcomputing/OpenDDS/runs/7439373375?check_suite_focus=true

Solution: Break out protection of shut_down_ flag to allow read access without waiting on factor lock.

### DIFF
--- a/dds/DCPS/Service_Participant.h
+++ b/dds/DCPS/Service_Participant.h
@@ -474,6 +474,9 @@ public:
 
 private:
 
+  /// Track shut down state
+  void set_shut_down(bool shut_down);
+
   /// Initialize default qos.
   void initialize();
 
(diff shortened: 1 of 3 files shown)


dds/DCPS/Service_Participant.h
```diff
@@ -761,6 +764,9 @@ class OpenDDS_Dcps_Export Service_Participant {
   /// Used to track state of service participant
   bool shut_down_;
 
+  /// Protects the internal shutdown state tracking
+  mutable ACE_Thread_Mutex shut_down_lock_;
+
```
jwillemsen, line 769: Why not make the shutdown flag an atomic?

simpsont-oci: Atomic flags are great, until you need two of them to represent your state. Without knowing more about what state is being tracked in Service_Participant, this seemed like the safer option. If we're confident it's just the one flag and it doesn't need to be coordinated with any other data structures, sure, I can switch to an atomic.

mitza-oci (approved):
Though I agree that atomic may be a good mechanism for this, the PR is an improvement on what's currently there and fixes a known issue.  So updating to atomic can come later.

PR #3632, opened 2022-07-22
Detach Conditions From WaitSets in Tests

simpsont-oci:


mitza-oci (approved):

tests/DCPS/Messenger/subscriber.cpp
```diff
@@ -176,6 +176,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
 
     DDS::ConditionSeq conditions;
     ret = ws->wait(conditions, timeout);
+    ws->attach_condition(gc);
```
sonndinh, line 179:
```suggestion
    ws->detach_condition(gc);
```

simpsont-oci: woops, good catch!

PR #3633, opened 2022-07-22
OpenDDS applications hang when dynamic library loading fails

jrw972:
Problem
-------

A user that fails to set the path to the OpenDDS shared libraries will
experience a deadlock when starting an application.  As part of #3342,
a condition variable was added that allows one thread to wait while
another thread loads the required transport type.  In the case of the
missing path, the wait will wait forever leading to the deadlock.

Solution
--------

* Preemptively load all of the shared libraries known by OpenDDS.
  This is currently done in the initialization of the Service
  Participant.

* Remove the condition variable.

Note
----

For dynamic libraries, it is theoretically okay for a
transport-like-thing, e.g., InfoRepoDiscovery, to depend on another
transport-like-thing, e.g., TCP transport, because the load order of
the libraries can be adjusted to accommodate the dependency.  However,
this may not work for static libraries because the load order is
indeterminate.  Thus, it may be necessary to refactor
InfoRepoDiscovery so its static initializer doesn't depend on TCP
being loaded.

### DIFF
--- a/dds/DCPS/transport/framework/TransportRegistry.cpp
+++ b/dds/DCPS/transport/framework/TransportRegistry.cpp
@@ -17,7 +17,7 @@
 #include "dds/DCPS/EntityImpl.h"
 #include "dds/DCPS/SafetyProfileStreams.h"
 #include <dds/DdsDcpsInfrastructureC.h>
-
+#include "dds/DCPS/debug.h"
 
 #include "ace/Singleton.h"
 #include "ace/OS_NS_strings.h"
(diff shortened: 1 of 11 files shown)


dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -71,20 +64,35 @@ const OPENDDS_STRING TransportRegistry::CUSTOM_ADD_DOMAIN_TO_PORT = "add_domain_
...
-  lib_directive_map_["udp"]       = "dynamic OpenDDS_Udp Service_Object * OpenDDS_Udp:_make_UdpLoader()";
-  lib_directive_map_["multicast"] = "dynamic OpenDDS_Multicast Service_Object * OpenDDS_Multicast:_make_MulticastLoader()";
-  lib_directive_map_["rtps_udp"]  = "dynamic OpenDDS_Rtps_Udp Service_Object * OpenDDS_Rtps_Udp:_make_RtpsUdpLoader()";
-  lib_directive_map_["shmem"]     = "dynamic OpenDDS_Shmem Service_Object * OpenDDS_Shmem:_make_ShmemLoader()";
+namespace {
+void load_dynamic_library(const char* directive)
+{
+  const ACE_TString d = ACE_TEXT_CHAR_TO_TCHAR(directive);
+  if (ACE_Service_Config::process_directive(d.c_str()) != 0) {
+    ACE_DEBUG((LM_NOTICE,
```
iguessthislldo, line 77: Shouldn't this be either an [error or info since we can't give feedback to the user](https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#logging-conditions-and-priority)?

jrw972: It's not an error if they chose to not build the library.

iguessthislldo: That's true, but in the guidelines we decided to pair `LM_NOTICE` with `ACE_ERROR` and give it a specific meaning as "reportable error".  So maybe this an info level message?

PR #3640, opened 2022-07-27
TSAN Fix And Remove Suppression For Lock-Order-Invserion Issues In Message_Block_Deleter

simpsont-oci:
Problem: Currently we suppress lock-order-inversion errors from ACE_Message_Block::release_no_delete, which seem to be caused by ACE_Message_Block chains with differently ordered locking_strategy_ pointers being called from different threads. In general, this is a problem with the locking strategy design of ACE_Message_Block, since ACE_Message_Block tries to acquire the locks of the chain sequentially as it follows the chain during a release.

Solution: Rather than trying to fundamentally change the design of ACE, which might break other users, we can optimize the behavior of Message_Block_Deleter to address the assumptions of OpenDDS, which is namely that it's acceptable to de-chain and individually release ACE_Message_Blocks, so they can lock and unlock their individual locking_strategy_ without introducing lock ordering issues. The PR also makes TransportSendStrategy's SendMode atomic to address a TSAN complaint that showed up during testing.

### DIFF
--- a/dds/DCPS/Message_Block_Ptr.h
+++ b/dds/DCPS/Message_Block_Ptr.h
@@ -17,6 +17,8 @@ namespace DCPS {
 struct Message_Block_Deleter
 {
   void operator()(ACE_Message_Block* ptr) const {
+    // In order to avoid locking order issues for message blocks with different locking strategies,
+    // it is safer to unlink elements in the chain before releasing them individually
     while (ptr) {
       ACE_Message_Block* cont = ptr->cont();
       ptr->cont(0);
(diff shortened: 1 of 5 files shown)


mitza-oci (changes requested):

dds/DCPS/Message_Block_Ptr.h
```diff
@@ -17,7 +17,12 @@ namespace DCPS {
 struct Message_Block_Deleter
 {
   void operator()(ACE_Message_Block* ptr) const {
-    ACE_Message_Block::release(ptr);
+    while (ptr) {
```
simpsont-oci, line 20:
```suggestion
    // In order to avoid locking order issues for message blocks with different locking strategies,
    // it is safer to unlink elements in the chain before releasing them individually
    while (ptr) {
```

mitza-oci (approved):
Are there uses of ACE_Message_Block::release which don't go through Message_Block_Ptr that we need to look at?

simpsont-oci:
TSAN should report them if they're problematic, now that we've removed the suppression.

PR #3643, opened 2022-07-28
Remove 'RecursiveDispatch' Unit Tests From valgrind CI Test Job

simpsont-oci:
Problem: They sometimes take too long under valgrind, and as somewhat 'stressful' tests they're not well suited to timely and predictable execution under valgrind.

Solution: Keep them in the unit tests, but don't run them under valgrind.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4157,7 +4157,7 @@ jobs:
         cd tests/unit-tests
         export ACE_RUN_VALGRIND_CMD="valgrind --leak-check=yes --show-leak-kinds=definite --errors-for-leak-kinds=definite --num-callers=50 --error-exitcode=86"
         export ACE_RUNTEST_DELAY=10
-        ./run_test.pl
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -4157,7 +4157,7 @@ jobs:
         cd tests/unit-tests
         export ACE_RUN_VALGRIND_CMD="valgrind --leak-check=yes --show-leak-kinds=definite --errors-for-leak-kinds=definite --num-callers=50 --error-exitcode=86"
         export ACE_RUNTEST_DELAY=10
-        ./run_test.pl
+        ./run_test.pl --gtest_filter=*-*RecursiveDispatch*
```
mitza-oci, line 4160: Do the `*` characters need to be protected from shell wildcard expansion?  I guess there's no way anything could match that...

simpsont-oci: Not sure, but I did run it locally to test and it seemed to work as expected.

mitza-oci (approved):

PR #3644, opened 2022-07-28
TransportRegistry::create_inst removed wait param

mitza-oci:
Also, return an error if we try to load a library via ACE and it fails.


jrw972 (approved):
There is a problem with this code that doesn't need to be fixed right now.  Since loading uses a reverse lock, it creates the possibility that two threads creating an inst at the same time will create an issue where one will unlock, load, relock, while the other will get a null inst.

PR #3652, opened 2022-08-01
Update TransportClient's use of PendingAssoc's lock

sonndinh:
While working on the ConfigTransports test's timeout error, a number of places in TransportClient were found where the PendingAssoc's lock was not acquired properly. This PR adds the PendingAssoc's lock in those places.


dds/DCPS/transport/framework/TransportClient.h
```diff
@@ -206,10 +208,11 @@ class OpenDDS_Dcps_Export TransportClient
     WeakRcHandle<TransportClient> client_;
 
     explicit PendingAssoc(TransportClient* tc)
-      : active_(false)
+      : reverse_mutex_(mutex_)
+      , active_(false)
       , scheduled_(false)
       , blob_index_(0)
-      , client_(RcHandle<TransportClient>(tc, inc_count()))
+      , client_(*tc)
```
jrw972, line 215: I suggest changing the constructor argument to a `RcHandle<TransportClient>`.  At the place where the PendingAssoc is constructed, you may need to use `rchandle_from(this)`.

jrw972 (changes requested):

dds/DCPS/transport/framework/TransportClient.h
mitza-oci: I suggested the change to `*tc` but using RcHandle as the argument type is probably cleaner.  There is a WeakRcHandle constructor from `T&` which should now be covered in the unit test.  I think it's doing the right thing in terms of the refcount but it would be good to check.

mitza-oci (changes requested):

mitza-oci (approved):

jrw972 (approved):

PR #3653, opened 2022-08-02
32-bit Fixes For MemoryPool Unit Test

simpsont-oci:


jrw972 (approved):

tests/unit-tests/dds/DCPS/MemoryPool.cpp
```diff
@@ -1306,14 +1306,14 @@ void test_too_large_find() {
   void setup(FreeIndex& index, size_t free_size = 1024) {
     memset(pool_ptr_, 0, sizeof(pool_ptr_));
     next_alloc_ = pool_ptr_ + (sizeof(AllocHeader*) - reinterpret_cast<uint64_t>(pool_ptr_) % sizeof(AllocHeader*));
-    EXPECT_EQ(reinterpret_cast<uint64_t>(next_alloc_) % 8, 0u);
+    EXPECT_EQ(reinterpret_cast<uint64_t>(next_alloc_) % sizeof (AllocHeader*), 0u);
```
mitza-oci, line 1309: A number of build have problems with `uint64_t` here

mitza-oci (changes requested):

tests/unit-tests/dds/DCPS/MemoryPool.cpp
simpsont-oci: Where are you seeing problems? I don't see them on scoreboard or in CI...

mitza-oci: I think it was the older Visual Studio versions.  Compilers that are new enough to have them do require `<cstdint>` to use these types.  But that may not be a solution to the portability problem.

mitza-oci (approved):

PR #3657, opened 2022-08-03
Fix gcc 4.4 Docker Build

simpsont-oci:
Problem: The UBSAN build changes caused issues with the gcc 4.4 (non-c++11) build on CI. The configure script has a bug which inadvertently turns on optimization for this build, which appears to repeatedly hit an internal compiler bug in gcc 4.4.

Solution: Fix the configure script to avoid `-O3` flag, add symlink to dockerfile to quiet erroneous "no g++" messages, and fix a few low-hanging compiler warnings.

### DIFF
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -903,14 +903,12 @@ bool Serializer::read_delimiter(size_t& size)
 {
   if (encoding().xcdr_version() == Encoding::XCDR_VERSION_2) {
     ACE_CDR::ULong dheader;
-    if (!(*this >> dheader)) {
-      return false;
+    if (*this >> dheader) {
+      size = dheader;
+      return true;
     }
-    size = dheader;
-  } else {
-    size = 0;
   }
-  return true;
+  return false;
 }
 
 ACE_INLINE
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -899,6 +899,15 @@ Serializer::align_cont_w()
 }
 
 ACE_INLINE
+bool Serializer::skip_delimiter()
+{
+  if (encoding().xcdr_version() == Encoding::XCDR_VERSION_2) {
+    return skip(uint32_cdr_size);
+  }
+  return true;
+}
+
+ACE_INLINE
 bool Serializer::read_delimiter(size_t& size)
 {
   if (encoding().xcdr_version() == Encoding::XCDR_VERSION_2) {
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -907,6 +907,8 @@ bool Serializer::read_delimiter(size_t& size)
       return false;
     }
     size = dheader;
+  } else {
+    size = 0;
   }
   return true;
 }
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -3261,7 +3261,7 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ${{ github.job }}.tar.xz
-        key: c01_${{ github.job }}_ace6tao2_${{ env.ACE_COMMIT }}_${{ env.COMPILER_VERSION }}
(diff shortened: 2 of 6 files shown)


mitza-oci (approved):

dds/DCPS/Serializer.inl
```diff
@@ -903,12 +903,12 @@ bool Serializer::read_delimiter(size_t& size)
...
-    if (!(*this >> dheader)) {
-      return false;
+    if (*this >> dheader) {
+      size = dheader;
+      return true;
     }
-    size = dheader;
   }
-  return true;
+  return false;
```
sonndinh, line 911: I didn't follow the discussion on this, but why does the return value for XCDR1 change from `true` to `false`?

mitza-oci: There is no delimiter so it can't be read, and the return value of `false` lets the caller know that `size` was not assigned to.

In both DynamicData and the `opendds_idl`-generated code this is not called unless the encoding type is XCDR2.

simpsont-oci: The initial reason for any change was to avoid a compiler warning about a calling function potentially using the `size` value passed to the function uninitialized. As in:
```
size_t dheader;
if (strm_.read_delimiter(dheader)) {
  do_something_else_with(dheader);
}
```
So the initial change simply set `size` to 0 when the encoding wasn't XCDR2.
But then Adam and I had a conversation about whether `read_delimiter` shouldn't just return false in that case, since it's not, in fact, reading a delimiter from the stream. And it looked like we were always checking for XCDR2 before calling it, so it seemed safe to change the behavior... but then the unit tests started failing.

sonndinh: When I worked on DynamicData, XCDR1 is partially supported. That's why there're unit test cases with XCDR1.

In DynamicData, it doesn't check whether XCDR1 or XCDR2 is used before calling `read_delimiter`, and so it could be called with both encoding kinds. Though, I agree that some places in DynamicData need to initialize the size argument passed to `read_delimiter` to properly handle cases when XCDR1 is used.

sonndinh: I do see the pros and cons of each approach. If we keep the current approach, the caller will need to initialize the size parameter to 0 in case XCDR1 is used which is error-prone. If we use the new approach and the method return `false`, then the caller must check whether it's because XCDR1 is used or XCDR2 is used but reading delimiter actually failed, because returning `false` due to XCDR1 doesn't indicate an error if both XCDR1 and 2 are allowed.

mitza-oci: At the risk of expanding scope...

It seems like there are different use cases for `read_delimiter` which means we should have different methods.

If the caller does this:
```
size_t dheader;
if (!strm_.read_delimiter(dheader)) {
```
...and then never uses the `dheader` value after the call to `read_delimiter`, consider changing it to use a `bool skip_delimiter()` (which doesn't exist yet) and returns true as long as the stream is valid and skips over a delimiter if there is one.

simpsont-oci: Would we just then be checking the encoding to see whether we try to call `skip_delimiter()` or `read_delimiter()`? Should we just make a single `read_or_skip_delimiter()` which only fails when it's expected by the encoding but not readable (i.e. what we had before)?

mitza-oci: That's not quite what I'm suggesting.  There are uses cases (especially in DynamicData, maybe elsewhere) where the calling code doesn't need to know the value of the delimiter or even if there is a delimiter present.  Make that operation its own method simplifies the calling code and avoids spurious warnings from old compilers.

sonndinh: I'm good with having a separate `skip_delimiter` function for skipping purpose. There are places in DynamicData where `read_delimiter` is used just to go past the delimiter's bytes. So if we don't care if there is a delimiter (i.e., XCDR2) or not (i.e., XCDR1), or don't care about its value if it's present, we can use `skip_delimiter`. That seems to suggest using `read_delimiter` only when XCDR2 is used.

PR #3686, opened 2022-08-17
Enums do not have extensibility set

jrw972:
Problem
-------

The IDL compiler does not set the extensibility for an enum.

Solution
--------

Set the extensibility in the IDL compiler.  For backwards
compatibility, 1) the extensibility will be set on received
TypeObjects and 2) the IDL compiler will have an option for not
setting the extensibility.

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscovery.h
+++ b/dds/DCPS/RTPS/RtpsDiscovery.h
@@ -623,8 +623,8 @@ public:
         return true;
       }
     }
-    if (log_level >= LogLevel::Notice) {
-      ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: RtpsDiscoveryConfig::use_xtypes: "
+    if (log_level >= LogLevel::Warning) {
+      ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: RtpsDiscoveryConfig::use_xtypes: "
                  "invalid XTypes configuration: %C\n", str));
     }
     return false;
--- a/dds/DCPS/XTypes/TypeLookupService.cpp
+++ b/dds/DCPS/XTypes/TypeLookupService.cpp
@@ -132,7 +132,10 @@ void TypeLookupService::add_type_objects_to_cache(const TypeIdentifierTypeObject
   for (unsigned i = 0; i < types.length(); ++i) {
     const TypeMap::iterator pos = type_map_.find(types[i].type_identifier);
     if (pos == type_map_.end()) {
-      type_map_.insert(std::make_pair(types[i].type_identifier, types[i].type_object));
+      TypeObject to = types[i].type_object;
+      if (set_type_object_defaults(to)) {
+        type_map_.insert(std::make_pair(types[i].type_identifier, to));
+      }
     }
   }
 }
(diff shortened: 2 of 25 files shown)


dds/DCPS/Service_Participant.cpp
```diff
@@ -97,6 +97,8 @@ void set_log_verbose(unsigned long verbose_logging)
 
 }
 
+extern char **environ;
```
iguessthislldo, line 100: I didn't do an extensive search on this, but I'm not sure this is standard, as hinted by the warnings. Is there something in ACE we can use? If not, it's not pretty as it looks like it would have to be parsed, but there is [a Win32 call for this](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentstrings?redirectedfrom=MSDN)

mitza-oci: It's not needed. The standard library has `getenv`.

iguessthislldo: He could use `getenv` if he limited himself to predefined environment variables, but he's using `environ` to scan all the environment variables:

https://github.com/objectcomputing/OpenDDS/blob/5754111cff81e70bf2e48145aae59b051130def2/dds/DCPS/Service_Participant.cpp#L839-L850

mitza-oci: The only variables that can effect the program are those that match a predefined list.  That list would be a in a compile-time constant data structure that could be used to drive command line, ini file, and environment variable handling.

dds/idl/be_global.h
```diff
@@ -260,6 +266,7 @@ class BE_GlobalData {
   bool root_default_nested_;
   bool warn_about_dcps_data_type_;
   ExtensibilityKind default_extensibility_;
+  bool default_enum_extensibility_zero_;
```
sonndinh, line 269: When this is `false`, is there another config option to set the default enum extensibility to final or appendable?

jrw972: No.  Appendable is the default.   I think we are asking the user to use the annotations.

dds/DCPS/RTPS/RtpsDiscovery.h
```diff
@@ -635,6 +635,68 @@ class OpenDDS_Rtps_Export RtpsDiscoveryConfig : public OpenDDS::DCPS::RcObject {
...
+    if (!XTypes::TypeLookupService::parse_extensibility(str, ex)) {
+      return false;
+    }
+
+    switch (ex) {
+    case DDS::APPENDABLE:
+    case DDS::FINAL:
+      type_lookup_service_default_bitmask_extensibility_ = ex;
+      return true;
+      break;
```
iguessthislldo, line 683:
```suggestion
```

dds/DCPS/RTPS/RtpsDiscovery.h
```diff
@@ -635,6 +635,68 @@ class OpenDDS_Rtps_Export RtpsDiscoveryConfig : public OpenDDS::DCPS::RcObject {
...
+    if (!XTypes::TypeLookupService::parse_extensibility(str, ex)) {
+      return false;
+    }
+
+    switch (ex) {
+    case DDS::APPENDABLE:
+    case DDS::FINAL:
+      type_lookup_service_default_enum_extensibility_ = ex;
+      return true;
+      break;
```
iguessthislldo, line 652:
```suggestion
```

iguessthislldo (changes requested):

dds/DCPS/Service_Participant.cpp
iguessthislldo: I didn't see anything for this anymore on my review just now. Are you shelving it for now?

jrw972: Yes.  @mitza-oci  and I agreed that we can pursue that in a different PR.

dds/DCPS/RTPS/RtpsDiscovery.h
```diff
@@ -635,6 +635,66 @@ class OpenDDS_Rtps_Export RtpsDiscoveryConfig : public OpenDDS::DCPS::RcObject {
...
+      }
+      return false;
+    }
+  }
+
+  DDS::ExtensibilityKind type_lookup_service_default_enum_extensibility() const
+  {
+    ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_, DDS::APPENDABLE);
+    return type_lookup_service_default_enum_extensibility_;
+  }
```
sonndinh, line 666: Returning `APPENDABLE` in case of failing to acquire the lock will hide the error to the caller. Will the caller care if the call succeeds or not?

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -73,6 +85,22 @@ class OpenDDS_Dcps_Export TypeLookupService : public virtual DCPS::RcObject {
...
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+    default_enum_extensibility_ = ek;
+  }
+
+  void default_bitmask_extensibility(DDS::ExtensibilityKind ek)
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+    default_bitmask_extensibility_ = ek;
+  }
```
sonndinh, line 102: The default enum and bitmask extensibility in opendds_idl seems to be fixed to `APPENDABLE` but is configurable here. What is an example use case for this?

iguessthislldo (approved):

dds/DCPS/RTPS/RtpsDiscovery.h
jrw972: If the lock cannot be acquired, the system is corrupt and all bets are off.  So, no, the caller will not care.

dds/DCPS/XTypes/TypeLookupService.h
jrw972: As a user, you may interoperate with a system that is not setting the extensibility correctly, e.g., a new and old OpenDDS.  In this case, it allows the newer system to set the flags.

dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -89,6 +89,8 @@ RtpsDiscoveryConfig::RtpsDiscoveryConfig()
   , secure_participant_user_data_(false)
   , max_type_lookup_service_reply_period_(5, 0)
   , use_xtypes_(XTYPES_MINIMAL)
+  , type_lookup_service_default_enum_extensibility_(DDS::APPENDABLE)
+  , type_lookup_service_default_bitmask_extensibility_(DDS::APPENDABLE)
```
mitza-oci, line 93: Is it necessary to have a `bitmask` option for this?  There aren't (yet) any opendds_idl-generated TypeObjects for bitmasks

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -560,6 +563,18 @@ void TypeLookupService::complete_to_dynamic_i(DynamicTypeImpl* dt,
   case TK_ENUM:
     td->kind(TK_ENUM);
     td->name(cto.enumerated_type.header.detail.type_name.c_str());
+    if (cto.enumerated_type.enum_flags & IS_FINAL) {
+      td->extensibility_kind(DDS::FINAL);
+    } else if (cto.enumerated_type.enum_flags & IS_APPENDABLE) {
+      td->extensibility_kind(DDS::APPENDABLE);
+    } else {
+      if (DCPS::log_level >= DCPS::LogLevel::Warning) {
```
mitza-oci, line 571: Should this be a warning?  It's expected behavior since the spec says that's how the enum_flags are used.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -1050,6 +1077,43 @@ void TypeLookupService::remove_guid_from_dynamic_map(const DCPS::GUID_t& guid)
 }
 #endif
 
+bool TypeLookupService::parse_extensibility(const DCPS::String& x, DDS::ExtensibilityKind& d)
+{
+  if (x == "FINAL") {
```
mitza-oci, line 1082: Should this allow case-insensitive matching?  Since the annotation is `@final`, users may be thinking of it in lower case.

dds/idl/be_global.cpp
```diff
@@ -854,11 +859,18 @@ ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
...
+    value = default_extensibility;
+    has_annotation = false;
   }
   return value;
 }
 
+ExtensibilityKind BE_GlobalData::extensibility(AST_Decl* node) const
+{
+  bool has_annotation = false;
+  return extensibility(node, default_extensibility_, has_annotation);
```
mitza-oci, line 871: Since `has_annotation` isn't read after it's written, should there be another overload like `extensibility(node, default_extensibility_)` that ignores has_annotation?

tests/DCPS/ZeroEnum/main.cpp
```diff
@@ -0,0 +1,157 @@
...
+  DDS::DomainParticipantFactory_var domain_participant_factory = TheParticipantFactoryWithArgs(argc, argv);
+  DDS::DomainParticipant_var participant =
+    domain_participant_factory->create_participant(ZeroEnum::HELLO_WORLD_DOMAIN,
+                                                   PARTICIPANT_QOS_DEFAULT,
+                                                   0,
+                                                   0);
+
+  ZeroEnum::MessageTypeSupport_var type_support = new ZeroEnum::MessageTypeSupportImpl();
+  type_support->register_type(participant, "");
+  CORBA::String_var type_name = type_support->get_type_name ();
```
mitza-oci, line 94:
```suggestion
  CORBA::String_var type_name = type_support->get_type_name();
```

tests/DCPS/ZeroEnum/main.cpp
```diff
@@ -0,0 +1,157 @@
...
+  data_reader_qos.representation.value.length(1);
+  data_reader_qos.representation.value[0] = DDS::XCDR2_DATA_REPRESENTATION;
+  data_reader_qos.reliability.kind = DDS::RELIABLE_RELIABILITY_QOS;
+
+  OpenDDS::DCPS::RcHandle<RecorderListenerImpl> listener = OpenDDS::DCPS::make_rch<RecorderListenerImpl>(distributed_condition_set, rtps_discovery->config()->type_lookup_service_default_enum_extensibility());
+
+  // Create Recorder
+  OpenDDS::DCPS::Recorder_var recorder =
+    TheServiceParticipant->create_recorder(participant,
+                                           typeless_topic.in(),
```
mitza-oci, line 125:
```suggestion
                                           typeless_topic,
```

mitza-oci (changes requested):

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -560,6 +563,18 @@ void TypeLookupService::complete_to_dynamic_i(DynamicTypeImpl* dt,
...
+      td->extensibility_kind(DDS::FINAL);
+    } else if (cto.enumerated_type.enum_flags & IS_APPENDABLE) {
+      td->extensibility_kind(DDS::APPENDABLE);
+    } else {
+      if (DCPS::log_level >= DCPS::LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING,
+                   "(%P|%t) WARNING: TypeLookupService::complete_to_dynamic_i: "
+                   "Invalid extensibility kind in TK_ENUM, using default\n"));
+      }
+      td->extensibility_kind(default_enum_extensibility_);
```
sonndinh, line 576: Follow up on my other comment on `default_enum_extensibility` function in `TypeLookupService`. So this happens when we receive a TypeObject from an old system but after the type compatibility check is already done using the received TypeObject?

It looks like we're providing compatibility with old systems only when the enum type isn't annotated and `opendds_idl` in the new system is called with `--default-enum-extensibility-zero`. If the enum type is annotated, I think it won't work.

dds/DCPS/RTPS/RtpsDiscovery.cpp
jrw972: I would say yes and more.  In general, we should be checking received type objects to ensure that _all_ flags and accouterments are set to valid values.

dds/DCPS/XTypes/TypeLookupService.cpp
jrw972: What section of the spec are you referring to?

mitza-oci: The one that caused this problem: `typedef TypeFlag EnumTypeFlag; // Unused. No flags apply`

dds/DCPS/XTypes/TypeLookupService.cpp
jrw972: You are right.  We need to reconsider the design here.

dds/DCPS/XTypes/TypeLookupService.cpp
jrw972: So, the IDL implies that they aren't used but the rest of the document does?

dds/DCPS/XTypes/TypeLookupService.h
```diff
@@ -21,12 +21,24 @@
 #pragma once
 #endif /* ACE_LACKS_PRAGMA_ONCE */
 
+#ifdef OPENDDS_SAFETY_PROFILE
+namespace DDS {
+  enum ExtensibilityKind {
```
mitza-oci, line 26: Should this come from the IDL?  Maybe move it from DdsDynamicData to a common IDL file that's already included?

tests/DCPS/ZeroEnum/README.rst
```diff
@@ -0,0 +1,15 @@
+#############
+ZeroEnum Test
+#############
+
+Previous versions of OpenDDS did not set the type flags for enums (they were zero) in minimal and complete type objects.
+The `-DCPSTypeLookupServiceDefaultEnumExtensibility` option was added to set the type flags for received type objects when they are not set.
```
mitza-oci, line 6: update based on latest changes

dds/DCPS/XTypes/TypeLookupService.h
jrw972: This may not be needed anymore.  I'll see if that is the case.

mitza-oci (approved):

PR #3690, opened 2022-08-22
Parsing partition names can result in buffer overflow

jrw972:
Problem
-------

Parsing comma-separated lists of partitions names assumes each
partition is less than 127 characters.  A partition name longer than
this will cause buffer overflow.

Solution
--------

Check the length of the partition name and log an error if it is too
long.

### DIFF
--- a/tests/unit-tests/dds/FACE/config/QosSettings.cpp
+++ b/tests/unit-tests/dds/FACE/config/QosSettings.cpp
@@ -56,6 +56,16 @@ void test_set_publisher_multiple_partitions() {
   EXPECT_TRUE(!strcmp(qos.partition.name[1], "Bar234"));
 }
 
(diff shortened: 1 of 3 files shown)


tests/unit-tests/dds/FACE/config/QosSettings.cpp
```diff
@@ -56,6 +56,16 @@ void test_set_publisher_multiple_partitions() {
   EXPECT_TRUE(!strcmp(qos.partition.name[1], "Bar234"));
 }
 
+void test_set_publisher_partition_too_long() {
```
mitza-oci, line 59: Where is this called?

jrw972: Right.  No using google test here.

PR #3691, opened 2022-08-22
Clone Google Test and RapidJSON if Not Found

iguessthislldo:


configure
```diff
@@ -1365,7 +1365,25 @@ if (exists $opts{'gtest'} || $tests) {
...
                               $is_windows ? 'googletest-distribution' : ''));
 
+    my $use_sm_dir = 0;
     if (-f File::Spec->catfile($sm_src, $gtest_sanity)) {
+      $use_sm_dir = 1;
+    }
+    elsif (-f File::Spec->catfile($sys_dir, $gtest_sanity)) {
+      $opts{'gtest'} = $sys_dir;
+      $gtest_root = $sys_dir;
+      setEnv('GTEST_ROOT', $gtest_root);
```
simpsont-oci, line 1375: Does this setEnv() need a print statement as well (similar to below?)

simpsont-oci (approved):

configure
iguessthislldo: It looks like it didn't because it's a special case for safety profile.

configure
```diff
@@ -425,6 +425,38 @@ sub run_command {
...
+  my $branch = shift;
+  print "Cloning git repo $url branch $branch\n";
+  my $rv = run_command(['git', 'clone', '-q', '--depth=1', '-b', $branch, $url, $dest]);
+  if (!$rv) {
+    run_command(['git', '--no-pager', 'log', '-1', '--oneline'], chdir => $dest);
+  }
+  return $rv;
+}
+
+sub git_sm_prop {
```
mitza-oci, line 440:
```suggestion
sub git_submodule_prop {
```

Or change the error message on 445, but I think the longer function name is more clear

mitza-oci (changes requested):
This part of the configure script could be tested in GitHub Actions by having at least one job that doesn't use built-in git submodule support (or just deletes the one that are already there).

mitza-oci (approved):

PR #3693, opened 2022-08-22
Windows Tests Warning Cleanup

simpsont-oci:
Problem: Warnings

Solution: Less Warnings


iguessthislldo (approved):
> Problem: Warnings
>
> Solution: Less Warnings

Very succinct

PR #3697, opened 2022-08-24
Updates for building host tools against the latest TAO 2.2a.

ocielliottc:


tools/scripts/create_universal_linux_dds_idl_binaries.sh
```diff
@@ -84,18 +84,21 @@ threads = 0
 debug = 0
 static_stdlibs = 1
 dynamic_loader = 0
+tao_idl_fe_gen = 1
```
mitza-oci, line 87: Could it work without this?  The release archive should already have the generated files.

ocielliottc: It didn't build without it.

jwillemsen: We don't need that option in our host builds using doc_group TAO 3

mitza-oci: @jwillemsen are you using this script?  Whatever the underlying issue is, it seems to be unique to this script or the environment it's run in.

I was able to build successfully without `tao_idl_fe_gen` on an Ubuntu VM.  I only changed buildbits to 64.

jwillemsen: Not with this script, but we do something similar, create a static host tree compilation of just the minimal parts we need on a host system, never ever used `tao_idl_fe_gen` on any build

mitza-oci (approved):

PR #3702, opened 2022-08-26
Fix RTPS SPDP for big endian.

Erich71:
Hi!

I discovered that in the RTPS SPDP transport the E-flag is always set (little endian identifier)
Data is sent in the byte order of the host. This will fail with big endian hosts.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -2360,7 +2360,15 @@ Spdp::SpdpTransport::SpdpTransport(DCPS::RcHandle<Spdp> outer)
   hdr_.vendorId = VENDORID_OPENDDS;
   DCPS::assign(hdr_.guidPrefix, outer->guid_.guidPrefix);
   data_.smHeader.submessageId = DATA;
+
+#if (ACE_BYTE_ORDER == ACE_LITTLE_ENDIAN)
+  // Little Endian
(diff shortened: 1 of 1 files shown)


mitza-oci:
This is by design.  It won't fail because the byte order is part of the protocol, not determined by the endianness of the host.

mitza-oci:
Maybe I misunderstood.  Do you have an example (pcap file) of the submessage header being written in a byte order the contradicts the flag?

Erich71:
> 

That's true for topic data. But in the RTPS SPDP part there is no byte swapping (I investigated that with Wireshark).
On a big endian host SPDP data is sent big endian with E-Flag set. Other DDS participant receive the data and decode the wrong data and subscription fails (received host endianess doesn't matter -> they alway decode what the E-flag says).

I have to dig for a wireshark pcap file which shows the problem in detail.

mitza-oci:
I looked through the code.  It should be taken care of by the generated serialization function:

`bool operator<<(Serializer& strm, const  ::OpenDDS::RTPS::SubmessageHeader& stru)`

which reads the Submessage Header's flags byte and sets the `swap_bytes` accordingly.

mitza-oci:
I ran a test on an arm-archv8a target (using QEMU) and didn't see a problem.  SPDP's Data Submessage is in little endian with the flag and the encoding matching.

PR #3704, opened 2022-08-26
Partition parsing is inconsistent

jrw972:
Problem
-------

Partition parsing uses a fixed size buffer.  All partitions except the
last one are subject to a length check based on the size of the
buffer.

Solution
--------

Dynamically allocate a buffer to not impose a length restriction.

### DIFF
--- a/dds/DCPS/StaticDiscovery.cpp
+++ b/dds/DCPS/StaticDiscovery.cpp
@@ -1689,26 +1689,12 @@ namespace {
   {
     // Value can be a comma-separated list
     const char* start = value.c_str();
-    char buffer[128];
-    std::memset(buffer, 0, sizeof(buffer));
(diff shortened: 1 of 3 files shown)


tottens:
The solution must be compatible with FACE Safety Profile, which does not support dynamic allocation (unless PARTITION QoS is not enabled in Safety Profile).

mitza-oci:
> The solution must be compatible with FACE Safety Profile, which does not support dynamic allocation (unless PARTITION QoS is not enabled in Safety Profile).

It can use dynamic in a few different cases:
* during process startup, dynamic allocation should be allowed (but this is hard to check with automated testing)
* custom allocators can be used to dynamically allocated from a predefined pool

mitza-oci (approved):

PR #3712, opened 2022-08-31
Use ACE::init() for AddressLookup test

simpsont-oci:
Problem: WinSock2 was not being properly initialized for some (all?) windows builds, causing interface lookup errors.

Solution: Use ACE::init() and ACE::fini() (via ACE_TMAIN) for the AddressLookup test.

### DIFF
--- a/tests/DCPS/AddressLookup/main.cpp
+++ b/tests/DCPS/AddressLookup/main.cpp
@@ -145,7 +145,7 @@ void address_info() {
 }
 
 
-int ACE_TMAIN(int argc, char* argv[]) {
+int ACE_TMAIN(int argc, ACE_TCHAR* argv[]) {
   ACE_UNUSED_ARG(argc);
   ACE_UNUSED_ARG(argv);
   const int attempts = 3;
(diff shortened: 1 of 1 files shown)


mitza-oci (commented):
I don't mind changing it but it seems like this shouldn't be necessary.
If you run the test (before these changes) on Windows and put a breakpoint in `ACE::init` it should land there because ACE defines a `main` macro to do just this.

```
>	ACEd.dll!ACE::init() Line 15	C++
 	ACEd.dll!ACE_Main_Base::ACE_Main_Base() Line 75	C++
 	main.exe!ACE_Main::ACE_Main()	C++
 	main.exe!main(int argc, char * * argv) Line 149	C++
```

simpsont-oci:
> I don't mind changing it but it seems like this shouldn't be necessary. If you run the test (before these changes) on Windows and put a breakpoint in `ACE::init` it should land there because ACE defines a `main` macro to do just this.
> 
> ```
> >	ACEd.dll!ACE::init() Line 15	C++
>  	ACEd.dll!ACE_Main_Base::ACE_Main_Base() Line 75	C++
>  	main.exe!ACE_Main::ACE_Main()	C++
>  	main.exe!main(int argc, char * * argv) Line 149	C++
> ```

It must not be getting picked up by VS2015 on hornet then for some reason, since this fixed the issue. Maybe related to headers?

mitza-oci:
> It must not be getting picked up by VS2015 on hornet then for some reason, since this fixed the issue. Maybe related to headers?

Try adding `#include <ace/OS_main.h>` ??

jrw972 (approved):

PR #3713, opened 2022-08-31
Prevent Use Of IPv4-mapped IPv6 'Any' Addresses In RTPS Locators

simpsont-oci:
Problem: Some machines give back IPv4-mapped IPv6 addresses from opened sockets, which weren't being caught by checking for `is_any()` before adding to the list of locators.

Solution:
- Update is_any() to check for IPv4-mapped IPv6 "any" (zeros) address.
- Add Unit Tests For This
- Explicitly check IPV6 addresses returned from IPV6 sockets and avoid subtle conversion to mapped IPV4 addresses.

### DIFF
--- a/dds/DCPS/NetworkAddress.cpp
+++ b/dds/DCPS/NetworkAddress.cpp
@@ -206,12 +206,11 @@ bool NetworkAddress::is_any() const
     return inet_addr_.in4_.sin_addr.s_addr == INADDR_ANY;
 #if defined (ACE_HAS_IPV6)
   } else if (inet_addr_.in6_.sin6_family == AF_INET6) {
-    const bool ipv6_unspec = IN6_IS_ADDR_UNSPECIFIED(&inet_addr_.in6_.sin6_addr);
-    if (!ipv6_unspec) {
-      const unsigned char buff[16] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF, 0, 0, 0, 0 };
-      return memcmp(buff, &inet_addr_.in6_.sin6_addr, sizeof (buff)) == 0;
+    if (IN6_IS_ADDR_UNSPECIFIED(&inet_addr_.in6_.sin6_addr)) {
+      return true;
     }
-    return true;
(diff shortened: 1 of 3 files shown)


dds/DCPS/NetworkAddress.cpp
```diff
@@ -206,7 +206,12 @@ bool NetworkAddress::is_any() const
     return inet_addr_.in4_.sin_addr.s_addr == INADDR_ANY;
 #if defined (ACE_HAS_IPV6)
   } else if (inet_addr_.in6_.sin6_family == AF_INET6) {
-    return IN6_IS_ADDR_UNSPECIFIED(&inet_addr_.in6_.sin6_addr);
+    const bool ipv6_unspec = IN6_IS_ADDR_UNSPECIFIED(&inet_addr_.in6_.sin6_addr);
+    if (!ipv6_unspec) {
+      const unsigned char buff[16] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF, 0, 0, 0, 0 };
+      return memcmp(buff, &inet_addr_.in6_.sin6_addr, sizeof (buff)) == 0;
+    }
+    return true;
```
mitza-oci, line 214:
```suggestion
    if (IN6_IS_ADDR_UNSPECIFIED(&inet_addr_.in6_.sin6_addr)) {
      return true;
    }
    static const unsigned char buff[16] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF, 0, 0, 0, 0 };
    return memcmp(buff, &inet_addr_.in6_.sin6_addr, sizeof (buff)) == 0;
```

mitza-oci (changes requested):

mitza-oci (approved):

sonndinh (approved):
I think the address this is working on is IPv4-mapped IPv6, and not IPv6-mapped IPv4 addresses as mentioned in the title and description, right?

simpsont-oci:
Ah, yeah, the mapping goes both ways. I got the terminology wrong. I can at least change the PR title.

PR #3718, opened 2022-09-03
Rewrite `DataWriterImpl[_T]` to use an Abstract `Sample` Class

iguessthislldo:
This moves a lot of the functionality of DataWriterImpl_T into
DataWriterImpl by using a new class called AbstractSample to avoid
working with the native C++ IDL-generated types directly as much as
possible. Some more work is needed on this front though. The ultimate
goal of this class is to be able to use it with DynamicData as well as
with the native types to work towards implementing a DynamicDataWriter
and DyanmicDataReader.


mitza-oci:
Merge conflict in `dds/DCPS/DataWriterImpl.cpp`

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+  virtual bool compare(const AbstractSample& other) const = 0;
+  virtual bool to_message_block(ACE_Message_Block& mb) = 0;
+  virtual bool from_message_block(const ACE_Message_Block& mb) = 0;
+  virtual AbstractSample_rch copy(bool read_only, bool key_only) const = 0;
+  AbstractSample_rch copy(bool read_only) const
+  {
+    return copy(read_only, key_only_);
+  }
+#ifndef OPENDDS_SAFETY_PROFILE
+  virtual DDS::DynamicData* get_dynamic_data(DDS::DynamicType_ptr type) = 0;
```
mitza-oci, line 68: For IDL interface types, let's consistently use either `*` or `_ptr`.... or use `_var`

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+  virtual ~AbstractSample() {}
+
+  bool owns_data() const
+  {
+    return owns_data_;
+  }
+
+  bool read_only() const
+  {
+    return owns_data_;
```
mitza-oci, line 48:
```suggestion
    return read_only_;
```

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+  typedef DDSTraits<NativeType> TraitsType;
+  typedef MarshalTraits<NativeType> MarshalTraitsType;
+  typedef RcHandle<NativeSample<NativeType> > Rch;
+  typedef KeyOnly<const NativeType> KeyOnlyType;
+  typedef KeyOnly<NativeType> MutableKeyOnlyType;
+#if OPENDDS_HAS_DYNAMIC_DATA_ADAPTER
+  typedef XTypes::DynamicDataAdapter<NativeType> DynamicDataImpl;
+#endif
+
+  NativeSample(const NativeType& data, bool key_only = false)
```
mitza-oci, line 100: Does this need to support implicit conversion?

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+
+struct OpenDDS_Dcps_Export AbstractSampleRchCmp {
+  bool operator()(const AbstractSample_rch& lhs, const AbstractSample_rch& rhs) const
+  {
+    return lhs->compare(*rhs);
+  }
+};
+
+template <typename NativeType>
+class NativeSample : public AbstractSample {
```
mitza-oci, line 89:
```suggestion
class LanguageBindingSample : public AbstractSample {
```

Alternatives?
XTypes calls this "Plain Language Binding" (7.5.1) but that seemed too long

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+
+#ifndef OPENDDS_SAFETY_PROFILE
+  DDS::DynamicData* get_dynamic_data(DDS::DynamicType_ptr type)
+  {
+#  if OPENDDS_HAS_DYNAMIC_DATA_ADAPTER
+    if (type && !dynamic_data_initialized_) {
+      dynamic_data_ = DynamicDataImpl(type, getMetaStruct<NativeType>(), *data_);
+      dynamic_data_initialized_ = true;
+    }
+    return &dynamic_data_;
```
mitza-oci, line 229: Returning an IDL interface type should increment the reference count

dds/DCPS/DataWriterImpl.h
```diff
@@ -119,6 +120,22 @@ class OpenDDS_Dcps_Export DataWriterImpl
...
+  {
+    skip_serialize_ = value;
+  }
+
+  bool get_marshal_skip_serialize() const
+  {
+    return skip_serialize_;
+  }
+
+  ACE_INLINE
```
mitza-oci, line 133:
```suggestion
```

dds/DCPS/DataWriterImpl.h
```diff
@@ -572,6 +607,84 @@ class OpenDDS_Dcps_Export DataWriterImpl
...
+    SerializedSizeBound key_only_bound_;
+  } encoding_mode_;
+
+  TypeSupportImpl* get_type_support() const
+  {
+    return type_support_;
+  }
+
+  DDS::ReturnCode_t instance_must_exist(
+    const char* const method_name,
```
mitza-oci, line 673:
```suggestion
    const char* method_name,
```

dds/DCPS/DataWriterImpl.h
```diff
@@ -572,6 +607,84 @@ class OpenDDS_Dcps_Export DataWriterImpl
...
+    , header_size_(encoding_.is_encapsulated() ? EncapsulationHeader::serialized_size : 0)
+    , bound_(ts->serialized_size_bound(encoding_))
+    , key_only_bound_(ts->key_only_serialized_size_bound(encoding_))
+    {
+    }
+
+    bool valid() const
+    {
+      return valid_;
+    }
```
mitza-oci, line 636:
```suggestion
    bool valid() const { return valid_; }
```
These are so short they can be 1-liners... but it's down to personal preference so either way is fine

tests/DCPS/FooTest4_0/Reader.cpp
```diff
@@ -140,9 +140,8 @@ Reader::read (const SampleInfoMap& si_map,
         for (CORBA::ULong i = 0 ; i < si.length() ; i++)
         {
           // skip the dispose notification sample.
-          if (si[i].valid_data == 0)
-          {
-            ACE_OS::printf ("got SampleInfo[%d] dispose sample\n", i);
+          if (si[i].valid_data == 0) {
```
mitza-oci, line 143:
```suggestion
          if (!si[i].valid_data) {
```

mitza-oci (changes requested):

dds/DCPS/AbstractSample.h
mitza-oci: It's not a great name so I asked for "Alternatives?"  I suppose leaving out "Plain" may be more confusing.  Could be "TypedSample" since it has a specific type attached.  Or "Sample_T"?

Or how about "static" since it's based on compile time code generation and static/dynamic is often used for that distinction.

iguessthislldo: I skipped over the "Alternatives?" apparently. I'm not sure about TypedSample because the dynamic ones are technically going to be typed, but in a different way. Sample_T doesn't sound bad to me. Could also be `GeneratedSample[_T]` or `CppSample[_T]`. If either of those don't sound good, Sample_T is alright with me.

Also if I'm changing`NativeSample`, I should probably also change `NativeTypeSupportImpl` to match?

mitza-oci: Yeah, when I thought about this some more I ended up at just `Sample<T>` (as in, no need to have the actual `_T` in its name).  `TypeSupportImpl` could follow a similar pattern.

dds/DCPS/AbstractSample.h
iguessthislldo: Probably not

dds/DCPS/AbstractSample.h
```diff
@@ -0,0 +1,262 @@
...
+typedef RcHandle<AbstractSample> AbstractSample_rch;
+
+/**
+ * Represents a sample that can either be an instance of a C++ generated type
+ * from opendds_idl or a DynamicData. This is meant to be used by
+ * DataReaderImpl and DataWriterImpl.
+ */
+class OpenDDS_Dcps_Export AbstractSample : public virtual RcObject {
+public:
+  AbstractSample(bool owns_data, bool read_only, bool key_only)
```
mitza-oci, line 32: Instead of bools, should it use a flags enum/integer?  It would be more self-documenting at the calling sites.

dds/DCPS/AbstractSample.h
iguessthislldo: Making it a var would solve your comment on the reference counting and let me undo the changes to make it possible to have a "null" `DyanmicDataAdaptor`.

dds/DCPS/AbstractSample.h
iguessthislldo: > `TypeSupportImpl` could follow a similar pattern.

An issue with this is the renamed `NativeTypeSupportImpl` would conflict with the existing [`TypeSupportImpl`](https://github.com/objectcomputing/OpenDDS/blob/1c8d3758d83bf2dfbe49e35cfcef1be4ef9dc76d/dds/DCPS/TypeSupportImpl.h#L74).

mitza-oci: In this case we could use the `_T` to match `DataReaderImpl_T` / `DataWriterImpl_T`

dds/DCPS/DataWriterImpl.h
iguessthislldo: I mostly do multiple lines to allow for expansion latter, but not always.

mitza-oci: > I mostly do multiple lines to allow for expansion later, but not always.

Changing it later is fine too, not much added complexity in reviewing that (future) diff.

dds/DCPS/AbstractSample.h
iguessthislldo: It would, but the values aren't together in the templated Sample constructor. I can make separate enums. Also I realized Dynamic Sample doesn't need `owns_data`, so I'm moving that to the templated Sample.

mitza-oci (approved):

dds/DCPS/DataWriterImpl.cpp
```diff
@@ -2830,6 +2822,419 @@ void DataWriterImpl::transport_discovery_change()
...
+  const char* const method_name,
+  const Sample& sample,
+  DDS::InstanceHandle_t& instance_handle,
+  bool remove)
+{
+  OPENDDS_ASSERT(sample.key_only());
+
+  ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, get_lock(), DDS::RETCODE_ERROR);
+
+  const InstanceValuesToHandles::const_iterator pos = find_instance(sample);
```
mitza-oci, line 3148:
```suggestion
  const InstanceValuesToHandles::iterator pos = find_instance(sample);
```

This should fix the CI errors.  It's a quirk of the pre-C++11 standard library that `map::erase` doesn't take a `const_iterator`.

iguessthislldo: Yea, I forgot about that. `erase` is a lot better in `C++11` between that and returning the next iterator.

PR #3719, opened 2022-09-03
Value Dispatcher Updates

simpsont-oci:
Problem: ValueWriterDispatcher is defined and made available as part of DataReader and DataWriter entity APIs, but no corresponding ValueReaderDispatcher is defined and there is no way to get access to any Value*Dispatcher short of having access to a reader or writer entity. This is problematic for generically using ValueReader and ValueWriter for conversion back and forth between C++ IDL types and other representations of the same values, as it requires knowing the type at compile time.

Solution: Combine the methods from ValueWriterDispatcher (and what would have been ValueReaderDispatcher) into a single templated object, ValueDispatcher, and move it to be part of the TypeSupportImpl object. This way, both directions of conversion are available as long as there's a handle to the TypeSupport object, and it allows conversion in both directions without knowing the type at compile time.


jrw972 (approved):
My only concern with this PR is the removal of the `vread` and `vwrite` declarations.  I struggled getting things to build and link without them.  However, if it works without them, then that is okay too.

PR #3723, opened 2022-09-07
Modifications to build the bulk of OpenDDS for VxWorks 21.03

ocielliottc:


dds/DCPS/InstanceHandle.h
```diff
@@ -36,7 +36,7 @@ class OpenDDS_Dcps_Export InstanceHandleGenerator {
 
 private:
 #ifdef ACE_HAS_CPP11
-  std::atomic<int32_t> sequence_;
+  std::atomic<long> sequence_;
```
simpsont-oci, line 39: For the sake of consistency, I think all these 'long' values should be InstanceHandle_t (which will be typedef'd as HANDLE_TYPE_NATIVE / long anyway, but because it shows up in the API for next, they should probably match.

mitza-oci: The IDL type `long` is different from C++'s `long`

simpsont-oci: Ah, sorry, I forgot the typedef was in the IDL file. Yes, it would all get mapped to CORBA::Long anyway... but that seems like further justification for having everything be specified in terms of InstanceHandle_t (or CORBA::Long).

iguessthislldo (approved):

mitza-oci:
The change in `dds/DCPS/RTPS/Sedp.cpp` and the ones in `dds/DCPS/Time_Helper.inl` look very similar but use different typedefs and cast operator/function.  Should they use the same approach, or are there differences in context?

ocielliottc:
> The change in `dds/DCPS/RTPS/Sedp.cpp` and the ones in `dds/DCPS/Time_Helper.inl` look very similar but use different typedefs and cast operator/function. Should they use the same approach, or are there differences in context?

I think those are different contexts.  The underlying types are different, as well.

PR #3724, opened 2022-09-08
GitHub Actions: Migrate Ubuntu 18 to 20

mitza-oci:
- All u18 jobs moved to u20

- JDK version upgraded: build_u20_i0_xer0_js0_j17 uses Java 17 (hence j17), upgraded from 12

- Compiler versions upgraded: build_u18_clang5_i0w1_sec became build_u20_clang12_i0w1_sec and build_u18_gcc6_d0w1_cpp03 became build_u20_gcc9_d0w1_cpp03

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1494,7 +1494,7 @@ jobs:
         ./configure \
           --ace=$GITHUB_WORKSPACE/ACE_TAO/ACE \
           --mpc=$GITHUB_WORKSPACE/MPC \
-          --compiler=clang++-5.0 \
+          --compiler=clang++-12 \
           --std=c++11 \
           --no-inline \
           --features=uses_wchar=1 \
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -213,11 +213,11 @@ jobs:
         name: ${{ github.job }}_artifact
         path: ${{ github.job }}.tar.xz
 
-  cmake_u18_i0_xer0_js0_j12:
+  cmake_u20_i0_xer0_js0_j17:
```
sonndinh, line 216: Looks like these u18 test jobs only run CMake tests. Do we want to run more tests on u20?

jrw972 (approved):

.github/workflows/build_and_test.yml
mitza-oci: There are some other jobs like https://github.com/objectcomputing/OpenDDS/runs/8242918577?check_suite_focus=true that run tests on u20.  We can extend test coverage in a future PR.

sonndinh (approved):

PR #3727, opened 2022-09-08
Implement DynamicData writing operations

sonndinh:


dds/DdsDynamicDataSeq.idl
```diff
@@ -42,12 +42,12 @@ module DDS {
...
   typedef sequence<long double> Float128Seq;
   typedef sequence<char> CharSeq;
-  typedef sequence<wchar> WcharSeq;
+  typedef sequence<wchar> WCharSeq;
   typedef sequence<boolean> BooleanSeq;
   typedef sequence<octet> ByteSeq;
   // Already defined in DdsDcpsCore.idl
   // typedef sequence<string> StringSeq;
-  typedef sequence<wstring> WstringSeq;
+  typedef sequence<wstring> WStringSeq;
```
mitza-oci, line 50: Is there a spec issue for this?

sonndinh: I don't know. I thought it was a typo when it was copied from the spec because the naming convention seems to be `WStringSeq` and `WCharSeq` from other places like CORBA. If it's not a typo, I'll revert these changes.

mitza-oci: I think we should have this match XTypes Annex C. It's unfortunate that the IDL in that section is inconsistent with other similar IDL in related specs.

sonndinh: I've reverted these changes.

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicDataBase.h
```diff
@@ -0,0 +1,164 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_XTYPES_DYNAMIC_DATA_BASE_H
```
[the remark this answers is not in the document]
sonndinh: I've added a cpp file for this.

dds/DCPS/Recorder.h
```diff
@@ -16,7 +16,7 @@
 
 #include <dds/DdsDcpsInfrastructureC.h>
 
-#include "XTypes/DynamicDataImpl.h"
+#include "XTypes/DynamicDataXcdrReadImpl.h"
```
mitza-oci, line 19:
```suggestion
#include "<dds/DdsDynamicDataC.h>
```
Only the interface types are needed in the header

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicDataBase.cpp
```diff
@@ -0,0 +1,157 @@
...
+               " Calling on an unexpected type %C\n", typekind_to_string(tk)));
+  }
+  return MEMBER_ID_INVALID;
+}
+
+bool DynamicDataBase::is_type_supported(TypeKind tk, const char* func_name)
+{
+  if (!is_primitive(tk) && tk != TK_STRING8 && tk != TK_STRING16) {
+    if (DCPS::log_level >= DCPS::LogLevel::Notice) {
+      ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: DynamicDataXcdrReadImpl::is_type_supported:"
```
mitza-oci, line 93:
```suggestion
      ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: DynamicDataBase::is_type_supported:"
```

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicDataImpl.h
```diff
@@ -61,524 +45,898 @@ class OpenDDS_Dcps_Export DynamicDataImpl : public virtual DCPS::LocalObject<DDS
...
 
-  static const ACE_CDR::ULong ITEM_COUNT_INVALID = ACE_UINT32_MAX;
+OpenDDS_Dcps_Export
+void serialized_size(const Encoding& encoding, size_t& size, const XTypes::DynamicDataImpl& data);
 
-  /// Cache the number of items (i.e., members or elements) in the data it holds.
-  ACE_CDR::ULong item_count_;
-};
+OpenDDS_Dcps_Export
+bool operator<<(Serializer& ser, const XTypes::DynamicDataImpl& data);
```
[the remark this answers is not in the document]
sonndinh: I still need these to take `DynamicDataImpl` because they are called internally in `DynamicDataImpl` class. So maybe we add another pair of `operator<<` and `serialized_size` which take `DDS::DynamicData` and delegate the serialization operations to the ones that take `DynamicDataImpl` when it's the specific DynamicData implementation being used?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
-        return descriptor->id();
-      }
-    }
+bool DynamicDataImpl::insert_single(DDS::MemberId id, const ACE_OutputCDR::from_int8& value)
+{
+  // The same member might be already written to complex_map_.
+  // Make sure there is only one entry for each member.
+  if (container_.complex_map_.count(id) > 0) {
+    container_.complex_map_.erase(id);
   }
```
mitza-oci, line 95: It's not necessary to check the `count` before `erase`.  Just calling `erase` is equivalent -- if nothing is there for that key then `erase` returns 0 (which is helpful for the `else if` on line 154 below).

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
-  if (DCPS::DCPS_debug_level >= 1) {
-    ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) DynamicDataImpl::get_member_id_at_index -")
-               ACE_TEXT(" Calling on an unexpected type %C\n"), typekind_to_string(tk)));
-  }
-  return MEMBER_ID_INVALID;
+  return insert_single(id, value);
 }
 
-ACE_CDR::ULong DynamicDataImpl::get_item_count()
+bool DynamicDataImpl::is_discriminator_type(TypeKind tk) const
```
mitza-oci, line 197: Should this be named something like is_valid_discriminator_type or is_possible_discriminiator_type?  It can be a `static` member since it doesn't use `this`.  That would make it clear that it's not checking if `tk` is the type of this object's discriminator.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
+DynamicDataImpl::SingleValue::SingleValue(CORBA::ULong uint32)
+  : kind_(TK_UINT32), uint32_(uint32)
+{}
 
-DDS::ReturnCode_t DynamicDataImpl::clear_value(DDS::MemberId /*id*/)
+DynamicDataImpl::SingleValue::SingleValue(ACE_OutputCDR::from_int8 value)
+  : kind_(TK_INT8)
 {
-  return DDS::RETCODE_UNSUPPORTED;
+  active_ = new(int8_) ACE_OutputCDR::from_int8(value.val_);
```
mitza-oci, line 262: use the member initializer list

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
-          value = new DynamicDataImpl(strm_, member_type);
-        }
-      }
-      break;
+bool DynamicDataImpl::set_complex_to_union(DDS::MemberId id, DDS::DynamicData_ptr value,
+                                           const DDS::TypeDescriptor_var& descriptor)
+{
+  if (id == DISCRIMINATOR_ID) {
+    DDS::DynamicType_var disc_type = get_base_type(descriptor->discriminator_type());
+    if (!disc_type->equals(value->type())) {
```
mitza-oci, line 1121: Use a _var to manage the reference returned by `value->type()` -- are there other places the same change is needed?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
+  : elem_kind_(TK_BOOLEAN)
+{
+  active_ = new(boolean_seq_) DDS::BooleanSeq(boolean_seq);
+}
 
-  num_skipped = 1;
-  const DDS::DynamicType_ptr member_type = md->type();
-  return member_type && skip_member(member_type);
+// TODO: Is this sufficient for string and wstring? or does it need to duplicate each
+// string/wstring in the sequence?
```
mitza-oci, line 1673: Sequence objects own their contents, so StringSeq acts just like `vector<string>`.  The TODO can probably be removed.

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
+  return true;
+}
+
+template<typename StringType>
+bool DynamicDataImpl::DataContainer::serialized_size_generic_string_sequence(
+  const DCPS::Encoding& encoding, size_t& size, const IndexToIdMap& index_to_id) const
+{
+  if (encoding.xcdr_version() == DCPS::Encoding::XCDR_VERSION_2) {
+    serialized_size_delimiter(encoding, size);
+  }
```
iguessthislldo, line 2787:
```suggestion
  serialized_size_delimiter(encoding, size);
```

These are redundant because the function already checks that.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -7,3263 +7,6479 @@
...
+    return false;
+  }
+  CORBA::ULong length = 1;
+  const DDS::BoundSeq& bound = descriptor->bound();
+  for (CORBA::ULong i = 0; i < bound.length(); ++i) {
+    length *= bound[i];
+  }
+
+  const DDS::DynamicType_var elem_type = get_base_type(descriptor->element_type());
+
```
iguessthislldo, line 5785: This should be able to use the same functions as XCDR2 like we do in normal serialization. Functions like `serialized_size_parameter_id` were designed to be able to handle either encoding, but I understand not covering it.

iguessthislldo, line 5958: I'm not 100 percent sure about this, but I don't think it matters what the ID is because we know the discriminator will always be first.

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: Yes, I think most of these functions can be reused for XCDR1. The main difference is optional members for which I think we don't have an existing function that handles both encodings.

sonndinh: I don't know why I couldn't find this comment on the file changes view. Could you let me know which lines is this comment for?

mitza-oci: I can't track it down, line 4809 in the version that GitHub is showing as relevant for this review (a8afeb7cdf4f9d8d896959c5b179eca7d323cf4b) doesn't look like it deals with the member ID of a union's discriminator.

iguessthislldo (approved):

dds/DCPS/XTypes/DynamicDataImpl.cpp
iguessthislldo: I didn't notice this was wrong. I can't imagine I picked the wrong lines, so I guess this is another instance of GitHub (or maybe a combination of this with Firefox?) bugging out on the diff for this file.

There were a few instances of this so this probably isn't the same as what I originally commented on, but they're all the same idea:

https://github.com/objectcomputing/OpenDDS/blob/7ad6752904118825507a4c0fcfe365235a07643f/dds/DCPS/XTypes/DynamicDataImpl.cpp#L6290-L6291

mitza-oci (approved):

PR #3728, opened 2022-09-08
Re-enable MacOS 12 builds

jrw972:


mitza-oci (approved):

jrw972:
MacOS 12 builds still fail due to corruption in large binary files.

PR #3746, opened 2022-09-19
Java ParticpantLocationTest sporadically fails

jrw972:
Problem
-------

The Java ParticipantLocationTest sporadically fails in CI; both in GitHub Actions and the scoreboard.  The root causes is not obvious from the test logging.

Solution
--------

* Update the test framework for Java processes.
* Add logging for BIT activities.

### DIFF
--- a/bin/PerlDDS/Run_Test.pm
+++ b/bin/PerlDDS/Run_Test.pm
@@ -582,6 +582,73 @@ sub unused_flags {
   return keys(%{$self->{_flags}->{unused}});
 }
 
+sub _process_common {
+  my $self = shift;
+  my $name = shift;
+  my $params = shift;
+  my $debug_logging = 1;
+
+  if ($$params !~ /-DCPSLogLevel / && $self->{dcps_log_level}) {
+    $$params .= " -DCPSLogLevel $self->{dcps_log_level}";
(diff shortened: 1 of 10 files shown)


bin/PerlDDS/Run_Test.pm
```diff
@@ -1025,6 +1052,30 @@ sub _create_process {
...
 
+sub _create_java_process {
+  my $self = shift;
+  my $main_class = shift;
+  my $params = shift;
+  my @jars = shift;
+  my $vmargs = shift;
+
+  $self->_info("TestFramework::_create_java_process creating executable="
+    . "w/ params=$params jars=@jars vmargs=$vmargs\n");
```
iguessthislldo, line 1063: Since this is actually an array reference it won't turn into anything useful (something like `ARRAY(0x564ba59fd4b8)`). Arrays of strings will become strings, but they don't have separators between the elements.

```suggestion
    . "w/ params=$params jars=" . join(':', @{$jars}) . " vmargs=$vmargs\n");
```

I put in `:` since that's how the Java classpaths are separated, but it could be comma, space, etc.

iguessthislldo (changes requested):

PR #3748, opened 2022-09-21
Unauthenticated participant leads to invalid iterator

jrw972:
Problem
-------

A participant with incompatible security attributes is discovered and then removed without resetting the iterator.  The invalid iterator then causes a run-time error.

Solution
--------

* Refactor the code in SPDP to make participant removal more obvious.
* Add/enable security tests that test this scenario.
* Fix the `SC1_sec_on_bad_perm_2_failure` scenario to use an access-controlled topic.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -380,10 +380,7 @@ Spdp::shutdown()
     }
 #endif
 
-    for (DiscoveredParticipantIter part = participants_.begin();
-         part != participants_.end();
-         /* Update in loop*/)
-    {
+    for (DiscoveredParticipantIter part = participants_.begin(); part != participants_.end(); ++part) {
 #ifdef OPENDDS_SECURITY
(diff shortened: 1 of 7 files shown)


mitza-oci (approved):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -396,7 +393,7 @@ Spdp::shutdown()
       }
       purge_handshake_deadlines(part);
 #endif
-      remove_discovered_participant(part++);
+      purge_discovered_participant(part);
     }
   }
```
simpsont-oci, line 399: Does it make sense to clear `participants_` here since purge doesn't do an erase like remove did?

jrw972: Since we are in `shutdown`, I don't think it is necessary.  However, if it makes the code more boring than I'm for it.

simpsont-oci (approved):

PR #3750, opened 2022-09-23
Attempt to fix MultiTopic test

sonndinh:


dds/DCPS/MultiTopicDataReader_T.cpp
```diff
@@ -104,33 +102,33 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::join(
       SampleInfo info;
       const ReturnCode_t ret = other_dri->read_instance_generic(other_data.ptr_,
         info, ih, READ_SAMPLE_STATE, ANY_VIEW_STATE, ALIVE_INSTANCE_STATE);
-      if (ret != RETCODE_OK && ret != RETCODE_NO_DATA) {
+      if (ret == DDS::RETCODE_NO_DATA || !info.valid_data) {
```
mitza-oci, line 105: This looks at the state of `info` even in the case of other (not "OK") return values.

mitza-oci (changes requested):

dds/DCPS/MultiTopicDataReader_T.cpp
jrw972: The `ret != RETCODE_OK` should be the first check.

dds/DCPS/MultiTopicDataReader_T.cpp
```diff
@@ -104,33 +102,33 @@ MultiTopicDataReader_T<Sample, TypedDataReader>::join(
...
+        return DDS::RETCODE_NO_DATA;
+      }
+
+      if (ret != RETCODE_OK) {
         throw std::runtime_error(
-         OPENDDS_STRING("In join(), incoming DataReader for ") + OPENDDS_STRING(other_topic) +
-         " read_instance_generic: " + retcode_to_string(ret));
-      } else if (ret == DDS::RETCODE_OK) {
+          OPENDDS_STRING("In join(), incoming DataReader for ") + OPENDDS_STRING(other_topic) +
```
jrw972, line 111: Do we really want to `throw` here?  What about returning the error?

sonndinh: I'm not sure why exception was used in the first place, but probably it could just use the return code.

mitza-oci (changes requested):
There is a warning on line 208 of MultiTopicDataReader_T.cpp because lines that end with `\` are joined with the next line by the preprocessor.  I suggest using a `/* */` comment for the whole block.

mitza-oci (approved):

PR #3754, opened 2022-09-27
Add `AtomicBool` Class

iguessthislldo:


dds/DCPS/ReplayerImpl.cpp
```diff
@@ -307,7 +307,7 @@ ReplayerImpl::enable()
     return DDS::RETCODE_OK;
   }
 
-  // if (this->publisher_servant_->is_enabled() == false) {
+  // if (this->publisher_servant_->is_enabled()) {
```
jrw972, line 310:
```suggestion
  // if (!this->publisher_servant_->is_enabled()) {
```

jrw972 (changes requested):

simpsont-oci (changes requested):
Please add a unit test for AtomicBool (multithreaded would be ideal (see ServiceEventDispatcher unit test for example of relatively easy multithreading)). A simple unit test for SafeBool_T would be nice as well, that way they'll both be sure to get tsan & ubsan coverage.

iguessthislldo:
> Please add a unit test for AtomicBool (multithreaded would be ideal (see ServiceEventDispatcher unit test for example of relatively easy multithreading)).

I think what I added for AtomicBool should be sufficient for tsan since I think it just has to be accessing the memory for it to check it and I don't actually have to assert the value.

simpsont-oci (approved):

jrw972 (approved):

PR #3755, opened 2022-09-28
Building for VxWorks on Windows

ocielliottc:
- Modified opendds_idl and tao_idl cmake files to build for VxWorks on Windows.
- Split library dependencies based on OPENDDS_XERCES3 and OPENDDS_SECURITY.

### DIFF
--- a/cmake/OpenDDSConfig.cmake
+++ b/cmake/OpenDDSConfig.cmake
@@ -76,8 +76,13 @@ endif()
 
 include(${CMAKE_CURRENT_LIST_DIR}/init.cmake)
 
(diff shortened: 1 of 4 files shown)


iguessthislldo (approved):

iguessthislldo (changes requested):
CMake tests appear to be broken

ocielliottc:
@iguessthislldo @mitza-oci It appears that I have moved past the cmake errors.  I could be wrong, but the run-time errors don't look like a result of my changes.  Could you please verify?  Thanks.

iguessthislldo (approved):
Yes. I think it looks alright now.

PR #3758, opened 2022-09-30
Fix macOS IPv6 Build Issue

simpsont-oci:
Problem: IPv6 compile error in `NetworkConfigModifier.cpp` (macOS).

Solution: Modify IPv6 path similar to previous changes to IPv4 path.


mitza-oci (changes requested):
Please add CI coverage

mitza-oci (approved):

jrw972 (approved):

PR #3759, opened 2022-09-30
`set_listener` creates logical data race for already received samples

jrw972:
Problem
-------

Samples can be received between the time a data reader is created and a listener is attached to it with `set_listener`.

Solution
--------

Invoke `on_data_available` after calling `set_listener`.  The listener needs to be thread safe since the middleware can invoke the listener any time after `set_listener`.


tests/DCPS/QueryCondition/QueryConditionTest.cpp
```diff
@@ -378,6 +378,9 @@ bool run_complex_filtering_test(const DomainParticipant_var& dp,
   MessageDataReader_var mdr2 = MessageDataReader::_narrow(dr2);
   mdr1->set_listener(ml1, DDS::DATA_AVAILABLE_STATUS);
   mdr2->set_listener(ml2, DDS::DATA_AVAILABLE_STATUS);
+  // Call on_data_available in case there are samples which are waiting
+  ml1->on_data_available(mdr1);
+  ml2->on_data_available(mdr2);
```
mitza-oci, line 383: The samples can't be waiting, they'll be written later in this same function

dds/FACE/FaceTSS.h
```diff
@@ -398,6 +398,8 @@ void register_callback(FACE::CONNECTION_ID_TYPE connection_id,
   } else {
     DDS::DataReaderListener_var listener = new Listener<Msg>(callback, connection_id);
     readers[connection_id]->dr->set_listener(listener, DDS::DATA_AVAILABLE_STATUS);
+    // TODO: Do we need to invoke the listener?
+    // TODO: Can calbacks be invoked concurrently?
```
simpsont-oci, line 402: Do we have answers for these questions?

jrw972: @mitza-oci Does FACE say we need to invoke the listener here?  If so, do we need to prevent a callback from being invoked concurrently?

jrw972:
The implication is that a user calling `set_listener` needs to invoke the listener if samples may already be available.

sonndinh:
> The implication is that a user calling `set_listener` needs to invoke the listener if samples may already be available.

Sounds like we need to update the devguide to add this. I think there'll be applications that don't know whether there is an arrived sample or not after it calls `set_listener`. If it calls `on_data_available`, there might be no data available. I think user should be aware of that when writing their `on_data_available`.

jrw972:
Agreed.

dds/FACE/FaceTSS.h
mitza-oci: I couldn't find any specific requirement for this.  My suggestion would be to leave the code in this function as it is.

mitza-oci (approved):

PR #3764, opened 2022-10-04
Clean Up Recorders / Replayers Before Removing Topics

simpsont-oci:
Problem: When recorders and replayers aren't cleaned up before topics, there's a chance to hit assertions in TopicDetails regarding the number of remaining local publications / subscriptions.

Solution: Clean up recorders & replayers after normal publications / subscriptions.


simpsont-oci:
For the failing check, see #3765 .

jrw972 (approved):

PR #3765, opened 2022-10-04
Fix Call To pending_messages in MessageTracker

simpsont-oci:
Problem: MessageTracker `wait_messages_pending` sometimes uses the wrong flavor of `pending_messages` and causes a double-lock. See https://github.com/objectcomputing/OpenDDS/actions/runs/3184202831/jobs/5193569354

Solution: Use `pending_messages_i` and be sure to set the DCPSPendingTimeout value for the DevGuideExamples tests.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -11989,6 +11989,13 @@ jobs:
         <command name="process_logs" options="prettify"/>
         </autobuild>
         EOF
+        echo "--- ip addr ---"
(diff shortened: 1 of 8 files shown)


simpsont-oci:
Well, adding DCPSPendingTimeout allowed it to time out but then it hit a deadlock. Fixing the deadlock then leads to the test simply not passing. So now I guess we need to fix that.

simpsont-oci:
Replacing with #3785.

PR #3766, opened 2022-10-05
Test and fix for problem with appendable types and additional written fields

jwillemsen:
…endable struct where the subscriber has added a field and check if the additional field has a correct zero value

    * tests/DCPS/XTypes/Publisher.cpp:
    * tests/DCPS/XTypes/Publisher.idl:
    * tests/DCPS/XTypes/Subscriber.cpp:
    * tests/DCPS/XTypes/Subscriber.idl:
    * tests/DCPS/XTypes/run_test.pl:

Added to try to reproduce an issue we see with AXCIOMA and OpenDDS, the additional fields are not zero in the subscriber when not send by the publisher


tests/DCPS/XTypes/Publisher.cpp
```diff
@@ -92,6 +92,23 @@ void write_additional_prefix_field_struct(const DataWriter_var& dw)
   }
 }
 
+void write_base_appendable_struct(const DataWriter_var& dw)
+{
+  BaseAppendableStructDataWriter_var typed_dw = BaseAppendableStructDataWriter::_narrow(dw);
+
+  BaseAppendableStruct apfs;
```
sonndinh, line 99: Looks like the variable name `apfs` is copied over from `write_additional_prefix_field_struct` function. Can you change to use a better name, something like `bas` for example?

tests/DCPS/XTypes/Subscriber.cpp
```diff
@@ -64,6 +64,31 @@ ReturnCode_t read_struct(const DataReader_var& dr, const T1& pdr, T2& data)
   return ret;
 }
 
+template<typename T1, typename T2>
+ReturnCode_t read_extended_struct(const DataReader_var& dr, const T1& pdr, T2& data)
+{
+  ReturnCode_t ret = RETCODE_OK;
+  if ((ret = read_i(dr, pdr, data)) == RETCODE_OK) {
```
mitza-oci, line 71:
```suggestion
  ReturnCode_t ret = read_i(dr, pdr, data);
  if (ret =  == RETCODE_OK) {
```

dds/idl/marshal_generator.cpp
```diff
@@ -2956,13 +2955,24 @@ namespace {
...
+            "    if (!";
         }
         expr += generate_field_stream(
           indent, field, ">> stru" + value_access, wrap_nested_key_only, intro);
         if (is_appendable) {
           expr += ") {\n"
-            "    return false;\n"
+            "      return false;\n"
+            "    }\n";
+          if (cond.empty())
```
mitza-oci, line 2974: Add `{ }` around the statement under the `if`

mitza-oci (changes requested):

dds/idl/marshal_generator.cpp
```diff
@@ -2934,10 +2936,7 @@ namespace {
         // ends before some fields on the reader side get their values.
```
iguessthislldo, line 2936: Can you remove this comment since this PR basically fixes this? (We don't need to involve try construct since the spec just says non-optional fields are set to default.)

dds/idl/marshal_generator.cpp
```diff
@@ -2956,13 +2955,24 @@ namespace {
...
         } else if (is_appendable) {
-          expr += "  if (!";
+          AST_Type* const type = field->field_type();
+          const string stru_field_name = "stru" + value_access + "." + field_name;
+          expr +=
+            "  if (end_of_stream) {\n";
+          expr += type_to_default("    ", type, stru_field_name, type->anonymous());
+          expr +=
+            "  } else {\n";
+          expr +=
```
iguessthislldo, line 2965:
```suggestion
            "  if (end_of_stream) {\n" +
            type_to_default("    ", type, stru_field_name, type->anonymous()) +
            "  } else {\n"
```

dds/idl/marshal_generator.cpp
```diff
@@ -2777,6 +2777,10 @@ namespace {
       be_global->impl_ <<
         "  const Encoding& encoding = strm.encoding();\n"
         "  ACE_UNUSED_ARG(encoding);\n";
+      if (is_appendable) {
+        be_global->impl_ <<
+          "  bool end_of_stream = false;\n";
```
iguessthislldo, line 2782: This name could be misleading since the struct is not the whole stream.

```suggestion
          "  bool reached_end_of_struct = false;\n";
```

iguessthislldo (changes requested):

dds/idl/marshal_generator.cpp
jwillemsen: Done

jwillemsen:
Our AXCIOMA DDS4CCM test for extensibility now works correct with this PR

mitza-oci (approved):

tests/DCPS/XTypes/Publisher.idl
```diff
@@ -64,6 +70,12 @@ struct ModifiedNameMutableStruct {
   @id(2) long additional_field_modified;
 };
 
+@topic
+@mutable
+struct NestedMutableStruct {
+  @id(1) ModifiedNameMutableStruct nested_struct;
+};
+
```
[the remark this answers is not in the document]
jwillemsen: Correct, that is a new one for https://github.com/objectcomputing/OpenDDS/issues/3780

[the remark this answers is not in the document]
sonndinh: Okay. I think the issue ticket already describes the problem and we don't need to keep this type in the test since the test doesn't do anything with it.

iguessthislldo (approved):

PR #3775, opened 2022-10-06
Update comments in CorbaSeq's IDL files

sonndinh:


mitza-oci:
I recommend removing these kind of comments.  And I'd make the same recommendation for #3772's change.

mitza-oci (approved):

PR #3791, opened 2022-10-11
Convert DevGuideExamples Messenger Test Runner Script To Use PerlDDS::TestFramework

simpsont-oci:
Problem: The run_test.pl test runner script in DevGuideExamples/DCPS/Messenger doesn't currently use the PerlDDS::TestFramework, which means it doesn't gain the nice error reporting functionality we've added for stuck / hung / crashed processes.

Solution: Since we're running these in CI, we should update the test script (and add DCPSPendingTimeout while we're at it).

### DIFF
--- a/DevGuideExamples/DCPS/Messenger/run_test.pl
+++ b/DevGuideExamples/DCPS/Messenger/run_test.pl
@@ -15,7 +15,10 @@ use strict;
 my $status = 0;
 my $rtps = 0;
 my $help = 0;
-my $common_opts = 0;
+
+my $dbg_opts = " -ORBDebugLevel 1";
+my $common_opts = "$dbg_opts";
+
 my $help_message = "usage: run_test.pl [-h|--help] [--rtps]\n";
 my $invalid_args = not GetOptions(
   "rtps" => \$rtps,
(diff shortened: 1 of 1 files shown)


mitza-oci (approved):

DevGuideExamples/DCPS/Messenger/run_test.pl
```diff
@@ -2,117 +2,62 @@
...
 use PerlDDS::Run_Test;
-
 use strict;
 
 my $status = 0;
-
 my $rtps = 0;
 my $help = 0;
-
+my $common_opts = 0;
```
iguessthislldo, line 18: 0?

DevGuideExamples/DCPS/Messenger/run_test.pl
```diff
@@ -2,117 +2,62 @@
...
-if (!(-e $subdir.$filename) && !(-e $subdir.$filename_exe)) {
-    print STDERR "ERROR: publisher does not exist. Subdir: $subdir\n";
-    exit 1;
-}
+$test->{dcps_debug_level} = 4;
+$test->{dcps_transport_debug_level} = 2;
+$test->{add_transport_config} = 0;
+my $dbg_lvl = '-ORBDebugLevel 1';
+my $pub_opts = "$dbg_lvl" . "$common_opts";
+my $sub_opts = "$dbg_lvl" . "$common_opts";
```
iguessthislldo, line 51: These are the same, replace with with `common_opts`?

DevGuideExamples/DCPS/Messenger/run_test.pl
```diff
@@ -2,117 +2,62 @@
...
+my $repo_bit_opt = "";
```
iguessthislldo, line 52: This isn't being used

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #3793, opened 2022-10-11
Add NetworkAdddress::is_multicast

simpsont-oci:
Problem: During consideration of recent algorithm changes for IPv6 multicast support, it was determined that NetworkAddress had no direct support for determining whether a destination address was multicast or not, meaning that it would require conversion of the address to something like ACE_INET_Addr in order to make this check, which is not create for performance and code clarity.

Solution: Add the check to NetworkAddress (w/ unit test).

### DIFF
--- a/dds/DCPS/NetworkAddress.h
+++ b/dds/DCPS/NetworkAddress.h
@@ -65,6 +65,7 @@ public:
 
   bool is_any() const;
   bool is_loopback() const;
+  bool is_multicast() const;
 
   bool is_private() const; // IPv4 only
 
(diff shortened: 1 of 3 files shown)


dds/DCPS/NetworkAddress.h
```diff
@@ -65,6 +65,7 @@ class OpenDDS_Dcps_Export NetworkAddress {
 
   bool is_any() const;
   bool is_loopback() const;
+  bool is_multicast() const;
```
mitza-oci, line 68: Is there any existing code in the repository that could be changed to call this?

jrw972 (approved):

PR #3794, opened 2022-10-11
RtpsSendQueue is not scalable

jrw972:
Problem
-------

The RtpsSendQueue maintains maps for heartbeats and ackancks that must be merged.  These maps can become very large leading to performance problems.

Solution
--------

Do a final deduplication after all submessages have been enqueued.


dds/DCPS/transport/rtps_udp/MetaSubmessage.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "MetaSubmessage.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+bool MetaSubmessage::operator==(const MetaSubmessage& other) const
+{
+  ACE_Message_Block left_block(1024);
```
mitza-oci, line 15: Instead of a fixed length with a hard-coded constant, this could get the serialized size of the object.
Or if that's not important, at least check if serialization succeeded.

dds/DCPS/transport/rtps_udp/MetaSubmessage.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  ACE_Message_Block right_block(1024);
+  Serializer right(&right_block, Encoding::KIND_XCDR2);
+  right << sm_;
+
+  return src_guid_ == other.src_guid_ &&
+    dst_guid_ == other.dst_guid_ &&
+    left_block.length() == right_block.length() &&
+    std::memcmp(left_block.rd_ptr(), right_block.rd_ptr(), left_block.length()) == 0 &&
+    ignore_ == other.ignore_;
```
mitza-oci, line 27: check ignore first to short-circuit the more expensive checks

dds/DCPS/transport/rtps_udp/MetaSubmessage.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+      if (x.sm_._d() != y.sm_._d()) {
+        return x.sm_._d() < y.sm_._d();
+      }
+
+      // Order heartbeats and acknacks by count, descending.
+      switch (x.sm_._d()) {
+      case RTPS::HEARTBEAT:
+        return x.sm_.heartbeat_sm().count.value > y.sm_.heartbeat_sm().count.value;
+        break;
```
mitza-oci, line 50:
```suggestion
```

dds/DCPS/transport/rtps_udp/MetaSubmessage.cpp
```diff
@@ -0,0 +1,97 @@
...
+      case RTPS::ACKNACK:
+        return x.sm_.acknack_sm().count.value > y.sm_.acknack_sm().count.value;
+        break;
```
mitza-oci, line 53:
```suggestion
```

tests/unit-tests/dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <dds/DCPS/RTPS/BaseMessageUtils.h>
+#include <dds/DCPS/RTPS/MessageTypes.h>
+
+using namespace OpenDDS::DCPS;
+using namespace OpenDDS::RTPS;
+
+namespace
+{
+
+  MetaSubmessage create_heartbeat(const RepoId& from, const RepoId& dst, ACE_INT64 first, ACE_INT64 last, ACE_INT32 count, bool ignore)
```
mitza-oci, line 21:
```suggestion
MetaSubmessage create_heartbeat(const RepoId& from, const RepoId& dst, ACE_INT64 first, ACE_INT64 last, ACE_INT32 count, bool ignore)
```

or indent contents of namespace (which is preferred when the namespace is not most/all of the file)

tests/unit-tests/dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.h>
+
+#include <dds/DCPS/RTPS/BaseMessageUtils.h>
+#include <dds/DCPS/RTPS/MessageTypes.h>
+
+using namespace OpenDDS::DCPS;
+using namespace OpenDDS::RTPS;
+
+namespace
+{
```
mitza-oci, line 19:
```suggestion
namespace {
```

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.h
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+* A Transactional Send Queue Class For MetaSubmessages
+*
+* This class is designed to collect submessages from various threads
+* in a transactional way so they can be more efficiently bundled.
+*/
+class OpenDDS_Rtps_Udp_Export TransactionalRtpsSendQueue {
+public:
+  TransactionalRtpsSendQueue(ThreadStatusManager& thread_status_manager);
```
mitza-oci, line 29:
```suggestion
  explicit TransactionalRtpsSendQueue(ThreadStatusManager& thread_status_manager);
```

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.h
```diff
@@ -0,0 +1,77 @@
...
+
+  /// Signal that a thread is ending a sequence of submessages.
+  /// Returns true if the queue is non-empty.
+  bool end_transaction();
+
+  /// Return the number of active transactions.
+  size_t active_transaction_count() const;
+
+  /// Swap the queue into the given vector.
+  /// Note: calls clear() on the input vector before swapping
```
mitza-oci, line 50: Since `swap` is typically used to exchange two objects of the same type, it would be more clear to use a different name for this.

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.h
```diff
@@ -0,0 +1,77 @@
...
+  void swap(MetaSubmessageVec& vec);
+
+  /// Remove all queued submessage with the given source and destination
+  void purge(const RepoId& local, const RepoId& remote);
+
+  /// Remove all queued submessage with the given destination (dst_guid_)
+  void purge_remote(const RepoId& id);
+
+  /// Remove all queued submessage with the given source (src_guid_)
+  void purge_local(const RepoId& id);
```
mitza-oci, line 60: Since these now set the ignore_ flag instead of removing objects from the internal data structure, consider renaming and updating comments.

mitza-oci (changes requested):

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
```diff
@@ -0,0 +1,104 @@
...
+  return active_transaction_count_;
+}
+
+void TransactionalRtpsSendQueue::swap(MetaSubmessageVec& vec)
+{
+  vec.clear();
+
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  while (active_transaction_count_ != 0) {
+    condition_variable_.wait(thread_status_manager_);
```
simpsont-oci, line 65: This worries me, since it means that the writing thread (in this case, the transport's single EventDispatcher thread) will block, potentially preventing it from running other scheduled events. The wait will probably be small (I think the transport's reactor thread is the only thing using transactions), but this may have unintended side effects when the transport thread is particularly busy (esp within a single transaction) or as threading design changes within the transport framework.

dds/DCPS/transport/rtps_udp/MetaSubmessage.cpp
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+#include "MetaSubmessage.h"
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+bool MetaSubmessage::operator==(const MetaSubmessage& other) const
```
simpsont-oci, line 13: This is for use in tests only, correct? Any chance we can move it to a test utility directory so we know it's never getting used during production?

jrw972: Yes and yes.

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
jrw972: That's fair.  I'll change it so the periodic timer will mark the queue ready for harvest and the last transaction to finish will actually drain the queue.

mitza-oci (approved):

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
simpsont-oci: I think this puts a large number of writes (probably the majority?) back onto the transport thread, which is also not great. Perhaps if the send queue is busy (mid-transaction), the datalink just sets a flag that the next scheduling of flush_send_queue should be immediate (have zero delay) and then a successful flush can clear that flag. That should keep the writes on the eventdispatcher thread.

jrw972: I updated it so that the event thread attempts to harvest the send queue.  If successful, it will send.  If it is not successful, then other transactions are still in progress.  In this case, the other threads will harvest the queue and a `flush_send_queue` will be scheduled.  I'm using a SporadicEvent here, but the delay is always zero.  So let me know if there is a more correct way of just scheduling an event for immediate execution.

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2626,38 +2633,71 @@ RtpsUdpDataLink::bundle_mapped_meta_submessages(const Encoding& encoding,
...
-  bundle_and_send_submessages(fsq_vec_);
-  fsq_vec_.clear();
+  flush_send_queue_i();
+}
+
+void
+RtpsUdpDataLink::flush_send_queue_i()
+{
+  for (size_t idx = 0; idx != fsq_vec_size_; ++idx) {
+    dedup(fsq_vec_[idx]);
```
simpsont-oci, line 2656: If you're calling `sq_.end_transaction()`  every time `disable_response_queue()` is called, I think you wind up potentially getting a vector of submessages in `fsq_vec_` for every inbound packet. And if that's the case, calling `dedup()` and `bundle_and_send_submessages()` on them individually doesn't seem like it would let de-duplication or bundling help out all that much, since there wouldn't be much accumulated by that point.

jrw972: `end_transaction` will only produce a vector if `ready_to_send` has been called.  This occurs at the frequency of the harvest task so we should not get a vector per processed packet.

dds/DCPS/transport/rtps_udp/TransactionalRtpsSendQueue.cpp
simpsont-oci: Not sure if you saw this TSAN issue from the last run since the actions were semi-hidden for some reason: https://github.com/objectcomputing/OpenDDS/actions/runs/3242904729/jobs/5321460350 ... looks like something is maybe up with the `fsq_vec_` indexing (maybe `fsq_vec_size_` is out-of-sync with `fsq_vec_`?).

jrw972: They were.  See https://github.com/objectcomputing/OpenDDS/pull/3794/commits/dbd54c2c04aa372c804107d66e4dddaef6df4926

simpsont-oci (approved):

PR #3796, opened 2022-10-11
Bump gcc/msvc "Problem Matcher" Action Version

simpsont-oci:
Problem: Warning annotations referencing deprecation of actions based on Node.js 12: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12

Solution: Bump to newer version of actions which are based on Node.js 16.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -177,7 +177,7 @@ jobs:
         cd OpenDDS
         . setenv.sh
         tools/scripts/show_build_config.pl
-    - uses: ammaraskar/gcc-problem-matcher@0.1
+    - uses: ammaraskar/gcc-problem-matcher@0.2.0
     - name: build OpenDDS
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


mitza-oci (approved):

simpsont-oci:
I bumped a single outdated `cache` action as well to avoid the warning from it, but unfortunately `run-vcpkg` doesn't have any node16 versions available yet so it will have to wait. Also worth noting is that `run-vcpkg` seems to have a big version jump from v7.4 (our current version) to v10 (v10.5) and the commands are different so it's not a simple version bump. Depending on how they choose to handle the node16 upgrades, we may need to figure out how to port our actions to v10 syntax or look for an alternative action for configuring vcpkg.

jwillemsen:
I already ported ace to run-vcpkg v10, see there for the changes

mitza-oci:
From GitHub:
> We plan to migrate all actions to run on Node16 by Summer 2023. 

So we have some time to move to the newer `run-vcpkg` or replace it.

mitza-oci:
> I already ported ace to run-vcpkg v10, see there for the changes

Thanks, I'm sure that will be helpful.
Even if we changed it now, `action.yml` in `lukka/run-vcpkg` (main branch) is using node v12 so it wouldn't resolve the warning.

mitza-oci:
> > I already ported ace to run-vcpkg v10, see there for the changes
> 
> Thanks, I'm sure that will be helpful. Even if we changed it now, `action.yml` in `lukka/run-vcpkg` (main branch) is using node v12 so it wouldn't resolve the warning.

related to https://github.com/lukka/run-vcpkg/issues/163

PR #3799, opened 2022-10-12
Improvements For Bench fan* Test Scenarios

simpsont-oci:
Problem: Bench fan_frag and (especially) fan_frag_ws test scenarios still fail with relatively high frequency in CI (where their execution has been limited) and on the DDS Scoreboard (especially on Windows VMs).

Partial Solution:
 - Improve Bench scenario logging
 - Improve coordination between Bench test processes by enforcing relative test phase times up-front in the test controller
 - And add a new per-writer expected sample count
 - Improve overall and per-writer expected sample count error checking
 - Limit RTPS to fragment replays when whole seq is NACKed (don't naively perform full replay)

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendBuffer.cpp
+++ b/dds/DCPS/transport/framework/TransportSendBuffer.cpp
@@ -372,6 +372,12 @@ SingleSendBuffer::check_capacity_i(BufferVec& removed)
 }
 
 bool
+SingleSendBuffer::has_frags(const SequenceNumber& seq) const
+{
+  return fragments_.find(seq) != fragments_.end();
+}
+
+bool
 SingleSendBuffer::resend(const SequenceRange& range, DisjointSequence* gaps)
 {
   ACE_GUARD_RETURN(LockType, guard, strategy_lock(), false);
--- a/dds/DCPS/transport/framework/TransportSendBuffer.cpp
+++ b/dds/DCPS/transport/framework/TransportSendBuffer.cpp
@@ -41,7 +41,6 @@ TransportSendBuffer::retain_all(const RepoId&)
 void
 TransportSendBuffer::resend_one(const BufferType& buffer)
 {
-  //ACE_DEBUG((LM_DEBUG, "(%P|%t) SSB::resend_one - &buffer = %@\n", &buffer));
   int bp = 0;
   strategy_->do_send_packet(buffer.second, bp);
 }
@@ -428,12 +427,9 @@ SingleSendBuffer::resend_i(const SequenceRange& range, DisjointSequence* gaps,
       } else {
         const FragmentMap::iterator fm_it = fragments_.find(it->first);
         if (fm_it != fragments_.end()) {
-          //ACE_DEBUG((LM_DEBUG, "(%P|%t) SSB::resend_i for seq %Q - resending all fragments\n", it->first.getValue()));
-          size_t count = 0;
           for (BufferMap::iterator bm_it = fm_it->second.begin();
-                count < 5 && bm_it != fm_it->second.end(); ++bm_it) {
+                bm_it != fm_it->second.end(); ++bm_it) {
             resend_one(bm_it->second);
-            ++count;
           }
         }
       }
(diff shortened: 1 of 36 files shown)


jrw972 (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
```diff
@@ -268,8 +268,9 @@ RtpsUdpSendStrategy::send_multi_i(const iovec iov[], int n,
 {
   ssize_t result = -1;
   typedef AddrSet::const_iterator iter_t;
+  const NetworkAddress empty = NetworkAddress();
   for (iter_t iter = addrs.begin(); iter != addrs.end(); ++iter) {
-    if (*iter == NetworkAddress()) {
+    if (*iter == empty) {
```
iguessthislldo, line 273: Grepping it looks like we do checks like this in several places, would it make sense to add a boolean overload to `NetworkAddress` instead of comparing it to an empty object that has to be constructed each time?

performance-tests/bench/node_controller/main.cpp
```diff
@@ -187,18 +189,27 @@ class SpawnedProcess {
...
     running_ = true;
   }
 
+  void log_process_creation()
+  {
+    std::stringstream ss;
+    ss << Bench::iso8601() << ": Node " << node_id_ << " (" << node_name_ << ") starting process " << pid_ << ":\n";
+    ss << " - original config: " << original_config_name_ << "\n";
+    ss << " - command line: " << command_line_ << "\n\n" << std::flush;
+    std::cerr << ss.str() << std::flush;
```
iguessthislldo, line 210: Why put everything into a string stream first and not into `cerr` directly? Also I don't think string stream has anything to flush unlike for file I/O.

iguessthislldo (changes requested):

mitza-oci:
Please update PR title since it's not limited to Bench

performance-tests/bench/node_controller/main.cpp
simpsont-oci: Since cerr is a shared resource, putting it all into a stringstream first reduces the number of operations on cerr and prevents having the lines from being split up when multiple threads and/or processes are writing at the the same time.

dds/DCPS/transport/rtps_udp/RtpsUdpSendStrategy.cpp
simpsont-oci: That's what this was intended to do, though I suppose you're asking about caching the empty value in a more globally accessible location. I'll add `operator bool()` and `bool operator!()` for `NetworkAddress` but I don't think this is right PR to switch everything over to using them.

dds/DCPS/transport/framework/TransportSendBuffer.cpp
```diff
@@ -370,6 +370,12 @@ SingleSendBuffer::check_capacity_i(BufferVec& removed)
   }
 }
 
+bool
+SingleSendBuffer::has_frags(const SequenceNumber& seq) const
+{
+  return fragments_.find(seq) != fragments_.end();
```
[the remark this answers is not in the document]
simpsont-oci: tomAYto / tomAHto

I think I prefer the aesthetic of using find rather than count, since I don't really care about the number. Looking at the implementation, count just wraps what I've written with an extra layer that returns `1` or `0` depending on the comparison. For C++14 or greater, there's also the potential confusion of accidentally triggering the templated version of count which allows for direct comparisons of non-key types against the keys in the map... something to avoid when we actually have the correct key type.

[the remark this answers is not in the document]
simpsont-oci: Looking at cppreference, it looks like find also gets the templated non-key version in C++14, so I guess it really boils down to whether the brevity of `count` is worth the wrapping.

performance-tests/bench/node_controller/main.cpp
iguessthislldo: Alright, that makes sense. What about the flush?

iguessthislldo (approved):

dds/DCPS/transport/framework/TransportSendBuffer.cpp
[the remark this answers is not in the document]
mitza-oci: Not a big deal at all (I didn't make it "requested changes") -- in general, without knowing anything about how it's implemented, prefer using a simpler function.  With map (as opposed to multimap) count semantically returns a bool (0 or 1) but the type is int to make it compatible with multimap.

simpsont-oci:
Barring any more review comments, I think this PR is probably ready. The latest TSAN failure probably needs to be looked into at some point but looks to be intermittent and is not caused by anything in this PR.

PR #3820, opened 2022-10-21
Foundation TAB mintues 2022-09-12

halflauren:
Add September 2022 Tab Meeting Minutes


iguessthislldo:
@halflauren I don't think there's anything wrong with this page, so you don't have to do anything about this.

It looks like adding a new foundation minutes page is breaking the website generation again. It looks different from #3556 though.

I think the problem is in what's generating the header links: https://github.com/objectcomputing/OpenDDS/blob/68fd3dd755ea5d94b7da05f6e0e2a0e77e96e548/_includes/header.html#L43

The resulting HTML has this at the start:

```
<ul class="trigger">
                <li><a class="page-link" href="/pages/objectcomputing/OpenDDS/foundation/minutes/2022-09-12.html">
                        
                    OpenDDS Foundation TAB Meeting Minutes 9/12/22 
                  
                </a></li>

            </ul>
```

That `</ul>` and probably that whole `<li>` for the new minutes shouldn't be here.

foundation/minutes/2022-09-12.md
```diff
@@ -0,0 +1,193 @@
+---
+layout: post
+title: "OpenDDS Foundation TAB Meeting Minutes 9/12/22"
+categories: news
+author: OpenDDS Foundation
```
iguessthislldo, line 5:
```suggestion
layout: foundation_minutes
date: "2022-09-12"
type: "foundation-minutes"
order: 206
sitemap: false
CSS: foundation-minutes.css
```

To match the other foundations minutes. I built the site locally with this change and it fixes the problem I described.

iguessthislldo (changes requested):
I can see the new page, but the markdown isn't rendering:

Screenshot from 2022-10-24 20-24-17

We either have to manually convert this to HTML or figure out how to get Jekyll to do it for us like it does for regular posts.

PR #3821, opened 2022-10-22
Add a `opendds_mwc.pl` Wrapper Script

iguessthislldo:
This script uses the same `mwc.pl` arguments (minus the workspace file) that OpenDDS was configured with. This is helpful for cases where ACE/TAO was separately configured in a generic way. In this case the default features file doesn't get updated and each feature has to be passed if running `mwc.pl` manually. This script will contain the features as well as the `-type` and `-static` if needed.

### DIFF
--- a/configure
+++ b/configure
@@ -21,6 +21,7 @@ use File::Temp ();
 use FileHandle;
 use Cwd;
 use POSIX qw(strftime);
+use B qw/perlstring/;
 
 use FindBin;
 use lib "$FindBin::RealBin/tools/scripts/modules";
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -2056,6 +2057,29 @@ sub get_features {
...
+#!/usr/bin/env perl
+if (!defined(\$ENV{ACE_ROOT}) || !defined(\$MPC_ROOT)) {
+  die("The enviroment needs to be setup.");
+}
+exec('perl', "\$ENV{ACE_ROOT}/bin/mwc.pl", $args, \@ARGV)
+  or die("Failed to exec mwc: \$!");
+EOF
+  $fd->close();
+  print "Wrote $path\n" if $opts{'verbose'};
+  dump_and_unlink($fd) if $opts{'dry-run'};
```
simpsont-oci, line 2078: If `opendds_mwc.pl` already exists and then `configure` is run with the `dry-run` flag... it looks like this makes a backup but deletes the old one? Is that the desired behavior? I guess theoretically you shouldn't be re-running configure, even with `dry-run` but it seems like you'd want to preserve the existing `opendds_mwc.pl` in that case.

simpsont-oci (approved):

configure
iguessthislldo: If it's dry run, then the file descriptor it's unlinking inside `dump_and_unlink` is for a temporary file: https://github.com/objectcomputing/OpenDDS/blob/e1d588f4559a2dfedf817c9844e132dfea00fad6/configure#L1022-L1024

PR #3831, opened 2022-10-25
[Bench] Worker Config Cleanup

simpsont-oci:
Problem: The bench worker config defines quasi-useless "Discovery" and "TransportInstance" configuration sections. These do not support multiple transports instances per config section, and they do not allow for the pass through of important configuration properties. Including the actual config sections from a real OpenDDS config file is much more useful / flexible.

Solution: Remove these configuration options, update config files to use real config sections.


iguessthislldo:
Does the documentation also need to be updated? https://github.com/objectcomputing/OpenDDS/blame/master/docs/internal/bench2.rst#L359

simpsont-oci:
Yeah, definitely, and the nightly configs, and I'm renaming the configs too so they will be more consistent and updating dcps_test.lst

iguessthislldo (approved):

PR #3834, opened 2022-10-27
Fix Coverity Issues

jrw972:


dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -188,6 +188,7 @@ void
 DataLink::invoke_on_start_callbacks(bool success)
 {
   const DataLink_rch link(success ? this : 0, inc_count());
+  RepoId remote = GUID_UNKNOWN;
```
mitza-oci, line 191: Why did this definition move?  In general, definition closest to when it's needed is preferred.

mitza-oci (approved):

PR #3836, opened 2022-10-27
Fix CI brew xerces-c root

simpsont-oci:
Problem: The xerces-c version installed by brew increased by a patch version, so the path we were using for it is no longer valid.

Solution: Fix the path issue by using find to locate the correct version directory of xerces-c 3.x


mitza-oci (approved):

mitza-oci:
Looks like this needs to be resolved before we merge https://github.com/objectcomputing/OpenDDS/actions/runs/3339867499/jobs/5530271204#step:12:37

PR #3840, opened 2022-10-28
RtpsRelay Admission Control

simpsont-oci:
Problem: Under periods of abnormally high levels of discovery traffic (such as during overall system redeployment), the RtpsRelay struggles to maintain an optimal rate of participant discovery, in effect worsening the experienced problem of slow discovery times.

Solution: Add an admission control queue in order to only allow a limited subset of the unknown remote participants to go through discovery at once. The admission control queue maximum size (total undergoing discovery at a time) and admittance duration (time considered sufficient for aiding discovery) are both configurable.

### DIFF
--- a/tools/rtpsrelay/RelayStatisticsReporter.h
+++ b/tools/rtpsrelay/RelayStatisticsReporter.h
@@ -151,6 +151,13 @@ public:
     report(now);
   }
 
+  void admission_deferral_count(const OpenDDS::DCPS::MonotonicTimePoint& now)
+  {
+    ++log_relay_statistics_.admission_deferral_count();
+    ++publish_relay_statistics_.admission_deferral_count();
+    report(now);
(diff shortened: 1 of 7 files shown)


tools/rtpsrelay/RelayStatisticsReporter.h
```diff
@@ -151,6 +151,13 @@ class RelayStatisticsReporter {
     report(now);
   }
 
+  void admission_deferral_count(const OpenDDS::DCPS::MonotonicTimePoint& now)
+  {
+    ++log_relay_statistics_.admission_deferral_count();
```
jrw972, line 156: Reset to zero after logging or publishing.

jrw972 (changes requested):

jrw972 (approved):

PR #3848, opened 2022-11-03
Add a set runtime to RtpsRelay and enforce clean shutdown

simpsont-oci:
Problem: For some testing use cases, it's helpful to have the RtpsRelay end "naturally" as opposed to being killed.

Solution: For these use cases, add a "RunTime" parameter which will cause the RtpsRelay to end naturally after the desired time.

### DIFF
--- a/tools/rtpsrelay/RtpsRelay.cpp
+++ b/tools/rtpsrelay/RtpsRelay.cpp
@@ -207,6 +207,9 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
     } else if ((arg = args.get_the_parameter("-AdmissionControlQueueDuration"))) {
       config.admission_control_queue_duration(OpenDDS::DCPS::TimeDuration(ACE_OS::atoi(arg)));
       args.consume_arg();
+    } else if ((arg = args.get_the_parameter("-RunTime"))) {
+      config.run_time(OpenDDS::DCPS::TimeDuration(ACE_OS::atoi(arg)));
+      args.consume_arg();
 #ifdef OPENDDS_SECURITY
     } else if ((arg = args.get_the_parameter("-IdentityCA"))) {
       identity_ca_file = file + arg;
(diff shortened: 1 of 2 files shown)


tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -953,10 +959,26 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
...
+
   } else {
     reactor->run_reactor_event_loop();
   }
 
+  application_participant->delete_contained_entities();
+  factory->delete_participant(application_participant.in());
+
+  relay_participant->delete_contained_entities();
+  factory->delete_participant(relay_participant.in());
```
mitza-oci, line 978:
```suggestion
  factory->delete_participant(application_participant);

  relay_participant->delete_contained_entities();
  factory->delete_participant(relay_participant);
```

mitza-oci (approved):

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -936,6 +939,9 @@ int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
   }
   ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) INFO: Meta Discovery listening on %C\n"), OpenDDS::DCPS::LogAddr(meta_discovery_addr).c_str()));
 
+  const bool has_run_time = config.run_time() != OpenDDS::DCPS::TimeDuration::zero_value;
```
iguessthislldo, line 942: `TimeDuration` can be implicitly used as a `bool` to check if it's zero, so you could replace `has_run_time` with  `config.run_time()`.

iguessthislldo: Or replace this line with:

```suggestion
  const bool has_run_time = !config.run_time().is_zero();
```

PR #3849, opened 2022-11-03
Minor Performance Improvements for TimeDuration

simpsont-oci:
From Profiling RtpsRelay. Fewer temporary objects, fewer function calls.


iguessthislldo (approved):
Yea, not exactly sure why I did this.

PR #3864, opened 2022-11-08
SignedDocument can't handle non-openssl documents

jrw972:
Problem
-------

Loading a governance file or permissions file signed by Bouncy Castle fails.  The root cause is the (lack of) MIME parsing.

Solution
--------

Use openssl to extract the signed content.

### DIFF
--- a/dds/DCPS/security/SSL/SignedDocument.cpp
+++ b/dds/DCPS/security/SSL/SignedDocument.cpp
@@ -9,6 +9,8 @@
 
 #include <dds/DCPS/security/CommonUtilities.h>
 
+#include <dds/DCPS/Definitions.h>
+
 #include <openssl/pem.h>
 
 #include <cstring>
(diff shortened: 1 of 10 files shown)


dds/DCPS/security/SSL/SignedDocument.h
```diff
@@ -42,73 +35,24 @@ class OpenDDS_Security_Export SignedDocument {
...
-  int verify_signature(const Certificate& ca) const;
-
-  /**
-   * @return int 0 on success; 1 on failure.
-   */
-  int serialize(DDS::OctetSeq& dst) const;
+  const DDS::OctetSeq& original() const {return original_;}
+  const std::string& content() const {return content_;}
+  bool verified() const { return verified_; }
+  const std::string filename() const {return filename_;}
```
mitza-oci, line 43:
```suggestion
  const std::string& filename() const { return filename_; }
```
OR
```suggestion
  std::string filename() const { return filename_; }
```

dds/DCPS/security/SSL/SignedDocument.cpp
```diff
@@ -28,59 +28,40 @@ namespace {
...
-  deserialize(src);
 }
 
 SignedDocument::SignedDocument(const SignedDocument& rhs)
-  : doc_(0)
-  , original_()
-  , verifiable_("")
+  : original_(rhs.original_)
+  , content_(rhs.content_)
+  , verified_(rhs.verified_)
```
mitza-oci, line 50: Looks like this turned in to the same as the compiler-generated copy constructor.  Could we just use that (similar for copy assignment below).

dds/DCPS/security/SSL/SignedDocument.cpp
```diff
@@ -127,172 +108,67 @@ bool SignedDocument::load(const std::string& uri, DDS::Security::SecurityExcepti
...
+            PKCS7_free(doc);
+          } else {
+            OPENDDS_SSL_LOG_ERR("SMIME_read_PKCS7 failed");
+          }
+        } else {
+          OPENDDS_SSL_LOG_ERR("BIO_write failed");
+        }
+        BIO_free(filebuf);
+      } else {
+        OPENDDS_SSL_LOG_ERR("BIO_new failed");
```
mitza-oci, line 155: Nesting this deep gets hard to read and the error logging (the line right here) is far away from the line that caused it.

Using RAII for the `*_free` functions allows the method to exit when it needs to.  An example is the `hkdf()` in `CryptoBuiltinImpl.cpp`

mitza-oci (changes requested):

mitza-oci (approved):

PR #3867, opened 2022-11-09
[Bench] Add Scenario and Test Run Parameters For Use In Performance Dashboard

simpsont-oci:
Problem: The [OpenDDS Performance Dashboard](https://github.com/OpenDDS/opendds-performance-dashboard) currently must attempt to deduce certain kinds of metadata about various bench test scenarios from output filenames. This makes it difficult to add new test scenarios and is prone to breakages when file names change.

Solution: Allow scenarios to specify both "test run" and "scenario" parameter metadata that will get passed through to the test result output, and then eventually through the report_parser and dashboard_summarizer for eventual consumption and use by the OpenDDS Performance Dashboard.

Disclaimer: Merging these changes will cause temporary incompatibilities with the existing performance dashboard, as the output format for the dashboard_summarizer has changed to more-generically represent the new parameter data (removing the old  "scenario" / "variety" hierarchy).

Disclaimer, part 2: Before merging these changes, any existing performance dashboard scenarios should be updated with appropriate scenario parameters (see updates to ci-*.json scenario files for examples) and the dashboard Jenkins instance should be updated to add test run parameters via the test_controller command line (most likely via the run scripts in the test environment).

### DIFF
--- a/performance-tests/bench/dashboard_summarizer/main.cpp
+++ b/performance-tests/bench/dashboard_summarizer/main.cpp
@@ -18,7 +18,9 @@ Value& find_or_create(Value& parent, const std::string& name, Type type, Documen
     pos = parent.FindMember(Value(name.c_str(), length));
     assert(pos != parent.MemberEnd());
   }
-  assert(pos->value.GetType() == type);
+  if (type != kNullType) {
+    assert(pos->value.GetType() == type);
+  }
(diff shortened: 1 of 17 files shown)


simpsont-oci:
Dashboard configs and scripts should be ready now.

PR #3887, opened 2022-11-21
ReceivedDataSample: fixed issues with swap and assign

mitza-oci:


dds/DCPS/transport/framework/ReceivedDataSample.h
```diff
@@ -112,6 +112,8 @@ class OpenDDS_Dcps_Export ReceivedDataSample {
     size_t rd_ptr_, wr_ptr_;
   };
 
+  friend void swap(MessageBlock& lhs, MessageBlock& rhs);
```
simpsont-oci, line 115: Make this a public member to avoid the friend declaration?

From [https://en.cppreference.com/w/cpp/algorithm/swap](https://en.cppreference.com/w/cpp/algorithm/swap): `The expected way to make a program-defined type swappable is to provide a non-member function swap in the same namespace as the type: see [_Swappable_](https://en.cppreference.com/w/cpp/named_req/Swappable) for details.

But if you go there, the example provided has `swap()` as a member function, where a global with the same name calls the public member.

simpsont-oci (approved):

PR #3908, opened 2022-12-02
Use WeakRcHandle For TransportImpl In DataLink

simpsont-oci:
Problem: Using a reference to TransportImpl in DataLink can cause memory access problems when the TransportImpl disappears for some reason (as part of shutdown) before the DataLink has been removed. See [this CI error](https://github.com/objectcomputing/OpenDDS/actions/runs/3587167129/jobs/6038276947) for an example.

Solution: Have DataLink use a WeakRcHandle to TransportImpl instead, and lock / check before accessing the TransportImpl. Also update a number of APIs that pass or use the reference to use RcHandle instead (and check when possible).


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -3230,8 +3230,13 @@ DataReaderImpl::add_link(const DataLink_rch& link, const RepoId& peer)
...
   TransportClient::add_link(link, peer);
-  TransportImpl& impl = link->impl();
-  OPENDDS_STRING type = impl.transport_type();
+  OPENDDS_STRING type;
+  {
+    RcHandle<TransportImpl> impl = link->impl();
+    if (impl) {
+      type = impl->transport_type();
+    }
+  }
```
jrw972, line 3239: I cried a little when I looked at the following code.

dds/DCPS/transport/framework/DataLink.h
```diff
@@ -250,7 +250,7 @@ class OpenDDS_Dcps_Export DataLink
   /// targets of this DataLink (see is_target()).
   GUIDSeq* target_intersection(const RepoId& pub_id, const GUIDSeq& in, size_t& n_subs);
 
-  TransportImpl& impl() const;
+  RcHandle<TransportImpl> impl() const;
```
jrw972, line 253: Return the `WeakRcHandle`.  This will force the using code to come to grips with the reality that the transport might not be there.

dds/DCPS/transport/framework/DataLink.cpp
```diff
@@ -971,7 +994,8 @@ DataLink::release_resources()
   DBG_ENTRY_LVL("DataLink", "release_resources", 6);
 
   this->prepare_release();
-  return impl_.release_link_resources(this);
+  RcHandle<TransportImpl> impl = impl_.lock();
+  return impl->release_link_resources(this);
```
mitza-oci, line 998: Does this need to be checked before using?  (Also line 1078 below)

dds/DCPS/transport/multicast/MulticastSendStrategy.cpp
```diff
@@ -16,7 +16,7 @@ namespace OpenDDS {
 namespace DCPS {
 
 MulticastSendStrategy::MulticastSendStrategy(MulticastDataLink* link)
-  : TransportSendStrategy(0, link->impl(),
+  : TransportSendStrategy(0, *link->impl(),
```
jrw972, line 19: Is this not setting up the same kind of problem?

dds/DCPS/transport/tcp/TcpConnection.cpp
```diff
@@ -593,19 +598,21 @@ OpenDDS::DCPS::TcpConnection::active_reconnect_i()
...
     TcpConnection* pconn = this;
     int ret;
     {
-      ACE_Reverse_Lock<LockType> rev_lock(this->reconnect_lock_);
-      ACE_Guard<ACE_Reverse_Lock<LockType> > guard(rev_lock);
-      // We need to temporarily release the lock here because the connect could occasionally be synchronous
-      // if the source and destination are on the same host. When the call become synchronous, active_reconnect_open()
-      // would be called and try to acquired the lock in the same thread.
-      ret = transport.connector_.connect(pconn, this->remote_address_,  ACE_Synch_Options::asynch);
+      RcHandle<TcpTransport> transport = dynamic_rchandle_cast<TcpTransport>(link_->impl());
```
mitza-oci, line 607: `ret` above this needs initialization

dds/DCPS/transport/tcp/TcpDataLink.h
```diff
@@ -26,7 +26,7 @@ class TcpDataLink : public DataLink {
 public:
 
   TcpDataLink(const ACE_INET_Addr& remote_address,
-              TcpTransport& transport_impl,
+              RcHandle<OpenDDS::DCPS::TcpTransport> transport_impl,
```
mitza-oci, line 29:
```suggestion
              RcHandle<TcpTransport> transport_impl,
```

mitza-oci (changes requested):

dds/DCPS/transport/tcp/TcpConnection.cpp
```diff
@@ -593,19 +598,21 @@ OpenDDS::DCPS::TcpConnection::active_reconnect_i()
...
+      if (transport) {
+        ACE_Reverse_Lock<LockType> rev_lock(this->reconnect_lock_);
+        ACE_Guard<ACE_Reverse_Lock<LockType> > guard(rev_lock);
+        // We need to temporarily release the lock here because the connect could occasionally be synchronous
+        // if the source and destination are on the same host. When the call become synchronous, active_reconnect_open()
+        // would be called and try to acquired the lock in the same thread.
+        ret = transport->connector_.connect(pconn, this->remote_address_,  ACE_Synch_Options::asynch);
```
jrw972, line 614: Initialize ret before if.

jrw972 (changes requested):
Inside the transport framework, I'm okay with strong references because the lifetime of the various objects should be bounded by the lifetime of the transport itself.  The bug you describe means that this assumption is being violated, i.e., the datalink is surviving past the transport.  Thus, fixing that would address the issue.  If that can't be handled in a clean way, then we should use weak references to the datalink itself so the rest of the middleware can correctly deal with situations where the datalink disappears.

dds/DCPS/transport/multicast/MulticastSendStrategy.cpp
simpsont-oci: Yes, probably, but I was hesitant to dig into the strategies as well. I guess I can add it.

dds/DCPS/transport/framework/DataLink.h
simpsont-oci: I think this would cause undue pain... we should be checking the return value, but having callers perform their own lock / check seems like it would generate a lot of clutter...

jrw972: The clutter already has to be there, i.e., every caller of `impl` has to check that it is null.  By returning a WeakRcHandle, the compiler ensures that every caller is performing the check.

mitza-oci: Both the strong and weak handles should be checked by the caller.  I agree there's some advantage to using the weak handle as the return type in order to force callers to do this.  However, it is different in terms of semantics of how the objects relate to each other.  Here's an example:

```
class A {
public:
  SomeHandle<B> op();
private:
  WeakRcHandle<B> b_;
  // other members, mutex
};
```

In this scenario, we know that A's data member has to be a weak handle.  Within the method `op()`, it acquires the mutex and examines other internal state.  As part of this method it locks the weak handle.  If this fails, the method must return a null handle (of either kind).  If it succeeds, returning the strong handle means that extending the lifetime of B was one of the postconditions of this method.  Returning the weak handle gives the caller the potential to use B, but if no other strong pointer to B exists then the object goes away before the caller can do its own locking.

dds/DCPS/Definitions.h
```diff
@@ -66,6 +66,9 @@
 #  define OPENDDS_ASSERT(C) assert(C)
 #endif
 
+#define INLINE_TEST_AND_CALL(TYPE, TEST, CALL) do { TYPE temp = TEST; if (temp) { temp->CALL; } } while (false);
+#define INLINE_TEST_AND_CALL_ASSIGN(TYPE, TEST, CALL, VAL) do { TYPE temp = TEST; if (temp) { VAL = temp->CALL; } } while (false);
```
mitza-oci, line 70: Use `OPENDDS_*` prefix on these names

dds/DCPS/transport/framework/DataLink.h
simpsont-oci: I've added checks for most uses now, but the config() calls are a bit trickier because the config signature (returns a reference) assumes it can't fail. This probably needs to become a RCH as well...

simpsont-oci: OK, configs are RCH now as well, and should be checked most of the time before use now.

iguessthislldo (changes requested):

dds/DCPS/transport/multicast/MulticastTransport.cpp
```diff
@@ -202,8 +207,10 @@ MulticastTransport::accept_datalink(const RemoteTransport& remote,
                                     const ConnectionAttribs& attribs,
                                     const TransportClient_rch& client)
 {
+  MulticastInst_rch cfg = config();
+
   // Check that the remote reliability matches.
-  if (get_remote_reliability(remote) != this->config().is_reliable()) {
+  if (!cfg || get_remote_reliability(remote) != cfg->is_reliable()) {
     return AcceptConnectResult();
```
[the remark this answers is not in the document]
simpsont-oci: Seems reasonable.

mitza-oci (approved):

iguessthislldo (approved):

PR #3912, opened 2022-12-05
Quick shutdown leads to error in LinuxNetworkConfigMonitor

jrw972:
Problem
-------

As part of startup, the Service Participant opens the socket for the LinuxNetworkConfigMonitor and then uses the ReactorInterceptor to register it with the reactor.  As part of shutdown, the Service Participant closes the socket for the LinuxNetworkConfigMonitor and then uses the ReactorInterceptor to unregister.  If the shutdown happens immediately after startup, then the reactor thread will see a closed socket, i.e., the order of operations is:

1. open
2. close
3. register -> error, invalid handle

Solution
--------

Defer all of the logic to the ReactorInterceptor commands.

Note
----

The NetworkConfigMonitor interface expects a synchronous open and close.  It does this so it can fall back from the
LinuxNetworkConfigMonitor to other implementations.  To satisfy this while ensuring the correct order of operations, conditions were added to the ReactorInterceptor commands.  In general, this is dangerous but none of the alternatives for fixing it are attractive at this point:

* Force it upon the user, i.e., the user is responsible for writing to the `network_interface_address_topic`
* Replacing ACE


mitza-oci (commented):
Is there an advantage to having a timeout on the waits?  Any drawbacks?

jrw972:
> Is there an advantage to having a timeout on the waits? Any drawbacks?

I don't think there is an advantage to adding a wait.  The code that gets dispatched to the reactor thread should execute promptly and if it doesn't, that means the thread spawn or something similarly awful has happened.

If the wait finished before the NetworkConfigMonitor is fully operational (because of a pokey system maybe), then we fall back and loose out on the dynamic configuration.

All in all, I would say we don't need to have a timeout on the waits.

mitza-oci (approved):

PR #3916, opened 2022-12-06
Fix warnings from Linux builds

ocielliottc:


dds/idl/langmap_generator.cpp
```diff
@@ -1421,6 +1421,25 @@ struct Cxx11Generator : GeneratorBase
   bool scoped_enum() { return true; }
   std::string enum_base() { return " : uint32_t"; }
 
+  void gen_union_pragma_pre()
+  {
+    // Older versions of gcc will complain because it appears that a primitive
+    // default constructor is not called for anonymous unions.
+    be_global->lang_header_ <<
+      "#ifdef __GNUC__\n"
```
jwillemsen, line 1429: Maybe generate a gcc version check? That way you know when to drop this workaround?

ocielliottc: Maybe.  I'd have to test which versions report the warning.

configure
```diff
@@ -1045,6 +1050,9 @@ sub write_config_h {
   $opts{'optimize'} = 0 if (!exists $opts{'optimize'} || exists $opts{'sanitize'});
 
   my $CFGH = backup_and_open("$buildEnv{'ACE_ROOT'}/ace/config.h");
+  for my $line (@default_configh) {
```
mitza-oci, line 1053: It would be good to have a new command-line option for configure that lets the user turn this off.

dds/DCPS/transport/multicast/MulticastInst.cpp
```diff
@@ -88,7 +88,8 @@ MulticastInst::load(ACE_Configuration_Heap& cf,
   GET_CONFIG_VALUE(cf, sect, ACE_TEXT("port_offset"),
                    this->port_offset_, u_short)
 
-  ACE_TString group_address_s;
+  // Explicitly initialize this string to stop gcc 11 from issuing a warning.
+  ACE_TString group_address_s(ACE_TEXT(""));
```
mitza-oci, line 92: What's the warning for this?  It seems odd that the compiler doesn't like it since the class has a default constructor.

configure
ocielliottc: Can do.

dds/DCPS/transport/multicast/MulticastInst.cpp
ocielliottc: It was this one:

‘void* __builtin_memcpy(void*, const void*, long unsigned int)’ specified bound 18446744073709551615 exceeds maximum object size 9223372036854775807

The default constructor didn't do anything wrong though.  I'm not 100% certain why initializing it to an empty string makes the warning disappear.

dds/idl/langmap_generator.cpp
jwillemsen: Instead of suppressing the warning, you can also use default/zero initialization, see TAOX11. See https://github.com/RemedyIT/idl2cpp11/blob/39ca4d118e9bf51e8f25cad906926472b857e6a2/examples/union/testC.h#L313 for some example code, I have the PR https://github.com/RemedyIT/taox11/pull/239 to reduce the amount of ifdef checks, only for the member that has zero initialization we need the ifdef

ocielliottc: I think this situation is slightly different.  This is an anonymous union relying on the default constructor.  So, when there is a string in the union, it calls that constructor.  But, when it does not contain a complex data type, it **should** call a constructor for one of the types.  I believe it actually does.  But, the gcc warning doesn't realize it?

jwillemsen: We had similar warnings related to unions and switched to use default which solved it for us. Maybe handcraft it and see if this also helps with opendds

ocielliottc: That's a good suggestion.  That may get done for v4.

jwillemsen: Also changing optimization level will probably help

dds/DCPS/transport/multicast/MulticastInst.cpp
mitza-oci: From what I can tell, it's warning on the operator= usage that's part of the macro expansion on the next line.  The GCC warning is that running this line from String_Base: ` this->rep_[len] = 0;` when rep_ points to the special "null string" value is invalid.  And if you only analyze the code at the level, it might make some sense...  
However, the warning is actually bogus since it's impossible to reach that line of code when rep_ points to NULL_String_.

The other error, the one with the very large numbers that points back to ACE's Configuration.cpp, seems to be a real bug in v6 which is fixed in v7.

ocielliottc: That seems like a reasonable explanation.  Nothing looked wrong to me, code-wise.

I just saw that ACE Configuration.cpp warning last night.  I will address that one this morning.

mitza-oci (approved):

PR #3919, opened 2022-12-07
Update XTypes test for dynamic writer and reader

sonndinh:
In the `run_test.pl` script, the options `--dynamic-writers` and `--dynamic-readers` are used to enable the test cases for dynamic data writers and dynamic data readers, respectively. Currently, it seems only one of these two options should be specified at a time, i.e., we are only testing either a dynamic writer communicating with a static reader or a static writer communicating with a dynamic reader.

Each test case in the test script can specify whether it supports the first type of testing by adding a `support_dynamic_writer` parameter to the test case's parameter list and/or the second type of testing by adding a `support_dynamic_reader` parameter. When both `--dynamic-writers` option is turned on and the `support_dynamic_writer` is specified, the publisher in that test case will use a dynamic data writer instead of a static one. That works similarly for testing dynamic data readers. This combination allows us to pick which test cases we want to use to test dynamic writers/readers.

In Publisher.cpp, each write function for a data type is enhanced to allow writing using a `DynamicDataWriter` for that type. This is so that whenever the test script wants to enable testing a dynamic writer for a type, it is already supported by the publisher. Similar changes should be done for Subscriber.cpp following this design changes, but this is currently a TODO.


sonndinh:
Closed as replaced  by #3993

PR #3920, opened 2022-12-07
DynamicDataReaderImpl: work around an error on older MSVC versions

mitza-oci:
Error is `C2487: 'DDS::DynamicDataReader::read_next_sample' : member of dll interface class may not be declared with dll interface`
Newer MSVC doesn't have the same problem, seems to be a compiler bug

### DIFF
--- a/dds/DCPS/XTypes/DynamicTypeSupport.h
+++ b/dds/DCPS/XTypes/DynamicTypeSupport.h
@@ -280,7 +280,14 @@ void DataReaderImpl_T<XTypes::DynamicSample>::dynamic_hook(XTypes::DynamicSample
 
 namespace XTypes {
 
-class OpenDDS_Dcps_Export DynamicDataReaderImpl
+#if defined _MSC_VER && _MSC_VER < 1700
+#define OPENDDS_MAYBE_EXPORT
+#else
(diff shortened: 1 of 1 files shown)


dds/DCPS/XTypes/DynamicTypeSupport.h
```diff
@@ -280,7 +280,14 @@ void DataReaderImpl_T<XTypes::DynamicSample>::dynamic_hook(XTypes::DynamicSample
 
 namespace XTypes {
 
-class OpenDDS_Dcps_Export DynamicDataReaderImpl
+#if defined _MSC_VER && _MSC_VER < 1700
```
ocielliottc, line 283: This is slightly confusing.  Does it not actually need to be exported?

mitza-oci: Based on what we currently have in the repo, we can get along without it.  So that was the other option - just remove the macro on all builds.
In general we'd like to export the these kind of classes so they can be added to unit tests.

ocielliottc: Ah, ok.  That makes sense then.

ocielliottc (approved):

Issue #1395, opened 2019-11-25
cmake OPENDDS_TARGET_SOURCES polluting global includes


joeroback:
`OPENDDS_TARGET_SOURCES` macro seems to pollute the global (directory) includes, moreover, it does not include the headers as "system" headers, so tools like `clang-tidy` include them (`clang-tidy` will ignore system headers with the right option), and my project's `clang-tidy` rules are being applied to the IDL generated headers.

I am not even sure why `opendds_include_dirs_once()` exists in the first place, when all the other options on the target use `target_compile_definitions` and `target_link_libraries`, why not use `target_include_directories`?

mitza-oci:
I don't think I can answer the question "why not use...?" but please go ahead with a pull request to make this change.

mitza-oci:
It looks like there are 2 distinct issues:

1. `opendds_include_dirs_once` doesn't need to be there, the function and the 1 place it's called can just be removed

2. The use "system" include paths could be configured by a CMake cache variable in order to give the end user control of this.  If I'm reading your comment correctly, this refers to header files generated by the IDL compilers and not header files that come with OpenDDS.  If so, I believe this is handled around lines 50-55 of `dds_idl_sources.cmake`.  The new cache variable could be used to determine if an extra `SYSTEM` argument is passed to `target_include_directories`.  One potential drawback is that the use of "system" includes may impact dependency tracking on certain platforms/buildsystems.

@joeroback please confirm if this will indeed fix the issue and let us know if you can create a pull request for one or both

joeroback:
correct for 1, for 2, not sure what to say, if you use tools like clang-tidy + cmake, if the headers are not marked as "system", clang-tidy will run over those headers and of course they do not follow so arbitrary coding standard that the project is using, so clang-tidy hates them. SYSTEM flag does change `-I` to `-isystem` and while that does change the behavior of how the compiler handles them, I have no experienced issues with generated headers being included as `-isystem` ... at least this project is windows / linux x86, and cross-compile linux aarch64 .. but that is not every system out there...

mitza-oci:
OK, I'm fine with an end-user configurable (CMake cache variable) for the SYSTEM flag.
Another approach would be to build a separate library containing just the generated code and don't run clang-tidy on it.

joeroback:
[...]

PR #3921, opened 2022-12-08
Disable warnings about deprecated functions.

ocielliottc:


mitza-oci (approved):
I don't have any issues with the peer review, but there are new test failures in CI that need to be resolved before we can merge.

cmake/OpenDDSConfig.cmake
```diff
@@ -137,6 +137,14 @@ if(OPENDDS_XERCES3)
   endif()
 endif()
 
+## ACE provides (or uses) many deprecated functions.
+## For now, we'll silence the warnings.
+if(MSVC)
+  add_compile_options(/wd4996)
+  add_compile_definitions(_CRT_SECURE_NO_WARNINGS
+                          _WINSOCK_DEPRECATED_NO_WARNINGS)
```
iguessthislldo, line 145: These should probably be attached to the ACE library if that's what they're meant for. This could be moved to the bottom after the ACE library is defined, probably at the end of the `if(OPENDDS_FOUND)` block and changed to something like:

```
target_compile_options(ACE::ACE INTERFACE /wd4996)
target_compile_definitions(ACE::ACE INTERFACE _CRT_SECURE_NO_WARNINGS _WINSOCK_DEPRECATED_NO_WARNINGS)
```

iguessthislldo (changes requested):

cmake/OpenDDSConfig.cmake
ocielliottc: They cannot be attached only to the ACE library.  Much, if not all, of what is warned about are in inline functions.  Otherwise, they wouldn't actually show up when building the cmake tests.

ocielliottc: I think I misunderstood what you were saying.  But, still, I'm not quite sure why it matters.

iguessthislldo: It matters because it's generally preferred to use the target-specific functions and avoid global ones. [We have had to remove a global function for adding includes in the past](https://github.com/objectcomputing/OpenDDS/issues/1395). In this case even if ACE is going to be used by everything we use, the user the might have non-ACE targets and these flags shouldn't be applied to those.

iguessthislldo (approved):

PR #3923, opened 2022-12-12
Added a copy constructor to DynamicDataImpl::SequenceValue

ocielliottc:
With this change, the unit-tests stopped crashing.


ocielliottc:
> Can you please provide a description of why the test crashed without the copy constructor? I'm curious to learn the root cause of this.

The SequenceValue struct has `void* active_` which is used to access the underlying union member.  The default copy constructor copies that value from the right hand side object (as well as the value for the union which is just an unsigned char array).  Writing to those values using the newly constructed SequenceValue will be writing to memory that it does not own.

PR #3925, opened 2022-12-12
Force ubuntu-20.04 For Lint Workflow

simpsont-oci:
Problem: `ubuntu-latest` has been shifted to use `ubuntu-22.04` which appears to fail with `actions-setup-perl` (latest version).

Solution: Fall back to `ubuntu-20.04`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


simpsont-oci:
`ubuntu-22.04` is not the issue. v1.20.0 of `actions-setup-perl` has problems.

PR #3926, opened 2022-12-12
Avoid actions-setup-perl v1.20.0

simpsont-oci:
Problem: `actions-setup-perl` v1.20.0 fails with error about missing "dot" file in runner home directory.

Solution: Falling back to v1.19.0 fixes the problem.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ env:
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


mitza-oci (approved):

simpsont-oci:
> Is there an issue in the Action's repository for this?

It looks like he's already fixed the issue in v1.20.1 (and updated v1 tag to match). I'll make a PR to try revering this now.

PR #3935, opened 2022-12-15
DynamicData Key Comparison

iguessthislldo:


dds/DCPS/XTypes/Utils.h
```diff
@@ -64,6 +64,9 @@ typedef OPENDDS_VECTOR(MemberPath) MemberPathVec;
 OpenDDS_Dcps_Export DDS::ReturnCode_t get_keys(DDS::DynamicType_ptr type, MemberPathVec& paths);
 OpenDDS_Dcps_Export DDS::ReturnCode_t key_count(DDS::DynamicType_ptr type, size_t& count);
 
+OpenDDS_Dcps_Export DDS::ReturnCode_t key_compare(
```
mitza-oci, line 67:
```suggestion
OpenDDS_Dcps_Export DDS::ReturnCode_t key_less_than(
```

If I'm reading this correctly.... assuming there's no error reported in `ReturnCode_t`, then `result` is `true` iff a's key is less than b's key.  Let's call these kind of functions "less than" to make it clear they're not the "compare"/"cmp" kind that return negative, zero, or positive.

iguessthislldo: I actually already thought of this and it's already called `key_less_than` in the cpp file, I just haven't updated the header.

mitza-oci (approved):

PR #3950, opened 2022-12-22
Updated ACE/TAO versions as used by the configure script

jwillemsen:
* README.md:

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,25 +114,25 @@ first -- the `configure` script will download it for you.
 
 There are three distributions of ACE/TAO that can be used with OpenDDS:
 
-* OCI ACE/TAO 2.2a patch 24 or later
+* OCI ACE/TAO 2.2a patch 26 or later
   * This will be automatically downloaded by default when using the configure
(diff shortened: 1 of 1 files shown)


simpsont-oci (approved):

jwillemsen:
Shouldn't this be done as part of the release script?

simpsont-oci:
It should probably be double-checked as part of the release process, but I would think we'd want to update the README in the same PR that updates the configure script so they're always kept in sync.

simpsont-oci:
There was already a list item for updating `README.md` with the latest dependency information in general as part of the release process, but I've updated it with a specific list item to double check that ACE and TAO versions match between `README.md` and the `configure` script.

jwillemsen:
I remembered that in the past I was asked to only update configure, see for example  https://github.com/objectcomputing/OpenDDS/pull/1664

simpsont-oci:
Ah, I see. @mitza-oci did you still want to delay updating `README.md` until release?

PR #3958, opened 2023-01-03
Code has references to Sun and Solaris

jrw972:
Problem
-------

Sun and Solaris are no longer supported.

Solution
--------

Remove references to Sun and Solaris.


dds/DCPS/SafetyProfileSequence.h
```diff
@@ -311,10 +311,6 @@ namespace SafetyProfile {
     // max_size() inherited from AllocPolicy
     bool empty() const { return !length_; }
 
-#ifndef __SUNPRO_CC
```
mitza-oci, line 314: This is `ifndef`, not `ifdef`

mitza-oci (changes requested):

dds/InfoRepo/FederatorManagerImpl.h
```diff
@@ -137,8 +137,6 @@ class OpenDDS_Federator_Export ManagerImpl
   //
   // UpdateProcessor<> methods.
   //
-  // "using" directive to fix "Hides the virtual function in virtual base" warning on
-  // SunOS compiler.
   using UpdateProcessor<OwnerUpdate>::processUpdateQos2;
```
ocielliottc, line 140: Should these be removed if they were put in place for the Sun compiler?

mitza-oci (approved):
Should be ready to go when CI passes.  See https://github.com/objectcomputing/OpenDDS/actions/runs/3832882423/jobs/6523708896

dds/InfoRepo/FederatorManagerImpl.h
iguessthislldo: Looks like Apple clang complains about it as well.

PR #3972, opened 2023-01-10
RepoId is confusing

jrw972:
Problem
-------

Usage of both RepoId and GUID_t causes confusion.

Solution
--------

Remove RepoId.


mitza-oci:
Did you consider changing internal function names like `get_repo_id()`? It could be deferred to another PR.

iguessthislldo:
> Did you consider changing internal function names like get_repo_id()?

That was my intention in the last part of: "... replace references to it with GUID_t, as well as any other appropriate adjustments that might be necessary.".

mitza-oci (approved):
Ready once CI issues are fixed https://github.com/OpenDDS/OpenDDS/actions/runs/3886409181/jobs/6631909838#step:14:3019 and one method is renamed

iguessthislldo (approved):

mitza-oci:
Looks like there are still problems compiling the Wireshark dissector

PR #3985, opened 2023-01-14
Relay STUN task dies during transient configuration

jrw972:
Problem
-------

Users can clear the RtpsRelay address at any time.  If the relay stun task in the RtpsUdpTransport executes during this time period, the task will not reschedule itself.  Setting the address to a new value does not schedule the task.

Solution
--------

Always reschedule the task.


simpsont-oci:
Isn't this what `RtpsUdpTransport::rtps_relay_address_change()` is for? What user is clearing the address? Could they call `rtps_relay_address_change()`?

PR #3988, opened 2023-01-17
Dynamic content subscription

mitza-oci:
Allow DynamicDataReaders to use QueryCondition and ContentFilteredTopic.
Allow DynamicDataWriters to do filtering on behalf of matched DataReaders that use ContentFilteredTopic.

Enum fields won't work in WHERE clause comparisons since the filter language treats them as strings while DynamicData treats them as integers.  A follow-on PR will address this.


iguessthislldo (approved):

dds/DCPS/XTypes/DynamicDataXcdrReadImpl.cpp
```diff
@@ -2833,6 +2834,98 @@ bool DynamicDataXcdrReadImpl::check_xcdr1_mutable_i(DDS::DynamicType_ptr dt, Dyn
...
+  case TK_CHAR8:
+    return get_some_value(value, id, *this, &DynamicDataXcdrReadImpl::get_char8_value);
+  case TK_CHAR16:
+    return get_some_value(value, id, *this, &DynamicDataXcdrReadImpl::get_char16_value);
+  case TK_STRING8:
+    return get_some_value(value, id, *this, &DynamicDataXcdrReadImpl::get_string_value);
+  case TK_STRING16:
+    return get_some_value(value, id, *this, &DynamicDataXcdrReadImpl::get_wstring_value);
+  case TK_ENUM:
+    return get_some_value(value, id, *this, &DynamicDataXcdrReadImpl::get_string_value);
```
sonndinh, line 2922: Should this use `get_int32_value` or equivalent functions instead?

mitza-oci: It's string because the filtering "language" treats enums as strings.
I realize it won't work yet, that will be in another PR.

PR #3989, opened 2023-01-17
`rtps_relay_address_change` deadlocks

jrw972:
Problem
-------

`rtps_relay_address_change` deadlocks if it causes the impl to be created.

Solution
--------

Don't create the impl.

There is no downside here, that is, when the impl is created, it will have the correct configuration.

### DIFF
--- a/dds/DCPS/transport/framework/TransportInst.cpp
+++ b/dds/DCPS/transport/framework/TransportInst.cpp
@@ -149,10 +149,7 @@ OpenDDS::DCPS::TransportImpl_rch
 OpenDDS::DCPS::TransportInst::get_impl()
 {
   ACE_GUARD_RETURN(ACE_SYNCH_MUTEX, g, lock_, TransportImpl_rch());
-  if (!shutting_down_) {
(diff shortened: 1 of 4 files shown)


dds/DCPS/transport/framework/TransportInst.cpp
```diff
@@ -145,6 +145,16 @@ OpenDDS::DCPS::TransportInst::impl()
...
 }
 
+OpenDDS::DCPS::TransportImpl_rch
+OpenDDS::DCPS::TransportInst::get_impl()
+{
+  ACE_GUARD_RETURN(ACE_SYNCH_MUTEX, g, lock_, TransportImpl_rch());
+  if (!shutting_down_) {
+    return impl_;
+  }
+  return TransportImpl_rch();
```
simpsont-oci, line 155:
```suggestion
  return impl_;
```
I don't think it matters if it's shutting down here. We're only check above to make sure we don't create a new one during shutdown.

simpsont-oci (approved):

PR #3993, opened 2023-01-19
Update XTypes Test to Use DynamicData

iguessthislldo:


tests/DCPS/XTypes/Publisher.cpp
```diff
@@ -5,247 +5,229 @@
...
-               OpenDDS::DCPS::retcode_to_string(ret)));
+  CORBA::String_var name = ts->get_type_name();
+  if (!dw) {
+    ACE_ERROR((LM_ERROR, "ERROR: typed DataWriter for %C was null\n", name.in()));
+    return;
   }
+  check_rc(dw->write(sample, HANDLE_NIL), std::string("write for ") + name.in() + " failed");
   if (verbose) {
-    ACE_DEBUG((LM_DEBUG, "writer: ModifiedFinalStruct\n"));
+    ACE_DEBUG((LM_DEBUG, "writer: {}\n", name.in()));
```
mitza-oci, line 23:
```suggestion
    ACE_DEBUG((LM_DEBUG, "writer: %C\n", name.in()));
```

tests/DCPS/XTypes/Subscriber.cpp
```diff
@@ -4,195 +4,205 @@
...
+  {
+#ifdef OPENDDS_SAFETY_PROFILE
+    success = false;
+    return false;
+#else
+    DDS::DynamicDataSeq seq;
+    if (!read_single<DDS::DynamicDataReader, DDS::DynamicDataSeq>(seq)) {
+      return false;
+    }
+    dd = DDS::DynamicData::_duplicate(seq[0]);
```
mitza-oci, line 132: This should also work without the `_duplicate` since the var-to-var assignment rule extends to the unspecified type of sequence elements that's "var-like", but I wouldn't be surprised if that was a gap in the implementation.

mitza-oci (changes requested):

tests/DCPS/XTypes/Subscriber.cpp
iguessthislldo: I don't think I gave any thought when writing it, but removing it causes the subscriber to segfault when trying to read values from the DynamicData, so something's broken. I probably bungled the reference counting somewhere.

iguessthislldo: Okay I've done over my code and it is this assignment because it's not incrementing the reference count without the duplicate.

mitza-oci (approved):

sonndinh (changes requested):

dds/DCPS/XTypes/DynamicDataXcdrReadImpl.cpp
```diff
@@ -678,12 +681,21 @@ bool DynamicDataXcdrReadImpl::read_value(ValueType& value, TypeKind tk)
...
   case TK_STRING16:
-    return strm_ >> value;
+    if (strm_ >> value) {
+      return true;
+    }
+    break;
+  default:
+    if (log_level >= LogLevel::Notice) {
+      ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: DynamicDataXcdrReadImpl::read_value: "
+                 "Calling on an unexpected element type %C\n", typekind_to_string(tk)));
```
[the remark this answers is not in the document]
iguessthislldo: Looks like I was copying the messages from `read_values` and missed that.

dds/DCPS/XTypes/Utils.h
```diff
@@ -11,17 +11,75 @@
...
+  bool consume(size_t by);
+
+  /**
+   * Put the next member name or subscript value in subpath and which kind in
+   * in_subscript. Returns false when the end of the path is reached or there
+   * was an error in which case error is set to true.
+   */
+  bool get_next_subpath(); bool
+
+  get_index(CORBA::UInt32& index);
```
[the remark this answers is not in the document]
iguessthislldo: :confused:  I'm not sure how this happened...

tests/DCPS/XTypes/run_test.pl
```diff
@@ -14,356 +14,343 @@
...
-      r_ini => "rtps_disc_no_xtypes.ini", w_ini => "rtps_disc_no_xtypes.ini",
+      expect_inconsistent_topic => 1, topic => "NoXTypesNoMatchTypeNames_nn",
+      ini => "rtps_disc_no_xtypes.ini",
     },
   );
 }
 
-my $status = 0;
+my %valid_params = ( # 1 = Required, 0 = Optional
+  'topic', => 1, 'reader_type' => 1, 'writer_type' => 1,
```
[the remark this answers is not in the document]
iguessthislldo: Good catch, I'm not sure how that's working perl though. I would've expected it to error on the '=>' following the comma.

sonndinh (approved):

PR #3995, opened 2023-01-20
Resolve Coverity Issues In AddressCache Unit Test

simpsont-oci:
Problem: Coverity doesn't like to see `memcmp` calls that exceed the object bounds of the input pointer types (even if that's the desired / optimized behavior).

Solution: Since it's a unit test, performance isn't critical. Split the `memcmp` calls up in such a way that Coverity can sleep at night.

### DIFF
--- a/tests/unit-tests/dds/DCPS/AddressCache.cpp
+++ b/tests/unit-tests/dds/DCPS/AddressCache.cpp
@@ -15,10 +15,18 @@ struct TestKey {
   TestKey(const GUID_t& from, const GUID_t& to) : from_(from), to_(to) {}
   TestKey(const TestKey& val) : from_(val.from_), to_(val.to_) {}
   bool operator<(const TestKey& rhs) const {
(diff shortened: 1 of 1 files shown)


tests/unit-tests/dds/DCPS/AddressCache.cpp
```diff
@@ -15,10 +15,18 @@ struct TestKey {
   TestKey(const GUID_t& from, const GUID_t& to) : from_(from), to_(to) {}
   TestKey(const TestKey& val) : from_(val.from_), to_(val.to_) {}
   bool operator<(const TestKey& rhs) const {
-    return std::memcmp(static_cast<const void*>(&from_), static_cast<const void*>(&rhs.from_), 2 * sizeof (GUID_t)) < 0;
+    const int from_comp = std::memcmp(static_cast<const void*>(&from_), static_cast<const void*>(&rhs.from_), sizeof (GUID_t));
```
mitza-oci, line 18: This doesn't need casts.  Every pointer converts to `void*`.

mitza-oci (approved):

jrw972 (approved):

PR #3999, opened 2023-01-20
Add RcObject (et al.) Stress Test

simpsont-oci:
### Problem
It appears as though there is a race condition between `WeakRcHandle::lock()` and `RcObject::_remove_ref()`, likely called from `RcHandle::~RcHandle()` or `RcHandle::reset()`.

### Solution
Create a stress test to expose the existing issue, then try to resolve it.

### DIFF
--- a/dds/DCPS/RcObject.h
+++ b/dds/DCPS/RcObject.h
@@ -9,17 +9,12 @@
 #include "dds/Versioned_Namespace.h"
 
 #include "dcps_export.h"
+#include "Atomic.h"
 #include "PoolAllocationBase.h"
 #include "RcHandle_T.h"
 
 #include <ace/Guard_T.h>
 #include <ace/Synch_Traits.h>
-
-#ifdef ACE_HAS_CPP11
-#  include <atomic>
-#else
-#  include <ace/Atomic_Op.h>
-#endif
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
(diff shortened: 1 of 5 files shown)


dds/DCPS/RcObject.h
```diff
@@ -31,10 +31,18 @@ namespace DCPS {
...
   {
   public:
+
+#ifdef ACE_HAS_CPP11
+    typedef std::atomic<long> AtomicCountType;
+#define GET_ATOMIC_COUNT_VALUE(v) v
+#else
+    typedef ACE_Atomic_Op<ACE_SYNCH_MUTEX, long> AtomicCountType;
+#define GET_ATOMIC_COUNT_VALUE(v) v.value()
+#endif
```
mitza-oci, line 41: Is this worth moving to its own header? Could it be useful in other places?

simpsont-oci: Possibly. I see we are using `int`, `long`, `unsigned long`, and `size_t` as atomic counts in other header files. Perhaps we could consolidate to a single header, if we could agree on the underlying integral type/types. Not sure which, if any, really need to be signed, but we could perhaps make one for `size_t` and `ssize_t` to be safe?

simpsont-oci: But that effort feels like a good candidate for a separate task / PR if we decide to go that way.

mitza-oci: It could be a template parameter in that header

mitza-oci: If it's not going to be otherwise changed on this PR, rename the macro to start with OPENDDS.

simpsont-oci: I've attempted to move the typedefs into their own header and opted to simply name them `Atomic`. I think I've removed the need for a macro through inheritance and addition of `operator T()` similar to `std::atomic`. Let me know if that's acceptable. I'm sure there are corner cases I'm not considering here, but for the most part I think it's sufficient for our needs for now. If you're happy with it, I can migrate other atomic counts to use it, or we can save that for a different task / PR.

mitza-oci (approved):

PR #4012, opened 2023-01-31
Factor out LocalEntities to make Sedp.h smaller

jrw972:


mitza-oci (approved):
warnings from CI need to be addressed

PR #4013, opened 2023-01-31
TypeLookup.idl has allocator issues in safety profile

jrw972:
Problem
-------

The TypeLookup.idl file is unconditionally processed by tao_idl which is unaware of the allocator support for safety profile.

Solution
--------

Don't compile TypeLookup.idl when safety profile is not enabled. (It this causes problems, we will compile it with opendds_idl for safety profile.)


dds/DCPS/XTypes/TypeObject.h
```diff
@@ -227,6 +244,16 @@ namespace XTypes {
   typedef ACE_CDR::Octet NameHash_slice;
   typedef Fake_TAO_Array_Forany_T<NameHash, NameHash_slice, NameHash_tag> NameHash_forany;
 
+  inline bool name_hash_equal(const NameHash& x, const NameHash& y)
```
mitza-oci, line 247: Why is a named function used here and not `operator==`?  Is it because this is a typedef so it wouldn't be a distinct type if there was another array of the same bound and element type?

mitza-oci (changes requested):
The equivalent of `ACE_CDR::ULongLong max_alignment;` in TypeIdentifier is needed in AnnotationParameterValue.  In that case, the max_align type should be `LongDouble`.

dds/DCPS/XTypes/TypeObject.h
jrw972:
```/Users/wilsonj/OpenDDS/OpenDDS-java-master/opendds-794/dds/DCPS/XTypes/TypeObject.h:247:15: error: overloaded
      'operator==' must have at least one parameter of class or enumeration type
  inline bool operator==(const NameHash& x, const NameHash& y)
```

mitza-oci: Yeah, that makes sense.  Alternatives include using `IDL::DistinctType` from `Serializer.h` or `std::array` (directly or emulated for older compilers).  But I don't think it needs to change in this PR.

mitza-oci (approved):

PR #4021, opened 2023-02-01
Tab Minutes 2022-12-05

halflauren:
Tab Minutes 2022-12-05

### DIFF
--- a/foundation/minutes/2022-12-05.html
+++ b/foundation/minutes/2022-12-05.html
@@ -1,0 +1,199 @@
+---
+layout: foundation_minutes
+date: "2022-12-05"
+type: "foundation-minutes"
+order: 207
+sitemap: false
+CSS: foundation-minutes.css
+---
+
+<p style="text-align: center;"><span id="docs-internal-guid-165a598f-7fff-fb98-8397-0ad425ba14c2"><span style="border: none; display: inline-block; overflow: hidden; width: 150px; height: 90px;"><img style="margin-left: 0px; margin-top: 0px;" src="https://lh5.googleusercontent.com/A1p6tjgDDSRxs8jpUMxmV8zkEeVQy5zCDolvMc7-tYndL3nnirnWAHNMf7selVpSuNcAhCn0m9kBVY35TR_zGnT0dEs8P5U2VqxReuI9cecMKV5H0jhXctdmnhVR6WRH0fvwQaGvUdSu3Upfzp8y7jI" width="150" height="90" /></span></span></p>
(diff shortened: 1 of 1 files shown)


foundation/minutes/2022-12-05.html
```diff
@@ -0,0 +1,199 @@
+---
+layout: foundation_minutes
+date: "2022-12-05"
+type: "foundation-minutes"
+order: 207
+sitemap: false
+CSS: foundation-minutes.css
+---
+
+<p style="text-align: center;"><span id="docs-internal-guid-165a598f-7fff-fb98-8397-0ad425ba14c2"><span style="border: none; display: inline-block; overflow: hidden; width: 150px; height: 90px;"><img style="margin-left: 0px; margin-top: 0px;" src="https://lh5.googleusercontent.com/A1p6tjgDDSRxs8jpUMxmV8zkEeVQy5zCDolvMc7-tYndL3nnirnWAHNMf7selVpSuNcAhCn0m9kBVY35TR_zGnT0dEs8P5U2VqxReuI9cecMKV5H0jhXctdmnhVR6WRH0fvwQaGvUdSu3Upfzp8y7jI" width="150" height="90" /></span></span></p>
```
iguessthislldo, line 10: I don't think we should be linking to external images like this that might go away. There's already a [Foundation logo in the website](https://github.com/OpenDDS/OpenDDS/blob/gh-pages/foundation/images/opendds-foundation-stacked.png). It's different layout, but if we want to use this one specifically then we should add it as well. 

Also for consistency, If we wanted to make it part of all minutes it should be added to the template for the page instead of the individual minutes pages.

foundation/minutes/2022-12-05.html
```diff
@@ -0,0 +1,199 @@
...
+<td>✓</td>
+</tr>
+<tr>
+<td>Andrew Job</td>
+<td>Founder and CEO, Plotlogic</td>
+<td>X</td>
+</tr>
+<tr>
+<td>Adam Mitz</td>
+<td>OpenDDS Product Lead and DISTINGUISHED ENGINEER, &nbsp; Object Computing</td>
```
iguessthislldo, line 39: Why are some are some OCI titles in all caps, while others are not? This also applies to Marc and Jason. [The last minutes didn't do this](https://github.com/OpenDDS/OpenDDS/blob/bff1e011263c5146d5df4c93c5a3c4517cb68624/foundation/minutes/2022-09-12.html#L50).

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4024, opened 2023-02-03
Use lazy initialization in DynamicDataImpl

sonndinh:


dds/DCPS/XTypes/DynamicDataBase.cpp
```diff
@@ -243,8 +235,6 @@ DDS::ReturnCode_t DynamicDataBase::check_member(
   bool invalid_tk = true;
   if (is_basic(cmp_type_kind)) {
     invalid_tk = cmp_type_kind != tk;
-  } else if (tk == TK_NONE) {
-    invalid_tk = !is_complex(type_kind);
```
jrw972, line 247: Why was this taken out?

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -745,6 +996,14 @@ void verify_string_union(DDS::DynamicType_var dt, const DataView& expected_cdr)
   EXPECT_EQ(ret, DDS::RETCODE_OK);
   ret = data.set_string_value(16, "def");
   EXPECT_EQ(ret, DDS::RETCODE_OK);
+
+  CORBA::Long disc_val;
+  EXPECT_EQ(DDS::RETCODE_OK, data.get_int32_value(disc_val, XTypes::DISCRIMINATOR_ID));
+  EXPECT_EQ(E_STRING8, disc_val);
+  char* str_val;
```
jrw972, line 1003: Use a string manager?

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -761,6 +1020,14 @@ void verify_wstring_union(DDS::DynamicType_var dt, const DataView& expected_cdr)
...
   ret = data.set_wstring_value(17, L"def");
   EXPECT_EQ(ret, DDS::RETCODE_OK);
+
+  CORBA::Long disc_val;
+  EXPECT_EQ(DDS::RETCODE_OK, data.get_int32_value(disc_val, XTypes::DISCRIMINATOR_ID));
+  EXPECT_EQ(E_STRING16, disc_val);
+  CORBA::WChar* wstr_val;
+  EXPECT_EQ(DDS::RETCODE_OK, data.get_wstring_value(wstr_val, 17));
+  EXPECT_STREQ(L"def", wstr_val);
+  CORBA::wstring_free(wstr_val);
```
jrw972, line 1030: Use a string manager?

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -305,23 +443,34 @@ void verify_int32_union(DDS::DynamicType_var dt, const DataView& expected_cdr)
 {
   XTypes::DynamicDataImpl data(dt);
   DDS::ReturnCode_t ret = data.set_int32_value(XTypes::DISCRIMINATOR_ID, E_INT32);
-  EXPECT_EQ(ret, DDS::RETCODE_OK);
+  EXPECT_EQ(ret, DDS::RETCODE_ERROR);
```
jrw972, line 446: I assume that either the test was wrong before or the new implementation breaks it.  Can you specify which one and why?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -1065,53 +1467,53 @@ bool DynamicDataImpl::set_value_to_union(DDS::MemberId id, const MemberType& val
...
-    if (!insert_valid_discriminator(md)) {
+      return insert_single(id, value);
+    } else {
+      // In case the union has implicit default member and the input discriminator value
+      // selects that implicit default member, store the discriminator value. The semantics
+      // of this is similar to the _default() method of the IDL-to-C++ mapping for union.
+      if (discriminator_selects_no_member(type_, disc_value)) {
+        return insert_single(id, value);
+      }
+      if (DCPS::log_level >= DCPS::LogLevel::Notice) {
```
jrw972, line 1486: Upgrade this to an error.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -1715,42 +2108,45 @@ bool DynamicDataImpl::set_complex_to_union(DDS::MemberId id, DDS::DynamicData_va
       if (selected_member->get_descriptor(selected_md) != DDS::RETCODE_OK) {
         return false;
       }
-      CORBA::Long disc_val;
-      const DynamicDataImpl* dd_impl = dynamic_cast<const DynamicDataImpl*>(value.in());
-      if (!dd_impl || !dd_impl->read_discriminator(disc_val)) {
+      if (!validate_discriminator(disc_val, selected_md)) {
+        if (DCPS::log_level >= DCPS::LogLevel::Notice) {
+          ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: DynamicDataImpl::set_complex_to_union:"
```
jrw972, line 2113: NOTICEs should go to ACE_DEBUG.

jrw972 (changes requested):

dds/DCPS/XTypes/DynamicDataBase.cpp
sonndinh: There were some uses of this function that relied on `tk` taking a default parameter value `TK_NONE`. But those uses are removed and we don't need this else block anymore.

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: Yes, the new implementation breaks it. Before we allowed setting the discriminator directly and that can leave the container of a union in a state where it has the discriminator but not the member it selects. I updated how we handle union to follow the IDL-to-C++ mapping more strictly, which doesn't let it set the discriminator in this case. The state of the container for a union now maintains an invariant that it either is empty or contains valid data for the union, meaning it has the discriminator and an activated member if that discriminator selects it. I think it'll be good to update the DevGuide for the behavior of union in DynamicData.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: I checked the guidelines and it has ACE_ERROR for NOTICE.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: The table in the guidelines has `LogLevel::Error` for unrecoverable error but this one looks recoverable to me. It will also return `DDS::RETCODE_ERROR` to the user, so maybe it's also reportable even though `DDS::RETCODE_ERROR` is not specific enough. But I'm not sure if it's really considered reportable in this case. If it's not, perhaps move it to `LogLevel::Warning`?

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: I'm not very familiar with string manager. Is that TAO's string manager?

dds/DCPS/XTypes/DynamicDataImpl.cpp
iguessthislldo: If it's a `ReturnCode_t` the user gets (or ultimately becomes one), it should be almost certainly be logged using notice. There could be an exception for code that's shared with something that's not returned to the user, but in that case it's probably more proper to have a separate warning/error at non-user caller.

dds/DCPS/XTypes/DynamicDataImpl.cpp
iguessthislldo: Yes, I think we were (or at least I was) kinda uncertain on the dividing line for ACE_ERROR/ACE_DEBUG. I think the rational for guidelines is that the difference only matters for ACE's error status, which we don't use and the warnings and notices are basically less serious errors.

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -714,6 +958,13 @@ void verify_bool_union(DDS::DynamicType_var dt, const DataView& expected_cdr)
   EXPECT_EQ(ret, DDS::RETCODE_OK);
   ret = data.set_boolean_value(15, CORBA::Boolean(false));
   EXPECT_EQ(ret, DDS::RETCODE_OK);
+
+  CORBA::Long disc_val;
+  EXPECT_EQ(DDS::RETCODE_OK, data.get_int32_value(disc_val, XTypes::DISCRIMINATOR_ID));
+  EXPECT_EQ(E_BOOL, disc_val);
+  CORBA::Boolean bool_val;
+  EXPECT_EQ(DDS::RETCODE_OK, data.get_boolean_value(bool_val, 15));
+  EXPECT_EQ(CORBA::Boolean(false), bool_val);
```
iguessthislldo, line 967: I'm not entirely sure what this C++03 error is about, but you could just change this to `EXPECT_FALSE(bool_val)` since `CORBA::Boolean` is always going to be the same thing as `bool`.

iguessthislldo (changes requested):

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
iguessthislldo: Yes, it's what the actual type of strings are in the TAO IDL mapping. I think they have implicit conversion operators for `char*&` and `const char*`for use in operation arguments, so that's what DynamicData uses.

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
mitza-oci: Is the empty state visible to the user?  If the user creates a dynamic union and then writes the sample is it an error or does it write some valid value?  (This is one place where the "C++ classic" and C++11 mappings differ, so there's no single right way to do it.)

dds/DCPS/XTypes/DynamicDataBase.cpp
sonndinh: In the master branch, `check_member` is called in `get_complex_value` to check for non-complex types. But that is not necessary since we can `get_complex_value` for any member type and get a `DynamicData` for that member.

tests/unit-tests/dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: It'll write a valid value constructed from the default value for the discriminator and the default value for a selected member if one is selected.

The empty state is probably not visible to the user since we're returning a default value when the user tries to get a member from an empty union.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: Sounds like using notice is what we want here since the user will receive `ReturnCode_t` in this case.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: I would say we keep NOTICEs with ACE_ERROR if ACE_ERROR and ACE_DEBUG are not much different for NOTICEs. I think we have been using NOTICEs with ACE_ERROR pretty consistently in this file and other files. If we update the guidelines, we can update them all at once.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -178,214 +177,422 @@ DDS::MemberId DynamicDataImpl::get_member_id_at_index(ACE_CDR::ULong index)
...
+      return 0;
+    }
+    return bound_total(descriptor);
+  }
+  case TK_STRUCTURE: {
+    const CORBA::ULong member_count = type_->get_member_count();
+    CORBA::ULong count = member_count;
+    // An optional member that hasn't been set is considered missing.
+    // All non-optional members are counted since they either are set directly
+    // or hold default values.
```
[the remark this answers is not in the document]
sonndinh: I'm quite sure I follow the part for testing if an optional member is set or not, but there is a check that if an optional member is not present in the container, then that means it hasn't been set by the user yet, and so is missing.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -1065,55 +1467,55 @@ bool DynamicDataImpl::set_value_to_union(DDS::MemberId id, const MemberType& val
...
-  if (memberSelected->is_default_label()) {
+  DDS::DynamicTypeMember_var member;
+  if (type_->get_member(member, id) != DDS::RETCODE_OK) {
+    return false;
+  }
+  DDS::MemberDescriptor_var md;
+  if (member->get_descriptor(md) != DDS::RETCODE_OK) {
+    return false;
+  }
+  member_type = get_base_type(md->type());
```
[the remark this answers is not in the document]
sonndinh: I tried replacing this but then realized I need `md` later in the same function.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -2147,500 +2532,1431 @@ DDS::ReturnCode_t DynamicDataImpl::set_wstring_values(DDS::MemberId id, const DD
...
+bool DynamicDataImpl::read_basic_value(CORBA::WChar*& value) const
 {
-  if (enum_string_helper(value, id)) {
-    return DDS::RETCODE_OK;
-  }
-
-  const char* orig = 0;
-  const DDS::ReturnCode_t rc = get_single_value(orig, id, TK_STRING8);
-  if (rc == DDS::RETCODE_OK) {
-    CORBA::string_free(value);
```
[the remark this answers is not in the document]
sonndinh: You're right, I need to update this.

dds/DCPS/XTypes/DynamicDataImpl.cpp
jrw972: You are correct on this.

dds/DCPS/XTypes/DynamicDataImpl.cpp
jrw972: You are correct on this.

dds/DCPS/XTypes/DynamicDataImpl.cpp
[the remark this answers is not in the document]
iguessthislldo: I was meaning from the user perspective. For example if you got a DynamicData from a DynamicDataReader and an optional member was null. I'm wondering how to test for that because that could be significant. Maybe returning `RETCODE_NO_DATA` for those cases?

iguessthislldo (approved):

PR #4028, opened 2023-02-04
Fix Random LargeSample Shmem Failures

iguessthislldo:
An initial message exchange [was previously added](https://github.com/OpenDDS/OpenDDS/pull/3549) to shared memory association that went from the reader to the writer and back to indicate to the writer that it could start. After this was added occasionally the LargeSample shared memory test would fail.

The main cause of this was a broken less than operator for a map tracking writers that readers should resend association messages to. Another cause that probably occurred a smaller part of the time was an unanticipated call path on the writer side that sent an acknowledgement message to the reader that messed up the exchange. The fix for this was to adjust how the writer association messages were sent and how data links were created.

Also:
- Some refactoring in Shmem, especially in `ShmemTransport` to cleanup shared code between `accept_datalink` and `connect_datalink`.
- Fix a data race TSAN identified in the LargeSample test

### DIFF
--- a/dds/DCPS/transport/shmem/ShmemDataLink.h
+++ b/dds/DCPS/transport/shmem/ShmemDataLink.h
@@ -19,6 +19,7 @@
 #include <dds/DCPS/transport/framework/DataLink.h>
 
 #include <string>
+#include <map>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
@@ -28,19 +29,16 @@ namespace DCPS {
 class ReceivedDataSample;
 
 struct ShmemData {
-  int status_;
+  enum Status {
+    Free = 0,
+    InUse = 1,
+    RecvDone = 2,
+    EndOfAlloc = -1
+  };
+
+  Status status_;
   char transport_header_[TRANSPORT_HDR_SERIALIZED_SZ];
   ACE_Based_Pointer_Basic<char> payload_;
-};
-
-/**
- * values for ShmemData::status_
- */
-enum {
-  SHMEM_DATA_FREE = 0,
-  SHMEM_DATA_IN_USE = 1,
-  SHMEM_DATA_RECV_DONE = 2,
-  SHMEM_DATA_END_OF_ALLOC = -1
 };
 
 class OpenDDS_Shmem_Export ShmemDataLink
@@ -50,9 +48,6 @@ public:
   ShmemDataLink(const ShmemTransport_rch& transport);
 
   bool open(const std::string& peer_address);
-
-  int make_reservation(const GUID_t& remote_sub, const GUID_t& local_pub,
-    const TransportSendListener_wrch& send_listener, bool reliable);
 
   int make_reservation(const GUID_t& remote_pub,
                        const GUID_t& local_sub,
@@ -102,7 +97,8 @@ private:
 
     bool operator<(const GuidPair& other) const
     {
-      return GUID_tKeyLessThan()(local, other.local) && GUID_tKeyLessThan()(remote, other.remote);
+      return GUID_tKeyLessThan()(local, other.local) ||
+        (local == other.local && GUID_tKeyLessThan()(remote, other.remote));
     }
   };
   ACE_Thread_Mutex assoc_resends_mutex_;
--- a/dds/DCPS/transport/framework/TransportClient.cpp
+++ b/dds/DCPS/transport/framework/TransportClient.cpp
@@ -1005,7 +1005,7 @@ TransportClient::send_i(SendStateDataSampleList send_list, ACE_UINT64 transactio
         VDBG_LVL((LM_DEBUG,"(%P|%t) DBG: Found DataLinkSet. Sending element %@.\n"
                   , cur), 5);
 
-  #ifndef OPENDDS_NO_CONTENT_SUBSCRIPTION_PROFILE
+#ifndef OPENDDS_NO_CONTENT_SUBSCRIPTION_PROFILE
 
         // Content-Filtering adjustment to the pub_links:
         // - If the sample should be filtered out of all subscriptions on a given
@@ -1057,7 +1057,7 @@ TransportClient::send_i(SendStateDataSampleList send_list, ACE_UINT64 transactio
           pub_links = subset;
         }
 
-  #endif /* OPENDDS_NO_CONTENT_SUBSCRIPTION_PROFILE */
+#endif
 
         // This will do several things, including adding to the membership
         // of the send_links set.  Any DataLinks added to the send_links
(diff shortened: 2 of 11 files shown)


dds/DCPS/GuidUtils.h
```diff
@@ -274,6 +272,28 @@ GUID_t bit_key_to_guid(const DDS::BuiltinTopicKey_t& key)
   return id;
 }
 
+struct OpenDDS_Dcps_Export GuidPair {
```
jrw972, line 275: Unit tests.

jrw972 (changes requested):

dds/DCPS/transport/shmem/ShmemDataLink.h
```diff
@@ -28,21 +29,18 @@ namespace DCPS {
...
 struct ShmemData {
-  int status_;
+  enum Status {
+    Free = 0,
+    InUse = 1,
+    RecvDone = 2,
+    EndOfAlloc = -1
+  };
+
+  Status status_;
```
mitza-oci, line 39:
```suggestion
  int status_;
```
This was chosen to make the in-memory representation independent of compiler/implementation decisions.
We can't guarantee that two processes running code built with different compilers can communicate over shmem, but we'll try to support it when possible.

mitza-oci (changes requested):

dds/DCPS/transport/shmem/ShmemDataLink.h
iguessthislldo: If that's the intention, wouldn't it be better if this was using something like `ACE_UINT8`?

dds/DCPS/GuidUtils.h
```diff
@@ -27,98 +25,104 @@ namespace OpenDDS {
...
-const EntityId_t ENTITYID_TL_SVC_REPLY_READER = { {0x00,0x03,0x01}, 0xc4};
+const EntityId_t ENTITYID_TL_SVC_REQ_WRITER = {{0x00, 0x03, 0x00}, 0xc3};
+const EntityId_t ENTITYID_TL_SVC_REQ_READER = {{0x00, 0x03, 0x00}, 0xc4};
+const EntityId_t ENTITYID_TL_SVC_REPLY_WRITER = {{0x00, 0x03, 0x01}, 0xc3};
+const EntityId_t ENTITYID_TL_SVC_REPLY_READER = {{0x00, 0x03, 0x01}, 0xc4};
 ///@}
 
 /// Nil value for GUID.
-const GUID_t GUID_UNKNOWN = { { 0 }, { { 0 }, 0 } };
+const GUID_t GUID_UNKNOWN = { {0}, {{0}, 0} };
```
simpsont-oci, line 58: Based on the latest discussion in from the dev guidelines PR, this just needs to pick one style or the other and go with it.

dds/DCPS/GuidUtils.h
```diff
@@ -274,6 +272,28 @@ GUID_t bit_key_to_guid(const DDS::BuiltinTopicKey_t& key)
...
+  GUID_t local;
+  GUID_t remote;
+
+  GuidPair(const GUID_t& local, const GUID_t& remote)
+  : local(local)
+  , remote(remote)
+  {
+  }
+
+  int cmp(const GuidPair& other) const
```
simpsont-oci, line 285: I think my biggest concern here is that we'll eventually want to optimize comparisons on some of these, and so a packed data structure makes sense (so we can use a single call to `std::memcmp`). But maybe that just needs to be a separate data structure if we ever get back to that point.

dds/DCPS/GuidUtils.h
iguessthislldo: This was based on my suggestion in that PR, but sure.

dds/DCPS/GuidUtils.h
mitza-oci: Does inline end up producing the same code?  godbolt.org may be helpful with this.

dds/DCPS/GuidUtils.h
mitza-oci: PR 4034 has `const GUID_t GUID_UNKNOWN = { { 0 }, { { 0 }, 0 } };`

dds/DCPS/RTPS/Sedp.h
```diff
@@ -1074,31 +1074,38 @@ class Sedp : public virtual DCPS::RcEventHandler {
     MonotonicTimePoint time_added_to_map;
   };
 
-  struct MatchingPair {
-    MatchingPair(GUID_t writer, GUID_t reader, DCPS::TypeObjReqCond* non_match_cond = 0)
-      : writer_(writer)
-      , reader_(reader)
-      , non_match_cond_(non_match_cond)
+  struct MatchingPair : public DCPS::GuidPair {
```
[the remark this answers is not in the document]
iguessthislldo: I generally think aggregation should be preferred when possible, but in a small class/struct like this I'm not sure it matter much. The only serious pro/con I can think of is aggregation would force access to the guids using the equivalent of something like `matching_pair.pair.local`, which I don't think is very elegant.

dds/DCPS/GuidUtils.h
iguessthislldo: > Does inline end up producing the same code? godbolt.org may be helpful with this.

- Single `memcmp`: https://godbolt.org/z/Mq35aMW7K
- 1 or 2 `memcmp`: https://godbolt.org/z/Y31EeqGrn

At `-O1` and higher, they look like they inline and optimize about as much as they can, but they don't take the leap that Tim wants. GCC and MSVC look the basically the same, but MSVC still includes the unused template instance definitions for some reason, even at higher `-O` levels. I haven't gone through Clang line by line to completely understand what it's doing, but it's interesting because it looks like it's inlining an unrolled implementation of `memcmp`.

If we can get away with this for now I'd prefer it so we're not prematurely optimizing. If it turns out this is a problem, we can change it later or bring back something like the packed `RoutedGuidPair`. As it is this PR should increase performance because, excluding the broken shmem `GuidPair` comparison, the other two `GuidPair`s were using less than calls.

dds/DCPS/RTPS/Sedp.h
```diff
@@ -1074,31 +1074,38 @@ class Sedp : public virtual DCPS::RcEventHandler {
...
+    bool local_is_reader;
+    DCPS::TypeObjReqCond* non_match_cond;
```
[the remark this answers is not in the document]
mitza-oci: Assuming it needs to be copyable, and is actually copied, is there a problem with the `non_match_cond` member?

dds/DCPS/GuidUtils.h
simpsont-oci: OK, let's leave it like this for now but make a mental note to revisit if we see these being used as map keys in performance-critical maps, etc.

simpsont-oci (approved):

mitza-oci (approved):

jrw972 (approved):

PR #4034, opened 2023-02-06
Update Dev Guidelines For Whitespace Around Braces

simpsont-oci:
Update the dev guidelines to reflect a consistent approach to whitespace surrounding braces.

### DIFF
--- a/docs/internal/dev_guidelines.rst
+++ b/docs/internal/dev_guidelines.rst
@@ -161,11 +161,14 @@ Whitespace
 * Namespace scopes that span most or all of a file do not cause indentation of their contents.
 * Otherwise lines ending in ``{`` indicate that subsequent lines should be indented one more level until ``}``.
 * Continuation lines (when a statement spans more than one line) can either be indented one more level, or indented to nest "under" an ``(`` or similar punctuation.
-* Add space around binary operators and after commas: ``a + b``
+* Add space around binary operators and after commas: ``a + b, c``
 * Do not add space around parentheses for function calls, a properly formatted function call looks like ``func(arg1, arg2, arg3);``
 * Do not add space around brackets for indexing, instead it should look like: ``mymap[key]``
-* In general, do not add space :) Do not add extra spaces to make syntax elements (that span lines/statements) line up.
-  This only causes unnecessary changes in adjacent lines as the code evolves.
+* For code that includes multiple braces appearing together in the same expression (such as initializer lists), there are two approved styles:
+  * spaces between braces and their enclosed (non-empty) sub-expression: ``const GUID_t GUID_UNKNOWN = { { 0 }, { { 0 }, 0 } };`` or ``{ a + b, {} }``
+  * no such spaces: ``const GUID_t GUID_UNKNOWN = {{0}, {{0}, 0}};`` or ``{a + b, {}}``
+* Do not add extra spaces to make syntax elements (that span lines/statements) line up; this only causes unnecessary changes in adjacent lines as the code evolves.
+* In general, do not add extra spaces unless doing so is covered by the rules above.
 
 Language Usage
 ==============
--- a/docs/internal/docs.rst
+++ b/docs/internal/docs.rst
@@ -88,7 +88,7 @@ RST/Sphinx Usage
 * `One sentence per line should be perfered. <https://rhodesmill.org/brandon/2012/one-sentence-per-line/>`__
   This makes it easier to see what changed in a ``git diff`` or GitHub PR and easier to move sentences around in editors like Vim.
   It also avoids inconsistencies involving what the maximum line length is.
-  This might make it more annoying to read the documentation raw, but that's not the indented way to do so anyway.
+  This might make it more annoying to read the documentation raw, but that's not the intended way to do so anyway.
 
 GitHub Links
 ============


docs/internal/dev_guidelines.rst
```diff
@@ -161,11 +161,11 @@ Whitespace
 * Namespace scopes that span most or all of a file do not cause indentation of their contents.
 * Otherwise lines ending in ``{`` indicate that subsequent lines should be indented one more level until ``}``.
 * Continuation lines (when a statement spans more than one line) can either be indented one more level, or indented to nest "under" an ``(`` or similar punctuation.
-* Add space around binary operators and after commas: ``a + b``
+* Add space around binary operators, between identical braces, and after commas: ``{ {}, {a + b} }``
```
iguessthislldo, line 164: I'm not sure it's necessarily saner to require the spacing. In array initialization, if it starts with a struct/arrayand ends in a scalar like in `EntityId_t` than it looks weird to have it on one side and not the other `{ {0}, 0}`. Also why curly braces and not the other kinds of braces/brackets?

docs/internal/dev_guidelines.rst
```diff
@@ -161,11 +161,11 @@ Whitespace
...
 * Do not add space around parentheses for function calls, a properly formatted function call looks like ``func(arg1, arg2, arg3);``
 * Do not add space around brackets for indexing, instead it should look like: ``mymap[key]``
-* In general, do not add space :) Do not add extra spaces to make syntax elements (that span lines/statements) line up.
-  This only causes unnecessary changes in adjacent lines as the code evolves.
+* Do not add extra spaces to make syntax elements (that span lines/statements) line up. This only causes unnecessary changes in adjacent lines as the code evolves.
```
iguessthislldo, line 167:
```suggestion
* Do not add extra spaces to make syntax elements (that span lines/statements) line up.
  This only causes unnecessary changes in adjacent lines as the code evolves.
```

This was right, there should only be one sentence per line.

iguessthislldo (changes requested):

docs/internal/dev_guidelines.rst
simpsont-oci: Do we have guidelines for the formatting of the guidelines?

iguessthislldo: Yes: https://opendds.readthedocs.io/en/master/internal/docs.html#rst-sphinx-usage

docs/internal/dev_guidelines.rst
simpsont-oci: It does look a little weird. I'd prefer a space at the end as well (really, between all curly braces), but I was going for the minimum change to address unreadable nonsense like `{0, {{0}, 0}}`. If we want to add a separate line item specifically for initializer lists I'm fine with that.

docs/internal/dev_guidelines.rst
simpsont-oci: Ironically, viewing them raw is the only way I tend to view them... but fine. We'll fix the typo and make it a single sentence.

docs/internal/dev_guidelines.rst
mitza-oci: Please add a link to the code that's motivating this change to the guidelines.

simpsont-oci: Even if we want to adopt Google's formatting guideline for initializer lists (their spacing should look just like function calls), I would like to see a separate line item in the guidelines for it. It's also worth noting that much of our generated code does not follow the guidelines for whitespace in this regard (no spaces after commas, etc).

iguessthislldo: >  I'd prefer a space at the end as well 

I don't prefer either. I don't think this is a problem that warrants any change to the dev guidelines.

simpsont-oci: > I don't think this is a problem that warrants any change to the dev guidelines.

The fact that we're disagreeing about what is implicitly allowed under the existing guidelines regarding intializer lists is strong evidence that we need the guidelines to be more explicit, one way or the other.

mitza-oci: There's always going to be some subjectivity to guidelines -- both in the contents of the guidelines and in how detailed/explicit they need to be.  The phrase "unreadable nonsense" is clearly one opinion but that's not to say it's invalid.  There's a lot of opinions codified in the style guidelines.  Their purpose is, at least in part, to provide some consistency to the project even though many developers have contributed over a long time.

I put `const GUID_t GUID_UNKNOWN = {{0}, {{0}, 0}};` into an online code formatter and it didn't add spaces when using a few common styles like "Google," "Chromium," and "LLVM."  It did add space using "Mozilla" settings.  This is not a scientific survey.

Practically speaking, I think (whether we've written it in or not), the guidelines have some leeway for developers to format certain code patterns at their discretion, especially uncommon ones that probably weren't considered when the guidelines were written.  This goes to what @iguessthislldo mentioned about not needing specific guidelines for this.

Maybe then the best way forward is to put something in the guidelines doc that describes the "leeway" idea -- I looked but I haven't found something similar yet.

Let's leave the issue of generated code aside for now, however we could consider adding a note to the guidelines that while we prefer generated code to follow the style there's no requirement for it to.  Nobody's going to be maintaining the code, but it does help for it to be readable/consistent.

docs/internal/docs.rst
```diff
@@ -88,7 +88,7 @@ RST/Sphinx Usage
 * `One sentence per line should be perfered. <https://rhodesmill.org/brandon/2012/one-sentence-per-line/>`__
   This makes it easier to see what changed in a ``git diff`` or GitHub PR and easier to move sentences around in editors like Vim.
   It also avoids inconsistencies involving what the maximum line length is.
-  This might make it more annoying to read the documentation raw, but that's not the indented way to do so anyway.
+  This might make it more annoying to read the documentation raw, but that's not the intended way to do so anyway.
```
mitza-oci, line 91: Just as one more data point -- I think I've only ever read it as a plaintext file, not rendered.

docs/internal/dev_guidelines.rst
mitza-oci: A more specific recommendation, if we want to see more detail in the guidelines:

There are two approved styles for code that includes multiple braces appearing together in the same expression (such as initializer lists): `const GUID_t GUID_UNKNOWN = { { 0 }, { { 0 }, 0 } };` and `{{0}, {{0}, 0}}`

iguessthislldo: I think I agree with Adam. Also I was a probably little blunt/hasty saying there shouldn't be a change. I'm mostly against a hard rule on this that'll force me to put spaces in when I really don't think it's necessary.

If we did want rule for this I think I'd be fine with a rule that says to prefer no spaces on trivial cases, spaces on more nested cases, but it's up the author's and reviewer's discretion. For my changes specifically, I think the case of `{{X}, X}` is trivial enough not to need spaces, but I'm okay with changing `GUID_UNKOWN` to `{ {0}, {{0}, 0} }`. Also longer, more complicated initializers, like for an exact GUID for example, could have "all the spaces" like Son and Tim seem to prefer if it makes it easier to read.

docs/internal/docs.rst
iguessthislldo: Yes, and that's fine. The intention in "intended" is what's going to be most friendly for end-users.

mitza-oci (approved):

iguessthislldo (approved):

PR #4035, opened 2023-02-07
Allow Reassembly of Overlapping Fragment Ranges

simpsont-oci:
### Problem
For the sake of robust handling of input data and maximum interoperability, the RTPS fragment reassembly algorithm (implemented by `TransportReassembly`) ought to be able to handle `DATA_FRAG` submessages that contain variable fragment ranges, including ones that overlap with previously sent fragment ranges. Currently, receiving such fragment ranges can cause the data to be permanently undeliverable, since it violates the assumptions of TransportReassembly and will be incorrectly inserted / tracked in the list of received fragments.

### Solution
Allow for overlapping fragment ranges by allowing incoming fragment samples to be "split" and (recursively) reinserted into the list of received samples. Reduce insertion complexity by removing combined samples and reinserting when certain complex boundaries change.


dds/DCPS/DataSampleHeader.h
```diff
@@ -172,6 +172,9 @@ struct OpenDDS_Dcps_Export DataSampleHeader : public PoolAllocationBase {
   ACE_UINT32 lifespan_duration_nanosec_;  // Corresponding IDL is unsigned.
   //@}
 
+  /// Fragment size used by this sample
+  ACE_UINT32 fragment_size_;
```
mitza-oci, line 176: What's the equivalent of this in the code as it is today? Is there another place this could be stored so we don't have to change DataSampleHeader?

simpsont-oci: I don't think there's an equivalent today, because the reader side doesn't have a need to know the fragment size. In general. It only worries about the fragment range associated with a buffer, and then tries to stitch the message back together based on those ranges. But now that we potentially need to split the buffer based on fragment, we need to know the fragment size.

If you don't want to put the value in this header, perhaps we could add another field to received data sample?

mitza-oci: The received data sample was one option I was considering.  Another would be to have the non-rtps transports set this value to 0 which would indicate to TransportReassembly that it should do the previous behavior

dds/DCPS/MessageBlock.h
```diff
@@ -0,0 +1,49 @@
...
+#include <dds/Versioned_Namespace.h>
+
+#include <ace/Message_Block.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+struct OpenDDS_Dcps_Export MessageBlock {
```
mitza-oci, line 19: This used to be nested inside another class, so there wasn't a need to make data private.  Now it should be an actual `class` with private member data.

dds/DCPS/transport/framework/ReceivedDataSample.cpp
```diff
@@ -145,28 +145,38 @@ void ReceivedDataSample::replace(const char* data, size_t size)
...
-  ACE_NEW_MALLOC(data_,
-    (ACE_Data_Block*) alloc->malloc(sizeof(ACE_Data_Block)),
-    ACE_Data_Block(size, ACE_Message_Block::MB_DATA, 0, alloc, 0,
-                   0, alloc));
-}
+  ReceivedDataSample result;
+  result.header_ = header_;
+
+  const ACE_UINT16 fsize = header_.fragment_size_;
+  const size_t start_offset = start_frag * fsize;
```
mitza-oci, line 155: CI warnings here

dds/DCPS/transport/framework/ReceivedDataSample.h
```diff
@@ -87,82 +87,16 @@ class OpenDDS_Dcps_Export ReceivedDataSample {
...
-  struct MessageBlock {
-    /// construct a MessageBlock that references the existing amb's ACE_Data_Block
-    explicit MessageBlock(const ACE_Message_Block& amb);
-
-    /// construct a MessageBlock with 'size' bytes allocated but not used (wr_ptr_ is 0)
-    explicit MessageBlock(size_t size);
-
-    /// construct a MessageBlock that points to external data, doesn't allocate or copy
-    MessageBlock(const char* data, size_t size);
+  typedef SequenceNumber::Value FragmentNumber;
```
mitza-oci, line 90: This is defined in two places which may be confusing or hard to maintain

dds/DCPS/transport/framework/ReceivedDataSample.h
```diff
@@ -87,82 +87,16 @@ class OpenDDS_Dcps_Export ReceivedDataSample {
...
+  static const FragmentNumber INVALID_FRAGMENT= -1;
```
mitza-oci, line 91:
```suggestion
  static const FragmentNumber INVALID_FRAGMENT = -1;
```

mitza-oci (changes requested):

mitza-oci:
The only concern I have at this point is https://github.com/OpenDDS/OpenDDS/pull/4035#discussion_r1102182372
I can look for some potential alternatives

dds/DCPS/DataSampleHeader.h
simpsont-oci: I can make the two "split" cases in the reassembly logic dependent on a non-zero fragment size as well, but currently the only place where fragment ranges could overlap is from a non-OpenDDS implementation of RTPS. OpenDDS transports, in general, won't ever hit those two cases in the reassembly logic. So the check would be slightly redundant.

simpsont-oci: Oh, this should have been removed. The fragment size is in ReceivedDataSample now.

simpsont-oci:
> The only concern I have at this point is [#4035 (comment)](https://github.com/OpenDDS/OpenDDS/pull/4035#discussion_r1102182372) I can look for some potential alternatives

Yeah, weird, I could have sworn I had moved this already. Should be out of DataSampleHeader now.

mitza-oci (approved):

PR #4036, opened 2023-02-08
Remove bound_max for now

iguessthislldo:


sonndinh (approved):

sonndinh:
I saw a CI build failed because of a call to `bitmask_bound` in `DynamicDataBase` that needs updates.

iguessthislldo:
Thanks, should've built it to check for that... now that I'm looking there was also a call in `DynamicDataImpl` and I also noticed the calling function in `DynamicDataBase` had the wrong return type.

sonndinh:
> Thanks, should've built it to check for that... now that I'm looking there was also a call in `DynamicDataImpl` and I also noticed the calling function in `DynamicDataBase` had the wrong return type.

Yeah, the `check_member` function. I noticed that too while working on #4024. Good to have it fixed.

sonndinh (approved):

PR #4037, opened 2023-02-09
[Bench] Improve Worker ForwardAction Queue Locking Performance

simpsont-oci:
### Problem
ForwardAction writes currently taking place under lock, preventing reads from updating queue and returning (potentially blocking transport reactor threads).

### Solution
Copy and/or swap data out of the queue before writing in order to be able to release action's lock.


iguessthislldo (approved):

simpsont-oci:
Still investigating the out of order deliver errors... there's probably still a bug in the ForwardAction queue logic.

PR #4038, opened 2023-02-09
Fixed a bug in content filtering with enum comparisons on serialized samples 

mitza-oci:
The code that’s used to evaluate content filters on serialized data samples (only used for a durable writer’s cached data when replaying to a filtered reader) incorrectly handled enum members.  If the sample has an enum member that’s used by the filter expression, the evaluator was treating it as an integer instead of a string.


iguessthislldo (approved):

sonndinh (commented):
There are some checks failed from the MetaStruct test. Looks like it read a string type but expected a different type.

PR #4041, opened 2023-02-14
Secure writer and reader in same participant do not associate

jrw972:
Problem
-------

A writer and reader in the same participant that uses non-trivial security fail to associate and exchange data.  The SingleParticipantWithSecurity tests this scenario but uses an "open" domain so all of the encryption and decryption logic was skipped.

Solution
--------

* Turn on encryption so that those code paths are exercised in the test.
* Fix code paths that cannot handle a remote writer being local.

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
@@ -412,9 +412,11 @@ RtpsUdpReceiveStrategy::receive_bytes(iovec iov[],
     static const int GuidPrefixOffset = 8; // "RTPS", Version(2), Vendor(2)
     std::memcpy(peer.guidPrefix, encBuf + GuidPrefixOffset, sizeof peer.guidPrefix);
     peer.entityId = RTPS::ENTITYID_PARTICIPANT;
-    const ParticipantCryptoHandle sender = link_->handle_registry()->get_remote_participant_crypto_handle(peer);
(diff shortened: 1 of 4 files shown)


iguessthislldo (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -412,9 +412,11 @@ RtpsUdpReceiveStrategy::receive_bytes(iovec iov[],
     static const int GuidPrefixOffset = 8; // "RTPS", Version(2), Vendor(2)
     std::memcpy(peer.guidPrefix, encBuf + GuidPrefixOffset, sizeof peer.guidPrefix);
     peer.entityId = RTPS::ENTITYID_PARTICIPANT;
-    const ParticipantCryptoHandle sender = link_->handle_registry()->get_remote_participant_crypto_handle(peer);
+    const ParticipantCryptoHandle sender = !equal_guid_prefixes(peer.guidPrefix, receiver_.local_) ?
+      link_->handle_registry()->get_remote_participant_crypto_handle(peer) :
+      link_->local_crypto_handle();
```
mitza-oci, line 417:
```suggestion
    const ParticipantCryptoHandle sender = equal_guid_prefixes(peer.guidPrefix, receiver_.local_) ?
      link_->local_crypto_handle() :
      link_->handle_registry()->get_remote_participant_crypto_handle(peer);
```
similarly below

PR #4042, opened 2023-02-14
Implement _refcount_value() on OpenDDS::DCPS::LocalObject by delegating to RcObject

simonmcqueen:
This looks like an oversight. As it's ref counting a simple reproducer isn't feasible but I'm medium certain the _refcount_value on this class is being picked up from CORBA::LocalObject when it seems the intent is it should be instead coming from RcObject.

See class Listener in dds/DCPS/MultiTopicDataReaderBase.cpp for a use of LocaObjectBase::_refcount_value().


dds/DCPS/LocalObject.h
```diff
@@ -34,6 +34,9 @@ class LocalObjectBase
   virtual void _remove_ref() {
     RcObject::_remove_ref();
   }
+  virtual CORBA::ULong _refcount_value() {
```
jwillemsen, line 37: const?

simpsont-oci (changes requested):

dds/DCPS/LocalObject.h
```diff
@@ -34,6 +34,9 @@ class LocalObjectBase
   virtual void _remove_ref() {
     RcObject::_remove_ref();
   }
+  virtual CORBA::ULong _refcount_value() const {
+    return RcObject::_refcount_value();
```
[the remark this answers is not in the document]
simonmcqueen: Done - thanks!

dds/DCPS/RcObject.h
```diff
@@ -87,7 +87,11 @@ namespace DCPS {
       }
     }
 
-    /// This accessor is purely for debugging purposes
+    virtual CORBA::ULong _refcount_value() const
+    {
+      return ref_count();
+    }
+
```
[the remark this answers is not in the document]
simonmcqueen: Done - thanks!

mitza-oci:
Not opposed to the change, but I'm also not sure there's a functional difference.  The `Listener` class in the anonymous namespace of `MultiTopicDataReaderBase.cpp` has a call to it, but it's just a forwarding function which nothing else calls (perhaps?).

Could this be unit-tested by comparing results of differently-qualified `_refcount_value()` calls?

simpsont-oci:
> Not opposed to the change, but I'm also not sure there's a functional difference. The `Listener` class in the anonymous namespace of `MultiTopicDataReaderBase.cpp` has a call to it, but it's just a forwarding function which nothing else calls (perhaps?).

I was assuming the motivation for this was some user code (potentially based on TAO/CORBA APIs) for checking / tracking reference counts. Nobody should be using this function beyond debugging (atomically grabbing the reference count tells you nothing about about future values), but it makes sense for the sake of completeness. From the perspective of "redirecting CORBA::Object reference counting APIs to use RcObject reference counts", this is (I believe) the last missing piece.

simpsont-oci (approved):

dds/DCPS/LocalObject.h
```diff
@@ -36,6 +36,9 @@ class LocalObjectBase
   {
     RcObject::_remove_ref();
   }
+  virtual CORBA::ULong _refcount_value() const {
```
mitza-oci, line 39:
```suggestion
  virtual CORBA::ULong _refcount_value() const
  {
```

simonmcqueen: Sorry for that! I see another like it lower down the file but am sure I just carelessly fell into the habit rather than trying to follow suit. Same line brace is/was the style used on Borland product code. Thanks.

mitza-oci: Not a problem.  Since the current code is a mix of different styles we realize it's not always clear.  Thanks for your contribution.

PR #4045, opened 2023-02-16
Add Hardening Features To RtpsRelay

simpsont-oci:
### Problem
RtpsRelay lacks certain features for gauging interaction with poorly behaving (either intentionally or unintentionally) participants, such as a limit for the maximum size of address sets associated with a participant and the ability to ignore/reject certain addresses which behave poorly (such as echoing back traffic) for a period of time.

### Solution
Implement these features, as well as some logging surrounding the data structure sizes.

### DIFF
--- a/tools/rtpsrelay/GuidAddrSet.cpp
+++ b/tools/rtpsrelay/GuidAddrSet.cpp
@@ -25,6 +25,9 @@ GuidAddrSet::record_activity(const Proxy& proxy,
   if (config_.restart_detection()) {
     Remote remote(remote_address.addr, src_guid);
     const auto result = remote_map_.insert(std::make_pair(remote, src_guid));
+    if (result.second) {
+      relay_stats_reporter_.remote_map_size(remote_map_.size(), now);
+    }
     if (!result.second && result.first->second != src_guid) {
       if (config_.log_activity()) {
         ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) INFO: GuidAddrSet::record_activity change detected %C -> %C\n"),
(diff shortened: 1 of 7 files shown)


tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -291,6 +291,11 @@ CORBA::ULong VerticalHandler::process_message(const ACE_INET_Addr& remote_addres
   {
     GuidAddrSet::Proxy proxy(guid_addr_set_);
     proxy.process_expirations(now);
+    if (!proxy.check_address(remote_address)) {
+      HANDLER_WARNING((LM_WARNING, ACE_TEXT("(%P|%t) WARNING: VerticalHandler::process_message %C dropping message from rejected address %C\n"),
```
jrw972, line 295: Remove HANDLER_WARNING.

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -272,4 +298,25 @@ void GuidAddrSet::remove(const Proxy& proxy,
...
+  OpenDDS::DCPS::MonotonicTimePoint expiration = now + config_.rejected_address_duration();
+  auto result = rejected_address_map_.insert(std::make_pair(OpenDDS::DCPS::NetworkAddress(addr), expiration));
+  if (result.second) {
+    if (config_.log_activity()) {
+      ACE_DEBUG((LM_INFO, ACE_TEXT("(%P|%t) INFO: GuidAddrSet::reject_address ")
+                 "Adding %C to list of rejected addresses.\n",
+                 OpenDDS::DCPS::LogAddr(addr).c_str()));
+    }
+    rejected_address_expiration_queue_.push_back(result.first);
+    relay_stats_reporter_.rejected_address_map_size(rejected_address_map_.size(), now);
```
jrw972, line 313: conversion warning.

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -99,6 +105,24 @@ GuidAddrSet::record_activity(const Proxy& proxy,
...
+                 "Rejected address %C expired %C ago, removing from rejeced address map.\n",
+                 OpenDDS::DCPS::LogAddr((*it)->first).c_str(),
+                 ago.str().c_str()));
+    }
+    rejected_address_map_.erase(*it);
+    rejected_address_expiration_queue_.erase(it++);
+    update_reject_stat = true;
+  }
+  if (update_reject_stat) {
+    relay_stats_reporter_.rejected_address_map_size(rejected_address_map_.size(), now);
```
jrw972, line 123: Conversion warning.

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -64,29 +67,32 @@ GuidAddrSet::record_activity(const Proxy& proxy,
...
-                   guid_to_string(src_guid).c_str(),
-                   OpenDDS::DCPS::LogAddr(remote_address.addr).c_str(),
-                   addr_set_stats.get_session_time(now).sec_str().c_str(),
-                   addr_set.size(),
-                   guid_addr_set_map_.size(),
-                   remote_map_.size(),
-                   deactivation_guid_queue_.size(),
-                   expiration_guid_addr_queue_.size(),
-                   admission_control_queue_.size()));
+    AddrSet& addr_set = *addr_set_stats.select_addr_set(remote_address.port, now);
```
jrw972, line 70: Why not just wrap the existing code with `if (config_.max_addr_set_size() == 0 || addr_set.size() <= config_.max_addr_set_size()) {`?

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -99,6 +105,24 @@ GuidAddrSet::record_activity(const Proxy& proxy,
 void GuidAddrSet::process_expirations(const Proxy& proxy,
                                       const OpenDDS::DCPS::MonotonicTimePoint& now)
 {
+  bool update_reject_stat = false;
+  auto it = rejected_address_expiration_queue_.begin();
+  while (it != rejected_address_expiration_queue_.end() && (*it)->second < now) {
+    if (config_.log_activity()) {
+      const auto ago = now - (*it)->second;
+      ACE_DEBUG((LM_INFO, "(%P|%t) INFO: GuidAddrSet::process_expirations "
+                 "Rejected address %C expired %C ago, removing from rejeced address map.\n",
```
jrw972, line 114:
```suggestion
                 "Rejected address %C expired %C ago, removing from rejected address map.\n",
```

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -133,15 +157,17 @@ void GuidAddrSet::process_expirations(const Proxy& proxy,
...
     const auto p = addr_set.find(ga.address);
     if (p == addr_set.end()) {
       continue;
     }
 
     if (p->second <= now) {
       addr_set.erase(p);
-      remote_map_.erase(Remote(ga.address.addr, ga.guid));
+      if (remote_map_.erase(Remote(ga.address.addr, ga.guid)) != 0) {
+        relay_stats_reporter_.remote_map_size(remote_map_.size(), now);
```
jrw972, line 169: conversion warning.

jrw972 (changes requested):

tools/rtpsrelay/GuidAddrSet.cpp
simpsont-oci: You don't want to just wrap the existing code because then when the `addr_set` size is equal to the max, the expiration updates won't get processed. So you have to check to see if it would be increasing the size first, and then limit the increases but not the updates. However, revisiting this code does highlight the fact that we don't technically need the external `if` block, so we can simplify that part.

tools/rtpsrelay/RelayHandler.cpp
simpsont-oci: Completely remove, or different guard? Assuming you mean remove completely, did we want to record the dropped packet somehow? I can't remember what we decided, if we wanted to collect a statistic for rejected messages, or maybe just lump them in with a different statistic (ignored)...

jrw972: Guard is correct.  The log message could flood so just record the message/bytes in the `ignore_*` fields of the handler.

jrw972 (changes requested):
Final step is converting log message to updating ignored_*.

jrw972 (approved):

PR #4048, opened 2023-02-17
configure: Added macOS cross-compile

mitza-oci:
This can be used to have an x86_64 macOS system generate arm64 builds.

### DIFF
--- a/configure
+++ b/configure
@@ -48,7 +48,7 @@ my @default_configh = (
 
 sub perlOS_to_host {
   return 'win32' if $^O eq 'MSWin32';
-  return 'macosx' if $^O eq 'darwin';
+  return 'macos' if $^O eq 'darwin';
   return $^O;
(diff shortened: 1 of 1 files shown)


iguessthislldo (approved):

jwillemsen:
Any idea why conda forge has problems cross compiling ACE, see https://github.com/conda-forge/ace-feedstock/issues/29

mitza-oci:
> Any idea why conda forge has problems cross compiling ACE, see [conda-forge/ace-feedstock#29](https://github.com/conda-forge/ace-feedstock/issues/29)

No, it's hard to tell from that issue what the problem was.
I haven't tried using it yet, but I have what looks like a valid libACE built on x86_64:
```
libACE.dylib: Mach-O 64-bit dynamically linked shared library arm64
```

PR #4049, opened 2023-02-20
Make TryConstruct Test Compatible With Safety Profile

ocielliottc:


mitza-oci (changes requested):
Update `tests/dcps_tests.lst` to have it run this test

PR #4056, opened 2023-02-21
Check usage of IDL unions in marshal_generator

ocielliottc:


mitza-oci:
Please add the C++11 mapping to the tests (it can use the same IDL).  See `tests/DCPS/Compiler/C++11` for examples.

ocielliottc:
Will do.

mitza-oci (approved):

PR #4075, opened 2023-03-10
Initializing DDS::DataWriterQos for constructors is difficult

jrw972:
Problem
-------

InternalDataWriters will be converted to use DDS::DataWriterQos. InternalDataWriters are often initialized in constructors meaning that we need a way to initialize the DDS::DataWriterQoS.

Solution
--------

Provide a builder to initialize DDS::DataWriterQos.

### DIFF
--- a/dds/DCPS/Qos_Helper.h
+++ b/dds/DCPS/Qos_Helper.h
@@ -9,6 +9,8 @@
 #define OPENDDS_DCPS_QOS_HELPER_H
 
 #include "dds/DdsDcpsInfrastructureC.h"
+#include "dds/DdsDcpsPublicationC.h"
+#include "dds/DdsDcpsTopicC.h"
 #include "Time_Helper.h"
 
 #if !defined (ACE_LACKS_PRAGMA_ONCE)
@@ -19,8 +21,6 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace DCPS {
-
-class Service_Participant;
 
 /**
  * @class Qos_Helper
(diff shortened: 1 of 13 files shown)


iguessthislldo (approved):

dds/DCPS/Qos_Helper.h
```diff
@@ -505,6 +505,304 @@ bool operator!=(const DDS::TypeConsistencyEnforcementQosPolicy& qos1,
                 const DDS::TypeConsistencyEnforcementQosPolicy& qos2);
 #endif
 
+class DataWriterQosBuilder {
+public:
+  DataWriterQosBuilder()
```
mitza-oci, line 510: What determines the state of the default-constructed builder?  If it's just the "initial" QoS from Service_Participant, use that to avoid duplication (and it's already portable to all supported compilers).  

In general, default QoS should be determined by the factory object or by the special "use topic QoS" value (though that may not apply to "internal") so the constructor could take a factory object, in this case Publisher.

jrw972: I agree with defining the default qos in one place.  However, I don't want that to be the Service Participant because it makes the qos builder dependent on the Service Participant which will complicate unit tests.  Thus, I'll make the builder provide the default and the Service Participant will initialize using the builder.

I would like to understand a little more about the use cases behind alternate constructors.

Changing the default qos (on anything) involves declaring a variable, getting the qos, changing it, and then setting it.  I'm not sure the builder improves this situation.  If we added methods to copy in from and copy out to the factory, then I suppose this could be done with a (complex) one-liner.

Changing from default qos involves declaring a variable, getting the qos, changing it, and then using it in a factory method or set qos call.  In this case, the builder can change the variable to a temporary by initializing it with the factory default.

More generally, do we think that users will benefit from using builders when constructing or modifying qos?  Some things that come to mind are changing history to keep last means that you should probably set the history depth at the same time and user data is commonly set from string so a helper here would be useful.

mitza-oci: One option would be to have 3 constructors for `DataWriterQosBuilder()`:
* default constructor uses initial values
* constructor that takes a Publisher object reference uses that publisher's default DataWriter QoS
* constructor that takes a Topic uses copy_from_topic_qos

PR #4092, opened 2023-03-28
Cleanup XTypes Section in Devguide

iguessthislldo:
- Changes in General:
  - Added `omgissue` RST Role
  - Make navigation sidebar fixed to top of page since otherwise it can easily go past the bottom of the page and become basically useless.
  - Random minor cleanup from conversion
  - Merge `docs/devguide/index.rst` into `docs/index.rst` and reduce tree depth.
  - Minor fixes for MPC syntax highlighter
- XTypes Section:
  - Extensive cleanup from conversion
  - Minor tweaks to content
  - Replaced ’ (U+2019) with normal single quote ('). Need to do similar things everywhere else.
  - Document information from https://github.com/OpenDDS/OpenDDS/discussions/4083#discussioncomment-5332350
  - Expand "Data Representation" Section to be more detailed and include unaligned CDR. (I thought this was something I already did, but apparently not)
  - Merged "Creating and Using a DynamicDataReader" and "Creating and Using a DynamicDataWriter" sections as they mostly contained the same exact text.
  - Converted OMG Issues to use `omgissue`


docs/devguide/xtypes.rst
```diff
@@ -429,15 +460,49 @@ Type Consistency Enforcement
...
+    * Changing a ``@final`` struct
+    * Adding a member in the middle of an ``@appendable`` struct
+
+  * Length bounds of strings and sequences are the same or greater
+  * Lengths of arrays are exactly the same
+  * The keys of the types match exactly
+  * Shared member IDs match when required, like when they are final or are being used as keys
+
+If the type objects are compatible then the match goes ahead.
+If one or both type objects are not available, then OpenDDS falls back to checking if the topic type names match.
```
mitza-oci, line 501: It might be more clear to call this type "TypeSupport type name" although they're both correct.

docs/devguide/xtypes.rst
```diff
@@ -568,9 +646,9 @@ Determining Extensibility
 ..
     Sect<16.6.3>
 
-The extensibility annotations determine how a type may be changed and still be compatible.
+The extensibility annotations can explicitly define the :ref:`extensibility <xtypes--extensibility>` of a type.
 If no extensibility annotation is set, the type will default to appendable.
-The default can be changed with the command line option --default-extensibility *type*, where *type* can be final, appendable, or mutable.
+This default can be changed with the :ref:`--default-extensibility <opendds_idl--default-extensibility>` opendds_idl option.
```
mitza-oci, line 651: I think this would be clearer if the middle sentence was something like "If no extensibility annotation is used, the compiler's default will take effect.  Without additional command-line options, the compiler's default is (appendable)."

docs/devguide/xtypes.rst
```diff
@@ -922,38 +998,38 @@ Reading Collections of Basic Types
...
-For instance, get_int32_value reads the value from a member of type long, and get_int32_values reads the value from a member of type sequence<long>.
-For the member ul_seq_field in our example, its value can be read as follows.
+For instance, ``get_int32_value`` reads the value from a member of type ``int32``/``long``, and ``get_int32_values`` reads the value from a member of type ``sequence<int32>``.
+For the member ``ul_seq_field`` in our example, its value can be read as follows.
 
-::
+.. code-block:: cpp
 
-    CORBA::ULongSeq my_ul_seq;
+    DDS::UInt32Seq my_ul_seq;
```
mitza-oci, line 1006: Has anyone tried this in a test?  I'm not sure it will directly work because different sequence typedefs in IDL generate distinct mapped classes.

mitza-oci (changes requested):

docs/devguide/opendds_idl.rst
```diff
@@ -313,9 +316,15 @@ Unlike when using the classic mapping, ``Foo.idl`` is not processed by ``tao_idl
...
   * Note that setting a union value through a modifier method automatically sets the discriminator.
     In cases where there are multiple possible values for the discriminator, a 2-argument modifier method is provided.
-    Using this is preferred to using _d().
+    Using this is preferred to using ``_d()``.
+
+  * .. compound::
+
+      If you chose to use the ``_d()`` method of the generated union types, note the following requirement from the specification:
+
+        The ``_d`` discriminator modifier can only be used to set the discriminant to a value within the same union member.
```
jwillemsen, line 325:
``` suggestion
        The ``_d`` discriminator modifier can only be used to set the discriminant to a value that has the same union member as currently selected.
```

docs/devguide/xtypes.rst
```diff
@@ -47,24 +47,27 @@ Extensibility
 ..
     Sect<16.2.1>
 
-There are 3 kinds of extensibility for types: Appendable, Mutable, and Final
+There are 3 kinds of extensibility for types:
 
-* *Appendable* denotes a constructed type which may have additional members added onto or removed from the end.
+*Appendable*
+  It denotes a constructed type which may have additional members added onto or removed from the end, but not both at the same time.
```
jrw972, line 53:
```suggestion
  Appendable denotes a constructed type which may have additional members added onto or removed from the end, but not both at the same time.
```

docs/devguide/xtypes.rst
```diff
@@ -47,24 +47,27 @@ Extensibility
...
+  It's used by default, but a type can be explicitly marked as appendable with the :ref:`@appendable <xtypes--appendable>` annotation.
```
jrw972, line 54:
```suggestion
  Appendable is the default extensibility.
  A type can be explicitly marked as appendable with the :ref:`@appendable <xtypes--appendable>` annotation.
```

docs/devguide/xtypes.rst
```diff
@@ -47,24 +47,27 @@ Extensibility
...
 
-* *Mutable* denotes a constructed type that allows for members to be added, removed, and reordered so long as the keys and the required members of the sender and receiver remain.
+*Mutable*
+  It denotes a constructed type that allows for members to be added, removed, and reordered so long as the keys and the required members of the sender and receiver remain.
```
jrw972, line 57:
```suggestion
  Mutable denotes a constructed type that allows for members to be added, removed, and reordered so long as the keys and the required members of the sender and receiver remain.
```

docs/devguide/xtypes.rst
```diff
@@ -47,24 +47,27 @@ Extensibility
...
   Mutable extensibility is accomplished by assigning a stable identifier to each member.
+  A type can be marked as mutable with the :ref:`@mutable <xtypes--mutable>` annotation.
 
-* *Final* denotes a constructed type that can not add, remove, or reorder members,.
+*Final*
+  It denotes a constructed type that can not add, remove, or reorder members.
```
jrw972, line 62:
```suggestion
  Final denotes a constructed type that can not add, remove, or reorder members.
```

docs/devguide/xtypes.rst
```diff
@@ -106,13 +109,13 @@ Compatible IDL types means that the types are structurally equivalent and serial
...
 Additionally, the XTypes-enabled participant needs to be set up as follows:
 
-* Types cannot use Mutable extensibility
+* Types cannot use mutable extensibility
 
-* Data Writers must have their Data Representation QoS policy set to DDS::XCDR_DATA_REPRESENTATION
+* Data Writers must have their Data Representation QoS policy set to ``DDS::XCDR_DATA_REPRESENTATION``
 
-* Data Readers must include DDS::XCDR_DATA_REPRESENTATION in the list of data representations in their Data Representation QoS (This is true by default)
+* Data Readers must include ``DDS::XCDR_DATA_REPRESENTATION`` in the list of data representations in their Data Representation QoS (This is true by default)
```
jrw972, line 116:
```suggestion
* Data Readers must include ``DDS::XCDR_DATA_REPRESENTATION`` in the list of data representations in their Data Representation QoS (true by default)
```

docs/devguide/xtypes.rst
```diff
@@ -313,9 +316,9 @@ Final Extensibility
 ..
     Sect<16.3.5>
 
-The third kind of extensibility is @final.
-Annotating a type with @final means that it will not be compatible with (assignable to/from) a type that's structurally different.
-The @final annotation can be used to define types for pre-XTypes compatibility or in situations where the overhead of @mutable or @appendable is unacceptable.
+The third kind of extensibility is final.
+Annotating a type with ``@final`` means that it will not be compatible with (assignable to/from) a type that's structurally different.
```
jrw972, line 320:
```suggestion
Annotating a type with ``@final`` means that it will not be compatible with (assignable to/from) a type that is structurally different.
```

docs/devguide/xtypes.rst
```diff
@@ -828,7 +906,7 @@ Interpreting Data Samples with DynamicData
 
 Together with DynamicType, DynamicData allows users to interpret a received data sample and read individual fields from it.
 Each DynamicData object is associated with a type, represented by a DynamicType object, and the data corresponding to an instance of that type.
-Let’s take a look at an example with the following type, described below in IDL:
+Let's take a look at an example with the following type, described below in IDL:
```
jrw972, line 909:
```suggestion
Consider the following example:
```

docs/devguide/xtypes.rst
```diff
@@ -1030,16 +1106,16 @@ To create a DynamicData object, use the DynamicDataFactory API defined by the XT
...
+The "type" input parameter to ``create_data()`` is an object that implements the ``DDS::DynamicType`` interface.
+The DynamicType representation of any type that's supported as a topic data type is available from its corresponding TypeSupport object (see :ref:`xtypes--obtaining-dynamictype-and-registering-typesupport`) using the ``get_type()`` operation.
 Once the application has access to that top-level type, the DynamicType interface can be used to obtain complete information about the type including nested and referenced data types.
 See the file :ghfile:`dds/DdsDynamicData.idl` in OpenDDS for the definition of the DynamicType and related interfaces.
 
 Once the application has created the DynamicData object, it can be populated with data members of any type.
 The operations used for this include the DynamicData operations named "set_*" for the various data types.
-They work similarly to the "get_*" operations that are described in section :ref:`xtypes--interpreting-data-samples-with-dynamicdata`.
+They work similarly to the "get_*" operations that are described in :ref:`xtypes--interpreting-data-samples-with-dynamicdata`.
```
jrw972, line 1118:
```suggestion
They are analogous to the "get_*" operations that are described in :ref:`xtypes--interpreting-data-samples-with-dynamicdata`.
```

jrw972 (changes requested):

docs/devguide/opendds_idl.rst
iguessthislldo: That would probably be a better way of phrasing it, but this is quoting the spec. I will add the section number it's quoting because it's missing, but I don't think the quote should be modified as long it's in the latest published spec.

mitza-oci: I don't think it has to directly quote the spec (but of course we'd need to reword the paragraph above) if there's a better way of explaining it in the DevGuide.

jwillemsen: I have created an OMG issue on the spec

docs/devguide/xtypes.rst
iguessthislldo: The `get_*_values` are defined using `DDS::*Seq`.

https://github.com/OpenDDS/OpenDDS/blob/3734398337a043e067fe6439e09100f974962e94/dds/DdsDynamicData.idl#L307-L308

mitza-oci: Right, I confused this with the other use of `DDS::` module

mitza-oci (approved):
There are a few other suggestions in review comments but I don't think any of them are must-haves

PR #4095, opened 2023-03-30
RtpsRelay: partition-dependent forwarding

mitza-oci:


tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -94,4 +94,25 @@ GuidPartitionTable::Result GuidPartitionTable::insert(const OpenDDS::DCPS::GUID_
...
+    return false;
+  }
+  const auto& found = iter->second;
+  // "partitions" and "found" are both StringSet and we're looking for any string
+  // that's common between the two.  Find the smaller collection to run the loop over.
+  const auto pair = partitions.size() < found.size()
+    ? make_pair(std::cref(partitions), std::cref(found))
+    : make_pair(std::cref(found), std::cref(partitions));
+  for (const auto& part : pair.first) {
+    if (pair.second.count(part)) {
```
jrw972, line 111: Simple string comparison won't work here because of wildcards.

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -94,4 +94,25 @@ GuidPartitionTable::Result GuidPartitionTable::insert(const OpenDDS::DCPS::GUID_
   return result;
 }
 
+bool GuidPartitionTable::has_any_partition(const OpenDDS::DCPS::GUID_t& guid, const StringSet& partitions) const
+{
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, mutex_, false);
+  const auto iter = guid_to_partitions_cache_.find(guid);
+  if (iter == guid_to_partitions_cache_.end()) {
```
jrw972, line 101: I think this should be treated as a cache miss and the cache populated.  The lookup function has code for doing this.

jrw972 (changes requested):

jrw972 (approved):

jrw972 (approved):

PR #4107, opened 2023-04-11
OpenDDS Foundation as author of DevGuide

jwillemsen:
Not 100% sure, but shouldn't have the DevGuide now have the OpenDDS Foundation as author?


jrw972 (approved):

jrw972:
Thanks @jwillemsen

Issue #4109, opened 2023-04-12
OpenDDS 3.24 opendds_idl generates not compilable code


jwillemsen:
Looks there is a regression in 3.24, will create a PR with a test extension soon

```
/home/johnny/ACE/OpenDDS/tests/DCPS/Compiler/union_defaults/union_defaultsTypeSupportImpl.cpp: In member function ‘virtual DDS::ReturnCode_t OpenDDS::XTypes::DynamicDataAdapterImpl<DDS_Native::Example::IDLConversionMessage, DDS_Native::Example::IDLConversionMessage>::get_raw_value(const char*, void*, DDS::TypeKind, DDS::MemberId)’:
/home/johnny/ACE/OpenDDS/tests/DCPS/Compiler/union_defaults/union_defaultsTypeSupportImpl.cpp:20362:67: error: ‘DDS_9Native_Example_B53_dda_tag’ was not declared in this scope; did you mean ‘DDS_9Native_Example_B56_dda_tag’?
20362 |         return get_complex_raw_value< ::DDS_Native::Example::B53, DDS_9Native_Example_B53_dda_tag>(method, dest, tk, value_.b_53, id);
      |                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                   DDS_9Native_Example_B56_dda_tag
/home/johnny/ACE/OpenDDS/tests/DCPS/Compiler/union_defaults/union_defaultsTypeSupportImpl.cpp:20362:99: error: no matching function for call to ‘OpenDDS::XTypes::DynamicDataAdapterImpl<DDS_Native::Example::IDLConversionMessage, DDS_Native::Example::IDLConversionMessage>::get_complex_raw_value<DDS_Native::Example::B53, <expression error> >(const char*&, void*&, DDS::TypeKind&, DDS_Native::Example::B53, DDS::MemberId&)’
20362 |         return get_complex_raw_value< ::DDS_Native::Example::B53, DDS_9Native_Example_B53_dda_tag>(method, dest, tk, value_.b_53, id);
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #4111, opened 2023-04-12
Add new complex types test

jwillemsen:
Reproducer test for issue #4109


iguessthislldo (changes requested):
It's overkill to add a new separate test for this when there's a number IDL tests like this already. For example there's keyword cases here, but we already have a test for that. There's an existing test for `DynamicDataAdatper` specifically in [`tests/unit-tests/dds/DCPS/XTypes/DynamicDataAdapter.idl`](https://github.com/OpenDDS/OpenDDS/blob/master/tests/unit-tests/dds/DCPS/XTypes/DynamicDataAdapter.idl), please add the failing cases there.

jwillemsen:
Closing, looks test cases are added as part of #4117

PR #4120, opened 2023-04-14
TcpDataLink fails to send association message

jrw972:
Problem
-------

The `send_association_msg` method creates a special `REQUEST_ACK`
control message to signal that the local reader is ready to receive
data.  On receipt of the message, the remote writer completes the
association and can start sending data.  `send_association_msg` calls
`send_i` which checks if the publication id of the
`TransportControlElement` is for a stopped (local) client.  The
publication id returned by the `TransportControlElement` is always
GUID_UNKNOWN.  Thus, if GUID_UNKNOWN gets put into the stopped client
set then the `TcpDataLink` will never send the association message and
association will stall on the writer side.

`TransportClient` caches its guid when an association is made.
Suppose the `TransportClient` is stopped before an association.  In
this case the guid is `GUID_UNNOWN` and this guid is put in the
stopped client set of the `TcpDataLink`.

Solution
--------

1. Implement support for the publisher id.
2. Guard against usages of an uninitialized `repo_id_` in `TransportClient`.

### DIFF
--- a/dds/DCPS/transport/framework/TransportClient.cpp
+++ b/dds/DCPS/transport/framework/TransportClient.cpp
@@ -203,6 +203,7 @@ TransportClient::associate(const AssociationData& data, bool active)
   ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, lock_, false);
 
   repo_id_ = repo_id;
+  OPENDDS_ASSERT(repo_id_ != GUID_UNKNOWN);
 
   if (impls_.empty()) {
     if (DCPS_debug_level) {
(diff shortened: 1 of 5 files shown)


iguessthislldo (approved):

simpsont-oci:
> In the TCP transport, the writing side performs the active part of the TCP connection.
I don't think this is always true, as it depends on if it can find a preexisting connection (which may have been "performed" by a remote writer to a local reader).

jrw972:
Do I need to qualify it like "in the event that a connection cannot be found".?

iguessthislldo (approved):

PR #4122, opened 2023-04-17
IDL-accessible encoding conversions for topic types

mitza-oci:
Extends the generated TypeSupport IDL with encode_to_string/bytes and decode_from_string/bytes.

Support for CDR is not yet implemented.
JSON is supported using the existing JsonValueWriter/Reader. 
Other types can be added as needed.


java/tests/vread_vwrite/run_test.pl
```diff
@@ -0,0 +1,32 @@
...
+use Env qw(ACE_ROOT DDS_ROOT);
+use lib "$DDS_ROOT/bin";
+use lib "$ACE_ROOT/bin";
+use PerlDDS::Run_Test;
+use PerlDDS::Process_Java;
+use strict;
+
+my $status = 0;
+
+#PerlACE::add_lib_path ("$DDS_ROOT/java/tests/complex_idl");
```
jrw972, line 16: Remove?

jrw972 (approved):

YosefBdr:
Hello, Thanks for your work.. I really appreciate it..
I was wondering if there is an example on how to use decode_from_bytes?

iguessthislldo:
I don't think we have any documentation or stand alone examples yet, but here is the relevant part of the test for it:

https://github.com/OpenDDS/OpenDDS/blob/bdfdde621cde0a056bdc79835dabcd25c4da51fe/tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.cpp#L21-L30

Also the [Java test for it](https://github.com/OpenDDS/OpenDDS/blob/bdfdde621cde0a056bdc79835dabcd25c4da51fe/java/tests/vread_vwrite/VreadVwrite.java#L51) in case you or someone else wants that.

One thing I should note since you asked for `decode_from_bytes` specifically instead of `decode_from_string` is I don't think formats other than `JSON_DATA_REPRESENTATION` are supported yet, so if you want to decode CDR data, you'll have to use the internal `Serializer` class. See [here](https://stackoverflow.com/questions/72015309/serialize-opendds-topic-to-a-stdstring) for an example of using it. `decode_from_bytes` will work with JSON though if that's what you want to use it for.

PR #4125, opened 2023-04-20
Generate News in Sphinx From Fragments

iguessthislldo:
# Problem

The current way of generating the news for releases mostly consists of generating a spreadsheet of PRs, editing that, and creating the new entries from that manually. More info on that process [here](https://opendds.readthedocs.io/en/master/internal/release.html#update-files-in-the-repo-as-needed) and [here](https://github.com/OpenDDS/OpenDDS/blob/f511b1f1582664ab7f49b3b012b968e684928aa2/tools/scripts/release_notes/README.md). News entries can be directly committed in the PR where the change is taking place, but doing that risks merge conflicts.

Overall this process is somewhat messy and limiting:

- Deciding what's newsworthy, what exactly to write, and reviewing the news is done all at once right before the release, sometimes months after the work was done. This makes it harder to remember what's newsworthy, what specific things needs to be pointed out to users, and what PRs should go together for single news item. This also means it takes more time to prepare the release and there's less time to spot and correct mistakes in the news or improve it.
- Most of the time the news item is left as just the title of PR. In the best case these might not need to be tweaked much or at all for changes that require little explanation. However this is certainly inadequate for explaining larger changes, for example like for [the XTypes fixes from PR4078]( https://github.com/OpenDDS/OpenDDS/blob/f511b1f1582664ab7f49b3b012b968e684928aa2/NEWS.md?plain=1#L49). It'd be very awkward to write that much in a spreadsheet.
- It's hard to link to documentation. This is better than it was before with the PDF devguide, when it was impossible, but this could still be improved on more. Linking would give more context to users and could immediately give them details on how to use a new feature.
- Contributors outside the OpenDDS maintainers basically have no direct input on what the news says for changes they contribute. Honestly I'm not sure if any have wanted to, but they couldn't if they did.

# Solution

The solution in this PR is committing the news of changes of a PR as a file in that PR. At release these fragments of the news are brought together automatically. There are still a few kinks to iron out, but even if those are mostly unresolved I think this system will improve the quality of the news.

## Other News Tools

The system is inspired by [Python's blurb tool](https://pypi.org/project/blurb/) and to a lesser extent tools like [towncrier](https://towncrier.readthedocs.io/en/stable/index.html). These tools are not bad, but they have some serious drawbacks. blurb is specifically tailored for CPython development. For example, it's oriented by GitHub issues, where as many of the changes we make are not prompted by a GitHub issue. towncrier really expects the project to be a Python project and has some quirks for some of use cases I was looking for. Specifically needing multiple identical files for to attribute a news item to multiple PRs and needing multiple files for a PR to have different kinds of changes. Also both rely on the files having a specific name, which seems unnecessary to me.

## Making the News from Fragments

Fragments are documented here: https://opendds--4125.org.readthedocs.build/en/4125/internal/docs.html#news

[This is the news page preview](https://opendds--4125.org.readthedocs.build/en/4125/news.html) for this PR. It has two duplicate 3.24.0 entries, because there's one from the test fragments that'll be removed before merging and the committed one to keep.

# TODO

Here are the two main issues I see with this system right now:

- [x] To do a PR with a news fragment in one commit, you basically have to know what the PR number is going to be before hand. Otherwise another commit is needed to add the PR number. The PR number could technically be manually added after the PR is merged, but I would consider that poor practice to this manually. One solution could be a placeholder in `news-pr` statement that an action automatically replaces with the PR number after the PR is merged. **I tried to this, but right after a PR is opened. The issue with that is that it's impossible for a workflow in a PR to push to a branch of a fork. It now it does what I originally imagined, which is replace the placeholder after the merge.** **(UPDATE: This is put off for now)**
- [X] How does this relate/integrate with `NEWS.md` and the GitHub release notes? I'm honestly a little stumped by this and unlike the other issue this needs to be figured out before this can be merged.
  - Something like pandoc could be used to convert the rst, but it would still need some manual intervention based on tests I did with the 3.24.0 news. **I ended up doing this, but using a modified Sphinx builder for Markdown instead of pandoc. This process is better than what pandoc could do because the markdown acts as though it was part of the Sphinx documentation because it is.**
  - The markdown version could just be a summarized version of the news, mostly consisting of highlights. This could be manually done or done with pandoc with human intervention. Also this summery could be what goes out in a prerelease announcement on social media.
  - The `NEWS.md` file could be also be done away with to simplify things. If that's the case, should news.rst live in the root directory and be called `NEWS.rst`? Is that going to case problems to try to include it in Sphinx?
  - The GitHub release notes could just link to `news.rst`, but I feel like they probably should at least have a summery.

Besides that there are some more things I needs to do, specifically:

- [X] Document either in the documentation guidelines or dev guidelines how to add to the news. **DONE**
- [X] Improve release entries, it needs the release date and output could be tweaked.
- [X] Maybe add two smaller examples just for "Additions" and "Fixes" without comments that are easier to use as templates. **Decided one simple example is probably enough, documenting everything that was in the bigger example in the docs.**
- [x] Integrate this into the release process:
  - Replace `NEWS.md` steps of release script with `news.py release`
  - Remove `NEWS.md` template
  - Update release process documentation
- [x] Before merge, remove 3.24.0 fragments, add any newer releases, and add any news fragments for a pending release.
- [x] Probably move https://github.com/iguessthislldo/sphinx-markdown-builder into the OpenDDS org since this process relies on it to generate the Markdown news.


docs/news.py
```diff
@@ -0,0 +1,164 @@
...
+            file=sys.stderr)
+    elif inserted:
+        print('Pushing changes...')
+        run('git', 'config', 'user.name', 'github-actions[bot]')
+        run('git', 'config', 'user.email', 'github-actions[bot]@users.noreply.github.com')
+        run('git', 'commit', '--all', '--message',
+            'Insert PR Number into News Fragment(s) [skip actions]')
+        run('git', 'push')
+    else:
+        print('No changes to push')
```
jrw972, line 138: Can/should this be moved to the github actions workflow?

iguessthislldo: It could, but then I'd have to get the distinction between no change/change/error out of the script. It could be done with the exist status, but this way just seems simpler.

docs/news.py
```diff
@@ -0,0 +1,164 @@
...
+    # Then try to use that to ask GitHub for the PR with that commit
+    pr_num = subprocess.check_output(
+        [
+            'gh', 'pr', 'list', '--search', sha, '--state', 'merged',
+            '--json', 'number', '--jq', '.[0].number'
+        ],
+        cwd=str(docs_path)).decode('utf-8').strip()
+    print('  pr_num', repr(pr_num))
+
+    return pr_num if pr_num else None
```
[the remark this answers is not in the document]
iguessthislldo: If there was just one PR per fragment, then sure I wouldn't bother with the automatic commits, but for follow on PRs how do we know what PRs belong to the fragment if the PR doesn't change them?

iguessthislldo:
@mitza-oci I added a news entry for #4122 as I think it's the only news-worthy PR that has been merged so far: https://github.com/OpenDDS/OpenDDS/blob/822e0c1d6dcb04ea4cfcb09e80929a2c0b1734a7/docs/news.d/sample_convert.rst

mitza-oci (approved):

docs/internal/release.rst
```diff
@@ -11,8 +11,8 @@ Notes About OpenDDS Releases
 - A micro (aka patch) (X.Y.Z) release is somewhat different from a minor (X.Y.0) release.
   The steps where this is the case will be marked as such.
 - All authoritative references to versions in the source code (like :ghfile:`VERSION.txt` or :ghfile:`dds/Version.h`) will be automatically updated by the release script, so there’s no need to do it manually.
-  This also applies to :ghfile:`NEWS.md` where the release script expects the start of the current entry to be exactly the same as when it inserted it and it will change it to the release version.
-  So the only thing that needs to manually edited is the list of changes.
+  Other files such as :ghfile:`NEWS.md` and some parts of :ghfile:`README.md` also shouldn't be updated manually.
```
jrw972, line 14:
```suggestion
  Other files such as :ghfile:`NEWS.md` and some parts of :ghfile:`README.md` shouldn't be updated manually.
```

docs/internal/docs.rst
```diff
@@ -193,8 +217,229 @@ omgissue
...
+
+``Security``
+  Fixes for issues that might compromise the security of a system using OpenDDS.
+  This doesn't need to be related to :doc:`DDS Security </devguide/dds_security>`.
+
+``Fixes``
+  User-relevant fixes not related to security or documentation.
+
+``Documentation``
+  Signifiant changes to documentation, either in the primary Sphinx documentation or in secondary documentation.
```
jrw972, line 319:
```suggestion
  Significant changes to documentation, either in the primary Sphinx documentation or in secondary documentation.
```

jrw972 (changes requested):

docs/internal/docs.rst
iguessthislldo: It's annoying that's a word. Either I must have either typed that as is or mispelled significant enough that it suggested "signifiant" first and I wasn't paying attention.

PR #4134, opened 2023-04-27
OpenDDS lacks a dynamic configuration store

jrw972:
Problem
-------

Configuration in OpenDDS has the following issues:

* Values that can be set via configuration file cannot be set via command line.
* Values that can be set via API cannot be set via configuration file.
* Configuration APIs are written in C++ and not fully exposed for other language bindings.
* Dynamic reconfiguration is messy.
* OpenDDS cannot be configured via environment variables.
* OpenDDS cannot take advantage of multiple configuration files.

Solution
--------

Implement a store of key-value pairs based on an Internal DDS Topic. The key is the canonicalized name of the configuration value and the value is the desired value stored as a string.  Using an Internal DDS Topic will allow different modules to subscribe to the configuration for dynamic updates.

The interface to the configuration store will be captured in IDL so that it can be exposed for different language bindings.

Subsequent work will apply the store to command-line arguments, configuration file processing, and environment variables.


jrw972:
@iguessthislldo @mitza-oci The IDL interface that I've proposed is quite narrow.  Should I add additional getters and setters and if so, what types do we want to support?  Assuming we do want additional types, should we be using aliases for fixed widths, plain IDL types, something from XTypes/DynamicData, etc.?

dds/DCPS/Time_Helper.h
```diff
@@ -10,6 +10,7 @@
 
 #include <dds/DdsDcpsCoreC.h>
 #include <dds/DdsDcpsInfoUtilsC.h>
+#include "PoolAllocator.h"
```
iguessthislldo, line 13:
```suggestion
#include "PoolAllocator.h"

#include <dds/DdsDcpsCoreC.h>
#include <dds/DdsDcpsInfoUtilsC.h>
```

dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -0,0 +1,383 @@
...
+ConfigStoreImpl::set(const char* key,
+                     const TimeDuration& value,
+                     IntegerTimeFormat format)
+{
+  switch (format) {
+  case Format_IntegerMilliseconds:
+    set_int32(key, value.value().msec());
+    break;
+  case Format_IntegerSeconds:
+    set_int32(key, value.value().sec());
```
iguessthislldo, line 307: Narrow warning here

dds/DCPS/ConfigStoreImpl.h
```diff
@@ -0,0 +1,148 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef OPENDDS_DCPS_CONFIG_STORE_IMPL_H
+#define OPENDDS_DCPS_CONFIG_STORE_IMPL_H
+
+#include "dcps_export.h"
+
```
iguessthislldo, line 10:
```suggestion
```

Export isn't a special header in terms of order. (In fact I could argue it should always go at the bottom to flush out places where we don't include it, but still use the export macro. I've found a couple of those cases before.)

iguessthislldo (changes requested):

dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -0,0 +1,383 @@
...
+  DCPS::InternalSampleInfoSequence infos;
+  config_reader_->read_instance(samples, infos, DDS::LENGTH_UNLIMITED, ConfigPair(key, ""),
+                                DDS::ANY_SAMPLE_STATE, DDS::ANY_VIEW_STATE, DDS::ALIVE_INSTANCE_STATE);
+  for (size_t idx = 0; idx != samples.size(); ++idx) {
+    const ConfigPair& sample = samples[idx];
+    const DDS::SampleInfo& info = infos[idx];
+    if (info.valid_data) {
+      CORBA::Long x = 0;
+      if (DCPS::convertToInteger(sample.value(), x)) {
+        value = x;
```
sonndinh, line 109: `break` after this?

dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -0,0 +1,383 @@
...
+      } else if (log_level >= LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING,
+                   ACE_TEXT("(%P|%t) WARNING: ConfigStoreImpl::get_int32: ")
+                   ACE_TEXT("failed to parse int32 for %C=%C\n"),
+                   sample.key().c_str(), sample.value().c_str()));
+      }
+    }
+  }
+
+  return value;
```
sonndinh, line 119: This returns the value provided by the caller in case the store doesn't have this key. What is an example use case for that? I thought it should return a default value defined by OpenDDS for this key instead, shouldn't it?

dds/DCPS/ConfigStoreImpl.cpp
jrw972: This is me being paranoid.  If we ever change the history depth qos, then we could end up with multiple values for one key.  This code ensures that the last (most recent) value is the one returned.

dds/DCPS/ConfigStoreImpl.cpp
jrw972: You are correct.  The code that is fetching a value will provide the OpenDDS default.  The next PR where the ConfigStore gets used should help illustrate.

PR #4137, opened 2023-05-04
Observer doesn't support dispose/unregister

jrw972:
Problem
-------

The Observer interface doesn't support dispose and unregister. Support for dispose and unregister is necessary for logs that must contain these activities.

Solution
--------

Implement support for dispose and unregister on both the writer and reader side.

Notes
-----

* The DataReader support for Observer has moved to `finish_store_instance_data`.  This means that observers will not receive samples that are rejected due to resource limits.

* The Observer Test has been rewritten as a single-process test.


dds/DCPS/DataReaderImpl.cpp
```diff
@@ -1429,9 +1425,9 @@ DataReaderImpl::data_received(const ReceivedDataSample& sample)
     bool is_new_instance = false;
     bool filtered = false;
     if (sample.header_.key_fields_only_) {
-      dds_demarshal(sample, publication_handle, instance, is_new_instance, filtered, KEY_ONLY_MARSHALING, false);
+      dds_demarshal(sample, publication_handle, instance, is_new_instance, filtered, KEY_ONLY_MARSHALING);
     } else {
-      real_data = dds_demarshal(sample, publication_handle, instance, is_new_instance, filtered, FULL_MARSHALING, observer && vd);
+      dds_demarshal(sample, publication_handle, instance, is_new_instance, filtered, FULL_MARSHALING);
```
mitza-oci, line 1430: Since these statements are essentially the same, collapse them down to one (either a named variable or a `? :` would work well for the one different parameter).

mitza-oci (changes requested):

iguessthislldo (approved):

dds/DCPS/TopicDescriptionImpl.h
```diff
@@ -94,39 +94,39 @@ class TopicDescriptionPtr
 {
 public:
   TopicDescriptionPtr(Topic* topic=0)
-    : topic_(topic)
   {
+    topic_ = topic;
```
mitza-oci, line 98: Does this work with the initializer list (also below on line 113)?

mitza-oci (approved):

PR #4144, opened 2023-05-18
Support JSON in CMake compilations

jmmorato:
Required to use `encode_to_string` in the code generated `TypeSupport` classes


iguessthislldo (approved):
I will have to rearrange this in #4140 because I refactored a lot of the CMake package there, but I wasn't aware of this problem, so this helps a lot.

PR #4151, opened 2023-05-22
`TransportInst` does not use `ConfigStore`

jrw972:
Problem
-------

`TransportInst` does not use `ConfigStore`.  See #4134.

Solution
--------

Convert `TransportInst` to use `ConfigStore`.

* `queue_messages_per_pool_` and `queue_initial_pools_` were removed because they are not used.

* Member access has been replaced with getters and setters.  This will break code that interacts with `TransportInst`.  The code should be updated to either use the getters or setters or use the ConfigStore.

* Members for dropping and counting messages have been relocated close to where they are used and updated via a listener.  If concurrent access becomes an issue, then they can be wrapped with `Atomic`.


docs/news.d/config_store.rst
```diff
@@ -0,0 +1,7 @@
+.. news-prs: 4134 4138 4151
+
+.. news-start-section: Additions
+- OpenDDS now stores configuration information in a key-value store.
+  - Configuration values can be set via API, config file, or command line.
+    - Currently applies to the `common` section and common transport configuration.
```
iguessthislldo, line 6:
```suggestion
- OpenDDS now stores configuration information in a key-value store.

  - Configuration values can be set via API, config file, or command line.
  
    - Currently applies to the ``common`` section and common transport configuration.
```

RST nested lists have to be separated by a blank line or else [they're not interpreted correctly](https://opendds--4151.org.readthedocs.build/en/4151/news.html#additions).

Also we could add a small section on this to the to [runtime config chapter](https://github.com/OpenDDS/OpenDDS/blob/master/docs/devguide/run_time_configuration.rst) in the devguide and link it here.

iguessthislldo: That's true, I'm okay with that.

dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -12,27 +12,52 @@
...
+  }
+
+  while (idx < key.size()) {
+    const char x = key[idx];
 
+    if (idx + 1 < key.size()) {
+      // Deal with camelcase;
+      const char y = key[idx + 1];
+
+      if (ACE_OS::ace_isupper(x) && ACE_OS::ace_islower(y) && !retval.empty() && retval[retval.size() - 1] != '_') {
```
iguessthislldo, line 39: Kebab case seems like it would have been easier to deal with.

iguessthislldo (approved):

PR #4154, opened 2023-05-25
Some language mappings don't support equality comparison

jrw972:
Problem
-------

The type support for C++11 does not include equality operators (`==` and `!=`).  The operators make the IDL defined types easier to use with generic programming techniques.

Solution
--------

Implement support for `==` and `!=`.

### DIFF
--- a/dds/idl/langmap_generator.cpp
+++ b/dds/idl/langmap_generator.cpp
@@ -370,7 +370,7 @@ struct GeneratorBase
           "    return *this->_u." << field_name << ";\n"
           "  }\n";
       } else {
-        std::cerr << "Unsupported type for union element\n";
+        idl_global->err()->misc_warning("Unsupported type for union element", field_type);
       }
     }
   };
@@ -430,7 +430,7 @@ struct GeneratorBase
       be_global->lang_header_ <<
         "    " << lang_field_type << "* " << branch->local_name()->get_string() << ";\n";
     } else {
-      std::cerr << "Unsupported type for union element\n";
+      idl_global->err()->misc_warning("Unsupported type for union element", field_type);
     }
   }
 
@@ -455,7 +455,7 @@ struct GeneratorBase
       ss <<
         "    this->_u." << name << " = (other._u." << name << ") ? new " << lang_field_type << "(*other._u." << name << ") : 0;\n";
     } else {
-      std::cerr << "Unsupported type for union element\n";
+      idl_global->err()->misc_warning("Unsupported type for union element", field_type);
     }
 
     return ss.str();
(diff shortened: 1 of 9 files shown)


tools/dds/rtpsrelaylib/Relay.idl
```diff
@@ -6,10 +6,18 @@
...
 
+  // Replicate DDS types to get C++11 versions.
+  struct Time_t {
+    long sec;
+    unsigned long nanosec;
+  };
+  typedef sequence<octet> OctetSeq;
+  struct UserDataQosPolicy {
+    OctetSeq value;
+  };
```
iguessthislldo, line 19: Why are these being duplicated?

dds/idl/langmap_generator.cpp
```diff
@@ -516,6 +516,24 @@ struct GeneratorBase
...
+  {
+    std::stringstream ss;
+
+    AST_Type* actual_field_type = resolveActualType(field_type);
+    const Classification cls = classify(actual_field_type);
+    if (cls & (CL_PRIMITIVE | CL_ENUM | CL_STRING | CL_ARRAY | CL_STRUCTURE | CL_UNION | CL_SEQUENCE | CL_FIXED)) {
+      ss <<
+        "    return this->_" << name << " == rhs._" << name << ";\n";
+    } else {
+      std::cerr << "Unsupported type for union element\n";
```
iguessthislldo, line 531: If this is supposed to be a warning, you can use:

https://github.com/DOCGroup/ACE_TAO/blob/ab3b192fcda5715a0c92e5627acdeb4e8f71911c/TAO/TAO_IDL/include/utl_err.h#L448

I think it would be:

```suggestion
      idl_global->err()->misc_warning("Unsupported type for union element", node);
```

iguessthislldo (changes requested):
There's also an error in the multi topic test. I think I converted that test to use the C++11 mapping at some point and probably implemented my own operators that now should probably be removed.

jwillemsen:
Does this work with valuetypes? The IDL to C++11 language mapping doesn’t define these operators as not all IDL types are comparable, but very likely the subset currently used by OpenDDS does by accident

jrw972:
> Does this work with valuetypes? The IDL to C++11 language mapping doesn’t define these operators as not all IDL types are comparable, but very likely the subset currently used by OpenDDS does by accident

No; `opendds_idl` is unaware of valuetypes.  Support for valuetypes will be necessary when `tao_idl` is phased out.

iguessthislldo (changes requested):

jwillemsen:
Any documentation extension needed (including warning that the generation of the operator is non-portable and has certain restrictions)?

PR #4157, opened 2023-05-31
Add `.editorconfig` and Document Lint Script

iguessthislldo:
[EditorConfig](https://editorconfig.org/) can help enforce our text file formatting. It is natively supported by many editors including Visual Studio and Neovim (the fork of Vim I use) with plugins available for other editors. It's also used by many big projects such as Jekyll, Git, Rust, Python, and CMake (where I recently first noticed it). It's also used by Micronaut. This means even if it's not natively supported by a contributor's editor, there's a chance they already have a plugin for it installed.

### DIFF
--- a/docs/internal/dev_guidelines.rst
+++ b/docs/internal/dev_guidelines.rst
@@ -80,6 +80,25 @@ These apply to C++ source code, Perl scripts, MPC files, and any other plaintext
   I don't think it makes sense to strictly enforce an 80-column limit, but overly long lines are harder to read.
   Try to keep lines to roughly 80 characters.
 
+The :ref:`lint script <dev_guidelines-lint-script>` will help check for most of these in a PR.
+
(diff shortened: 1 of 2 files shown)


docs/internal/dev_guidelines.rst
```diff
@@ -80,6 +80,25 @@ These apply to C++ source code, Perl scripts, MPC files, and any other plaintext
...
+
+.. _dev_guidelines-lint-script:
+
+***********
+Lint Script
+***********
+
+The :ghfile:`lint script <tools/scripts/lint.pl>` is a Perl script that is run on every PR.
+It checks for some mistakes both in coding and style.
+See the default checks by running ``./tools/scripts/lint.pl --list``.
```
mitza-oci, line 96: In general I'm not in favor of using `./` prefixes for commands.
We should not assume any particular kind of shell / command interpreter is being used.

PR #4162, opened 2023-06-08
`RtpsUdpInst` does not use `ConfigStore`

jrw972:
Problem
-------

`RtpsUdpInst` does not use `ConfigStore`.  See #4134.

Solution
--------

Convert `RtpsUdpInst` to use `ConfigStore`.

* Member access has been replaced with getters and setters.

* Dynamic configuration related to the RtpsRelay is handled via a listener.

* The `RtpsUdpCore` caches dynamic configuration values and values that are used frequently.

Notes
-----

This fixes a bug in `ConfigStoreImpl` where the last valid value was returned instead of the provided default in the event of an error.


dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -103,6 +104,11 @@ DDS::Boolean
 ConfigStoreImpl::get_boolean(const char* key,
                              DDS::Boolean value)
 {
+  if (DCPS_debug_level) {
```
iguessthislldo, line 107: This seems like it should be higher.

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -734,13 +663,11 @@ void
...
 {
 #ifdef OPENDDS_SECURITY
-  RtpsUdpInst_rch cfg = config();
-
-  if (cfg && cfg->use_ice()) {
-    stop_ice();
-  }
+    if (core_.use_ice()) {
+      stop_ice();
+    }
```
iguessthislldo, line 668:
```suggestion
  if (core_.use_ice()) {
    stop_ice();
  }
```

dds/DCPS/transport/rtps_udp/RtpsUdpTransport.cpp
```diff
@@ -734,13 +663,11 @@ void
...
 
-  relay_stun_task_->cancel();
+    relay_stun_task_->cancel();
```
iguessthislldo, line 670:
```suggestion
  relay_stun_task_->cancel();
```

iguessthislldo (changes requested):

dds/DCPS/ConfigStoreImpl.cpp
jrw972: Ideally, the user would only see configuration messages at start-up and during runtime if they change something.  I thought it was better to set these low so the configuration would be more obvious when debugging.

iguessthislldo: I ran DevGuide Messenger on this branch and `ConfigStore::get` lines make up almost half the lines in the publisher log (299/620). There are a lot at the beginning, but also at least once on each write. I really think getting a value should be much much higher. Going by the [`DCPS_debug_level` list](https://github.com/OpenDDS/OpenDDS/blob/72462b97d5ee53f18e699260406487ae419e2ffd/dds/DCPS/debug.h#L69-L74) something happening every writer should be least an 8, if not it's own category like the special transport or security ones. Getting a value also doesn't say the value, so it doesn't seem that useful other than saying something wanted that value.

Setting a value should probably be lower than that since it happens a lot less often (27/620). However a lot of it is duplicating other logging, like when transports dump their settings, so I'm not sure it should be `DCPS_debug_level>=1`. Maybe at least 2?

jrw972: I'll reproduce your experiment.  My goal is to _not_ have a get for every write.

I'll clean up the logging so that it prints the key and value for `DCPS_debug_level >= 2`.

iguessthislldo (approved):

tests/unit-tests/dds/DCPS/ConfigStoreImpl.cpp
```diff
@@ -173,57 +184,146 @@ TEST(dds_DCPS_ConfigStoreImpl, set_get_TimeDuration_milliseconds)
...
+TEST(dds_DCPS_ConfigStoreImpl, take_has_prefix)
+{
+  ConfigTopic_rch topic = make_rch<ConfigTopic>();
+  ConfigStoreImpl store1(topic);
+  ConfigReader_rch reader = make_rch<ConfigReader>(store1.datareader_qos());
+  topic->connect(reader);
+
+  store1.set("PREFIX1_key", "value");
+  EXPECT_TRUE(take_has_prefix(reader, "PREFIX1"));
+  EXPECT_FALSE(take_has_prefix(reader, "PREFIX1"));
```
iguessthislldo, line 270: This isn't really related to this code, but reading it made me think of it. Does the config store warn if a value wasn't used? If not, would it be possible? There's different things grabbing things from the key store, so I assume that could be tracked as bool on `ConfigPair` and at some point you'd know if something was used or not. If someone misunderstood how to make a key value or mistyped one, which could happen to a user or one of us, then it seems like it would be good for some sort of feedback.

iguessthislldo (changes requested):

dds/DCPS/Time_Helper.inl
```diff
@@ -349,12 +349,22 @@ namespace {
...
+    return "DDS::INFINITE_DURATION";
+  }
+
   return to_dds_string(duration.sec) + '.' + left_pad(to_dds_string(duration.nanosec), '0', 9);
 }
 
 ACE_INLINE OpenDDS_Dcps_Export
 bool from_dds_string(const String& str, DDS::Duration_t& duration)
 {
+  if (str == "DDS::INFINITE_DURATION") {
```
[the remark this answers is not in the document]
jrw972: The [JSON spec](https://www.omg.org/spec/DDS-JSON/1.0/PDF) puts forth `DURATION_INFINITY`.  Do you think that's an acceptable option?

tests/unit-tests/dds/DCPS/ConfigStoreImpl.cpp
jrw972: Good question.  The InternalDataReader keeps the view state and sample state.  So, if everything accesses the config through the same reader, then those states would reflect if a value was accessed or not.  Writing the function to do the scan would be easy.  The challenge would be when to invoke it.

dds/DCPS/Time_Helper.inl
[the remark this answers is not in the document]
iguessthislldo: Yes, I think it's alright.

docs/news.d/config_store.rst
```diff
@@ -0,0 +1,19 @@
...
+  -  use_multicast_
+  -  ttl_
+  -  multicast_interface_
+  -  anticipated_fragments_
+  -  max_message_size_
+  -  nak_depth_
+  -  nak_response_delay_
+  -  heartbeat_period_
+  -  receive_address_duration_
+  -  responsive_mode_
```
iguessthislldo, line 17: These should be quoted because they're symbol names. I don't care so much about that in a list like this and wasn't going say anything about this but [trailing underscores are part of the hyperlink syntax in RST](https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets), so this broke the sphinx_strict check.

```suggestion
  -  ``send_buffer_size_``
  -  ``rcv_buffer_size_``
  -  ``use_multicast_``
  -  ``ttl_``
  -  ``multicast_interface_``
  -  ``anticipated_fragments_``
  -  ``max_message_size_``
  -  ``nak_depth_``
  -  ``nak_response_delay_``
  -  ``heartbeat_period_``
  -  ``receive_address_duration_``
  -  ``responsive_mode_``
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4168, opened 2023-06-12
Transport logging changes, fixed typo

jwillemsen:


iguessthislldo:
I'm going to close this because it looks like #4171 has the same commits.

jwillemsen:
#4171 has a fix included, this was purely logging/typo.

jwillemsen:
Closed, this is not a complete fix, only the creator of the shared memory segment should use `shmctl` to cleanup, but the other processes using the shared memory do use `shmat` and they should use `shmdt` when they don't use the shared memory anymore. Looks the `ACE_Shared_Memory_Pool` used by OpenDDS doesn't do use `shmdt` anywhere

PR #4171, opened 2023-06-13
Release the used shared memory in the datalink

### LINKED ISSUE #4163: Shared memory transport doesn't release memory on publisher shutdown
When we have a subscriber which uses the shared memory transport to receive data from a publisher using a very large shared memory buffer we observe that when the publisher does shutdown cleanly the subscriber doesn't release the shared memory transport, on each restart of the publisher the memory usage of the subscriber increases each time.

jwillemsen:
At the moment the datalink for shared memory is closed we must release the shared memory in order to let the OS free it when it is not used anymore 

Fixed #4163


dds/DCPS/transport/shmem/ShmemDataLink.cpp
```diff
@@ -233,7 +233,10 @@ ShmemDataLink::stop_i()
   {
     ACE_GUARD(ACE_Thread_Mutex, g, peer_alloc_mutex_);
     if (peer_alloc_) {
-      peer_alloc_->release(0 /*don't close*/);
+      if (peer_alloc_->release(1 /*close*/) == -1) {
```
mitza-oci, line 236: I haven't looked at the ACE PR in detail yet, but would having that fix remove the need to pass `1` here?

jwillemsen: I tested again, but a 1 is required, only than `release` on the underlying `ACE_Shared_Memory_Pool`, see https://github.com/DOCGroup/ACE_TAO/blob/5ea3ce0f676f3ae3d9347cf43b762d09c32745fd/ACE/ace/Malloc_T.inl#L65. That will release the shared memory from this process.

mitza-oci: OK, I need to review in more detail then.  There should be some distinction between detaching from a peer's allocator and completely removing one's own allocator.

jwillemsen: Only when called with `1` `ACE_Malloc_T` will call the underlying `release of `ACE_Shared_Memory_Pool`, the decision to destroy it is done based on the reference counting in the PI control block. With just `0` nothing happens to the `ACE_Shared_Memory_Pool`, looks the API between `ACE_Malloc_T` and `ACE_Shared_Memory_Pool` aren't 100% fit but we can't change that. 

Did just a test using `DevGuideExamples/DCPS/Messenger`, configured it to use shared memory, keep the subscriber running for some minutes and restarted the publisher multiple times, shared memory segment for the subscriber stays at 1, when the subscriber exists its own shared memory segment is also destroyed.

jwillemsen:
There is a remark in `ACE_Shared_Memory_Pool` that it is better to use `ACE_MMAP_Memory_Pool`, maybe something for OpenDDS 4, but very likely switching will cause interoperability issues when OpenDDS 3 and OpenDDS 4 based applications want to use shared memory between them on the same host

jrw972:
Please prepare a [news fragment](https://opendds.readthedocs.io/en/master/internal/docs.html#news) for this PR.

dds/DCPS/transport/shmem/ShmemDataLink.cpp
mitza-oci: Right, I didn't have the ACE code in front of me when I first reviewed it so it wasn't clear that the Malloc_T object treats this parameter differently.

mitza-oci:
> There is a remark in `ACE_Shared_Memory_Pool` that it is better to use `ACE_MMAP_Memory_Pool`, 

I'm not sure mmap is better for this use case (persistence doesn't matter; getting the filesystem involved adds complexity).

> maybe something for OpenDDS 4, but very likely switching will cause interoperability issues when OpenDDS 3 and OpenDDS 4 based applications want to use shared memory between them on the same host

I would not want to guarantee expect 3-to-4 interoperability for shmem.  It seems like that would be a very rare scenario to have two different applications in the same domain on the same host using different major versions of the libraries.

jwillemsen:
> Please prepare a [news fragment](https://opendds.readthedocs.io/en/master/internal/docs.html#news) for this PR.

Done

mitza-oci:
Would a user of the shmem transport be impacted by the bug fixed in https://github.com/DOCGroup/ACE_TAO/pull/2077?  Or is it an independent issue since the ACE_Shared_Memory_Pool created by the transport doesn't use multiple segments?

jwillemsen:
For a 100% fix the user should use an ACE version with the proposed changes there, shmdt should be called

mitza-oci:
> For a 100% fix the user should use an ACE version with the proposed changes there, shmdt should be called

OK, please add that to the NEWS

jwillemsen:
Done

jrw972 (approved):

docs/news.d/shmem.rst
```diff
@@ -0,0 +1,6 @@
+.. news-prs: 4171
+.. news-start-section: Fixes
+- Fixed shared memory transport
+
+  - Fix the leaked shared memory by the shared memory transport. For a 100% fix also a new ACE version must be used.
```
jrw972, line 5:
```suggestion
  - Fix the leaked shared memory by the shared memory transport. For a 100% fix, a new ACE version including https://github.com/DOCGroup/ACE_TAO/pull/2077 must be used.
```

PR #4187, opened 2023-06-30
Multiple SpdpSendAddrs line support in ini files

smnrgrs:
Add support for multiple rtps.ini file lines specifying SpdpSendAddrs. In large systems on networks where Multicast discovery is impossible, it would be nice to be able to split the  SpdpSendAddrs entries across multiple lines.

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscoveryConfig.h
+++ b/dds/DCPS/RTPS/RtpsDiscoveryConfig.h
@@ -418,7 +418,7 @@ public:
   void spdp_send_addrs(const AddrVec& addrs)
   {
     ACE_GUARD(ACE_Thread_Mutex, g, lock_);
-    spdp_send_addrs_ = addrs;
+    spdp_send_addrs_.insert(spdp_send_addrs_.end(), addrs.begin(), addrs.end());
   }
 
   DCPS::TimeDuration max_auth_time() const


dds/DCPS/RTPS/RtpsDiscoveryConfig.h
```diff
@@ -418,7 +418,7 @@ class OpenDDS_Rtps_Export RtpsDiscoveryConfig : public OpenDDS::DCPS::RcObject {
   void spdp_send_addrs(const AddrVec& addrs)
   {
     ACE_GUARD(ACE_Thread_Mutex, g, lock_);
-    spdp_send_addrs_ = addrs;
+    spdp_send_addrs_.insert(spdp_send_addrs_.end(), addrs.begin(), addrs.end());
```
jrw972, line 421: This will break existing usage patterns as users are expecting set instead of insert semantics.

jrw972 (commented):
This will break existing usage patterns as users are expecting set instead of insert semantics.

jrw972:
Please update `tests/DCPS/ConfigFile/ConfigFile.cpp` to demonstrate that your idea works as you expect.  If it does not, then you may want to consider alternatives like:

1. Having an external file with all of the addresses that are parsed and set using the existing API.
2. Extending `SpdpSendAddrs` to take the name of such a file.

smnrgrs:
> This will break existing usage patterns as users are expecting set instead of insert semantics.

So are you saying that RtpsDiscoveryConfig::spdp_send_addrs is not just used when parsing the ini file. Users might be calling this directly and so would expect the existing 'set' functionality?

PR #4197, opened 2023-07-11
RTPS Parameter parsing erroneously succeeds when size is 0

jrw972:
Problem
-------

The specialized RTPS Parameter parsing code in `marshal_generator.cpp` accepts parameters that declare a zero size by setting the discriminator in the union and returning successfully.  If the discriminator is for a declared case of the union, then this creates a problem since the corresponding member of the union is not initialized.  This is especially problematic for members that require dynamic allocation like strings and sequences since the code accessing the Parameter will see a null pointer.

Solution
--------

Remove the special handling for size==0.  This causes known cases to be parsed.  If size==0 and the discriminator is a known case, then the parse will fail.


mitza-oci (approved):

tests/DCPS/RtpsMessages/RtpsMessagesTest.cpp
```diff
@@ -470,6 +470,22 @@ bool test_messages()
...
+      0x59, 0x00, // id: PID_PROPERTY_LIST
+      0x00, 0x00, // length: 0 (this is a problem)
+    };
+    ACE_Message_Block mb(reinterpret_cast<const char*>(input),
+                         sizeof(input));
+    mb.wr_ptr(sizeof(input));
+    Serializer ser(&mb, le_encoding);
+    // This should fail.
+    if (ser >> param) {
+      ok = false;
```
iguessthislldo, line 485: I think the cases above print something when they fail, this should probably do the same.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4201, opened 2023-07-14
Update subject name parsing

sonndinh:


dds/DCPS/security/SSL/SubjectName.cpp
```diff
@@ -145,28 +139,204 @@ namespace SSL {
...
+    while (start_pos < in_.size() && !got_value) {
+      std::string::size_type delimiter_pos = in_.find_first_of(",+", start_pos);
+      if (delimiter_pos == std::string::npos) {
+        av = in_.substr(pos_, in_.size() - pos_);
+        pos_ = in_.size();
+        got_value = true;
+      } else {
+        std::string::size_type tmp_pos = delimiter_pos - 1;
+        size_t esc_count = 0;
+        while (tmp_pos >= 0 && in_[tmp_pos--] == '\\') {
```
iguessthislldo, line 286:
```suggestion
        while (tmp_pos > 0 && in_[tmp_pos--] == '\\') {
```

dds/DCPS/security/SSL/SubjectName.h
```diff
@@ -10,17 +10,50 @@
...
+
+  // Current position in the input string.
+  std::string::size_type pos_;
+};
+
 class OpenDDS_Security_Export SubjectName {
- private:
-    typedef std::map<std::string, std::string> AttrMap;
+private:
+  typedef std::vector<std::pair<std::string, std::string> > AttrVec;
```
jrw972, line 56:
```suggestion
  typedef std::vector<std::map<std::string, std::string> > AttrVec;
```

See PLUS in https://datatracker.ietf.org/doc/html/rfc4514.

dds/DCPS/security/SSL/SubjectName.cpp
```diff
@@ -145,28 +139,204 @@ namespace SSL {
...
   bool SubjectName::operator==(const SubjectName& rhs) const
   {
-    bool result = (map_.size() == rhs.map_.size());
-    for (AttrMap::const_iterator i1 = map_.begin(), i2 = rhs.map_.begin();
-         result == true && i1 != map_.end() && i2 != rhs.map_.end();
-         ++i1, ++i2) {
-      if (i1->first.compare(i2->first) != 0 ||
-          i1->second.compare(i2->second) != 0) {
-        result = false;
+    if (attr_vec_.size() != rhs.attr_vec_.size()) {
```
jrw972, line 148: Can this should be simplified to `attr_vec_ == rhs.attr_vec_`?

dds/DCPS/security/SSL/SubjectName.cpp
```diff
@@ -145,28 +139,204 @@ namespace SSL {
...
+    }
+
+    unescape(av);
+    return validate_attribute_value(av);
+  }
+
+  void Parser::unescape(std::string& av) const
+  {
+    // Unescape the escaped characters from the attribute value.
+    replace_all(av, "\\\\", "\\");
```
jrw972, line 312: I think this line needs to be last.  Otherwise, you could be creating a new escape sequence.

jrw972 (changes requested):

dds/DCPS/security/SSL/SubjectName.cpp
sonndinh: Good catch!

dds/DCPS/security/SSL/SubjectName.cpp
sonndinh: Yes, I think that works.

dds/DCPS/security/SSL/SubjectName.cpp
sonndinh: I think this is fine because it's unescaping. But I've moved it to the last to be safe.

jrw972 (approved):

PR #4203, opened 2023-07-17
Building OpenDDS with CMake

iguessthislldo:


dds/DCPS/XTypes/DynamicTypeSupport.cpp
```diff
@@ -208,6 +208,31 @@ namespace OpenDDS {
...
+    template <>
+    DDS::ReturnCode_t DataReaderImpl_T<XTypes::DynamicSample>::read_generic(
+      GenericBundle&,
+      DDS::SampleStateMask,
+      DDS::ViewStateMask,
+      DDS::InstanceStateMask,
+      bool)
+    {
+      return DDS::RETCODE_UNSUPPORTED;
+    }
```
iguessthislldo, line 222: @mitza-oci I moved this and `DataReaderImpl_T<XTypes::DynamicSample>::take` here from `DynamicDataReaderImpl.cpp` after getting this very unhelpful link error while fixing a different issue with `scrap_vs.pl` in the static CMake Windows CI build (`build_cmake_w22_stat`):

```
OpenDDS_Dcps.lib(unity_13_cxx.obj) : error LNK2019: unresolved external symbol "public: virtual int __cdecl OpenDDS::DCPS::DataReade
rImpl_T<class OpenDDS::XTypes::DynamicSample>::read_generic(struct OpenDDS::DCPS::DataReaderImpl::GenericBundle &,unsigned int,unsig
ned int,unsigned int,bool)" (?read_generic@?$DataReaderImpl_T@VDynamicSample@XTypes@OpenDDS@@@DCPS@OpenDDS@@UEAAHAEAUGenericBundle@D
ataReaderImpl@23@III_N@Z) referenced in function "[thunk]:public: virtual int __cdecl OpenDDS::DCPS::DataReaderImpl_T<class OpenDDS:
:XTypes::DynamicSample>::read_generic`vtordisp{4294967292,8}' (struct OpenDDS::DCPS::DataReaderImpl::GenericBundle &,unsigned int,un
signed int,unsigned int,bool)" (?read_generic@?$DataReaderImpl_T@VDynamicSample@XTypes@OpenDDS@@@DCPS@OpenDDS@@$4PPPPPPPM@7EAAHAEAUG
enericBundle@DataReaderImpl@23@III_N@Z) [C:\Users\Fred\oci\w19_re_o1p1_sec\OpenDDS\build\DevGuideExamples\DCPS\Messenger\subscriber.
vcxproj]
OpenDDS_Dcps.lib(unity_13_cxx.obj) : error LNK2019: unresolved external symbol "public: virtual int __cdecl OpenDDS::DCPS::DataReade
rImpl_T<class OpenDDS::XTypes::DynamicSample>::take(class OpenDDS::DCPS::AbstractSamples &,unsigned int,unsigned int,unsigned int)"
(?take@?$DataReaderImpl_T@VDynamicSample@XTypes@OpenDDS@@@DCPS@OpenDDS@@UEAAHAEAVAbstractSamples@23@III@Z) referenced in function "[
thunk]:public: virtual int __cdecl OpenDDS::DCPS::DataReaderImpl_T<class OpenDDS::XTypes::DynamicSample>::take`vtordisp{4294967292,8
}' (class OpenDDS::DCPS::AbstractSamples &,unsigned int,unsigned int,unsigned int)" (?take@?$DataReaderImpl_T@VDynamicSample@XTypes@
OpenDDS@@@DCPS@OpenDDS@@$4PPPPPPPM@7EAAHAEAVAbstractSamples@23@III@Z) [C:\Users\Fred\oci\w19_re_o1p1_sec\OpenDDS\build\DevGuideExamp
les\DCPS\Messenger\subscriber.vcxproj]
C:\Users\Fred\oci\w19_re_o1p1_sec\OpenDDS\build\DevGuideExamples\DCPS\Messenger\Debug\subscriber.exe : fatal error LNK1120: 2 unreso
lved externals [C:\Users\Fred\oci\w19_re_o1p1_sec\OpenDDS\build\DevGuideExamples\DCPS\Messenger\subscriber.vcxproj]
```

This was mostly a shot in the dark, but because this was a unity build I looked and saw `DynamicDataReaderImpl.cpp` was in `unity_12_cxx.cxx` and not `unity_13_cxx` mentioned in the error. In `unity_13_cxx.cxx`, I think `DynamicTypeSupport.cpp` is the only file that uses `DynamicDataReader` so I moved the functions here and that solved the issue. I'm not entirely sure I understand the problem through because normally in a non-unity build they are not in the same translation unit, so this shouldn't matter as far as I know. If you have any alternative suggestions I'd be happy to try them out.

mitza-oci: I'm not sure, what do I need to do to enable unity build?
In general I'd rather turn unity build off for now

dds/CMakeLists.txt
```diff
@@ -0,0 +1,635 @@
...
+    CorbaSeq/WStringSeq.idl
+  SKIP_TAO_IDL
+  OPENDDS_IDL_OPTIONS -SI -Wb,tao_include_prefix=tao/
+  ${opendds_target_sources_args}
+)
+
+opendds_target_sources(OpenDDS_Dcps
+  PUBLIC
+    DdsDcps.idl
+    DdsDcpsConditionSeq.idl # TODO
```
mitza-oci, line 596: TODOs?

dds/DCPS/Service_Participant.cpp
```diff
@@ -431,6 +431,15 @@ Service_Participant::get_domain_participant_factory(int &argc,
         }
 
       } else {
+        // Convenient way to run tests in a different place from ini files.
+        const char* const config_dir = ACE_OS::getenv("OPENDDS_TEST_CONFIG_DIR");
```
mitza-oci, line 435:
```suggestion
        const char* const config_dir = ACE_OS::getenv("OPENDDS_CONFIG_DIR");
```
Why not make it more general and add to devguide?

dds/DCPS/security/Utility.h
```diff
@@ -1,32 +1,28 @@
...
+#  pragma once
+#endif
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
 namespace OpenDDS {
 namespace Security {
 
-class OpenDDS_Security_Export Utility : public DCPS::RcObject {
+class OpenDDS_Dcps_Export Utility : public DCPS::RcObject {
```
mitza-oci, line 25: Is this right? Is it moving from OpenDDS_Security to OpenDDS_Dcps?

dds/DCPS/transport/multicast/ReliableSession.cpp
```diff
@@ -26,8 +26,8 @@ namespace OpenDDS {
 namespace DCPS {
 
 namespace {
-  const Encoding::Kind encoding_kind = Encoding::KIND_UNALIGNED_CDR;
-  const Encoding encoding_unaligned_native(encoding_kind);
+  const Encoding::Kind reliable_session_encoding_kind = Encoding::KIND_UNALIGNED_CDR;
```
mitza-oci, line 29: Unity build strikes again?  Do we need something in dev guidelines?

dds/idl/be_produce.cpp
```diff
@@ -271,6 +271,22 @@ void postprocess(const char* fn, ostringstream& content,
...
+      prefix[i] = tolower(prefix[i]);
+    }
+    out <<
+      "\n"
+      "#ifdef OPENDDS_IDL_FILE_SPECIFIC\n"
+      "#  undef OPENDDS_IDL_FILE_SPECIFIC\n"
+      "#endif\n"
+      "#define OPENDDS_IDL_FILE_SPECIFIC(base, index) " << prefix << "##_##base##index\n"
+      "\n";
+  }
```
mitza-oci, line 288: Instead of generating code for the preprocessor to determine the name, how about generating that name directly in the be_*?

docs/devguide/building/cmake.rst
```diff
@@ -560,7 +579,7 @@ Functions
     opendds_export_header(<target> [USE_EXPORT_VAR <use-export-var-name>])
 
   Generates a header that is compatible with `ACE's generate_export_file.pl <https://github.com/DOCGroup/ACE_TAO/blob/master/ACE/bin/generate_export_file.pl>`__ for exporting symbols in shared libraries.
-  The header will able to be included as ``<target>_export.h`` and macro will be to use on symbols to will be ``<target>_Export``.
+  The header will able to be included as ``<target>_export.h`` and the macro that can be used to export symbols wll be named ``<target>_Export``.
```
mitza-oci, line 582:
```suggestion
  The header will able to be included as ``<target>_export.h`` and the macro that can be used to export symbols will be named ``<target>_Export``.
```

docs/devguide/building/index.rst
```diff
@@ -527,3 +529,171 @@ Users of OpenDDS are strongly encouraged to select one of the two options listed
...
+  4 was used as an example for the max number of parallel jobs.
+  If unsure what this number should be, use the number of CPU cores on the machine.
+  This can be combined with unity builds.
+
+.. tab:: Windows
+
+  .. code-block::
+
+    cmake -B build -DCMAKE_UNITY_BUILD=TRUE
+    cmake --build build
```
mitza-oci, line 568: These are very similar.  Could they be refactored to just show the commands with an optional `[-- <build tool args>]` part at the end?

docs/devguide/building/index.rst
```diff
@@ -527,3 +529,171 @@ Users of OpenDDS are strongly encouraged to select one of the two options listed
...
+=======================
+
+- The following features are planned, but not implemented yet:
+
+  - The ability to use MPC for building user applications.
+  - Safety profile
+  - Java Mapping
+
+- CMake-build OpenDDS libraries and executables will currently be ignored by :ref:`find_package(OpenDDS COMPONENTS ...) <cmake-components>`.
+  Passing feature should work.
```
mitza-oci, line 699: I'm not sure what this line means

iguessthislldo: Besides it should saying "features", I might have worded it poorly. Here is the full explanation:

In an MPC-built OpenDDS [features, libraries, and executables can be passed to find_package to require that they are enabled/disable/available](https://opendds.readthedocs.io/en/master/devguide/building/cmake.html#components). For example `find_package(OPENDDS REQUIRED safety_profile=OFF OpenDDS::Security OpenDDS::RtpsRelay)` will require that safety profile is disabled and the relay and security library can be used if OpenDDS is found. This also works within a CMake-build OpenDDS.

Once the CMake-built OpenDDS is installed though a different target direct importing mechanism has to be used, which is decided [here](https://github.com/iguessthislldo/OpenDDS/blob/d1cdb48a76bcefca6702fccda4332b791998477a/cmake/OpenDDSConfig.cmake#L222-L234). I haven't worked out exactly how the installed mechanism interacts with `find_package` yet. Theoretically it shouldn't be too hard, just check if the target exists, but I just haven't gotten around to it. ACE/TAO targets always work because they're always imported indirectly, which I should probably mention. Features also always work because these are just variables and don't really have much to do with the target importing mechanisms.

dds/CMakeLists.txt
iguessthislldo: I think these are safety profile TODOs. I will double check this and note that if that's the case.

cmake/OpenDDSConfig.cmake
```diff
@@ -219,39 +219,56 @@ if(_opendds_tao_find_libs OR _opendds_tao_find_exes)
...
+      set(_found_all FALSE)
+      break()
+    endif()
+  endforeach()
+
+  if(_found_all)
+    add_library(OpenDDS::OpenDDS INTERFACE IMPORTED)
+    set_target_properties(OpenDDS::OpenDDS
+      PROPERTIES
+        INTERFACE_LINK_LIBRARIES "${_opendds_core_libs}")
```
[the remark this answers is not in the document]
iguessthislldo: This has always been the behavior of `OpenDDS::OpenDSS`. I've kept it as is mostly for backwards compatibity, but I'm not found of it.

cmake/dds_idl_sources.cmake
```diff
@@ -136,7 +139,8 @@ function(_opendds_target_idl_sources target)
     get_property(generated_dependencies SOURCE ${idl_file}
       PROPERTY OPENDDS_IDL_GENERATED_DEPENDENCIES SET)
 
-    if(generated_dependencies)
+    # TODO Fix this
```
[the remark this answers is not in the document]
iguessthislldo: I think this has to be left for later. Enabling this again breaks the OpenDDS build because of [these two `opendds_target_sources` calls in the Rtps CMakeLists](https://github.com/iguessthislldo/OpenDDS/blob/43f0679b10cb66981cfe74125251bee76ac42f3d/dds/DCPS/RTPS/CMakeLists.txt#L52-L69). This should be valid, but if this check wasn't disabled then it only runs the `tao_idl`-only. I think this is an attempt to keep the IDL compilers from running more than once in cases like here:

https://github.com/OpenDDS/OpenDDS/blob/eb011d68c99cf6096465bb34f3539eda9a5a6ac7/tests/cmake/Messenger/Messenger_1/CMakeLists.txt#L39-L50

However the caller could pass different options like in RTPS and that should be valid.  This doesn't completely fix that issue though if there are conflicting options on two different targets. That would require per-target isolated generation which the `opendds_target_syntax` implies, but doesn't do.

I will update the comment with this information.

dds/idl/be_produce.cpp
iguessthislldo: I think tried that first, but went with this because it seemed slightly cleaner for the code in opendds_idl.

cmake/build_ace_tao.cmake
```diff
@@ -0,0 +1,238 @@
...
+include(FetchContent)
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.24)
+  cmake_policy(SET CMP0135 NEW)
+endif()
+
+set(ACE_IS_BEING_BUILT MPC CACHE INTERNAL "")
+set(TAO_IS_BEING_BUILT MPC CACHE INTERNAL "")
+set(url_base "https://github.com/DOCGroup/ACE_TAO/releases/download/")
+set(ace_ver "7.1.1")
```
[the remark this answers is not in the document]
iguessthislldo: I moved it so they are next to each other. I guess updating them could be automated, but I'm not sure what that looks like right now.

docs/devguide/building/index.rst
iguessthislldo: They could, but I don't really want to force someone to have to try to piece together what the right commands are, especially if they're not familiar with them.

docs/devguide/building/index.rst
iguessthislldo: I have implemented the check, so this note is no longer necessary.

dds/DCPS/security/Utility.h
mitza-oci: That seems odd... static builds don't have `Export` macros expand to anything.  Does this mean the CMake static build is significantly different from the MPC static build?

dds/idl/be_produce.cpp
mitza-oci: Maybe at some point (not this PR) we'll move it to a common header so it can be used in other places.

docs/devguide/building/index.rst
mitza-oci: Just a philosophical point about a "cross-platform make" (CMake).  We could also leave out the parallel build part and let that be an exercise for the reader.

dds/idl/be_produce.cpp
iguessthislldo: Another option is [`UNITY_BUILD_UNIQUE_ID`](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD_UNIQUE_ID.html#prop_tgt:UNITY_BUILD_UNIQUE_ID).

dds/DCPS/security/Utility.h
iguessthislldo: Yes, it's very odd. For the most part the CMake should follow MPC's lead, but the new automatic [`StaticIncludes.h`](https://github.com/iguessthislldo/OpenDDS/blob/igtd/cmake-build/dds/DCPS/StaticIncludes.h) is probably the biggest difference that's relevant. It might have a hand in this since it looks like I did this not too long after. I will try to look at it in detail again tomorrow and see if I can see exactly why this was apparently necessary and if there's a better solution.

mitza-oci: The command line you copied above doesn't link to libOpenDDS_Security.  So with the new code in StaticIncludes, the application has an undefined symbol reference to something in libOpenDDS_Security.

dds/InfoRepo/DCPSInfo_i.cpp
```diff
@@ -2260,6 +2260,13 @@ int TAO_DDS_DCPSInfo_i::init_transport(int listen_address_given,
     OpenDDS::DCPS::TransportInst_rch inst =
       OpenDDS::DCPS::TransportRegistry::instance()->create_inst(inst_name,
                                                                "tcp");
+    if(!inst) {
```
mitza-oci, line 2263:
```suggestion
    if (!inst) {
```

dds/DCPS/security/Utility.h
iguessthislldo: Yes, it shouldn't have the initializer in the first place. I think I figured it out though:

- There's probably many ways to get there depending on include order, but somehow [`security/Utility.h` gets included in `security/framework/SecurityConfig.h`](https://github.com/iguessthislldo/OpenDDS/blob/d2767ceb20bf8a6c0396195df6829f23d9ddfc06/dds/DCPS/security/framework/SecurityConfig.h#L22).
- `Utility.h` includes the security export header which sets [`OPENDDS_SECURITY_HAS_DLL` to 0](https://github.com/iguessthislldo/OpenDDS/blob/d2767ceb20bf8a6c0396195df6829f23d9ddfc06/dds/DCPS/security/OpenDDS_Security_Export.h#L12).
- If `StaticIncludes.h` is included first after that, this looks the same as if someone linked the security library ([this is where that's done in CMake](https://github.com/iguessthislldo/OpenDDS/blob/d2767ceb20bf8a6c0396195df6829f23d9ddfc06/cmake/opendds_build_helpers.cmake#L50-L52)), so [it includes the initialization header](https://github.com/iguessthislldo/OpenDDS/blob/d2767ceb20bf8a6c0396195df6829f23d9ddfc06/dds/DCPS/StaticIncludes.h#L38-L40).

This had to be what I figured out last time. I chose to reuse the `*_HAS_DLL` to control `StaticIncludes` because it would also allow overriding `ACE_AS_STATIC_LIBS` in the individual export files without any changes to them. However I guess that depended on there not being a publicly visible include of an export header for a library that's not linked.

Having `Utility` be part of Security really doesn't make any sense as it's part of the interface offered by the security framework. Otherwise there's not really a much of a point for there to be a separate `Utility` and `UtilityImpl` classes. So I do think `Utility` should be moved from Security to Dcps, except more properly. Looking at the commit again where I did this I think it was a quick fix and I meant to follow up on it and get feedback, but forgot about it.

mitza-oci (approved):
👍 👍 👍 
I left a few of the review comments unresolved since they may be things we come back to later

cmake/dds_idl_sources.cmake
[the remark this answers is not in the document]
iguessthislldo: #4306

dds/DCPS/XTypes/DynamicTypeSupport.cpp
iguessthislldo: #4308

dds/idl/be_produce.cpp
iguessthislldo: #4307

PR #4208, opened 2023-07-19
Fix build_u22_stat_qs_ws_sec build

jrw972:


.github/workflows/build_and_test.yml
```diff
@@ -5406,6 +5329,12 @@ jobs:
     steps:
     - name: install xerces
       run: sudo apt-get -y install libxerces-c-dev
+    - name: update apt
+      run: sudo apt-get update
```
mitza-oci, line 5333: Move this up ahead of installing xerces-c.
Consider combining the install steps to potentially save time.

PR #4223, opened 2023-08-04
[fix] WaitSet not notified when DataWriter liveliness lost

jmmorato:


tests/unit-tests/dds/DCPS/WaitSet.cpp
```diff
@@ -7,8 +7,11 @@
 
 #include <gtest/gtest.h>
 
+#include <dds/DCPS/Service_Participant.h>
```
jrw972, line 10: As this test creates particpants, waits, etc., move it to `/tests/DCPS` and add it to the `*.lst` files as appropriate.

jrw972 (changes requested):

tests/unit-tests/dds/DCPS/WaitSet.cpp
jmmorato: Test moved to `/tests/DCPS/WaitSetLivelinessLost`

jrw972 (approved):

PR #4226, opened 2023-08-08
Support sending `DynamicDataAdapter` sample via dynamic writer

sonndinh:
- Move the internal XCDR serialization functions for `DynamicDataImpl` out of its `DataContainer`. `DataContainer` now is just a container of the sample data which would facilitate serializing to other formats.
- Add new interfaces for XCDR serialization to `DynamicDataBase` that allow dynamic writers to send dynamic samples constructed from `DynamicDataAdapter` in addition to `DynamicDataImpl`.

### DIFF
--- a/dds/DCPS/XTypes/DynamicDataImpl.cpp
+++ b/dds/DCPS/XTypes/DynamicDataImpl.cpp
@@ -333,9 +333,7 @@ DDS::ReturnCode_t DynamicDataImpl::clear_all_values()
 
 void DynamicDataImpl::clear_container()
 {
-  container_.single_map_.clear();
-  container_.sequence_map_.clear();
-  container_.complex_map_.clear();
+  container_.clear();
 }
 
 DDS::ReturnCode_t DynamicDataImpl::clear_nonkey_values()
@@ -4150,7 +4148,7 @@ bool DynamicDataImpl::serialized_size_string(const DCPS::Encoding& encoding, siz
   const bool is_empty = container_.single_map_.empty() && container_.complex_map_.empty();
   if (!is_empty) {
     CORBA::ULong largest_index;
-    if (!get_largest_index_basic(largest_index)) {
+    if (!container_.get_largest_index_basic(largest_index)) {
       return false;
     }
     primitive_serialized_size_ulong(encoding, size);
@@ -4206,7 +4204,7 @@ bool DynamicDataImpl::serialized_size_wstring(const DCPS::Encoding& encoding, si
   const bool is_empty = container_.single_map_.empty() && container_.complex_map_.empty();
   if (!is_empty) {
     CORBA::ULong largest_index;
-    if (!get_largest_index_basic(largest_index)) {
+    if (!container_.get_largest_index_basic(largest_index)) {
       return false;
     }
     primitive_serialized_size_ulong(encoding, size);
@@ -5997,7 +5995,7 @@ bool DynamicDataImpl::get_index_to_id_from_complex(IndexToIdMap& index_to_id,
   CORBA::ULong length = 0;
   if (!complex_map_.empty()) {
     CORBA::ULong largest_index;
-    if (!get_largest_complex_index(largest_index)) {
+    if (!container_.get_largest_complex_index(largest_index)) {
       return false;
     }
     length = largest_index + 1;
@@ -6030,7 +6028,7 @@ bool DynamicDataImpl::serialized_size_sequence(const DCPS::Encoding& encoding,
     CORBA::ULong length = 0;
     if (!is_empty) {
       CORBA::ULong largest_index;
-      if (!get_largest_index_basic(largest_index)) {
+      if (!container_.get_largest_index_basic(largest_index)) {
         return false;
       }
       length = largest_index + 1;
@@ -6071,7 +6069,7 @@ bool DynamicDataImpl::serialized_size_sequence(const DCPS::Encoding& encoding,
       CORBA::ULong length = 0;
       if (!is_empty) {
         CORBA::ULong largest_index;
-        if (!get_largest_index_basic_sequence(largest_index)) {
+        if (!container_.get_largest_index_basic_sequence(largest_index)) {
           return false;
         }
         length = largest_index + 1;
@@ -6114,7 +6112,7 @@ bool DynamicDataImpl::serialize_sequence(DCPS::Serializer& ser, DCPS::Sample::Ex
     CORBA::ULong length = 0;
     if (!is_empty) {
       CORBA::ULong largest_index;
-      if (!get_largest_index_basic(largest_index)) {
+      if (!container_.get_largest_index_basic(largest_index)) {
         return false;
       }
       length = largest_index + 1;
@@ -6145,7 +6143,7 @@ bool DynamicDataImpl::serialize_sequence(DCPS::Serializer& ser, DCPS::Sample::Ex
       CORBA::ULong length = 0;
       if (!is_empty) {
         CORBA::ULong largest_index;
-        if (!get_largest_index_basic_sequence(largest_index)) {
+        if (!container_.get_largest_index_basic_sequence(largest_index)) {
           return false;
         }
         length = largest_index + 1;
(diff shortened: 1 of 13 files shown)


jrw972 (changes requested):
Keep going.  We intended to move all serialization-related code out of the container _and_ the DynamicDataImpl.  The "to_xcdr" function should take a DynamicData, Serializer, and whatever else is necessary.  With a free function, the same code can serialize either a DynamicDataImpl or a DynamicDataAdapter.

     @iguessthislldo

sonndinh:
> Keep going. We intended to move all serialization-related code out of the container _and_ the DynamicDataImpl. The "to_xcdr" function should take a DynamicData, Serializer, and whatever else is necessary. With a free function, the same code can serialize either a DynamicDataImpl or a DynamicDataAdapter.
> 
> ```
>  @iguessthislldo 
> ```

The serialization code for DynamicDataImpl was written to work specifically with the data container in that class. If we move it out of DynamicDataImpl, we'll need to provide a way to access the data container, e.g. via a public getter, but I think we don't want to let end user uses that interface too. Another way I can think of is making the "to_xcdr" function a friend of DynamicDataImpl. But this approach seems a bit clunky to me since many of the serialization functions called by it also need access to the data container as well.

It doesn't look like we can reuse the same serialization code of DynamicDataImpl for DynamicDataAdapter. Since the internal data of DynamicDataAdapter is a C++ value of the corresponding type, can we just use the generated `operator<<` from type support for it?

dds/DCPS/XTypes/DynamicDataImpl.h
```diff
@@ -971,18 +628,15 @@ class OpenDDS_Dcps_Export DynamicDataImpl : public DynamicDataBase {
   };
 
   // Copy a value of a basic member from single map to a DynamicData object.
-  bool move_single_to_complex(const DataContainer::const_single_iterator& it,
-                              DynamicDataImpl* data);
-  bool move_single_to_complex_i(const DataContainer::const_single_iterator& it,
-                                DynamicDataImpl* data, const TypeKind treat_as);
+  bool move_single_to_complex(const const_single_iterator& it, DynamicDataImpl* data);
+  bool move_single_to_complex_i(const const_single_iterator& it, DynamicDataImpl* data,
+                                const TypeKind treat_as);
```
mitza-oci, line 633:
```suggestion
                                TypeKind treat_as);
```

dds/idl/dynamic_data_adapter_generator.cpp
```diff
@@ -450,26 +454,133 @@ namespace {
...
 
+      be_global->header_ <<
+        "  DDS::DynamicData_ptr clone();\n"
+        "  bool serialized_size(const OpenDDS::DCPS::Encoding& enc, size_t& size, OpenDDS::DCPS::Sample::Extent ext) const;\n"
+        "  bool serialize(OpenDDS::DCPS::Serializer& ser, OpenDDS::DCPS::Sample::Extent ext) const;\n"
+        "\n";
+
+      const bool is_topic_type = be_global->is_topic_type(node);
+      const bool forany = needs_forany(dynamic_cast<AST_Type*>(node));
+      const bool distinct_type = needs_distinct_type(dynamic_cast<AST_Type*>(node));
```
mitza-oci, line 465: Could this do the cast just once, or change the `needs_` functions to take ACE_Decl?

dds/idl/dynamic_data_adapter_generator.cpp
```diff
@@ -450,26 +454,133 @@ namespace {
...
         "};\n"
         "\n";
     }
 
     {
       NoSafetyProfileGuard nspg;
-
-      wrapper.generate_tag();
+      if (!generate) {
+        wrapper.generate_tag();
```
mitza-oci, line 582: It's seem counter-intuitive that when generate is false this calls a generate_* function.

dds/idl/dynamic_data_adapter_generator.cpp
```diff
@@ -400,12 +400,16 @@ namespace {
     }
 
     if (generate) {
+      NoSafetyProfileGuard nspg;
+      wrapper.generate_tag();
```
iguessthislldo, line 404: This is already generated below.

dds/idl/dynamic_data_adapter_generator.cpp
```diff
@@ -400,12 +400,16 @@ namespace {
...
       DynamicDataAdapterGuard ddag;
 
-      be_global->impl_ <<
+      std::string impl_classname = "DynamicDataAdapterImpl<" + cpp_name + tag + " >";
+
+      be_global->header_ <<
         "template <>\n"
-        "class DynamicDataAdapterImpl<" << cpp_name << tag << " > : public DynamicDataAdapter_T<"
-          << cpp_name << "> {\n"
+        "class " << impl_classname << " : public DynamicDataAdapter_T<" << cpp_name << "> {\n"
```
iguessthislldo, line 412: I explicitly avoided putting these in the headers because almost every single struct and union gets these and everything put in headers slows down builds.

dds/idl/dynamic_data_adapter_generator.cpp
```diff
@@ -450,26 +454,133 @@ namespace {
           "\n";
       }
 
+      be_global->header_ <<
+        "  DDS::DynamicData_ptr clone();\n"
+        "  bool serialized_size(const OpenDDS::DCPS::Encoding& enc, size_t& size, OpenDDS::DCPS::Sample::Extent ext) const;\n"
+        "  bool serialize(OpenDDS::DCPS::Serializer& ser, OpenDDS::DCPS::Sample::Extent ext) const;\n"
+        "\n";
+
+      const bool is_topic_type = be_global->is_topic_type(node);
```
iguessthislldo, line 463: A lot of the code below is can be simplified by making full use of `RefWrapper`. Here's one example:

https://github.com/OpenDDS/OpenDDS/blob/4d1d6a37ed7d86978600edf631e7bb9f1772a5aa/dds/idl/marshal_generator.cpp#L1607-L1612

It will make some decisions based on the type and others based on traits set before calling `done`.

iguessthislldo (changes requested):

dds/idl/dynamic_data_adapter_generator.cpp
sonndinh: I updated to use cast only once.

dds/idl/dynamic_data_adapter_generator.cpp
sonndinh: Refactored to merge this block with the `if (generate)` block before this. It will make this clearer.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -3266,7 +3266,7 @@ bool DynamicDataImpl::move_single_to_complex(const const_single_iterator& it, Dy
 }
 
 bool DynamicDataImpl::move_single_to_complex_i(const const_single_iterator& it,
-                                               DynamicDataImpl* data, const TypeKind treat_as)
+                                               DynamicDataImpl* data, TypeKind treat_as)
```
mitza-oci, line 3269: The function definition can use `const` here, just not in the header.  I think there's a note on this in the gudelines, and if not then we can add one.

sonndinh: The guidelines have a related guide (https://opendds.readthedocs.io/en/latest-release/internal/dev_guidelines.html#language-usage, point 6th) but it doesn't explicitly say when to use `const`.

jrw972 (approved):

dds/idl/dynamic_data_adapter_generator.cpp
sonndinh: `RefWrapper` is used for types that need `IDL::DistinctType`. The names with `_forany` suffix are manually constructed because `RefWrapper` will generate references (I think this is needed by marshal_generator) while I need to get non-reference types to instantiate values for them. I think it's simple enough to construct those type names manually without needing to change `RefWrapper`.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: I updated the guidelines for this.

iguessthislldo (approved):

PR #4230, opened 2023-08-11
ICE Agent doesn't expose local agent/endpoint info topic

jrw972:
Problem
-------

1. The ICE Agent has bugs and is exceedingly difficult to test.

2. Password changes cause publication/subscription messages to be sent which causes unnecessary overhead.

The long-term solution to this problems is to refactor the ICE Agent and test it mercilessly.

Solution
--------

The first step toward this solution (this commit) is to replace the AgentInfoListener interface with internal DDS topics.  The next step is to introduce SEDP endpoints for ICE to exchange the agent/endpoint info and feed this information to the ICE agent using internal DDS topics.


docs/news.d/ice.rst
```diff
@@ -0,0 +1,6 @@
+.. news-prs: 4230
+
+.. news-start-section: Fixes
+- ICE doesn't use IPV6-map IPV4 addresses.
```
iguessthislldo, line 4:
```suggestion
- ICE doesn't use IPv6-mapped IPv4 addresses anymore.
```

The ways it's currently phrased as a fact without any relation to how it worked before sounded a little weird to me.

dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1639,18 +1639,13 @@ bool from_param_list(const ParameterList& param_list,
     case PID_OPENDDS_ICE_CANDIDATE: {
       const IceCandidate_t& ice_candidate = parameter.ice_candidate();
       ICE::Candidate candidate;
-#if IPV6_V6ONLY
       // https://tools.ietf.org/html/rfc8445
 
       // IPv4-mapped IPv6 addresses SHOULD NOT be included in the
       // address candidates unless the application using ICE does not
       // support IPv4 (i.e., it is an IPv6-only application
       // [RFC4038]).
```
iguessthislldo, line 1647: Is this comment still accurate? It looks like it never uses mapped addresses now.

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4231, opened 2023-08-14
Remove -Grapidjson flag [opendds_idl]

Idan-Neeman:
The -Grapidjson flag has been removed.


iguessthislldo (approved):
For reference the RapidJSON generated code was removed in #3260. Thanks for noticing this was still being referenced.

PR #4238, opened 2023-08-22
Remove Unneeded Non-ASCII From DevGuide

iguessthislldo:
- ``` `` s``` -> ``` ``\s ```
- `‘` (Left Single Quotation Mark U+2018) and `’` (Right Single Quotation Mark U+2019) -> `'` (U+0027) or to code formating depending on the context.
- (Non-breaking space U+00A0) -> Space
- `–` (En Dash U+2013) and `—` (Em Dash U+2014) -> `--`
- Other misc. changes.


docs/devguide/run_time_configuration.rst
```diff
@@ -3483,7 +3485,7 @@ It can also be set in application code using:
 
 The *level* defaults to a value of 0 and has values of 0 to 10 as defined below:
 
-* 0 – debug logging is disabled
+* 0 -- debug logging is disabled
```
mitza-oci, line 3488:
```suggestion
* 0 - debug logging is disabled
```
inconsistent with the items below (so perhaps change those instead?)

mitza-oci (changes requested):

docs/devguide/run_time_configuration.rst
iguessthislldo: Many of these were just search and replace, but I didn't consider they'd be inconsistent, which I should've known.

mitza-oci (approved):

PR #4202, opened 2023-07-14
RtpsRelay: updated tracking of client network addresses

mitza-oci:

PR #4254, opened 2023-09-05
RtpsRelay: fixed bug in record_activity's use of remove in GuidAddrSet

mitza-oci:
Bug was introduced in PR #4202

### DIFF
--- a/tools/rtpsrelay/GuidAddrSet.cpp
+++ b/tools/rtpsrelay/GuidAddrSet.cpp
@@ -24,7 +24,7 @@ GuidAddrSet::record_activity(const Proxy& proxy,
 
   if (config_.restart_detection()) {
     Remote remote(remote_address.addr, src_guid);
-    const auto result = remote_map_.insert(std::make_pair(remote, src_guid));
+    auto result = remote_map_.insert(std::make_pair(remote, src_guid));
     if (result.second) {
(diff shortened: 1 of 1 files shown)


tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -36,18 +36,22 @@ GuidAddrSet::record_activity(const Proxy& proxy,
       }
       rtps_discovery_->remove_domain_participant(config_.application_domain(), config_.application_participant_guid(), result.first->second);
       const auto pos = guid_addr_set_map_.find(result.first->second);
+      const auto prev_guid = result.first->second;
       if (pos != guid_addr_set_map_.end()) {
-        remove(proxy, result.first->second, pos, now, &relay_participant_status_reporter_);
+        remove(proxy, prev_guid, pos, now, &relay_participant_status_reporter_);
+        result.first = remote_map_.find(remote); // remote_map_ was updated by remove()
+      }
+      if (result.first != remote_map_.end()) {
```
jrw972, line 44: I think this needs an else that inserts.  Alternatively, you could combine line 42-45 with something like
```
const auto result2 = remote_map_.insert(std::make_pair(remote, src_guid));
if (!result2.second) {
  result2.first->second = src_guid;
}
```

jrw972 (changes requested):

jrw972 (approved):

PR #4292, opened 2023-10-02
Initialize variables in TypeObject to silence warnings

Idan-Neeman:


jrw972 (approved):

jrw972:
@Idan-Neeman What compiler and settings are you using?  If it is one that is publicly available, we might consider adding it to the CI.

Idan-Neeman:
> @Idan-Neeman What compiler and settings are you using? If it is one that is publicly available, we might consider adding it to the CI.

I'm using Microsoft Visual C++ with the default settings. However, I load the class in another project that gives the warnings.

PR #4294, opened 2023-10-03
Build and install instructions are outside of DevGuide

jrw972:
Problem
-------

The build and install instruction are outside of the DevGuide proper. There is also duplication between the introduction and build documentation.

Solution
--------

Move "Building and Installing" after the introduction and move content from the introduction into "Building and Installing".


docs/devguide/building/dependencies.rst
```diff
@@ -47,6 +47,9 @@ It is included in the release archive of ACE/TAO, which is downloaded by the con
 ACE/TAO
 =======
 
+OpenDDS is compatible with the current DOC Group micro release in the ACE 6.x / TAO 2.x series.
+See the :ghfile:`README.md` file for details.
+
```
iguessthislldo, line 52:
```suggestion
```

I'm not sure what details about ACE/TAO version are in the README that are not here. This sections is already very detailed.

docs/devguide/building/index.rst
```diff
@@ -317,3 +471,62 @@ Tests
 Tests are not built by default, ``--tests`` must be passed to the configure script.
 All tests can be run using :ghfile:`tests/auto_run_tests.pl`.
 See :doc:`/internal/running_tests` for running all tests or individual tests.
+
+..
+    Sect<1.3>
```
iguessthislldo, line 476: These comments should come after an old section header. It's missing so I don't see a point in keeping the comment.

NEWS.md
```diff
@@ -26,19 +26,19 @@ Read [the documenation for this release on Read the Docs](https://opendds.readth
...
+        * Added [`opendds_target_sources(SKIP_TAO_IDL)`](https://opendds.readthedocs.io/en/dds-3.25/devguide/building/cmake.html#func-arg-opendds_target_sources-SKIP_TAO_IDL) to disable `tao_idl`. ([PR #4140](https://github.com/OpenDDS/OpenDDS/pull/4140))
+        * Added [`opendds_target_sources(SKIP_OPENDDS_IDL)`](https://opendds.readthedocs.io/en/dds-3.25/devguide/building/cmake.html#func-arg-opendds_target_sources-SKIP_OPENDDS_IDL) to disable `opendds_idl`. ([PR #4140](https://github.com/OpenDDS/OpenDDS/pull/4140))
+        * Added [`opendds_target_sources(USE_EXPORT)`](https://opendds.readthedocs.io/en/dds-3.25/devguide/building/cmake.html#func-arg-opendds_target_sources-USE_EXPORT) to allow overriding the generated export header with an existing one. ([PR #4160](https://github.com/OpenDDS/OpenDDS/pull/4160))
```
iguessthislldo, line 37: Nothing should change in this file because it's using fixed older versions.

iguessthislldo (changes requested):

docs/devguide/building/index.rst
jrw972: Should I remove all such comments?

iguessthislldo: Probably just this one here. The idea behind them is that we or someone else could find a section using an old devguide section number.  I also used it during the conversion to help compare sections. I'm not sure that's worth the cost of keeping these comments, but I think that should be a different discussion/PR.

iguessthislldo (approved):

PR #4310, opened 2023-10-16
Documentation for next release

jrw972:
I'm not sure that we need to include the GHA changes.

### DIFF
--- a/docs/news.d/config_store.rst
+++ b/docs/news.d/config_store.rst
@@ -1,13 +1,5 @@
-.. news-prs: 4162 4241 4242 4255 4243 4249
+.. news-prs: 4162 4241 4242 4255 4243 4249 4272 4270
+
 .. news-start-section: Additions
-- OpenDDS now stores ``rtps_udp`` transport configuration in the key-value store.
-
-- OpenDDS now stores ``multicast`` transport configuration in the key-value store.
-
-- OpenDDS now stores ``shmem`` transport configuration in the key-value store.
-
-- OpenDDS now stores ``tcp`` transport configuration in the key-value store.
-
-- OpenDDS now stores ``udp`` transport configuration in the key-value store.
-
+- Convert transport configurations (``rtps_udp``, ``multicast``, ``shmem``, ``tcp``, ``udp``) uses key-value store.
 .. news-end-section
--- a/docs/news.d/4216.rst
+++ b/docs/news.d/4216.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4216
+
+.. news-start-section: Fixes
+- Remove from TypeLookupService when remote endpoint is removed from SEDP
+.. news-end-section
--- a/docs/news.d/4226.rst
+++ b/docs/news.d/4226.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4226
+
+.. news-start-section: Additions
+- Support sending DynamicDataAdapter sample via DynamicDataWriter
+.. news-end-section
--- a/docs/news.d/4253.rst
+++ b/docs/news.d/4253.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4253
+
+.. news-start-section: Fixes
+- Efficiency: Remove per-element locking in JobQueue
+.. news-end-section
--- a/docs/news.d/4254.rst
+++ b/docs/news.d/4254.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4254
+
+.. news-start-section: Fixes
+- RtpsRelay: fixed bug in record_activity's use of remove in GuidAddrSet
+.. news-end-section
--- a/docs/news.d/4262.rst
+++ b/docs/news.d/4262.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4262
+
+.. news-start-section: Fixes
+- Fix warnings in typeobject_generator when using TAO 3
+.. news-end-section
--- a/docs/news.d/4265.rst
+++ b/docs/news.d/4265.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4265
+
+.. news-start-section: Fixes
+- Fix null pointer when participant is absent when updating locators
+.. news-end-section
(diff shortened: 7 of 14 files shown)


docs/news.d/config_store.rst
```diff
@@ -1,4 +1,4 @@
-.. news-prs: 4162 4241 4242 4255 4243 4249
+.. news-prs: 4162 4241 4242 4255 4243 4249 4272 7270
```
iguessthislldo, line 1: Do you mean 4270?

docs/news.d/devguide-improvments.rst
```diff
@@ -1,7 +1,16 @@
-.. news-prs: 4281
+.. news-prs: 4281 4294 4296 4279 4238 4234 4237 4231 4233 4210
```
iguessthislldo, line 1: To be honest, this looks weird in the news as each item is getting all of these PRs and this isn't really how I envisioned this being used. The idea was that each PR would have a news fragment for every set of related new-worthy changes, then subsequent PRs could tweak them. This was to try to mirror the way PRs worked. Do we need a way to declare multiple different PR sets per file as a convenience?

docs/news.d/gha.rst
```diff
@@ -0,0 +1,8 @@
+.. news-prs: 4299 4240 4208 4209
+
+.. news-start-section: Notes
+- GitHub Actions: use apt-get update for certain builds
```
iguessthislldo, line 4: This is probably debatable, I don't consider the CI to be news-worthy

docs/news.d/cmake-build.rst
```diff
@@ -1,4 +1,4 @@
-.. news-prs: 4203
+.. news-prs: 4203 4304
```
iguessthislldo, line 1:
```suggestion
.. news-prs: 4203
```

4304 is related to CMake, but it's a minor documentation change that I don't consider to be news-worthy.

docs/news.d/gha.rst
iguessthislldo, line 4: This is probably debatable, I don't consider the CI to be news-worthy

iguessthislldo (changes requested):

docs/news.d/devguide-improvments.rst
jrw972: I don't think so.  I can break them out as individual fragments.

docs/news.d/subjectname_parsing.rst
```diff
@@ -1,4 +1,4 @@
-.. news-prs: 4201
+.. news-prs: 4201 4220
```
sonndinh, line 1: Should we leave PR #4220 out since it was only to add this news fragment file?

docs/news.d/general.rst
```diff
@@ -0,0 +1,37 @@
+.. news-prs: 4298 4292 4262 4295 4226 4293 4253 4252 4250 4265 4254 4219 4216
+
+.. news-start-section: Additions
+- Support sending DynamicDataAdapter sample via DynamicDataWriter
+.. news-end-section
+
+.. news-start-section: Fixes
+- Improvements
```
sonndinh, line 8: Is this a relevant place to use subsections (https://opendds.readthedocs.io/en/master/internal/docs.html#docs-news-subsections)? Or maybe this does the same thing?

docs/news.d/general.rst
```diff
@@ -0,0 +1,37 @@
+.. news-prs: 4298 4292 4262 4295 4226 4293 4253 4252 4250 4265 4254 4219 4216
```
sonndinh, line 1: This fragment seems to collect changes to multiple aspects of OpenDDS. Should we break it up into smaller fragments?

docs/news.d/subjectname_parsing.rst
iguessthislldo: I would lean towards leaving it out, but I'm not sure this sort of situation been discussed before.

docs/news.d/general.rst
iguessthislldo: Subsections are similar to manually creating a list, but subsections can be shared by fragments, like they share "Additions", "Fixes", etc., while a RST list can not be shared. After separating this fragment into multiple fragments, I think you could use subsections if they had common topics, but I'm not sure these really do.

docs/news.d/general.rst
```diff
@@ -0,0 +1,37 @@
...
+.. news-start-section: Fixes
+- Improvements
+
+  - Added export macro to ConditionImpl
+
+  - RtpsRelay: Use ACE_Message_Block's locking strategy for cached SPDP
+
+  - Remove per-element locking in JobQueue
+
+  - Remove unused constructor in WeakObject
```
iguessthislldo, line 16: This doesn't seem news-worthy.

docs/news.d/general.rst
```diff
@@ -0,0 +1,37 @@
...
+
+  - Make use of unique_ptr
+
+- Fixes
```
iguessthislldo, line 20: I'm not sure this makes sense to have a section in Fixes called "Fixes".

docs/news.d/config_store.rst
```diff
@@ -1,4 +1,4 @@
-.. news-prs: 4162 4241 4242 4255 4243 4249
+.. news-prs: 4162 4241 4242 4255 4243 4249 4272 4270
 .. news-start-section: Additions
 - OpenDDS now stores ``rtps_udp`` transport configuration in the key-value store.
```
iguessthislldo, line 3: Could all of these combined into one item? Also could you like to the documentation of the config store so users can see what that is?

docs/news.d/general.rst
iguessthislldo: Yes, probably. Fragments, at least how I imagined them are a related set of changes.

iguessthislldo (changes requested):

docs/news.d/subjectname_parsing.rst
jrw972: I'll leave it out.

iguessthislldo (approved):

PR #4314, opened 2023-10-17
Fixes and Tweaks for CMake Documentation

iguessthislldo:


docs/devguide/building/cmake.rst
```diff
@@ -733,50 +737,51 @@ These variables can be used to override default behavior of the CMake package.
 Config Variables
 -----------------
 
-These variables are set by the configure script and normally shouldn't be changed.
+These variables are set by the ``configure`` script  in a MPC-built OpenDDS and normally shouldn't be changed.
+They can be changed when configuring a :ref:`CMake-built OpenDDS <cmake-building>` using ``-D``, but should not be chagned after that.
```
mitza-oci, line 741:
```suggestion
These variables are set by the ``configure`` script in an MPC-built OpenDDS and normally shouldn't be changed.
They can be changed when configuring a :ref:`CMake-built OpenDDS <cmake-building>` using ``-D``, but should not be changed after that.
```

mitza-oci (changes requested):

mitza-oci (approved):

mitza-oci (approved):

jrw972 (approved):

jrw972 (approved):

PR #4316, opened 2023-10-19
Add some tests to cmake 

tmayoff:
This PR is less about the particular test but I just wanted to work on getting something similar to the maps tests working in cmake. I split this off to keep the discussions focused.

The issue in the maps branch in the CI seems to be gtest, I'll see here if it's the same thing, and work on it


tests/DCPS/Compiler/char_literals/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+find_package(OpenDDS REQUIRED)
+include(opendds_testing)
+
+set(target_prefix "${PROJECT_NAME}_")
+set(src "${CMAKE_CURRENT_SOURCE_DIR}")
+set(dst ${CMAKE_CURRENT_BINARY_DIR})
+set(opendds_libs
+  OpenDDS::Dcps
+  TAO::TAO
```
iguessthislldo, line 13: `TAO::TAO` is already be linked through:

https://github.com/OpenDDS/OpenDDS/blob/18103482a923d95ecb1370cf235d992e9529c0a8/dds/CMakeLists.txt#L5-L11

and:

https://github.com/OpenDDS/OpenDDS/blob/18103482a923d95ecb1370cf235d992e9529c0a8/dds/CMakeLists.txt#L561

I assume you're trying to fix:

> /home/runner/work/OpenDDS/OpenDDS/OpenDDS/build/tests/DCPS/Compiler/char_literals/opendds_compiler_char_literals_exe: error while loading shared libraries: libTAO_AnyTypeCode.so.3.1.1: cannot open shared object file: No such file or directory

This is weird. I'm assuming the Windows cmake builds are having the same problem. I'm not sure why this is only an issue in this test and if it really should be looking for the library why it can't find it as the TAO `lib` that should be in `PATH`/`LD_LIBRARY_PATH`.

I'll try to look at this in detail some time tomorrow. It might be that `TAO::AnyTypeCode` has to be added to `dcps_lib_deps` in Dcps' CMakeLists.

tmayoff: Yeah I was, I saw it was linked from the Dcps lib, but just checked to see if it would work, doing the same with TAO::AnyTypeCode now as a long shot here

tmayoff:
Also I tried adding the HelloWorld test, and I think something's missing, I had to change the subscriber/publisher code, 

https://github.com/OpenDDS/OpenDDS/blob/6fbaec6f205cf03dd30042cd9cdbe268389ade6b/tests/DCPS/HelloWorld/publisher.cpp#L34 I had to add a `.c_str()`, the the topic name, and https://github.com/OpenDDS/OpenDDS/blob/6fbaec6f205cf03dd30042cd9cdbe268389ade6b/tests/DCPS/HelloWorld/publisher.cpp#L52 and change .value -> ._value. Is there something missing in the build, in general or did I miss something?

tests/DCPS/Compiler/char_literals/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  RUNTIME_OUTPUT_DIRECTORY "${idl_dir}"
+  LIBRARY_OUTPUT_DIRECTORY "${idl_dir}"
+)
+
+# Assert the mapping used was C++11
+get_property(mappings TARGET ${idl} PROPERTY OPENDDS_LANGUAGE_MAPPINGS)
+if(NOT ("C++11" IN_LIST mappings))
+  message(FATAL_ERROR "${idl}: C++11 NOT in mapping list: ${mappings}")
+endif()
+
```
iguessthislldo, line 31: This is something for testing that `opendds_target_sources` is setting `OPENDDS_LANGUAGE_MAPPINGS` correctly and not every test needs this.

```suggestion
```

tests/DCPS/Compiler/char_literals/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+add_executable(opendds_compiler_char_literals_exe
+  "${src}/main.cpp"
+)
+set_target_properties(opendds_compiler_char_literals_exe PROPERTIES
+  RUNTIME_OUTPUT_DIRECTORY "${dst}"
+)
+target_link_libraries(opendds_compiler_char_literals_exe ${opendds_libs} ${idl} gtest)
+
+add_test(opendds_compiler_char_literals_test opendds_compiler_char_literals_exe)
```
iguessthislldo, line 40: It took a while to realize you didn't use `opendds_add_test`. It sets up the environment, like `PATH`/`LD_LIBRARY_PATH` correctly, which is why it can't find ACE/TAO libs. Ideally I'd more of what it does to be handled by the Perl test framework, but for now this is needed. It also is going to be better for testing for it to use `run_test.pl` instead of running the executable directly, though that means it has to conform to how it's built in MPC.

```suggestion
set_target_properties(opendds_compiler_char_literals_exe PROPERTIES
  OUTPUT_NAME char-literals-cpp11
  RUNTIME_OUTPUT_DIRECTORY "${dst}/cpp11"
)
target_link_libraries(opendds_compiler_char_literals_exe ${opendds_libs} ${idl} gtest)

configure_file(run_test.pl . COPYONLY)
opendds_add_test(COMMAND perl run_test.pl cpp11 EXTRA_LIB_DIRS "${idl_dir}")
```

tests/DCPS/Compiler/char_literals/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+set(opendds_libs
+  OpenDDS::Dcps
+)
+
+# IDL Library
+set(idl "${target_prefix}idl")
+add_library(${idl})
+opendds_target_sources(${idl} PUBLIC "${src}/test.idl" OPENDDS_IDL_OPTIONS "-Lc++11")
+target_link_libraries(${idl} PUBLIC OpenDDS::Dcps)
+set(idl_dir "${dst}")
```
iguessthislldo, line 20: Part of using `run_test.pl`.

```suggestion
set(idl_dir "${dst}/cpp11")
```

tests/DCPS/Compiler/char_literals/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+cmake_minimum_required(VERSION 3.8...3.27)
+project(opendds_compiler_char_literals CXX)
+enable_testing()
+
+find_package(OpenDDS REQUIRED)
+include(opendds_testing)
+
+set(target_prefix "${PROJECT_NAME}_")
+set(src "${CMAKE_CURRENT_SOURCE_DIR}")
```
iguessthislldo, line 9:
```suggestion
```

You can get rid of this and where it's mentioned. Input file paths are assumed to be relative to `CMAKE_CURRENT_SOURCE_DIR` most, if not all the time. We have it in the CMake tests because they're using source files located in a very different directories from the `CMakeLists.txt`.

tests/DCPS/HelloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+add_executable(${subscriber}
+  subscriber.cpp
+)
+target_link_libraries(${subscriber} ${opendds_libs})
+
+# Testing
+configure_file(run_test.pl . COPYONLY)
+configure_file(rtps.ini . COPYONLY)
+opendds_add_test(NAME info_repo COMMAND perl run_test.pl)
+opendds_add_test(NAME rtps COMMAND perl run_test.pl --rtps)
```
iguessthislldo, line 41: Using RTPS is a bit different from the Messenger.

```suggestion
opendds_add_test(NAME rtps COMMAND perl run_test.pl ini=rtps.ini)
```

tests/DCPS/HelloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  publisher.cpp
+)
+target_link_libraries(${publisher} ${opendds_libs})
+
+# Subscriber
+set(subscriber "${target_prefix}subscriber")
+add_executable(${subscriber}
+  subscriber.cpp
+)
+target_link_libraries(${subscriber} ${opendds_libs})
```
iguessthislldo, line 35: You'll have to set the `OUTPUT_NAME` target properties to `publisher` and `subscriber` to match how MPC builds them.

```suggestion
target_link_libraries(${publisher} ${opendds_libs})
set_target_properties(${publisher} PROPERTIES
  OUTPUT_NAME publisher
)

# Subscriber
set(subscriber "${target_prefix}subscriber")
add_executable(${subscriber}
  subscriber.cpp
)
target_link_libraries(${subscriber} ${opendds_libs})
set_target_properties(${subscriber} PROPERTIES
  OUTPUT_NAME subscriber
)
```

iguessthislldo (changes requested):

tmayoff:
@iguessthislldo, do you know why the build_cmake_w22, might be failing, 

The return code on this line, [`2023-11-01 19:59:02: ERROR: test finished and returned 3221225781`](https://github.com/OpenDDS/OpenDDS/actions/runs/6723900707/job/18275083017?pr=4316#step:13:110), seems very wrong, but don't know the cause.

mitza-oci:
> @iguessthislldo, do you know why the build_cmake_w22, might be failing,
> 
> The return code on this line, [`2023-11-01 19:59:02: ERROR: test finished and returned 3221225781`](https://github.com/OpenDDS/OpenDDS/actions/runs/6723900707/job/18275083017?pr=4316#step:13:110), seems very wrong, but don't know the cause.

https://gist.github.com/mbrownnycnyc/24cb49aa2cba35be4764b1fa37bf9063#file-nterr-h-errors-L311

tmayoff:
I tried looking into this but no luck, not sure if it's my unfamiliarity with cmake but I don't know, could always disable gtest tests with windows, but that's probably not ideal

iguessthislldo:
For the char_literals test, the problem is that Google Test isn't in the `PATH` when the test runs. I'm working on some changes to how the environment is setup for tests. It would probably also be good to use CMake's builtin support for Google Test. Still not sure what's wrong with the HelloWorld test yet though.

tmayoff:
Would it be okay to maybe skip gtest tests on windows for now? I'd like to be able to continue the maps PR with gtest in there.

.github/workflows/build_and_test.yml
```diff
@@ -12971,4 +12971,4 @@ jobs:
       shell: cmd
       run: |
         cd OpenDDS\build
-        ctest -C Debug
+        ctest --output-on-failure -C Debug
```
iguessthislldo, line 12974: This should be handled by `CTEST_OUTPUT_ON_FAILURE` [here](https://github.com/OpenDDS/OpenDDS/blob/6d807de50b1e7c9614b9c691375d0446b2657335/.github/workflows/build_and_test.yml#L66). Is that not working?

cmake/opendds_testing.cmake
```diff
@@ -20,7 +20,7 @@ function(opendds_add_test)
   list(APPEND arg_COMMAND ${arg_ARGS})
   add_test(NAME "${test_name}" ${arg_UNPARSED_ARGUMENTS} COMMAND ${arg_COMMAND})
 
-  set(env "ACE_ROOT=${ACE_ROOT}" "TAO_ROOT=${TAO_ROOT}")
+  set(env "ACE_ROOT=${ACE_ROOT}" "TAO_ROOT=${TAO_ROOT}" "DDS_ROOT=${DDS_ROOT}")
```
iguessthislldo, line 23:
```suggestion
  set(env "ACE_ROOT=${ACE_ROOT}" "TAO_ROOT=${TAO_ROOT}")
```

I'm explicitly not setting this because it shouldn't be necessary and I'm trying to eventually get rid of it.

iguessthislldo (changes requested):
I did some more testing on Windows to get Google Test working and have some suggested changes in https://github.com/tmayoff/OpenDDS/pull/3.

.github/workflows/build_and_test.yml
tmayoff: It is, the first few CI tests didn't seem to be outputting but was failing in a different way or something.

cmake/opendds_testing.cmake
```diff
@@ -27,6 +27,13 @@ function(opendds_add_test)
...
     set(env_var_name LD_LIBRARY_PATH)
   endif()
+  if(DEFINED _OPENDDS_GOOGLE_TEST_DIR)
+    _opendds_path_list(lib_dir_list APPEND "${_OPENDDS_GOOGLE_TEST_DIR}")
+    if(CMAKE_CONFIGURATION_TYPES)
+      _opendds_path_list(lib_dir_list APPEND
+        "${_OPENDDS_GOOGLE_TEST_DIR}$<$<BOOL:$<CONFIG>>:/$<CONFIG>>")
+    endif()
+  endif()
   _opendds_path_list(lib_dir_list "$ENV{${env_var_name}}" "${TAO_LIB_DIR}")
```
iguessthislldo, line 37: Somehow the `if(DEFINED _OPENDDS_GOOGLE_TEST_DIR)...` and the lone ` _opendds_path_list` got swapped between testing and when I submitted the PR. This should fix the char_literals test.

```suggestion
  _opendds_path_list(lib_dir_list "$ENV{${env_var_name}}" "${TAO_LIB_DIR}")
  if(DEFINED _OPENDDS_GOOGLE_TEST_DIR)
    _opendds_path_list(lib_dir_list APPEND "${_OPENDDS_GOOGLE_TEST_DIR}")
    if(CMAKE_CONFIGURATION_TYPES)
      _opendds_path_list(lib_dir_list APPEND
        "${_OPENDDS_GOOGLE_TEST_DIR}$<$<BOOL:$<CONFIG>>:/$<CONFIG>>")
    endif()
  endif()
```

tests/DCPS/HelloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+set_target_properties(${subscriber} PROPERTIES
+  OUTPUT_NAME subscriber
+)
+
+# Testing
+configure_file(run_test.pl . COPYONLY)
+configure_file(rtps.ini . COPYONLY)
+opendds_add_test(NAME info_repo COMMAND perl run_test.pl)
+opendds_add_test(NAME rtps COMMAND perl run_test.pl init=rtps.ini)
```
iguessthislldo, line 47:
```suggestion
opendds_add_test(NAME rtps COMMAND perl run_test.pl ini=rtps.ini)
```

This is another typo, but I'm not sure fixing this will mean rtps will work since info_repo isn't, unless the InfoRepo isn't working at all for some reason?

iguessthislldo (changes requested):

.github/workflows/build_and_test.yml
iguessthislldo: This is a guess, but you might have been looking at MPC-built OpenDDS builds. In those CMake tests are still run using ctest, but using different logging and reporting mechanism that gets put into artifacts. The CMake-built OpenDDS builds just use ctest directly. It's on a TODO list somewhere to decide what to do about this difference.

iguessthislldo:
Okay, finally char_literals is working. It's also nice to see that HelloWorld RTPS is also working. HelloWorld InfoRepo/TCP isn't working though. It works fine on my machine, so there's something particular about GitHub Actions.

@jrw972 do you have any ideas?

https://github.com/OpenDDS/OpenDDS/actions/runs/6948505054/job/18904789877?pr=4316

My only thought is maybe the combination of waiting for the InfoRepo to spin up, the timeout on `get_fully_qualified_hostname`, waiting on DCS all running on GHA is reaching the timeout. Maybe setting [`DCPSDefaultAddress`](https://github.com/OpenDDS/OpenDDS/issues/453) will fix this? If possible I'd like a general solution for all the tests as I mentioned in the chat a little while a go. I did have this FQDN issue with the timeout on my machine, but it appears I fixed it by setting the DNS suffix to my tailscale domain since I don't have a real Windows domain for it to join.

iguessthislldo:
https://github.com/tmayoff/OpenDDS/pull/4

Increasing the default to 60s also seems to have worked:

https://github.com/iguessthislldo/OpenDDS/actions/runs/6950846671/job/18911765231

iguessthislldo (approved):

iguessthislldo (approved):

PR #4327, opened 2023-10-30
Automatically Update ACE/TAO Versions

iguessthislldo:
These changes introduce `acetao.ini` that the configure script, CMake, and Sphinx can get the ACE/TAO versions and related info from. This can be updated manually using:

    tools/scripts/gitrelease.pl --update-ace-tao

or by a workflow that will open a PR when the file needs to be updated.

The PRs will look like this: https://github.com/iguessthislldo/OpenDDS/pull/43


jwillemsen:
Would be really great when running this would be automated, maybe use https://github.com/marketplace/actions/changed-files, that could watch the ACE/TAO repo, when something changed, create the PR?

iguessthislldo:
The way I have it setup is more direct than that because it's checking GitHub releases in the GitHub API. If, for some reason, the release archives and MD5s were changed after a release was made, it should detect that and make a PR for that. Also it would always have to look at the GitHub releases for the MD5s, so looking at the repo itself doesn't really add anything.

jrw972 (approved):

PR #4339, opened 2023-11-07
MemberDescriptor implementation is incomplete

jrw972:
Problem
-------

The MemberDescriptor implementation is incomplete.  Completing it is necessary to support user-constructed DynamicTypes.

Solution
--------

Complete MemberDescriptorImpl.

The largest part of the code is devoted to checking that default values are a valid IDL literal for the member type.  This prompted the creation of a primitive scanner.

### DIFF
--- a/dds/DCPS/XTypes/IdlScanner.cpp
+++ b/dds/DCPS/XTypes/IdlScanner.cpp
@@ -1,0 +1,148 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include <DCPS/DdsDcps_pch.h>
+
+#ifndef OPENDDS_SAFETY_PROFILE
+#  include "IdlScanner.h"
+
+#  include "DynamicTypeImpl.h"
+
+#  include <ace/Basic_Types.h>
+
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+namespace OpenDDS {
+namespace XTypes {
+
+IdlToken
+IdlScanner::scan_token(DDS::DynamicType_ptr type)
+{
+  OPENDDS_ASSERT(type);
+
+  DDS::DynamicType_var bt = get_base_type(type);
+
+  switch (bt->get_kind()) {
+  case TK_BOOLEAN: {
+    const IdlToken token = scan_identifier();
+    if (token.is_boolean()) {
+      return token;
+    }
+    break;
+  }
+  case TK_BYTE: {
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= ACE_UINT8_MAX) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT8: {
+    const IdlToken token = scan_numeric_literal();
+    const ACE_UINT64 limit = static_cast<ACE_UINT64>(ACE_INT8_MAX) + (token.is_signed() ? 1 : 0);
+    if (token.is_integer() && token.integer_value() <= limit) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT16: {
+    const IdlToken token = scan_numeric_literal();
+    const ACE_UINT64 limit = static_cast<ACE_UINT64>(ACE_INT16_MAX) + (token.is_signed() ? 1 : 0);
+    if (token.is_integer() && token.integer_value() <= limit) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT32: {
+    const IdlToken token = scan_numeric_literal();
+    const ACE_UINT64 limit = static_cast<ACE_UINT64>(ACE_INT32_MAX) + (token.is_signed() ? 1 : 0);
+    if (token.is_integer() && token.integer_value() <= limit) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT64: {
+    const IdlToken token = scan_numeric_literal();
+    const ACE_UINT64 limit = static_cast<ACE_UINT64>(ACE_INT64_MAX) + (token.is_signed() ? 1 : 0);
+    if (token.is_integer() && token.integer_value() <= limit) {
+      return token;
+    }
+    break;
+  }
+  case TK_UINT8: {
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= ACE_UINT8_MAX) {
+      return token;
+    }
+    break;
+  }
+  case TK_UINT16: {
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= ACE_UINT16_MAX) {
+      return token;
+    }
+    break;
+  }
+  case TK_UINT32: {
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= ACE_UINT32_MAX) {
+      return token;
+    }
+    break;
+  }
+  case TK_UINT64: {
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= ACE_UINT64_MAX) {
+      return token;
+    }
+    break;
+  }
+  case TK_FLOAT32:
+  case TK_FLOAT64:
+  case TK_FLOAT128: {
+    const IdlToken token = scan_numeric_literal();
+    // TODO: Bounds checking.
+    if (token.is_float()) {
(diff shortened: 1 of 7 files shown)


dds/DCPS/XTypes/IdlScanner.cpp
```diff
@@ -0,0 +1,147 @@
...
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include <DCPS/DdsDcps_pch.h>
+
+#ifndef OPENDDS_SAFETY_PROFILE
+#  include "IdlScanner.h"
+
+# include "DynamicTypeImpl.h"
```
iguessthislldo, line 11:
```suggestion
#  include "DynamicTypeImpl.h"
```

dds/DCPS/XTypes/IdlScanner.cpp
```diff
@@ -0,0 +1,147 @@
+/*
```
iguessthislldo, line 1: You need to add the new header and cpp file to `dds/CMakeLists.txt`.

dds/DCPS/XTypes/IdlScanner.cpp
```diff
@@ -0,0 +1,147 @@
...
+    const IdlToken token = scan_numeric_literal();
+    const uint64_t limit = token.is_signed() ? 2147483648 : 2147483647;
+    if (token.is_integer() && token.integer_value() <= limit) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT64: {
+    const IdlToken token = scan_numeric_literal();
+    const uint64_t limit = token.is_signed() ? 9223372036854775808UL : 9223372036854775807;
```
iguessthislldo, line 58: `uint64_t` isn't C++03. Also use either C, C++, or ACE macros/functions for the limits?

iguessthislldo (changes requested):

dds/DCPS/XTypes/IdlScanner.cpp
iguessthislldo: Or for this specifically I guess it looks like you're trying to stuff signed into unsigned. That seems like it might be problematic. Why not a separate signed field/type?

dds/DCPS/XTypes/IdlScanner.h
```diff
@@ -0,0 +1,890 @@
...
+      return IdlToken::make_eoi();
+    }
+
+    const char c = scanner_.peek();
+    if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z')) {
+      return scan_identifier();
+    }
+
+    switch (c) {
+    case '\'':
```
iguessthislldo, line 260: I know we're probably not supporting it at least for the moment, but does this handle being given a wide string or char?

jrw972: I was uninterested in wide characters since it is unclear how `default_value` would store a wide character or string.

dds/DCPS/XTypes/IdlScanner.cpp
jrw972: Getting the limits from a macro is a good idea.  C/C++/ACE is going to give me something that's for the compiler and not IDL.  I looked at ACE and TAO for limit for IDL/CDR but didn't find anything.  Let me know if I missed something.

dds/DCPS/XTypes/IdlScanner.h
```diff
@@ -0,0 +1,890 @@
...
+public:
+  enum Kind {
+    Error,
+    EOI,
+    BooleanLiteral,
+    IntegerLiteral,
+    FloatLiteral,
+    StringLiteral,
+    CharacterLiteral,
+    Identifier
```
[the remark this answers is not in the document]
jrw972: Identifiers can appear for boolean literals (`TRUE` or `FALSE`) and for enum values.  Section 7.5.2.7.3 shows an illegal `default_value` based on an enum.  While the example is illegal, it suggests that enum values can be treated like literals.

dds/DCPS/XTypes/IdlScanner.h
```diff
@@ -0,0 +1,890 @@
...
+    struct {
+      bool signed_;
+      ACE_UINT64 integer_value_;
+      ACE_UINT64 digits_after_decimal_;
+      bool exponent_signed_;
+      ACE_UINT64 exponent_value_; // TODO:  Handle overflow.
+    };
+  };
+  std::string string_literal_;
+  std::string identifier_;
```
sonndinh, line 882: In another comment on `IdlToken`, my thought was that union may be used to make it clear that an instance of `IdlToken` (and `IdlScanner` if I understand it right) can hold only one of those values at a time. But looks like `string` can't be included in a union. I'm not sure what an alternative would be.

Also, there are some similarities between the data members of `IdlToken` and `IdlScanner`. Could they share a common data type for that?

dds/DCPS/XTypes/IdlScanner.h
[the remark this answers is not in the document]
sonndinh: Got it now. They sound like `Constant` to me than `Identifier` but I'm fine with it.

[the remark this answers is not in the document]
jrw972: If I were to implement a parser for IDL, I would populate the symbol table with constants for `TRUE` and `FALSE` so that the identifier would return the appropriate constant.

dds/DCPS/XTypes/MemberDescriptorImpl.cpp
```diff
@@ -89,7 +154,7 @@ bool test_equality(DDS::MemberDescriptor* lhs, DDS::MemberDescriptor* rhs, Dynam
     std::strcmp(lhs->name(), rhs->name()) == 0 &&
     lhs->id() == rhs->id() &&
     test_equality(lhs_type, rhs_type, dt_ptr_pair) &&
-    lhs->default_value() == rhs->default_value() &&
+    std::strcmp(lhs->default_value(), rhs->default_value()) == 0 &&
```
iguessthislldo, line 157: This is probably okay, but doesn't seem like it's technically correct since there could be multiple string representations for some values, eg `0x10` and `16`.

tests/unit-tests/dds/DCPS/XTypes/MemberDescriptorImpl.cpp
```diff
@@ -0,0 +1,175 @@
...
+  dti->set_descriptor(td);
+
+  MemberDescriptorImpl md1;
+  EXPECT_EQ(md1.copy_from(0), DDS::RETCODE_BAD_PARAMETER);
+
+  MemberDescriptorImpl md2;
+  md2.name("name");
+  md2.id(1);
+  md2.type(dt);
+  md2.default_value("true");
```
iguessthislldo, line 155: Is this supposed to be `TRUE` or does it not matter in this context?

iguessthislldo (changes requested):

dds/DCPS/XTypes/IdlScanner.cpp
```diff
@@ -0,0 +1,146 @@
...
+    const IdlToken token = scan_numeric_literal();
+    if (token.is_integer() && !token.is_signed() && token.integer_value() <= 255) {
+      return token;
+    }
+    break;
+  }
+  case TK_INT16: {
+    const IdlToken token = scan_numeric_literal();
+    const ACE_UINT64 limit = token.is_signed() ? 32768 : 32767;
+    if (token.is_integer() && token.integer_value() <= limit) {
```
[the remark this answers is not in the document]
jrw972: If `is_signed` is true, then the minimum is checked.

dds/DCPS/XTypes/IdlScanner.h
```diff
@@ -0,0 +1,898 @@
...
+      return IdlToken::make_eoi();
+    }
+
+    const char c = scanner_.peek();
+    if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z')) {
+      return scan_identifier();
+    }
+
+    switch (c) {
+    case '\'':
```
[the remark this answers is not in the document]
jrw972: Yes.  This method could be used to parse actual IDL so while not strictly necessary for this PR, it forced me to think about what was common for parsing `default_value` and parsing IDL.  The other `scan_token` method is using for `default_value`.

tests/unit-tests/dds/DCPS/XTypes/MemberDescriptorImpl.cpp
jrw972: Yes.

iguessthislldo (approved):

PR #4343, opened 2023-11-08
Remove disposed instances from DataReader containers

smnrgrs:
Testing of publishing and then unregistering/disposing a large number of instances shows that DataReader maps grow in size and never recover, taking more and more memory. 

We've memory profiled our test apps and then made these modifications to clean up the DataReader containers. 

We're not familiar with the inner workings of OpenDDS so it would be good if someone who knows what they're doing could review these changes.

https://github.com/OpenDDS/OpenDDS/discussions/4334


jrw972:
@smnrgrs I put my suggestions on https://github.com/smnrgrs/OpenDDS/pull/1

jrw972 (approved):

PR #4344, opened 2023-11-08
Publish/subscribe actions do not have a validity section

jrw972:
Problem
-------

In a collaborative DDS Domain, such as those possible with the DDS Permissions Manager, different organization will contribute different publish/subscribe actions for the same application (subject).  These publish/subscribe actions will be valid in different timeframes.

Solution
--------

A desirable solution would be to have a permissions document with multiple grants referring to the same application.  However, the "default action" in a DDS Security grant effectively means that an application (subject) can have a single grant.  Thus, all of the publish/subscribe actions for an application must be combined under a single grant.  This means that either 1) the validity for the grant must be crafted carefully from the validity assigned to each action or 2) each action needs a validity interval.

Crafting the overall grant validity is possible but can lead to pathological behavior with respect to renewing a permissions document. That is, the application must new the permissions document at every point in time (not_before or not_after) for every publish/subscribe action.  To avoid this complexity, each publish/subscribe action is given a validity which is independent from the grant validity.

This is an OpenDDS extension to DDS Security.

### DIFF
--- a/docs/news.d/action-validity.rst
+++ b/docs/news.d/action-validity.rst
@@ -1,5 +1,5 @@
 .. news-prs: 4344
 
 .. news-start-section: Additions
-- It is now possible to specify the validity for individual publish/subscribe actions in DDS Security Permission documents.  This is an OpenDDS extension.
+- It is now possible to specify the :ref:`validity for individual publish/subscribe actions <dds_security--psr-validity>` in DDS Security Permission documents.  This is an OpenDDS extension.
 .. news-end-section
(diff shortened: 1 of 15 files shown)


docs/news.d/action-validity.rst
```diff
@@ -0,0 +1,5 @@
+.. news-prs: 4344
+
+.. news-start-section: Additions
+- It is now possible to specify the validity for individual publish/subscribe actions in DDS Security Permission documents.  This is an OpenDDS extension.
```
iguessthislldo, line 4:
```suggestion
- It is now possible to specify the :ref:`validity for individual publish/subscribe actions <dds_security--psr-validity>` in DDS Security Permission documents.  This is an OpenDDS extension.
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4373, opened 2023-11-29
Implement `create_sample` and `create_dynamic_sample`

iguessthislldo:


MPC/modules/IDLBase.pm
```diff
@@ -115,7 +115,11 @@ sub do_cached_parse {
   $self->{'default_nested'} = 1;
 
   ## Set up the macros and include paths supplied in the command flags
-  my %macros = ('__OPENDDS_IDL' => 1, '__OPENDDS_MPC' => 1);
+  my %macros = (
+    __OPENDDS_IDL => 1,
+    __OPENDDS_MPC => 1,
+    OPENDDS_HIDE_DYNMIC_DATA => 1,
```
mitza-oci, line 121:
```suggestion
    OPENDDS_HIDE_DYNAMIC_DATA => 1,
```
and all other occurrences

dds/DCPS/TypeSupportImpl.h
```diff
@@ -218,6 +222,69 @@ class TypeSupportImpl_T : public TypeSupportImpl {
...
+
+  /// The IDL is `NativeType create_sample(DDS::DynamicData src)`, but in the
+  /// C++ mapping this can return NativeType* or just NativeType depending on if
+  /// the type is "variable length". If it has something like a sequence then
+  /// it's variable length and returns a pointer else it's fixed and returns on
+  /// the stack. opendds_idl will wrap this in the correct form.
+  DDS::ReturnCode_t create_sample_rc(DDS::DynamicData_ptr src, NativeType& dst)
+  {
+    OpenDDS::DCPS::set_default(dst);
+#  if OPENDDS_HAS_DYNAMIC_DATA_ADAPTOR
```
mitza-oci, line 234:
```suggestion
#  if OPENDDS_HAS_DYNAMIC_DATA_ADAPTER
```

docs/news.d/create-sample.rst
```diff
@@ -0,0 +1,6 @@
+.. news-prs: 4373
+
+.. news-start-section: Additions
+- Implemented the ``create_dynamic_sample`` and ``create_sample`` methods on ``TypeSupport`` to convert samples to and from ``DynamicData``.
+  Also documented the existing ``DynamicDataAdaptor`` class for wrapping IDL-generated types in a ``DynamicData``.
```
mitza-oci, line 5:
```suggestion
  Also documented the existing ``DynamicDataAdapter`` class for wrapping IDL-generated types in a ``DynamicData``.
```

docs/devguide/xtypes.rst
```diff
@@ -1364,13 +1505,11 @@ Type System
...
-* IDL map type
-
-* IDL bitmask type
+* IDL ``map`` type
 
-* .. _xtypes--refheading-toc32438-4273764768:
+* IDL ``bitmask`` type
 
-  Struct and union inheritance
+* Struct and union inheritance
```
iguessthislldo, line 1512: Unless I'm missing something, I don't think union inheritance is in the specs.

dds/DCPS/TypeSupportImpl_T.h
```diff
@@ -0,0 +1,150 @@
...
+    get_type_from_type_lookup_service();
+    return TypeSupportImpl::get_type();
+  }
+
+  /// The IDL is `NativeType create_sample(DDS::DynamicData src)`, but in the
+  /// C++ mapping this can return NativeType* or just NativeType depending on if
+  /// the type is "variable length". If it has something like a sequence then
+  /// it's variable length and returns a pointer else it's fixed and returns on
+  /// the stack. opendds_idl will wrap this in the correct form.
+  DDS::ReturnCode_t create_sample_rc(DDS::DynamicData_ptr src, NativeType& dst)
```
iguessthislldo, line 89: I arbitrarily chose source and destination, but do we have a convention for what order these go in? When writing the examples I accidentally did destination and source like `XTypes::copy` and `memset`, so I'm wondering if I should switch them.

docs/devguide/xtypes.rst
```diff
@@ -1315,10 +1331,135 @@ They provides the same operations as any other DataWriter or DataReader, but wit
...
+
+The XTypes specification defines ``create_sample`` and ``create_dynamic_sample`` on the ``TypeSupport`` for every :term:`topic type` to interchange ``DynamicData`` and IDL-generated types.
+In addition to this, OpenDDS adds ``create_sample_rc`` and ``create_dynamic_sample_rc`` to indicate failure and provide a more consistent interface.
+The standard methods will only warn if there was an error and can't indicate that an error occured.
+
+``create_sample`` and ``create_sample_rc``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``create_sample`` takes a ``DynamicData`` and returns the IDL-generated type.
+One thing to note about this method is the C++ to IDL specification requires that returns like this can be either returned on the stack or the heap depending on if they are :omgspec:`"variable-length" <cpp03:5.11 Mapping for Structured Types>`.
```
mitza-oci, line 1357:
```suggestion
One thing to note about this method is the IDL-to-C++ specification requires that returns like this can be either returned on the stack or the heap depending on if they are :omgspec:`"variable-length" <cpp03:5.11 Mapping for Structured Types>`.
```

docs/devguide/xtypes.rst
```diff
@@ -1315,10 +1331,135 @@ They provides the same operations as any other DataWriter or DataReader, but wit
...
+  };
+
+.. code-block:: cpp
+
+  FooTypeSupport_var type_support = new FooTypeSupportImpl();
+  DynamicData_var data = /* ... */;
+
+  Foo foo = type_support->create_sample(data);
+
+If the type contains a ``vector`` or ``string`` anywhere, then then it returns a pointer that's meant to go inside a ``_var`` class:
```
mitza-oci, line 1373:
```suggestion
If the type contains a ``sequence`` or ``string`` anywhere, then then it returns a pointer that's meant to go inside a ``_var`` class:
```
Are there other IDL constructs that can be variable-length and can appear in a DDS topic type?

docs/devguide/xtypes.rst
```diff
@@ -1315,10 +1331,135 @@ They provides the same operations as any other DataWriter or DataReader, but wit
...
+
+.. _xtypes-dynamic-data-adapter:
+
+Wrapping Data using ``DynamicDataAdapter``
+------------------------------------------
+
+.. versionadded:: 3.29.0 (but unofficially present in its current form in 3.24.0)
+
+``DynamicDataAdapter`` is an OpenDDS-specific implementation of ``DynamicData`` that wraps an IDL-generated type.
+This allows direct interaction instead of copying all the data everytime.
```
mitza-oci, line 1430:
```suggestion
This allows direct interaction instead of copying all the data every time.
```

mitza-oci (changes requested):

docs/devguide/xtypes.rst
iguessthislldo: Why do I keep doing that?

docs/devguide/xtypes.rst
mitza-oci: It's proposed for future versions https://issues.omg.org/issues/spec/IDL#issue-43778
That distinction was probably missed when this part of the devguide was first authored

dds/DCPS/TypeSupportImpl_T.h
mitza-oci: `std::copy` is the reverse of `std::strcpy` so there's no right answer

docs/devguide/xtypes.rst
iguessthislldo: If everything in the specs, `tao_idl`, and OpenDDS was ideal and consistent, then probably `map`. [Everything else](https://www.omg.org/spec/CPP/1.3/PDF#G6.69275) doesn't look relevant to DDS.

mitza-oci: How about putting a comment in the rst (if it supports comments) that map should be included here when it's ready?

mitza-oci (approved):

PR #4382, opened 2023-12-08
Improve SpdpSendAddrs parsing in ini file

smnrgrs:
For large deployments with no multicast discovery, the current SpdpSendAddrs syntax in the ini file is cumbersome.

Allow multiple ports per host to be easily specified. Each SpdpSendAddrs entry can now be: <address>:<first_port>[:<count>[:<increment>]]

Example:
  SpdpSendAddrs=100.11.13.15:13910:16:2

  This will specify 16 endpoints on 100.11.13.15, starting from port 13910, counting up two at a time.


jrw972:
Sorry for getting to this so late.  This code has moved to ConfigStoreImpl.  Please add a unit test when you update the pull request and we'll get this in.

smnrgrs:
Thanks, reworked as #4400

PR #4389, opened 2023-12-12
`[DomainRange/*]` does not use `ConfigStore`

jrw972:
Problem
-------

`[DomainRange/*]` does not use ConfigStore. See #4134.

Solution
--------

Convert `[DomainRange/*]` to use ConfigStore.


dds/DCPS/Service_Participant.cpp
```diff
@@ -1525,14 +1524,23 @@ Service_Participant::load_configuration(
   // Domain config is loaded after Discovery (see below). Since the domain
   // could be a domain_range that specifies the DiscoveryTemplate, check
   // for config templates before loading any config information.
```
sonndinh, line 1526: Should this comment be moved to a proper place?

PR #4390, opened 2023-12-12
Quick Starts do not link to DevGuide

jrw972:
Problem
-------

The Quick Starts have been migrated to the DevGuide but the website has not been update.

Solution
--------

Update the website to link to the DevGuide.


iguessthislldo (approved):

iguessthislldo:
Actually me and Adam were wondering if you could replace the quickstarts with redirects in case there are existing links to them.

jrw972:
> Actually me and Adam were wondering if you could replace the quickstarts with redirects in case there are existing links to them.

That makes sense.

PR #4398, opened 2023-12-18
`create_datawriter/datareader` do not check topic participant

jrw972:
Problem
-------

Sections 2.2.2.4.1.5 and 2.2.2.5.2.5 of the DDS Specification say that `create_datawriter` and `create_datareader` should fail if the topic does not belong to the same participant as the publisher/subscriber.

Solution
--------

Add the necessary checks to `create_datawriter` and `create_datareader`.

### DIFF
--- a/docs/news.d/create-writer-reader-topic.rst
+++ b/docs/news.d/create-writer-reader-topic.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4398
+
+.. news-start-section: Fixes
+- ``create_datawriter`` and ``create_datareader`` check if the topic belongs to the same participant as the publisher/subscriber.
+.. news-end-section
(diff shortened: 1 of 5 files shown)


mitza-oci (approved):

docs/news.d/create-writer-reader-topic.rst
```diff
@@ -0,0 +1,5 @@
+.. news-prs: 4398
+
+.. news-start-section: Fixes
+- `create_datawriter` and `create_datareader` check if the topic belongs to the same participant as the publisher/subscriber.
```
iguessthislldo, line 4:
```suggestion
- ``create_datawriter`` and ``create_datareader`` check if the topic belongs to the same participant as the publisher/subscriber.
```

iguessthislldo (changes requested):

sonndinh (approved):

iguessthislldo (approved):

PR #4399, opened 2023-12-18
Add `vwrite` for `DynamicData`

sonndinh:


tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.cpp
```diff
@@ -276,11 +261,306 @@ TEST(VreadVwriteTest, SerializeTest)
...
+
+  // bt
+  ASSERT_EQ(DDS::RETCODE_OK, dt->get_member_by_name(dtm, "bt"));
+  DDS::DynamicData_var bt_dd;
+  ASSERT_EQ(DDS::RETCODE_OK, dd.get_complex_value(bt_dd, dtm->get_id()));
+  DDS::DynamicType_var bt_dt = bt_dd->type();
+
+  ASSERT_EQ(DDS::RETCODE_OK, bt_dt->get_member_by_name(dtm, "o"));
+  ASSERT_EQ(DDS::RETCODE_OK, bt_dd->set_byte_value(dtm->get_id(), sample.bt.o));
+#ifdef OPENDDS_HAS_EXPLICIT_INTS
```
mitza-oci, line 337: Does this still need to be conditional or can we assume that all builds have support for IDL explicit ints?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -4654,12 +4654,50 @@ void serialized_size_wstring_value(const Encoding& encoding, size_t& size, const
 }
 #endif
 
-bool check_rc_from_get(DDS::ReturnCode_t rc, DDS::MemberId id, DDS::TypeKind tk, const char* fn_name)
+const char* log_level_to_string(LogLevel::Value log_level)
```
mitza-oci, line 4657: This function and the one below aren't related to DynamicData - could they be moved somewhere else? Or, do they already exist somewhere else in the library?

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -5962,6 +6000,640 @@ bool operator<<(Serializer& ser, const KeyOnly<DDS::DynamicData_ptr>& key)
   return serialize(ser, key.value, Sample::KeyOnly);
 }
 
+} // namespace DCPS
+
+namespace DCPS {
```
mitza-oci, line 6005:
```suggestion
} // namespace DCPS

namespace DCPS {
```

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -5962,6 +6000,640 @@ bool operator<<(Serializer& ser, const KeyOnly<DDS::DynamicData_ptr>& key)
...
+    rc = enum_bound(item_type, tk);
+    if (rc != DDS::RETCODE_OK) {
+      if (log_level >= LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: vwrite_item: enum_bound failed (%C)\n",
+                   retcode_to_string(rc)));
+      }
+      return;
+    }
+  }
+  if (item_tk == TK_BITMASK) {
```
mitza-oci, line 6032: This could be an `else if` but the compiler is probably going to figure that out on its own -- I'd prefer having it in there so that readers realize that it can't take both "then" branches.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -5962,6 +6000,640 @@ bool operator<<(Serializer& ser, const KeyOnly<DDS::DynamicData_ptr>& key)
...
+      ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: vwrite_element: get_member_id_at_index %u failed\n", idx));
+    }
+    return;
+  }
+  vwrite_item(vw, value, id, elem_dt);
+}
+
+// When we're sure that the concrete DynamicData class has implemented get_*_values operations
+// for primitive sequences, we can use ValueWriter's write_*_array functions.
+// For now, vwrite for array and sequence will write elements one by one.
```
mitza-oci, line 6569: Which implementations are missing those functions?  One strategy here would be to call the _values function and if it returns an error fall back to the one-at-a-time method.

mitza-oci (changes requested):

dds/DCPS/ValueWriter.h
```diff
@@ -110,6 +110,8 @@ class OpenDDS_Dcps_Export ValueWriter {
     write_enum(name, static_cast<ACE_CDR::Long>(value));
   }
 
+  virtual void write_absent_value() = 0;
```
jrw972, line 113: I would like to understand more about the intent of this method.  From the code, it seems like it is called in an error condition.  However, this may have been how you implemented optional.

If it is just for errors, then I think the method can be removed and the resulting stream would just be in a bad state.  It may be useful to add return codes to the `vwrite` and `ValueWriter` family.

If it is for optional, I would again favor not having this method as the member can just be omitted.

tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.cpp
sonndinh: It looks to me all the builds in CI have support for explicit ints. But it's configurable from TAO, so I think it still needs to be conditional here.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: I don't see the difference from the suggested change, but I assume the suggestion is to merge these two `namespace DCPS` blocks?

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: `DynamicDataImpl` and `DynamicDataAdapter` haven't had these methods yet. Yeah, the fall back strategy sounds good to me.

dds/DCPS/XTypes/DynamicDataImpl.cpp
mitza-oci: Yes, that's what I meant

tests/DCPS/Compiler/vread_vwrite/VreadVwriteTest.cpp
mitza-oci: OK, I guess removing the conditionals can be a separate PR

dds/DCPS/ValueWriter.h
mitza-oci: ValueWriter having awareness of an "empty optional" may be useful for an SomeOtherValueWriter even if JsonValueWriter doesn't use it.

sonndinh: It is for optional member that is missing. For `JsonValueWriter` and `PrinterValueWriter`, I think it would be good to have a missing optional member noted as absent in the resulting stream, e.g. something like `OptionalField: null`. This way we know that the stream has entries for all members and don't have to check with the type definition to see what members are missing.

The side effect of this is that the `vwrite` function for dynamic data (and the generated ones for IDL types when we support `@optional` in IDL) can keep the same form as it is now. That is,
```
vw.begin_struct();
// Iterate over all members
for (...) {
  vw.begin_struct_member(...);
  vw.end_struct_member();
}
vw.end_struct();
```
Otherwise, I think for each member it'll need to check if the member doesn't have a value, and skip the calls to `begin_struct_member` and `end_struct_member` if so.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: There seems no similar function elsewhere. I moved them to XTypes/Utils.h. We could move them to a common utility file in dds/DCPS but there seems no use for them right now in DCPS.

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: Looking at this comment again, it was more of an optimization for primitive sequence/array with a XcdrValueWriter when we support it because `Serializer` has functions to write arrays and they're more efficient. For `JsonValueWriter` and `PrinterValueWriter`, their write_*_array functions also write elements one at a time. So I'd prefer to defer this to when we implement a XcdrValueWriter, preferably in #4392.

dds/DCPS/XTypes/DynamicDataImpl.cpp
mitza-oci: Is it the same as `LogLevel::get_as_string()`?

sonndinh: It's very similar. Maybe make them static member functions in `LogLevel`?

mitza-oci: Given a value `v` of type `LogLevel::Value`, the existing code appears to support `LogLevel(v).get_as_string()`

dds/DCPS/ValueWriter.h
jrw972: From discussion with @sonndinh and @mitza-oci.

An XCDR2 ValueWriter with a final or appendable struct must write a boolean indicating the status of an optional member.  In order for this to work, `begin_struct_member` and `end_struct_member` must be called.

Two options exist for conveying the presence of the optional member:
1. Pass a flag to `begin_struct_member`.
2. Call a method like `write_absent_value`.

Option 2 was selected as it restrains the scope of `begin_struct_member` to dealing with the type of the member instead of the status/value of the member.

dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -5962,6 +5964,637 @@ bool operator<<(Serializer& ser, const KeyOnly<DDS::DynamicData_ptr>& key)
...
+    if (!check_rc(rc, id, tk, "vwrite_item")) {
+      return;
+    }
+    if (rc == DDS::RETCODE_OK) {
+      if (item_tk == TK_ENUM) {
+        write_enum(vw, item_type, val);
+      } else {
+        vw.write_int8(val);
+      }
+    } else {
```
jrw972, line 6017: Should these explicitly check for NO_DATA?

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: Yes, I realized that as well. I was hesitant to use it this way because the constructor of `LogLevel` does other setups that aren't relevant to the purpose of `to_string`. The downside though is that there is some similarity between `get_as_string` and `to_string`.

@iguessthislldo Fred, can you take a look at these changes to `LogLevel`?

dds/DCPS/XTypes/DynamicDataImpl.cpp
sonndinh: At this point, we're sure that it's either OK or NO_DATA, so it doesn't need to check.

dds/DCPS/ValueWriter.h
sonndinh: I checked the DDS-JSON spec and it says "Unset optional members shall be omitted from the sample representation." So I think if we conform with the spec, `begin_struct_member` will need to know the status of the member so that it doesn't print the member name.

sonndinh: I still see some benefits of printing something like `OptionalField: null`, for example, for human inspection. But I also see the benefits of omitting absent values like the DDS-JSON spec does to reduce the sample size when sending over the wire.

dds/DCPS/XTypes/DynamicDataImpl.cpp
iguessthislldo: We shouldn't change the spelling of the levels like that because they're part of the public API documented in the devguide. Maybe add optional boolean argument to `get_as_string` to capitalize the name? I'm also fine with the non-static method called a static common function.

mitza-oci (approved):

iguessthislldo (approved):

PR #4400, opened 2023-12-19
Improve SpdpSendAddrs parsing

smnrgrs:
For large deployments with no multicast discovery, the current SpdpSendAddrs syntax in the ini file is cumbersome. Allow multiple ports per host to be easily specified.

Each SpdpSendAddrs entry can now be:
<addr>:<port>

or
<addr>:<base_port>:<num>  (defaults to interval = 2)

or
<addr>:<base_port>:<num>:<port_interval>

Example:
SpdpSendAddrs=100.11.13.15:13910:16:2
This will specify 16 endpoints on 100.11.13.15, starting from port 13910, counting up two at a time


jrw972:
@smnrgrs See the build and test failures.

PR #4411, opened 2024-01-05
Discovery templates are not reused

jrw972:
Problem
-------

One of the goals of using the ConfigStore (#4134) is to minimize the number of entries in the ConfigStore that are generated by the framework itself.  Such entries are basically leaked.

The main culprit here is template support which
instatiates new `[domain]`, `[rtps_discovery]`, `[config]`, and `[transport]` sections.  This commit removes the need to instantiate `[rtps_discovery]` sections.

Solution
--------

Parameterize `RtpsDiscoveryConfig::default_multicast_group` on the domain so that it can be customized on the fly.

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscovery.cpp
+++ b/dds/DCPS/RTPS/RtpsDiscovery.cpp
@@ -41,125 +41,102 @@ RtpsDiscovery::~RtpsDiscovery()
 {
 }
 
-namespace {
-  const ACE_TCHAR RTPS_SECTION_NAME[] = ACE_TEXT("rtps_discovery");
-}
-
 int
-RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
+RtpsDiscovery::Config::discovery_config()
 {
-  const ACE_Configuration_Section_Key &root = cf.root_section();
-  ACE_Configuration_Section_Key rtps_sect;
+  RcHandle<DCPS::ConfigStoreImpl> config_store = TheServiceParticipant->config_store();
 
-  if (cf.open_section(root, RTPS_SECTION_NAME, 0, rtps_sect) == 0) {
+  const DCPS::ConfigStoreImpl::StringList sections = config_store->get_section_names("OPENDDS_RTPS_DISCOVERY");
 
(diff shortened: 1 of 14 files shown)


dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -46,120 +46,101 @@ namespace {
...
 
 int
-RtpsDiscovery::Config::discovery_config(ACE_Configuration_Heap& cf)
+RtpsDiscovery::Config::discovery_config()
 {
-  const ACE_Configuration_Section_Key &root = cf.root_section();
-  ACE_Configuration_Section_Key rtps_sect;
+  RcHandle<DCPS::ConfigStoreImpl> config_store = TheServiceParticipant->config_store();
 
-  if (cf.open_section(root, RTPS_SECTION_NAME, 0, rtps_sect) == 0) {
```
iguessthislldo, line 54: `RTPS_SECTION_NAME` above is unused now.

dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
```diff
@@ -391,13 +391,51 @@ RtpsDiscoveryConfig::multicast_interface(const OPENDDS_STRING& mi)
...
+                                               DCPS::NetworkAddress("239.255.0.1:0"),
+                                               DCPS::ConfigStoreImpl::Format_No_Port,
+                                               DCPS::ConfigStoreImpl::Kind_IPV4);
+  // Customize.
+  const String customization_name = TheServiceParticipant->config_store()->get(config_key("CUSTOMIZATION").c_str(), "");
+  if (!customization_name.empty()) {
+    const String directive = TheServiceParticipant->config_store()->get(String("OPENDDS_CUSTOMIZATION_" + customization_name + "_INTEROP_MULTICAST_OVERRIDE").c_str(), "");
+    if (directive == "AddDomainId") {
+      String addr = DCPS::LogAddr(na, DCPS::LogAddr::Ip).str();
+      size_t pos = addr.find_last_of(".");
```
iguessthislldo, line 408:
```suggestion
      const size_t pos = addr.find_last_of(".");
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4419, opened 2024-01-11
BUGREPORT: ignored durability_service field in topic QoS leads to uninitialized structure fields

paocela:
[PROBLEM-REPORT-FORM.txt](https://github.com/OpenDDS/OpenDDS/files/13901799/PROBLEM-REPORT-FORM.txt)


mitza-oci:
Posting attachment from @paocela as comment body:

This is OpenDDS version 3.27.0-dev (NOT A RELEASE)

TAO VERSION: 6.5.20
HOST MACHINE and OPERATING SYSTEM: 6.1.46-gentoo-x86_64
COMPILER NAME AND VERSION (AND PATCHLEVEL): gcc (Gentoo 12.3.1_p20230526 p2) 12.3.1 20230526

CONTENTS OF $ACE_ROOT/ace/config.h:

    #define ACE_DISABLE_MKTEMP
    #define ACE_DISABLE_READDIR_R
    #define ACE_DISABLE_TEMPNAM
    #define TAO_HAS_UIOP 0
    #include "ace/config-linux.h"

    optimize = 0
    include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU

    no_cxx11=0
    no_rapidjson=0
    no_rapidjson=0
    no_rapidjson=0
    xerces3=1
    no_rapidjson=0
    xerces3=1

    MEDIUM

AREA/CLASS/EXAMPLE AFFECTED:
    None

DOES THE PROBLEM AFFECT:
COMPILATION? NO
LINKING? NO
EXECUTION?  YES, my application is affected

SYNOPSIS:
When using QoS policies via XML file, specifically for the topic policies, the  ``durability_service`` field is ignored by the QOS_XML_Handler component. This results in passing to the topic a ``TopicQos`` structure where the ``durability_service`` field is uninitialized, causing a failure in the topic's creation.

REPEAT BY:
    Running test at ``tests/DCPS/TestDurabilityServiceTopic``

SAMPLE FIX/WORKAROUND:
The problem seems to lie in the use of function ``DwDrTpBase::read_qos`` instead of function ``DwTpBase::read_qos`` in file ``dds/DCPS/QOS_XML_Handler/QOS_Topic_T.cpp:23``. If instead the documentation wrong, it should be updated, together with the example in ``tests/DCPS/QoS_XML/dump`` (where the qos.xml file was initially found)

jwillemsen:
It has been very long ago that I worked on this. Very important is that you need to first retrieve the default QoS for your DDS entity from DDS, the QoS XML Handler will overwrite the ones set, so you can't just pass in an empty QoS as you do in your test, see AXCIOMA at https://github.com/RemedyIT/ciaox11/blob/df06afbe574946d9354225d96919a999823cab9c/ddsx11/vendors/opendds/dds/opendds_domain_participant_factory.cpp#L70 for a similar use case

paocela:
Thanks for the quick reply. 

Yes, that definitely solves uninitialized elements and correct DDS entity creation (topic in my case), but it still doesn't help me passing all policies specified in the qos.xml file (in my case the `durability_service` one). For that one it can only be the default one, or a user defined one manually inserted inside the structure, in no case one populated by the QoS XML Handler, correct?

jwillemsen:
durability_service is datawriter/topic, so handling it in QOS_DwTp_Base_T is correct, but very likely QOS_Topic_T should derive from QOS_DwTp_Base_T. Looks the first step is to extend tests/DCPS/QOS_XML/dump with durability_service in the xml and validate that at topic level

paocela:
> durability_service is datawriter/topic, so handling it in QOS_DwTp_Base_T is correct, but very likely QOS_Topic_T should derive from QOS_DwTp_Base_T. 

This is what I did to temporarely fix the problem, and it seems to work (at least for the durability_service)

> Looks the first step is to extend tests/DCPS/QOS_XML/dump with durability_service in the xml and validate that at topic level

See last commit, it fails as expected. 
My concern is that other fields might be affected as well because the test seems incomplete. I just investigated the one raising an issue for me.

PR #4421, opened 2024-01-11
Add an XCDR2 Value Writer

sonndinh:
- Added `Xcdr2ValueWriter` that has two operation modes: one for computing the total serialized size of a sample and its component sizes needed to fill its headers, and another for serializing the sample using the cached sizes. This value writer can be passed to the generated `vwrite` to serialize a C++ object or the `vwrite` for dynamic data to serialize a dynamic data object using XCDR2 encoding.
- Moved the `vwrite` function for dynamic data to its own files. Updated it to handle optional members that are not present and enabled bulk write for primitive arrays and sequences.
- Fixed a bug in the generated `vwrite` for enum that can cause name conflicts.
- Added utility for flattening the indexes to a multi-dimensional array.


dds/DCPS/Xcdr2ValueWriter.h
```diff
@@ -0,0 +1,187 @@
...
+  void write_float64_array(const ACE_CDR::Double* x, size_t length);
+  void write_float128_array(const ACE_CDR::LongDouble* x, size_t length);
+  void write_char8_array(const ACE_CDR::Char* x, size_t length);
+  void write_char16_array(const ACE_CDR::WChar* x, size_t length);
+
+  size_t get_serialized_size() const;
+  const std::vector<size_t>& get_serialized_sizes() const;
+
+  // Reset the operation mode to only computing serialization size.
+  void reset()
```
mitza-oci, line 87: What's the expected use case for `reset()`?  Is it used in tests?

mitza-oci (commented):
This is a partial review

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  DDS::MemberDescriptor_var disc_md;
+  rc = dtm->get_descriptor(disc_md);
+  if (rc != DDS::RETCODE_OK) {
+    if (log_level >= LogLevel::Warning) {
+      ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: vwrite_union: get_descriptor failed (%C)\n",
+                 retcode_to_string(rc)));
+    }
+    return;
+  }
+  // Discriminator Id is always 0?
```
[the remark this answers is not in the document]
sonndinh: Discriminator doesn't have Id in TypeObject and when converting to DynamicType, we use DISCRIMINATOR_ID as a special Id used to access it. When serializing the union, it needs to get the actual Id to write the Emheader but we don't have that. I think we're using the Id of 0 everywhere for the discriminator for the purpose of serialization right now.

I'm not sure if it's specified in the IDL spec, but it seems to me discriminator is different from the regular members since it's always the first thing in the union, and so an Id is not required. Maybe that's why TypeObject for unions has a separate field for discriminator that doesn't contain Id. If that's the case, we can use an arbitrary value for the discriminator Id when serializing.

There's a related spec issue to this: https://issues.omg.org/issues/DDSXTY14-59.

dds/DCPS/Xcdr2ValueWriter.h
sonndinh: The intended use case is to allow reusing the value writer object for serializing a different object. The usage would be like:
```
Xcdr2ValueWriter vw(xcdr2); // The default mode is computing serialized size
Sample sample;
vwrite(vw, sample); // Compute the sizes.
vw.set_serializer(ser); // Switch to serialization mode
vwrite(vw, sample); // Serialize the sample
...
Sample sample2;
vw.reset(); // Switch back to the serialized size mode
vwrite(vw, sample2);
vw.set_serializer(ser2);
vwrite(vw, sample2);
```
I think I don't have a test for it yet. I can add one.

dds/DCPS/XTypes/DynamicVwrite.cpp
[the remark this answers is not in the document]
mitza-oci: It looks like in the generated code this is covered by:
```
if (!strm.write_parameter_id(0, size)) {
```
I suggest changing the `0` for both generated and dynamic cases to a named constant which can have comment describing if/how the spec is lacking.

dds/DCPS/Xcdr2ValueWriter.h
mitza-oci: There are a few options: 
- no reset method, user must create a new object
- no reset method needed since it automatically changes back to the "reset" state
- provide a reset method
- others?

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1042 @@
...
+  vw.begin_struct(extensibility);
+  for (CORBA::ULong i = 0; i < dt->get_member_count(); ++i) {
+    DDS::DynamicTypeMember_var dtm;
+    rc = dt->get_member_by_index(dtm, i);
+    if (rc != DDS::RETCODE_OK) {
+      if (log_level >= LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: vwrite_struct: get_member_by_index %u failed (%C)\n",
+                   i, retcode_to_string(rc)));
+      }
+      return;
```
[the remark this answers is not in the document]
sonndinh: As the first step, I think the value writer interface should change to return something like `bool` for every method now.
When an error like this happens, the `Xcdr2ValueWriter` object has an invalid state because it expects the matching `end_struct`, and so we can't reuse it.

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1042 @@
...
+      if (log_level >= LogLevel::Warning) {
+        ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: vwrite_discriminator: enum_bound failed (%C)\n",
+                   retcode_to_string(rc)));
+      }
+      return;
+    }
+  }
+
+  switch (treat_disc_as) {
+  case TK_BOOLEAN: {
```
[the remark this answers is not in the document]
sonndinh: Possibly, but I'd prefer to keep them separate. `write_item` includes calls to either the `begin_struct_member`/`end_struct_member` pair or the `begin_union_member`/`end_union_member` pair, and it also handles more types than the types allowed for the discriminator. I think merging it to `write_item` will make it more convoluted and harder to understand.

dds/DCPS/ValueWriter.h
```diff
@@ -24,6 +25,47 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
+struct VWriterMemberParam {
```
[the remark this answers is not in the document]
sonndinh: I haven't thought through how the value reader will look like yet. I think it can skip `must_understand` and `optional` because the caller probably has it already from the type info.

dds/DCPS/Xcdr2ValueWriter.h
sonndinh: I forgot to mention that right now, if the object is only used in the serialized size mode, then it doesn't need to call `reset`. This should be fine:
```
Xcdr2ValueWriter vw(xcdr2);
Sample sample;
vwrite(vw, sample);
vw.get_serialized_size(); // size of sample
Sample sample2;
vwrite(vw, sample2);
vw.get_serialized_size(); // size of sample2
```
We only need to call `reset` if the mode changes.

dds/DCPS/XTypes/DynamicVwrite.cpp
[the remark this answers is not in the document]
jrw972: Returning a bool makes sense.

dds/DCPS/ValueWriter.h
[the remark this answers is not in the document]
jrw972: I lean toward `MemberParam`.

dds/DCPS/ValueWriter.h
```diff
@@ -40,23 +82,23 @@ class OpenDDS_Dcps_Export ValueWriter {
   ValueWriter()  {}
   virtual ~ValueWriter() {}
 
-  virtual void begin_struct() {}
+  virtual void begin_struct(Extensibility /*extensibility*/) {}
   virtual void end_struct() {}
-  virtual void begin_struct_member(const DDS::MemberDescriptor& /*descriptor*/) {}
+  virtual void begin_struct_member(VWriterMemberParam /*params*/) {}
```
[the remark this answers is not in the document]
sonndinh: `VWriterMemberParam` is pretty lightweight so I think it's fine to pass by value so that we can be free of any invalid reference issue.

mitza-oci (changes requested):
another partial review -- I still haven't looked at arrays (single or multi dimensional)

dds/DCPS/XTypes/Utils.h
```diff
@@ -238,6 +240,27 @@ inline DCPS::Sample::Extent nested(DCPS::Sample::Extent ext)
   return ext == DCPS::Sample::KeyOnly ? DCPS::Sample::NestedKeyOnly : ext;
 }
 
+// Convert the index vector to an element in a multi-dimensional array into a flat index.
+// See description for ARRAY_TYPE in XTypes 1.3, page 139.
+DDS::ReturnCode_t flat_index(CORBA::ULong& flat_idx, const DDS::BoundSeq& idx_vec,
+                             const DDS::BoundSeq& dims, CORBA::ULong up_to_ndims);
```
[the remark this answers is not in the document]
sonndinh: No, it is only called internally. It was intended to get the index of a subarray in a multi-dimensional array represented by a dynamic data object. But we're going with a different approach now and there is no need for it anymore. I think I can delete this and only keep the one with 3 parameters.

iguessthislldo (changes requested):

dds/DCPS/Xcdr2ValueWriter.cpp
```diff
@@ -0,0 +1,836 @@
...
+    return *ser_ << value;
+  }
+}
+
+bool Xcdr2ValueWriter::write_wstring(const ACE_CDR::WChar* value, size_t length)
+{
+  if (mode_ == SERIALIZATION_SIZE_MODE) {
+    size_t& size = size_states_.top().total_size;
+    primitive_serialized_size_ulong(encoding_, size);
+    if (value) {
```
mitza-oci, line 606: If value is null isn't this an error?  Not sure if it's one that needs to be detected here.

dds/DCPS/Xcdr2ValueWriter.cpp
```diff
@@ -0,0 +1,836 @@
...
+}
+
+// Array and sequence can be treated similar to final/appendable struct.
+// If the element type is primitive, it is similar to a final struct with no delimiter.
+// If the element type is not primitive, it is similar to an appendable struct with a delimiter.
+// In both case, elements are serialized back-to-back similarly to final/appendable struct.
+// One difference is that sequence has a length field ahead of the elements.
+bool Xcdr2ValueWriter::begin_array(XTypes::TypeKind elem_tk)
+{
+  // TODO: Revisit for multi-dimensional array with typedef.
```
mitza-oci, line 309: Does the belong to Xcdr2ValueWriter, DynamicVwrite, or both?

mitza-oci (changes requested):

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -1680,6 +1680,35 @@ bool has_explicit_keys(DDS::DynamicType* dt)
...
+  }
+
+  CORBA::ULong ret_val = 0;
+  CORBA::ULong factor = 1;
+  for (CORBA::ULong i = dims.length() - 1; i > 0; --i) {
+    const CORBA::ULong dim = dims[i];
+    if (idx_vec[i] >= dim) {
+      if (DCPS::log_level >= DCPS::LogLevel::Notice) {
+        ACE_ERROR((LM_WARNING, "(%P|%t) WARNING: flat_index: Index %u-th is invalid for"
+                   " the %u-th dimension (%u)", i, i, dim));
```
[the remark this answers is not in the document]
sonndinh: Yes. I tweaked it a bit to make it more clear.

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1488 @@
...
+    }
+  }
+
+  return vw.end_array();
+}
+
+DDS::ReturnCode_t vwrite_primitive_collection(
+  ValueWriter& vw, DDS::DynamicData_ptr value, DDS::MemberId id,
+  XTypes::TypeKind elem_kind, XTypes::TypeKind orig_elem_kind, XTypes::TypeKind col_tk,
+  const DDS::BoundSeq& dims, XTypes::TypeKind containing_tk, MemberParam* params)
```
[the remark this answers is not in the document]
sonndinh: No, because `params` is updated in a call to `begin_member_helper`.

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1488 @@
...
+                   retcode_to_string(rc)));
+      }
+      return rc;
+    }
+  }
+  return DDS::RETCODE_OK;
+}
+
+bool vwrite_primitive_array_helper(
+  ValueWriter& vw, void* buffer, XTypes::TypeKind elem_kind, XTypes::TypeKind orig_elem_kind,
```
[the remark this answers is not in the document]
sonndinh: Yes, using a template made the code more concise.

dds/DCPS/Xcdr2ValueWriter.cpp
sonndinh: Yes, I think it's an error. `DynamicDataImpl` hasn't checked for a null string in `set_string_value` and `set_wstring_value`. I think it should do that.

dds/DCPS/XTypes/DynamicVwrite.cpp
```diff
@@ -0,0 +1,1488 @@
...
+{
+  using namespace OpenDDS::XTypes;
+  const DDS::TypeKind item_tk = item_type->get_kind();
+  DDS::TypeKind treat_as;
+  if (get_equivalent_kind(item_type, treat_as) != DDS::RETCODE_OK) {
+    return false;
+  }
+
+  DDS::ReturnCode_t rc = DDS::RETCODE_OK;
+  switch (treat_as) {
```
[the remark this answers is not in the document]
sonndinh: I think the bits in a bitmask have names, so it makes sense to have an equivalent to `write_enum`. I think we haven't seen bitmasks a lot, so it was neglected. I can add something like `write_bitmask` to the `ValueWriter` and its derived classes if we agree to do it here.

dds/DCPS/ValueCommon.h
```diff
@@ -0,0 +1,173 @@
...
+OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
+
+namespace OpenDDS {
+namespace DCPS {
+
+class EnumHelper {
+public:
+  virtual ~EnumHelper() {}
+  virtual bool get_value(ACE_CDR::Long& value, const char* name) const = 0;
+  virtual bool get_name(const char*& name, ACE_CDR::Long value) const = 0;
```
mitza-oci, line 22: Since we're going to have `optional` available at some point in the not-too-distant future (either from `std::` or our own), this would be a good use of it.
Doesn't need to be in this PR.

dds/DCPS/ValueCommon.h
```diff
@@ -0,0 +1,173 @@
...
+
+  bool get_value(ACE_CDR::Long& value, const char* name) const
+  {
+    for (OPENDDS_VECTOR(Pair)::const_iterator it = pairs_.begin(); it != pairs_.end(); ++it) {
+      if (std::strcmp(it->name, name) == 0) {
+        value = it->value;
+        return true;
+      }
+    }
+    return false;
```
mitza-oci, line 58: This seems like a lot to have inlined in the header file.

dds/DCPS/ValueCommon.h
```diff
@@ -0,0 +1,173 @@
...
+        return false;
+      }
+      rtn |= 1 << it->second;
+    }
+    value = rtn;
+    return true;
+  }
+
+  // Return an estimated length of a string constructed by the returned flag names
+  // with a delimiter character between two consecutive flags.
```
mitza-oci, line 140: Move this comment to the base class definition (assuming it applies to the interface and not the implementation)

dds/DCPS/ValueWriter.h
```diff
@@ -189,7 +195,7 @@ class OpenDDS_Dcps_Export ValueWriter {
   ///@}
 
 protected:
-  std::string bit_string(ACE_CDR::ULongLong value, ACE_CDR::ULong bitbound);
+  String bitflag_string(ACE_CDR::ULongLong value, const BitmaskHelper& helper);
```
mitza-oci, line 198: Could the name of the function be `bitmask_string` or `bitmask_to_string`?

dds/DCPS/ValueWriter.h
```diff
@@ -154,14 +155,19 @@ class OpenDDS_Dcps_Export ValueWriter {
...
-  bool write_enum(const char* name, const T& value, XTypes::TypeKind as_int = XTypes::TK_INT32)
+  bool write_enum(const T& value, const EnumHelper& helper)
   {
-    return write_enum(name, static_cast<ACE_CDR::Long>(value), as_int);
+    return write_enum(static_cast<ACE_CDR::Long>(value), helper);
   }
 
-  virtual bool write_bitmask(ACE_CDR::ULongLong value, ACE_CDR::ULong bitbound) = 0;
+  virtual bool write_bitmask(ACE_CDR::ULongLong value, const BitmaskHelper& helper) = 0;
+  template <typename T>
```
mitza-oci, line 166:
```suggestion
  virtual bool write_bitmask(ACE_CDR::ULongLong value, const BitmaskHelper& helper) = 0;

  template <typename T>
```

mitza-oci (changes requested):

dds/DCPS/ValueWriter.h
sonndinh: `bitmask_to_string` sounds good to me.

dds/DCPS/ValueCommon.h
sonndinh: Can you elaborate on how `optional` can be used here?

dds/DCPS/ValueCommon.h
```diff
@@ -99,59 +84,25 @@ class MapBitmaskHelper : public BitmaskHelper {
   typedef PosToNameMap::const_iterator ptn_iterator;
   typedef NameToPosMap::const_iterator ntp_iterator;
 
-  MapBitmaskHelper(ACE_CDR::UShort bit_bound, XTypes::TypeKind as_uint)
-    : bit_bound_(bit_bound)
-    , as_uint_(as_uint)
+  MapBitmaskHelper(XTypes::TypeKind as_uint)
```
mitza-oci, line 87:
```suggestion
  explicit MapBitmaskHelper(XTypes::TypeKind as_uint)
```

mitza-oci (changes requested):

iguessthislldo (approved):

dds/DCPS/ValueCommon.h
```diff
@@ -0,0 +1,124 @@
...
+
+class EnumHelper {
+public:
+  virtual ~EnumHelper() {}
+  virtual bool get_value(ACE_CDR::Long& value, const char* name) const = 0;
+  virtual bool get_name(const char*& name, ACE_CDR::Long value) const = 0;
+  virtual XTypes::TypeKind get_equivalent_int() const = 0;
+};
+
+class OpenDDS_Dcps_Export ListEnumHelper : public EnumHelper {
```
jrw972, line 26: Add a unit test for this class.

dds/DCPS/ValueCommon.h
```diff
@@ -0,0 +1,124 @@
...
+  virtual bool get_value(ACE_CDR::ULongLong& value, const OPENDDS_VECTOR(const char*)& names) const = 0;
+
+  // Return an estimated length of a string constructed by the returned flag names
+  // with a delimiter character between two consecutive flags.
+  virtual size_t get_names(OPENDDS_VECTOR(const char*)& names, ACE_CDR::ULongLong value) const = 0;
+
+  virtual XTypes::TypeKind get_equivalent_uint() const = 0;
+};
+
+class OpenDDS_Dcps_Export MapBitmaskHelper : public BitmaskHelper {
```
jrw972, line 75: Add a unit test for this class.

dds/DCPS/ValueWriter.h
```diff
@@ -40,100 +83,125 @@ class OpenDDS_Dcps_Export ValueWriter {
...
+  virtual bool write_uint64_array(const ACE_CDR::ULongLong* x, ACE_CDR::ULong length);
+  virtual bool write_float32_array(const ACE_CDR::Float* x, ACE_CDR::ULong length);
+  virtual bool write_float64_array(const ACE_CDR::Double* x, ACE_CDR::ULong length);
+  virtual bool write_float128_array(const ACE_CDR::LongDouble* x, ACE_CDR::ULong length);
+  virtual bool write_char8_array(const ACE_CDR::Char* x, ACE_CDR::ULong length);
+  virtual bool write_char16_array(const ACE_CDR::WChar* x, ACE_CDR::ULong length);
   ///@}
+
+protected:
+  String bitmask_to_string(ACE_CDR::ULongLong value, const BitmaskHelper& helper);
```
jrw972, line 200: I don't think this should be a concern of `ValueWriter`.

jrw972 (changes requested):

jrw972 (approved):

iguessthislldo (approved):

mitza-oci (approved):

PR #4425, opened 2024-01-12
Make seconds of Time_t 64bit

### LINKED ISSUE #4365: Update DDS::Time_t and DDS::Duration_t for 64bit
The OMG DDS issue for 64bit time has a comment that the RTF agreed to make seconds 64bit, see https://issues.omg.org/browse/DDS15-13. Would be nice when OpenDDS would support this 64bit time/duration

jwillemsen:
Fixed #4365

### DIFF
--- a/dds/DCPS/Time_Helper.inl
+++ b/dds/DCPS/Time_Helper.inl
@@ -126,7 +126,7 @@ operator>=(const DDS::Time_t& t1, const DDS::Time_t& t2)
 ACE_INLINE DDS::Time_t
 operator+(const DDS::Time_t& t1, const DDS::Duration_t& d1)
 {
-  CORBA::Long sec = static_cast<CORBA::Long>(static_cast<CORBA::ULong>(t1.sec) + static_cast<CORBA::ULong>(d1.sec));
+  CORBA::LongLong sec = t1.sec + d1.sec;
   CORBA::ULong nanosec = t1.nanosec + d1.nanosec;
 
   while (nanosec >= ACE_ONE_SECOND_IN_NSECS) {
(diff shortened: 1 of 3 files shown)


jrw972:
This will break the interoperability of applications that use `DDS::Time_t`.  We will definitely make this change for OpenDDS 4.  Can you wait for OpenDDS 4?

jwillemsen:
This could break portability, interoperability is handled by RTPS and there is another different definition on time. The risk is very small, all Time_t is created by the OpenDDS core and passed to user code where they can use it. At the moment they use Time_t as type nothing will happen, the only risk is when they use another 32bit type to store it, at that moment they would get a warning, easily fixed by changing to 64bit also or add a cast. This is a break in portability we should do asap to my idea, not wait because we maybe cause a warning somewhere for just a few users.

jrw972:
The risk is for applications that use `DDS::Time_t` like `tools/ dds/rtpsrelay/Relay.idl`.

iguessthislldo:
We have added this to [the roadmap under OpenDDS 4.0](https://github.com/OpenDDS/OpenDDS/wiki/Roadmap).

jwillemsen:
> There are `LongLong` to `Long` warnings from MSVC in:
> 
> For DataReader, DataWriter, and Replayer, it looks like this should be changed to 64 bits to match `Time_t`:
> 
> https://github.com/OpenDDS/OpenDDS/blob/bcdc249199ca63fe0d8ab4f0bfdcf97a91061a0f/dds/DCPS/DataSampleHeader.h#L160

Wouldn't updating DataSampleHeader break interoperability, I see serialization in this code, I think at RTPS layer the second field stays 32bit, correct? Something special required here?

PR #4440, opened 2024-01-23
`TransportInst::load` is unnecessary

jrw972:
Problem
-------

`TransportInst::load` is no longer necessary and can be removed.

Solution
--------

Remove `TransportInst::load`.


tests/DCPS/MultiDPTest/subscriber.cpp
```diff
@@ -110,21 +112,14 @@ void init_dcps_objects(int i)
       ACE_TCHAR nak_depth[8];
```
iguessthislldo, line 112:
```suggestion
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4451, opened 2024-01-30
`SedpLocalAddress` does not respect `DCPSDefaultAddress`

jrw972:
Problem
-------

`SedpLocalAddress` does not use `DCPSDefaultAddress` as a default. This prevents restricting all RTPS traffic to a specific network interface by setting `DCPSDefaultAddress`.  The user can set the local address for everyting if they can't use `DCPSDefaultAddress`.

Solution
--------

Update `SedpLocalAddress` to work like `SpdpLocalAddress`.


dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
```diff
@@ -321,9 +321,12 @@ RtpsDiscoveryConfig::sedp_multicast(bool sm)
 DCPS::NetworkAddress
 RtpsDiscoveryConfig::sedp_local_address() const
 {
+  const DCPS::NetworkAddress d1(TheServiceParticipant->default_address());
+  const DCPS::NetworkAddress d2("0.0.0.0:0");
+
   return TheServiceParticipant->config_store()->get(config_key("SEDP_LOCAL_ADDRESS").c_str(),
-                                                    DCPS::NetworkAddress::default_IPV4,
-                                                    DCPS::ConfigStoreImpl::Format_Required_Port,
+                                                    d1 ? d1 : d2,
```
mitza-oci, line 328: Why not use default_IPv4?

jrw972: Are you asking if `d2` can be initialized to `default_IPV4`?

mitza-oci: Either that or just use the expression `DCPS::NetworkAddress::default_IPV4` in place of d2 since that's what the code did before this change and it seems to be equivalent.

iguessthislldo (approved):

iguessthislldo:
The [devguide entry](https://github.com/OpenDDS/OpenDDS/blob/30553135c68a55321d9c63158879a224c04d0c3f/docs/devguide/run_time_configuration.rst#L1051) and probably the news should be updated to reflect this change.

PR #4472, opened 2024-02-14
Post 3.27.0 Release Script Improvements

iguessthislldo:
- Replace `--upload-shapes-demo` with `--upload-artifacts`, which will now upload the RTPS interop test in addition to the shapes demo.
- `--upload-shapes-demo`, now `--upload-artifacts` was unable to download artifacts from workflows. [This GitHub discussion answer](https://github.com/orgs/community/discussions/88698#discussioncomment-8252538) provided an overview of the issue. The artifact download URL was being redirected to an internal Microsoft one which rejects GitHub authentication. The solution was to use the `download` function instead of PitHub and make a special case where it detects the failed redirect and retries it without authentication.
- Replace the multiple `die_with_stack_trace` functions with a single implementation called `trace`. Also fixed the confusing off-by-one error in the output.
- Replaced invalid usage of `$?` with `$!`.

### DIFF
--- a/docs/internal/release.rst
+++ b/docs/internal/release.rst
@@ -376,7 +376,7 @@ Upload Artifacts from Release Workflows
 
     This should only be done for the :ref:`latest release <release-latest-release>`.
 
-During the release script there are steps that trigger :ref:`release workflows <release-workflows>` on GitHub Actions and print out links to the runs.
+The release script has steps that trigger :ref:`release workflows <release-workflows>` on GitHub Actions and print out links to the runs.
 After they have finished successfully, run the release script with the version and workspace arguments and the ``--upload-artifacts`` option.
 If the workflows are still in progress it will say so and give the links again.
 If the workflows were successful, it will download the artifacts, package them, and upload them to GitHub.
--- a/docs/internal/release.rst
+++ b/docs/internal/release.rst
@@ -161,6 +161,8 @@ To do this follow these steps:
 
 The release script will merge ``website-next-release`` into ``gh-pages`` on the ``OpenDDS/OpenDDS`` repository during the release process.
 
+.. _release-workflows:
+
 Check if GHA Workflows need Updating
 ====================================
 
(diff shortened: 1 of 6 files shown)


tools/scripts/gitrelease.pl
```diff
@@ -286,25 +277,46 @@ sub create_dummy_release_file {
...
 sub new_pithub {
   my $settings = shift();
+  my %args = @_;
+  my $needs_token = $args{needs_token} // 0;
+  my $user = $args{user};
+  my $repo = $args{repo};
+
+  if (defined($user) || defined($repo)) {
+    # We need to create another Pithub object, make a copy of settings for this.
+    $settings = {%{$settings}};
```
mitza-oci, line 294: I don't know if I've seen this Perl construct before. Is it the same as `$settings = \%{$settings};` (if that even works)?
May also be good to use two different variables.

docs/internal/release.rst
```diff
@@ -208,8 +210,10 @@ Before Running the Release Script
     - Git version 2.5 or later
     - :ref:`Python 3 for News Generation <docs-requirements>`
 
-  - Your GitHub account has been added as a member of the `OpenDDS organization <https://github.com/OpenDDS>`__ with the appropriate permissions.
+  - Your GitHub account meets the following requirements:
 
+    - It has been added as a member of the `OpenDDS organization <https://github.com/OpenDDS>`__ with the appropriate permissions.
+    - It has permissions to update the release artifacts for the `OMG RTPS Interop repo <https://github.com/omg-dds/dds-rtps>`__ for :ref:`release-upload-artifacts`.
```
mitza-oci, line 216: If the release script command-line arguments indicate that the user want to try uploading, but the user doesn't actually have permissions, does the release continue with a warning or error out?

jrw972 (approved):

tools/scripts/gitrelease.pl
iguessthislldo: I'm not sure what `$settings = \%{$settings};` is supposed to do, but here I'm dereferencing settings and putting the values in a new hash. I don't think I can make two separate variables because usage of copied settings is conditional.

docs/internal/release.rst
iguessthislldo: This is happening after the release proper and it will error when it tries to upload to GitHub.

mitza-oci (approved):

iguessthislldo:
This needs `--help` changed and probably would be good if could detect if one the artifacts was already uploaded. Right now it gives an error if it tries to upload an artifact again.

mitza-oci (approved):

PR #4475, opened 2024-02-14
`load_transport_configuration` uses `ACE_Configuration_Heap`

jrw972:
Problem
-------

`load_transport_configuration` uses `ACE_Configuration_Heap` and `ACE_Configuration_Heap` is being removed where possible as part of #4134.

Solution
--------

Simplify the arguments to only have the file name.

### DIFF
--- a/dds/DCPS/transport/framework/TransportRegistry.cpp
+++ b/dds/DCPS/transport/framework/TransportRegistry.cpp
@@ -86,236 +86,122 @@ TransportRegistry::TransportRegistry()
   lib_directive_map_["repository"] = "dynamic OpenDDS_InfoRepoDiscovery Service_Object * OpenDDS_InfoRepoDiscovery:_make_IRDiscoveryLoader()";
 }
 
-int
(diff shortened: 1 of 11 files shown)


dds/DCPS/transport/framework/TransportRegistry.cpp
```diff
@@ -86,236 +86,122 @@ TransportRegistry::TransportRegistry()
...
+       pos != limit; ++pos) {
+    TransportInst_rch inst = get_inst(*pos);
+    if (!inst) {
+      if (log_level >= LogLevel::Error) {
+        ACE_ERROR((LM_ERROR,
+                   "(%P|%t) TransportRegistry::load_transport_configuration: "
+                   "The inst (%C) in [config/%C] section is undefined.\n",
+                   pos->c_str(), config->name().c_str()));
+      }
+      return -1;
```
iguessthislldo, line 160:
```suggestion
      return false;
```

iguessthislldo (changes requested):

PR #4487, opened 2024-02-23
Allow Compiling with CMake and C++03

iguessthislldo:
This continues the work started in https://github.com/OpenDDS/OpenDDS/pull/4481 and tries to correctly handle C++ standards in all CMake support. Previously the CMake support was agnostic towards C++ standards, but this meant that building with CMake didn't have a way of differentiating C++ standards. Now we can detect C++ standard support and should be able to build ACE accordingly.

Also:
- Added `OPENDDS_CXX_STD_YEAR` CMake variable for easy checking of C++ standard support.
- In addition to passing `CMAKE_CXX_STANDARD` to the ACE/TAO MPC build, it will also now pass `CMAKE_CXX_COMPILER`.
- Removed need for `cmake/detect_ace.pl`
- Some of the changes originally in https://github.com/OpenDDS/OpenDDS/pull/4374
  - MPC location message will now say when it wasn't set.
  - Split adding of tests between different directories.


CMakeLists.txt
```diff
@@ -102,7 +102,7 @@ if(OPENDDS_XERCES3)
 endif()
 if(OPENDDS_SECURITY)
   add_subdirectory(dds/DCPS/security)
-  if(OPENDDS_BUILT_IN_TOPICS)
+  if(OPENDDS_BUILT_IN_TOPICS AND NOT OPENDDS_CXX_STD_YEAR LESS 2011)
     add_subdirectory(tools/dds/rtpsrelaylib)
     add_subdirectory(tools/rtpsrelay)
```
mitza-oci, line 107: I don't mind this strategy but I think we should consider (maybe for a future PR) adding all the subdirectories and letting per-target properties determine what gets built

cmake/configure_ace_tao.pl
```diff
@@ -100,6 +101,12 @@ sub read_file {
   my $platform_macros_path = "$values{ace}/include/makeinclude/platform_macros.GNU";
   open(my $platform_macros_file, '>', $platform_macros_path)
     or die("Failed to open $platform_macros_path: $!");
+  if ($values{compiler}) {
+    for my $var ('CC', 'CXX', 'LD') {
+      print $platform_macros_file ("$var = $values{compiler}\n");
+    }
+  }
+  print $platform_macros_file ("CCFLAGS += \$(opendds_cmake_std)\n");
```
mitza-oci, line 109: These look kind of odd with a parenthesized expression that's just a single string

CMakeLists.txt
iguessthislldo: This pattern comes from the CMake tests. I think the rational was if you wanted to configure a project individually, it would check the requirements and fail to configure. You can't really configure the non-test OpenDDS projects individually though, so maybe... The only issue I have with it is I'd like the tests current setup to remain because I think it definitely makes sense for the tests. So having having them act differently might be confusing.

cmake/configure_ace_tao.pl
iguessthislldo: A bit, but it's how Perl requires parentheses to be used with `print`. It's not much different looking from `foreach` loops though.

CMakeLists.txt
mitza-oci: I think it would be easier to manage if it were consistently done on a target/properties evaluation basis and not imperative logic around which subdirectories are added.  A variable that's something like `BUILD_THE_TESTS` is very clearly tied to `add_subdirectory(tests)` so that's fine.

cmake/configure_ace_tao.pl
mitza-oci: I guess was expecting no parens when they don't add meaning or clarity, but it's subjective.  Perl is TMTOWTDI to a fault.

PR #4489, opened 2024-02-26
Make Generated Output in CMake More Robust

iguessthislldo:
The generated output functions in CMake didn't have any sort of testing and could give incorrect results for where to put the export header file. This tries to correct both of those issues.

Also added workaround for https://github.com/DOCGroup/ACE_TAO/issues/2202


mitza-oci (approved):

cmake/opendds_target_sources.cmake
```diff
@@ -216,7 +216,7 @@ function(opendds_export_header target)
     return()
   endif()
 
-  # Need come before configure_file
+  # Need to be defined before configure_file
```
mitza-oci, line 219:
```suggestion
  # Needs to be defined before configure_file
```
? not entirely sure what this is referring to

iguessthislldo: `uppercase_target` and `export_macro` are used in the export template. I moved them after `configure_file` on my machine and was confused why my builds where breaking like they were completely missing the export headers.

mitza-oci (approved):

PR #4491, opened 2024-02-27
OpenDDS cannot be configured with environment variables

jrw972:
Problem
-------

OpenDDS cannot be configured with environment variables.  Configuring with environment variables is useful for containers and cloud deployments.

Solution
--------

Parse environment variables and add them to the ConfigStore.  Only environment variables starting with OPENDDS_ are parsed.  The remainder of the name is used as the key for the ConfigStore. Environment variables are processed before command-line arguments and the config file.

### DIFF
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -54,6 +54,10 @@
 #include "Service_Participant.inl"
 #endif /* __ACE_INLINE__ */
 
+#if !defined (ACE_WIN32)
+extern char **environ;
+#endif
+
 namespace {
 
 void set_log_file_name(const char* fname)
@@ -388,6 +392,8 @@ Service_Participant::get_domain_participant_factory(int &argc,
         }
       }
 
+      parse_env();
+
       if (parse_args(argc, argv) != 0) {
         return DDS::DomainParticipantFactory::_nil();
       }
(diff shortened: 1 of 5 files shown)


dds/DCPS/Service_Participant.cpp
```diff
@@ -54,6 +54,8 @@
 #include "Service_Participant.inl"
 #endif /* __ACE_INLINE__ */
 
+extern char **environ;
```
iguessthislldo, line 57: There's windows linking warnings on this. Alternatively, we could use [Win32's function for this](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentstrings).

dds/DCPS/Service_Participant.cpp
```diff
@@ -388,6 +390,10 @@ Service_Participant::get_domain_participant_factory(int &argc,
         }
       }
 
+      if (!parse_env()) {
+        return DDS::DomainParticipantFactory::_nil();
```
iguessthislldo, line 394: I know it doesn't fail right now, but If this fails, what does that mean? If fails because we couldn't get environment because the platform doesn't have one, then do we really have to fail to create the `DomainParticipantFactory` because of it?

iguessthislldo (changes requested):

iguessthislldo (approved):

dds/DCPS/Service_Participant.cpp
```diff
@@ -578,6 +584,56 @@ Service_Participant::get_domain_participant_factory(int &argc,
...
+      ACE_ERROR((LM_ERROR,
+                 "(%P|%t) ERROR: Service_Participant::parse_env: Could not get environment strings\n"));
+    }
+    return;
+  }
+
+  LPTSTR env_string = (LPTSTR) env_strings;
+
+  while (*env_string) {
+    parse_env(ACE_TEXT_ALWAYS_CHAR(env_string));
```
jrw972, line 606: @mitza-oci Does this make sense?

mitza-oci: Try adding a uses_wchar Windows build to CI

mitza-oci (commented):
This could be tested by setting the environment in the Perl test driver or spawning a child process with a modified environment.

dds/DCPS/Service_Participant.cpp
jrw972: For posterity...

build_and_test.yml is at the limit that GHA will allow so we will need to refactor it before we can add another build.

Converting an existing Windows build failed with build errors.  One of the root causes was wide character not being supported correctly for ACE's XML support.  So, we will need to be selective about what features to include in such a build.

I will record these points in the Wiki.

PR #4492, opened 2024-02-27
Allow compile-time configuration of CLOCK_BOOTTIME as the monotonic time source

mitza-oci:

PR #4498, opened 2024-02-29
Conditionally include OpenDDSConfig.h, fixes #4497

### LINKED ISSUE #4497: Master requires dds/OpenDDSConfig.h header
Master is broken for anyone not using configure, like our builds, see https://download.remedy.nl/buildlogs/fc_opendds_versioned/. This was caused by https://github.com/OpenDDS/OpenDDS/pull/4482, this should be fixed by a separate PR see https://github.com/OpenDDS/OpenDDS/pull/4482#discussion_r1503173420

Easiest looks to be to add back the following

```
ifdef __has_include
#  if __has_include(<dds/OpenDDSConfig.h>)
```

jwillemsen:
* dds/DCPS/Definitions.h:


iguessthislldo:
We can't change it back to that because it requires C++17 and we don't want to require that just to use #4492. You can read more about the config file design [here](https://github.com/OpenDDS/OpenDDS/wiki/Proposal:-Header-file-for-compile%E2%80%90time-configuration). We did discuss an alternative where a macro could be used to ignore the config file. Something like:

```C++
#ifndef OPENDDS_IGNORE_CONFIG_FILE
#  include <dds/OpenDDSConfig.h>
#endif
```

jwillemsen:
A macro will also work, I can add that to the ace/config.h file. for ACE7 I want to move to c++17 this year, so we could like this pr to that upgrade

iguessthislldo (changes requested):
This looks okay to us. The only change I think is needed is to add it to the [news file](https://github.com/OpenDDS/OpenDDS/blob/aee0af4556a4e92d1ac73c7fc063948a7cff2100/docs/news.d/openddsconfig.rst?plain=1#L1). Please add this PR number to the top (`.. news-prs: 4482 4497`) and this as an alternative in the manual configuration line.

docs/news.d/openddsconfig.rst
```diff
@@ -1,8 +1,10 @@
-.. news-prs: 4482
+.. news-prs: 4482 4498
 
 .. news-start-section: Additions
 - A new header, ``dds/OpenDDSConfig.h`` is generated by configure or CMake.
 
-  - Users manually configuring a build will need to create this file, which may be empty.
+  - Users manually configuring a build will need to create this file, which may be empty, or
+    add the define `OPENDDS_IGNORE_OPENDDSCONFIG_H_FILE` to their `ace/config.h` file
```
iguessthislldo, line 7:
```suggestion
    add the define ``OPENDDS_IGNORE_OPENDDSCONFIG_H_FILE`` to their ``ace/config.h`` file.
```

iguessthislldo (changes requested):

iguessthislldo (approved):

PR #4500, opened 2024-03-05
Update ValueReader interface

sonndinh:


tests/unit-tests/dds/DCPS/ValueCommon.cpp
```diff
@@ -63,13 +63,14 @@ TEST(dds_DCPS_ValueCommon, MapBitmaskHelper)
...
+  EXPECT_EQ(0x32ull, value);
 
   value = 0x65;
   EXPECT_GT(helper.get_names(names, value), (size_t)0);
   EXPECT_EQ(names.size(), (size_t)4);
-  EXPECT_STREQ("FLAG0", names[0]);
-  EXPECT_STREQ("FLAG2", names[1]);
-  EXPECT_STREQ("FLAG5", names[2]);
-  EXPECT_STREQ("FLAG6", names[3]);
+  EXPECT_STREQ("FLAG0", names[0].c_str());
```
mitza-oci, line 72: Is `.c_str()` needed?  It seems like this should use `operator==` which supports comparing `const char*` with `std::string`

sonndinh: Yeah, using `EXPECT_EQ` doesn't need the `c_str()` call.

iguessthislldo (approved):

PR #4519, opened 2024-03-13
Support the IDL4 `@value` annotation in IDL-to-C++11 and use it for RTPS SubmessageKind

mitza-oci:
Adds limited support for IDL4 "standardized" value annotation. When opendds_idl is used to generate the language mapping for an enum, each enumerator annotated with `@value(x)` will be assigned value `x` in the generated C++ code.

Moves RTPS SubmessageKind to its own IDL file to use the value annotation.

Also changes opendds_idl options used for the IDL that's internal to the RTPS library in order to minimize footprint of generated code.


dds/DCPS/RTPS/RtpsSubmessageKind.idl
```diff
@@ -0,0 +1,40 @@
...
+
+  // SubmessageKinds 0x30-3f are reserved for DDS Security
+  @value(0x30) SEC_BODY,
+  @value(0x31) SEC_PREFIX,
+  @value(0x32) SEC_POSTFIX,
+  @value(0x33) SRTPS_PREFIX,
+  @value(0x34) SRTPS_POSTFIX
+
+  // SubmessageKinds 0x80 and above are vendor-specific
+};
```
sonndinh, line 36: Since this uses the IDL-to-C++11 mappings, does it only work with compilers with C++11 support?

dds/idl/marshal_generator.cpp
```diff
@@ -153,36 +153,29 @@ namespace {
...
-    const std::string idl_name = canonical_name(name);
     be_global->impl_ <<
-      "    if (CORBA::ULong(enumval) >= " << vals.size() << ") {\n"
-      "      if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Warning) {\n"
-      "        ACE_ERROR((LM_WARNING, \"(%P|%t) WARNING: "
-        "%u is an invalid enumerated value for " << idl_name << "\\n\", enumval));\n"
-      "      }\n"
-      "      return false;\n"
-      "    }\n"
-      "  return strm << static_cast<CORBA::ULong>(enumval);\n";
```
iguessthislldo, line 175: Is it okay to take out this warning? It seems reasonable. Is the check for validity really that costly?

docs/devguide/xtypes.rst
```diff
@@ -820,6 +820,35 @@ The ``@key`` annotation marks a member used to determine the Instances of a topi
 See :ref:`getting_started--keys` for more details on the general concept of a Key.
 For XTypes specifically, two types can only be compatible if each contains the members that are keys within the other.
 
+Customizing the values of enumerators
+=====================================
+
+..
+    Sect<16.6.7>
```
iguessthislldo, line 827:
```suggestion
```

The section comments were for debugging the conversion script, but I kept in case someone wanted to search using an old section reference. I don't think we need to add numbers for new sections.

docs/news.d/enum-value-annotation.rst
```diff
@@ -0,0 +1,6 @@
+.. news-prs: 4519
+
+.. news-start-section: Additions
+- The ``@value(x)`` annotation is now supported on IDL enumerators when using the IDL-to-C++11 mapping.
```
iguessthislldo, line 4:
```suggestion
- The ``@value(x)`` annotation is now supported on IDL enumerators when using the IDL-to-C++11 mapping.

  - See :ref:`xtypes--value` for details.
```

iguessthislldo (changes requested):

dds/DCPS/RTPS/RtpsSubmessageKind.idl
iguessthislldo: It actually looks like he's compiling this file using `-Lspcpp`, which is the mapping for safety profile. Both use `langmap_generator`, so I'm guessing `@value` works for both C++11 and Safety profile mappings now and there's nothing stopping us from using the safety profile mapping like this in a normal build.

mitza-oci: Lspcpp is a minimal generator backend that's not useful for end users building their IDL for use as topic types.  It has enough of the IDL-to-C++(11) mapping to generate the code that's needed for safety profile when tao_idl isn't available.
In this case, we have IDL files in the OpenDDS_Rtps library that are just used for implementing the protocol and not as DDS topic types.

docs/devguide/xtypes.rst
mitza-oci: Of course I just copied and pasted the section above it.  I'll update this.

dds/idl/marshal_generator.cpp
mitza-oci: Since this is the output side, it's not necessary to check.  But checking may be helpful/desirable.
Currently it's just a `O(N)` walk through the list of valid options.  For some enums this might be noticeable since it's on the critical path.

sonndinh (approved):

iguessthislldo (approved):

jwillemsen:
Found yesterday some new errors in the AXCIOMA tests, I think this change introduced a small issue, see https://github.com/OpenDDS/OpenDDS/pull/4537

PR #4520, opened 2024-03-14
Improve QoS Documentation

iguessthislldo:
- Quality of Service
  - Rearranged to remove large default value tables.
  - Added Data Representation and Type Consistency Enforcement.
  - Every policy now has a box that says if it's mutable, if it affects writer-reader association, its default values, and links to the spec definition.
  - Added many links to QoS policies
- Move fnmatch expressions into an "annex" file so it can be common between security and partitions QoS.
- Add definitions for "instance", "unregister", and "dispose" to the glossary.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -152,7 +152,7 @@ intersphinx_mapping = {
     'cmake': ('https://cmake.org/cmake/help/latest', None),
 }
 
-manpages_url = 'https://linux.die.net/man/{section}/{page}'
+manpages_url = 'https://manpages.debian.org/{page}({section})'
 
 
 # -- Options for Markdown output ---------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -152,6 +152,9 @@ intersphinx_mapping = {
     'cmake': ('https://cmake.org/cmake/help/latest', None),
 }
 
+manpages_url = 'https://linux.die.net/man/{section}/{page}'
+
+
 # -- Options for Markdown output ---------------------------------------------
 # This builder is just used to generate the release notes for GitHub
 
--- a/docs/devguide/annex.rst
+++ b/docs/devguide/annex.rst
@@ -1,0 +1,37 @@
+.. _annex:
+
+#####
+Annex
+#####
+
+Documentation that doesn't fit with other sections, but isn't big enough to justify having a separate section.
+
+.. _fnmatch-exprs:
+
+*******************
+Fnmatch Expressions
+*******************
+
+A wildcard-capable string used to match one or more names from a set.
+This is used in :ref:`qos-partition` and in :ref:`sec` documents.
+Recognized values will conform to POSIX ``fnmatch()`` function as specified in POSIX 1003.2-1992, Section B.6.
+This is a subset of UNIX shell file matching and is similar to, but separate from standard regular expressions.
+
+Simplified, this consists of the following:
+
+``?``
+  Will match any single character.
+  For example ``ab?`` matches ``abc`` and ``abb``.
+
+``*``
+  Will match any zero or more characters.
+  For example ``*`` will match anything and ``a*`` matches ``a``, ``abc``, and ``aaaaa``.
+
+``[]``
+  Will match a single character specified in the brackets.
+  For example ``a[bc]`` matches ``ab`` and ``ac``.
+  Can also use ranges, for example ``a[b-d]`` matches ``ab``, ``ac``, and ``ad``.
+
+``\``
+  Will escape the following character.
+  For example ``\?`` just matches ``?`` and ``\\`` matches ``\``.
--- a/docs/devguide/annex.rst
+++ b/docs/devguide/annex.rst
@@ -4,8 +4,6 @@
 Annex
 #####
 
-Documentation that doesn't fit with other sections, but isn't big enough to justify having a separate section.
-
 .. _fnmatch-exprs:
 
 *******************
--- a/docs/devguide/quality_of_service.rst
+++ b/docs/devguide/quality_of_service.rst
@@ -557,7 +557,7 @@ Partition QoS
     Sect<3.2.8>
 
 The partition QoS policy allows the creation of logical partitions within a :term:`domain`.
-It only allows data readers and data writers to be associated if they have matched partition strings.
+Data readers and data writers can only be associated if they have at least one matched partition string.
 This policy applies to the publisher and subscriber entities via the ``partition`` member of their respective QoS structures.
 
 .. important::
--- a/docs/devguide/quality_of_service.rst
+++ b/docs/devguide/quality_of_service.rst
@@ -286,8 +286,8 @@ History QoS
     Sect<3.2.4>
 
 The history QoS policy determines how :term:`sample`\s are held in the data writer and data reader for a particular :term:`instance`.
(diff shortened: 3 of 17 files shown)


iguessthislldo:
The main change, the info boxes for the QoS, can be seen here https://opendds--4520.org.readthedocs.build/en/4520/devguide/quality_of_service.html#reliability-qos

iguessthislldo:
Links to all the changes here: https://opendds--4520.org.readthedocs.build/en/4520/news.html#documentation

docs/devguide/annex.rst
```diff
@@ -0,0 +1,37 @@
...
+
+.. _fnmatch-exprs:
+
+*******************
+Fnmatch Expressions
+*******************
+
+A wildcard-capable string used to match one or more names from a set.
+This is used in :ref:`qos-partition` and in :ref:`sec` documents.
+Recognized values will conform to POSIX ``fnmatch()`` function as specified in POSIX 1003.2-1992, Section B.6.
```
mitza-oci, line 17: This could link to POSIX spec from The Open Group

docs/conf.py
```diff
@@ -152,6 +152,9 @@ def setup(app):
     'cmake': ('https://cmake.org/cmake/help/latest', None),
 }
 
+manpages_url = 'https://linux.die.net/man/{section}/{page}'
```
mitza-oci, line 155: Is there an official source we can link to?

docs/devguide/quality_of_service.rst
```diff
@@ -120,1047 +103,881 @@ The following examples illustrate how to obtain the default policies for publish
...
+The value of the policy is available in respective :ref:`built-in topic data <bit>`.
+The value's use is defined by the application.
+For example, the application could attach security credentials via the policy that can be used by the remote application to authenticate the source. [#user_data_ex]_
 
-     - ``BEST_EFFORT_RELIABILITY_QOS``
+.. warning::
 
-       ``DURATION_INFINITE_SEC``
+  When using :ref:`dds_security`, the user data of a participant can be leaked in unsecured discovery messages.
+  Enabling ``[rtps_discovery]SecureParticipantUserData`` will only send and provide the real user data when it can be securely sent.
```
mitza-oci, line 763: This looks strange in the rendered document.  The `[rtps_discovery]SecureParticipantUserData` should look like it appears in the ini format, or just link to the other section.

mitza-oci (changes requested):

jrw972 (changes requested):

docs/devguide/quality_of_service.rst
iguessthislldo: This is what what the rendered config domain links look like. [Here is the example RST](https://github.com/OpenDDS/OpenDDS/blob/e4e9498857e3a025a8665cfc0f933e4147633a17/docs/internal/docs.rst?plain=1#L565) for the config domain. Right now the links looks the same as the input expect for sections discriminators appear like ``[transport] default_address (rtps_udp)``.

docs/conf.py
iguessthislldo: It's a common one that pops up in searches and looks like they've been around awhile, but yes, I agree a distro's site would be better. I originally tried Ubuntu, but weirdly it redirects `sched_get_priority_max` to the POSIX section 3 entry instead of the Linux-specific section 2 entry, which it's still doing:

https://manpages.ubuntu.com/manpages/en/man2/sched_get_priority_min
https://manpages.ubuntu.com/manpages/en/man2/sched_get_priority_max

Looking at Debian now though, it looks like it works fine:

https://manpages.debian.org/sched_get_priority_min(2)
https://manpages.debian.org/sched_get_priority_max(2)

docs/devguide/annex.rst
iguessthislldo: I think was going to when I originally wrote this for security, but I couldn't figure out what to link to. https://pubs.opengroup.org/onlinepubs/9695969399/toc.pdf doesn't have a section B. It mentions `fnmatch` and talks about what is basically the same pattern matching in section 2.13, but I didn't know if that was enough. I ended up just documenting the ACE implementation of `fnmatch` since that's what we are using anyway.

docs/devguide/quality_of_service.rst
mitza-oci: Maybe this is something we can address in a follow-on PR?

mitza-oci (approved):

docs/devguide/quality_of_service.rst
```diff
@@ -120,1363 +103,1542 @@ The following examples illustrate how to obtain the default policies for publish
...
+The property QoS policy contains sequences of key-value pairs for the :term:`DomainParticipant`.
 
-The ``PRESENTATION`` QoS policy controls how changes to instances by publishers are presented to data readers.
-It affects the relative ordering of these changes and the scope of this ordering.
-Additionally, this policy introduces the concept of coherent change sets.
-Here is the IDL for the Presentation QoS:
+.. important::
 
-.. code-block:: omg-idl
+  This policy is :ref:`immutable <qos-changing>` and affects association indirectly through security.
```
iguessthislldo, line 1437: I almost forgot about this, but I couldn't find a answer to if the DomainParitcipant properties QoS is actually immutable or not. I just assumed it was immutable, but it doesn't say either in the security spec or in the QoS code:

https://github.com/OpenDDS/OpenDDS/blob/33f65f988808d814f3828bf090ba0eab7c2b9c07/dds/DCPS/Qos_Helper.inl#L1129-L1135

I guess going by the code it's mutable if it's not checked.

mitza-oci: It's considered "Changeable" (using the term from the core spec) even though this isn't explicitly stated.
One way to see it is that it would be awkward for other specs to introduce additional constraints on the core's behavior.  The core spec offers no framework for extensibility of QoS by implementations or by other specs.

iguessthislldo: Okay, so if the actual QoS values are changeable I assume this doesn't mean security will start using a different set of security documents that were passed in? Even if the specs aren't ideal about this, I'd like to document behavior the user might care about it.

mitza-oci: The propertyQos is changeable since you can use it for anything (or a combination of things).  There are some cases where it would make sense to change the security properties, but I think they're not currently implemented in OpenDDS.

docs/devguide/quality_of_service.rst
```diff
@@ -120,1363 +103,1542 @@ The following examples illustrate how to obtain the default policies for publish
...
+===========
 
-       ``DESTINATIONORDER_QOS``
+..
+    Sect<3.2.4>
 
-   * - ``HISTORY``
+The history QoS policy determines how :term:`sample`\s are held in the data writer and data reader for a particular :term:`instance`.
+For data writers these values are held until the publisher retrieves them and successfully sends them to all connected subscribers.
+For data readers these values are held until :ref:`taken <getting_started--reading-multiple-samples>` by the application.
```
sonndinh, line 290:
```suggestion
For data writers these samples are held until the publisher retrieves them and successfully sends them to all connected subscribers.
For data readers these samples are held until :ref:`taken <getting_started--reading-multiple-samples>` by the application.
```

PR #4551, opened 2024-03-29
Add missing news

jrw972:


docs/news.d/config_store3.rst
```diff
@@ -1,5 +1,11 @@
...
+.. news-prs: 4454 4515 4505 4491 4488 4479 4475 4469
 
 .. news-start-section: Additions
 - Convert ``domain`` sections to key-value store.
+- The ConfigStore is available in Java.
+- OpenDDS now supports multiple config files.  DevGuide update to follow.
+- OpenDDS can now be configured with environment variables.  DevGuide update to follow.
+- Security configuration now uses ConfigStore.
+- Additional transport configuration now performed through ConfigStore.
+- Additional discovery configuration now performed through ConfigStore.
```
iguessthislldo, line 10: Putting in a bunch of news items under different PRs like this doesn't look sane in the resulting news:

https://opendds--4551.org.readthedocs.build/en/4551/news.html#additions

I would suggest:

These can probably stay in this file:

```suggestion
.. news-prs: 4454 4469 4475 4479 4488

.. news-start-section: Additions
.. news-start-section: ``ConfigStore``
- Converted the transport and discovery loading, and ``domain`` and ``security`` sections to use ``ConfigStore``.
.. news-end-section
.. news-end-section
```

A side note: I should have noticed this when I reviewed the PR, but the security configuration from #4479 doesn't looks like it's tested (outside of a unit test) or documented in the devguide. If it's not documented, does it need to be? If it doesn't need to be, then does the fact that is was converted to use ConfigStore need to be in the news?

The rest should  probably have separate files, but could also be in a ConfigStore section to group them:

```
.. news-prs: 4515

.. news-start-section: ``ConfigStore``
- The ConfigStore is available in Java.
.. news-end-section
.. news-end-section
```

```
.. news-prs: 4505

.. news-start-section: ``ConfigStore``
- OpenDDS now supports multiple config files.  DevGuide update to follow.
.. news-end-section
.. news-end-section
```

```
.. news-prs: 4491

.. news-start-section: ``ConfigStore``
- OpenDDS can now be configured with environment variables.  DevGuide update to follow.
.. news-end-section
.. news-end-section
```

iguessthislldo (changes requested):

docs/news.d/config_store3.rst
jrw972: I was similarly confused about the security section when I was working on it.  I think the right thing to do is remove it from the news and think about the feature itself.  Users are definitely not using this feature so I'm wondering why we have it or maybe it comes from the spec.  If it becomes useful to the user, I think that's the right time to document it.

iguessthislldo (approved):

PR #4554, opened 2024-04-02
Generate `vread` and `vwrite` for key-only samples

sonndinh:
- Add gereration of `vread` and `vwrite` functions for KeyOnly and NestedKeyOnly
- KeyOnly serialization of union that doesn't have an explicit key now has a delimiter
- Remove the generation of serialization functions for KeyOnly wrapper of non-topic union types


dds/idl/dds_generator.h
```diff
@@ -1443,6 +1443,50 @@ inline std::string type_kind(AST_Type* type)
   }
 }
 
+inline
+std::string key_only_type_name(const std::string& type_name, FieldFilter field_filter, bool writing)
```
iguessthislldo, line 1447: This looks like it's duplicating logic in the `RefWrapper`, expect it doesn't look like the `RefWrapper` has `KeyOnly`. I think either it should be using this or the other way around.

jrw972 (approved):

dds/DCPS/ValueDispatcher.h
```diff
@@ -21,8 +21,8 @@ struct OpenDDS_Dcps_Export ValueDispatcher {
   virtual void* new_value() const = 0;
   virtual void delete_value(void* data) const = 0;
 
-  virtual bool read(ValueReader& value_reader, void* data) const = 0;
-  virtual bool write(ValueWriter& value_writer, const void* data) const = 0;
+  virtual bool read(ValueReader& value_reader, void* data, bool key_only = false) const = 0;
+  virtual bool write(ValueWriter& value_writer, const void* data, bool key_only = false) const = 0;
```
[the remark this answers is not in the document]
sonndinh: It looks like these functions are called on topic types, so it only needs to differentiate between full sample vs. key-only sample, and a bool would be sufficient.

dds/idl/dds_generator.h
sonndinh: Are you suggesting to add `KeyOnly` to `RefWrapper`?

dds/DCPS/ValueDispatcher.h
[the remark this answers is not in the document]
mitza-oci: A `bool` is certainly sufficient.  Bool parameters are often avoided in API design because the functional call expression lacks context about what the `true` and `false` constants mean.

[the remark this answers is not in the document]
sonndinh: Got it. Not sure if we have an enum that only has `Full` and `KeyOnly`. If not, I think using `Sample::Extent` is okay; it would just not be called with `Sample::Extent::NestedKeyOnly`.

tests/DCPS/Compiler/xcdr/xcdr.cpp
```diff
@@ -2533,9 +2548,7 @@ void build_expected_union(DataVec& expected, FieldFilter field_filter, bool keye
   if (include_unkeyed) {
     non_keys = DataView(key_only_union_non_keys_expected_base);
   }
-  if (include_possible_keyed) {
-    serialize_u32(expected, keys.size + non_keys.size);
-  }
+  serialize_u32(expected, keys.size + non_keys.size);
```
iguessthislldo, line 2551: Is the serialization of unions changing?

iguessthislldo: Oh, this is the keyed delimiter... If we're doing this, at the very least we have to announce in the news that their serialization might break.

sonndinh: Yes, it's for the change to KeyOnly serialization of union. I'll add a news item.

docs/news.d/keyonly_vread_vwrite.rst
```diff
@@ -0,0 +1,5 @@
+.. news-prs: 4554
+
+.. news-start-section: Fixes
+- KeyOnly serialization of union that has no key now has a delimiter for appendable and mutable extensibility.
```
mitza-oci, line 4: Should this specify XCDR2?

sonndinh: Yes, I'll update that.

iguessthislldo (approved):

PR #4556, opened 2024-04-03
Configuration with ConfigStore is not documented

jrw972:
Problem
-------

The new capabilities from the ConfigStore are not documented for users.

Solution
--------

Add to the configuration documentation.

### DIFF
--- a/docs/devguide/run_time_configuration.rst
+++ b/docs/devguide/run_time_configuration.rst
@@ -12,17 +12,6 @@ Run-time Configuration
 
 .. _run_time_configuration--configuration-approach:
 
-**********************
-Configuration Approach
-**********************
-
-..
-    Sect<7.1>
-
-OpenDDS includes a file-based configuration framework for configuring global options and options related to specific publishers and subscribers such as discovery and transport configuration.
-OpenDDS also allows configuration via the command line for a limited number of options and via a configuration API.
-This section summarizes the configuration options supported by OpenDDS.
-
 OpenDDS configuration is concerned with three main areas:
 
 #. **Common Configuration Options** -- configure the behavior of DCPS entities at a global level.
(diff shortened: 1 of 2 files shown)


docs/devguide/run_time_configuration.rst
```diff
@@ -36,17 +25,49 @@ OpenDDS configuration is concerned with three main areas:
...
+Values are stored as strings and converted to other types as necessary.
+
+A key has the following parts:
+
+#. **section** -- The section describes the area of functionality that is being configured.
+
+#. **instance** -- (optional) The instance names a particular collection of configuration values.
+   Configuration keys that do not have an instance imply a singleton or global.
+
+#. **property** -- The property names a variable relative to the section and instance.
```
iguessthislldo, line 47: I think we said we'd propbably be fine leaving the key vs property terms alone for now, but the fact that a ConfigStore key is explicitly defined to contain a property seems like it could be confusing. Should I change it so it's always referred to as a *property* so *key* is a sort of protected term in configuration that's specific to the ConfigStore?

docs/devguide/run_time_configuration.rst
```diff
@@ -36,17 +25,49 @@ OpenDDS configuration is concerned with three main areas:
...
+**********************
+Configuration Approach
+**********************
+
+..
+    Sect<7.1>
+
+Most of OpenDDS is configured through a key-value store.
+Keys are strings that are converted to a canonical form before being stored by 1) converting camel case to snake case, 2) capitalizing all alphabetic characers, 3) replacing all non-alphanumeric characters with underscores, and 4) stripping leading, trailing, and duplicate underscores.
+For example, the key ``!SectionInstance__PROPERTY`` is canonicalized into ``SECTION_INSTANCE_PROPERTY`` before being stored.
```
iguessthislldo, line 37: It might be helpful to note that the canonical names of each key are in their documentation.

docs/news.d/config_store_documentation.rst
```diff
@@ -0,0 +1,6 @@
+.. news-prs: 4556
+.. news-start-section: Documentation
+.. news-start-section: ``ConfigStore``
+- Add configuration capabilities to DevGuide.
```
iguessthislldo, line 4:
```suggestion
- Add :ref:`configuration capabilities <run_time_configuration--configuration-approach>` to DevGuide.
```

iguessthislldo (changes requested):

docs/devguide/run_time_configuration.rst
jrw972: I think that's the right way to go.  I'm happy to change whatever I can in this PR if you describe what needs to be done.  Is is just looking for "key" and changing to property as necessary?

iguessthislldo: I think I can change it in my next set of changes since I don't think these changes don't have any instances of that.

iguessthislldo: #4564

PR #4557, opened 2024-04-03
Update vcpkg to fix a failed download

mitza-oci:


jrw972 (approved):

iguessthislldo:
It looks like it's still trying to download from https://github.com/tukaani-project/xz

iguessthislldo (changes requested):
In https://github.com/microsoft/vcpkg/issues/37839 I think they're saying they reverted the versions, but the xz repo's still gone and [they're not going to change the repo](https://github.com/microsoft/vcpkg/pull/37957#pullrequestreview-1977640176). So I think our options are either to patch vcpkg like it seems people are doing or somehow get it to not need xz if that's even possible.

mitza-oci:
> It looks like it's still trying to download from https://github.com/tukaani-project/xz

Right, ce1916404fc6f2b645f419a6d47b7ebafe686582 changed the version back to 5.4.4 but that's apparently insufficient.

We'll need to wait for https://github.com/microsoft/vcpkg/pull/37957

mitza-oci:
Or something in the dependency chain has to change... (the identifiers in brackets are enabled features for each package)
```
qt5-tools
qt5-declarative[platform-default-features, d3d12]
qt5-imageformats[tiff, webp]
tiff[jpeg, lzma, zip]
liblzma
```

We install "qt5-winextras", "qt5-tools", "qt5-svg", "qt5-multimedia", "qt5-declarative" to satisfy Wireshark's build-time dependencies.

mitza-oci:
Wireshark's own CI appears to use https://github.com/jurplel/install-qt-action

Issue #4550, opened 2024-03-29
Update compiler warning levels used for CI


mitza-oci:
Downstream projects may use higher warning levels than the compiler defaults.  An example is https://github.com/OpenDDS/opendds-monitor/blob/39c38579479f88411f287956ef2b66c2048a9524/test/CMakeLists.txt#L6  We would like to catch these problems before the PR is merged.

Update the main OpenDDS CI workflows so at least some builds use higher warning levels and treat warnings as errors.

PR #4558, opened 2024-04-03
Update compiler warning levels

jrw972:
See #4550.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -90,6 +90,7 @@ jobs:
           -DBUILD_SHARED_LIBS=TRUE \
           -DCMAKE_BUILD_TYPE=Debug \
           -DOPENDDS_BOOTTIME_TIMERS=TRUE \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
           -DOPENDDS_SECURITY=TRUE \
           -DOPENDDS_ACE_TAO_SRC=$GITHUB_WORKSPACE/OpenDDS/build/ACE_TAO \
           -DOPENDDS_MPC=$GITHUB_WORKSPACE/MPC \
@@ -132,6 +133,7 @@ jobs:
           -DCMAKE_UNITY_BUILD=TRUE \
           -DBUILD_SHARED_LIBS=FALSE \
           -DCMAKE_BUILD_TYPE=Debug \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
           -DOPENDDS_ACE_TAO_SRC=$GITHUB_WORKSPACE/OpenDDS/build/ACE_TAO \
           -DOPENDDS_MPC=$GITHUB_WORKSPACE/MPC \
           -DOPENDDS_BUILD_TESTS=TRUE \
@@ -176,6 +178,7 @@ jobs:
           -DCMAKE_UNITY_BUILD=TRUE \
           -DBUILD_SHARED_LIBS=TRUE \
           -DCMAKE_BUILD_TYPE=Debug \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
           -DOPENDDS_ACE_TAO_SRC=$GITHUB_WORKSPACE/OpenDDS/build/ACE_TAO \
           -DOPENDDS_MPC=$GITHUB_WORKSPACE/MPC \
           -DOPENDDS_BUILD_TESTS=TRUE \
@@ -228,6 +231,7 @@ jobs:
         cmake -B build-host \
           -G Ninja \
           -DBUILD_SHARED_LIBS=FALSE \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
           -DOPENDDS_JUST_BUILD_HOST_TOOLS=TRUE
     - uses: ammaraskar/gcc-problem-matcher@0.3.0
     - name: Build Host Tools
@@ -242,6 +246,7 @@ jobs:
           -DBUILD_SHARED_LIBS=TRUE \
           -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-29 \
           --toolchain ${{ steps.setup-ndk.outputs.ndk-path }}/build/cmake/android.toolchain.cmake \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
           -DOPENDDS_CMAKE_VERBOSE=all \
           -DOPENDDS_HOST_TOOLS=$(realpath build-host)
     - name: Build Target
@@ -287,6 +292,7 @@ jobs:
         cd OpenDDS
         cmake -B build^
           -DBUILD_SHARED_LIBS=TRUE^
+          -DOPENDDS_COMPILE_WARNINGS=WARNING^
           -DOPENDDS_ACE_TAO_SRC=${{ github.workspace }}\OpenDDS\build\ACE_TAO^
           -DOPENDDS_MPC=${{ github.workspace }}\MPC^
           -DOPENDDS_XERCES3=${{ env.VCPKG_INSTALLED_DIR }}/x64-windows^
@@ -343,6 +349,7 @@ jobs:
         cmake -B build^
           -DCMAKE_UNITY_BUILD=TRUE^
           -DBUILD_SHARED_LIBS=FALSE^
+          -DOPENDDS_COMPILE_WARNINGS=WARNING^
           -DOPENDDS_ACE_TAO_SRC=${{ github.workspace }}\OpenDDS\build\ACE_TAO^
           -DOPENDDS_MPC=${{ github.workspace }}\MPC^
           -DOPENDDS_XERCES3=${{ env.VCPKG_INSTALLED_DIR }}/x64-windows^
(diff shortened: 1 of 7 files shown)


cmake/opendds_build_helpers.cmake
```diff
@@ -60,7 +60,9 @@ function(_opendds_library target)
   endif()
 
   if(MSVC AND arg_MSVC_BIGOBJ)
-    target_compile_options(${target} PRIVATE /bigobj)
+    target_compile_options(${target} PRIVATE /bigobj /W4)
+  else()
+    target_compile_options(${target} PRIVATE $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>,$<CXX_COMPILER_ID:GNU>>: -Wall -Wpedantic -Wno-unused -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-include-dirs -Woverloaded-virtual -Wredundant-decls -Wshadow -Wsign-conversion -Wstrict-overflow=5 -Wundef -Werror> $<$<CXX_COMPILER_ID:MSVC>: /W4>)
```
iguessthislldo, line 65: A couple of thoughts about this:

1. This should be separate from bigobj because they're not related at all besides having to do with compiler options.
2. It should be broken into into separate variables because generator expressions this long are a pain to look at. I'm also not sure it needs to be a generator expression because compilers are fixed at configuration, so it could just be normal CMake code, which is (relatively) easier to read and maintain.
3. `-Werror` shouldn't be unconditional or default, it should be something we have to opt into for the CI builds. If a user in the future wants to build with a newer compiler or different settings that causes new warnings, then the build fails when it doesn't have to. GoogleTest had [this issue](https://github.com/google/googletest/issues/3447) and I think the version we use still has it.
4. This is only in `_opendds_library` and not `_opendds_executable`. They should be in common for both, probably in `_opendds_alias`.

So all together, something like:

```cmake
function(_opendds_target target) # Renamed _opendds_alias
  # This is the name the target should be exported from CMake-build OpenDDS as
  # or imported in the MPC-built OpenDDS CMake package. For consistency and
  # convenience, alias it so we can use it within the OpenDDS CMake build.
  string(REPLACE "OpenDDS_" "" name "${target}")
  set(name "OpenDDS::${name}")
  get_target_property(target_type ${target} TYPE)
  if(target_type MATCHES "LIBRARY")
    add_library("${name}" ALIAS "${target}")
  else()
    add_executable("${name}" ALIAS "${target}")
  endif()
  set_target_properties(${target} PROPERTIES EXPORT_NAME "${name}")
  
  if(OPENDDS_COMPILE_WARNINGS)
    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
      target_compile_options(${target} PRIVATE
        -Wall -Wpedantic -Wno-unused -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy
        -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-include-dirs -Woverloaded-virtual
        -Wredundant-decls -Wshadow -Wsign-conversion -Wstrict-overflow=5 -Wundef
      )
      if(OPENDDS_COMPILE_WARNINGS STREQUAL "ERROR")
        target_compile_options(${target} PRIVATE -Werror)
      endif()
    elseif(MSVC)
      target_compile_options(${target} PRIVATE /W4)
    endif()
  endif()
endfunction()
```

Then `OPENDDS_COMPILE_WARNINGS` could be default `TRUE` and you can turn off the warnings using `-DOPENDDS_COMPILE_WARNINGS=FALSE` or turn them into errors using `-DOPENDDS_COMPILE_WARNINGS=ERROR`.

`_opendds_alias` calls have to be renamed and `OPENDDS_COMPILE_WARNINGS` option has to be set in `init.cmake` and documented. We also have to set `OPENDDS_COMPILE_WARNINGS` in the workflows. Do you want me to do all that?

iguessthislldo (changes requested):

mitza-oci (changes requested):
* add new CMake option to documentation
* update MPC build(s)

docs/devguide/building/index.rst
```diff
@@ -672,6 +672,13 @@ These are all the variables that are exclusive to building OpenDDS with CMake:
   See :ref:`cmake-running-tests` for how to run them.
   The default for this is ``TRUE``.
 
+.. cmake:var:: OPENDDS_COMPILE_WARNINGS
```
iguessthislldo, line 675: Can you also add this to the news?

iguessthislldo (changes requested):

iguessthislldo:
> update MPC build(s)

How do we want to do this? In the configure script?

cmake/opendds_build_helpers.cmake
```diff
@@ -18,6 +18,21 @@ function(_opendds_alias target)
...
+  if(OPENDDS_COMPILE_WARNINGS)
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      target_compile_options(${target} PRIVATE
+        -Wall -Wpedantic -Wno-unused -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy
+        -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-include-dirs -Woverloaded-virtual
+        -Wredundant-decls -Wshadow -Wsign-conversion -Wstrict-overflow=5 -Wundef
+      )
+      if(OPENDDS_COMPILE_WARNINGS STREQUAL "ERROR")
+        target_compile_options(${target} PRIVATE -Werror)
+      endif()
```
mitza-oci, line 31: This is missing on the MSVC branch
See https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170

.github/workflows/cmake.yml
```diff
@@ -89,6 +89,7 @@ jobs:
           -DCMAKE_UNITY_BUILD=TRUE \
           -DBUILD_SHARED_LIBS=TRUE \
           -DCMAKE_BUILD_TYPE=Debug \
+          -DOPENDDS_COMPILE_WARNINGS=WARNING \
```
mitza-oci, line 92: Would `=ON` be more clear and concise?

docs/devguide/building/index.rst
```diff
@@ -672,6 +672,13 @@ These are all the variables that are exclusive to building OpenDDS with CMake:
...
+
+   If not empty, enables compiler warnings when compiling OpenDDS.
```
mitza-oci, line 677: "enables more compiler warnings" ?  Aren't some enabled by default either in CMake, its specific Generators, or the compilers themselves?

mitza-oci (changes requested):

mitza-oci:
> > update MPC build(s)
> 
> How do we want to do this? In the configure script?

I think that's a good option

cmake/opendds_build_helpers.cmake
iguessthislldo: Are you talking about `/WX`?

iguessthislldo (approved):

cmake/opendds_build_helpers.cmake
```diff
@@ -18,6 +18,24 @@ function(_opendds_alias target)
     add_executable("${name}" ALIAS "${target}")
   endif()
   set_target_properties(${target} PROPERTIES EXPORT_NAME "${name}")
+
+  if(OPENDDS_COMPILE_WARNINGS STREQUAL "WARNING" OR OPENDDS_COMPILE_WARNINGS STREQUAL "ERROR")
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+      target_compile_options(${target} PRIVATE
+        -Wall -Wpedantic -Wno-unused -Wextra -Wcast-align -Wcast-qual -Wctor-dtor-privacy
+        -Wdisabled-optimization -Wformat=2 -Winit-self -Wmissing-include-dirs -Woverloaded-virtual
+        -Wredundant-decls -Wshadow -Wsign-conversion -Wstrict-overflow=5 -Wundef
```
mitza-oci, line 27: Instead of duplicating these, what about putting them in a separate file which is read by both CMake and configure?

iguessthislldo: Like acetao.ini does now? I don't know if we want to put them there or not though.

cmake/opendds_build_helpers.cmake
```diff
@@ -18,6 +31,13 @@ function(_opendds_alias target)
     add_executable("${name}" ALIAS "${target}")
   endif()
   set_target_properties(${target} PROPERTIES EXPORT_NAME "${name}")
+
+  if(OPENDDS_COMPILE_WARNINGS STREQUAL "WARNING" OR OPENDDS_COMPILE_WARNINGS STREQUAL "ERROR")
+    target_compile_options(${target} PRIVATE ${CMAKE_CXX_COMPILER_ID}-warning)
```
jrw972, line 36: @iguessthislldo What is the necessary syntax here?

iguessthislldo: I think what you want is:

```suggestion
    target_compile_options(${target} PRIVATE ${${CMAKE_CXX_COMPILER_ID}-warning)}
```

iguessthislldo: I'm assuming a space delimited list will be interpreted the same as a normal CMake `;`-delimited list though.

mitza-oci (approved):

PR #4568, opened 2024-04-09
CLOCK_BOOTTIME timers (Linux only)

mitza-oci:
Allow configuration of timers that use the CLOCK_BOOTTIME clock offered by Linux.


jrw972 (approved):

jrw972 (approved):

configure
```diff
@@ -227,6 +227,7 @@ my @specs = # Array of array-refs, each inner array is an option group which
     'macros=s@', 'Extra text for platform_macros.GNU',
     'features=s@', 'Extra text for default.features',
     'mpcopts=s@', 'Extra command-line args for MPC',
+    'boottime|boottime!', 'Use CLOCK_BOOTTIME for timers (no)',
```
iguessthislldo, line 230: Does this need to be `boottime|boottime`?

dds/DCPS/PeriodicTask.cpp
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+void PeriodicTask::enable_i(bool reenable, const TimeDuration& per)
+{
+  if (!enabled_) {
+    timer_ = Timers::schedule(reactor(), *this, 0, TimeDuration(), per);
+    if (timer_ == Timers::InvalidTimerId) {
+      ACE_ERROR((LM_ERROR, "(%P|%t) PeriodicTask::enable_i"
+                  " failed to schedule timer %p\n", ACE_TEXT("")));
```
iguessthislldo, line 47: Since we're chancing these, we should convert these to use `LogLevel::Error`, there are a few more in `Multitask.cpp`.

iguessthislldo (changes requested):

configure
iguessthislldo: As opposed to just `boottime!`.

iguessthislldo (approved):

dds/DCPS/Timers.cpp
```diff
@@ -0,0 +1,121 @@
+/*
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#include "DCPS/DdsDcps_pch.h" //Only the _pch include should start with DCPS/
+
+#include "Timers.h"
+
```
sonndinh, line 9:
```suggestion
```

PR #4572, opened 2024-04-10
Pass Xerces Path to ACE/TAO in CMake

iguessthislldo:


cmake/configure_ace_tao.pl
```diff
@@ -113,6 +115,13 @@ sub read_file {
 $ENV{MPC_ROOT} = File::Spec->rel2abs($values{mpc});
 $ENV{ACE_ROOT} = File::Spec->rel2abs($values{ace});
 $ENV{TAO_ROOT} = File::Spec->rel2abs($values{tao});
+if ($values{'env'}) {
+  for my $name (keys(%{$values{'env'}})) {
```
mitza-oci, line 119: Is this used?  I don't mind having it even if it's not... just trying to follow the logic of the changes

iguessthislldo: It was used for https://github.com/OpenDDS/OpenDDS/pull/4572/commits/8b529a326f5891732193e641cda0567e6eb9d38d. It isn't used since I removed setting `XERCESC_LIBDIR`, but it could be used for something else.

mitza-oci (approved):

jrw972 (approved):

PR #4574, opened 2024-04-11
`--mpcopts` values are split on spaces

jrw972:
Problem
-------

`--mpcopts` is a `configure` script option that accumulates additional options for MPC.  These values are split on spaces so `--mpcopts="-features no_opendds_testing_features=0"` results in two arguments to MPC: `-features` and `no_opendds_testing_features=0`.

This behavior complicates situations where one of the arguments to MPC contains spaces.  For example, `--mpcopts="-value_template build_flags+=-Wall -Werror" results in three arguments to MPC (and an error): `-value_template`, `build_flags+=Wall`, and `-Werror`.

Solution
--------

Remove the splitting functionality.

The examples above would now be written as
* `--mpcopts=-features --mpcopts=no_opendds_testing_features=0`
* `--mpcopts=-value_template --mpcopts=build_flags+="-Wall -Werror"`

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1870,7 +1870,7 @@ jobs:
           --tests \
           --security \
           --no-rapidjson \
-          --mpcopts="-features no_opendds_testing_features=0"
+          --features no_opendds_testing_features=0
     - name: check build configuration
       shell: bash
       run: |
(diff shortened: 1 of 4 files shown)


mitza-oci:
> We could support something like this to make usage a little more convenient
> 
> ```
>     --mpc:features no_opendds_testing_features=0
>     --mpc:value_template build_flags+="-Wall -Werror"
> ```

Since the script already has `--features` (which is for the default.features file), we wouldn't do exactly what's quoted above -- but the idea may be useful for other options.  Or we could just do `--mpc-value-template var+=val` ?

build_and_test.yml has a few examples of `--mpcopts=-features ...` which should probably just be `--features` unless they are for special cases that the configure script doesn't implement.

jrw972:
This PR still has a number of issues

- [ ] Are the MPC option processing doing what we want?
- [ ] How to document `--mpx:X Y` for the configure script?
- [ ] Is the news correct?
- [ ] Do we need to update the devguide?
- [ ] Do we need to update the github actions?

mitza-oci (approved):

PR #4603, opened 2024-04-25
rtps_udp sends heartbeats when sending fragments

jrw972:
Problem
-------

Suppose a writer has an agressive heartbeat period and then writes a sample that is highly fragmented.  It is possible (probable) that the heartbeat timer will go off before all fragments have been sent.  The reader can then nack fragments that haven't even been sent.  The writer would then resend the requested fragments.

Solution
--------

Before sending a heartbeat, check that the user is not in the middle of a send.

The solution does not make a change for resends due to reliability and/or durability.  The reason for this is that sending heartbeats and nack responses are mutually exclusive based on locking.

### DIFF
--- a/dds/DCPS/transport/framework/TransportSendStrategy.h
+++ b/dds/DCPS/transport/framework/TransportSendStrategy.h
@@ -326,7 +326,14 @@ public:
 
   /// Access the current sending mode.
   SendMode mode() const;
-protected:
+
+  bool is_sending(const GUID_t& guid) const
+  {
+    GuardType g(is_sending_lock_);
+    return is_sending_ == guid;
+  }
+
+ protected:
   /// Implement framework chain visitations to remove a sample.
   virtual RemoveResult do_remove_sample(const GUID_t& pub_id,
     const TransportQueueElement::MatchCriteria& criteria, bool remove_all = false);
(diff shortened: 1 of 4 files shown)


jrw972:
@jwillemsen It appears https://github.com/DOCGroup/ACE_TAO/pull/2224 broke our CMake builds.

dds/DCPS/transport/framework/TransportSendStrategy.h
```diff
@@ -438,6 +445,28 @@ class OpenDDS_Dcps_Export TransportSendStrategy
...
+    }
+
+    ~BeginEndSend()
+    {
+      GuardType g(tss_.is_sending_lock_);
+      tss_.is_sending_ = GUID_UNKNOWN;
+    }
+
+    TransportSendStrategy& tss_;
+    const GUID_t guid_;
```
mitza-oci, line 465:
```suggestion
```

Is this needed?

jrw972: No.  Good catch.

jwillemsen:
Created https://github.com/OpenDDS/OpenDDS/pull/4605 to raise C++ standard to C++17 with ACE 7

mitza-oci (approved):

iguessthislldo (approved):

PR #4605, opened 2024-04-26
ACE 7 requires C++17

jwillemsen:
* cmake/init.cmake:


iguessthislldo:
I don't want to unconditionally require 2017 by default and we're still trying to evaluate what we want to do in both OpenDDS and ACE/TAO.

jwillemsen:
When you want to keep ACE7/TAO3 at C++14 I am ok with releasing ACE8/TAO4 as next release with C++17 as minimum required level.

jrw972:
@jwillemsen We believe it would simplify things for users if we maintain the same requirements of the C++ language standard for all of the ACE7/TAO3 releases.  Changing to C++17 in ACE8/TAO4 would be perfectly acceptable.

jwillemsen:
Ok, will update ACE/TAO master that the next release will be a major one

jwillemsen:
https://github.com/DOCGroup/ACE_TAO/pull/2230

PR #4606, opened 2024-04-26
ValueDispatcher assumes pointer is KeyOnly

jrw972:
Problem
-------

The ValueDispatcher code assumes the value indicated by the void* is a KeyOnly<T>*.  This is incorrect as the void* is a T*.

Solution
--------

Assume the void* is a T* and construct the KeyOnly.


sonndinh (approved):
Thanks for fixing this.

PR #4609, opened 2024-04-30
Samples from a DynamicDataReader are immutable

jrw972:
Problem
-------

Users cannot read a sample from a DynamicDataReader, modify it, and then write it.

Solution
--------

Update the DynamicSample implementation to use DynamicDataImpl with a DynamicDataXcdrReadImpl backing store.


dds/DCPS/XTypes/DynamicSample.cpp
```diff
@@ -93,7 +93,7 @@ bool DynamicSample::deserialize(Serializer& ser)
   mb->wr_ptr(len);
 
   const DDS::DynamicType_var type = data_->type();
-  data_ = new DynamicDataXcdrReadImpl(mb.get(), ser.encoding(), type, extent_);
+  data_ = new DynamicDataImpl(type, new DynamicDataXcdrReadImpl(mb.get(), ser.encoding(), type, extent_));
```
mitza-oci, line 96: Construct each object in its own statement with its own smart pointer

dds/DCPS/XTypes/DynamicDataBase.cpp
```diff
@@ -346,9 +346,93 @@ DDS::ReturnCode_t DynamicDataBase::unsupported_method(const char* method_name, b
...
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_char8_value);
+  case TK_CHAR16:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_char16_value);
+  case TK_ENUM:
+  case TK_STRING8:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_string_value);
+  case TK_STRING16:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_wstring_value);
+  default:
+    return unsupported_method("DynamicDataBase::get_simple_value");
```
iguessthislldo, line 434: This should probably use the same logging as the `DynamicDataImpl` one. It will make it more obvious what went wrong if this case is ever hit.

dds/DCPS/XTypes/DynamicDataBase.cpp
```diff
@@ -346,9 +346,93 @@ DDS::ReturnCode_t DynamicDataBase::unsupported_method(const char* method_name, b
...
+  case TK_INT64:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_int64_value);
+  case TK_UINT8:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_uint8_value);
+  case TK_UINT16:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_uint16_value);
+  case TK_UINT32:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_uint32_value);
+  case TK_UINT64:
+    return get_some_value(value, id, *this, &DDS::DynamicData::get_uint64_value);
```
iguessthislldo, line 417: I might be wrong about this, but because `Value` just takes either a int64 or uint64, I think you could combine the signed and unsigned cases into two cases and the casting in `get_int64_value` and `get_uint64_value` will take care of the rest.

```suggestion
  case TK_INT8:
  case TK_INT16:
  case TK_INT32:
  case TK_INT64:
    return get_some_value(value, id, *this, &DDS::DynamicData::get_int64_value);
  case TK_UINT8:
  case TK_UINT16:
  case TK_UINT32:
  case TK_UINT64:
    return get_some_value(value, id, *this, &DDS::DynamicData::get_uint64_value);
```

dds/DCPS/XTypes/DynamicDataBase.cpp
```diff
@@ -346,9 +346,93 @@ DDS::ReturnCode_t DynamicDataBase::unsupported_method(const char* method_name, b
...
-  return unsupported_method("DynamicDataBase::get_simple_value");
+  DDS::DynamicTypeMember_var dtm;
+  if (type_->get_member(dtm, id) != DDS::RETCODE_OK) {
+    return DDS::RETCODE_ERROR;
+  }
+  DDS::MemberDescriptor_var md;
+  if (dtm->get_descriptor(md) != DDS::RETCODE_OK) {
+    return DDS::RETCODE_ERROR;
+  }
+  DDS::DynamicType_var member_type = get_base_type(md->type());
```
iguessthislldo, line 395:
```suggestion
  DDS::DynamicType_var member_type;
  DDS::ReturnCode_t rc = get_member_type(member_type, type_, id);
  if (rc != DDS::RETCODE_OK) {
    return rc;
  }
```

iguessthislldo (changes requested):

dds/DCPS/XTypes/DynamicDataBase.cpp
mitza-oci: I don't think this is correct: "Value just takes either a int64 or uint64"

iguessthislldo: I meant as far as integer values as concerned.

iguessthislldo (approved):

dds/DCPS/XTypes/DynamicDataBase.cpp
mitza-oci: Integer types of different sizes are supported - see constructors of `DCPS::Value`

iguessthislldo (approved):

PR #4611, opened 2024-04-30
Improvements to configure Script

iguessthislldo:
- Fix uninitialized value warning
- Tweak configure script output:
  - Bring `--help` mostly back to 80-ish columns.
  - Try to make new options a little more clear.
  - Use "cross-compile" instead of "cross compile", "GoogleTest" instead of "Google Test", "CMake" instead of "cmake", etc.
- `--mpc:OPT` should accept `--mpc:*=OPT=VALUE`.
- Try to make how features are handled a little bit more sane by putting them all in `@features`. It's still not great and I know I'm at least partially to blame.
- Try to put all features not in `default.features` in `opendds_mwc.pl`.
- Add `OpenDDSCOnfig.h.bak*` to git ignore.

### DIFF
--- a/configure
+++ b/configure
@@ -105,7 +105,7 @@ my %platforminfo =
                     'no_host' => 1,
                     'aceplatform' => 'lynxos',
                     'compiler_root_env' => 'ENV_PREFIX',
-                    'usage' => ['Set up the cross-compile using '.
+                    'usage' => ['Set up the cross compile using '.
                                 'the script from LynxOS',
                                ],
                    },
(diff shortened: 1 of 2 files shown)


jrw972 (approved):

configure
```diff
@@ -1191,16 +1209,20 @@ sub write_config_h {
...
+sub default_features {
   my %buildEnv = %{shift()};
   my @feat;
   if ($buildEnv{'build'} eq 'target') {
     push(@feat, 'ipv6=1') if $opts{'ipv6'};
-    push(@feat, @{$opts{'features'}}) if $opts{'features'};
-  }
-  elsif ($opts{'java'}) {
-    push(@feat, 'java=1');
+    my @normalized = map {/=/ ? $_: "$_=1"} @features;
```
mitza-oci, line 1217:
```suggestion
    my @normalized = map {/=/ ? $_ : "$_=1"} @features;
```

configure
```diff
@@ -436,6 +448,11 @@ if (exists($opts{'std'})) {
   # Accept any of --std=17, --std=stdcpp17, --std=c++17, etc.
   $cxx_std =~ s/^(std)?(cpp|c\+\+)//;
 }
+my @mpcopts;
+push(@mpcopts, @{$opts{'mpcopts'}}) if exists($opts{'mpcopts'});
+my @features;
+push(@features, @{$opts{'features'}}) if exists($opts{'features'});
+push(@features, 'java=1') if $opts{'java'};
```
mitza-oci, line 455: Can this be set where other Java-related settings are handled?  If it has to be here, maybe add a comment.

iguessthislldo: I put it here to be safe, but looking over how `features` is used, I think it can go with the `java_pre_jpms` code.

PR #4614, opened 2024-05-01
Internal documentation lacks dependency management policy

jrw972:
Problem
-------

The internal (developer) documentation lacks the policy used for managing the dependencies of OpenDDS.

Solution
--------

Add it.

### DIFF
--- a/docs/internal/dev_guidelines.rst
+++ b/docs/internal/dev_guidelines.rst
@@ -53,6 +53,41 @@ Dependencies
 
   :doc:`/devguide/building/dependencies` for all dependencies and details on how these are used in OpenDDS.
 
+*********************
+Dependency Management
+*********************
+
+ACE/TAO
+=======
+
+By default, the :ghfile:`configure` script and :ref:`CMake support<cmake-building>` use released versions of ACE/TAO.
+The same versions are used for pull requests.
+The versions are documented in :ghfile:`acetao.ini`.
+Currently, the :ghfile:`configure` scripts uses a release of ACE6 and CMake uses a release of ACE7.
+
+The versions are updated by the GitHub Actions workflow in :ghfile:`.github/workflows/update-ace-tao.yml` for micro/patch releases.
+When a new version of ACE/TAO is available, the workflow creates a pull request based on the new version.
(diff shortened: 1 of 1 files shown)


iguessthislldo:
This makes it sound like we're not using any branch of ACE/TAO in PR CI at all. With that, I'm concerned about having newly merged features in ACE/TAO and able to test that those are working correctly with an OpenDDS PR CI that's trying to use them.

jrw972:
> This makes it sound like we're not using any branch of ACE/TAO in PR CI at all. With that, I'm concerned about having newly merged features in ACE/TAO and able to test that those are working correctly with an OpenDDS PR CI that's trying to use them.

If a PR requires an unreleased feature of ACE/TAO, then the PR author is free to change the ACE/TAO versions from release to the branches for the purpose of CI.  However, I would push back against merging such a PR until the feature makes it into a released version of ACE/TAO at which point the ACE/TAO versions could be updated to a release version.

iguessthislldo:
I would suggest something like an environment variable in the workflows to make it easier to switch the PRs to use git branches and maybe a lint check to make sure we don't accidentally merge that.

docs/internal/dev_guidelines.rst
```diff
@@ -53,6 +53,36 @@ Dependencies
...
+ACE/TAO
+=======
+
+By default, the :ghfile:`configure` script and :ref:`CMake support<cmake-building>` use released versions of ACE/TAO.
+The same versions are used for pull requests.
+The versions are documented in :ghfile:`acetao.ini`.
+Currently, the :ghfile:`configure` scripts uses a release of ACE6 and CMake uses a release of ACE7.
+
+The versions are updated by the GitHub Actions workflow in :ghfile:`.github/workflows/update-ace-tao.yml`.
+When a new version of ACE/TAO is available, the workflow creates a pull request based on the new versions.
```
iguessthislldo, line 69: This might or might not be worth mentioning here, but the update ACE/TAO workflow currently only automatically updates the version if it's a micro/patch release. Other releases require changing the `version` property and running the release script manually.

iguessthislldo (approved):

PR #4618, opened 2024-05-02
Test ACE's Minimum Required C++ Standard in CMake

iguessthislldo:
Since https://github.com/OpenDDS/OpenDDS/pull/4487, CMake requires knowledge of the minimum C++ standard of ACE. It currently uses the ACE version to assume the minimum C++ standard, but the required standard changed in ACE to C++17. To support this correctly in CMake, we could do a version check, but master currently is 7.1.4, the last ACE7 release.

These changes try to compile against `ace/Global_Macros.h` to get the minimum C++ standard of ACE like how we currently check the compiler default. This will allow us to get any standard ACE requires as long it's a standard we're aware of. We will not need this after https://github.com/OpenDDS/OpenDDS/issues/4578, because CMake can directly tell us what ACE/TAO requires.

These changes still don't cover a case like ACE/TAO 7.1.4, where ACE/TAO was compiled with something other the compiler default or the minimum standard of ACE. CMake can't detect that, so that requires manually setting the standard in either CMake and/or the ACE/TAO build so they match.

Also:

- Removed unused ACE/TAO and MPC checkouts in the CMake Android job. They are currently using the CMake default ACE7/TAO3 release. I don't think I did this on purpose. They could be changed to use checkouts, but:
  1. ACE/TAO would need to be checked out twice for both host and target or use the clone tree utility.
  2. We're trying to evaluate what specific ACE/TAO we want to use in CI (https://github.com/OpenDDS/OpenDDS/pull/4614) so just remove the unused checkouts for now.
- Fix check for default C++ standard in MSVC because it apparently doesn't define `__cplusplus` correctly.


cmake/cplusplus.cpp
```diff
@@ -1,4 +1,13 @@
-#if __cplusplus < OPENDDS_TEST_CPLUSPLUS
+// msvc doesn't define __cplusplus by default for some reason
```
mitza-oci, line 1: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

cmake/ace_cppstd.cpp
```diff
@@ -0,0 +1,7 @@
+#define ACE_DOESNT_DEFINE_MAIN 1
+#include <ace/Global_Macros.h>
+
+int main()
+{
+  return 0;
+}
```
mitza-oci, line 7: This could be combined with cplusplus.cpp and have different behavior based on a preprocessor macro

cmake/cplusplus.cpp
iguessthislldo: So the issue we saw earlier in standup where ace6tao3 was defaulting to C++03 was caused by this trying to pass `/Zc:__cplusplus`. When that option is manually set in visual studio, it works as intended. Trying to pass it via `CMAKE_CXX_FLAGS` to `try_compile` like the PR is currently trying to do results in the option being ignored because `try_compile` is overriding it. It seems like the correct way to either set the option using the `CXXFLAGS` environment variable or to make a `CMakeLists.txt` just for this cpp file. Neither are appealing to me. This combined with the fact that `/Zc:__cplusplus` doesn't work with all of the VS versions we claim to support, suggests this should  use `_MSVC_LANG`.

mitza-oci (approved):

PR #4634, opened 2024-05-15
Debugging output not available for Messenger test

jrw972:
Problem
-------

Debugging is not turned on for the Messenger test.

Solution
--------

Turn on debugging.

### DIFF
--- a/bin/PerlDDS/Run_Test.pm
+++ b/bin/PerlDDS/Run_Test.pm
@@ -595,7 +595,7 @@ sub _process_common {
   my $self = shift;
   my $name = shift;
   my $params = shift;
-  my $debug_logging = $self->{dcps_log_level};
+  my $debug_logging = 1;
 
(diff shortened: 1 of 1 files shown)


jrw972:
The ancillary goal is to debug some failures on different PRs.

iguessthislldo (changes requested):
I don't think this is necessary. I think we already discussed this. The better solution would be changing [this value](https://github.com/OpenDDS/OpenDDS/blob/81660e1d67b2230b029c0cbe2430ab4317129ea2/bin/PerlDDS/Run_Test.pm#L598) back to 1 so all tests work as expected.

mitza-oci:
Related change is https://github.com/OpenDDS/OpenDDS/commit/2605ead31cb4334c388f305e0fee89aa838bf857

iguessthislldo (approved):

Issue #4432, opened 2024-01-19
Inherit IDL Include Paths in `opendds_target_sources`


iguessthislldo:
In MPC, if you have a library with IDL that you want to use in the IDL of other targets, you can do something like [this](https://github.com/OpenDDS/OpenDDS/blob/b548b158a4297019040177fc5702d53b86c768df/MPC/config/dcps.mpb#L2). CMake doesn't have an equivalent to that though. There is a hard coded include for this specific include [here](https://github.com/OpenDDS/OpenDDS/blob/b548b158a4297019040177fc5702d53b86c768df/cmake/dds_idl_sources.cmake#L329), but not a generic solution. You'd have to pass `-I` to every usage of `opendds_target_sources` that needs to include the IDL.

One way to have the CMake support handle this would be to leverage [`INTERFACE_LINK_LIBRARIES`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html). Here's an example:

```cmake
add_library(x)
opendds_target_sources(x PUBLIC xidl/x/x.idl INCLUDE_BASE xidl)

add_library(y)
target_link_libraries(y PUBLIC x)
opendds_target_sources(y
  PUBLIC yidl/y/y.idl # includes "x/x.idl"
  INCLUDE_BASE yidl
)
```

[...]

PR #4659, opened 2024-05-22
Inconsistent setting of `OPENDDS_SECURITY` corrupts builds

jrw972:
Problem
-------

`OPENDDS_SECURITY` is a macro that appears in public headers. If a user enables security during the build of OpenDDS but then does not set it during the build of their application, then the build is inconsistent and the application fails in unexpected ways.  Regardless of the build system (MPC, CMake, or something else), the idea that security is enabled must be preserved to generate consistent builds.

Solution
--------

See https://github.com/OpenDDS/OpenDDS/issues/4616 for background.

Add `OPENDDS_CONFIG_SECURITY` to `dds/OpenDDSConfig.h`.

Update `dds/DCPS/Definitions.h` to accommodate both `OPENDDS_SECURITY` and `OPENDDS_CONFIG_SECURITY`.


dds/OpenDDSConfigWrapper.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef OPENDDS_DDS_OPENDDSCONFIG_WRAPPER_H
+#define OPENDDS_DDS_OPENDDSCONFIG_WRAPPER_H
+
+// OpenDDSConfig.h is generated by the configure script or by CMake
+// If you're manually configuring OpenDDS, start by creating an empty file
+// named OpenDDSConfig.h in a dds/ directory on the include path.
+// See the OpenDDS Developer's Guide and dds/OpenDDSConfig.h.in for details
+// regarding what can be customized.
+#ifndef OPENDDS_IGNORE_OPENDDSCONFIG_H_FILE
+#  include <dds/OpenDDSConfig.h>
```
iguessthislldo, line 15: Looks like the IDL compilers needs to `-I` the directory for this in CMake. It's generated here:

https://github.com/OpenDDS/OpenDDS/blob/85561d54d10af95f6ff05b06e34a279983550600/CMakeLists.txt#L62-L81

Possible options:

- Maybe if we had #4432, we could use it on the `OpenDDS_Config` target. I have the start of include inheriting as a prototype, but it needs a bit more design work, so it probably isn't viable for this PR.

- We could add `OPENDDS_IDL_OPTIONS -I${config_out_dir} TAO_IDL_OPTIONS -I${config_out_dir}` to the `opendds_target_sources` calls that need the config header.

- We just assume all IDL files should be able to include this and set `config_out_dir` to a cache variable that `opendds_target_sources` checks for and includes it. I think this one's the one I favor.

iguessthislldo (changes requested):

dds/OpenDDSConfigWrapper.h
iguessthislldo: https://github.com/OpenDDS/OpenDDS/blob/85561d54d10af95f6ff05b06e34a279983550600/cmake/opendds_target_sources.cmake#L387

iguessthislldo: https://github.com/OpenDDS/OpenDDS/blob/85561d54d10af95f6ff05b06e34a279983550600/cmake/opendds_target_sources.cmake#L353

Actual place were the check should go.

iguessthislldo (approved):

PR #4677, opened 2024-06-05
Updates to the InternalThreadStatus built-in topic and boottime timer support

mitza-oci:


docs/news.d/thread-status-boottime.rst
```diff
@@ -0,0 +1,12 @@
...
+
+.. news-start-section: Additions
+- Added a new data member, monotonic_timestamp to the InternalThreadBuiltInTopic IDL struct.
+  - monotonic_timestamp is the time of the sample was written (time of last update of this instance) on the monotonic clock.
+  - On systems that don't support a monotonic clock, this will be the same value as the corresponding SampleInfo's source_timestamp.
+.. news-end-section
+
+.. news-start-section: Fixes
+- When :cfg:prop:`DCPSThreadStatusInterval` is enabled, threads that run the ACE Reactor now use timers instead of a time-limited select() system call to update the InternalThreadBuiltInTopic.
+  - This allows the InternalThreadBuiltInTopic to be updated accurately on systems that suspend/resume and are configured for boottime timers.
```
iguessthislldo, line 11:
```suggestion
- Added a new data member, ``monotonic_timestamp`` to the ``InternalThreadBuiltInTopic`` IDL struct.

  - ``monotonic_timestamp`` is the time of the sample was written (time of last update of this instance) on the monotonic clock.
  - On systems that don't support a monotonic clock, this will be the same value as the corresponding ``SampleInfo``'s ``source_timestamp``.
.. news-end-section

.. news-start-section: Fixes
- When :cfg:prop:`DCPSThreadStatusInterval` is enabled, threads that run the ACE Reactor now use timers instead of a time-limited ``select()`` system call to update the ``InternalThreadBuiltInTopic``.

  - This allows the ``InternalThreadBuiltInTopic`` to be updated accurately on systems that suspend/resume and are configured for boottime timers.
```

dds/DCPS/ThreadStatusManager.h
```diff
@@ -188,6 +199,7 @@ class OpenDDS_Dcps_Export ThreadStatusManager {
   mutable ACE_Thread_Mutex lock_;
 };
 
+
```
iguessthislldo, line 202:
```suggestion
```

dds/DCPS/Time_Helper.h
```diff
@@ -23,14 +24,25 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
+  t.sec = ACE_Utils::truncate_cast<CORBA::Long>(tv.sec());
+  t.nanosec = ACE_Utils::truncate_cast<CORBA::ULong>(tv.usec() * 1000);
+}
 
 ACE_INLINE OpenDDS_Dcps_Export
 DDS::Time_t time_value_to_time(const ACE_Time_Value& tv);
 
 ACE_INLINE OpenDDS_Dcps_Export
-MonotonicTime_t time_value_to_monotonic_time(const ACE_Time_Value& tv);
+MonotonicTime_t time_value_to_time(const ACE_Time_Value_T<ACE_Monotonic_Time_Policy>& tv);
```
iguessthislldo, line 45:
```suggestion
MonotonicTime_t time_value_to_time(const ACE_Time_Value_T<MonotonicClock>& tv);
```

I don't know if this would actually cause a problem, but the clock type should probably be consistent with `MonotonicTimePoint`.

iguessthislldo (changes requested):

dds/DCPS/Time_Helper.h
mitza-oci: When I tried this change it failed to compile https://github.com/OpenDDS/OpenDDS/actions/runs/9417476930/job/25942737566  So for now it's reverted.

I agree with the idea, we would just need to work out the declaration order of the relevant pieces so that it doesn't lead to a cycle of `#include`s.

jrw972 (approved):

dds/DCPS/Time_Helper.h
iguessthislldo: Oh, yea, I guess `TimeTypes.h` needs this header... The clock definitions probably would need to go in their own header then?

iguessthislldo (approved):

PR #4681, opened 2024-06-07
Make CMake and Configure Fallback To Last Known VS

iguessthislldo:


iguessthislldo:
The Windows build failure should be fixed by either these changes, which will use the last known version or #4677, which explicitly adds the new cl version.

mitza-oci (approved):

jrw972 (approved):

PR #4683, opened 2024-06-07
Include ace/config.h before OpenDDSConfigWraper.h so that any setting…

jwillemsen:
…s in config.h are used

    * dds/DCPS/security/framework/HandleRegistry.cpp:

### DIFF
--- a/dds/DCPS/RTPS/ICE/AgentImpl.h
+++ b/dds/DCPS/RTPS/ICE/AgentImpl.h
@@ -5,6 +5,7 @@
  * See: http://www.opendds.org/license.html
  */
 
+#include <ace/config.h>
 #include "dds/OpenDDSConfigWrapper.h"
 
 #if OPENDDS_CONFIG_SECURITY
--- a/dds/DCPS/RTPS/ICE/Checklist.cpp
+++ b/dds/DCPS/RTPS/ICE/Checklist.cpp
@@ -5,6 +5,7 @@
  * See: http://www.opendds.org/license.html
  */
 
+#include <ace/config.h>
 #include <dds/OpenDDSConfigWrapper.h>
 
 #if OPENDDS_CONFIG_SECURITY
--- a/dds/DCPS/RTPS/ICE/EndpointManager.h
+++ b/dds/DCPS/RTPS/ICE/EndpointManager.h
@@ -5,6 +5,8 @@
  * See: http://www.opendds.org/license.html
  */
 
+#include <ace/config.h>
+#include <dds/OpenDDSConfigWrapper.h>
 
 #if OPENDDS_CONFIG_SECURITY
 #ifndef OPENDDS_DCPS_RTPS_ICE_ENDPOINTMANAGER_H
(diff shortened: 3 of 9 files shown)


iguessthislldo (changes requested):
This looks fragile. Even if these lines didn't change, we could easily add new checks that didn't include `ace/config.h`. I think `macros += OPENDDS_IGNORE_OPENDDSCONFIG_H_FILE` should be added to the `!openddsconfig_h_file` feature. Alternatively we could also add `ace/config.h` to the wrapper config, but we were worried about including `ace/config.h` in IDL. Also adding the macro to the feature makes it the one thing to do to avoid using the config header instead of having to both add the macro to `ace/config.h` and disable `openddsconfig_h_file`.

jwillemsen:
The fact that a header is included in a IDL file is in itself fragile. Maybe have a separate file for including in IDL? This PR worked for me

jwillemsen:
Including ace/config.h would expose a lot of ACE defines and headers to the IDL compilers, that will probably cause a lot of issues

iguessthislldo:
> The fact that a header is included in a IDL file is in itself fragile.

It's no different from an IDL file as long as it's just preprocessor statements. tao_idl has had [a shared C++/IDL header](https://github.com/DOCGroup/ACE_TAO/blob/master/TAO/tao/idl_features.h) for a while.

> This PR worked for me

Yes, I don't doubt that, but we will be converting more build config flags to this style eventually. I don't want your builds to break because we forgot to include `ace/config.h` somewhere. Adding the macro to the feature helps avoid that and simplifies things on your end.

jwillemsen:
When you don’t include ace/config.h everywhere before OpenDDSConfigWrapper.h there is a risk that a setting done by the user in config.h is not used at some point, not sure about the impact right now, but this could also cause issues later

iguessthislldo:
`ace/config.h` can't really be included in IDL like we've already touched on, so it's not possible to define `OPENDDS_CONFIG_SECURITY` in `ace/config.h` and have it show up in IDL where it's needed. This means the `dds/OpenDDSConfig*.h` files should be considered independent of `ace/config*.h` files, otherwise we wouldn't be able to consistently use the macros. The [design document](https://github.com/OpenDDS/OpenDDS/wiki/Proposal:-Header-file-for-compile%E2%80%90time-configuration#goals) doesn't explicitly say this, but I think it was an assumption (IDL issue as already mentioned, "The same file should be used for all configuration methods, platforms, build systems").

So I think OpenDDS-specific macros have to be defined either through `-D` options like before or in `dds/OpenDDSConfig.h`. OpenDDS should only include `ace/config*.h` when it needs a ACE/TAO-specific macro.

jwillemsen:
I am closing this PR, I am going to extend our build/axcioma tooling to generate `dds/OpenDDSConfig.h` as this new file is not just a configuration file in addition to `ace/config.h`, but a replacement. For example we enable DDS security when SSL is available on the system, that is something that has to be enabled in this new file `dds/OpenDDSConfig.h`

PR #4693, opened 2024-06-13
Publish the ParticipantLocation BIT before participant discovery completes

mitza-oci:
Applications can use ParticipantLocation to get notified that discovery
is in progress.  The spec-defined Participant BIT won't be published
until participant discovery is complete.


tests/DCPS/ParticipantLocationTopic/ParticipantLocationTest.cpp
```diff
@@ -359,6 +362,27 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
     // Call on_data_available in case there are samples which are waiting
     sub_listener->on_data_available(sub_loc_dr);
 
+    DDS::DataReader_var sub_part_dr = sub_bit_subscriber->lookup_datareader(OpenDDS::DCPS::BUILT_IN_PARTICIPANT_TOPIC);
+    if (!sub_loc_dr) {
+      ACE_ERROR((LM_ERROR,
+                 ACE_TEXT("%N:%l main()")
+                 ACE_TEXT(" ERROR: Could not get %C DataReader\n"),
+                 OpenDDS::DCPS::BUILT_IN_PARTICIPANT_TOPIC));
+      ACE_OS::exit(EXIT_FAILURE);
```
iguessthislldo, line 371: These could be returns?

iguessthislldo (approved):

tests/DCPS/ParticipantLocationTopic/ParticipantLocationTest.cpp
mitza-oci: Sure, I was just keeping it the same as what the test was already doing for the other BIT reader.

jrw972 (approved):

PR #4694, opened 2024-06-13
Update char_literals Tests to Check for `wchar` Escape Sequence Support in tao_idl

iguessthislldo:
https://github.com/DOCGroup/ACE_TAO/pull/2244 sets `TAO_IDL_HAS_OCTET_AND_WCHAR_UNION_DISCS` to 1, but the char_literals wchar union contains all the escape sequences literals, which are not supported in TAO_IDL: https://github.com/DOCGroup/ACE_TAO/issues/1284. This changes it to check the planned macro name.

Also removed the hex and octet escape sequences for wchar, which are probably not indented to work with wchar.

### DIFF
--- a/tests/DCPS/Compiler/char_literals/test.idl
+++ b/tests/DCPS/Compiler/char_literals/test.idl
@@ -1,8 +1,14 @@
 #if defined __TAO_IDL_FEATURES
 #  include __TAO_IDL_FEATURES
 #  define WCHAR_DISC TAO_IDL_HAS_OCTET_AND_WCHAR_UNION_DISCS
+#  ifdef TAO_IDL_HAS_WCHAR_ESCAPE_SEQUENCES
+#    define WCHAR_ESC_SEQ TAO_IDL_HAS_WCHAR_ESCAPE_SEQUENCES
+#  endif
 #else
 #  define WCHAR_DISC 0
+#endif
+#ifndef WCHAR_ESC_SEQ
+#  define WCHAR_ESC_SEQ  0
 #endif
 
 const char c_a = 'a';
@@ -23,9 +29,12 @@ const char c_hex_1 = '\x1';
 const char c_hex_254 = '\xfe';
 
 @topic
+const char c_const = '*';
 union CharUnion switch (@key char) {
 case 'a':
   long a;
+case c_const:
+  long const_branch;
 case '\n':
   long newline;
 case '\t':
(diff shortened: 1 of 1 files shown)


tests/DCPS/Compiler/char_literals/test.idl
```diff
@@ -1,9 +1,15 @@
...
 #  include __TAO_IDL_FEATURES
 #  define WCHAR_DISC TAO_IDL_HAS_OCTET_AND_WCHAR_UNION_DISCS
+#  ifdef TAO_IDL_HAS_WCHAR_ESCAPE_SEQUENCES
+#    define WCHAR_ESC_SEQ TAO_IDL_HAS_WCHAR_ESCAPE_SEQUENCES
+#  endif
 #else
 #  define WCHAR_DISC 0
 #endif
+#ifndef WCHAR_ESC_SEQ
+#  define WCHAR_ESC_SEQ  0
```
mitza-oci, line 11:
```suggestion
#  define WCHAR_ESC_SEQ 0
```

mitza-oci:
> Also removed the hex and octet escape sequences for wchar, which are probably not indented to work with wchar.

The spec allows them - why are they not going to work (or not intended to work)?

iguessthislldo:
To add to what I said as we were leaving, I wasn't sure, but was guessing they weren't allowed because 7.2.6.2.2 mentions only sequences lengths for a byte:

> The escape \ooo consists of the backslash followed by one, two, or three octal digits that are taken to specify the value
of the desired character. The escape \xhh consists of the backslash followed by x followed by one or two hexadecimal
digits that are taken to specify the value of the desired character.

I read the section again and I don't see something explicitly allowing them, but don't mind putting them back and allowing them when I get around to working to my changes in ACE/TAO again.

mitza-oci:
> To add to what I said as we were leaving, I wasn't sure, but was guessing they weren't allowed because 7.2.6.2.2 mentions only sequences lengths for a byte:
> 
> > The escape \ooo consists of the backslash followed by one, two, or three octal digits that are taken to specify the value
> > of the desired character. The escape \xhh consists of the backslash followed by x followed by one or two hexadecimal
> > digits that are taken to specify the value of the desired character.
> 
> I read the section again and I don't see something explicitly allowing them, but don't mind putting them back and allowing them when I get around to working to my changes in ACE/TAO again.

The spec goes out of its way to say that `\u` may not be used in char.  Based on what's there, `\o` and `\x` can represent values between 0 and 255 either in char or in wchar.  In wchar they are widened (0 padding) to the size of the wchar.

mitza-oci (approved):

PR #4695, opened 2024-06-13
Updated Java build (non-Android)

mitza-oci:
Generated makefile needs to see a .class file so it doesn't need to build every time make is run


jrw972 (approved):

iguessthislldo (approved):

mitza-oci:
`make install` for Java needs to get fixed before we merge this
https://github.com/OpenDDS/OpenDDS/actions/runs/9507279461/job/26206396526?pr=4695

PR #4685, opened 2024-06-07
Fix various compile warnings

jrw972:

PR #4700, opened 2024-06-14
Fixed warnings related to int-to-string conversions

mitza-oci:


dds/DCPS/SafetyProfileStreams.cpp
```diff
@@ -14,85 +14,67 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
...
-  } else {
-    fmt = "%u";
-    const int buff_size = 20 + 1; // note +1 for null terminator
-    char buf[buff_size];
-    ACE_OS::snprintf(&buf[0], buff_size, fmt, to_convert);
-    return String(buf);
-  }
+  static const int buff_size = 20 + 1; // note +1 for null terminator
+  char buf[buff_size];
+  ACE_OS::snprintf(&buf[0], buff_size, as_hex ? "%x" : "%u", to_convert);
```
jrw972, line 45: Did you intend to now use `%08x`?

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -281,7 +281,7 @@ void GuidAddrSet::remove(const OpenDDS::DCPS::GUID_t& guid,
   relay_stats_reporter_.local_active_participants(guid_addr_set_map_.size(), now);
 
   if (config_.log_activity()) {
-    ACE_DEBUG((LM_INFO, "(%P|%t) INFO: GuidAddrSet::remove_i "
+    ACE_DEBUG((LM_INFO, "(%P|%t) INFO: GuidAddrSet::remove "
```
jrw972, line 284:
```suggestion
    ACE_DEBUG((LM_INFO, "(%P|%t) INFO: GuidAddrSet::remove: "
```

jrw972 (changes requested):

dds/DCPS/SafetyProfileStreams.cpp
mitza-oci: No, I meant to come back to it after I had a better idea of what the formats should be.

tools/rtpsrelay/GuidAddrSet.cpp
mitza-oci: No other logging statement in this file uses a `:` here, so if we're going to change it we should make it consistent either by file or by project.

jrw972: Oh.  :(. Nevermind then.

iguessthislldo (approved):

mitza-oci:
- [ ] Update from @jwillemsen comment on #4685 
- [ ] Make hex formats fixed-width using leading 0s
- [ ] Add unit tests

jrw972 (approved):

PR #4707, opened 2024-06-21
`EncapsulationHeader::to_encoding_i` error messages lack detail

jrw972:
Problem
-------

The error messages from `EncapsulationHeader::to_encoding_i` lack enough detail to actually debug the error.  Typically, the issue is a mismatch in encoding, however, it is unclear which topic, writer, reader, etc. is involved.

Solution
--------

Add the writer and reader guids to the error messages.


mitza-oci (changes requested):
Instead of moving additional details into the EncapsulationHeader I think we should move to_encoding out to another class or function, maybe in XTypes dir/namespace.

iguessthislldo:
I'm not sure sure about that. `EncapsulationHeader`, like `Encoding` and `Serializer` is aware of XTypes, but isn't specific to XTypes.

mitza-oci:
> I'm not sure sure about that. `EncapsulationHeader`, like `Encoding` and `Serializer` is aware of XTypes, but isn't specific to XTypes.

I was referring to moving the `to_encoding` algorithm, not the `EncapsulationHeader`.  It doesn't have to move to XTypes, but it also doesn't have to be in `Serializer.h`.

iguessthislldo (approved):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3132,11 +3132,13 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
         if (data.smHeader.flags & (FLAG_D | FLAG_K_IN_DATA)) {
           DCPS::EncapsulationHeader encap;
           DCPS::Encoding enc;
-          if (!(ser >> encap) || !encap.to_encoding(enc, DCPS::MUTABLE) || enc.kind() != Encoding::KIND_XCDR1) {
+          if (!(ser >> encap) || !to_encoding(enc, encap, DCPS::MUTABLE)  || enc.kind() != Encoding::KIND_XCDR1) {
```
mitza-oci, line 3135:
```suggestion
          if (!(ser >> encap) || !to_encoding(enc, encap, DCPS::MUTABLE) || enc.kind() != Encoding::KIND_XCDR1) {
```

mitza-oci (changes requested):

mitza-oci (approved):

iguessthislldo (approved):

PR #4710, opened 2024-06-25
Parse InfoDestinationSubmessages in incoming SPDP messages

mitza-oci:
If InfoDestinationSubmessage indicates this isn't directed at the current participant, the following DataSubmessage(s) are dropped.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -3100,6 +3100,8 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
       message.hdr = header;
     }
 
+    DCPS::GuidPrefix_t destinationGuidPrefix = {0};
+
     while (buff_.length() > 3) {
       const char subm = buff_.rd_ptr()[0], flags = buff_.rd_ptr()[1];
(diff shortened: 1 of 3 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3100,6 +3100,8 @@ Spdp::SpdpTransport::handle_input(ACE_HANDLE h)
       message.hdr = header;
     }
 
+    DCPS::GuidPrefix_t destinationGuidPrefix = {0};
```
jrw972, line 3103:
```suggestion
    DCPS::GuidPrefix_t destinationGuidPrefix = DCPS::GUIDPREFIX_UNKNOWN;
```

jrw972: Also, it appears that this variable is not set.

jrw972 (changes requested):

dds/DCPS/RTPS/Spdp.cpp
mitza-oci: > (diff with GUID_PREFIX_UNKNOWN)

Unfortunately, arrays can't be initialized that way.

jrw972 (approved):

PR #4712, opened 2024-06-27
[fix] RtpsDiscoveryConfig RESEND_PERIOD

jmmorato:
Use milliseconds in order to don't lose the microseconds in the TimeDuration


dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
```diff
@@ -32,15 +32,15 @@ RtpsDiscoveryConfig::resend_period() const
   // see RTPS v2.1 9.6.1.4.2
   return TheServiceParticipant->config_store()->get(config_key("RESEND_PERIOD").c_str(),
                                                     TimeDuration(30 /*seconds*/),
-                                                    DCPS::ConfigStoreImpl::Format_IntegerSeconds);
+                                                    DCPS::ConfigStoreImpl::Format_IntegerMilliseconds);
```
iguessthislldo, line 35: `Format_FractionalSeconds`? I think `Format_IntegerMilliseconds` would make all existing usage of `ResendPeriod` much much shorter.

iguessthislldo (changes requested):

iguessthislldo (changes requested):
Alright, the only things I think are left are that the documentation and news should be updated. For documentation, the only other property that's using fractional seconds is `AuthResendPeriod`:

https://github.com/OpenDDS/OpenDDS/blob/e0cd8c8bdeceef9c67bf70e38395f88fd84c1938/docs/devguide/run_time_configuration.rst?plain=1#L1381-L1385

It says it's a floating point value, so we should copy that line to `ResendPeriod`'s description:

https://github.com/OpenDDS/OpenDDS/blob/e0cd8c8bdeceef9c67bf70e38395f88fd84c1938/docs/devguide/run_time_configuration.rst?plain=1#L1106-L1109

Then for the news, I don't know if remember if you've made a PR since we added systems for the new and config properties, so I'll give it to you. Create a rst file in https://github.com/OpenDDS/OpenDDS/blob/master/docs/news.d along the lines of:

```rst
.. news-prs: 4712

- :prop:`[rtps_discovery]ResendPeriod` now accepts fractions of a second.
.. news-end-section

jrw972 (approved):

iguessthislldo (approved):

PR #4716, opened 2024-06-28
Removed the "expand" section of rapidjson.mpb

mitza-oci:
Expanding the variable here is useful because it provides a default value when RAPIDJSON_ROOT is not set in the environment.

However, it always results in makefiles that hard-code the expansion value (absolute path to rapidjson, may be in DDS_ROOT).

With this change, if the MPC feature no_rapidjson is 0 then the RAPIDJSON_ROOT environment variable needs to be set either by the configure script or manually.

### DIFF
--- a/MPC/config/rapidjson.mpb
+++ b/MPC/config/rapidjson.mpb
@@ -1,10 +1,5 @@
 project {
   avoids += no_rapidjson
-
-  expand(RAPIDJSON_ROOT) {
-    $RAPIDJSON_ROOT
-    $(DDS_ROOT)/tools/rapidjson
-  }
   includes += $(RAPIDJSON_ROOT)/include
   macros   += OPENDDS_RAPIDJSON
 }


jrw972 (approved):

jwillemsen:
Is there no option in MPC to not let expand the variable in the generated makefile?

mitza-oci:
> Is there no option in MPC to not let expand the variable in the generated makefile?

Not that I found.  A specific variable used with `expand` seems to take precedence over other options that apply more generally.

jrw972 (approved):

mitza-oci:
@jwillemsen is this going to work with the builds you maintain?

jwillemsen:
Just back from a week of vacation, I will check this on Monday @mitza-oci

iguessthislldo (approved):

jwillemsen:
Maybe related to https://github.com/DOCGroup/MPC/issues/102, will check our builds next week

mitza-oci:
> As far as I can tell it will break our builds because RAPIDJSON_ROOT is not by our build scripts, but I can add that when there is no other solution for this issue.

It's not urgent, so I'll see what else we can do before merging.

If we merge this then the build setup would be similar to what's done for googletest with GTEST_ROOT.

PR #4725, opened 2024-07-11
Remove Broken Links, Fix TypeLookup Secure Entity Names

iguessthislldo:
There were links that should have gone to https://issues.omg.org/issues/DDSSEC12-86 that were wrong. On closer inspection, the issue was fixed in v1.2 of the DDS security spec, which just came out. It looks like what we have matches what's in 1.2, so I removed the links.

I also switched the "reader"/"writer" and "secure" in names because it's what's in 1.2 and matches the other builtin entities. Also it's bugged me before.


mitza-oci:
Are there other sections of the devguide that need to change to refer to the 1.2 version of DDS Security?

iguessthislldo:
That depends on if we want to "upgrade" the spec version. I don't know the status of the other spec issues and if there's something new in 1.2 we need to explicitly say we don't support. If we do, I can only see one explicit reference to 1.1 here:

https://github.com/OpenDDS/OpenDDS/blob/7d3a82fbba24e3914d6d30a8bfe27b939343db46/docs/devguide/dds_security.rst?plain=1#L936

We update all other links and references by just changing the version here:

https://github.com/OpenDDS/OpenDDS/blob/7d3a82fbba24e3914d6d30a8bfe27b939343db46/docs/conf.py#L54

That might break spec section references in the devguide. There's also more references to 1.1 in the code, but I don't think I'm as concerned about those.

jrw972 (approved):

PR #4729, opened 2024-07-15
RtpsRelay tests don't have CheckSourceIp=false

jrw972:
Problem
-------

The RtpsRelay tests are a starting poing to users that want to use the RtpsRelay.  In any realistic deployment, the user must have `CheckSourceIp=false` in the configuration file.  Since the tests don't have this line, the user will have to learn "the hard way."

Solution
--------

Add `CheckSourceIp=false` to the configuration files.

### DIFF
--- a/tests/DCPS/RtpsRelay/Smoke/relay1.ini
+++ b/tests/DCPS/RtpsRelay/Smoke/relay1.ini
@@ -23,3 +23,4 @@ RtpsRelayOnly=1
 SedpMaxMessageSize=1400
 UndirectedSpdp=0
 PeriodicDirectedSpdp=1
+CheckSourceIp=false
--- a/tests/DCPS/RtpsRelay/Smoke/relay2.ini
+++ b/tests/DCPS/RtpsRelay/Smoke/relay2.ini
@@ -23,3 +23,4 @@ RtpsRelayOnly=1
 SedpMaxMessageSize=1400
 UndirectedSpdp=0
 PeriodicDirectedSpdp=1
+CheckSourceIp=false


dds/DCPS/Serializer.inl
```diff
@@ -830,15 +830,15 @@ bool Serializer::align_w(size_t al)
...
+  return static_cast<unsigned char>((reinterpret_cast<size_t>(index) - start) % align);
 }
 
 ACE_INLINE void
 Serializer::align_cont_r()
 {
   const size_t max_align = encoding().max_align();
   const size_t thisblock =
-    max_align ? (ptrdiff_t(current_->rd_ptr()) - align_rshift_) % max_align : 0;
+    max_align ? static_cast<size_t>((ptrdiff_t(current_->rd_ptr()) - align_rshift_) % max_align) : 0;
```
mitza-oci, line 841: This line is generating warnings

mitza-oci (approved):

dds/DCPS/Serializer.inl
```diff
@@ -783,7 +783,7 @@ bool Serializer::align_r(size_t al)
   }
   al = (std::min)(al, encoding().max_align());
   const size_t len =
-    (al - ptrdiff_t(current_->rd_ptr()) + align_rshift_) % al;
+    (al - reinterpret_cast<size_t>(current_->rd_ptr()) + align_rshift_) % al;
```
mitza-oci, line 786: This previously used `ptrdiff_t` since that's an integer type guaranteed to be as wide as the pointer type.  For portability / future compatibly I'd like to maintain that property.  Is there a way to use `ptrdiff_t` and avoid warnings?

mitza-oci (changes requested):

dds/DCPS/Serializer.inl
jrw972: The problem is that `al` is unsigned so whatever `current->rd_ptr()` is cast to needs to be unsigned to prevent the warning.

mitza-oci: Maybe the whole expression needs a re-write?  It really only matters what the low few bytes of the rd_ptr are (which is `rd_ptr % al` with `al` being a power of two.

mitza-oci (approved):

PR #4731, opened 2024-07-18
Fix Locator Memory Leak in RTPS/UDP Transport

iguessthislldo:
Locators were not being cleaned up correctly because there were two functions for cleaning up after remotes in `RtpsUdpDataLink`. These changes merge those two functions.

Also:
- Don't create locator entries for unknown remotes in `RtpsUdpDataLink::update_locators`
- Change `RemoteInfo::ref_count_` from `size_t` to `DDS::UInt32` to save a bit of space in 64-bit builds. For example, if the RtpsRelay SEDP transport is serving 10,000 participants with a `RemoteInfo` for each participant's 12 remote builtin entities, then this saves almost half a megabyte of memory.
- Methods for getting GUIDs from `DiscoveredParticipant` to simplify code
- Remove an unused function in Sedp

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
@@ -957,14 +957,14 @@ RtpsUdpDataLink::release_reservations_i(const GUID_t& remote_id,
     RemoteInfoMap::iterator pos = locators_.find(remote_id);
     if (pos != locators_.end()) {
       OPENDDS_ASSERT(pos->second.ref_count_ > 0);
-
-      --pos->second.ref_count_;
-      if (pos->second.ref_count_ == 0) {
+      if (--pos->second.ref_count_ == 0) {
         locators_.erase(pos);
       }
     } else if (Transport_debug_level >= 4) {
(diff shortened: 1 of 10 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -956,6 +951,25 @@ RtpsUdpDataLink::release_reservations_i(const GUID_t& remote_id,
...
 
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, locators_lock_);
+
+    RemoteInfoMap::iterator pos = locators_.find(remote_id);
+    if (pos != locators_.end()) {
+      OPENDDS_ASSERT(pos->second.ref_count_ > 0);
+
+      --pos->second.ref_count_;
+      if (pos->second.ref_count_ == 0) {
```
jwillemsen, line 962: combined in one line? 
```
if (--pos->second.ref_count_ == 0) {
```

jwillemsen, line 962: ``` suggestion ```
      if (--pos->second.ref_count_ == 0) {
```

mitza-oci (approved):

jrw972 (approved):

PR #4747, opened 2024-07-25
Address warnings

sonndinh:
This addresses most of the warnings that come from OpenDDS, except this warning in `dds/DCPS/Hash.cpp` that needs further look: `[build_m13_j_FM-1f: dds/DCPS/Hash.cpp#L135]
cast from 'const unsigned char *' to 'const MD5_u32plus *' (aka 'const unsigned int *') increases required alignment from 1 to 4 [-Wcast-align]`

The rest of the warnings come from ACE, TAO, and generated code by `tao_idl` which are the majority of all warnings.

### DIFF
--- a/dds/idl/metaclass_generator.cpp
+++ b/dds/idl/metaclass_generator.cpp
@@ -524,8 +524,8 @@ metaclass_generator::gen_typedef(AST_Typedef*, UTL_ScopedName* name,
   AST_Type* type, const char*)
 {
   AST_Array* arr = dynamic_cast<AST_Array*>(type);
-  AST_Sequence* seq = 0;
-  if (!arr && !(seq = dynamic_cast<AST_Sequence*>(type))) {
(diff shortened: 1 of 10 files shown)


dds/idl/metaclass_generator.cpp
```diff
@@ -524,8 +524,8 @@ metaclass_generator::gen_typedef(AST_Typedef*, UTL_ScopedName* name,
   AST_Type* type, const char*)
 {
   AST_Array* arr = dynamic_cast<AST_Array*>(type);
-  AST_Sequence* seq = 0;
-  if (!arr && !(seq = dynamic_cast<AST_Sequence*>(type))) {
+  AST_Sequence* seq = dynamic_cast<AST_Sequence*>(type);
+  if (!arr && !seq) {
```
mitza-oci, line 528: I don't think this matters all that much, but the modified code doesn't take advantage of the short-circuit `&&` control flow that we had before.

jrw972 (approved):

PR #4754, opened 2024-07-31
Release process lacks detail on micro releases and using containers

jrw972:
Problem
-------

- The release process cannot be performed on MacOS.

- There were some assumptions about micro releases that were not stated.

Solution
--------

- Add guidance for using a container to perform the release.

- Add detail about preparing the release branch for a micro release.

### DIFF
--- a/docs/internal/release.rst
+++ b/docs/internal/release.rst
@@ -236,6 +236,8 @@ Before Running the Release Script
 
         cpan -T -i Pithub Net::SFTP::Foreign Time::Piece LWP::UserAgent LWP::Protocol::https
 
+    If Pithub gives errors about https url not be supported, then it might be necessary to install `Net::SSLeay <https://metacpan.org/pod/Net::SSLeay>`__ which also requires openssl.
+
 - Choose a directory for the ``WORKSPACE`` argument.
   It doesn't have to exist but the release script must be able to create it if it doesn't.
   It should not contain files created by previous release (mocked or otherwise).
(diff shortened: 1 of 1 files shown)


docs/internal/release.rst
```diff
@@ -187,6 +187,15 @@ The workflows are:
...
 These will be triggered by the release script, then must be :ref:`uploaded after release <release-upload-artifacts>`.
 
+Preparing for a Micro Release
+=============================
+
+As stated earlier, a micro release only contains fixes.
+If a bug is found in a release version, the fix should first be applied to the development branch for the corresponding major version and the master branch if they are not the same.
+To start a micro release, cherry pick one or more fixes from the development branch to the release branch.
+Release branches have the form ``branch-DDS-X.Y``.
+The cherry-picked commits should include the relevant news files.
```
iguessthislldo, line 197: This should link to (or be merged with) the other section describing micro releases: https://github.com/OpenDDS/OpenDDS/blob/abe12bcb6807b560b58a241f8a88e0d818298cb9/docs/internal/release.rst?plain=1#L340
Also there's a sub command (`--cherry-pick-prs`) to help with cherry picking.

docs/internal/release.rst
```diff
@@ -257,6 +266,37 @@ Before Running the Release Script
...
+Performing the release from a container is recommended as it provides a standardized and repeatable environment.
+
+The following script demonstrates how to prepare a container to perform a release:
+
+  .. code-block:: bash
+
+      podman run --rm -ti ubuntu /bin/bash
+      apt update
+      apt install zip git g++ make libssl-dev zlib1g-dev rustc cargo python3 python3-venv vim
+      cpan -T -i Pithub Net::SFTP::Foreign Time::Piece LWP::UserAgent LWP::Protocol::https Net::SSLeay
```
iguessthislldo, line 281: Why is `Net::SSLeay` needed?

docs/internal/release.rst
```diff
@@ -257,6 +266,37 @@ Before Running the Release Script
...
       export READ_THE_DOCS_TOKEN=ff00ff00ff00ff00ff00ff00ff00ff00ff00ff00
 
+Releasing from a Container
+==========================
+
+Performing the release from a container is recommended as it provides a standardized and repeatable environment.
+
+The following script demonstrates how to prepare a container to perform a release:
+
+  .. code-block:: bash
```
iguessthislldo, line 276:
```suggestion
.. code-block:: bash
```

Extra indent inserts a quote block: https://opendds--4754.org.readthedocs.build/en/4754/internal/release.html#releasing-from-a-container

iguessthislldo (changes requested):

docs/internal/release.rst
jrw972: In my container, Pithub fails without it.

iguessthislldo: Do you know if this is a new requirement of PitHub? If it is, we should update the instructions for Perl dependencies.

jrw972: It is not listed as a dependency at https://metacpan.org/pod/Pithub.
So it must be an optional dependency somewhere down the chain.
It was definitely necessary when using the container.

Maybe we mention it as an optional thing that you may need to install?

iguessthislldo: I just deleted my `local::lib` dir and reinstalled the dependencies from scratch and it didn't install `Net::SSLeany`. Running the script with no options is fine. Is it a runtime error when actually trying to use PitHub during the release process?

Part of me wants to know why it's needed, but it's probably okay to go forward with this, especially if it's only required when running in a container.

iguessthislldo (approved):

docs/internal/release.rst
jrw972: It was definitely a runtime issue.
I don't have an example of the failure but it contained the url and something like "https protocol not supported."

iguessthislldo: I ran some PitHub code and it works without it on my machine so it's not always required.

iguessthislldo (approved):

PR #4772, opened 2024-08-16
Fixed Leak in `AddressCache` for RTPS/UDP Transport

iguessthislldo:
AddressCache has a `map_` that maps keys (local-remote GUID pairs in RTPS/UDP) to cache entries and a `id_map_` that maps GUIDs to a vector of related keys.

`id_map_` has two problems:
- The vector of keys are not checked if elements are being duplicated. These changes make it a set.
- `remove_id` removes the `id_map_` keys for the GUID argument (the remote GUID in RTPS/UDP), but not any keys that contain that GUID for other vectors in `id_maps_` (the entries under local GUIDs in RTPS/UDP). These changes try to clean up all the relevant GUIDs.

Also:
- A change to release reservation I meant to add in the other leak PR. If the release reservation is for a specific local-remote pair, it doesn't seem right to remove all messages for the remote from the send queue.
- Small fixes to the QoS documentation that also I had in this local repo for some reason.

### DIFF
--- a/dds/DCPS/AddressCache.h
+++ b/dds/DCPS/AddressCache.h
@@ -17,9 +17,6 @@
 #include "PoolAllocator.h"
 #include "RcObject.h"
 #include "TimeTypes.h"
-#include "dcps_export.h"
-
-#include <ace/INET_Addr.h>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 
@@ -30,15 +27,18 @@ typedef OPENDDS_SET_CMP(GUID_t, GUID_tKeyLessThan) GuidSet;
 
 struct AddressCacheEntry : public virtual RcObject {
 
-  AddressCacheEntry() : addrs_(), expires_(MonotonicTimePoint::max_value)
+  AddressCacheEntry()
+    : expires_(MonotonicTimePoint::max_value)
 #if defined ACE_HAS_CPP11
     , addrs_hash_(0)
 #endif
   {}
 
-  AddressCacheEntry(const NetworkAddressSet& addrs, const MonotonicTimePoint& expires) : addrs_(addrs), expires_(expires)
+  AddressCacheEntry(const NetworkAddressSet& addrs, const MonotonicTimePoint& expires)
+    : addrs_(addrs)
+    , expires_(expires)
 #if defined ACE_HAS_CPP11
-  , addrs_hash_(calculate_hash(addrs_))
+    , addrs_hash_(calculate_hash(addrs_))
 #endif
   {}
 
@@ -209,7 +209,7 @@ public:
   bool remove(const Key& key)
   {
     ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
-    return map_.erase(key) != 0;
+    return remove_i(key);
   }
 
   void remove_id(const GUID_t& val)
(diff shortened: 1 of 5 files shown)


iguessthislldo:
TODO:
- Confirm this fixes relay memory leaks.
- News
- `AddressCache` has unit tests that should be updated to test this.

simpsont-oci:
A few thoughts:
 - `id_map_` is purely there as a performance enhancement... the slow way would be to iterate over the entire `map_` and remove any keys that had the GUID in the set returned by `get_contained_guids`. But that's prohibitively slow, so let's not do that.
 - The original intent was for the GUID argument to `remove_id` to either be a remote or local GUID (not just remote). This is why `get_contained_guids` currently always adds both the local and remote GUID (see BundlingCacheKey, LocatorCacheKey, and TestKey implementations). But ultimately it's up to the key type... someone could theoretically make a key type that only added one or the other, and then `remove_id` would only remove keys where the remote or local GUID matched.
 - As for the reason why `id_map_` contains vectors and not sets: it's been a long time but I'm pretty sure this was done for performance reasons based on profiling. I agree that set is a more natural fit here, but if I recall correctly, I think I saw a lot of profiling overhead when deleting large sets (the cache can purge entries fairly often, either based on time or activity) compared to deleting a vector with a single large chunk of contiguous memory.
 - Unfortunately, I didn't document this decision or capture the actual numbers to justify it. I believe I was using the larger bench `disco` tests to profile the cache in the centipede test environment, as those seemed to put a lot of strain on both the bundling and locator caches. And without the performance dashboard (which I assume no longer exists), it'll be hard to tell how much of an impact a set would have versus a vector.

dds/DCPS/AddressCache.h
```diff
@@ -222,16 +216,41 @@ class AddressCache {
...
-        map_.erase(*it);
-      }
+      keys.swap(pos->second);
       id_map_.erase(pos);
     }
+
+    for (typename KeySet::iterator key = keys.begin(), limit = keys.end(); key != limit; ++key) {
+      map_.erase(*key);
+
+      // Undo insert_ids(key)
```
simpsont-oci, line 228: Ah, yeah, I see the problem now. Removing one GUID's `id_map_` entry still potentially leaves ejected keys in other GUID's `id_map_` entry. It doesn't impact cache correctness, but those keys could potentially stick around for a long time. The need to remove the keys here makes me think it really does need to be a set (and not just a sorted vector or whatever).

simpsont-oci:
As for a change to the tests... perhaps expose an "isEmpty" or a "contains(GUID)" method on the address cache and then make a small test like:
 - Add key(GUID_A, GUID_B)
 - remove_id(GUID_B)
 - check "isEmpty" is true and/or "contains(GUID_A)" is false

dds/DCPS/AddressCache.h
iguessthislldo: Yea, if there's a ever-changing set of remote participants, the vectors under the local GUIDs grow without limit because I don't think there's anyway for them to be removed.

iguessthislldo:
> * `id_map_` is purely there as a performance enhancement... the slow way would be to iterate over the entire `map_` and remove any keys that had the GUID in the set returned by `get_contained_guids`. But that's prohibitively slow, so let's not do that.
>
> * The original intent was for the GUID argument to `remove_id` to either be a remote or local GUID (not just remote). This is why `get_contained_guids` currently always adds both the local and remote GUID (see BundlingCacheKey, LocatorCacheKey, and TestKey implementations). But ultimately it's up to the key type... someone could theoretically make a key type that only added one or the other, and then `remove_id` would only remove keys added GUID (remote XOR local) matched.

A train of thought I had was maybe there was an alternative design that would avoid `id_map_` using `lower_bound` on `map_` if we assumed the remote was the first in the key. I haven't put much more though into that though other than that this probably doesn't work on unordered maps and this fix is kind of critical so I'm fine with leaving `id_map_` mostly how it is.

mitza-oci (approved):

iguessthislldo:
> probably doesn't work on unordered map

Yea, `std::unordered_map` doesn't have `lower_bound`, which makes sense. Maybe it's worth using ordered maps, but I could see the cost of maintaining that ordered map might outweigh the cost of removal.

mitza-oci (approved):

jrw972 (approved):

PR #4774, opened 2024-08-19
Added -AdmissionMaxParticipantsRange to RtpsRelay options

mitza-oci:
provides another mechanism for detecting load on each RtpsRelay instance

### DIFF
--- a/docs/devguide/internet_enabled_rtps.rst
+++ b/docs/devguide/internet_enabled_rtps.rst
@@ -293,6 +293,12 @@ The command-line options for the RtpsRelay:
   New client participants in the :option:`admission control queue <-AdmissionControlQueueSize>` that are taking longer than this many seconds to perform discovery are removed from the queue.
   Default is 0.
 
+.. option:: -AdmissionMaxParticipantsRange <low>-<high>
+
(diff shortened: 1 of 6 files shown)


docs/devguide/internet_enabled_rtps.rst
```diff
@@ -293,6 +293,12 @@ The command-line options for the RtpsRelay:
   New client participants in the :option:`admission control queue <-AdmissionControlQueueSize>` that are taking longer than this many seconds to perform discovery are removed from the queue.
   Default is 0.
 
+.. option:: -AdmissionMaxParticipantsRange <low>-<high>
+
+  If this option is used, the relay's "admitting" state will be set to false when the number of active participants
+  reaches the <high> value.  After the number of active participants falls below the <low> value, this metric will
+  cease to cause the "admitting" state to be false, although it may remain false due to other controls.
```
jrw972, line 300:
```suggestion
  cease to cause the "admitting" state to be true, although it may remain false due to other controls.
```

jrw972 (changes requested):

jrw972 (approved):

PR #4777, opened 2024-08-20
Restrict Vector Reserve Count in `RtpsUdpDataLink`

iguessthislldo:
If the number of remotes and `meta_submessages` is large, then this will waste a lot of space.


mitza-oci (approved):

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.cpp
```diff
@@ -2436,6 +2436,7 @@ RtpsUdpDataLink::build_meta_submessage_map(MetaSubmessageVec& meta_submessages,
   size_t cache_misses = 0;
   size_t addrset_min_size = std::numeric_limits<size_t>::max();
   size_t addrset_max_size = 0;
+  const size_t reserve_count = std::min(meta_submessages.size(), 64);
```
simpsont-oci, line 2439: The 64 should probably be a constant defined somewhere?

mitza-oci (approved):

PR #4779, opened 2024-08-22
Opendds4 prep - Do not merge

jrw972:


iguessthislldo:
You can remove: https://github.com/OpenDDS/OpenDDS/blob/00d1d5cff0fae21e1787df6a2e9ccddfb2ee732c/tests/auto_run_tests.pl#L89-L92

bin/PerlDDS/Run_Test.pm
```diff
@@ -376,6 +376,8 @@ sub new {
   my $class = shift;
   my $self = bless {}, $class;
 
+  $self->{configs} = shift;
+  $self->{config} = shift;
```
iguessthislldo, line 380: I don't know if Perl has a term for this, but if these are optional consider making these "keyword" arguments as Python calls them. This way they are explicit and it makes it much easier to add more arguments in the future in a backwards compatible way. The idea is because arrays and hashes are basically the same thing in Perl you can do:

```perl
my $test = new PerlDDS::TestFramework(configs => \%configs, config => 'ir_tcp');
```

Then treat `@_` as a hash that can be joined with a default argument hash.

Here's an example of this I've implemented before:

https://github.com/OpenDDS/OpenDDS/blob/07ee200864c1b58616cf5e3ab3dc865fb71d84ca/tools/scripts/modules/command_utils.pm#L264-L282

Here I'd recommend to something like:

```perl
# At the top we'd need to import trace since we're currently
# not using in the test framework, but really should:
use FindBin;
use lib "$FindBin::RealBin/../../tools/scripts/modules"
use misc_utils qw(trace);

# ...

  # At where I'm commenting replace the bless, configs, and config lines with:
  my %valid_args = (
    configs => {},
    config => undef,
  );

  my %args = (%valid_args, @_);
  my @invalid_args = grep { !exists($valid_args{$_}) } keys(%args);
  trace("invalid arguments: ", join(', ', @invalid_args)) if (scalar(@invalid_args));
  my $self = bless(\%args, $class);
```

Because `args` is the basis of the `bless`, `configs` and `config` will be in `$self`. We could also move most or all of the other values below into the `valid_args` hash to allow overriding them from the constructor.

bin/PerlDDS/Run_Test.pm
```diff
@@ -420,6 +422,12 @@ sub new {
     my $flag_value_str = defined($flag_value) ? "\"$flag_value\"" : 'undef';
     $self->_info("TestFramework storing \"$flag_name\"=$flag_value_str\n");
     $self->{_flags}->{$flag_name} = $flag_value;
+
+    if (exists $self->{configs}->{$flag_name}) {
+        $self->_info("TestFramework switching to config $flag_name\n");
+        $self->{config} = $flag_name;
```
iguessthislldo, line 428: I know this and many Perl files aren't consistent, but [we should prefer 2 spaces to 4 spaces](https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#perl-coding-style).

iguessthislldo (changes requested):

bin/PerlDDS/Run_Test.pm
```diff
@@ -454,6 +462,29 @@ sub new {
...
+      my $config = $self->{configs}{$self->{config}}{file};
+      keys %$config;
+      while (my ($s, $t) = each %$config) {
+          keys %$t;
+
+          if ($s =~ /\//) {
+              my @parts = split /\//, $s;
+              my $key = "OPENDDS_${s}";
+              my $value = "\@$parts[1]";
+              $ENV{$key} = $value;
```
iguessthislldo, line 477: I'm pretty sure this isn't going to work in a remote-controlled cross-compiled target.

jrw972: That's true.  PerlACE TestTargets have `SetEnv` which we can use to set environment variables in this case.

PR #4786, opened 2024-08-29
Turn Off "Fail on Warning" for RTD

iguessthislldo:


iguessthislldo:
Read the docs injects a sphinx extension that integrates the docs with the rest of the site. This is good, but if this extension causes any warnings, then the RTD build fails for no good reason like it did here: https://readthedocs.org/projects/opendds/builds/25463180/

We also already have a "strict" Sphinx job in GHA that checks for warnings.

jrw972 (approved):

PR #4799, opened 2024-09-17
Fix OMG Spec Links Page

iguessthislldo:
[This page](https://opendds.readthedocs.io/en/master/internal/omg_spec_links.html) is supposed to list links to all the sections in all the spec, not just the specs like it does now. It was causing a lot of unwanted output in the link checker, so I tried to make it conditional in https://github.com/OpenDDS/OpenDDS/pull/4669. Unfortunately it looks like I broke the page so it never lists all the possible spec section links. These changes fix that.

Also added `--nitpicky`, `--keep-going`, and `--show-traceback` to the `sphinx-build` command of `build.py strict` to make output more useful.

### DIFF
--- a/docs/build.py
+++ b/docs/build.py
@@ -118,11 +118,16 @@ class DocEnv:
 
     def do_strict(self):
         self.do(['test'], because_of='strict')
-        self.sphinx_build('dummy', '-W')
+        self.sphinx_build('dummy',
+            '--nitpicky',
+            '--fail-on-warning',
+            '--keep-going',
+            '--show-traceback',
+        )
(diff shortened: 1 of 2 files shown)


iguessthislldo:
To give a little more background, these links are part of the [`omgspec` role](https://opendds--4799.org.readthedocs.build/en/4799/internal/docs.html#role-omgspec) and this is what the page is supposed to look like: https://opendds--4799.org.readthedocs.build/en/4799/internal/omg_spec_links.html

jrw972 (approved):

PR #4801, opened 2024-09-19
Fix `ctest-to-auto-run-tests.py` Output on Windows

iguessthislldo:
There's a check in the CTest scripts on Windows for the build configuration. If the configuration isn't valid, the script would return invalid results and lead to the fake auto_run_tests name being truncated. These changes pass the build configuration from `auto_run_tests.pl` and add extra checks to prevent this from happening again.

Also changed `auto_run_tests.pl` CMake output to make it linkable in the matrix changes in https://github.com/DOCGroup/autobuild/pull/74.

### DIFF
--- a/tests/auto_run_tests.pl
+++ b/tests/auto_run_tests.pl
@@ -381,7 +381,7 @@ foreach my $test_lst (@file_list) {
 }
 
 if ($cmake) {
-    my $fake_name = "Run CMake Tests";
+    my $fake_name = "tests/cmake/CMakeLists.txt";
     if (!$list_tests) {
         mark_test_start($fake_name);
     }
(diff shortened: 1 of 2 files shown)


mitza-oci (approved):

iguessthislldo:
I checked the autobuild output and it looks good:

```
auto_run_tests: DevGuideExamples/DCPS/Messenger/CMakeLists.txt run_test.pl
...
auto_run_tests: DevGuideExamples/DCPS/Messenger/CMakeLists.txt run_test.pl --rtps
....
```

Once this is merged, we can merge https://github.com/DOCGroup/autobuild/pull/74 and https://github.com/OpenDDS/scoreboard/pull/15.

jrw972 (approved):

PR #4828, opened 2024-10-22
Remove Spurious Error Message When Parsing Member Path

sonndinh:


iguessthislldo (approved):
I'm not sure how I didn't notice this before. It looks like it's not unit tested though, so maybe that's why. :face_with_diagonal_mouth:

jrw972:
> I'm not sure how I didn't notice this before. It looks like it's not unit tested though, so maybe that's why. 🫤

What would it take to add a unit test?

iguessthislldo:
I think a lot of variation is being handled by the existing `MemberPathParser` unit testing, so maybe just a handful of cases? I think you'd need to have a probably preexisting type or types to resolve different paths and assert the vector of member ids is what is expected. Rough example:

```c++
struct Point {
  uint32 x;
  uint32 y;
};

struct Triangle {
  uint32 some_value;
  Point points[3];
};

DynamicType_var dt; // = Triangle
MemberPath path1;
ASSERT_RC_OK(path1.resolve_string_path(dt, "some_value"));
ASSERT_EQ(path1.ids[0], 0);
MemberPath path2;
ASSERT_RC_OK(path2.resolve_string_path(dt, "points[2].y"));
ASSERT_EQ(path2.ids[0], 1);
ASSERT_EQ(path2.ids[1], 2);
ASSERT_EQ(path2.ids[2], 1);
```

tests/unit-tests/dds/DCPS/XTypes/Utils.cpp
```diff
@@ -236,6 +236,46 @@ TEST_F(dds_DCPS_XTypes_Utils, get_keys)
...
+  }
+
+  // TODO: path with subscript is not supported. Add these when it is supported.
+  // add_type<SimpleKeyArray>();
+  // DDS::DynamicType_var dt2 = get_dynamic_type<SimpleKeyArray>();
+
+  // {
+  //   MemberPath path;
+  //   EXPECT_RC_EQ(path.resolve_string_path(dt2, "values[0].key"));
+  // }
```
iguessthislldo, line 276: Forgot about that. I would like inspect to use member path to filter parts of samples, so I wanted this eventually but was only adding it for the XTypes test which didn't need it. I would expect that failure for now.

```suggestion
  // TODO: path with subscript is not supported. Add these when it is supported.
  add_type<SimpleKeyArray>();
  DDS::DynamicType_var dt2 = get_dynamic_type<SimpleKeyArray>();

  {
    MemberPath path;
    EXPECT_RC_EQ(path.resolve_string_path(dt2, "values[0].key"), DDS::RETCODE_UNSUPPORTED);
  }
```

tests/unit-tests/dds/DCPS/XTypes/Utils.cpp
```diff
@@ -236,6 +236,46 @@ TEST_F(dds_DCPS_XTypes_Utils, get_keys)
...
+    EXPECT_EQ(path.ids[2], 0u);
+  }
+  {
+    MemberPath path;
+    EXPECT_RC_EQ(path.resolve_string_path(dt, "fmas."), DDS::RETCODE_BAD_PARAMETER);
+  }
+  {
+    MemberPath path;
+    EXPECT_NE(path.resolve_string_path(dt, "fmas.fs.invalid"), DDS::RETCODE_OK);
+  }
```
iguessthislldo, line 267: Would be nicer if it can use the exact error with `EXPECT_RC_EQ`, although I wish this method returned `DDS::RETCODE_BAD_PARAMETER` in this case.

```suggestion
  {
    MemberPath path;
    EXPECT_RC_EQ(path.resolve_string_path(dt, "fmas.fs.invalid"), DDS::RETCODE_ERROR);
  }
```

iguessthislldo (changes requested):

tests/unit-tests/dds/DCPS/XTypes/Utils.cpp
sonndinh: I didn't look into what error code can be returned by the `get_member_by_name` and `get_descriptor` calls inside this function. The only error code is `RETCODE_ERROR`, so this will work.

iguessthislldo (approved):

PR #4830, opened 2024-10-25
Updated ThreadStatusManager for iterative event handling

mitza-oci:
Some event loops process entire batches of events at once. Before this change, the batch of data would be considered an "Event" for the ThreadStatusManager.  With this change, each element within the batch counts as its own event for the ThreadStatusManager. That way the thread appears to be active and not stuck.

### DIFF
--- a/dds/DCPS/JobQueue.cpp
+++ b/dds/DCPS/JobQueue.cpp
@@ -23,7 +23,8 @@ JobQueue::JobQueue(ACE_Reactor* reactor)
 
 int JobQueue::handle_exception(ACE_HANDLE /*fd*/)
 {
-  ThreadStatusManager::Event ev(TheServiceParticipant->get_thread_status_manager());
+  ThreadStatusManager& thread_status_manager = TheServiceParticipant->get_thread_status_manager();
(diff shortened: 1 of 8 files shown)


jrw972 (approved):

dds/DCPS/JobQueue.cpp
```diff
@@ -33,6 +34,7 @@ int JobQueue::handle_exception(ACE_HANDLE /*fd*/)
   }
 
   for (Queue::const_iterator pos = q.begin(), limit = q.end(); pos != limit; ++pos) {
+    ThreadStatusManager::Event ev(thread_status_manager);
```
iguessthislldo, line 37: Windows is warning about the shadowed declaration here, should there be two of them?

mitza-oci: It is supposed to have both "events" since they support nesting.  However, now that I look at the code again in ThreadStatusManager, it needs to store the timestamp even for a nested update.
And then I'll change the variable name to fix the warning.

jrw972 (approved):

PR #4831, opened 2024-10-25
DevGuide lacks cmake example for MacOS that works by default

jrw972:
Problem
-------

The cmake example for MacOS doesn't work by default.

Solution
--------

Add an example that works.

### DIFF
--- a/docs/devguide/building/index.rst
+++ b/docs/devguide/building/index.rst
@@ -549,13 +549,25 @@ Configuring and Building
 
 To configure and build:
 
-.. tab:: Linux, macOS, BSDs, etc.
+.. tab:: Linux, BSDs, etc.
 
   .. code-block::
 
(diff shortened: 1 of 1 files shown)


mitza-oci (approved):

iguessthislldo (changes requested):
This sounds like something is wrong with CMake standard stuff. You shouldn't have to specify the standard.

iguessthislldo:
> The cmake example for MacOS doesn't work by default.

How exactly?

jrw972:
> > The cmake example for MacOS doesn't work by default.
> 
> How exactly?

I think I wrote this and stashed it before the cmake support handled the C++ versions the way it does now.
At that point, cmake would use ACE 7 which needed at least C++11 but the system default was 98 so things would not compile.

PR #4834, opened 2024-10-30
Adding Optional Annotations to IDL Test

masu7056:
[PROBLEM-REPORT-FORM.txt](https://github.com/user-attachments/files/17578932/PROBLEM-REPORT-FORM.txt)
Attached is the Problem Report Form.


jrw972:
https://github.com/OpenDDS/OpenDDS/tree/master/tests/DCPS/HelloWorld is probably a better starting point as it is simpler and more modern.  I would start by just copying and renaming to ensure that your build is correct.  Then, add the `@optional` annotation, change to the C++11 language mapping, and you should be good to go.

PR #4835, opened 2024-11-01
CLOCK_BOOTTIME timers ignore zero delay with interval

jrw972:
Problem
-------

Scheduling a timer with an interval but no delay causes the first firing of the timer to occur after the interval instead of as soon as possible.  Without CLOCK_BOOTTIME, the timer fires immediately and then starts to fire on the requested interval.

Solution
--------

Address the delay independent of the interval.

### DIFF
--- a/docs/news.d/boottime-timers.rst
+++ b/docs/news.d/boottime-timers.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4835
+
+.. news-start-section: Fixes
+- Fix periodic timers with ``--boottime`` so that the initial delay is honored.
+.. news-end-section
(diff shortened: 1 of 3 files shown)


mitza-oci:
Is there something you could add or change in `tests/unit-tests/dds/DCPS/Timers.cpp` to validate correct behavior?

mitza-oci (approved):
Add NEWS for this

docs/news.d/boottime-timers.rst
```diff
@@ -0,0 +1,5 @@
+.. news-prs: 4835
+
+.. news-start-section: Fixes
+- Fix periodic timers with `--boottime` so that the initial delay is honored.
```
iguessthislldo, line 4:
```suggestion
- Fix periodic timers with ``--boottime`` so that the initial delay is honored.
```

PR #4837, opened 2024-11-04
MacOS 12 is deprecated in GHA

jrw972:
Problem
-------

The MacOS 12 runners are going away in GHA.

Solution
--------

Upgrade the MacOS 12 jobs to use MacOS 14.


mitza-oci (approved):

.github/workflows/build_and_test.yml
```diff
@@ -13267,7 +13267,7 @@ jobs:
         <command name="check_compiler" options="gcc"/>
         <command name="print_perl_version"/>
         <command name="print_autobuild_config"/>
-        <command name="auto_run_tests" options="script_path=tests dir=$GITHUB_WORKSPACE/OpenDDS -Config STATIC_MESSENGER -Config STATIC_COMPILER -Config STATIC_COMPILER2 -Config STATIC_UNIT_TESTS -Config XERCES3 -Config OPENDDS_SECURITY -Config CXX11 -Config RAPIDJSON -Config NO_SHMEM -Config DDS_NO_ORBSVCS -Config GH_ACTIONS_M12 --no-dcps tests/core_ci_tests.lst"/>
```
sonndinh, line 13270: The `GH_ACTIONS_M12` flag doesn't seem to exist in the *.lst files. The most similar option I found is `GH_ACTIONS_M10`. I think it was used to block some tests that failed on MacOS for some reason that we hadn't found out. The GitHub Actions document (https://github.com/OpenDDS/OpenDDS/blob/master/docs/internal/github_actions.rst) mentions it. I think the document needs to be updated at some point for that part.

Issue #4842, opened 2024-11-11
Visual Studio 2015 failed with cmake


hyh19962008:
Building OpenDDS 3.28.1 with Visual Studio 2015 x64 and CMake 3.22, I got this error:
```
CMake Error at cmake/build_ace_tao.cmake:155 (message):
  Not Sure how to build ACE/TAO for this system (Windows/Visual Studio 14
  2015/vc14)
Call Stack (most recent call first):
  CMakeLists.txt:47 (include)
```

There is this check in cmake/build_ace_tao.cmake:
`elseif(_OPENDDS_MPC_TYPE MATCHES "^vs")`
But in cmake/get_ace_tao.cmake, "Visual Studio 14 2015" was set to "vc14".
So I modified it to pass the check:
`elseif(_OPENDDS_MPC_TYPE MATCHES "^vs" OR _OPENDDS_MPC_TYPE MATCHES "^vc" )`

hyh19962008:
Alright, I created PR #4843 .

PR #4843, opened 2024-11-12
fix VisualStudio 2015 cmake check error

hyh19962008:
#4842


iguessthislldo (changes requested):
Looks good to me. There's a failing Windows CMake build in CI, but it's from [this issue](https://github.com/DOCGroup/ACE_TAO/pull/1357#issuecomment-2161417666) and is unrelated to the change here. The only thing that needs to be done is to add news for it. Can you add as a new file called something like `docs/news.d/cmake-2015.rst`? (The name isn't really important, only that it's in that directory.)


```rst
.. news-prs: 4843
.. news-start-section: Platform Support and Dependencies
.. news-start-section: CMake
- Fixed an issue :ghissue:`where OpenDDS won't build ACE/TAO with Visual Studio 2015 <4842>`.
.. news-end-section
.. news-end-section
```

hyh19962008:
OK, it's done.

iguessthislldo (approved):
Alright, thanks!

PR #4864, opened 2024-12-09
Fixed some warnings

mitza-oci:
Also improved configure script's use of warning settings on VC++. This avoids "warning D9025: overriding '/W3' with '/W4'"

### DIFF
--- a/dds/DCPS/RTPS/ICE/Stun.cpp
+++ b/dds/DCPS/RTPS/ICE/Stun.cpp
@@ -283,7 +283,7 @@ bool operator>>(DCPS::Serializer& serializer, AttributeHolder& holder)
       return false;
     }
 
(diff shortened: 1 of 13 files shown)


jrw972 (approved):

dds/DCPS/RTPS/ICE/Stun.cpp
```diff
@@ -283,7 +283,7 @@ bool operator>>(DCPS::Serializer& serializer, AttributeHolder& holder)
       return false;
     }
 
-    ACE_UINT16 code = class_ * 100 + num;
+    ACE_UINT16 code = static_cast<ACE_UINT16>(class_ * 100 + num);
```
sonndinh, line 286: Looks like `code` can be a `const`?

PR #4867, opened 2024-12-11
opendds_idl: error for struct member with unsupported type

### LINKED ISSUE #4866: Belongs to type seems missing in warning message
I have this in IDL

```
typedef string<128> color_type;
typedef map<string,long> StringLongMap;

@nested(FALSE)
struct ShapeType {
  color_type color; //@key
  long x;
  long y;
  long shapesize;
  StringLongMap slm;
};
```

Compiled with `bin/opendds_idl -DOPENDDS_NO_MULTI_TOPIC -Sa -St -Wb,opendds_sequence_suffix=OpenDDSSeq --unknown-annotations ignore -o GeneratedCode -Wb,export_macro=SHAPES_TYPESUPPORT_Export -Wb,export_include=shapes_typesupport_export.h GeneratedCode/shapetype_dds.idl`

Gives this warning
```
WARNING: field slm can not be serialized.  The struct or union it belongs to () can not be used in an OpenDDS topic type.
```

The type is belongs to `()` seems to be missing. Would recommend to also add quotes around `slm`

mitza-oci:
These are cases with valid IDL syntax but types that aren't (yet) supported in DDS topic structs, such as maps.

Resolves #4866


jwillemsen:
When it works ok for me, hopefully at some point https://github.com/OpenDDS/OpenDDS/pull/3603 will be finished

dds/idl/typeobject_generator.cpp
```diff
@@ -1732,8 +1732,7 @@ typeobject_generator::get_minimal_type_identifier(AST_Type* type)
   }
 
   HashTypeIdentifierMap::const_iterator pos = hash_type_identifier_map_.find(type);
-  OPENDDS_ASSERT(pos != hash_type_identifier_map_.end());
-  return pos->second.minimal;
+  return pos == hash_type_identifier_map_.end() ? OpenDDS::XTypes::TypeIdentifier() : pos->second.minimal;
```
jrw972, line 1735: Is there a concern that callers expect a non-default TypeIdentifier?

mitza-oci: When run with an unsupported type it hit the assert.  It's probably worth a look to see if there would be any issues with supported types.

mitza-oci: I reviewed the calling code and didn't see any issues.  It may not generate useful typeobjects but those are cases where opendds_idl exits with an error anyway.

jrw972 (approved):

PR #4871, opened 2024-12-20
BOOTTIME timers use too many file descriptors

jrw972:
Problem
-------

OpenDDS applications use a lot of timers.  The BOOTTIME timer support allocates a file descriptor for each timer.  When the number of timers exceeds the limit for `select` system calls, then things stop working correctly.

Solution
--------

Allocate a single file descriptor and use a timer heap.


dds/DCPS/ReactorTask.cpp
```diff
@@ -121,6 +122,11 @@ int ReactorTask::open_reactor_task(ThreadStatusManager* thread_status_manager,
     reactor_->timer_queue(timer_queue_);
   }
 
+  if (!reactor_wrapper_.open(reactor_) && log_level >= LogLevel::Error) {
+    ACE_ERROR((LM_ERROR, "(%P|%t) NOTICE: ReactorTask::open_reactor_task: could not open ReactorWrapper\n"));
+    return -1;
+  }
```
iguessthislldo, line 128:
```suggestion
  if (!reactor_wrapper_.open(reactor_)) {
    if (log_level >= LogLevel::Error) {
      ACE_ERROR((LM_ERROR, "(%P|%t) NOTICE: ReactorTask::open_reactor_task: could not open ReactorWrapper\n"));
    }
    return -1;
  }
```

dds/DCPS/ReactorTask.cpp
```diff
@@ -480,6 +462,42 @@ int ReactorWrapper::remove_handler(ACE_Event_Handler* event_handler,
...
+  ts.it_interval = ACE_Time_Value::zero;
+  // expiration in the past, execute as soon as possible
+  // avoid zeros since that would disarm the timer
+  ts.it_value = (delay <= ACE_Time_Value::zero) ? one_ns : delay;
+  if (timerfd_settime(timer_fd_, 0, &ts, 0) == -1) {
+    if (log_level >= LogLevel::Notice) {
+      ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: ReactorWrapper::arm: timerfd_settime %m\n"));
+    }
+  }
+}
```
iguessthislldo, line 498: This should probably be a working because it's not returning any error status.

iguessthislldo (approved):

PR #4884, opened 2025-02-07
Fix CMake iShapes and Comply with CMP0165

iguessthislldo:
Fixes https://github.com/OpenDDS/OpenDDS/issues/4849

Changes from https://github.com/OpenDDS/OpenDDS/pull/4487 require C++ language to be enabled in CMake, which it basically is all the time through `project(... LANGUAGES CXX)` expect in ishapes. There `find_project(OpenDDS)` was used to get the OpenDDS version before `project` enabled C++ which results in `try_compile` failing fatally. These changes both make sure C and C++ are enabled early in CMake initialization and switch the `project` and `find_package` to comply with https://cmake.org/cmake/help/latest/policy/CMP0165.html#policy:CMP0165

Also changed the C++ standard checks to include the min C++ standard (currently C++98/03), so it would never mistake a broken compiler setup with a compiler that only supports the min standard.

### DIFF
--- a/cmake/ace_group.cmake
+++ b/cmake/ace_group.cmake
@@ -67,7 +67,6 @@ if(MSVC AND OPENDDS_STATIC)
   _opendds_vs_force_static()
 endif()
 
-enable_language(C)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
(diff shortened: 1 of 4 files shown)


iguessthislldo:
Need to add news and adding a ishapes CMake build to CI would probably be a good idea.

jrw972 (approved):

mitza-oci (approved):

PR #4901, opened 2025-02-18
iShapes demo doesn't have complete type objects

jrw972:
Problem
-------

The iShapes demo doesn't have complete type objects which is something other vendors expect.

Solution
--------

Add complete type objects.


mitza-oci (approved):

iguessthislldo (changes requested):
Should also be added to the CMake file: https://github.com/OpenDDS/OpenDDS/blob/a22596a24cf9c5183c17729ebbfa27805ae5544c/examples/DCPS/ishapes/CMakeLists.txt#L48

```cmake
opendds_target_sources(ishapes ShapeType.idl OPENDDS_IDL_OPTIONS -Gxtypes-complete)
```

iguessthislldo (approved):

PR #4908, opened 2025-02-24
RtpsDiscovery thread can be unresponsive during cleanup

jrw972:
Problem
-------

The leases for remote participants can expire within a short amount of time.  The current cleanup logic processes all participants that have expired up until the most recent sampling of now.  If participants remain, the cleanup task is rescheduled by comparing the first participant to expire against the most recent sample of now.  This delay can be negative due to the time needed for cleanup and the sampling of now.  Consequently, the cleanup task can preempt other tasks from running, such as those for thread monitoring.

Solution
--------

Process one expired remote participant at a time and enforce at least a 1ms delay between cleanup attempts.


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3921,41 +3921,40 @@ Spdp::process_lease_expirations(const DCPS::MonotonicTimePoint& now)
...
   ACE_GUARD (ACE_Thread_Mutex, g, lock_);
 
-  for (TimeQueue::iterator pos = lease_expirations_.begin(), limit = lease_expirations_.end();
-       pos != limit && pos->first <= now;) {
+  // Process one at a time to keep the thread responsive.
+  TimeQueue::iterator pos = lease_expirations_.begin();
+  if (pos != lease_expirations_.end() && pos->first <= now) {
     DiscoveredParticipantIter part = participants_.find(pos->second);
     // Pre-emptively erase so purge_discovered_participant will not modify lease_expirations_.
     lease_expirations_.erase(pos++);
```
mitza-oci, line 3929:
```suggestion
  if (pos != lease_expirations_.end() && pos->first <= now) {
    DiscoveredParticipantIter part = participants_.find(pos->second);
    // Pre-emptively erase so purge_discovered_participant will not modify lease_expirations_.
    lease_expirations_.erase(pos);
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3921,41 +3921,40 @@ Spdp::process_lease_expirations(const DCPS::MonotonicTimePoint& now)
 {
   ACE_GUARD (ACE_Thread_Mutex, g, lock_);
 
-  for (TimeQueue::iterator pos = lease_expirations_.begin(), limit = lease_expirations_.end();
-       pos != limit && pos->first <= now;) {
+  // Process one at a time to keep the thread responsive.
+  TimeQueue::iterator pos = lease_expirations_.begin();
```
mitza-oci, line 3925:
```suggestion
  const TimeQueue::iterator pos = lease_expirations_.begin();
```

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3921,41 +3921,40 @@ Spdp::process_lease_expirations(const DCPS::MonotonicTimePoint& now)
...
-    purge_discovered_participant(part);
-    participants_.erase(part);
+      purge_discovered_participant(part);
+      participants_.erase(part);
+    }
   }
 
   if (!lease_expirations_.empty()) {
-    tport_->lease_expiration_task_->schedule(lease_expirations_.begin()->first - now);
+    const TimeDuration delay = std::max(lease_expirations_.begin()->first - now, TimeDuration::from_msec(1));
```
mitza-oci, line 3956: Wouldn't there be cases where a shorter min delay constant would be desired?  If there are many expirations, and they can be processed quickly, requiring 1ms can keep more data in memory than needed.

Also how about resampling the clock instead of using the passed-in `now`?  When determining the time for the next event, the passed-in time may not be that helpful.

mitza-oci (changes requested):

dds/DCPS/SporadicTask.cpp
```diff
@@ -14,13 +14,22 @@ namespace DCPS {
 
 void SporadicTask::schedule(const TimeDuration& delay)
 {
-  const MonotonicTimePoint next_time = time_source_.monotonic_time_point_now() + delay;
+  schedule_i(time_source_.monotonic_time_point_now() + delay);
+}
+
+void SporadicTask::schedule_max(const MonotonicTimePoint& release,
+                                const TimeDuration& delay)
```
mitza-oci, line 21: Rename parameters here?

docs/devguide/run_time_configuration.rst
```diff
@@ -1137,6 +1137,21 @@ Those properties, along with options specific to OpenDDS's RTPS discovery implem
...
+
+  .. prop:: MinimumCleanupSeparation=<msec>
+    :default: ``1`` (1 millisecond)
+
+    Ensure that attempts to clean up expired participants are separated by this amount of time.
+
+  .. prop:: SecurityUnsecureLeaseDuration=<sec>
+    :default: ``30`` (30 seconds)
+
+    Limit the lease duration requested by a discovred participant when bootstrapping discovery from an SPDP message.
```
mitza-oci, line 1153:
```suggestion
    Limit the lease duration requested by a discovered participant when bootstrapping discovery from an SPDP message.
```

docs/news.d/rtps-cleanup-separation.rst
```diff
@@ -0,0 +1,8 @@
+.. news-prs: 4908
+
+.. news-start-section: Fixes
+- Introduced a time separation when processed expired discoverd participants with RtpsDiscovery.
```
mitza-oci, line 4:
```suggestion
- Introduced a time separation when processed expired discovered participants with RtpsDiscovery.
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #4928, opened 2025-02-26
RtpsRelay output queue may be ill-behaved

jrw972:
Problem
-------

The RtpsRelay uses an output queue.  When a message needs to be sent, the message is placed on a queue and a handler is registered with the reactor if necessary.  When the handler is invoked, a message is removed from the queue and sent. This avoid errors from the send buffer being full since the socket is non-blocking.

This design has shown to have undesirable behavior which stems from a lack of fairness between input and output processing.  That is, given that both tasks are ready, there is no guarantee that one will not be systematically preferred over another.  Given a high enough input rate, the output task will fall behind and the queue will grow.

Solution
--------

Introduce a `-SynchronousOutput` option that sends messages as soon as possible instead of queueing them and waiting on the reactor.  In this configuration, it is possible that the output socket will not be ready and return an error indicating that a write would block.

### DIFF
--- a/docs/news.d/rtpsrelay-synchronous-output.rst
+++ b/docs/news.d/rtpsrelay-synchronous-output.rst
@@ -1,0 +1,5 @@
+.. news-prs: 4928
+
+.. news-start-section: Additions
+- Added :option:`RtpsRelay -SynchronousOutput` option to RtpsRelay.
+.. news-end-section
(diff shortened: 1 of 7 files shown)


docs/news.d/rtpsrelay-synchronous-output.rst
```diff
@@ -0,0 +1,5 @@
+.. news-prs: 4928
+
+.. news-start-section: Additions
+- Added ``-SynchronousOutput`` option to RtpsRelay.
```
iguessthislldo, line 4:
```suggestion
- Added :option:`RtpsRelay -SynchronousOutput` option.
```

mitza-oci (commented):
Consider renaming `RelayHandler::enqueue_message` since it doesn't always enqueue.

mitza-oci (approved):

iguessthislldo (approved):

PR #4937, opened 2025-03-04
[v4] Remove udp transport

jrw972:


iguessthislldo:
These removals will need news items, are we doing them later?

jrw972:
> These removals will need news items, are we doing them later?

I think we will do those later.

iguessthislldo (approved):

PR #4938, opened 2025-03-05
Add Class Dependencies For Strongly Connected Components

simpsont-oci:
Problem:
 - Not all StronglyConnectedComponent types in the same class are being added to list of type dependencies, preventing the TypeLookupService from sending enough type information for complex types.
 - This can be seen, for example, when looking at the opendds-monitor complex topic type instances (which currently cause the monitor to crash)

Solution:
 - Add all the StronglyConnectedComponent class typeids when one is encountered during dependency calculation


jrw972:
Replaced by #4940

PR #4940, opened 2025-03-06
TypeLookupService doesn't support SCCs

jrw972:
Problem
-------

The TypeLookupService doesn't support strongly connected components (SCCs) correctly.

Solution
--------

* Add the SCCId to dependencies when necessary (fixes `getTypeDependencies`).
* Update `getTypes` to consolidate type identifiers and send all components of an SCC.

### DIFF
--- a/dds/DCPS/XTypes/TypeLookupService.cpp
+++ b/dds/DCPS/XTypes/TypeLookupService.cpp
@@ -57,11 +57,31 @@ TypeLookupService::~TypeLookupService()
 void TypeLookupService::get_type_objects(const TypeIdentifierSeq& type_ids,
                                          TypeIdentifierTypeObjectPairSeq& types) const
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  TypeIdentifierSet tids;
+
   for (unsigned i = 0; i < type_ids.length(); ++i) {
-    TypeMap::const_iterator pos = type_map_.find(type_ids[i]);
-    if (pos != type_map_.end()) {
-      types.append(TypeIdentifierTypeObjectPair(pos->first, pos->second));
+    const TypeIdentifier& tid = type_ids[i];
+    if (tid.kind() != TI_STRONGLY_CONNECTED_COMPONENT) {
+      tids.insert(tid);
+    } else {
+      TypeIdentifier scc_id(tid);
+      for (ACE_CDR::Long j = 1; j <= scc_id.sc_component_id().scc_length; ++j) {
+        scc_id.sc_component_id().scc_index = j;
+        tids.insert(scc_id);
+      }
+    }
+  }
+
+  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  for (TypeIdentifierSet::const_iterator pos = tids.begin(), limit = tids.end(); pos != limit; ++pos) {
+    const TypeIdentifier& tid = *pos;
+    TypeMap::const_iterator pos2 = type_map_.find(tid);
+    if (pos2 != type_map_.end()) {
+      types.append(TypeIdentifierTypeObjectPair(tid, pos2->second));
+    } else if (DCPS::log_level >= DCPS::LogLevel::Warning) {
+      ACE_ERROR((LM_WARNING,
+                 "(%P|%t) WARNING: TypeLookupService::set_type_objects: "
+                 "unknown type\n"));
     }
   }
 }
@@ -103,25 +123,15 @@ void TypeLookupService::get_type_dependencies(const TypeIdentifierSeq& type_ids,
 void TypeLookupService::get_type_dependencies_i(const TypeIdentifierSeq& type_ids,
   TypeIdentifierWithSizeSeq& dependencies) const
 {
-  OPENDDS_SET(TypeIdentifier) tmp;
   for (unsigned i = 0; i < type_ids.length(); ++i) {
-    const TypeIdentifierWithSizeSeqMap::const_iterator it = type_dependencies_map_.find(type_ids[i]);
+    const TypeIdentifier tid = make_scc_id_or_default(type_ids[i]);
+    const TypeIdentifierWithSizeSeqMap::const_iterator it = type_dependencies_map_.find(tid);
     if (it != type_dependencies_map_.end()) {
       for (unsigned j = 0; j < it->second.length(); ++j) {
-        tmp.insert(it->second[j].type_id);
+        const ACE_CDR::ULong idx = dependencies.length();
+        dependencies.length(idx + 1);
+        dependencies[idx] = it->second[j];
       }
-    }
-  }
-
-  // All dependent TypeIdentifiers are expected to have an entry in the TypeObject cache.
-  dependencies.length(static_cast<unsigned>(tmp.size()));
-  OPENDDS_SET(TypeIdentifier)::const_iterator iter = tmp.begin();
-  for (unsigned i = 0; iter != tmp.end(); ++i, ++iter) {
-    const TypeMap::const_iterator tobj_it = type_map_.find(*iter);
-    if (tobj_it != type_map_.end()) {
-      dependencies[i].type_id = *iter;
-      const size_t sz = DCPS::serialized_size(get_typeobject_encoding(), tobj_it->second);
-      dependencies[i].typeobject_serialized_size = static_cast<unsigned>(sz);
     }
   }
 }
--- a/dds/DCPS/XTypes/TypeObject.cpp
+++ b/dds/DCPS/XTypes/TypeObject.cpp
@@ -466,49 +466,64 @@ bool has_type_object(const TypeIdentifier& ti)
     ti.kind() != TK_NONE;
 }
 
+TypeIdentifier make_scc_id_or_default(const TypeIdentifier& tid)
+{
+  if (tid.kind() != TI_STRONGLY_CONNECTED_COMPONENT) {
+    return tid;
+  }
+
+  TypeIdentifier scc_id(tid);
+  scc_id.sc_component_id().scc_index = 0;
+  return scc_id;
+}
+
 const TypeMap TypeMapBuilder::EmptyMap;
 
 namespace {
   void compute_dependencies_i(const TypeMap& type_map,
                               const TypeIdentifier& type_identifier,
-                              OPENDDS_SET(TypeIdentifier)& dependencies);
+                              TypeIdentifierSet& dependencies);
+
+  void compute_dependencies_j(const TypeMap& type_map,
+                              const TypeIdentifier& type_identifier,
+                              TypeIdentifierSet& dependencies);
 
   template <typename T>
   void compute_dependencies_i(const TypeMap& type_map,
                               const Sequence<T>& type,
-                              OPENDDS_SET(TypeIdentifier)& dependencies);
+                              TypeIdentifierSet& dependencies);
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const CommonAliasBody& type,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, type.related_type, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const MinimalAliasBody& type,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, type.common, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const MinimalAliasType& type,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, type.body, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const AppliedAnnotation& ann,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, ann.annotation_typeid, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const OPENDDS_OPTIONAL_NS::optional<AppliedAnnotationSeq>& ann_seq,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     if (ann_seq) {
       compute_dependencies_i(type_map, ann_seq.value(), dependencies);
@@ -517,21 +532,21 @@ namespace {
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const CompleteTypeDetail& type_detail,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, type_detail.ann_custom, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const CompleteAliasHeader& header,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, header.detail, dependencies);
   }
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const CompleteAliasBody& body,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, body.common, dependencies);
     compute_dependencies_i(type_map, body.ann_custom, dependencies);
@@ -539,7 +554,7 @@ namespace {
 
   void compute_dependencies_i(const TypeMap& type_map,
                               const CompleteAliasType& type,
-                              OPENDDS_SET(TypeIdentifier)& dependencies)
+                              TypeIdentifierSet& dependencies)
   {
     compute_dependencies_i(type_map, type.header, dependencies);
     compute_dependencies_i(type_map, type.body, dependencies);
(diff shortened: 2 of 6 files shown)


jrw972:
See https://github.com/OpenDDS/OpenDDS/pull/4938

simpsont-oci:
I haven't had a time to do a full review, but I've built this branch locally and tested it against the monitor & (unmanaged) monitor test app and unfortunately the monitor still crashes. I've got a bunch of debugging changes stashed somewhere, I'll try to salvage those and see if I can figure out what's not getting requested / sent (assuming it's even crashing in the same spot).

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -471,616 +471,719 @@ const TypeMap TypeMapBuilder::EmptyMap;
...
     for (typename Sequence<T>::const_iterator pos = type.begin(), limit = type.end(); pos != limit; ++pos) {
-      compute_dependencies_i(type_map, *pos, dependencies);
+      compute_dependencies_i(type_map, *pos, dependencies, sccs);
+    }
+  }
+
+  TypeIdentifier make_scc_id(const TypeIdentifier& tid)
+  {
+    if (tid.kind() != TI_STRONGLY_CONNECTED_COMPONENT) {
+      return tid;
```
sonndinh, line 1093: Should this return an error instead? It is returning the input type which isn't an SCC which can be misleading given the function name `make_scc_id`.

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -471,616 +471,719 @@ const TypeMap TypeMapBuilder::EmptyMap;
...
+      compute_dependencies_i(type_map, type_identifier.map_sdefn(), dependencies, sccs);
+      break;
+    case TI_PLAIN_MAP_LARGE:
+      compute_dependencies_i(type_map, type_identifier.map_ldefn(), dependencies, sccs);
+      break;
+    case TI_STRONGLY_CONNECTED_COMPONENT:
+      {
+        // An SCCId (identifies the entire component.)
+        TypeIdentifier scc_id = make_scc_id(type_identifier);
+        sccs.insert(scc_id);
```
sonndinh, line 1156: The returned `scc_id` has `scc_index` value of 0. I think the SCC index starts from 1?

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -57,11 +57,27 @@ TypeLookupService::~TypeLookupService()
...
+      }
+    }
+  }
+
+  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  for (TypeIdentifierSet::const_iterator pos = tids.begin(), limit = tids.end(); pos != limit; ++pos) {
+    const TypeIdentifier& tid = *pos;
+    TypeMap::const_iterator pos2 = type_map_.find(tid);
+    if (pos2 != type_map_.end()) {
+      types.append(TypeIdentifierTypeObjectPair(tid, pos2->second));
```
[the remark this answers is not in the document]
jrw972: It's possible but I don't think that's the problem based on what @simpsont-oci has said.

dds/DCPS/XTypes/TypeObject.cpp
jrw972: The function does exactly what I want it to do, return the SCCId if it's a component of an SCC or the given type identifier if not.  Would `make_scc_id_or_default` be better?

sonndinh: Yeah, I think so.

simpsont-oci:
> I haven't had a time to do a full review, but I've built this branch locally and tested it against the monitor & (unmanaged) monitor test app and unfortunately the monitor still crashes. I've got a bunch of debugging changes stashed somewhere, I'll try to salvage those and see if I can figure out what's not getting requested / sent (assuming it's even crashing in the same spot).

I was able to restore my debugging output... it looks like what's happening is that the SCC type ids are never getting sent or requested. I see an error message on the testapp in `populate_dependencies_i` which makes it look like the zero-index SCC type identifiers can't be found in type_map and thus can't be serialized (can't create TypeIdentifierWithSize), and I think that's generally true for our implementation, only the individual SCC components get added to the type_map, but maybe that needs to change?

jrw972:
I added a test for an SCC topic type.
For the IDL I'm using, it should have 5 types in the SCC and it does.
They are indexed from 1 to 5 and they are inserted into the type map.

`getTypeDependencies` should return the index 0 SCCId (see below) when queried with any non-index SCCId or the index 0 SCCId.
I updated the calling code in OpenDDS to the index 0 SCCId based on the spec.
`getTypes` should return type objects for all of the components of an SCCId.

To me, the spec is vague about returning the type in its dependency list.
If you query for the dependencies of X should you
1. Always get back X
2. Never get back X
3. Only get back the index 0 SCCId of X if X is part of an SCC

Right now, we are doing 3.

simpsont-oci:
Poked around with the latest changes a bit more, and it looks like what's happening for the monitor is this:
 - Complex topic type is not an SCC, but depends on one.
 - Request for that topic's dependencies seems to correctly compute the the dependency, but when it goes calculate the response size in `TypeSupportImpl::populate_dependencies_i`, the the 0-index SCC doesn't exist in the type map so we get an error message and the and the SCC dependency gets omitted from the response.
 - If we make a special check for 0-index SCCs in `TypeSupportImpl::populate_dependencies_i`, then a similar thing happens in `TypeLookupService::get_type_dependencies_i` when it tries to actually serialize the response.
 - If we also make a special check for 0-index SCCs in `TypeLookupService::get_type_dependencies_i`, it then looks like the monitor is able to receive the 0-index SCC and then make a request for it, getting all the individual components and not crash because of a null descriptor when trying to subscribe to the complex topic.

Even with those changes, however, I also see a bunch of these warnings, which I don't remember seeing before, but I'm not sure if these changes could have caused them:
```
(1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
(1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
(1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
(1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
```

simpsont-oci:
> Even with those changes, however, I also see a bunch of these warnings, which I don't remember seeing before, but I'm not sure if these changes could have caused them:

Just double checked against master, and these warnings/errors are there too, just hidden in the log window. The mapping seems odd in the monitor (TK_UINT8 becomes tk_ushort?) so it's likely a bug there.

jrw972:
> Poked around with the latest changes a bit more, and it looks like what's happening for the monitor is this:
> 
> * Complex topic type is not an SCC, but depends on one.
> * Request for that topic's dependencies seems to correctly compute the the dependency, but when it goes calculate the response size in `TypeSupportImpl::populate_dependencies_i`, the the 0-index SCC doesn't exist in the type map so we get an error message and the and the SCC dependency gets omitted from the response.
> * If we make a special check for 0-index SCCs in `TypeSupportImpl::populate_dependencies_i`, then a similar thing happens in `TypeLookupService::get_type_dependencies_i` when it tries to actually serialize the response.
> * If we also make a special check for 0-index SCCs in `TypeLookupService::get_type_dependencies_i`, it then looks like the monitor is able to receive the 0-index SCC and then make a request for it, getting all the individual components and not crash because of a null descriptor when trying to subscribe to the complex topic.
> 
> Even with those changes, however, I also see a bunch of these warnings, which I don't remember seeing before, but I'm not sure if these changes could have caused them:
> 
> ```
> (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> ```

Excellent detective work.

So, it seems like the missing piece is the size of SCCId(0).
And the spec is not going to help us.
I see three possible options:
1. Let the size be 0.
2. The sum of the serialized sizes of the components.
3. The serialized size of a sequence of type objects of the components.  (This is probably the one implied by the spec.)

simpsont-oci:
> > Poked around with the latest changes a bit more, and it looks like what's happening for the monitor is this:
> > 
> > * Complex topic type is not an SCC, but depends on one.
> > * Request for that topic's dependencies seems to correctly compute the the dependency, but when it goes calculate the response size in `TypeSupportImpl::populate_dependencies_i`, the the 0-index SCC doesn't exist in the type map so we get an error message and the and the SCC dependency gets omitted from the response.
> > * If we make a special check for 0-index SCCs in `TypeSupportImpl::populate_dependencies_i`, then a similar thing happens in `TypeLookupService::get_type_dependencies_i` when it tries to actually serialize the response.
> > * If we also make a special check for 0-index SCCs in `TypeLookupService::get_type_dependencies_i`, it then looks like the monitor is able to receive the 0-index SCC and then make a request for it, getting all the individual components and not crash because of a null descriptor when trying to subscribe to the complex topic.
> > 
> > Even with those changes, however, I also see a bunch of these warnings, which I don't remember seeing before, but I'm not sure if these changes could have caused them:
> > ```
> > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> > ```
> 
> Excellent detective work.
> 
> So, it seems like the missing piece is the size of SCCId(0). And the spec is not going to help us. I see three possible options:
> 
>     1. Let the size be 0.
> 
>     2. The sum of the serialized sizes of the components.
> 
>     3. The serialized size of a sequence of type objects of the components.  (This is probably the one implied by the spec.)
> 
> 
> https://issues.omg.org/browse/DDSXTY14-87

I'm not sure I'm following... wouldn't it just be the size of the 0-index SCC TypeIdentifier?

I take it from the 2nd-to-last paragraph of section 7.6.3.2.1 that it's fine to just return the 0-index SCC TypeIdentifier in the list of dependencies for the queried type/s, and then from that single 0-index SCC TypeIdentifier, the interested participant can follow up with a type request containing all of the associated component SCC TypeIdentifiers, at which point it will get the actual TypeObjects for them.

simpsont-oci:
For example, the code I added to `populate_dependencies_i` looks like this (ignore the DEBUG statements):
```
   for (; it != dependencies.end(); ++it) {
-    XTypes::TypeMap::const_iterator iter = type_map.find(*it);
-    if (iter != type_map.end()) {
-      const size_t tobj_size = serialized_size(XTypes::get_typeobject_encoding(), iter->second);
+    ACE_DEBUG((LM_DEBUG, ACE_TEXT("TypeSupportImpl::populate_dependencies_i - found dependency %C\n"), it->to_string().c_str()));
+    if (it->kind() == XTypes::TI_STRONGLY_CONNECTED_COMPONENT && it->sc_component_id().scc_index == 0) {
+      const size_t tobj_size = serialized_size(XTypes::get_typeobject_encoding(), *it);
       XTypes::TypeIdentifierWithSize ti_with_size(*it, static_cast<ACE_CDR::ULong>(tobj_size));
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("TypeSupportImpl::populate_dependencies_i - found SCC dependency with calculated size of %d\n"), tobj_size));
       deps_with_size.append(ti_with_size);
-    } else if (XTypes::has_type_object(*it)) {
-      log_ti_not_found("populate_dependencies_i", name(), *it);
+    } else {
+      XTypes::TypeMap::const_iterator iter = type_map.find(*it);
+      if (iter != type_map.end()) {
+        const size_t tobj_size = serialized_size(XTypes::get_typeobject_encoding(), iter->second);
+        XTypes::TypeIdentifierWithSize ti_with_size(*it, static_cast<ACE_CDR::ULong>(tobj_size));
+        deps_with_size.append(ti_with_size);
+      } else if (XTypes::has_type_object(*it)) {
+        log_ti_not_found("populate_dependencies_i", name(), *it);
+      }
     }
```

and the code I added to `get_type_dependencies_i` looks like this (again, ignore DEBUG stuff):
```
   for (unsigned i = 0; iter != tmp.end(); ++i, ++iter) {
-    const TypeMap::const_iterator tobj_it = type_map_.find(*iter);
-    if (tobj_it != type_map_.end()) {
+    if (iter->kind() == TI_STRONGLY_CONNECTED_COMPONENT && iter->sc_component_id().scc_index == 0) {
       dependencies[i].type_id = *iter;
-      const size_t sz = DCPS::serialized_size(get_typeobject_encoding(), tobj_it->second);
+      const size_t sz = serialized_size(get_typeobject_encoding(), *iter);
+      ACE_DEBUG((LM_DEBUG, ACE_TEXT("TypeLookupService::get_type_dependencies_i - found SCC dependency with calculated size of %d\n"), sz));
       dependencies[i].typeobject_serialized_size = static_cast<unsigned>(sz);
+    } else { 
+      const TypeMap::const_iterator tobj_it = type_map_.find(*iter);
+      if (tobj_it != type_map_.end()) {
+        dependencies[i].type_id = *iter;
+        const size_t sz = DCPS::serialized_size(get_typeobject_encoding(), tobj_it->second);
+        dependencies[i].typeobject_serialized_size = static_cast<unsigned>(sz);
+      }
     }
   }
```

jrw972:
> > > Poked around with the latest changes a bit more, and it looks like what's happening for the monitor is this:
> > > 
> > > * Complex topic type is not an SCC, but depends on one.
> > > * Request for that topic's dependencies seems to correctly compute the the dependency, but when it goes calculate the response size in `TypeSupportImpl::populate_dependencies_i`, the the 0-index SCC doesn't exist in the type map so we get an error message and the and the SCC dependency gets omitted from the response.
> > > * If we make a special check for 0-index SCCs in `TypeSupportImpl::populate_dependencies_i`, then a similar thing happens in `TypeLookupService::get_type_dependencies_i` when it tries to actually serialize the response.
> > > * If we also make a special check for 0-index SCCs in `TypeLookupService::get_type_dependencies_i`, it then looks like the monitor is able to receive the 0-index SCC and then make a request for it, getting all the individual components and not crash because of a null descriptor when trying to subscribe to the complex topic.
> > > 
> > > Even with those changes, however, I also see a bunch of these warnings, which I don't remember seeing before, but I'm not sure if these changes could have caused them:
> > > ```
> > > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> > > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> > > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.u8 id 0 kind uint8 (uint8) as an invalid kind uint16
> > > (1749885|1749885) NOTICE: DynamicDataImpl::get_value_from_struct: trying to get test::BasicTypes.i8 id 1 kind int8 (int8) as an invalid kind int16
> > > ```
> > 
> > 
> > Excellent detective work.
> > So, it seems like the missing piece is the size of SCCId(0). And the spec is not going to help us. I see three possible options:
> > ```
> > 1. Let the size be 0.
> > 
> > 2. The sum of the serialized sizes of the components.
> > 
> > 3. The serialized size of a sequence of type objects of the components.  (This is probably the one implied by the spec.)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > https://issues.omg.org/browse/DDSXTY14-87
> 
> I'm not sure I'm following... wouldn't it just be the size of the 0-index SCC TypeIdentifier?
> 
> I take it from the 2nd-to-last paragraph of section 7.6.3.2.1 that it's fine to just return the 0-index SCC TypeIdentifier in the list of dependencies for the queried type/s, and then from that single 0-index SCC TypeIdentifier, the interested participant can follow up with a type request containing all of the associated component SCC TypeIdentifiers, at which point it will get the actual TypeObjects for them.

There is no TypeObject corresponding to the SCCIdentifier.  The closest there is the sequence of component type objects from the hashing algorithm.
7.6.3.2.1 and elsewhere says you can only use the SCCIdentifier, except when returning `getTypes` at which point you have to use the identifiers for the individual components.

To summarize:

`getTypeDependencies`
* in: TypeIdentifier (minimal/complete) or SCCIdentifier (index 0).
* out: TypeIdentifierWithSize (same as above) <-- This is where the spec lacks clarity.

`getTypes`
* in: TypeIdentifier (minimal/complete) or SCCIdentifier (index 0).
* out: TypeIdentifier and TypeObject (SCCComponentIds must have index != 0 and all must be present)

simpsont-oci:
> There is no TypeObject corresponding to the SCCIdentifier. The closest there is the sequence of component type objects from the hashing algorithm. 7.6.3.2.1 and elsewhere says you can only use the SCCIdentifier, except when returning `getTypes` at which point you have to use the identifiers for the individual components.
> 
> To summarize:
> 
> `getTypeDependencies`
> 
>     * in: TypeIdentifier (minimal/complete) or SCCIdentifier (index 0).
> 
>     * out: TypeIdentifierWithSize (same as above) <-- This is where the spec lacks clarity.
> 
> 
> `getTypes`
> 
>     * in: TypeIdentifier (minimal/complete) or SCCIdentifier (index 0).
> 
>     * out: TypeIdentifier and TypeObject (SCCComponentIds must have index != 0 and all must be present)

Ah, I see now, I didn't realize the "Size" part of TypeIdentifiersWithSize was supposed to be the size of the TypeObject rather than the TypeIdentifier. Which makes more sense now that I think about it. In the code I shared I think I'm just returning the size of the serialized TypeIdentifier.

dds/DCPS/XTypes/TypeLookupService.cpp
```diff
@@ -103,27 +123,17 @@ void TypeLookupService::get_type_dependencies(const TypeIdentifierSeq& type_ids,
...
 {
-  OPENDDS_SET(TypeIdentifier) tmp;
   for (unsigned i = 0; i < type_ids.length(); ++i) {
-    const TypeIdentifierWithSizeSeqMap::const_iterator it = type_dependencies_map_.find(type_ids[i]);
+    const TypeIdentifier tid = make_scc_id_or_default(type_ids[i]);
+    const TypeIdentifierWithSizeSeqMap::const_iterator it = type_dependencies_map_.find(tid);
     if (it != type_dependencies_map_.end()) {
       for (unsigned j = 0; j < it->second.length(); ++j) {
-        tmp.insert(it->second[j].type_id);
+        const size_t idx = dependencies.length();
```
simpsont-oci, line 131:
```suggestion
        const auto idx = dependencies.length();
```

dds/DCPS/XTypes/TypeObject.cpp
```diff
@@ -997,24 +1012,32 @@ namespace {
...
 
-  void compute_dependencies_i(const TypeMap&,
-                              const StronglyConnectedComponentId&,
-                              OPENDDS_SET(TypeIdentifier)&)
+  void compute_dependencies_i(const TypeMap& type_map,
+                              const StronglyConnectedComponentId& type,
+                              TypeIdentifierSet& dependencies)
   {
-    // Do nothing.
+    // An SCCId (identifies the entire component.)
```
simpsont-oci, line 1026: If I have a topic type T which (eventually) depends on an SCC, and I request the dependencies of just type T... doesn't this code mean I'll _always_ get all the dependencies for the whole SCC (even if I haven't asked explicitly asked for it yet?). Is that what's intended / desired?

dds/DCPS/XTypes/TypeLookupService.cpp
simpsont-oci: signed / unsigned warning

simpsont-oci (changes requested):

dds/DCPS/XTypes/TypeObject.cpp
jrw972: That is the intended behavior.  If you ask for the dependencies of T, you should get all of the dependencies of T.

PR #4947, opened 2025-03-11
FlexibleTypes delay association

jrw972:
Problem
-------

Enabling flexible types causes new publications and subscriptions to be written.  These are outside of the durable data which means they go through a delay, heartbeat, and acknack cycle.  This delays association since there is definitely a match but the local match cannot proceed without confirmation that the remote received the local publication/subscription.  The goal is to eliminate the delay.

Solution
--------

- Keep track of endpoints that write samples when enabling flexible types.
- Request acknowledgement on these endpoints.
  - Enable support for request acknowledgements in RTPS.
- Update the progressive association logic to only require the reader to acknowledge the relevant publication/subscription.

### DIFF
--- a/dds/DCPS/transport/framework/DataLinkSet.cpp
+++ b/dds/DCPS/transport/framework/DataLinkSet.cpp
@@ -107,11 +107,26 @@ void OpenDDS::DCPS::DataLinkSet::terminate_send_if_suspended()
   }
 }
 
+OpenDDS::DCPS::SequenceNumber OpenDDS::DCPS::DataLinkSet::cur_cumulative_ack(const GUID_t& writer_id,
+                                                                             const GUID_t& reader_id) const
+{
+  SequenceNumber sn = SequenceNumber::ZERO();
+
+  GuardType guard(this->lock_);
+  // Note: Only one DataLink will handle the writer/reader pair.  The other DataLinks will return 0.
+  for (MapType::const_iterator pos = map_.begin(), limit = map_.end(); pos != limit; ++pos) {
+    sn = std::max(sn, pos->second->cur_cumulative_ack(writer_id, reader_id));
+  }
+
+  return sn;
+}
+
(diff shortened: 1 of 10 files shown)


dds/DCPS/transport/framework/DataLinkSet.cpp
```diff
@@ -107,6 +107,19 @@ void OpenDDS::DCPS::DataLinkSet::terminate_send_if_suspended()
...
 }
 
+OpenDDS::DCPS::SequenceNumber OpenDDS::DCPS::DataLinkSet::cur_cumulative_ack(const GUID_t& writer_id,
+                                                                             const GUID_t& reader_id) const
+{
+  SequenceNumber sn = SequenceNumber::ZERO();
+
+  GuardType guard(this->lock_);
+  for (MapType::const_iterator pos = map_.begin(), limit = map_.end(); pos != limit; ++pos) {
+    sn = std::max(sn, pos->second->cur_cumulative_ack(writer_id, reader_id));
```
mitza-oci, line 117: What's the meaning of the maximum cumulative ack across multiple DataLinks?  It seems like there would be only one DataLink that handles this writer/reader pair -- maybe add a comment?

dds/DCPS/transport/rtps_udp/RtpsUdpDataLink.h
```diff
@@ -382,6 +388,7 @@ class OpenDDS_Rtps_Udp_Export RtpsUdpDataLink
     void expunge_durable_data();
     bool expecting_durable_data() const;
     SequenceNumber acked_sn() const { return cur_cumulative_ack_.previous(); }
+    SequenceNumber cur_cumulative_ack () const { return cur_cumulative_ack_; }
```
mitza-oci, line 391:
```suggestion
    SequenceNumber cur_cumulative_ack() const { return cur_cumulative_ack_; }
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #4954, opened 2025-03-20
Consolidate and Correct CMake Export Header Code

iguessthislldo:
- This is to fix an issue on Windows where public export macro should be set to override the ACE default.
- Added a function `opendds_use_existing_export_header` to give users more flexibility when using a committed export header.
- Added a function to set `/bigobj` on targets that need it on Windows.


docs/devguide/building/cmake.rst
```diff
@@ -654,6 +738,15 @@ Functions
 
   .. versionadded:: 3.26
 
+.. cmake:func:: opendds_bigobj
+
+  ::
+
+    opendds_bigobj(<target>)
+
+  A small function that adds `/bigobj <https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file>`__ to the compile options of the targets when using Visual Studio.
```
mitza-oci, line 747:
```suggestion
  A function that adds `/bigobj <https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file>`__ to the compile options of the targets when using Visual Studio.
```

docs/devguide/building/cmake.rst
```diff
@@ -109,7 +109,7 @@ Components
 .. versionadded:: 3.25
 
 By default the package will search for all libraries and executables, but only require the bare minimum for ACE/TAO and OpenDDS.
-Arguments can be passed to ``find_package(OpenDDS COMPONENTS <argument>...)`` (or alternatively ``find_package(OpenDDS REQUIRED [COMPONENTS] <argument>...)``) add to what's required.
+Arguments can be passed to ``find_package(OpenDDS COMPONENTS <argument>...)`` (or alternatively ``find_package(OpenDDS REQUIRED [COMPONENTS] <argument>...)``) require additional executives, libraries, and features.
```
sonndinh, line 112:
```suggestion
Arguments can be passed to ``find_package(OpenDDS COMPONENTS <argument>...)`` (or alternatively ``find_package(OpenDDS REQUIRED [COMPONENTS] <argument>...)``) require additional executables, libraries, and features.
```

docs/devguide/building/cmake.rst
```diff
@@ -601,13 +604,50 @@ Functions
...
+    public:
+      int x_;
+      Doohickey(int x): x_(x) {}
+    };
+
+    Doohickey_Export void proccess_doohickey(Doohickey& doohickey);
+
+  .. cmake:func:arg:: EXISTING
+
+    By default a header file will generated, but the IDL-generated code can also use an existing compatible export header if that is preferred.
```
sonndinh, line 647:
```suggestion
    By default a header file will be generated, but the IDL-generated code can also use an existing compatible export header if that is preferred.
```

iguessthislldo:
I'm not sure why it was just one build, but CMake 4.0 was released and that build got it. I sorta knew this was coming, but our GoogleTest doesn't work with CMake 4.0 by default. I've made the fix in https://github.com/OpenDDS/googletest/pull/1.

jrw972 (approved):

PR #4959, opened 2025-03-31
Better Compatibility with CMake 4 

iguessthislldo:
- Update GoogleTest submodule to https://github.com/OpenDDS/googletest/pull/1
    
- Add a check for `cmake_minimum_required` to the lint script.
    
- Fix relative path issues when working within symlinked directory, including a warning about https://cmake.org/cmake/help/latest/policy/CMP0177.html.


mitza-oci (approved):

jrw972 (approved):

iguessthislldo:
I can reproduce the issue with [the failing build](https://github.com/OpenDDS/OpenDDS/actions/runs/14178898079/job/39720264840?pr=4959) locally. I'm 100% sure why it's happening, but I know it's related to setting the `cmake_minimum_required`'s "max policy" to 3.28 or later. It looks like [this CMake policy](https://cmake.org/cmake/help/latest/policy/CMP0154.html#policy:CMP0154) is what's causing it because the policy affects build order and "no features" is the only one to use Ninja and the latest CMake though. When I force the policy off, it still happens. Complying with it is kinda difficult because `opendds_target_sources` is agnostic to the installation method of files. Even after I changed it to use file sets, it still hasn't worked so far.

PR #4961, opened 2025-03-31
Remove virtual inheritance of RcObject in cases that don't need it

mitza-oci:


jrw972 (approved):

iguessthislldo (approved):

jrw972 (approved):

simpsont-oci:
Question: Shouldn't we default to virtual inheritance of RcObject to avoid accidentally instantiating two atomic reference counts?

mitza-oci:
> Question: Shouldn't we default to virtual inheritance of RcObject to avoid accidentally instantiating two atomic reference counts?

There's additional runtime cost for virtual inheritance and certain refcount operations are bound to be on the hot path (in addition to these odd warnings from certain compilers / architectures), so I don't think there's a single strategy that will be best in all cases.

If we need to reconsider the change in general, I'm OK with that.  In that case we would probably replace 32-bit Windows builds with 64-bit since it's hard to imagine that 32-bit is that useful to anyone.

We can also look at specific cases where an RcObject-derived class is meant to be a mix-in base class (as we saw with NetworkConfigMonitor before) and roll those back.

PR #4972, opened 2025-04-14
RtpsRelay: Add sizes of key data structures to reported statistics

mitza-oci:


sonndinh (approved):
I mostly reviewed style only.

PR #4974, opened 2025-04-16
[v4] Definitions missing from OpenDDSConfig.h

jrw972:
Problem
-------

OpenDDSConfig.h doesn't contain all build-time definitions.

Solution
--------
Add
- `OPENDDS_CONFIG_BUILT_IN_TOPICS`
- `OPENDDS_CONFIG_CONTENT_FILTERED_TOPIC`
- `OPENDDS_CONFIG_CONTENT_SUBSCRIPTION`
- `OPENDDS_CONFIG_MULTI_TOPIC`
- `OPENDDS_CONFIG_QUERY_CONDITION`
- `OPENDDS_CONFIG_BUILT_IN_TOPICS`
- `OPENDDS_CONFIG_CONTENT_FILTERED_TOPIC`
- `OPENDDS_CONFIG_CXX11`
- `OPENDDS_CONFIG_MULTI_TOPIC`
- `OPENDDS_CONFIG_OBJECT_MODEL_PROFILE`
- `OPENDDS_CONFIG_OWNERSHIP_PROFILE`
- `OPENDDS_CONFIG_PERSISTENCE_PROFILE`
- `OPENDDS_CONFIG_QUERY_CONDITION`
- `OPENDDS_CONFIG_RAPIDJSON`
- `OPENDDS_CONFIG_SAFETY_PROFILE`


cmake/init.cmake
```diff
@@ -520,6 +520,7 @@ else()
 endif()
 _opendds_feature(CXX17 "${_opendds_cxx17}" MPC_INVERTED_NAME no_cxx17 DOC "Build assumes C++17 support")
 _opendds_feature(STD_OPTIONAL "${OPENDDS_CXX17}" CONFIG DOC "Use C++17's std::optional")
+_opendds_feature(RAPIDJSON "${OPENDDS_RAPIDJSON}" CONFIG DOC "Enable rapidjson features" TYPE PATH)
```
iguessthislldo, line 523: So it's blank. After thinking about it, the problem with rapidjson in the u22 build is probably in the root `CMakeLists.txt`:

https://github.com/OpenDDS/OpenDDS/blob/62c78ee53f7556b99c517d192fe4fa2d99f394d5/CMakeLists.txt#L67-L97

Here we're loading `init.cmake`, then setting RapidJSON because originally `init.cmake` didn't originally do anything with RapidJSON, so the order didn't matter. Now we're setting `OPENDDS_RAPIDJSON` in `init.cmake`, so it's skipping it later after it comes back from `init.cmake`. I think the solution is to move the RapidJSON setup to before the ACE/TAO and `init.cmake` lines I linked to (aka swap these lines around). The RapidJSON lines don't depend on the ACE/TAO/`init.cmake` lines, but now `init.cmake` does depend on the RapidJSON being set correctly by default. It's the same with MPC built OpenDDS, but in that case the `config.cmake` is already read first beforehand.

iguessthislldo:
> --   OPENDDS_RAPIDJSON=/home/runner/work/OpenDDS/OpenDDS/OpenDDS/tools/rapidjson

Looks like it's going to work now.

cmake/init.cmake
mitza-oci, line 523:
```suggestion
_opendds_feature(RAPIDJSON "${OPENDDS_RAPIDJSON}" CONFIG DOC "Enable RapidJSON features" TYPE PATH)
```

Based on docs on https://github.com/miloyip/rapidjson.git

configure
```diff
@@ -1480,14 +1480,37 @@ sub write_opendds_configh {
...
+    'OPENDDS_CONFIG_CONTENT_FILTERED_TOPIC' => $opts{'content-filtered-topic'} // ($opts{'safety-profile'} ? 0 : 1),
+    'OPENDDS_CONFIG_CONTENT_SUBSCRIPTION' => $opts{'content-subscription'} // ($opts{'safety-profile'} ? 0 : 1),
+    'OPENDDS_CONFIG_MULTI_TOPIC' => $opts{'multi-topic'} // ($opts{'safety-profile'} ? 0 : 1),
+    'OPENDDS_CONFIG_OBJECT_MODEL_PROFILE' => $opts{'object-model-profile'} // ($opts{'safety-profile'} ? 0 : 1),
     'OPENDDS_CONFIG_OWNERSHIP_KIND_EXCLUSIVE' => $exclusive,
+    'OPENDDS_CONFIG_OWNERSHIP_PROFILE' => $opts{'ownership-profile'} // 1,
+    'OPENDDS_CONFIG_PERSISTENCE_PROFILE' => $opts{'persistence-profile'} // ($opts{'safety-profile'} ? 0 : 1),
+    'OPENDDS_CONFIG_QUERY_CONDITION' => $opts{'query-condition'} // ($opts{'safety-profile'} ? 0 : 1),
+    'OPENDDS_CONFIG_RAPIDJSON' => ($opts{'rapidjson'} // 0) ? 1 : 0,
+    'OPENDDS_CONFIG_SAFETY_PROFILE' => ($buildEnv{'build'} eq 'host') ? 0 : ($opts{'safety-profile'} ? 1 : 0),
```
mitza-oci, line 1503: Refactor common subexpression `$opts{'safety-profile'} ? 1 : 0)`

configure
```diff
@@ -1480,14 +1480,37 @@ sub write_opendds_configh {
...
+    if ($f eq 'no_cxx11=0') {
+      $cxx11 = 1;
+    }
+  }
+
   my %config = (
     'OPENDDS_CONFIG_AUTO_STATIC_INCLUDES' => 0,
     'OPENDDS_CONFIG_BOOTTIME_TIMERS' => $opts{'boottime'} // 0,
+    'OPENDDS_CONFIG_BUILT_IN_TOPICS' => $opts{'built-in-topics'} // 1,
+    'OPENDDS_CONFIG_CONTENT_FILTERED_TOPIC' => $opts{'content-filtered-topic'} // ($opts{'safety-profile'} ? 0 : 1),
```
mitza-oci, line 1494: If we have a `--no-content-subscription` configuration, is `OPENDDS_CONFIG_CONTENT_FILTERED_TOPIC` set to 0 (along with multi topic and query condition)?

dds/DCPS/Service_Participant.h
```diff
@@ -59,10 +59,10 @@ const char COMMON_DCPS_BIDIR_GIOP[] = "COMMON_DCPS_BIDIR_GIOP";
 const bool COMMON_DCPS_BIDIR_GIOP_default = true;
 
 const char COMMON_DCPS_BIT[] = "COMMON_DCPS_BIT";
-#ifdef DDS_HAS_MINIMUM_BIT
-const bool COMMON_DCPS_BIT_default = false;
-#else
+#if OPENDDS_CONFIG_BUILT_IN_TOPICS
 const bool COMMON_DCPS_BIT_default = true;
+#else
+const bool COMMON_DCPS_BIT_default = false;
```
mitza-oci, line 65: This could be a single line: `const bool COMMON_DCPS_BIT_default = OPENDDS_CONFIG_BUILT_IN_TOPICS;`
Which also means we can eliminate `COMMON_DCPS_BIT_default` and just use `OPENDDS_CONFIG_BUILT_IN_TOPICS` (unless it does something significant with type deduction so 0/1 would be different from false/true).

.github/workflows/build_and_test.yml
```diff
@@ -5909,7 +5909,7 @@ jobs:
         cd OpenDDS
         . setenv.sh
         cd tests/DCPS/Messenger
-        perl $ACE_ROOT/bin/mwc.pl -type gnuace -static
+        perl $DDS_ROOT/bin/opendds_mwc.pl -type gnuace -static
```
mitza-oci, line 5912: Is `-static` needed or is it already in `opendds_mwc.pl`?  Same for `-type gnuace`.

dds/InfoRepo/DCPSInfoRepoServ.cpp
```diff
@@ -41,10 +42,10 @@
 InfoRepo::InfoRepo(int argc, ACE_TCHAR *argv[])
 : ior_file_(ACE_TEXT("repo.ior"))
 , listen_address_given_(0)
-#ifdef DDS_HAS_MINIMUM_BIT
-, use_bits_(false)
-#else
+#if OPENDDS_CONFIG_BUILT_IN_TOPICS
 , use_bits_(true)
+#else
+, use_bits_(false)
```
mitza-oci, line 48:
```suggestion
, use_bits_(OPENDDS_CONFIG_BUILT_IN_TOPICS)
```

And remove `#endif` below (GitHub won't make it part of the suggested change).

docs/devguide/building/index.rst
```diff
@@ -104,6 +104,68 @@ If configure runs successfully it will end with a message about the next steps f
...
+A template can be found in :ghfile:`dds/OpenDDSConfig.h.in`.
+
+The configuration header is not used directly.
+Instead, :ghfile:`dds/OpenDDSConfigWrapper.h` includes it, provides defaults and backwards compatibility, and performs some sanity checks.
+For consistency and to avoid build errors, applications that have code that is conditioned on OpenDDS features should include :ghfile:`dds/OpenDDSConfigWrapper.h`.
+
+The following macros are available in the config header:
+
+``OPENDDS_CONFIG_AUTO_STATIC_INCLUDES``
+    TODO
```
mitza-oci, line 124: TODO

tools/scripts/face_pp.rb
mitza-oci, line 1: I'm not sure what this file is used for -- time to remove it?

mitza-oci (changes requested):
I haven't reviewed files in `*tests/`

.github/workflows/build_and_test.yml
iguessthislldo: Yes, both are included:

https://github.com/OpenDDS/OpenDDS/blob/62c78ee53f7556b99c517d192fe4fa2d99f394d5/configure#L2530-L2533

Then later:

https://github.com/OpenDDS/OpenDDS/blob/62c78ee53f7556b99c517d192fe4fa2d99f394d5/configure#L2549

iguessthislldo: Also it should be in the PATH, so it could just be: `opendds_mwc.pl`

configure
jrw972: No.  There is a check that content subscription is enabled if any of content filtered, multi topic, or query condition are enabled.  https://github.com/jrw972/OpenDDS/blob/4e1251a65634bb9058e0fa3b1ede156264608e26/dds/OpenDDSConfigWrapper.h#L201

If we want to deal with this in the configure script, then I think it would be better to enable content subscription if any of the dependents are enabled.

.github/workflows/build_and_test.yml
jrw972: When I peeked at GHA, it looked like `-type gnuace` was there but `-static` was not.  Probably just missed it.

docs/devguide/building/index.rst
jrw972: Yeah.  So what is a decent blurb for this?

iguessthislldo: This is the current explanation in a section that talks about the libraries:

https://github.com/OpenDDS/OpenDDS/blob/62c78ee53f7556b99c517d192fe4fa2d99f394d5/docs/devguide/introduction.rst?plain=1#L458

configure
mitza-oci: Everything defaults to enabled (except with safety profile), so I think the more common use case is the one I described above.  The logic in `dcps_optional_features.mpb` that this PR is removing/replacing says that if `content_subscription == 0` then the 3 sub-features get turned off.

So it may be good to change the configure script, or maybe change the wrapper to do the inverse of this block:

https://github.com/jrw972/OpenDDS/blob/4e1251a65634bb9058e0fa3b1ede156264608e26/dds/OpenDDSConfigWrapper.h#L195

If the user configures `OPENDDS_CONFIG_CONTENT_SUBSCRIPTION == 0` then it turns off the 3 sub-features.

.github/workflows/build_and_test.yml
mitza-oci: > Also it should be in the PATH, so it could just be: `opendds_mwc.pl`

That's right, but also keep in mind that on Windows sometimes the `*.pl` association doesn't work so it needs an explicit `perl %DDS_ROOT%\bin\opendds_mwc.pl`

configure
mitza-oci: Maybe you have this already in the latest revision?

.github/workflows/build_and_test.yml
jrw972:
``` my $static = (($opts{'static'} && $is_windows) ||
    ($cross_compile && $buildEnv->{'build'} eq 'host' && $mpctype ne 'gnuace'));```

From this, it seems like `-static` is superfluous on the non-windows builds.

File: configure (hunk 3 above)
Comment (mitza-oci) [reply configure]: See `MPC\config\dcps_optional_features.mpb` -- the umbrella feature is not a preprocessor macro, just a shortcut to get all 3 other macros set

File: MPC/modules/IDLBase.pm [hunk 8]
```diff
@@ -839,13 +839,27 @@ sub preprocess {
   return $contents, $ts_str, $ts_pragma;
 }
 
+# expand $(TAO_ROOT) and $(DDS_ROOT).
+sub expand {
+  my($self, $path) = @_;
+
+  if ($path =~ m/^\$\(([^)]+)\)/ && exists $ENV{$1}) {
+      my $replacement = $ENV{$1};
+      $path =~ s/^\$\([^)]+\)/$replacement/;
```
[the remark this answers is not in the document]
jrw972: I think `$1` can be set by the `s` operator so I thought this approach was easier to understand.

docs/devguide/building/index.rst
iguessthislldo: Yes, it's what's doing the "auto including" if `OPENDDS_CONFIG_AUTO_STATIC_INCLUDES` is true:

https://github.com/OpenDDS/OpenDDS/blob/62c78ee53f7556b99c517d192fe4fa2d99f394d5/dds/DCPS/StaticIncludes.h#L6-L12

iguessthislldo (approved):

sonndinh (approved):

PR #4977, opened 2025-04-18
configure script option --static-runtime

mitza-oci:
This is meant to be used in combination with --static.

Using Visual C++, default is to link statically to the platform runtime libraries so specify --no-static-runtime to use DLLs.

Using gnuace, --static-runtime adds static_stdlibs=1 to platform_macros.GNU, which some platforms use.


cmake/ace_group.cmake
```diff
@@ -63,7 +63,7 @@ function(_opendds_vs_force_static)
   endif()
 endfunction()
 
-if(MSVC AND OPENDDS_STATIC)
+if(MSVC AND OPENDDS_STATIC AND NOT ${OPENDDS_STATIC_RUNTIME} STREQUAL "OFF")
```
iguessthislldo, line 66:
```suggestion
if(MSVC AND OPENDDS_STATIC AND NOT OPENDDS_STATIC_RUNTIME)
```

iguessthislldo (changes requested):

cmake/ace_group.cmake
mitza-oci: If the user doesn't specify the option to the configure script, the variable will be the empty string instead of OFF or ON.  CMake would treat the empty string as a false value.

mitza-oci:
@iguessthislldo When the build of ACE+TAO is launched via CMake and `configure_ace_tao.pl`, we need to determine how to handle static vs. DLL runtime libs.  Should it be based on https://cmake.org/cmake/help/v3.31/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html?  See https://github.com/OpenDDS/OpenDDS/actions/runs/14574647381/job/40878098939?pr=4977

cmake/ace_group.cmake
iguessthislldo: What about:

```suggestion
if(MSVC AND OPENDDS_STATIC AND NOT (OPENDDS_STATIC_RUNTIME STREQUAL "" OR  OPENDDS_STATIC_RUNTIME))
```

I think it'd be better if it's not dependent on a specific false value.

iguessthislldo: Actually this still seems wrong...

iguessthislldo:
I think so because we should be translating the CMake options to MPC options. However, I think that CMake failure is because the `OPENDDS_STATIC_RUNTIME` check is using `${OPENDDS_STATIC_RUNTIME}` instead of `OPENDDS_STATIC_RUNTIME`. For example:

```CMake
foreach(OPENDDS_STATIC TRUE FALSE)
  if(OPENDDS_STATIC AND NOT ${OPENDDS_STATIC_RUNTIME} STREQUAL "OFF")
    message("Set static runtime stuff")
  else()
    message("Don't set static runtime stuff")
  endif()
endforeach()

foreach(OPENDDS_STATIC TRUE FALSE)
  if(OPENDDS_STATIC AND NOT OPENDDS_STATIC_RUNTIME STREQUAL "OFF")
    message("Set static runtime stuff")
  else()
    message("Don't set static runtime stuff")
  endif()
endforeach()

```

```
$ cmake -P test.cmake 
Don't set static runtime stuff
Don't set static runtime stuff
Set static runtime stuff
Don't set static runtime stuff
```

So it doesn't work right when `OPENDDS_STATIC_RUNTIME` isn't defined. I think when `OPENDDS_STATIC_RUNTIME` isn't defined it's the same as `"NOT" STREQUAL "OFF"`?

cmake/ace_group.cmake
iguessthislldo: I just realized I inverted `OPENDDS_STATIC_RUNTIME` when I originally commented.

iguessthislldo: Maybe we should only set `OPENDDS_STATIC_RUNTIME` when `--static-runtime` or `--no-static-runtime` is passed so it's only ever true, false, or undefined. Then this could be:

```suggestion
if(MSVC)
  if(OPENDDS_STATIC AND NOT DEFINED OPENDDS_STATIC_RUNTIME)
    set(OPENDDS_STATIC_RUNTIME TRUE)
  endif()
  if(OPENDDS_STATIC_RUNTIME)
```

cmake/ace_group.cmake
```diff
@@ -63,7 +63,9 @@ function(_opendds_vs_force_static)
   endif()
 endfunction()
 
-if(MSVC AND OPENDDS_STATIC)
+if(MSVC AND OPENDDS_STATIC AND DEFINED OPENDDS_STATIC_RUNTIME AND (OPENDDS_STATIC_RUNTIME STREQUAL "" OR OPENDDS_STATIC_RUNTIME))
```
sonndinh, line 66: It looks like `OPENDDS_STATIC_RUNTIME` will always be set in the generated config.cmake, is that correct? If so, I think this line can skip the `DEFINED` check.

configure
```diff
@@ -206,6 +206,7 @@ my @specs =
     'optimize!', 'Optimization (no)',
     'inline!', 'Inlining (yes)',
     'static!', 'Static libraries (no)',
+    'static-runtime!', 'Link to static platform runtime libraries',
```
sonndinh, line 209: My understanding is that this option should only be used with `--static`. If that is correct, should we document that here or somewhere else?

cmake/ace_group.cmake
mitza-oci: The check is required since this code is used for cases that don't have a `config.cmake` (these are cases where OpenDDS is built with CMake).

configure
mitza-oci: It will only take effect with `--static`.  I think the existing mechanisms for understanding the configure script's behavior, namely using `--verbose`, running `tools/scripts/show_build/config.pl`, and reading the code are sufficient, but some may disagree.

cmake/build_ace_tao.cmake
```diff
@@ -21,6 +21,9 @@ set(_opendds_std "$<TARGET_PROPERTY:${_opendds_std_target},CXX_STANDARD>")
 find_package(Perl REQUIRED)
 if(OPENDDS_STATIC)
   list(APPEND _OPENDDS_CONFIGURE_ACE_TAO_ARGS --static=1)
+  if(NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY OR CMAKE_MSVC_RUNTIME_LIBRARY MATCHES "DLL")
```
iguessthislldo, line 24: If we're accounting for `CMAKE_MSVC_RUNTIME_LIBRARY` here, we should also probably do it in `_opendds_vs_force_static`.

cmake/ace_group.cmake
iguessthislldo: Yes, we do need it work without it being defined, but the way it's done here changes the behavior of the CMake-built OpenDDS. Now it won't ever force it by default.... which might be fine since it now leaves it to CMake to decide. Is that intentional?

I'm also still not comfortable with how the empty string is being used as a pseudo-true value. If we're assuming the current behavior, this could be:

```suggestion
if(MSVC AND OPENDDS_STATIC AND DEFINED OPENDDS_STATIC_RUNTIME AND OPENDDS_STATIC_RUNTIME)
```

Then the default value in the configure script would be changed to `1`.

iguessthislldo (changes requested):

cmake/ace_group.cmake
mitza-oci: The `configure` script is reporting one of three states, it's not boolean.  Empty, ON, and OFF are each their own values.

Maybe that's a non-optimal design but I'm not that interested in fixing it in this PR.

cmake/build_ace_tao.cmake
iguessthislldo: This was made assuming the `OPENDDS_STATIC_RUNTIME` change was unintentional. Since it's not, this code is mutually exclusive with the `_opendds_vs_force_static`.

cmake/ace_group.cmake
iguessthislldo: > The `configure` script is reporting one of three states, it's not boolean. Empty, ON, and OFF are each their own values.

> I guess that didn't really answer the question -- we're also trying to determine if the configure script was used at all.

Okay, but the code doesn't distinguish between `OPENDDS_STATIC_RUNTIME` being empty and `OPENDDS_STATIC_RUNTIME` being true, so there's no point to breaking a standard convention here.

iguessthislldo (changes requested):

cmake/ace_group.cmake
mitza-oci: Empty being treated as true is only the case when `MSVC` is also true.  Other platforms have their own defaults for static vs. shared linking to standard library runtime libraries.

The `OPENDDS_STATIC_RUNTIME` value is just relaying to CMake how `configure` was run (user specified `--static-runtime`, user specified `--no-static-runtime`, or neither), it's not adding additional logic for defaults or specific platforms.

iguessthislldo: If we want leave it open to cover those cases, we could leave it undefined if it wasn't specified, then:

```suggestion
if(MSVC AND OPENDDS_STATIC)
  if(NOT DEFINED OPENDDS_STATIC_RUNTIME AND NOT (_OPENDDS_CMAKE_BUILT_AND_INSTALLED OR OPENDDS_IS_BEING_BUILT))
    set(OPENDDS_STATIC_RUNTIME TRUE)
  endif()
  if(OPENDDS_STATIC_RUNTIME)
```

mitza-oci: That would require more changes to `configure`, and I don't see any cases where it already handles an `undef` Perl value or has other logic for omitting it from `config.cmake`.

PR #4983, opened 2025-04-24
A CMake Variable for if OpenDDS was CMake-built

iguessthislldo:
Also fixed `version-added` in CMake docs.


configure
```diff
@@ -2889,7 +2889,9 @@ sub write_cmake_file {
     $name = uc("OPENDDS_$name");
     $name =~ s/-/_/g;
 
-    print $fh "set($name ", to_cmake_value($value), ")\n";
+    if(defined($value)) {
```
mitza-oci, line 2892:
```suggestion
    if (defined($value)) {
```

mitza-oci (changes requested):

mitza-oci (approved):

PR #4964, opened 2025-04-03
Update CMake support for custom export header

sonndinh:

PR #4986, opened 2025-04-26
Misc CMake Fixes

iguessthislldo:
- Fixed reconfigured cross-compiled builds trying to use `TRUE` as the path to host tools.
- Fixed `OPENDDS_COMPILE_WARNINGS=ERROR`, which didn't work before.
  - As part of this, added common INI file reading function for CMake.
- Restructure CMake News
- Try to improve news and documentation from #4964


docs/news.d/try_compile.rst
```diff
@@ -2,10 +2,12 @@
 
 .. news-start-section: Platform Support and Dependencies
 .. news-start-section: CMake
+.. news-start-section: Building OpenDDS with CMake
 - The :ref:`shapes-demo` will now be built when :ref:`cmake-building` and :cmake:var:`OPENDDS_QT` is set to ``TRUE``.
+- Invalid :cmake:var:`OPENDDS_ACE_TAO_KIND` values will now be explictly rejected.
```
mitza-oci, line 7: Is this user-facing?

iguessthislldo: [Yes](https://opendds.readthedocs.io/en/master/devguide/building/index.html#cmake-var-OPENDDS_ACE_TAO_KIND), it's how the user would choose what ACE/TAO to get. Whether or not this is news worthy could be debated though.

mitza-oci: Right, I was thinking that it doesn't need to be in the NEWS

sonndinh (approved):

PR #4989, opened 2025-04-30
Fixed invalid CMake variable name generated by configure

mitza-oci:
Looks like a bug introduced in #4487 resulted in a duplicated OPENDDS_ prefix on this one variable.

### DIFF
--- a/configure
+++ b/configure
@@ -2979,7 +2979,7 @@ sub write_cmake_file {
     }
 
     if (defined($cxx_std)) {
-      print_cmake_config($fh, 'OPENDDS_CXX_STD', $cxx_std);
+      print_cmake_config($fh, 'CXX_STD', $cxx_std);
     }
 
     # TODO(iguessthislldo): Move to a smarter system that can use existing


iguessthislldo:
Looks like this won't have worked correctly even if I inserted the CMake variable correctly. [`03` needs to be translated to `98`.](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

mitza-oci:
> Looks like this won't have worked correctly even if I inserted the CMake variable correctly. [`03` needs to be translated to `98`.](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

Can/should that be handled in `cmake/init.cmake`?

iguessthislldo (approved):

sonndinh (approved):

cmake/init.cmake
```diff
@@ -382,7 +382,11 @@ endif()
 
 function(_opendds_cxx_std target scope)
   if(OPENDDS_CXX_STD AND NOT (scope STREQUAL INTERFACE AND CMAKE_VERSION VERSION_LESS "3.11.0"))
-    target_compile_features(${target} ${scope} "cxx_std_${OPENDDS_CXX_STD}")
+    if(OPENDDS_CXX_STD STREQUAL 03)
+      target_compile_features(${target} ${scope} "cxx_std_98")
```
iguessthislldo, line 386: `OPENDDS_CXX_STD=03` should have been turned into `98` above. Did this not work?

iguessthislldo (changes requested):

cmake/init.cmake
mitza-oci: https://github.com/OpenDDS/OpenDDS/actions/runs/14762691597/job/41484095095

iguessthislldo: I checked out this branch and played with it. I thought these lines would normalize it, but its not:

https://github.com/OpenDDS/OpenDDS/blob/ab5b693fc009a5b1d80af539233c91cc456ddc08/cmake/init.cmake#L371-L373

Values at this point:

```
cxx_std_year: 1998
cxx_std: 98
-- OPENDDS_CXX_STD: 03 (from OPENDDS_CXX_STD)
```

Apparently this is a behavior of CMake I was not aware of where the [normal variables have priority over the cache variables](https://stackoverflow.com/questions/78151540/set-variable-with-cache-force-cache-does-not-update-the-cmake-variable). [(CMP0126)](https://cmake.org/cmake/help/latest/policy/CMP0126.html).

I'm a little concerned about this in existing code because I would have assumed cached values would override the normal variables. This might require changes to how variables are set in `init.cmake` and used elsewhere.

In the mean time, can you remove this and add `unset(OPENDDS_CXX_STD PARENT_SCOPE)` before these lines?

https://github.com/OpenDDS/OpenDDS/blob/ab5b693fc009a5b1d80af539233c91cc456ddc08/cmake/init.cmake#L372-L373

mitza-oci: Done in d8ee7bdfd9

iguessthislldo (approved):

PR #4990, opened 2025-04-30
Use an InternalTopic to report statistics about the internals of RtpsDiscovery in the RtpsRelay

mitza-oci:


dds/DCPS/RTPS/RtpsDiscovery.cpp
```diff
@@ -783,6 +800,19 @@ void RtpsDiscovery::request_remote_complete_type_objects(
   spdp->request_remote_complete_type_objects(remote_entity, remote_type_info, cond);
 }
 
+void RtpsDiscovery::write_stats(const MonotonicTimePoint&) const
+{
+  ACE_Guard<ACE_Thread_Mutex> guard(participants_lock_);
+  DCPS::InternalStatistics statistics;
+  for (DomainParticipantMap::const_iterator domain = participants_.begin(); domain != participants_.end(); ++domain) {
+    for (ParticipantMap::const_iterator part = domain->second.begin(); part != domain->second.end(); ++part) {
+      statistics.id = ("RtpsDiscovery " + DCPS::to_dds_string(domain->first) + ' ' + DCPS::to_string(part->first)).c_str();
```
jrw972, line 809: Do we want to allow the user to give a friendly name to a participant and use it (if set) here?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -4762,6 +4781,68 @@ bool Spdp::has_domain_participant(const GUID_t& remote) const
...
+  stats[Stats_Index_TotalWriterAssociated].name = "TotalWriterAssociated";
+  stats[Stats_Index_TotalReaderPending].name = "TotalReaderPending";
+  stats[Stats_Index_TotalReaderAssociated].name = "TotalReaderAssociated";
+  for (DDS::UInt32 i = 0; i < sedp_template.length(); ++i) {
+    stats[Stats_Len + i].name = sedp_template[i].name;
+  }
+  return stats;
+}
+
+void Spdp::fill_stats(DCPS::InternalStatisticSeq& stats) const
```
jrw972, line 4826: What about `directed_guids_`?

jrw972 (commented):
In terms of naming, I think we should think about dropping "internal" from the name.  Yes, they are statistics internal to OpenDDS but they are intended for user consumption.

Should the periodic tasks that collect the statistics also log the statistics when configured to do so?

mitza-oci:
> In terms of naming, I think we should think about dropping "internal" from the name. Yes, they are statistics internal to OpenDDS but they are intended for user consumption.

I'm fine with that but it may be confusing with other things named similarly, like `Stats_T.h` and `WriterStats` in `DataReaderImpl.h`

> Should the periodic tasks that collect the statistics also log the statistics when configured to do so?

How about making that its own reader/listener?  That way the logging happens in one place instead of scattered.

dds/DCPS/RTPS/RtpsDiscovery.cpp
mitza-oci: For many applications, including RtpsRelay, the domain ID is sufficient.

I think it could be an option for future development.

jrw972: For posterity, when using security, it could be the subject name.

mitza-oci: Two participants could use the same security credentials

mitza-oci:
> > In terms of naming, I think we should think about dropping "internal" from the name. Yes, they are statistics internal to OpenDDS but they are intended for user consumption.
> 
> I'm fine with that but it may be confusing with other things named similarly, like `Stats_T.h` and `WriterStats` in `DataReaderImpl.h`

There's a naming conflict here https://github.com/OpenDDS/OpenDDS/actions/runs/14782684162/job/41504880416?pr=4990#step:6:49739  --  other ideas @jrw972 ?

jrw972:
> > > In terms of naming, I think we should think about dropping "internal" from the name. Yes, they are statistics internal to OpenDDS but they are intended for user consumption.
> > 
> > 
> > I'm fine with that but it may be confusing with other things named similarly, like `Stats_T.h` and `WriterStats` in `DataReaderImpl.h`
> 
> There's a naming conflict here https://github.com/OpenDDS/OpenDDS/actions/runs/14782684162/job/41504880416?pr=4990#step:6:49739 -- other ideas @jrw972 ?

It's either rename `Statistics` in `monitor.idl` or revert back to internal statistics.  I don't have a strong preference.  We can always rename in v4.

mitza-oci:
> > > > In terms of naming, I think we should think about dropping "internal" from the name. Yes, they are statistics internal to OpenDDS but they are intended for user consumption.
> > > 
> > > 
> > > I'm fine with that but it may be confusing with other things named similarly, like `Stats_T.h` and `WriterStats` in `DataReaderImpl.h`
> > 
> > 
> > There's a naming conflict here https://github.com/OpenDDS/OpenDDS/actions/runs/14782684162/job/41504880416?pr=4990#step:6:49739 -- other ideas @jrw972 ?
> 
> It's either rename `Statistics` in `monitor.idl` or revert back to internal statistics. I don't have a strong preference. We can always rename in v4.

Since monitor is in its own shared library, we can put it in its own namespace.  Other code isn't depending on the namespace.

jrw972 (approved):

iguessthislldo (approved):

iguessthislldo:
Need to add the new header to the header set so it's installed:

https://github.com/OpenDDS/OpenDDS/blob/9f2a9e8dd4e750de4dd01f849e800cda46d400ef/dds/CMakeLists.txt#L196-L197

tests/unit-tests/dds/DCPS/Statistics.cpp
```diff
@@ -0,0 +1,23 @@
+#include <dds/DCPS/Statistics.h>
+
```
sonndinh, line 2:
```suggestion
```

tests/unit-tests/dds/DCPS/Statistics.cpp
```diff
@@ -0,0 +1,23 @@
...
+using namespace OpenDDS::DCPS;
+
+TEST(dds_DCPS_Statistics, key_compare)
+{
+  typedef InternalTraits<Statistics>::KeyCompare Comparator;
+  const Comparator cmp = Comparator();
+
+  const StatisticSeq empty, nonempty(1, 1, StatisticSeq::allocbuf(1), true);
+  const Statistics is1 = {"a", empty}, is2 = {"a", nonempty}, is3 = {"b", empty};
```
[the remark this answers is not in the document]
mitza-oci: They could be, but it's just in the unit test

tests/unit-tests/dds/DCPS/Statistics.cpp
mitza-oci: This is a common pattern in unit tests - the thing we're testing is its own "group"

tests/unit-tests/dds/DCPS/Statistics.cpp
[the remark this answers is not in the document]
sonndinh: There is a similar name in RelayStatisticsReporter.cpp, but since it's only used in a short loop, I think its use is clear.

sonndinh (approved):

tests/unit-tests/dds/DCPS/Statistics.cpp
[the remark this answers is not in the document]
mitza-oci: I can change that in the next PR

PR #4997, opened 2025-05-04
GHA Annotations and Better Link Check for Sphinx

iguessthislldo:
This should make Sphinx output the correct annotation in RST files in the PR in most situations except for when the content causing the error is in a news fragment. News fragments are used to generate two files at runtime, so the source info that Sphinx has for these files doesn't lead the correct annotation in the PR. See below for what the workaround looks like.

Also tried to improve link checking so it fails less often. This includes:

- Ignoring `:ghfile:` links. It's not really needed because they're checked locally. There's also a lot of them so it takes a while to check them and there's a greater chance for something going wrong.
- Ignoring links that fail with 409 HTTP status code. They happen a lot.
- Turning redirects on GitHub PR and issue links into errors because that means something is wrong with the link.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -242,7 +242,7 @@ Read [the documentation for this release on Read the Docs](https://opendds.readt
 
 ### Documentation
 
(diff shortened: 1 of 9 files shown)


iguessthislldo:
Examples of the annotations:

Screenshot from 2025-05-03 23-39-47

Screenshot from 2025-05-03 23-48-08

iguessthislldo:
Example of the annotation for content in the news fragment. These were on the job summary page, but not in the "File Changed" page.

Screenshot from 2025-05-03 23-55-11

PR #5009, opened 2025-05-14
CMake: Set the FOLDER property on custom targets

mitza-oci:
When CMake generates projects for an IDE, all custom targets for IDL code generation will be grouped together in a folder.

### DIFF
--- a/cmake/dds_idl_sources.cmake
+++ b/cmake/dds_idl_sources.cmake
@@ -20,6 +20,7 @@ function(_opendds_target_idl_sources target)
     SKIP_OPENDDS_IDL
     AUTO_INCLUDES
     INCLUDE_BASE
+    FOLDER
   )
   set(multi_value_args TAO_IDL_FLAGS DDS_IDL_FLAGS IDL_FILES)
   cmake_parse_arguments(arg "" "${one_value_args}" "${multi_value_args}" ${ARGN})
@@ -274,7 +275,7 @@ function(_opendds_target_idl_sources target)
     set_target_properties(${target} PROPERTIES _OPENDDS_IDL_FILE_COUNT ${idl_file_count})
     set(idl_target "_opendds_codegen_${idl_file_count}_for_${target}")
     add_custom_target(${idl_target} DEPENDS ${generated_files})
-    set_target_properties(${idl_target} PROPERTIES FOLDER IDL)
+    set_target_properties(${idl_target} PROPERTIES FOLDER ${arg_FOLDER})
     add_dependencies(${target} ${idl_target})
 
     set_source_files_properties(${idl_files} ${h_files}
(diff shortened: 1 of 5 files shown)


iguessthislldo:
Out of curiosity, what does this look like in Visual Studio?

mitza-oci:
> Out of curiosity, what does this look like in Visual Studio?
image

mitza-oci:
@doug1234 would it be better to have something like `IDL Code Generation` instead of just `IDL` as the label?

iguessthislldo:
Yikes! Yea, I wouldn't have imagined it would look like that, assuming those were mixed with the other projects.

I thought I was clever by solving a few problems by doing this. These were supporting the `codegen` target and replacing the "bridge" target system. I did some searching and It would be nice if CMake had [implemented the ability to hide targets from the IDE](https://gitlab.kitware.com/cmake/cmake/-/issues/23566). Assuming that doesn't come. It might be nice to give these slightly better names if they are going to be visible. I did number to be safe, but maybe base them off the IDL file name since they all are one file each? There maybe could also be only one codegen target per normal target, but I feel like there was a reason I didn't do that, I just don't remember what it was at the moment.

In the mean time, this might have to do.

iguessthislldo:
> @doug1234 would it be better to have something like `IDL Code Generation` instead of just `IDL` as the label?

I think it's better. Also to properly support Doug's current solution, we could either provide a property with the names of the custom IDL targets or provide an option to `opendds_target_sources` to override the `FOLDER` property of the targets.

mitza-oci:
> Yikes! Yea, I wouldn't have imagined it would look like that, assuming those were mixed with the other projects.

They were not interleaved because sorting alphabetically put them together, but they weren't set apart in their own folder.  The folders should default to collapsed so it's less intrusive.

> I thought I was clever by solving a few problems by doing this. These were supporting the `codegen` target and replacing the "bridge" target system. I did some searching and It would be nice if CMake had [implemented the ability to hide targets from the IDE](https://gitlab.kitware.com/cmake/cmake/-/issues/23566). Assuming that doesn't come. It might be nice to give these slightly better names if they are going to be visible. I used numbers to be safe, but maybe base them off the IDL file name since they all are one file each? There maybe could also be only one codegen target per normal target, but I feel like there was a reason I didn't do that, I just don't remember what it was at the moment.
> 
> In the mean time, this might have to do.

doug1234:
So... if you look at idl2library it will now create two levels in visual studio.  At the IDL level are the projects that generate the code for each idl file specified.  Under that in the generated file each of the generated targets is located (like in the image you showed).

doug1234:
The way you have it now isn't quite ideal.  Can you make it so I can set the folder the dependency targets end up in?

mitza-oci:
> The way you have it now isn't quite ideal. Can you make it so I can set the folder the dependency targets end up in?

Yes, if we implement what @iguessthislldo suggested in the comment here https://github.com/OpenDDS/OpenDDS/pull/5009#issuecomment-2880599901

How about making it an option to `opendds_target_sources` and  a cache variable that sets the default?

jrw972 (approved):

PR #5018, opened 2025-05-22
[v4] C++17 is minimum version for configure/MPC/CMake

jrw972:


NEWS.md
```diff
@@ -47,7 +47,7 @@ Read [the documentation for this release on Read the Docs](https://opendds.readt
 - OpenDDS's use of `std::optional` or an emulation is determined by the configuration file ([PR #4976](https://github.com/OpenDDS/OpenDDS/pull/4976))
   - See `OPENDDS_CONFIG_STD_OPTIONAL` in `dds/OpenDDSConfig.h`
   - Default is to use `std::optional` on compilers that support it
-  - See configure script's `--no-std-optional` or CMake's [`OPENDDS_STD_OPTIONAL`](https://opendds.readthedocs.io/en/v3.32.0/devguide/building/index.html#cmake-var-OPENDDS_STD_OPTIONAL)
+  - See configure script's `--no-std-optional` or CMake's `OPENDDS_STD_OPTIONAL`
```
sonndinh: The readthedocs link for v3.32 still exists, so it may be good to keep it here.

iguessthislldo: Yes, that's what I imagined when creating the news fragment system. Remove/change links in `news.d/_releases`, but leave `NEWS.md` alone because it pointed to the specific version. I tried to think about a better way to handle this kind of situation, but never came up with anything.

sonndinh (approved):

PR #5029, opened 2025-06-09
Use IDL maps in DdsDynamicData.idl

mitza-oci:
Not ready to merge until `tao_idl` changes are ready.


dds/DCPS/XTypes/DynamicDataImpl.cpp
```diff
@@ -477,6 +474,9 @@ DDS::ReturnCode_t DynamicDataImpl::clear_value(DDS::MemberId id)
           }
         }
         set_backing_store(0);
+#else
```
jrw972, line 477: Why is the old code not in the `else`? (Same question for other files.)

mitza-oci: The `#else` is supposed to be for very rare cases.  We don't suggest building with this old version of ACE_TAO, but some user may be doing so and it was decided not to make it completely unsupported until OpenDDS v4.  Because it's rare, we don't want to maintain blocks of code which are essentially dead.  These odd builds won't have support for dynamic data.

We could go back to the old code in the `#else` case if we want to maintain working dynamic data features with ACE v7 / TAO v3.

PR #5034, opened 2025-06-27
[v4] Removed support for ACE version 7

mitza-oci:
As of this PR, ACE version 6.5.x (latest release on that branch) and ACE version 8 (master branch) are both still supported.
In a future PR for the v4 branch, ACE version 6.5.x will also be removed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -95,7 +95,7 @@ is_release = opendds_version_info.is_release
 if is_release:
     github_links_release_tag = opendds_version_info.tag
 ace6tao2_version = opendds_version_info.ace6tao2_version
-ace7tao3_version = opendds_version_info.ace7tao3_version
+ace8tao4_version = opendds_version_info.ace8tao4_version
 if opendds_version_info.release_year:
     copyright = f'{opendds_version_info.release_year} {copyright}'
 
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -95,7 +95,7 @@ def setup(app):
 if is_release:
     github_links_release_tag = opendds_version_info.tag
 ace6tao2_version = opendds_version_info.ace6tao2_version
-ace7tao3_version = opendds_version_info.ace7tao3_version
+ace8tao4_version = opendds_version_info.ace8tao4_version
```
iguessthislldo: I don't think these variables for ACE/TAO version are used anymore. It looks like they used to be used in the dependencies page:

https://github.com/OpenDDS/OpenDDS/blob/1df2cdcb742b266a9266de8835238b2a690b3dad/docs/building/dependencies.rst?plain=1#L44-L45

But now they use a RST role specifically made for this:

https://github.com/OpenDDS/OpenDDS/blob/c377df3ae4a73d9e6927636d907cbabee921ddf3/docs/devguide/building/dependencies.rst?plain=1#L56-L57

And that uses the `opendds_version_info` variable:

https://github.com/OpenDDS/OpenDDS/blob/c377df3ae4a73d9e6927636d907cbabee921ddf3/docs/sphinx_extensions/links.py#L178-L183

So I think these two lines can be removed.

mitza-oci: I was wondering where they were referenced.

PR #5035, opened 2025-06-27
[v4] Updated GoogleTest submodule URL and version

mitza-oci:


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	openddsConfigureCommit = fd3dc29a5c2852df569e1ea81dbde2c412ac5051
 [submodule "tests/googletest"]
 	path = tests/googletest
-	url = https://github.com/OpenDDS/googletest
-	branch = v1.8.x
+	url = https://github.com/google/googletest
+	branch = 1.17.x
```
iguessthislldo: Will also need to update this separately for CMake unfortunately:

https://github.com/OpenDDS/OpenDDS/blob/a55b551aaff76e4b023fed69b226b4b22e0fc126/tests/CMakeLists.txt#L24-L25

Added an issue for this: #5037.

iguessthislldo (approved):

configure
```diff
@@ -1995,9 +1995,7 @@ if ($build_gtest) {
   my $install_dir = $build_dir . $slash . 'install';
   my $cmake = $opts{'cmake'};
   my @cmake_platform_args = $is_windows ?
-    ('-Dgtest_force_shared_crt=ON',
```
iguessthislldo, line 1998: Do we need to remove this in CMake as well?

https://github.com/OpenDDS/OpenDDS/blob/3e10bff784217b60f2cbb54b2d24db7e090c8f4f/tests/CMakeLists.txt#L33-L37

mitza-oci: We can, I think that it doesn't really take effect in this version though.

PR #5053, opened 2025-07-21
Enable DDS Security on GitHub Actions CMake Windows builds

mitza-oci:


.github/workflows/cmake.yml
```diff
@@ -354,7 +355,8 @@ jobs:
           -DOPENDDS_ACE_TAO_SRC=${{ github.workspace }}\OpenDDS\build\ACE_TAO^
           -DOPENDDS_MPC=${{ github.workspace }}\MPC^
           -DOPENDDS_XERCES3=${{ env.VCPKG_INSTALLED_DIR }}/x64-windows^
```
iguessthislldo, line 357: Can we also remove the Xerces variables?

mitza-oci: It didn't work when I tried

https://github.com/OpenDDS/OpenDDS/pull/5053/commits/a30d1d02b6dbed8d086a92edc7d7a8bfdc79d5f5

iguessthislldo: Oh, yea, I actually fixed that and a few other things in my windows VM at home. I changed it to use the `LOCATION` property on the `XercesC::XercesC` target instead of whatever `XercesC_LIBRARY` is doing. I meant to clean it up and make a PR on the way, but forgot to transfer the patch file before leaving. I've been meaning to setup remote access to that VM but I keep putting it off.

iguessthislldo: If you guys merge this today I can make a PR tonight with my fixes and I can try removing the variables.

PR #5054, opened 2025-07-23
Fix RapidJSON and Xerces Issues in CMake Build

iguessthislldo:
Sort of a follow up to https://github.com/OpenDDS/OpenDDS/pull/5053.

- Fixed accidental use of RapidJSON's `CMakeLists.txt` file when building OpenDDS with CMake. This caused issues such as forcing the `CMAKE_CXX_STANDARD` to C++11 and breaking the ACE C++ standard detection code.
- Fixed configure issue with Xerces from vcpkg when building OpenDDS with CMake.


iguessthislldo:
Vcpkg toolchain file looks like it worked.

For reference, rapidjson change was made in [#4973](https://github.com/OpenDDS/OpenDDS/pull/4973/commits/38786edec90080979808959174c1a070756ffbaa), when `FetchContent_Populate(rapidjson)` was replaced with `FetchContent_MakeAvailable(rapidjson)`.

PR #5055, opened 2025-07-23
InternalDataWriter doesn't set `source_timestamp`

jrw972:
Problem
-------

InternalDataWriter doesn't provide a `source_timestamp` to the InternalDataReader for the SampleInfo.

Solution
--------

Provide the `source_timestamp`.

### DIFF
--- a/dds/DCPS/TimeSource.h
+++ b/dds/DCPS/TimeSource.h
@@ -23,10 +23,15 @@ class TimeSource {
 public:
   virtual ~TimeSource() {}
 
-  virtual MonotonicTimePoint monotonic_time_point_now() const {
+  virtual MonotonicTimePoint monotonic_time_point_now() const
+  {
     return MonotonicTimePoint::now();
   }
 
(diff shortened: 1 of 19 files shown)


dds/DCPS/TimeSource.h
```diff
@@ -27,6 +27,10 @@ class TimeSource {
     return MonotonicTimePoint::now();
   }
 
+  virtual DDS::Time_t dds_time_t_now() const{
```
mitza-oci:
```suggestion
  virtual DDS::Time_t dds_time_t_now() const
  {
```

Also fix the code above please

mitza-oci (changes requested):
Does this qualify for a NEWS fragment?

jrw972:
> Does this qualify for a NEWS fragment?

If we are expecting users to use them (perhaps for statistics?) then yes.

mitza-oci (approved):

PR #5060, opened 2025-07-25
Fixes and Improvements for `TestFramework`

iguessthislldo:
- Fix DCS `wait_for` using wrong path and exit with an error status when the wait failed.
- Add `parse_func_opts` for Python keyword arg-like functionality in Perl. Make use of that and `trace` in `TestFramework`.
- Remove `DCS` directory after in addition to before test framework runs. Remove redundant `rmtree('./DCS')` from various tests
- Call `finish()` automatically on object destruction.
- Don't crash when process object has already been destroyed.
- Fix import usage between the two packages in `Run_Test.pm`.
- Allow changing any `TestFramework` member value from `flag=value` arguments.


bin/PerlDDS/Run_Test.pm
```diff
@@ -7,13 +6,22 @@ use warnings;
 
 package PerlDDS;
 
+# NOTE: There is another package called TestFramework in this file. If you need
+# imports for that package, add them there, not here.
```
mitza-oci, line 10: Should we split this into two files and have one import the other?

tests/DCPS/TransientLocalMultiInstanceTest/run_test.pl
```diff
@@ -47,11 +47,11 @@
 
 $test->start_process('sub1');
 
-$test->wait_for('driver', 'sub1', 'ready', 5);
+$test->wait_for('sub1', 'ready', max_wait => 5);
```
mitza-oci, line 50: #5063 (merged today) has similar code in the RtpsRelay test script.  Does it also need to change?

iguessthislldo: Yes, I will merge master and make the change.

bin/PerlDDS/Run_Test.pm
iguessthislldo: I thought about it, but I wasn't sure it was worth the work or not.

PR #5130, opened 2025-10-07
Consolidate cached RtpsRelay config values

sonndinh:
**Problem**
RtpsRelay reads config values from command-line arguments and stores them in the `ServiceParticipant`'s config store. Cached config values are populated via a config listener which is called asynchronously. The `Config::set_defaults()` function, however, immediately checks whether each cached value is empty, and if so, sets a default value for the cached parameter. This causes a race since the result of the check depends on whether the listener callback finishes before the check. Consequently, the config values from the command-line arguments can be overwritten by the default values.

**Solution**
Set the cached config values from the command-line arguments at the same time as storing them in the config store.

### DIFF
--- a/tools/rtpsrelay/Config.h
+++ b/tools/rtpsrelay/Config.h
@@ -146,11 +146,7 @@ public:
   void lifespan(const OpenDDS::DCPS::TimeDuration& value)
   {
     lifespan(value.to_dds_duration());
-  }
-
-  void lifespan(const DDS::Duration_t& value)
-  {
-    TheServiceParticipant->config_store()->set_duration(RTPS_RELAY_LIFESPAN, value);
+    cached_lifespan_.set(value);
   }
 
   OpenDDS::DCPS::TimeDuration lifespan() const
@@ -161,11 +157,7 @@ public:
   void inactive_period(const OpenDDS::DCPS::TimeDuration& value)
   {
     inactive_period(value.to_dds_duration());
-  }
-
-  void inactive_period(const DDS::Duration_t& value)
-  {
-    TheServiceParticipant->config_store()->set_duration(RTPS_RELAY_INACTIVE_PERIOD, value);
+    cached_inactive_period_.set(value);
   }
 
   OpenDDS::DCPS::TimeDuration inactive_period() const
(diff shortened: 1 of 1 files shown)


tools/rtpsrelay/Config.h
```diff
@@ -457,6 +470,16 @@ class Config : public OpenDDS::DCPS::ConfigListener {
 
   bool parse_admission_participants_range(const char* arg);
 
+  void lifespan(const DDS::Duration_t& value)
+  {
+    TheServiceParticipant->config_store()->set_duration(RTPS_RELAY_LIFESPAN, value);
```
mitza-oci, line 473: Should this also updated `cached_lifespan_`?

mitza-oci:
I don't think this needs a separate NEWS entry, but please add this PR# to the one in `docs/news.d/rtpsrelay-dynamic-config.rst`

tools/rtpsrelay/Config.h
sonndinh: Actually, now I think setting the cache here makes more sense since the other `lifespan` function, which takes a `TimeDuration` parameter, calls this function.

tools/rtpsrelay/Config.h
```diff
@@ -473,12 +496,19 @@ class Config : public OpenDDS::DCPS::ConfigListener {
...
+    }
+
     template <typename Fn>
-    void default_if_empty(Fn&& func) const
+    void default_if_empty(Fn&& func)
     {
       ACE_Guard<ACE_Thread_Mutex> g{lock_};
       if (!value_) {
         func(default_value_);
+        value_ = default_value_;
```
mitza-oci, line 511: The `lock_` is held during the callback -- does that work?  The callback is going to end up back in the set() function based on the other changes to this class.  This might need to be redesigned.

mitza-oci: Oh, I didn't see the other changes when I reviewed this revision.  So the callback is only calling config store.  Now we have code duplication so probably not ideal but OK.

sonndinh: I'm setting up a build to test but I think it works. The problem was `default_if_empty` calls the added `set` function (in the lambda `func`) and both try to hold lock. This breaks that by not calling the newer `set` from `default_if_empty`.

PR #5132, opened 2025-10-21
More details for thread error messages in RtpsRelay

sonndinh:


tools/rtpsrelay/RelayThreadMonitor.cpp
```diff
@@ -102,8 +102,8 @@ int RelayThreadMonitor::svc()
     for (const auto idx : late_thread_indexes) {
       const SystemTimePoint timestamp(infos[idx].source_timestamp);
       const auto& thread = datas[idx];
-      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: RelayThreadMonitor::svc thread %C (%d, %d) last update %#T.  Aborting...\n",
-                 thread.thread_id.in(), thread.detail1, thread.detail2, &timestamp.value()));
+      ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: RelayThreadMonitor::svc thread %C (%C, %d) last update %#T.  Aborting...\n",
+                 thread.thread_id.in(), decompose_thread_detail1(thread.detail1).c_str(), thread.detail2, &timestamp.value()));
```
mitza-oci, line 106: Currently, the detail data members are only used by the relay handler threads.  But other threads could use them with their own arbitrary integer IDs.

tools/rtpsrelay/RelayThreadMonitor.cpp
```diff
@@ -145,4 +145,40 @@ void RelayThreadMonitor::on_data_available(DDS::DataReader_ptr /*reader*/)
...
 }
 
+std::string RelayThreadMonitor::decompose_thread_detail1(int detail) const
+{
+  std::string result;
+  if (detail & ACE_Event_Handler::READ_MASK) {
+    result += "READ_MASK";
+  }
+  if (detail & ACE_Event_Handler::WRITE_MASK) {
+    result += " | WRITE_MASK";
```
mitza-oci: It might look odd for the resulting string to start with `| `.

tools/rtpsrelay/RelayThreadMonitor.cpp
sonndinh: The detail data members look like something we can generalize/categorize so different threads can communicate their events uniformly. But maybe it's not imminent?

tools/rtpsrelay/RelayThreadMonitor.cpp
```diff
@@ -148,35 +148,37 @@ void RelayThreadMonitor::on_data_available(DDS::DataReader_ptr /*reader*/)
 std::string RelayThreadMonitor::decompose_thread_detail1(int detail) const
 {
   std::string result;
+  bool first = true;
```
mitza-oci: The current implementation doesn't set `first = false` on all branches.  This could be replaced with `result.empty()` so the state doesn't need to be kept in two different variables.

tools/rtpsrelay/RelayThreadMonitor.cpp
mitza-oci: We can leave that for a future PR

PR #5135, opened 2025-10-28
Support denied partitions in RtpsRelay

sonndinh:
Add a new topic for specifying a list of partitions to deny. Also a new command-line option is added to specify how long a denied partition will remain in the RtpsRelay instance.

### DIFF
--- a/tools/rtpsrelay/GuidAddrSet.cpp
+++ b/tools/rtpsrelay/GuidAddrSet.cpp
@@ -586,7 +586,7 @@ void GuidAddrSet::populate_relay_status(RelayStatus& relay_status)
 void GuidAddrSet::deny(const OpenDDS::DCPS::GUID_t& guid)
 {
   const auto it = guid_addr_set_map_.find(guid);
-  if (it != guid_addr_set_map_.end()) {
+  if (it != guid_addr_set_map_.end() && it->second.allow_stun_responses) {
     it->second.allow_stun_responses = false;
     it->second.in_denied_partition = true;
     ++mark_count_;


mitza-oci (approved):

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -149,4 +156,71 @@ void GuidPartitionTable::lookup(StringSet& partitions, const OpenDDS::DCPS::GUID
...
+    for (auto it = denied_partitions_.begin(); it != denied_partitions_.end();) {
+      if (it->second <= cutoff_time) {
+        it = denied_partitions_.erase(it);
+      } else {
+        ++it;
+      }
+    }
+
+    if (!denied_partitions_.empty()) {
+      denied_partitions_cleanup_task_->schedule(config_.denied_partitions_timeout());
```
simpsont-oci, line 208: I'm guessing the requests to add denied partitions can come in whenever? If so, we might want a bit finer granularity on when we're rescheduling cleanup. The "first time" (line 181) won't happen until the timeout expires (this seems OK), but if for some reason we have a second set of denied partitions coming in afterwards, they won't be currently be cleaned up until the next "multiple" of the timeout value passes (currently defaults to 15 minutes)... which might be nearly 15 minutes later than you'd expect if the second set came in shortly after the first.

mitza-oci: Right, each partition should have its own expiry time so they don't keep getting extended.

sonndinh: Right. This approach trades off exact cleanup time for simpler implementation. I think an important point is the denied partitions will eventually be cleaned up and the delay is bounded by at most 1 timeout value, e.g., denied partitions from a request received at minute 1, ideally would be cleaned up at minute 16, is actually cleaned up at minute 30 and the delay is 14 mins. Since it's bounded by the timeout value, we have control over it using the command-line option.

mitza-oci (approved):
Review comment has been resolved

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -224,7 +224,7 @@ GuidAddrSet::record_activity(const AddrPort& remote_address,
 
   switch (drain_state_) {
   case DrainState::DS_NORMAL:
-    if (!addr_set_stats.allow_stun_responses) {
+    if (!addr_set_stats.allow_stun_responses && !addr_set_stats.in_denied_partition) {
```
simpsont-oci: I'm not doubting this is correct, but can someone explain the logic here?

What I'm following so far is that if we're not draining (DrainState::DS_NORMAL) and we see some activity from an address (src_guid -> cass -> addr_set_stats), then we would normally toggle on `allow_stun_responses` decrement `mark_count_`, but now we don't do that if the address was in a denied partition.

What are `mark_count_` and `mark_budget_`?

PR #5138, opened 2025-11-11
Support RelayDeniedPartitions topic in RtpsRelayControl & add denied partitions test for RtpsRelay

sonndinh:


tests/DCPS/RtpsRelay/Smoke/publisher.cpp
```diff
@@ -371,7 +371,7 @@ int ACE_TMAIN(int argc, ACE_TCHAR *argv[])
 
       } else if (args.stress_test) {
         stress_test(dw, participant, type_name, pub);
-      } else if (args.drain_test) {
+      } else if (args.drain_test || args.reject_partition_test) {
```
mitza-oci: Should this be "deny" (or "denied") for consistency with other changes in this PR?

tools/rtpsrelay/control/RtpsRelayControl.cpp
```diff
@@ -214,6 +218,15 @@ int run(int argc, ACE_TCHAR* argv[])
   participant->get_default_subscriber_qos(subscriber_qos);
   subscriber_qos.partition.name = publisher_qos.partition.name;
 
+  DDS::Publisher_var publisher;
+  if (!config.config().empty() || !denied_partitions.empty()) {
+    publisher = participant->create_publisher(publisher_qos, nullptr, 0);
```
mitza-oci: The denied partitions writer could use a publisher that doesn't have partition QoS so it writes to all instances.

tools/rtpsrelay/control/RtpsRelayControl.cpp
```diff
@@ -171,6 +173,8 @@ int run(int argc, ACE_TCHAR* argv[])
       } else {
         throw std::runtime_error{"Argument to -Set must be NAME=Value: " + arg_str};
       }
+    } else if ((arg = args.get_the_parameter("-Deny"))) {
```
mitza-oci, line 177: Update devguide and NEWS for this?

mitza-oci (changes requested):

tools/rtpsrelay/control/RtpsRelayControl.cpp
mitza-oci: And this would include waiting for publication matched > 1

sonndinh: When `-RelayId` is not specified, this publisher will use `"*"` in the partition QoS. I think that should make it write to all relay instances.

mitza-oci: Writing to > 1 reader (relay) is a design change from the previous version.  In order to do so correctly it will need changes to the publication_matched part of the code -- this is not a durable topic.  It will probably need some way to know how many readers to expect.

sonndinh: I'm using a different publisher for denied partitions, separate from the publisher for the relay config control topic, since the denied partitions are published to all RtpsRelay instances.

A new command-line option is added to specify the minimum number of readers before sending.

tools/rtpsrelay/control/RtpsRelayControl.cpp
```diff
@@ -224,9 +308,9 @@ int run(int argc, ACE_TCHAR* argv[])
       return EXIT_FAILURE;
     }
 
-    DDS::Publisher_var publisher{participant->create_publisher(publisher_qos, nullptr, 0)};
+    DDS::Publisher_var publisher = participant->create_publisher(publisher_qos, nullptr, 0);
```
mitza-oci: Why did this line change?

sonndinh: I manually moved it back from a previous commit and caused this change accidentally. I'll revert it back to the original style.

PR #5161, opened 2026-01-06
Remove reactor's `notify` within `handle_exception`

sonndinh:
- Problem:

`JobQueue`'s `handle_exception` notification to reactor is likely redundant, as a notification has already been inserted in `enqueue` method. This may add spurious notifications, for which the job queue doesn't have any jobs, to the reactor.

`ReactorTask` uses a similar pattern that notifies the reactor in its `handle_exception` method. This is due to `execute_or_enqueue` not always notifying the reactor when a new command is added -- it doesn't notify if the reactor thread is processing existing commands, so the notification is postponed until `handle_exception` finishes processing the commands in that case. The reactor is thus notified in two places.

- Solution:

Remove the redundant `notify` call from `JobQueue`'s `handle_exception`.

Decouple the reactor thread's state (`RS_PROCESSING`) from the reactor object's notification state (`RS_NONE` and `RS_NOTIFIED`). This helps simplify the code a bit and also avoids calling `notify` within `handle_exception`.

### DIFF
--- a/dds/DCPS/JobQueue.cpp
+++ b/dds/DCPS/JobQueue.cpp
@@ -38,12 +38,6 @@ int JobQueue::handle_exception(ACE_HANDLE /*fd*/)
     (*pos)->execute();
   }
 
-  ACE_GUARD_RETURN(ACE_Thread_Mutex, guard, mutex_, -1);
-  if (!job_queue_.empty()) {
-    guard.release();
-    reactor()->notify(this);
-  }
-
   return 0;
 }
 


dds/DCPS/ReactorTask.cpp
```diff
@@ -442,30 +443,28 @@ int ReactorTask::handle_exception(ACE_HANDLE /*fd*/)
...
 {
   Queue cq;
   ACE_Reverse_Lock<ACE_Thread_Mutex> rev_lock(lock_);
 
-  reactor_state_ = RS_PROCESSING;
+  processing_ = true;
   if (!command_queue_.empty()) {
     cq.swap(command_queue_);
+    // The current notification is being processed, reset to allow a new notification
+    reactor_state_ = RS_NONE;
```
simpsont-oci: Since the only two states are now `RS_NONE` and `RS_NOTIFIED`, I think I would expect any call to `process_command_queue_i` to set `reactor_state_` to `RS_NONE`. Is there a reason it's only reset to `RS_NONE` when `command_queue_` is not empty?

dds/DCPS/ReactorTask.h
```diff
@@ -209,16 +209,18 @@ class OpenDDS_Dcps_Export ReactorTask
   typedef OPENDDS_VECTOR(CommandPtr) Queue;
   Queue command_queue_;
 
+  // Whether the reactor has been notified about pending commands
   enum ReactorState {
```
simpsont-oci: At this point, `bool notified_` might make more sense than a `ReactorState`

dds/DCPS/ReactorTask.cpp
sonndinh: I think the state should be `RS_NONE` already if the command queue is empty, so it doesn't need to reset in that case. My understanding is, semantically when the command queue is empty, there should be no outstanding notification in the reactor.

dds/DCPS/ReactorTask.h
sonndinh: Yea, that makes sense.

PR #5166, opened 2026-02-02
Harvest thread status with a separate thread

sonndinh:


simpsont-oci:
FYI, on my branch I migrated the thread status task to event dispatcher in this commit (may have other stuff as well): https://github.com/OpenDDS/OpenDDS/pull/5160/changes/3742d156884e50897e90901be30af77b9244a68a

I think the only other thing you might have to do is expose the event dispatcher from the sedp transport... Pretty sure I did that in a different commit.

PR #5169, opened 2026-02-05
Add --profiling flag to configure

simpsont-oci:
Add a `--profiling` flag to the configure script which, similar to `--sanitize`, caps the optimization to a reasonable level, uses `-fno-omit-frame-pointer` and forces the inclusion of debug symbols regardless of build type. Useful for when running a build through a profiling tool like valgrind's callgrind or anything that needs the ability to pull a clean stack trace.

### DIFF
--- a/docs/news.d/configure-script-profiling-flag.rst
+++ b/docs/news.d/configure-script-profiling-flag.rst
@@ -1,0 +1,10 @@
+.. news-prs: 5169
+
+.. news-start-section: Additions
+- Added a '--profiling' flag to the configure script to enable common flags used during profiling (gcc / clang)
+
+  - This enables some optimization ('-O3') but also includes debuging info ('-ggdb') and protects frame pointers ('-fno-omit-frame-pointer')
+    in order to help in generating clean stack traces for profilers and other utilities which examine the call stack.
+
+.. news-end-section
+


configure
```diff
@@ -3099,6 +3113,12 @@ sub write_cmake_file {
       print_cmake_config($fh, 'SANITIZER_COMPILER_ARGS', \@sanitizer_compiler_args);
       print_cmake_config($fh, 'SANITIZER_LINKER_ARGS', \@sanitizer_linker_args);
     }
+
+    print $fh "\n# Profiling\n";
```
mitza-oci: Should this be within the `if` block below?

configure
```diff
@@ -3099,6 +3113,12 @@ sub write_cmake_file {
...
+    if ($opts{'profiling'}) {
+      print_cmake_config($fh, 'PROFILING_COMPILER_ARGS', \@profiling_compiler_args);
+      print_cmake_config($fh, 'PROFILING_LINKER_ARGS', \@profiling_linker_args);
```
mitza-oci, line 3121: Should these add to other variables instead of creating new variables? Otherwise how would it be used from CMake?

simpsont-oci: Hmm. Not sure. I guess we'd have to update CMakeLists.txt to pick these up. I initially started this as an addition to the sanitizer flags but it felt a little weird for the command line flag to be something like `--sanitize=profiling` and then would have had to hack things to make sure it wasn't trying to pass an `-fsanitize` argument to the compiler.

simpsont-oci: @iguessthislldo What's the prefered way to handle this with our cmake files?

configure
simpsont-oci: I was following the pattern of printing "\n# Sanitizers\n" above, which doesn't seem conditional on anything... but maybe it should be as well?

mitza-oci: Right, I don't know why the comment needs to be in the output file if the thing it's commenting on is not

configure
simpsont-oci: I added a separate comment for enabling / disabling of sanitizers (always present) vs sanitizer args (optionally present) and followed the same (optionally present) pattern for profiling flags. Also added actual support for profiling flags to cmake config.

simpsont-oci: @iguessthislldo Can you take a quick look and verify the changes I added to init.cmake are sufficient for cmake support of the new profiling flag?

PR #5170, opened 2026-02-05
`PeriodicEvent` can create two chains of events

sonndinh:
- Problem
`PeriodicEvent::enable` can trigger two chains of events when `immediate_dispatch` is set to true

- Solution
Either dispatch or schedule depending on `immediate_dispatch`

### DIFF
--- a/dds/DCPS/PeriodicEvent.cpp
+++ b/dds/DCPS/PeriodicEvent.cpp
@@ -41,7 +41,7 @@ bool PeriodicEvent::enable(const TimeDuration& period, bool immediate_dispatch, 
         }
         return okay;
       } else {
-        long id = dispatcher->schedule(rchandle_from(this), expiration);
+        const long id = dispatcher->schedule(rchandle_from(this), expiration);
         if (id > 0) {
           period_ = period;
           expiration_ = expiration;
@@ -62,11 +62,11 @@ void PeriodicEvent::shorten_current_wait(const TimeDuration& period)
   const MonotonicTimePoint now = MonotonicTimePoint::now();
   ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
   if (timer_id_ > 0) {
-    EventDispatcher_rch dispatcher = dispatcher_.lock();
+    const EventDispatcher_rch dispatcher = dispatcher_.lock();
     if (dispatcher) {
       const MonotonicTimePoint expiration = now + period;
       if (expiration < expiration_ && dispatcher->cancel(timer_id_)) {
-        long id = dispatcher->schedule(rchandle_from(this), expiration);
+        const long id = dispatcher->schedule(rchandle_from(this), expiration);
         if (id > 0) {
           expiration_ = expiration;
           timer_id_ = id;
@@ -109,10 +109,10 @@ void PeriodicEvent::handle_event_scheduling()
   ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
   timer_id_ = 0;
   if (enabled_i()) {
-    EventDispatcher_rch dispatcher = dispatcher_.lock();
+    const EventDispatcher_rch dispatcher = dispatcher_.lock();
     if (dispatcher) {
       const MonotonicTimePoint expiration = (strict_timing_ ? expiration_ : MonotonicTimePoint::now()) + period_;
-      long id = dispatcher->schedule(rchandle_from(this), expiration);
+      const long id = dispatcher->schedule(rchandle_from(this), expiration);
       if (id > 0) {
         expiration_ = expiration;
         timer_id_ = id;


dds/DCPS/PeriodicEvent.h
```diff
@@ -72,6 +72,11 @@ class OpenDDS_Dcps_Export PeriodicEvent : public EventBase {
 
   void handle_event_scheduling();
 
+  bool enabled_i() const
+  {
+    return !period_.is_zero();
```
simpsont-oci, line 86: Overall this looks good, but the shift from using `timer_id_` to using `period_` to determine whether things are enabled makes me realize we probably ought to allow `enable` to fail (i.e. return `true` or `false` for whether the internal events were successfully scheduled / dispatched (both of which also can return `true` or `false`). This might happen, for example, if a periodic task decided to call `enable` just after the EventDispatcher had called `shutdown`.

simpsont-oci: So if you can, please add that an a small unit test.

sonndinh: Yes, that makes sense. I guess `EventDispatcher` has an independent lifecycle so different events, like periodic or sporadic events, from different places can share the same event dispatcher?

There is some case where using `timer_id_` may not be able to correctly decide the enable status. For example,  `enable` is called with immediate dispatch, followed by another `enable` call. The immediate dispatch from the first call puts the object to the dispatcher's queue but its `handle_event` might be processed after the second `enable` call. So the second `enable` call still sees the `timer_id_` of 0, and will trigger another event chain. Similarly, if a `disable` call immediately follows the first `enable` call, it might not be able to cancel the timer which will be scheduled later when the first dispatch from the `enable` call is processed. I think using `period_` prevents those cases since it's set directly in the `enable` call while `timer_id_` is set asynchronously by the one of the dispatcher threads.

dds/DCPS/PeriodicEvent.cpp
```diff
@@ -23,25 +23,57 @@ PeriodicEvent::PeriodicEvent(EventDispatcher_rch dispatcher, EventBase_rch event
...
+          strict_timing_ = strict_timing;
+          return true;
+        } else if (log_level >= LogLevel::Warning) {
+          ACE_ERROR((LM_WARNING, "(%P|%t) PeriodicEvent::enable: failed to schedule\n"));
+        }
+      }
+    }
+    return false;
+  }
+  return true;
```
simpsont-oci:
```suggestion
  }
  return false;
```

simpsont-oci: This is somewhat subjective, but I would think we want calls to enable to fail (return false) when it's already enabled, especially if the period / configuration flags are different, since after the call the event won't be behaving as expected.

dds/DCPS/PeriodicEvent.cpp
```diff
@@ -23,25 +23,57 @@ PeriodicEvent::PeriodicEvent(EventDispatcher_rch dispatcher, EventBase_rch event
...
+}
+
+void PeriodicEvent::shorten_current_wait(const TimeDuration& period)
+{
+  const MonotonicTimePoint now = MonotonicTimePoint::now();
+  ACE_Guard<ACE_Thread_Mutex> guard(mutex_);
+  if (timer_id_ > 0) {
```
simpsont-oci, line 64: Based on your changes, so that things match, I think the logic here needs to be:
```
if (enabled_i())
```
Not sure about the rest of it.

simpsont-oci (changes requested):

dds/DCPS/PeriodicEvent.cpp
sonndinh: I think in case of immediate dispatch, it's possible that `PeriodicEvent` is enabled (`enabled_i` return true) and the event is pushed into the event dispatcher's queue but it may still wait to be processed. In that case, `timer_id_` isn't set yet and can't be cancelled?

simpsont-oci: Oh, I see what you're saying. Yes, shortening the current wait only makes sense when there's an active timer. We can leave this as is.

simpsont-oci (approved):

dds/DCPS/PeriodicEvent.cpp
```diff
@@ -23,25 +23,56 @@ PeriodicEvent::PeriodicEvent(EventDispatcher_rch dispatcher, EventBase_rch event
...
         strict_timing_ = strict_timing;
-        if (immediate_dispatch) {
-          dispatcher->dispatch(rchandle_from(this));
+        const bool okay = dispatcher->dispatch(rchandle_from(this));
+        if (!okay) {
+          period_ = TimeDuration::zero_value;
+        }
+        return okay;
+      } else {
+        long id = dispatcher->schedule(rchandle_from(this), expiration);
```
mitza-oci:
```suggestion
        const long id = dispatcher->schedule(rchandle_from(this), expiration);
```

mitza-oci: Also on lines 65, 69, 112, 115

PR #5171, opened 2026-02-06
Callbacks when threads are started and finished

iguessthislldo:
Added callbacks for when an OpenDDS thread starts and finishes to control thread behavior. For example on Linux, using a user could use `pthread_setaffinity_np` to set which CPU core the thread runs on in order to dedicate cores to OpenDDS.

### DIFF
--- a/dds/DCPS/ReactorTask.cpp
+++ b/dds/DCPS/ReactorTask.cpp
@@ -221,7 +221,7 @@ int ReactorTask::svc()
 
     if (thread_status_timer_ == ReactorWrapper::InvalidTimerId) {
       if (log_level >= LogLevel::Notice) {
-        ACE_ERROR((LM_NOTICE, "(%P|%t) NOTICE: ReactorTask::svc: failed to "
+        ACE_ERROR((LM_ERROR, "(%P|%t) NOTICE: ReactorTask::svc: failed to "
                               "schedule timer for ThreadStatusManager::Updater\n"));
       }
     }
--- a/dds/DCPS/ThreadStatusManager.cpp
+++ b/dds/DCPS/ThreadStatusManager.cpp
@@ -178,7 +178,7 @@ void ThreadStatusManager::ThreadContainer::update(
   const bool update_thread_details = manager_info_.thread_status_interval();
   if (!update_thread_details && !finished) {
     // We don't need to continue if we have a listener and we're not publishing
-    // the thead status topic and the thread is not finished. For the listener
+    // the thread status topic and the thread is not finished. For the listener
     // we are only waiting for the thread to finish.
     return;
   }


dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -131,10 +157,14 @@ void ThreadStatusManager::update_i(Thread::ThreadStatus status, bool finished,
...
   if (pos != map_.end()) {
-    pos->second.update(m_now, s_now, status, bucket_limit_, nested, detail1, detail2);
+    Thread& thread = pos->second;
+    thread.update(m_now, s_now, status, bucket_limit_, nested, detail1, detail2);
     if (finished) {
-      finished_.push_back(pos->second);
+      finished_.push_back(thread);
       map_.erase(pos);
+      if (thread_status_listener_) {
+        thread_status_listener_->on_thread_finished(thread.info());
```
sonndinh: This `Thread` entry has been deleted from `map_` at line 164, so accessing it here will cause issue.

simpsont-oci:
```suggestion
        thread_status_listener_->on_thread_finished(finished_.back());
```

dds/DCPS/ThreadStatusManager.h
```diff
@@ -98,9 +111,11 @@ class OpenDDS_Dcps_Export ThreadStatusManager {
 
   bool update_thread_status() const
   {
-    return thread_status_interval_ > TimeDuration::zero_value;
+    return thread_status_interval_ > TimeDuration::zero_value || thread_status_listener_;
```
sonndinh: If I remember correctly, some period task or event currently is set up based on the return from `update_thread_status()` so that the task/event is created when `thread_status_interval_` is greater than zero. With this change, can it run to a case where `thread_status_interval_` is zero but `thread_status_listener_` exists, and so the task/event may be created but doesn't have a valid interval to schedule timers on?

iguessthislldo: It also needs the ["harvest thread" QoS property](https://opendds.readthedocs.io/en/master/devguide/built_in_topics.html#openddsinternalthread-topic) enabled to start the periodic task. I think I tried to explain this when I talked about the changes in standup originally a while ago, but probably did so poorly. I also was only vaguely aware that's what was stopping it though. Looking at it in detail this is the code that controls it:

First it needs to init `thread_status_event_` in `Spdp::SpdpTransport::open()`:

https://github.com/OpenDDS/OpenDDS/blob/567fc54bc0a9d710fa82d887b61eaf032ef9a90d/dds/DCPS/RTPS/Spdp.cpp#L2546-L2548

https://github.com/OpenDDS/OpenDDS/blob/567fc54bc0a9d710fa82d887b61eaf032ef9a90d/dds/DCPS/RTPS/Spdp.cpp#L2561-L2573

Then it needs to enable it in `Spdp::SpdpTransport::enable_periodic_tasks()`:

https://github.com/OpenDDS/OpenDDS/blob/567fc54bc0a9d710fa82d887b61eaf032ef9a90d/dds/DCPS/RTPS/Spdp.cpp#L2661-L2664

So it needs `harvest_thread_status_` enabled at each point. I think it will do something like what you describe if you enabled harvest status, used my callbacks, but left status interval disabled right now. I want to say that the periodic task wouldn't fire, but I'm not 100% sure about that. There should probably be a function that checks both the interval and harvest status config. I agree it shouldn't be using `update_thread_status`, if that's what you're thinking. Ultimately this is a different operation from updating the threads, especially with my changes.

As a side note there's also another path to `init_thread_status_task` through a config topic:

https://github.com/OpenDDS/OpenDDS/blob/567fc54bc0a9d710fa82d887b61eaf032ef9a90d/dds/DCPS/RTPS/Spdp.cpp#L3739-L3748

It mostly looks okay as is, because it's only depending on `init_thread_status_task`. The one thing I did notice about it is it's not changing the interval in the `ThreadStatusManager`, I don't know how important that is though.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3737,15 +3756,8 @@ void Spdp::SpdpTransport::on_data_available(DCPS::ConfigReader_rch)
...
-      init_thread_status_event();
-      if (thread_status_event_) {
-        if (val) {
-          thread_status_event_->enable(val);
-        } else {
-          thread_status_event_->disable();
-        }
-      }
+      init_thread_status_event(TimeDuration(std::atoi(sample.value().c_str())));
+      enable_disable_thread_status_event();
```
sonndinh: New intervals will be set for the `ThreadStatusManager` object through `Service_Participant` (https://github.com/OpenDDS/OpenDDS/blob/master/dds/DCPS/Service_Participant.cpp#L2393-L2394). I think not having to store and set `harvest_thread_status_interval_` here would simplify functions such as `init_thread_status_event`.

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -130,35 +150,102 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
...
 {
-  if (!update_thread_status()) {
-    return;
+  for (size_t i = 0; i < NUM_CONTAINERS; ++i) {
+    containers_[i].set_manager_info(copy);
   }
+}
 
+void ThreadStatusManager::add_thread(const String& name)
+{
```
sonndinh, line 246: `ThreadContainer::add_thread` does check for the return from `update_thread_status`, but could this retain the early check for that?

This method is only called at most once per thread, so probably this is okay.

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -138,20 +138,8 @@ bool read_status(DDS::WaitSet_var ws, InternalThreadBuiltinTopicDataDataReader_v
   return found;
 }
 
-int ACE_TMAIN(int argc, ACE_TCHAR* argv[])
+int tms_test(InternalThreadBuiltinTopicDataDataReader_var itbtd_reader, DDS::WaitSet_var ws)
```
sonndinh:
```suggestion
int tsm_test(InternalThreadBuiltinTopicDataDataReader_var itbtd_reader, DDS::WaitSet_var ws)
```

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -130,35 +150,102 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
...
+}
+
+void ThreadStatusManager::ThreadContainer::update(
+  Thread::ThreadStatus status, bool finished, bool nested, int detail1, int detail2,
+  const MonotonicTimePoint& m_now, const SystemTimePoint& s_now, const ThreadId& thread_id)
+{
+  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  if (!manager_info_.update_thread_status()) {
+    return;
+  }
```
sonndinh, line 176: This will allow a thread to report its status even if the `thread_status_interval_` is zero (but `thread_status_listener_` was set). But the thread statuses won't be harvested since the interval is zero. What would be an use case for that?

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -130,35 +150,102 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
...
+  const Map::iterator pos = map_.find(thread_id);
+  const bool use_finished = manager_info_.thread_status_interval();
+  if (pos != map_.end()) {
+    Thread& thread = pos->second;
+    thread.update(m_now, s_now, status, manager_info_.bucket_limit(), nested, detail1, detail2);
+    if (finished) {
+      manager_info_.on_thread_finished(thread);
+      if (use_finished) {
+        finished_.push_back(thread);
+      }
```
sonndinh: Why does this need to be conditional on `use_finished`?

iguessthislldo: It doesn't, but if we're not publishing to the thread status topic, it doesn't need to use the time or space to track them.

dds/DCPS/ThreadStatusManager.cpp
iguessthislldo: Harvesting is only required for the thread status topic. The whole use case for this is for when you don't care about the thread status topic at all, but just want to know when OpenDDS starts and is done with a thread.

dds/DCPS/RTPS/Spdp.cpp
iguessthislldo: I'm not fully sure what you mean in the first part. It's true, but it doesn't have much to do with what's happening here.

I do agree `harvest_thread_status_interval_` isn't great though... I think what it can do run `init_thread_status_event` run once and have it only use `harvest_thread_status_` since it doesn't change. It was using interval before, but it doesn't really need it.

iguessthislldo: I shouldn't say it doesn't have much with what's happening here I guess, it's part of the same overall operation, but they're working independently.

dds/DCPS/ThreadStatusManager.cpp
iguessthislldo: Sure

dds/DCPS/ThreadStatusManager.cpp
iguessthislldo: A practical use case for this is to call https://man7.org/linux/man-pages/man3/pthread_setaffinity_np.3.html in `on_thread_started` to allow the user to assign threads to dedicated cores so they don't have to experience context switching.

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -130,35 +150,102 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
...
+  const Thread thread(name, thread_id, ACE_OS::thr_self());
 
   if (DCPS_debug_level > 4) {
     ACE_DEBUG((LM_DEBUG, "(%P|%t) ThreadStatusManager::add_thread: "
-               "adding thread %C\n", bit_key.c_str()));
+               "adding thread %C\n", thread.bit_key().c_str()));
+  }
+
+  if (!update_thread_status()) {
+    return;
```
sonndinh: What I meant in a previous comment was, in the current code, this check was at the beginning of this function so in case `update_thread_status()` returns false, it could return earlier and avoid executing `get_thread_id()`, constructing the `thread` object, etc.

dds/DCPS/ThreadStatusManager.cpp
sonndinh: Right. I guess I was wondering in the cases where we only care about the start and finish events, it would be good to skip all other events so that the thread doesn't incur overhead for the events it doesn't want to record. Current master branch is doing that with a check on the enable/disable state (https://github.com/OpenDDS/OpenDDS/blob/master/dds/DCPS/ThreadStatusManager.cpp#L174-L176).

Maybe in case the interval is zero (harvesting is not enabled) but the listener is set, it can call `on_thread_finished` and return immediately and skip the rest of the function?

dds/DCPS/ThreadStatusManager.cpp
sonndinh: Okay. I think this aligns with my other comment on this function about skipping status updates if it's not publishing to the topic. This checks is helpful for `finished_` but we can skip the whole update with an earlier check.

iguessthislldo: You mean skip calling `thread.update` if the interval is zero? I suppose yes, there is no point in updating the threads. It could go further and not add them at all.

iguessthislldo: Nevermind, that was dumb, I started to code that and realized `on_thread_finished` needs the thread information from the map. Yes, it still doesn't need to update the thread.

dds/DCPS/ThreadStatusManager.cpp
iguessthislldo: No, I mentioned this below, we need the thread object from the map to pass to `on_thread_finished` and we still need to remove the thread object from the map. We also need to clear the `finished_` in case the thread static interval was turned off mid run but the listener is still being used, but I'm thinking this could probably be done when the interval is set to zero because if the listener is null, then `finished_` is never cleared.

dds/DCPS/ThreadStatusManager.cpp
iguessthislldo: I guess we can skip if `thread_status_interval()` is zero and `finished` is false though.

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -130,35 +144,115 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
...
+{
+  ACE_GUARD(ACE_Thread_Mutex, g, mutex_);
+  if (!manager_info_.update_thread_status()) {
+    return;
+  }
+
+  const bool update_thread_details = manager_info_.thread_status_interval();
+  if (!update_thread_details && !finished) {
+    // We don't need to continue if we have a listener and we're not publishing
+    // the thead status topic and the thread is not finished. For the listener
```
simpsont-oci:
```suggestion
    // the thread status topic and the thread is not finished. For the listener
```

dds/DCPS/ReactorTask.cpp
simpsont-oci:
```suggestion
        ACE_ERROR((LM_ERROR, "(%P|%t) NOTICE: ReactorTask::svc: failed to "
```

PR #5172, opened 2026-02-10
Reduce contention on `ThreadStatusManager`

sonndinh:
- Problem
OpenDDS threads check in with the global `ThreadStatusManager` object to record their activities. `ThreadStatusManager` saves the thread statuses in an internal store protected by a lock. Thus, threads contend for this single lock, and this can hinder their progress.

- Solution
Partition the internal store of `ThreadStatusManager` into multiple slots, each has its own lock. Only threads assigned to the same slot will contend for the lock associated with that slot.

### DIFF
--- a/dds/DCPS/Hash.h
+++ b/dds/DCPS/Hash.h
@@ -43,7 +43,7 @@ inline uint32_t one_at_a_time_hash(const uint8_t* key, size_t length, uint32_t s
 
 inline unsigned fnv_1a_hash(const unsigned char* data, size_t length)
 {
-  // FNV-1a hash method
+  // Refer to https://en.wikipedia.org/wiki/Fowler-Noll-Vo_hash_function for more information.
   unsigned hash = 2166136261u;
   for (size_t i = 0; i < length; ++i) {
     hash ^= data[i];
--- a/dds/DCPS/ThreadStatusManager.cpp
+++ b/dds/DCPS/ThreadStatusManager.cpp
@@ -93,7 +93,7 @@ void ThreadStatusManager::thread_status_interval(const TimeDuration& thread_stat
     ACE_GUARD(ACE_Thread_Mutex, g, lock_);
     thread_status_interval_ = thread_status_interval;
     bucket_limit = bucket_limit_ = thread_status_interval / static_cast<double>(Thread::BUCKET_COUNT);
-    enabled = enabled_ = thread_status_interval_ > TimeDuration::zero_value ? true : false;
+    enabled = enabled_ = thread_status_interval_;
   }
 
   for (size_t i = 0; i < NUM_CONTAINERS; ++i) {
@@ -130,14 +130,14 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
 #endif
 }
 
-size_t ThreadStatusManager::get_container(ThreadId tid)
+ThreadStatusManager::ThreadContainer& ThreadStatusManager::get_container(ThreadId tid)
 {
 #if defined (ACE_WIN32) || defined (ACE_HAS_GETTID)
-  return static_cast<size_t>(tid) % NUM_CONTAINERS;
+  return containers_[static_cast<size_t>(tid) % NUM_CONTAINERS];
 #else
   const unsigned char* data = reinterpret_cast<const unsigned char*>(tid.c_str());
   const unsigned hash = fnv_1a_hash(data, tid.length());
-  return static_cast<size_t>(hash) % NUM_CONTAINERS;
+  return containers_[static_cast<size_t>(hash) % NUM_CONTAINERS];
 #endif
 }
 
@@ -156,7 +156,7 @@ void ThreadStatusManager::add_thread(const String& name)
                "adding thread %C\n", bit_key.c_str()));
   }
 
-  ThreadContainer& container = containers_[get_container(thread_id)];
+  ThreadContainer& container = get_container(thread_id);
   ACE_GUARD(ACE_Thread_Mutex, g, container.mutex_);
   container.map_.insert(std::make_pair(thread_id, Thread(bit_key)));
 }
@@ -168,7 +168,7 @@ void ThreadStatusManager::update_i(Thread::ThreadStatus status, bool finished,
   const SystemTimePoint s_now = SystemTimePoint::now();
   const ThreadId thread_id = get_thread_id();
 
-  ThreadContainer& container = containers_[get_container(thread_id)];
+  ThreadContainer& container = get_container(thread_id);
   ACE_GUARD(ACE_Thread_Mutex, g, container.mutex_);
 
   if (!container.enabled_) {
--- a/dds/DCPS/ThreadStatusManager.h
+++ b/dds/DCPS/ThreadStatusManager.h
@@ -180,7 +180,23 @@ public:
 
 private:
   static ThreadId get_thread_id();
-  size_t get_container(ThreadId tid);
+
+  struct ThreadContainer {
+    ThreadContainer() : enabled_(false) {}
+
+    // Store copies of these from the parent to avoid needing to access it.
+    TimeDuration thread_status_interval_;
+    TimeDuration bucket_limit_;
+    bool enabled_;
+
+    Map map_;
+    List finished_;
+    mutable ACE_Thread_Mutex mutex_;
+
+    void cleanup(const MonotonicTimePoint& now);
+  };
+
+  ThreadContainer& get_container(ThreadId tid);
   void add_thread(const String& name);
 
   void update_i(Thread::ThreadStatus status, bool finished = false,
@@ -212,21 +228,6 @@ private:
   mutable ACE_Thread_Mutex lock_;
 
   static const size_t NUM_CONTAINERS = 11;
-
-  struct ThreadContainer {
-    ThreadContainer() : enabled_(false) {}
-
-    // Store copies of these from the parent to avoid needing to access it.
-    TimeDuration thread_status_interval_;
-    TimeDuration bucket_limit_;
-    bool enabled_;
-
-    Map map_;
-    List finished_;
-    mutable ACE_Thread_Mutex mutex_;
-
-    void cleanup(const MonotonicTimePoint& now);
-  };
   ThreadContainer containers_[NUM_CONTAINERS];
 };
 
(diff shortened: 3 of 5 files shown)


dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -97,6 +129,22 @@ ThreadStatusManager::ThreadId ThreadStatusManager::get_thread_id()
 #endif
 }
 
+size_t ThreadStatusManager::get_container(ThreadId tid)
+{
+#if defined (ACE_WIN32) || defined (ACE_HAS_GETTID)
+  return static_cast<size_t>(tid) % NUM_CONTAINERS;
+#else
+  const unsigned char* data = reinterpret_cast<const unsigned char*>(tid.c_str());
+  // FNV-1a hash method
```
simpsont-oci: Put this functionality in `dds/DCPS/Hash.h`?

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -112,57 +160,60 @@ void ThreadStatusManager::add_thread(const String& name)
...
 void ThreadStatusManager::harvest(const MonotonicTimePoint& start,
                                   ThreadStatusManager::List& running,
                                   ThreadStatusManager::List& finished) const
 {
-  ACE_GUARD(ACE_Thread_Mutex, g, lock_);
-
-  for (Map::const_iterator pos = map_.begin(), limit = map_.end(); pos != limit; ++pos) {
-    if (pos->second.last_update() > start) {
-      running.push_back(pos->second);
+  for (size_t i = 0; i < NUM_CONTAINERS; ++i) {
```
simpsont-oci, line 193: This will change the order of the reported threads, but maybe that doesn't matter? Will it make the logs more confusing to review? If so, you could potentially also change the return types from a list to a set which would sort the results for you.

dds/DCPS/ThreadStatusManager.h
```diff
@@ -183,33 +174,54 @@ class OpenDDS_Dcps_Export ThreadStatusManager {
...
-  Map map_;
-  List finished_;
-
+  bool enabled_;
   mutable ACE_Thread_Mutex lock_;
+
+  static const size_t NUM_CONTAINERS = 11;
+
+  struct ThreadContainer {
+    ThreadContainer() {}
```
simpsont-oci: Do you need to initialize `enabled_` in the default constructor?
```suggestion
    ThreadContainer() : enabled_(false) {}
```

dds/DCPS/ThreadStatusManager.h
```diff
@@ -183,33 +174,54 @@ class OpenDDS_Dcps_Export ThreadStatusManager {
...
+
+    // Store copies of these from the parent to avoid needing to access it.
+    TimeDuration thread_status_interval_;
+    TimeDuration bucket_limit_;
+    bool enabled_;
+
+    Map map_;
```
simpsont-oci: Now that harvested results aren't in order... maybe we should be using OPENDDS_UNORDERED_MAP as the type for `Map` when it's available (ACE_HAS_CPP11)?

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -0,0 +1,256 @@
...
+  String id() const { return id_; }
+
+private:
+  int svc()
+  {
+    ThreadStatusManager& tsm = TheServiceParticipant->get_thread_status_manager();
+
+    // For each event, wait for the main thread to signal to proceed
+    ThreadStatusManager::Start start(tsm, id_);
+    if (cv_.wait() != 0) {
```
simpsont-oci, line 39: This thread needs to lock the ConditionVariable's mutex before calling wait()... so you'll likely want to pass in the mutex to the constructor as well so you have a reference to it and can lock it in `svc()`.

dds/DCPS/ThreadStatusManager.h
sonndinh: You're spot on -- there is an ubsan test error because it wasn't initialized.

dds/DCPS/ThreadStatusManager.cpp
sonndinh: I think the order doesn't matter. The only usage of thread status that I'm aware of is in RtpsRelay which checks for the presence of each thread's status. There is an optional log for thread status but they are ordered using thread Ids, so it doesn't matter much.

dds/DCPS/ThreadStatusManager.h
sonndinh: I updated to make it conditional on ACE_HAS_CPP11.

tests/DCPS/ThreadStatusManager/main.cpp
sonndinh: That's right. A tsan build also was able to catch this. I updated accordingly.

dds/DCPS/Hash.h
```diff
@@ -41,6 +41,17 @@ inline uint32_t one_at_a_time_hash(const uint8_t* key, size_t length, uint32_t s
 }
 #endif
 
+inline unsigned fnv_1a_hash(const unsigned char* data, size_t length)
+{
+  // FNV-1a hash method
```
iguessthislldo: I'm not sure this comment is adding much since it's just the name of the function. Maybe consider removing it or turning it into a documenting comment and linking to something like the [wikipedia page](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) for it.

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  // The thread status sample corresponding to the thread end event is a disposed sample,
+  // so thread_end is just passed here, but its value is not used for anything.
+  if (!read_status(ws, itbtd_reader, thread_end, task, true)) {
+    ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main - failed to read test thread's end status\n"));
+    return EXIT_FAILURE;
+  }
+
+  // Check that the timestamps are in the correct order
+  bool ret = EXIT_SUCCESS;
```
iguessthislldo:
```suggestion
  int ret = EXIT_SUCCESS;
```

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -0,0 +1,259 @@
...
+    case DDS::ALIVE_INSTANCE_STATE:
+      return "ALIVE";
+    case DDS::NOT_ALIVE_DISPOSED_INSTANCE_STATE:
+      return "NOT_ALIVE_DISPOSED";
+    case DDS::NOT_ALIVE_NO_WRITERS_INSTANCE_STATE:
+      return "NOT_ALIVE_NO_WRITERS";
+    default:
+      return "UNKNOWN";
+  }
+}
```
iguessthislldo: I wish this was in a better place, but there is a static method you can call to do this: https://github.com/OpenDDS/OpenDDS/blob/934b71affc48fb56087b413397823e2d6d4d7e96/dds/DCPS/InstanceState.h#L151-L152

tests/DCPS/ThreadStatusManager/ThreadStatusManager.mpc
```diff
@@ -0,0 +1,5 @@
+project: dcpsexe, dcps_test, dcps_rtps_udp {
+  exename = threadstatusmanager
+  requires += built_in_topics
+  idlflags += -SS
```
iguessthislldo:
```suggestion
```

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -0,0 +1,259 @@
+#include <dds/DCPS/ThreadStatusManager.h>
+#include <dds/DCPS/RTPS/RtpsDiscovery.h>
+#include <dds/DCPS/Qos_Helper.h>
+#include <dds/DCPS/WaitSet.h>
+
+#include <ace/Task.h>
+
+#ifndef DDS_HAS_MINIMUM_BIT
```
iguessthislldo: If you have `requires += built_in_topics`, I don't really see the point of having this. Also if it was defined, the main would go away and cause a linking error.

dds/DCPS/ThreadStatusManager.cpp
```diff
@@ -84,6 +85,38 @@ double ThreadStatusManager::Thread::utilization(const MonotonicTimePoint& now) c
...
 
+void ThreadStatusManager::thread_status_interval(const TimeDuration& thread_status_interval)
+{
+  TimeDuration bucket_limit;
+  bool enabled;
+  {
+    ACE_GUARD(ACE_Thread_Mutex, g, lock_);
+    thread_status_interval_ = thread_status_interval;
+    bucket_limit = bucket_limit_ = thread_status_interval / static_cast<double>(Thread::BUCKET_COUNT);
+    enabled = enabled_ = thread_status_interval_ > TimeDuration::zero_value ? true : false;
```
iguessthislldo: TimeDuration inherits from SafeBool

```suggestion
    enabled = enabled_ = thread_status_interval_;
```

tests/DCPS/ThreadStatusManager/main.cpp
```diff
@@ -0,0 +1,259 @@
...
+  ws->attach_condition(read_cond);
+
+  MonotonicTimePoint thread_start, event1_start, event1_end, event2_start, event2_end, thread_end;
+
+  // Notify the test thread to proceed after reading the thread status in each step
+  if (!read_status(ws, itbtd_reader, thread_start, task)) {
+    ACE_ERROR((LM_ERROR, "(%P|%t) ERROR: main - failed to read test thread's start status\n"));
+    return EXIT_FAILURE;
+  }
+  cv.broadcast();
```
iguessthislldo: Move the error logging and broadcast into `read_status`?

dds/DCPS/ThreadStatusManager.h
```diff
@@ -183,33 +180,54 @@ class OpenDDS_Dcps_Export ThreadStatusManager {
 
 private:
   static ThreadId get_thread_id();
+  size_t get_container(ThreadId tid);
```
iguessthislldo: Make static?

sonndinh: Is there a reason that you want to make this a static function?

iguessthislldo: Not particularity, only that it could be static (or const). A different approach would be maybe to make actually return the container. It would then better fit the name.

```suggestion
  ThreadContainer& get_container(ThreadId tid);
```

sonndinh: Yea, that sounds good to me.

PR #5173, opened 2026-02-10
Deserialize primitive sequences without zero initization

iguessthislldo:
When sequences of primitives are deserialized, the generated code calls length before the elements are read in. The issue with length is it initializes the elements. This is wasted effort for primitives in this case because we read the number of elements from the CDR stream so unless the stream ends we know we're setting every element afterwards. In samples with large number of primitives setting the length (ultimately calling std::fill in TAO) can take up a significant amount of time.

### DIFF
--- a/dds/DCPS/ResizeSeqNoInit.h
+++ b/dds/DCPS/ResizeSeqNoInit.h
@@ -38,7 +38,7 @@ namespace SafetyProfile {
 
 namespace DCPS {
 
-bool OpenDDS_Dcps_Export get_init_in_optional_init_allocator();
+bool OpenDDS_Dcps_Export get_optional_init_allocator_must_init();
 
 /// Resize unbounded FACE sequences without zero-initializing new elements.
 template <typename T, typename Bounds, typename Elts>
@@ -81,11 +81,11 @@ void resize_bounded_seq_no_init(
 }
 
 #ifdef ACE_HAS_CPP11
-struct OpenDDS_Dcps_Export InitInOptionalInitAllocator {
+struct OpenDDS_Dcps_Export OptionalInitAllocatorScopedSetter {
   const bool prev_value;
 
-  explicit InitInOptionalInitAllocator(bool value);
-  ~InitInOptionalInitAllocator();
+  explicit OptionalInitAllocatorScopedSetter(bool value);
+  ~OptionalInitAllocatorScopedSetter();
 };
 
 template <typename T, typename Alloc = std::allocator<T>>
@@ -105,7 +105,7 @@ public:
   template <typename U>
   void construct(U* ptr) noexcept(std::is_nothrow_default_constructible<U>::value)
   {
-    if (get_init_in_optional_init_allocator()) {
+    if (get_optional_init_allocator_must_init()) {
       ::new (static_cast<void*>(ptr)) U();
     } else {
       ::new (static_cast<void*>(ptr)) U;
@@ -132,10 +132,10 @@ void resize_unbounded_seq_no_init(std::vector<T, Alloc>& vec, typename std::vect
   // of elements because std::vector itself doesn't give a way to avoid
   // initialization. This will almost certainly be slower than a vanilla
   // std::vector unless optimization is enabled. Here
-  // InitInOptionalInitAllocator sets a global variable that
+  // OptionalInitAllocatorScopedSetter sets a global variable that
   // OptionalInitAllocator checks to see if elements should be initialized or
   // not.
-  InitInOptionalInitAllocator skip_init(false);
+  OptionalInitAllocatorScopedSetter skip_init(false);
   vec.resize(new_length);
 }
 
--- a/dds/DCPS/ResizeSeqNoInit.cpp
+++ b/dds/DCPS/ResizeSeqNoInit.cpp
@@ -14,22 +14,22 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
-thread_local bool init_in_optional_init_allocator = true;
+thread_local bool optional_init_allocator_must_init = true;
 
-OpenDDS_Dcps_Export bool get_init_in_optional_init_allocator()
+OpenDDS_Dcps_Export bool get_optional_init_allocator_must_init()
 {
-  return init_in_optional_init_allocator;
+  return optional_init_allocator_must_init;
 }
 
-InitInOptionalInitAllocator::InitInOptionalInitAllocator(bool value)
-: prev_value(init_in_optional_init_allocator)
+OptionalInitAllocatorScopedSetter::OptionalInitAllocatorScopedSetter(bool value)
+: prev_value(optional_init_allocator_must_init)
 {
-  init_in_optional_init_allocator = value;
+  optional_init_allocator_must_init = value;
 }
 
-InitInOptionalInitAllocator::~InitInOptionalInitAllocator()
+OptionalInitAllocatorScopedSetter::~OptionalInitAllocatorScopedSetter()
 {
-  init_in_optional_init_allocator = prev_value;
+  optional_init_allocator_must_init = prev_value;
 }
 
 } // namespace DCPS


dds/DCPS/ResizeSeqNoInit.h
```diff
@@ -0,0 +1,200 @@
...
+  typename OpenDDS::SafetyProfile::Sequence<T, Bounds, Elts>::size_type new_length)
+{
+  resize_unbounded_seq_no_init(seq, new_length);
+}
+
+#ifdef ACE_HAS_CPP11
+struct OpenDDS_Dcps_Export InitInOptionalInitAllocator {
+  const bool prev_value;
+
+  InitInOptionalInitAllocator(bool value);
```
mitza-oci:
```suggestion
  explicit InitInOptionalInitAllocator(bool value);
```

dds/DCPS/ResizeSeqNoInit.h
```diff
@@ -0,0 +1,200 @@
...
+  seq.length(new_length);
+}
+
+/// Resize bounded FACE sequences without zero-initializing new elements.
+template <typename T, typename Bounds, typename Elts>
+void resize_bounded_seq_no_init(
+  OpenDDS::FaceTypes::Sequence<T, Bounds, Elts>& seq,
+  typename OpenDDS::FaceTypes::Sequence<T, Bounds, Elts>::size_type new_length)
+{
+  resize_unbounded_seq_no_init(seq, new_length);
```
mitza-oci, line 61: Why does bounded call unbounded?  Maybe add a comment?

dds/DCPS/ResizeSeqNoInit.h
```diff
@@ -0,0 +1,200 @@
...
+
+template <typename T, typename Alloc = std::allocator<T>>
+class OptionalInitAllocator : public Alloc {
+protected:
+  using Traits = std::allocator_traits<Alloc>;
+
+public:
+  template<typename U>
+  struct rebind {
+      using other = OptionalInitAllocator<U, typename Traits::template rebind_alloc<U>>;
```
mitza-oci:
```suggestion
    using other = OptionalInitAllocator<U, typename Traits::template rebind_alloc<U>>;
```

dds/DCPS/ResizeSeqNoInit.h
```diff
@@ -0,0 +1,200 @@
...
+};
+
+template <typename T>
+using OptionalInitVector = std::vector<T, OptionalInitAllocator<T>>;
+
+/// Try to resize unbounded std::vector without zero-initializing new elements.
+template <typename T, typename Alloc>
+void resize_unbounded_seq_no_init(std::vector<T, Alloc>& vec, typename std::vector<T, Alloc>::size_type new_length)
+{
+  // This is the most painful version of resize_seq_no_init. The std::vector
```
mitza-oci: Clarify what's painful and why?  Or maybe just reduce this comment be more succinct?

docs/devguide/opendds_idl.rst
```diff
@@ -301,3 +301,47 @@ When using MPC to generate projects, the ``opendds_cxx11`` base project should b
...
+
+.. _opendds_idl--annotations:
+
+Annotations
+===========
+
+..
+    Sect<16.6>
+
+In IDL, annotations are a way of indicating meta-information about a type or part of a type.
```
mitza-oci, line 319: Should this part link to the xtypes section?

tests/dcps_tests.lst
```diff
@@ -88,6 +88,8 @@ tests/DCPS/Compiler/explicit_ints/run_test.pl classic: !DCPS_MIN
 tests/DCPS/Compiler/explicit_ints/run_test.pl cpp11: !DCPS_MIN CXX11 !OPENDDS_SAFETY_PROFILE
 tests/DCPS/Compiler/char_literals/run_test.pl classic: !DCPS_MIN
 tests/DCPS/Compiler/char_literals/run_test.pl cpp11: !DCPS_MIN CXX11 !OPENDDS_SAFETY_PROFILE
+tests/DCPS/Compiler/no_init_before_deserialize/run_test.pl classic: !DCPS_MIN
+tests/DCPS/Compiler/no_init_before_deserialize/run_test.pl cpp11: !DCPS_MIN CXX11 !OPENDDS_SAFETY_PROFILE
```
mitza-oci, line 92: Should this be added to any of the more-specific lst files as well?

dds/DCPS/ResizeSeqNoInit.h
```diff
@@ -0,0 +1,200 @@
...
+template <typename T, typename Bounds, typename Elts>
+void resize_bounded_seq_no_init(
+  OpenDDS::SafetyProfile::Sequence<T, Bounds, Elts>& seq,
+  typename OpenDDS::SafetyProfile::Sequence<T, Bounds, Elts>::size_type new_length)
+{
+  resize_unbounded_seq_no_init(seq, new_length);
+}
+
+#ifdef ACE_HAS_CPP11
+struct OpenDDS_Dcps_Export InitInOptionalInitAllocator {
```
simpsont-oci: I know naming is hard, but `InitInOptionalInitAllocator` seems fairly confusing to me given what this seems to be intended do. Perhaps rename it to something like `ScopedOptionalInitBehaviorSetter`?

docs/devguide/opendds_idl.rst
```diff
@@ -301,3 +301,47 @@ When using MPC to generate projects, the ``opendds_cxx11`` base project should b
...
+
+.. _opendds_idl-opendds-no_init_before_deserialize:
+
+``@OpenDDS::no_init_before_deserialize``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Applies to: sequence typedefs, anonymous sequence struct members, anonymous sequence union members (only when :ref:`opendds_idl--using-the-idl-to-c-11-mapping`)
+
+When deserializing a very long sequence of primitive values, a significant portion of time could be spent zero-initializing the memory of the expanded receiving type before inserting the actual values.
+The sequences other mappings can avoid doing this without any user intervention, but the :ref:`IDL-to-C++11 mapping <opendds_idl--using-the-idl-to-c-11-mapping>` requires a special ``std::vector`` type to do this.
```
sonndinh: Does this mean to say "The other sequence mappings can avoid ..."?

docs/devguide/opendds_idl.rst
```diff
@@ -301,3 +301,47 @@ When using MPC to generate projects, the ``opendds_cxx11`` base project should b
...
+This type will not be assignable with normal ``std::vector`` types because it's using a custom allocator.
+In addition, it should be noted that builds using the IDL-to-C++11 mapping must be optimized to get any benefit from this, otherwise the deserialization will probably be slower.
```
sonndinh: This sounds like an important note for using this. It would be nice to have it highlighted.

docs/news.d/no-init-primitive-seq.rst
```diff
@@ -0,0 +1,9 @@
+.. news-prs: 5173
+
+.. news-start-section: Additions
+
+- As an optimization, In all IDL mappings OpenDDS will now try to avoid zero-initializing memory of sequences of primitive types while deserializing a sample.
```
sonndinh:
```suggestion
- As an optimization, in all IDL mappings OpenDDS will now try to avoid zero-initializing memory of sequences of primitive types while deserializing a sample.
```

dds/DCPS/ResizeSeqNoInit.h
iguessthislldo: Maybe annoying is a better way to phrase it? I think the comment is already pretty clear, having to change the type of `std::vector` is very annoying.

docs/devguide/opendds_idl.rst
iguessthislldo: I want to move all the annotation from XTypes here, but didn't want to do it in this PR.

dds/DCPS/ResizeSeqNoInit.h
iguessthislldo: It's supposed to be "init in `OptionalInitAllocator`", but yeah I agree it's not great, especially since it might imply it's an allocator itself. Maybe  `OptionalInitAllocatorScopedSetter`?

docs/devguide/opendds_idl.rst
iguessthislldo: I might have just accidentally left out "of": "The sequences of other mappings ...". However rereading it, I gave room to FACE/Safety profile mappings when they are not even mentioned, have always worked like this, and are not used very much. I rewrote it to just say "While the classic TAO mapping can avoid this....".

docs/news.d/no-init-primitive-seq.rst
iguessthislldo: I decided to take the "in" out.

docs/devguide/opendds_idl.rst
iguessthislldo: Using the `important` directive to highlight it

dds/DCPS/ResizeSeqNoInit.h
simpsont-oci: That works too.

sonndinh (approved):

PR #5174, opened 2026-02-10
Add move constructors and assignment operators to improve fragment reassembly

simpsont-oci:
Problem:
 - For data samples with large numbers of fragments, reassembly often is rebuilding temporary MessageBlock objects, forcing calls to ACE_Data_Block::duplicate()

Solution:
 - For C++11 builds, we can use move constructors and assignment operators to improve the behavior for these temporary objects

### DIFF
--- a/dds/DCPS/transport/framework/TransportReassembly.cpp
+++ b/dds/DCPS/transport/framework/TransportReassembly.cpp
@@ -27,11 +27,20 @@ FragKey::FragKey(const GUID_t& pubId,
 GUID_tKeyLessThan FragKey::compare_;
 
 TransportReassembly::FragSample::FragSample(const FragmentRange& fragRange,
-                                          const ReceivedDataSample& data)
+                                            const ReceivedDataSample& data)
   : frag_range_(fragRange)
   , rec_ds_(data)
 {
 }
+
+#ifdef ACE_HAS_CPP11
+TransportReassembly::FragSample::FragSample(const FragmentRange& fragRange,
+                                            ReceivedDataSample&& data)
+  : frag_range_(fragRange)
+  , rec_ds_(std::move(data))
+{
+}
+#endif
 
 TransportReassembly::TransportReassembly(const TimeDuration& timeout)
   : timeout_(timeout)
@@ -375,6 +384,13 @@ namespace {
   }
 }
 
+#ifdef ACE_HAS_CPP11
+#define OPENDDS_MAYBE_MOVE std::move
+#else
+#define OPENDDS_MAYBE_MOVE
+#endif
+
+
 bool
 TransportReassembly::FragInfo::insert(const FragmentRange& fragRange, ReceivedDataSample& data)
 {
@@ -403,7 +419,7 @@ TransportReassembly::FragInfo::insert(const FragmentRange& fragRange, ReceivedDa
     FragSample& fr = *it;
     if (next < fr.frag_range_.first) {
       // insert before 'it'
-      sample_finder_[fragRange.second] = sample_list_.insert(it, FragSample(fragRange, data));
+      sample_finder_[fragRange.second] = sample_list_.insert(it, FragSample(fragRange, OPENDDS_MAYBE_MOVE(data)));
       data.clear();
       VDBG((LM_DEBUG, "(%P|%t) TransportReassembly::insert: (SN: %q) inserted %q-%q on the left of %q-%q\n", sn, fragRange.first, fragRange.second, fr.frag_range_.first, fr.frag_range_.second));
       return true;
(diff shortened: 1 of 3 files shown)


dds/DCPS/transport/framework/ReceivedDataSample.cpp
```diff
@@ -32,6 +32,42 @@ ReceivedDataSample::ReceivedDataSample(const ACE_Message_Block& payload)
...
+    header_ = rhs.header_;
+    fragment_size_ = rhs.fragment_size_;
+    blocks_ = rhs.blocks_;
+  }
+  return *this;
+}
+
+#if defined (ACE_HAS_CPP11)
+ReceivedDataSample::ReceivedDataSample(ReceivedDataSample&& other)
+  : header_(std::move(other.header_))
```
jwillemsen: Can’t you use default?

simpsont-oci: I can!

dds/DCPS/transport/framework/TransportReassembly.cpp
```diff
@@ -33,6 +33,30 @@ TransportReassembly::FragSample::FragSample(const FragmentRange& fragRange,
...
+  if (this != &rhs) {
+    frag_range_ = rhs.frag_range_;
+    rec_ds_ = rhs.rec_ds_;
+  }
+  return *this;
+}
+
+#ifdef ACE_HAS_CPP11
+TransportReassembly::FragSample::FragSample(const FragmentRange& fragRange,
+                                          ReceivedDataSample&& data)
```
iguessthislldo:
```suggestion
                                            ReceivedDataSample&& data)
```

dds/DCPS/transport/framework/TransportReassembly.cpp
```diff
@@ -403,8 +458,8 @@ TransportReassembly::FragInfo::insert(const FragmentRange& fragRange, ReceivedDa
     FragSample& fr = *it;
     if (next < fr.frag_range_.first) {
       // insert before 'it'
-      sample_finder_[fragRange.second] = sample_list_.insert(it, FragSample(fragRange, data));
-      data.clear();
+      sample_finder_[fragRange.second] = sample_list_.insert(it, FragSample(fragRange, OPENDDS_MAYBE_MOVE(data)));
+      // data.clear();
```
iguessthislldo: Does this need to be dependent on C++03? I would assume we want to clear it for a clean API and so it does the same thing as C++11 code. Or do we not want to bother because it would take a little time  clear it?

dds/DCPS/transport/framework/TransportReassembly.h
```diff
@@ -122,6 +122,17 @@ class OpenDDS_Dcps_Export TransportReassembly : public RcObject {
     FragSample(const FragmentRange& fragRange,
               const ReceivedDataSample& data);
 
+
+    FragSample(const FragSample&);
+    FragSample& operator=(const FragSample&);
+
+#ifdef ACE_HAS_CPP11
+    FragSample(const FragmentRange& fragRange,
+              ReceivedDataSample&& data);
```
iguessthislldo, line 127:
```suggestion
               ReceivedDataSample&& data);
```

dds/DCPS/transport/framework/ReceivedDataSample.cpp
```diff
@@ -32,6 +32,23 @@ ReceivedDataSample::ReceivedDataSample(const ACE_Message_Block& payload)
   } while (amb);
 }
 
+ReceivedDataSample::ReceivedDataSample(const ReceivedDataSample& other)
```
jwillemsen: Couldn't with C++11 and newer you could declare them as default in the header and let the compiler generate the implementation?

simpsont-oci: Ah, yeah, I'd forgotten that I'd had to explicitly add these as well once I added the move versions. I'll make them default as well and remove the implementations.

dds/DCPS/transport/framework/TransportReassembly.cpp
```diff
@@ -366,6 +375,30 @@ TransportReassembly::FragInfo::operator=(const FragInfo& rhs)
   return *this;
 }
 
+#ifdef ACE_HAS_CPP11
+TransportReassembly::FragInfo::FragInfo(FragInfo&& val)
```
jwillemsen: Can't these two be default?

simpsont-oci: Ah, yep. Realized the const& versions were different implementation from the default, but default behavior should be OK for these.

jwillemsen (approved):

PR #5176, opened 2026-02-13
Fix null pointer dereference in DynamicSample when data_ is null

smtbhd32-ABX:
The data_ pointer in DynamicSample is null when deserializing samples from
a participant being removed, causing a null pointer dereference at data_->type().

So, added a check. 
https://github.com/OpenDDS/OpenDDS/issues/5167

### DIFF
--- a/docs/news.d/fix-dynamic-sample-null-ptr.rst
+++ b/docs/news.d/fix-dynamic-sample-null-ptr.rst
@@ -1,8 +1,5 @@
 .. news-prs: 5176
 
 .. news-start-section: Fixes
-- Fixed a null pointer dereference in :ghfile:`dds/DCPS/XTypes/DynamicSample.cpp`
-  and :ghfile:`dds/DCPS/XTypes/DynamicSample.h` when ``data_`` is null in
-  ``DynamicSample::deserialize`` and ``DynamicSample::copy``, which could occur
-  when deserializing samples from a participant being removed.
+- Fixed a null pointer dereference which could occur when deserializing ``DynamicData`` samples from a participant being removed.
 .. news-end-section


dds/DCPS/XTypes/DynamicSample.cpp
```diff
@@ -92,6 +92,13 @@ bool DynamicSample::deserialize(Serializer& ser)
   }
   mb->wr_ptr(len);
 
+  if (!data_) {
```
simpsont-oci, line 95: This looks fine to me. Since the default constructor can always allow `data_` to be initialized to null, we probably ought to have checks in place to prevent de-referencing `data_` when it's null. Along that line of thought, can you please also add a similar check to `DynamicSample::copy` in the header? In that case, I would maybe change the line:
```
DDS::DynamicData_var dd = data_->clone();
```
to something like:
```
DDS::DynamicData_var dd = data_ ? data_->clone() : NULL;
```

smtbhd32-ABX: I have updated it as per your request.

dds/DCPS/XTypes/DynamicSample.h
```diff
@@ -62,7 +62,7 @@ class OpenDDS_Dcps_Export DynamicSample : public DCPS::Sample {
 
   DCPS::Sample_rch copy(DCPS::Sample::Mutability mutability, DCPS::Sample::Extent extent) const
   {
-    DDS::DynamicData_var dd = data_->clone();
+    DDS::DynamicData_var dd = data_ ? data_->clone() : NULL;
```
mitza-oci:
```suggestion
    DDS::DynamicData_var dd = data_ ? data_->clone() : 0;
```

docs/news.d/fix-dynamic-sample-null-ptr.rst
```diff
@@ -0,0 +1,8 @@
+.. news-prs: 5176
+
+.. news-start-section: Fixes
+- Fixed a null pointer dereference in :ghfile:`dds/DCPS/XTypes/DynamicSample.cpp`
+  and :ghfile:`dds/DCPS/XTypes/DynamicSample.h` when ``data_`` is null in
+  ``DynamicSample::deserialize`` and ``DynamicSample::copy``, which could occur
+  when deserializing samples from a participant being removed.
```
iguessthislldo:
```suggestion
- Fixed a null pointer dereference which could occur when deserializing ``DynamicData`` samples from a participant being removed.
```

I think it's ok as is, but there will a link to this PR if someone reading the news wants more information, so there's doesn't have to explain the bug scenario in detail.

smtbhd32-ABX: Sure

PR #5160, opened 2025-12-27
Migrate Reactor-Based Tasks To EventDispatcher

simpsont-oci:
Problems:
 - Transport reactor threads are often performing work not strictly related to the reading in of data. This has potential performance impacts on machines which could be simultaneously doing this other work on other threads.
 - These reactor threads are likewise not particularly well suited to general purpose "jobs" and "tasks" as ACE::Reactor is not safe for re-entrant code.
 - The EventDispatcher threads, which already exist within each transport, are underutilized.
 - We don't have great internal multi-threading capabilities, because the current transport reactors run in a single threaded fashion.

Solution:
 - In addition to the transport EventDispatcher instances, add a global EventDispatcher alongside the global ReactorTask.
 - Switch as many "job" and "task" type operations to use EventDispatcher, rather than JobQueue or reactor "Tasks" (PeriodicTask, SporadicTask, MultiTask).
 - For "Events" which really must run on the reactor thread (e.g. Multicast joins), introduce "ReactorEvent".
 - Since this inherently makes the internals more multi-threaded than previously, as a precautionary measure bump up the potential number of EventDispatcher threads and resolve as many exposed threading issues as possible before putting the default thread count for EventDispatchers back to one.

PR #5180, opened 2026-02-24
ConcurrentAuthLimit test ought to have resend logic

simpsont-oci:
Problem:
- a single UDP send may get lost during testing, causing the test to fail to behave as expected

Solution:
- send at a regular interval while waiting for SPDP processing in order to ensure entities are correctly discovered


simpsont-oci:
This seems to address the issues we're seeing in #5160, as evidenced by this workflow run: https://github.com/OpenDDS/OpenDDS/actions/runs/22391473459 (based on the same branch as #5160, but with these changes).

Issue #4757, opened 2024-08-01
Replace `SafeBool_T` with explicit operator bool


jwillemsen:
With C++11 `SafeBool_T` can be replaced with `explicit operator bool`

PR #5193, opened 2026-03-07
Use `SafeBool_T` for `RcHandle`

sonndinh:
- Avoid implicit conversion from `RcHandle` to types such as integer by making it a `SafeBool_T`
- Better following the safe bool pattern by disallowing comparison operators by default. Derived classes can still define comparison operators as member functions.


sonndinh:
> With C++11 this could be reworked, see #4757

This is for v3 which doesn't have C++11

mitza-oci (approved):

simpsont-oci (approved):

PR #5208, opened 2026-04-13
Add EventDispatcher thread count config flags

### LINKED ISSUE #5183: Add config flag for global EventDispatcher thread count
Currently, the global event dispatcher is hard-coded to use a single thread. This should be user-configurable, with a single thread as the lowest possible value.

### LINKED ISSUE #5184: Add config flag for transport-level EventDispatcher thread count
Currently, the transport-level event dispatcher is hard-coded to use a single thread. This should be user-configurable, per transport instance, with a single thread as the lowest possible value.

### LINKED ISSUE #5185: Add config flag to allow transport-level EventDispatcher to be replaced by global EventDispatcher
Currently, event dispatchers are created at both the global level and per transport instance. For applications which desire to more tightly control the the number of OpenDDS threads, the transports should be allowed to configurably use the global level EventDispatcher rather than creating their own. This way an application with a large number of transport instances could have them all use the global event dispatcher and (pending #5183) control the thread count of the global event dispatcher as desired.

simpsont-oci:
Problems:
 - EventDispatcher thread counts were hard-coded at one for both global and per-transport EventDispatcher instances. This doesn't allow much control over thread creation.

Solution:
 - Add two configuration flags:
   - DCPSEventDispatcherThreads - specify the number of threads created by the global (Service_Participant) EventDispatcher instance. Default and minimum value remains one.
   - event_dispatcher_threads (per transport) - specify the number of threads created by the transport instance's EventDispatcher instance. Defaults to one. A value of zero replaces use of the local per-transport EventDispatcher instance with use of the global (Service_Participant) EventDispatcher instance.

### DIFF
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -1779,6 +1779,8 @@ Service_Participant::event_dispatcher_thread_count() const
                  value,
                  COMMON_DCPS_EVENT_DISPATCHER_THREADS_default));
     }
+    config_store_->set_uint32(COMMON_DCPS_EVENT_DISPATCHER_THREADS,
+                              static_cast<DDS::UInt32>(COMMON_DCPS_EVENT_DISPATCHER_THREADS_default));
     return COMMON_DCPS_EVENT_DISPATCHER_THREADS_default;
   }
 
--- a/dds/DCPS/Service_Participant.cpp
+++ b/dds/DCPS/Service_Participant.cpp
@@ -38,6 +38,8 @@
 #include <ace/Configuration_Import_Export.h>
 #include <ace/Malloc_Allocator.h>
 #include <ace/OS_NS_ctype.h>
+
+#include <limits>
 #include <ace/OS_NS_sys_utsname.h>
 #include <ace/OS_NS_unistd.h>
 #include <ace/Reactor.h>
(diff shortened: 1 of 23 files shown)


dds/DCPS/Service_Participant.cpp
```diff
@@ -1753,6 +1753,36 @@ Service_Participant::liveliness_factor() const
...
+  if (value < COMMON_DCPS_EVENT_DISPATCHER_THREADS_default) {
+    if (log_level >= LogLevel::Warning) {
+      ACE_ERROR((LM_WARNING,
+                 "(%P|%t) WARNING: Service_Participant::event_dispatcher_thread_count: "
+                 "configured value %B is invalid, using %B\n",
+                 value,
+                 COMMON_DCPS_EVENT_DISPATCHER_THREADS_default));
+    }
+    return COMMON_DCPS_EVENT_DISPATCHER_THREADS_default;
+  }
```
sonndinh, line 1781: Should this also set the thread count to the default value so it won't print out the warning again when it's called?

Alternatively, I think this check can be done in `Service_Participant::ConfigReaderListener::on_data_available` so it happens once for each write to the `COMMON_DCPS_EVENT_DISPATCHER_THREADS` key.

docs/devguide/run_time_configuration.rst
```diff
@@ -474,6 +474,12 @@ For example:
 
     See :ref:`config-disc` for details about configuring discovery.
 
+  .. prop:: DCPSEventDispatcherThreads=<n>
+    :default: ``1``
```
sonndinh, line 478: These new config options probably need a news file?

simpsont-oci: Ah, yes. Will do.

dds/DCPS/Service_Participant.cpp
simpsont-oci: I've added it to `Service_Participant::ConfigReaderListener::on_data_available` but will keep this check here in case folks are setting / checking it directly (or via the config store directly). Once I'm sure those changes are fine I'll add a write-back as well to prevent future messages.

sonndinh (approved):

PR #5210, opened 2026-04-14
Improve RtpsSampleHeader Length Checking

simpsont-oci:
Problem:
 - RtpsSampleHeader was not consistently checking whether input message block's rd_ptr() was valid before reading bytes.

Solution:
 - Improve validity checking for deserialization

### DIFF
--- a/dds/DCPS/transport/rtps_udp/RtpsSampleHeader.cpp
+++ b/dds/DCPS/transport/rtps_udp/RtpsSampleHeader.cpp
@@ -160,7 +160,8 @@ RtpsSampleHeader::init(ACE_Message_Block& mb)
 #undef CASE_SMKIND
 
   if (valid_) {
-    if (message_length_ < SMHDR_SZ) {
+    const size_t available_size = message_length_ ? message_length_ : starting_length;
+    if (available_size < SMHDR_SZ) {
       valid_ = false;
       return;
     }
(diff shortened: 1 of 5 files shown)


dds/DCPS/transport/rtps_udp/RtpsUdpReceiveStrategy.cpp
```diff
@@ -32,6 +32,13 @@ OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
 namespace OpenDDS {
 namespace DCPS {
 
+namespace {
+  bool has_valid_cursor(const ACE_Message_Block& mb)
+  {
+    return mb.rd_ptr() && mb.wr_ptr() && mb.rd_ptr() <= mb.wr_ptr();
+  }
```
sonndinh: RtpsSampleHeader.cpp has the same function. Could it be moved to a common place so both file can share?

dds/DCPS/transport/rtps_udp/RtpsSampleHeader.cpp
```diff
@@ -60,23 +60,36 @@ operator==(const StatusInfo_t& lhs, const StatusInfo_t& rhs)
...
+
+  const size_t contiguous_length = static_cast<size_t>(mb.wr_ptr() - mb.rd_ptr());
+  if (contiguous_length == 0 || mb.total_length() < SMHDR_SZ) {
     return;
   }
 
-  const SubmessageKind kind = static_cast<SubmessageKind>(*mb.rd_ptr());
+  // Manually grab the first two bytes for the SubmessageKind and the byte order.
+  const SubmessageKind kind =
+    static_cast<SubmessageKind>(static_cast<ACE_CDR::Octet>(*mb.rd_ptr()));
```
sonndinh, line 88: This reads the byte pointed to by `rd_ptr` and not two bytes as the comment noted? This seems to assume `SubmessageKind`'s size is 1 byte, is that always the case?

simpsont-oci: The comment moved from earlier in the function call. But the two bytes referenced here are the SubmessageKind (which is an enum of unspecified size locally, but according to the RTPS spec is currently always represented with a a single byte) and the flags from the Submessage header. I'll clean up the comments a bit to make this clearer.

sonndinh (approved):

PR #5212, opened 2026-04-20
Support asynchronous discovery in RtpsRelay

sonndinh:
**Current**
RtpsRelay uses partition information to route client's messages. That requires it to learn the partitions of a client, in particular, through endpoint discovery, before it can forward the client's messages to other clients. Hence, before a client can exchange its first data message with another client, it has to complete discovery with the relay and then with the other client sequentially, in that order.

**Proposed change**
This aims to expedite this process by caching client's partitions learned from a previous connection with the relay and using it for subsequent connections from the same client. This allows the relay to forward the client's messages immediately to already connected clients with matching partitions. The client can proceed its discovery with the relay and with other clients in parallel.


docs/devguide/internet_enabled_rtps.rst
```diff
@@ -390,6 +390,47 @@ The command-line options for the RtpsRelay:
   Expected amount of time a denied partition will remain in the denied partitions list before being cleaned up.
   The default is 900 seconds.
 
+.. option:: -CertificateIdPattern <regex>
+
+  This requires security to be enabled.
+  Specify a regex pattern used to match client participant's certificate subject name to enable the asynchronous discovery feature.
+  For matched client, the first group match will be used as the key to cache the client's partitions.
+  The cached partitions will be used to forward the client's messages immediately in subsequent sessions before the client
```
simpsont-oci: Is the "first group match" is in reference to the regex? If so:

```suggestion
  Specify a regex pattern used to match a client participant's certificate subject name to enable the asynchronous discovery feature.
  For a matched client, the first regex group match will be used as the key to cache the client's partitions.
  The cached partitions will be used to forward the matching client's messages immediately in subsequent sessions before the client
```
Also, is there a good reason why we're not simply using the whole subject name as the cache key? I'm not sure the minor performance bump will outweigh the confusion caused if the regex user accidentally uses two not-quite-unique-enough keys and gives themselves name collisions.

docs/devguide/internet_enabled_rtps.rst
```diff
@@ -390,6 +390,47 @@ The command-line options for the RtpsRelay:
   Expected amount of time a denied partition will remain in the denied partitions list before being cleaned up.
   The default is 900 seconds.
 
+.. option:: -CertificateIdPattern <regex>
```
simpsont-oci, line 393: What's the default value / behavior?

docs/devguide/internet_enabled_rtps.rst
```diff
@@ -390,6 +390,47 @@ The command-line options for the RtpsRelay:
...
+  Specify the duration for an inactive cache entry to be removed from the asynchronous discovery cache for local client participants.
+  The default is 7 days.
+
+.. option:: -AsyncDiscoveryRemoteCacheTimeout <seconds>
+
+  :ref:`Config store key <config-store-keys>`: ``RTPS_RELAY_ASYNC_DISCOVERY_REMOTE_CACHE_TIMEOUT``
+
+  Specify the duration for an inactive cache entry to be removed from the asynchronous discovery cache for remote client participants,
+  i.e., those that connect to other relay instances.
+  The default is 30 days.
```
simpsont-oci, line 425: What's the reasoning on why this is different / longer? If RTPS_RELAY_SYNCHRONIZE_ASYNC_DISCOVERY_CACHE is enabled and the local cache has evicted but the remote cache has not yet evicted, how is the difference resolved?

tests/DCPS/RtpsRelay/Smoke/Args.h
```diff
@@ -29,13 +29,17 @@ struct Args {
   bool terminate_on_data = false;
   bool drain_test = false;
 
+  // For publisher, it means this is the second publisher instance.
+  // For subscriber, it means the subscriber should read from both publisher instances before exits.
```
simpsont-oci:
```suggestion
  // For a publisher, 'second_pub' means this is the second publisher instance.
  // For a subscriber, 'second_pub' means the subscriber should read from both publisher instances before exiting.
```

tests/DCPS/RtpsRelay/Smoke/publisher.cpp
```diff
@@ -36,6 +36,8 @@
 #include <ace/OS_NS_stdlib.h>
 #include <ace/OS_NS_unistd.h>
 
+#include <sstream>
```
simpsont-oci: Does this get used? Or added for testing / debug and not needed anymore?

tests/DCPS/RtpsRelay/Smoke/run_async_disc_test.pl
```diff
@@ -0,0 +1,77 @@
...
+                 $async_discovery_opts .
+                 " -LogDiscovery 1" .
+                 " -LogActivity 1" .
+                 " -LogRelayStatistics 3" .
+                 " -DCPSConfigFile relay1.ini" .
+                 " -ApplicationDomain 42" .
+                 " -VerticalAddress 4444" .
+                 " -HorizontalAddress 127.0.0.1:11444" .
+                 " -UserData thisrelay" .
+                 " -MetaDiscoveryAddress 127.0.0.1:0" .
```
simpsont-oci: What does "127.0.0.1:0" mean here? I see we do that in the lease duration test as well, but it's not clear to me when / why we'd specify this. Is this just a way of disabling it?

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -221,4 +227,377 @@ void GuidPartitionTable::lookup(StringSet& partitions, const OpenDDS::DCPS::GUID
...
   }
 
+void
+GuidPartitionTable::update_cert_partitions_cache(const std::string& key, const StringSet& partitions, const OpenDDS::DCPS::GUID_t& guid)
+{
+  if (key.empty() || !config_.async_discovery_enabled()) {
+    return;
+  }
+
+  // The corresponding participant becomes local, remove its from the remote cache if exists.
```
simpsont-oci:
```suggestion
  // The corresponding participant becomes local, remove it from the remote cache if it exists.
```

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -358,10 +358,37 @@ CORBA::ULong VerticalHandler::process_message(const ACE_INET_Addr& remote_addres
...
       guid_partition_table_.lookup(to_partitions, src_guid);
+      // Denial decision is based only on the "ground truth" routing table and
+      // not the "heuristic" partition cache (looked up below) that may contain
+      // stale partitions for this guid and may cause it to be denied incorrectly.
       if (guid_partition_table_.is_denied(to_partitions)) {
         proxy.deny(src_guid);
       }
-      sent += send(proxy, src_guid, to_partitions, to, send_to_application_participant, msg, now);
+
+      // Initiate async discovery if applies, i.e., the relay has not learned about any partitions
```
simpsont-oci:
```suggestion
      // Initiate async discovery if applicable, i.e., the relay has not learned about any partitions
```

tools/rtpsrelay/EndpointListener.cpp
```diff
@@ -0,0 +1,31 @@
...
+      proxy.cleanup_peers_pending_recipients(iter);
+      iter->second.initiated_async_discovery_with.clear();
+      cert_id = iter->second.identity_info.cert_id();
+    }
+  }
+
+  // Cache all partitions corresponding to this participant for async discovery
+  StringSet all_partitions;
+  guid_partition_table_.lookup(all_partitions, part_guid);
+  guid_partition_table_.update_cert_partitions_cache(cert_id, all_partitions, part_guid);
```
simpsont-oci, line 30: From Claude:

```
  Empty default-partition silently erases the async discovery cache entry

  update_partitions_info() calls guid_partition_table_.lookup(all_partitions, part_guid) and passes the result unconditionally to update_cert_partitions_cache(). When the relay is configured with allow_empty_partition=false (the
  default), GuidPartitionTable::lookup strips the "" entry before returning, so all_partitions is empty for any participant using only the default partition. AsyncDiscoveryCache::update_i then unconditionally erases the
  cert_to_partitions_ entry but skips re-insertion (if (!partitions.empty())). The valid cached entry from the prior session is gone.
  
  Fix: Guard the call — skip update_cert_partitions_cache when all_partitions.empty(), or change update_i to not erase when partitions is empty.
```

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -631,9 +652,91 @@ int run(int argc, ACE_TCHAR* argv[])
...
+      config.certificate_id_pattern().c_str(), e.what()));
+    return EXIT_FAILURE;
+  }
+
+  DDS::Topic_var async_disc_cache_update_topic;
+  DDS::Topic_var async_disc_cache_prune_topic;
+  AsyncDiscoveryCacheUpdateDataWriter_var async_disc_cache_update_writer;
+  AsyncDiscoveryCachePruneDataWriter_var async_disc_cache_prune_writer;
+
+  if (config.synchronize_async_discovery_cache()) {
```
simpsont-oci: From Claude:
```
  Runtime toggle of -SynchronizeAsyncDiscoveryCache from 0→1 dereferences nil writers (crash)

  async_disc_cache_update_writer_ and async_disc_cache_prune_writer_ are only created in RtpsRelay.cpp inside if (config.synchronize_async_discovery_cache()) at startup. If the flag is false at startup, these are nil DDS::DataWriter_var
  members. However, Config::on_data_available (added in this PR) handles the RTPS_RELAY_SYNCHRONIZE_ASYNC_DISCOVERY_CACHE config-store key as a CachedValue<bool>, making it live-reconfigurable. If an operator toggles the key at
  runtime, the next update_cert_partitions_cache or cleanup_local_async_disc_cache call hits async_disc_cache_update_writer_->write(...) on a nil CORBA object — crash.
  
  Fix: Either remove RTPS_RELAY_SYNCHRONIZE_ASYNC_DISCOVERY_CACHE from on_data_available (document it as startup-only), or add nil-guards before each writer use.
```

tools/rtpsrelay/RtpsRelay.cpp
```diff
@@ -542,6 +554,15 @@ int run(int argc, ACE_TCHAR* argv[])
     append(application_properties, DDS::Security::Properties::AuthPrivateKey, identity_key_file);
     append(application_properties, DDS::Security::Properties::AccessGovernance, governance_file);
     append(application_properties, DDS::Security::Properties::AccessPermissions, permissions_file);
+
+    OpenDDS::Security::SSL::Certificate ca_cert(identity_ca_file);
+    std::string ca_subject_name;
+    if (ca_cert.subject_name_to_str(ca_subject_name) == 0 && !ca_subject_name.empty()) {
```
simpsont-oci, line 560: From Claude:

```
  CA subject name extraction failure silently kills async discovery with only a WARNING

  When subject_name_to_str() fails, the else branch logs LM_WARNING and falls through — expected_ca_subject_name_ stays empty. Meanwhile config_.async_discovery_enabled() returns true (it checks only certificate_id_pattern_). In
  process_message, the async-discovery code path is gated on !config_.expected_ca_subject_name().empty(), which is permanently false. The partition cache is never consulted for any client. No operator-visible error is generated, and the
  feature appears functional.
  
  Fix: Return EXIT_FAILURE (with LM_ERROR) when async discovery is requested but CA subject extraction fails, consistent with every other startup-configuration failure in that function.
```

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -221,4 +227,377 @@ void GuidPartitionTable::lookup(StringSet& partitions, const OpenDDS::DCPS::GUID
...
+{
+  StringSet expired_keys;
+  const auto sizes = local_async_disc_cache_.remove_expired(now, config_.async_discovery_cache_timeout(),
+    expired_keys, config_.log_async_discovery() || config_.synchronize_async_discovery_cache());
+
+  relay_stats_reporter_.async_discovery_local_cache_size(sizes.first, now);
+  relay_stats_reporter_.async_discovery_local_expiration_map_size(sizes.second, now);
+
+  if (sizes.first) {
+    const auto next_fire_in = local_async_disc_cache_.earliest_last_access() + config_.async_discovery_cache_timeout() - now;
```
simpsont-oci: From Claude:

```
  TOCTOU between remove_expired and earliest_last_access() can produce max_value + timeout overflow, permanently disabling cleanup

  Both calls independently acquire and release AsyncDiscoveryCache::mutex_. Between them, another reactor thread can drain the remaining entries. earliest_last_access() then returns MonotonicTimePoint::max_value
  (ACE_Time_Value(time_t::max, 999999)). Adding any positive TimeDuration to that overflows ACE_Time_Value with no saturation protection, and SporadicEvent::schedule() has no cap. The cleanup task either fires at a nonsensical time or
  not at all — the cache grows without bound from that point.
  
  Fix: Either hold the mutex across both operations (expose a combined method), or skip scheduling when earliest_last_access() == max_value.
```

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -682,6 +720,18 @@ CORBA::ULong VerticalHandler::send(GuidAddrSet::Proxy& proxy,
...
+            // subsequent messages from the target can be forwarded to the source, thus
+            // allowing them to discover each other.
+            p->second.pending_recipients.insert(src_guid);
+          }
+        }
+      }
+      if (async_discovery) {
+        guids.erase(src_guid);
+        if (iter != proxy.end()) {
+          iter->second.initiated_async_discovery_with.insert(guids.begin(), guids.end());
```
simpsont-oci: From Claude:

```
  initiated_async_discovery_with over-counts: records GUIDs for offline targets that never received src_guid in their pending_recipients

  Inside the send loop, p->second.pending_recipients.insert(src_guid) only runs when p != proxy.end() (target is active). But after the loop, initiated_async_discovery_with.insert(guids.begin(), guids.end()) runs for all
  partition-matched GUIDs unconditionally. A target that was in the routing table but had no active proxy entry gets into initiated_async_discovery_with without ever gaining src_guid in its pending_recipients.
  cleanup_peers_pending_recipients later calls erase(src_guid) on that target — a no-op — but the asymmetry means src_guid never learns of messages from that target via the async path.
```

tools/rtpsrelay/GuidAddrSet.cpp
```diff
@@ -489,6 +489,7 @@ void GuidAddrSet::remove(const OpenDDS::DCPS::GUID_t& guid,
     --mark_count_;
   }
 
+  cleanup_peers_pending_recipients(it);
```
simpsont-oci, line 492: From Claude:

```
  pending_recipients / initiated_async_discovery_with unbounded growth under session churn

  Both sets are cleaned up by GuidAddrSet::remove (session expiry) or EndpointListener::update_partitions_info (endpoint discovery completion). A participant that repeatedly initiates async discovery sessions but never publishes or
  subscribes (endpoint discovery never completes) will have its entry in the routing table accumulate one pending_recipients GUID per peer per session with no intermediate pruning until the peer also disconnects. High-churn relay
  deployments with poorly-behaved clients could see these per-participant sets grow proportionally to session count.
```

simpsont-oci (changes requested):

docs/devguide/internet_enabled_rtps.rst
sonndinh: The default is the empty string that will disable the async discovery feature. I updated the docs for that.

docs/devguide/internet_enabled_rtps.rst
sonndinh: Yes, it's in reference to the regex.

The regex provides a way for the user to control which part of the subject name they want to use as the key. In some existing use cases, a key identifying a unique user might appear in two slightly different subject names, for example, each associated with a device from that user. So if we use the whole subject name as the key, a user can have more than one entry in the cache which could be unexpected.

docs/devguide/internet_enabled_rtps.rst
sonndinh: There is a new topic called "Relay Async Discovery Cache Prune" introduced. When the cache entries for some local users have been evicted, this relay instance will publish to this topic and update other relays so they can remove the users from their caches for remote users.

Besides explicit pruning based on received samples from that topic, pruning "inactive" remote users, i.e., remote users whose updates have not been received for a specified period, is based on a timer with this timeout value. The reasoning for a longer value is that if there is no update for a remote user for a long enough period, with updates being either the remote user has moved to another remote relay or it has been evicted from its current relay, this relay knows that the user has attached to the same relay for that longer period of time and can assume that there's a good chance it will continue attaching to the same relay. With that, it can remove the remote user from its remote cache.

Though it's not perfect since the same user can later move to this relay which had already removed the user's entry, I think it's a good trade-off for a simpler implementation to handle an edge case when a relay can crash and lose its cache, and hence won't be able to explicitly tell other relays to prune the user. Plus, if the user moves to this relay, the cost only happens for the initial connection when the relay needs to build the partition cache for the user -- essentially, it loses the optimization benefit only for the first session.

tests/DCPS/RtpsRelay/Smoke/publisher.cpp
sonndinh: I probably added it for debugging and forgot to remove. I'll remove it.

tests/DCPS/RtpsRelay/Smoke/run_async_disc_test.pl
sonndinh: I probably copied it from the lease duration test .pl file. My guess is that for the lease duration test, there are two relays so specifying port 0 makes each relay choose a random, non-conflicting port for listening to meta discovery requests. Otherwise, both will select the same default port and one of them will fail.

For this test, it doesn't matter since there is one relay, and the metachecker process is not used, so I'll probably remove that `-MetaDiscoveryAddress` line.

tools/rtpsrelay/EndpointListener.cpp
sonndinh: The default behavior is to allow the empty partition unless specified otherwise with the `-AllowEmptyPartition` command-line option.

When the relay learns about updated partitions for a client, via endpoint discovery, I think it's good to update the cache immediately so it contains the latest partition info. In case the client's partition list becomes empty and the relay is configured to disallow the empty partition, it's valid to remove the client's cache entry because it's consistent with the client's input and the relay config.

tools/rtpsrelay/RtpsRelay.cpp
sonndinh: Yea, this is a good point. I'll fix this.

tools/rtpsrelay/GuidPartitionTable.cpp
sonndinh: I updated to check whether `earliest_last_access` returns `max_value` and schedule the timer only when it doesn't.

tools/rtpsrelay/GuidAddrSet.cpp
sonndinh: Claude may be confused here. Async discovery works in an asymmetric way -- only participants whose partitions haven't been learned by the relay from endpoint discovery can initiate async discovery with other clients who have already completed endpoint discovery with the relay.

When a participant initiates async discovery, it adds itself (i.e., its GUID) to the target participants' pending recipients so that messages from those participants can be forwarded back to the initiating participant. It also adds those target participants to the list of participants it has initiated async discovery with; this list is used to find the target clients for cleanup. With that, I don't think these two sets can unboundedly grow.

tools/rtpsrelay/RtpsRelay.cpp
sonndinh: Updated to exit when failing to extract RtpsRelay's CA subject name. Client's CA subject name must match the relay's CA subject name for its messages to be forwarded while async discovery is in effect.

simpsont-oci (approved):

tests/tsan_tests.lst
```diff
@@ -168,6 +168,8 @@ tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl: !DCPS_MIN CXX11 RTPS !NO_
 tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl reverse: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
 tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl secure: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
 tests/DCPS/RtpsRelay/Smoke/run_lease_duration_test.pl reverse secure: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
+tests/DCPS/RtpsRelay/Smole/run_async_disc_test.pl: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
+tests/DCPS/RtpsRelay/Smole/run_async_disc_test.pl cross_relay: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
```
simpsont-oci:
```suggestion
tests/DCPS/RtpsRelay/Smoke/run_async_disc_test.pl: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
tests/DCPS/RtpsRelay/Smoke/run_async_disc_test.pl cross_relay: !DCPS_MIN CXX11 RTPS !NO_BUILT_IN_TOPICS !OPENDDS_SAFETY_PROFILE RAPIDJSON
```

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -221,4 +227,395 @@ void GuidPartitionTable::lookup(StringSet& partitions, const OpenDDS::DCPS::GUID
...
+  update_i(key, partitions, now, prev_partitions);
+  return {cert_to_partitions_.size(), expiration_map_.size()};
+}
+
+std::pair<size_t, size_t>
+GuidPartitionTable::AsyncDiscoveryCache::update(const AsyncDiscoveryCacheEntrySeq &entries, const OpenDDS::DCPS::MonotonicTimePoint &now)
+{
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, mutex_, std::make_pair(0, 0));
+  for (const auto &entry : entries) {
+    const auto &key = entry.key();
```
mitza-oci:
```suggestion
GuidPartitionTable::AsyncDiscoveryCache::update(const AsyncDiscoveryCacheEntrySeq& entries, const OpenDDS::DCPS::MonotonicTimePoint& now)
{
  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, mutex_, std::make_pair(0, 0));
  for (const auto& entry : entries) {
    const auto& key = entry.key();
```
A few more of these below

tools/rtpsrelay/GuidPartitionTable.cpp
```diff
@@ -221,4 +227,395 @@ void GuidPartitionTable::lookup(StringSet& partitions, const OpenDDS::DCPS::GUID
...
+    if (remove_i(key)) {
+      removed_keys.insert(key);
+    }
+  }
+  return {cert_to_partitions_.size(), expiration_map_.size()};
+}
+
+std::pair<size_t, size_t>
+GuidPartitionTable::AsyncDiscoveryCache::remove_expired(const OpenDDS::DCPS::MonotonicTimePoint &now, const OpenDDS::DCPS::TimeDuration &timeout,
+                                    StringSet &expired_keys, bool record_expired_keys)
```
mitza-oci: Fix whitespace here and on similar lines below

mitza-oci (changes requested):

mitza-oci:
This appears to be based on the NEWS update in this branch https://github.com/OpenDDS/OpenDDS/actions/runs/27786939059/job/82225913366?pr=5212

sonndinh:
> This appears to be based on the NEWS update in this branch https://github.com/OpenDDS/OpenDDS/actions/runs/27786939059/job/82225913366?pr=5212

I noticed these errors started happening around the time of the latest OpenDDS release, if I remember correctly. The two files where the errors are reported from, docs/news.rst and docs/this-release.rst, don't exist in this branch. I'm not sure why it's checking those files.

mitza-oci (approved):

PR #5215, opened 2026-04-24
TCP Transport Optional Buffer Size Configuration

simpsont-oci:
Problem:
 - THE TCP transport currently hard-codes socket buffer sizes of `ACE_DEFAULT_MAX_SOCKET_BUFSIZ` which is only 65535 bytes.
   - On modern systems, this is actually a fairly small TCP buffer size.
   - But perhaps even worse, when we automatically pin TCP buffer sizes like this, the underlying host will no longer attempt to provide dynamic window sizing for TCP connections, which generally allow for fine-tuned and better performance than simply setting a single buffer size.

Solution:
 - Make TCP socket buffer sizes configurable, but the default '0' value is used to avoid setting an explicit buffer size for the TCP entirely, allowing system "default" behavior, which for modern platforms often includes fine-tuned performance behaviors like dynamic window sizing.


dds/DCPS/transport/tcp/TcpInst.cpp
```diff
@@ -100,6 +102,30 @@ OpenDDS::DCPS::TcpInst::enable_nagle_algorithm() const
   return TheServiceParticipant->config_store()->get_boolean(config_key("ENABLE_NAGLE_ALGORITHM").c_str(), false);
 }
 
+void
+OpenDDS::DCPS::TcpInst::send_buffer_size(ACE_INT32 sbs)
+{
+  TheServiceParticipant->config_store()->set_int32(config_key("SEND_BUFFER_SIZE").c_str(), sbs);
```
sonndinh, line 108: Would it be good to check the validity of the input `sbs` here, for example, to prevent negative values? Or maybe the interface could use an unsigned integer parameter?

sonndinh (approved):

dds/DCPS/transport/tcp/TcpInst.cpp
simpsont-oci: Hmm. The send and receive buffer sizes are already checked for whether they're greater than 0 in `TcpConnection::set_sock_options()` since we want to avoid setting the buffer sizes in the default (0) case. We could also add a check here, but it's not adding much value and it doesn't follow the pattern of the other config store options listed in this section.

sonndinh: I'm good with the current changes -- if we decide to add input validity check later, we can go through other options as well.

sonndinh (approved):

PR #5217, opened 2026-05-01
Add XTypes Runtime XML/JSON Support for Interop Tests

simpsont-oci:
This adds the OpenDDS-side runtime XTypes pieces needed to drive the dds-xtypes interop harness with OpenDDS participants.

- Adds DDS-XML type loading under OpenDDS::XTypes, gated on Xerces support.
- Adds DynamicData JSON load/write helpers, including union discriminator format handling for DDS-JSON and dds-xtypes test data.
- Implements and tests DynamicData::equals() for runtime data comparison.
- Adds focused unit coverage for XML type loading, JSON DynamicData population/round-trip, and DynamicData equality.
- Updates build/CI plumbing so Xerces-dependent XML type loading is only built and tested where Xerces is available.


dds/DCPS/XTypes/DynamicDataJson.h
```diff
@@ -0,0 +1,70 @@
...
+#if defined OPENDDS_RAPIDJSON && !defined OPENDDS_SAFETY_PROFILE
+
+enum DynamicDataJsonDiscriminatorFormat {
+  /**
+   * Input accepts either {"$discriminator": ..., "member": ...} or
+   * {"member": ...}.  Output uses the active-member-only representation.
+   */
+  DYNAMIC_DATA_JSON_DISCRIMINATOR_AUTO,
+
+  /// Include/require the OpenDDS-style "$discriminator" JSON member.
```
mitza-oci: It's in the OMG DDS-JSON spec

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -799,6 +1079,23 @@ namespace {
...
+{
+  if (!lhs || !rhs) {
+    return lhs == rhs;
+  }
+  DDS::DynamicType_var lhs_type = lhs->type();
+  DDS::DynamicType_var rhs_type = rhs->type();
+  DDS::DynamicType_var lhs_base = safe_base_type(lhs_type);
+  DDS::DynamicType_var rhs_base = safe_base_type(rhs_type);
+  if (!lhs_base || !rhs_base ||
+      lhs_base->get_kind() != rhs_base->get_kind() ||
```
sonndinh: This may already be covered by the `lhs_base->equals(rhs_base)` call in the next line.

dds/DCPS/XTypes/DynamicDataJson.h
```diff
@@ -0,0 +1,70 @@
...
+  DYNAMIC_DATA_JSON_DISCRIMINATOR_FIELD,
+
+  /// Use the dds-xtypes interop active-member-only union representation.
+  DYNAMIC_DATA_JSON_DISCRIMINATOR_ACTIVE_MEMBER
```
mitza-oci:
```suggestion
  DYNAMIC_DATA_JSON_DISCRIMINATOR_ACTIVE_MEMBER,
```

dds/DCPS/XTypes/DynamicDataJson.h
```diff
@@ -0,0 +1,70 @@
...
+};
+
+OpenDDS_Dcps_Export DDS::ReturnCode_t dynamic_data_from_json(
+  DDS::DynamicData_ptr data,
+  const char* json,
+  const DynamicDataJsonOptions& options = DynamicDataJsonOptions());
+
+OpenDDS_Dcps_Export DDS::ReturnCode_t dynamic_data_from_json_file(
+  DDS::DynamicData_ptr data,
+  const ACE_TString& json_file,
```
mitza-oci: Why ACE_TString?

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +509,283 @@ namespace {
...
+    const char* const src = float128_bytes(value);
+#if defined(ACE_BIG_ENDIAN)
+    ACE_OS::memcpy(bytes, src, 16);
+#else
+    for (size_t i = 0; i != 16; ++i) {
+      bytes[i] = src[15 - i];
+    }
+#endif
+
+#if ACE_SIZEOF_LONG_DOUBLE == 16 && LDBL_MANT_DIG == 64
```
mitza-oci: This part doesn't look right to me.  ACE_CDR::LongDouble uses an array of chars so it's just bytes, it doesn't matter if the local platform uses x86 long double.

dds/DCPS/XTypes/DynamicDataJson.h
mitza-oci: These kind of lines should always end in a comma so that future revisions that add to the list don't need to change them.

dds/DCPS/XTypes/XmlTypeProvider.cpp
```diff
@@ -0,0 +1,1504 @@
...
+    if (!parse(error)) {
+      report(error, type_name);
+      return DDS::RETCODE_ERROR;
+    }
+
+    const std::string requested = strip_leading_colons(type_name);
+    if (types_.find(requested) == types_.end()) {
+      std::ostringstream os;
+      os << "type '" << requested << "' was not found";
+      report(os.str(), requested);
```
mitza-oci: Why use ostringstream for string concatenation?

dds/DCPS/XTypes/DynamicDataJson.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  const DynamicDataJsonOptions& options,
+  const std::string& path,
+  JsonValue::ConstMemberIterator discriminator)
+{
+  DDS::TypeDescriptor_var td;
+  if (!get_type_descriptor(td, type)) {
+    return DDS::RETCODE_BAD_PARAMETER;
+  }
+  DDS::DynamicType_var discriminator_type = get_base_type(td->discriminator_type());
+  DDS::Int32 discriminator_value = 0;
```
mitza-oci, line 828: Does this need to handle 64-bit discriminators?

dds/DCPS/XTypes/DynamicDataJson.cpp
```diff
@@ -0,0 +1,1292 @@
...
+  }
+
+  switch (base->get_kind()) {
+  case TK_STRUCTURE:
+    return populate_structure(data, base, value, options, path);
+  case TK_UNION:
+    return populate_union(data, base, value, options, path);
+  case TK_ARRAY:
+  case TK_SEQUENCE:
+    return populate_collection(data, base, value, options, path);
```
mitza-oci, line 1075: Should this (across the entire PR, not just here) support TK_MAP as well?

mitza-oci (changes requested):

dds/DCPS/XTypes/DynamicDataJson.cpp
```diff
@@ -0,0 +1,1290 @@
...
+}
+
+std::string path_member(const std::string& path, const char* name)
+{
+  return path.empty() ? std::string(name) : path + "." + name;
+}
+
+std::string path_index(const std::string& path, DDS::UInt32 index)
+{
+  return path + '[' + std::to_string(index) + ']';
```
sonndinh: Some builds fail because `to_string` requires C++11

dds/DCPS/XTypes/DynamicDataJson.h
```diff
@@ -0,0 +1,70 @@
...
+
+namespace OpenDDS {
+namespace XTypes {
+
+#if defined OPENDDS_RAPIDJSON && !defined OPENDDS_SAFETY_PROFILE
+
+enum DynamicDataJsonDiscriminatorFormat {
+  /**
+   * Input accepts either {"$discriminator": ..., "member": ...} or
+   * {"member": ...}.  Output uses the active-member-only representation.
```
mitza-oci: Why was this chosen as the output format?  It would be good to avoid a lossy conversion here.

dds/DCPS/XTypes/DynamicDataJson.cpp
```diff
@@ -0,0 +1,1298 @@
...
+  if (parse_unsigned_text(text, numeric)) {
+    output = numeric;
+    return true;
+  }
+
+  output = 0;
+  const char* token_begin = text;
+  while (token_begin && *token_begin) {
+    const char* token_end = std::strchr(token_begin, '|');
+    const std::string flag(token_begin, token_end ? token_end - token_begin : std::strlen(token_begin));
```
mitza-oci: warnings here

mitza-oci (changes requested):

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+  {
+    return type && type->get_descriptor(descriptor) == DDS::RETCODE_OK && descriptor;
+  }
+
+  DDS::DynamicType_var safe_base_type(DDS::DynamicType_ptr type)
+  {
+    DDS::TypeDescriptor_var descriptor;
+    if (!get_type_descriptor(descriptor, type)) {
+      return DDS::DynamicType::_nil();
+    }
```
sonndinh: `descriptor` is obtained here but not used after that -- looks like it is used to check if a valid TypeDescriptor is present?

`get_base_type` call at line 544 has similar checks to the new `get_type_descriptor` function, so it looks like there are some duplicates here. Could we use the existing `get_base_type` function in places where `safe_base_type` is being used?

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+      DDS::DynamicData_var lhs_nested;
+      DDS::DynamicData_var rhs_nested;
+      DDS::ReturnCode_t lhs_rc = lhs->get_complex_value(lhs_nested, lhs_id);
+      DDS::ReturnCode_t rhs_rc = rhs->get_complex_value(rhs_nested, rhs_id);
+      if (lhs_rc != rhs_rc) {
+        return false;
+      }
+      if (lhs_rc != DDS::RETCODE_OK) {
+        return true;
+      }
```
sonndinh, line 706: Should this return `false` because this looks like an error case when it fails to get the nested dynamic data objects to continue comparing?

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+          return false;
+        }
+        DDS::MemberDescriptor_var md;
+        if (member->get_descriptor(md) != DDS::RETCODE_OK) {
+          return false;
+        }
+        const bool member_equal = data_equal_i(lhs, rhs, md->id(), md->id(), md->type());
+        if (!member_equal) {
+          return false;
+        }
```
sonndinh:
```suggestion
        if (!data_equal_i(lhs, rhs, md->id(), md->id(), md->type())) {
          return false;
        }
```

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+      DDS::TypeDescriptor_var td;
+      if (!get_type_descriptor(td, base)) {
+        return false;
+      }
+      DDS::DynamicType_var element_type = safe_base_type(td->element_type());
+      if (!element_type) {
+        return false;
+      }
+      DDS::UInt32 lhs_count = lhs->get_item_count();
+      DDS::UInt32 rhs_count = rhs->get_item_count();
```
sonndinh:
```suggestion
      const DDS::UInt32 lhs_count = lhs->get_item_count();
      const DDS::UInt32 rhs_count = rhs->get_item_count();
```

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+      if (kind == TK_ARRAY) {
+        lhs_count = rhs_count = bound_total(td);
```
sonndinh: I think `lhs_count` and `rhs_count` have already been set by the above `get_item_count` calls, so this may be redundant.

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -506,6 +508,272 @@ namespace {
...
+    }
+
+    DDS::ReturnCode_t left_rc = DDS::RETCODE_ERROR;
+    DDS::ReturnCode_t right_rc = DDS::RETCODE_ERROR;
+    switch (base->get_kind()) {
+    case TK_BOOLEAN: {
+      CORBA::Boolean left = false, right = false;
+      left_rc = lhs->get_boolean_value(left, lhs_id);
+      right_rc = rhs->get_boolean_value(right, rhs_id);
+      return left_rc == right_rc && (left_rc != DDS::RETCODE_OK || left == right);
```
sonndinh: This returns `true` when `left_rc == right_rc` and both are not `RETCODE_OK`. Shouldn't it return `false` instead in that case? The other switch cases have the same logic.

sonndinh (changes requested):

dds/DCPS/XTypes/DynamicDataJson.cpp
simpsont-oci: The actual discriminator DynamicData is still populated using the discriminator DynamicType, including 64-bit types. This local Int32 is only the normalized union case label used for branch lookup; UnionCaseLabelSeq stores labels as 32-bit signed values. json_to_discriminator() parses 64-bit JSON values and rejects values that can’t be represented as a union label. I'll add a comment to make that distinction clearer.

mitza-oci (approved):

dds/DCPS/XTypes/XmlTypeProvider.cpp
```diff
@@ -0,0 +1,1710 @@
...
+  bool wide_string;
+  LBound string_bound;
+  std::string named_type;
+  std::string scope;
+  bool has_sequence;
+  LBound sequence_bound;
+  bool has_array;
+  OPENDDS_VECTOR(LBound) array_dimensions;
+  TypeSpec* map_key;
+  TypeSpec* map_value;
```
sonndinh: Would `OpenDDS::DCPS::unique_ptr` be a good use to manage these two pointers?

dds/DCPS/XTypes/XmlTypeProvider.cpp
```diff
@@ -0,0 +1,1710 @@
...
+  if (initialized) {
+    return true;
+  }
+  if (failed) {
+    error = "Xerces-C initialization previously failed";
+    return false;
+  }
+
+  try {
+    XERCES_CPP_NAMESPACE::XMLPlatformUtils::Initialize();
```
sonndinh: Should there be a matching call to `XMLPlatformUtils::Initialize()` somewhere to clean up the resources from `Initialize`?

dds/DCPS/XTypes/XmlTypeProvider.cpp
```diff
@@ -0,0 +1,1710 @@
...
+};
+
+} // namespace
+
+DDS::ReturnCode_t load_xml_type(DDS::DynamicType_var& type,
+                                const ACE_TString& xml_file,
+                                const std::string& fully_qualified_type_name)
+{
+  XmlTypeLoader loader(xml_file);
+  return loader.load(type, fully_qualified_type_name);
```
sonndinh, line 1725: Is it possible that `load_xml_type` is called more than once with the same `xml_file` but different `fully_qualified_type_name` to load different types from the same file? If so, it looks like each time the same XML file will be loaded and parsed -- should the parsed result for a file be cached for subsequent queries?

simpsont-oci: If repeated loads become a real cost, we can add an explicit reusable loader/cache API later, but for now I don't think it's worth adding the complexity since you'd have to worry about things like file invalidation and shared state.

sonndinh (approved):

PR #5223, opened 2026-05-06
Misc Coverity Cleanup

simpsont-oci:
This PR is intended to address a number of small Coverity issues, namely:
- 1648943 - JsonValueWriter.h::to_json - Check vwrite() / writer return values in the string-returning to_json() overloads; return empty string on serialization failure. 
- 1527474 - TypeSupportImpl.h::TypeSupportInitializer - Check register_type() result, assert in debug if it fails, and only unregister during destruction if registration succeeded. 
- 1573405, 1573403, 1573398, 1573396 - XTypes/IdlScanner.h - Initialize IdlToken::numeric_value_ fields in the default constructor so token factory paths do not leave scalar members
 indeterminate. 
- 1594377 - XTypes/Utils.cpp::MemberPathParser::get_next_subpath - Remove the size_t decrement / underflow-prone loop logic and make consumption explicit. 
- 1594403 - DataSampleHeader.cpp::operator<< - Replace boolean left-shifts with DataSampleHeader::mask_flag() accumulation for serialized flag bytes. 
- 1594437, 1594418 - ReliableSession.cpp::send_naks - Avoid overflow around nak_delay_intervals_ + 1, and replace the floating-point threshold test with integer helper logic. 
- 1644788, 1644783 - typeobject_generator.cpp::generate_map_type_identifier - Initialize map collection_flag fields. 
- 1515269, 1515232 - typeobject_generator.cpp::generate_array_type_identifier - Initialize array collection_flag fields. 
- 1515195, 1515173 - typeobject_generator.cpp::generate_sequence_type_identifier - Initialize sequence collection_flag fields. 
- 1515285, 1515220 - typeobject_generator.cpp::generate_alias_type_identifier - Initialize alias flags and related-type flags. 
- 1515319, 1515260, 1515240 - typeobject_generator.cpp::generate_struct_type_identifier  - Build struct members via initialized XTypes constructors instead of partially filling default-constructed members.
- 1515230, 1515197, 1515149 - typeobject_generator.cpp::generate_union_type_identifier - Same idea for union discriminator/members: use initialized constructor paths and explicit flag variables. 
- 1515206, 1515193, 1515152 - typeobject_generator.cpp::generate_enum_type_identifier - Build enum literals from initialized common/detail objects. 
- 1644767, 1510264, 1510260, 1510258, 1510256, 1510253, 1510251, 1510250, 1510248, 1510245, 1510243, 1500892, 1500891, 1500884, 1500867 - typeobject_generator.cpp unchecked dynamic_cast sites - Add ast_cast() helper that checks the cast and aborts with an IDL generator error if the AST type is unexpected.
- 1643839 - typeobject_generator.cpp::set_builtin_member_annotations - Avoid direct unchecked cast to HashidAnnotation; use be_global->hashid() instead.

### DIFF
--- a/dds/DCPS/XTypes/IdlScanner.h
+++ b/dds/DCPS/XTypes/IdlScanner.h
@@ -74,10 +74,7 @@ public:
   IdlToken()
   : kind_(Error)
   {
-    numeric_value_.signed_ = false;
-    numeric_value_.integer_value_ = 0;
-    numeric_value_.exponent_signed_ = false;
-    numeric_value_.exponent_value_ = 0;
+    numeric_value_ = Numeric();
   }
 
   static IdlToken make_error()
--- a/dds/DCPS/XTypes/IdlScanner.h
+++ b/dds/DCPS/XTypes/IdlScanner.h
@@ -73,7 +73,12 @@ public:
 
   IdlToken()
   : kind_(Error)
-  {}
+  {
+    numeric_value_.signed_ = false;
+    numeric_value_.integer_value_ = 0;
+    numeric_value_.exponent_signed_ = false;
+    numeric_value_.exponent_value_ = 0;
+  }
 
   static IdlToken make_error()
   {
--- a/dds/DCPS/XTypes/Utils.cpp
+++ b/dds/DCPS/XTypes/Utils.cpp
@@ -155,7 +155,7 @@ bool MemberPathParser::get_next_subpath()
   size_t got = 0; // Char count to use for the result
   char c = '\0';
   bool scan = true;
-  for (; i < left && scan; ++i) {
+  while (i < left && scan) {
     c = path[i];
     switch (c) {
     case '.':
@@ -168,14 +168,16 @@ bool MemberPathParser::get_next_subpath()
         error = true;
         return false;
       }
-      --i; // Don't consume, leave for next iteration
-      // fallthrough
+      scan = false; // Don't consume, leave for next iteration
+      break;
     case ']':
       scan = false;
+      ++i;
       break;
 
     default:
       ++got;
+      ++i;
     }
   }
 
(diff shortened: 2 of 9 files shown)


dds/DCPS/XTypes/IdlScanner.h
```diff
@@ -73,7 +73,12 @@ class OpenDDS_Dcps_Export IdlToken {
 
   IdlToken()
   : kind_(Error)
-  {}
+  {
+    numeric_value_.signed_ = false;
+    numeric_value_.integer_value_ = 0;
+    numeric_value_.exponent_signed_ = false;
+    numeric_value_.exponent_value_ = 0;
+  }
```
mitza-oci, line 78: Wouldn't this be cleaner as a default constructor of `Numeric`?

dds/DCPS/XTypes/Utils.cpp
```diff
@@ -155,7 +155,7 @@ bool MemberPathParser::get_next_subpath()
   size_t got = 0; // Char count to use for the result
   char c = '\0';
   bool scan = true;
-  for (; i < left && scan; ++i) {
+  while (i < left && scan) {
```
mitza-oci, line 158: Not sure why `for` -> `while` here is better, it duplicates the `++i` below.

simpsont-oci: The point is to avoid `--i` for the sake of an underflow, which was only ever required because the loop unconditionally incremented but some branches didn't want to increment `i`. So this spreads out the loop logic a bit, but makes it clear that `i` only ever increments. A stylistic tradeoff, to be sure, but at least this way doesn't have the potential for UB.

mitza-oci: I thought that was handled by the `break;`

mitza-oci: I misread it - I now see that the `break;` belongs to `switch`
Maybe `if` is better than `switch` in this case, but it's fine either way

dds/DCPS/XTypes/IdlScanner.h
simpsont-oci: Sure.

mitza-oci (approved):

PR #5225, opened 2026-05-06
Check for infinite interval before scheduling Liveliness timer

sonndinh:
- Problem
`LivelinessTimer::add_adjust` updates the liveliness interval of a `DomainParticipant` based on the input data writer's QoS policy. When the writer's liveliness kind does not match the kind of the specific `LivelinessTimer` object from `DomainParticipantImpl`, or when the writer's liveliness lease duration is infinite, the timer is scheduled with infinite duration, causing overflow when computing expiration time.

- Solution
Add a check and return when either of these conditions happens.

### DIFF
--- a/dds/DCPS/DomainParticipantImpl.cpp
+++ b/dds/DCPS/DomainParticipantImpl.cpp
@@ -2324,7 +2324,7 @@ DomainParticipantImpl::LivelinessTimer::~LivelinessTimer()
 void
 DomainParticipantImpl::LivelinessTimer::add_adjust(OpenDDS::DCPS::DataWriterImpl* writer)
 {
-  TimeDuration writer_interval = writer->liveliness_check_interval(kind_);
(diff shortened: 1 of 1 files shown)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -2324,6 +2324,11 @@ DomainParticipantImpl::LivelinessTimer::~LivelinessTimer()
 void
 DomainParticipantImpl::LivelinessTimer::add_adjust(OpenDDS::DCPS::DataWriterImpl* writer)
 {
+  TimeDuration writer_interval = writer->liveliness_check_interval(kind_);
```
simpsont-oci:
```suggestion
  const TimeDuration writer_interval = writer->liveliness_check_interval(kind_);
```

simpsont-oci (changes requested):

simpsont-oci (approved):

PR #5240, opened 2026-06-01
Check for null `ACE_Message_Block` in `peek_helper`

sonndinh:
**Problem**
For XCDR2 mutable data, with length code values of 5, 6, or 7, `read_parameter_id` calls `peek` to preview the value of the nextint field. Malformed data that doesn't have nextint can cause `Serializer` to have a null `current_` data member and crash the process.

**Solution**
Check for null in `peek_helper`. Also added tests for malformed input to `read_parameter_id`.

### DIFF
--- a/dds/DCPS/Serializer.h
+++ b/dds/DCPS/Serializer.h
@@ -679,6 +679,10 @@ public:
   template <typename T>
   bool peek_helper(ACE_Message_Block* const block, size_t bytes, T& t)
   {
+    if (!block) {
+      return false;
+    }
+
     bool result = false;
     char* const rd_ptr = block->rd_ptr();
     const size_t length = block->length();
--- a/tests/unit-tests/dds/DCPS/Serializer.cpp
+++ b/tests/unit-tests/dds/DCPS/Serializer.cpp
@@ -591,7 +591,7 @@ TEST(dds_DCPS_Serializer, read_parameter_id_xcdr2)
 
   const Encoding enc(Encoding::KIND_XCDR2, ENDIAN_BIG);
   ACE_Message_Block mb(sizeof(xcdr));
-  mb.copy((const char*)xcdr, sizeof(xcdr));
+  mb.copy(reinterpret_cast<const char*>(xcdr), sizeof(xcdr));
   Serializer ser(&mb, enc);
   unsigned id;
   size_t size;
@@ -651,7 +651,7 @@ namespace {
   {
     const Encoding enc(Encoding::KIND_XCDR2, ENDIAN_BIG);
     ACE_Message_Block mb(size);
-    if (xcdr && mb.copy(reinterpret_cast<const char *>(xcdr), size) != 0) {
+    if (xcdr && mb.copy(reinterpret_cast<const char*>(xcdr), size) != 0) {
       ACE_ERROR((LM_ERROR, "read_parameter_id_xcdr2: failed to copy data to message block!\n"));
       return false;
     }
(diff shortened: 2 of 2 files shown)


dds/DCPS/Serializer.h
```diff
@@ -679,6 +679,10 @@ class OpenDDS_Dcps_Export Serializer {
   template <typename T>
   bool peek_helper(ACE_Message_Block* const block, size_t bytes, T& t)
   {
+    if (!block) {
```
jwillemsen, line 682: Why not move the `if` around the usage of it so that there is only one return?

tests/unit-tests/dds/DCPS/Serializer.cpp
```diff
@@ -645,3 +645,151 @@ TEST(dds_DCPS_Serializer, read_parameter_id_xcdr2)
   EXPECT_FALSE(must_understand);
   ASSERT_TRUE(ser.skip(size));
 }
+
+namespace {
+  bool read_parameter_id_xcdr2(const unsigned char* xcdr, size_t size)
+  {
+    const Encoding enc(Encoding::KIND_XCDR2, ENDIAN_BIG);
+    ACE_Message_Block mb(size);
+    mb.copy((const char *)xcdr, size);
```
jwillemsen: copy has a return value which is not checked

dds/DCPS/Serializer.h
mitza-oci: Not sure what makes that better - I personally think "just one return statement" is counterproductive in C++.  I like the pattern of checking preconditions early and then not having those checks impact the rest of the function.

sonndinh: > Why not move the if around the usage of it so that there is only one return?

Do you mean to move the check to `peek` instead?

The `peek_helper` function has most of the logic for `peek`, so putting the check here would be more consistent with how they are currently designed. Also, there are two versions of `peek`, a template and an overload function, with basically similar code, so if it's moved to `peek`, there would be the same check in both versions.

tests/unit-tests/dds/DCPS/Serializer.cpp
```diff
@@ -645,3 +645,153 @@ TEST(dds_DCPS_Serializer, read_parameter_id_xcdr2)
   EXPECT_FALSE(must_understand);
   ASSERT_TRUE(ser.skip(size));
 }
+
+namespace {
+  bool read_parameter_id_xcdr2(const unsigned char* xcdr, size_t size)
+  {
+    const Encoding enc(Encoding::KIND_XCDR2, ENDIAN_BIG);
+    ACE_Message_Block mb(size);
+    if (xcdr && mb.copy(reinterpret_cast<const char *>(xcdr), size) != 0) {
```
mitza-oci:
```suggestion
    if (xcdr && mb.copy(reinterpret_cast<const char*>(xcdr), size) != 0) {
```

tests/unit-tests/dds/DCPS/Serializer.cpp
```diff
@@ -645,3 +645,153 @@ TEST(dds_DCPS_Serializer, read_parameter_id_xcdr2)
...
+      0x50, 0x00, 0x00, 0x01,
+      0x00, 0x00, 0x00, 0x04
+    };
+    test_read_parameter_id_xcdr2_ok(xcdr, sizeof(xcdr));
+
+    // Splitted message into multiple blocks should also work
+    const Encoding enc(Encoding::KIND_XCDR2, ENDIAN_BIG);
+    const size_t total_size = sizeof(xcdr);
+    ACE_Message_Block mb(total_size - 3);
+    mb.copy((const char *)xcdr, total_size - 3);
```
mitza-oci: Use the same style of cast consistently in this file

mitza-oci (changes requested):

mitza-oci (approved):

PR #5242, opened 2026-06-16
Free up RtpsRelay admission queue more aggressively

sonndinh:
Ideally, an admission queue entry corresponding to a participant that has disconnected or failed to connect to the relay could be removed to allow new incoming participants to join. However, there isn't a good way to detect promptly whether a participant has disconnected or failed to connect -- by the time that is detected, it's likely the admission queue entry has already been removed due to the queue duration expiry.

This PR adds a heuristic to remove admission queue entries more opportunistically: when the relay receives a disposed/unregister message for a user data writer or reader from a client participant, it takes it as a signal that the client is disconnecting and removes its entry.


tools/rtpsrelay/PublicationListener.cpp
```diff
@@ -92,6 +92,10 @@ void PublicationListener::on_data_available(DDS::DataReader_ptr reader)
                      idx, infos.length()));
         }
         guid_partition_table_.remove(repoid);
+        {
+          GuidAddrSet::Proxy proxy(*guid_addr_set_);
+          proxy.freeup_admission_queue(repoid.guidPrefix);
```
simpsont-oci, line 97: This is more of a general comment, but I'll add it here, since this (and the SubscriptionListener equivalent line) seem most suspicious to me.

I'm not sure that the deletion of any reader or writer is, generally, a good indicator that a participant no longer belongs on the admission queue. Some applications may know and define all their endpoints up front, but for those that create and destroy them dynamically, these lines mean those kinds of applications might inadvertently remove themselves from the admission queue before authentication / discovery is complete. This may not even totally stop that participant from authenticating / finishing discovery, depending on where in the process they are, but it will definitely allow additional participants into the queue, depending on the configured max size.

And while it's probably a stretch, a "malicious" application could intentionally churn its endpoints on connection with many instances, in an attempt to flood / overwhelm the admission queue protections.

sonndinh: Yea, this is a "loose" heuristic for deciding whether a participant is disconnecting based on my understanding that when a participant disconnects, the relay would receive NOT_ALIVE_NO_WRITERS or NOT_ALIVE_DISPOSED samples for its readers and writers (but please correct me if I'm wrong). You're right that it could be a false indicator when a participant dynamically removes one of its writers or readers and still stays alive and connected to the relay. In that case, the effect is that the "handle" corresponding to the participant in the admission queue is removed earlier, without having to wait until the full admission queue duration elapses before its entry can be removed. That opens an available slot in the admission queue to accept a new participant earlier than if the existing participant had stayed in the queue for a longer period of time. The admission queue is a rate limit tool, and so I think the only effect of false removals would be allowing a higher rate of participants to join the relay.

I agree, though, that a malicious application can trick the admission queue into admitting new participants freely, effectively disabling its rate-limit function. I'm open to other alternative approaches.

sonndinh: Following from the earlier discussion, I added a config option to allow the removal based on reader or writer deletion to be turned on and off. The default is disabled, which would be the same as the current relay's behavior.

simpsont-oci (approved):

PR #5243, opened 2026-06-17
XmlTypeProvider library layout changes to address macOS dynamic load issue

simpsont-oci:
Problem:
- node-opendds macOS CI started failing after OpenDDS added Xerces-backed XmlTypeProvider support directly into OpenDDS_Dcps. The failures were misleading at first because they appeared in security-related node-opendds builds, not in obvious XTypes tests.
- We narrowed it down by testing a branch that removed the Xerces dependency from OpenDDS_Dcps, which cleared up the node-opendds CI issues. That showed the issue was not within node-opendds itself, but the base OpenDDS DCPS library gaining a Xerces dependency. The failure was especially visible on macOS ARM64 runners. See here: https://github.com/OpenDDS/node-opendds/actions/runs/27484673404

Solution:
- This solution keeps XmlTypeProvider, but moves it out of OpenDDS_Dcps and into a separate optional library OpenDDS_XTypes_Xml.
- That means:
  - OpenDDS_Dcps no longer depends on Xerces.
  - Consumers that need XML type loading can link OpenDDS_XTypes_Xml.
  - CMake and MPC builds both know how to build/install the new library.
  - Unit tests that exercise XML/XTypes now link the new library when Xerces is enabled.
  - node-opendds can continue using the base OpenDDS libraries without inheriting the problematic Xerces dependency, while XTypes interoperability tooling still has access to XmlTypeProvider.
- See here for node-opendds test branch / PR with green CI: https://github.com/OpenDDS/node-opendds/pull/88


mitza-oci (commented):
Any documentation / NEWS updates needed?

CMakeLists.txt
```diff
@@ -131,6 +131,9 @@ if(NOT CMAKE_CROSSCOMPILING OR OPENDDS_JUST_BUILD_HOST_TOOLS)
 endif()
 
 add_subdirectory(dds) # OpenDDS_Dcps
+if(OPENDDS_XERCES3 AND NOT OPENDDS_SAFETY_PROFILE)
+  add_subdirectory(dds/DCPS/XTypes) # OpenDDS_XTypes_Xml
```
sonndinh, line 135: Would it be better organized to use a subdirectory within dds/DCPS/XTypes for the new library? Maybe not now but something to consider later.

simpsont-oci: I think we should leave it as is for now. The new library is only splitting out XmlTypeProvider.cpp while keeping the public headers under dds/DCPS/XTypes, so another subdirectory would add build/layout churn without much payoff. If this grows beyond the XML type provider, then I agree a subdirectory would make more sense.

sonndinh (approved):

PR #5246, opened 2026-06-24
[RtpsRelay] Forward messages without `GuidAddrSet` lock

sonndinh:
**Problem**
The current `VerticalHandler::send` function holds the `GuidAddrSet` lock while sending IO network messages, which could cause it to hold the lock for a prolonged duration.

**Solution**
Break up the `send` function into a new `prepare_send` function that collects all target IP addresses for a message, and a new `send` function that does the actual forwarding to those addresses. The `prepare_send` function still needs the `GuidAddrSet` lock to collect the addresses, but the new `send` function forwards the message outside of the lock.

Note: this change applies mainly to the forwarding of messages from client participants. There are two cases when the lock is still held while the relay forwards the messages:
- Messages from the application participants (forwarded to the specified local client participants)
- The first RTPS message of the newly admitted client (forwarded to the application participant)

### DIFF
--- a/tools/rtpsrelay/RelayHandler.cpp
+++ b/tools/rtpsrelay/RelayHandler.cpp
@@ -331,64 +331,74 @@ CORBA::ULong VerticalHandler::process_message(const ACE_INET_Addr& remote_addres
       (remote_address == application_participant_addr_) &&
       (src_guid == config_.application_participant_guid());
 
-    GuidAddrSet::Proxy proxy(guid_addr_set_);
-    OpenDDS::DCPS::ThreadStatusManager::Event evLocked(statusManager, READ_MASK | SIGNAL_MASK, handle_as_int);
-
-    bool already_checked_admit = false;
-    if (!record_activity(proxy, addr_port, now, src_guid, from_application_participant, already_checked_admit)) {
-      stats_reporter_.ignored_message(msg_len, now, type);
-      return 0;
-    }
-
-    cache_message(proxy, src_guid, to, msg, now);
-
-    bool admitted = false;
-    if (proxy.defer_client(from_application_participant, src_guid, now, already_checked_admit, admitted)) {
-      stats_reporter_.ignored_message(msg_len, now, type);
-      return 0;
-    }
-
     CORBA::ULong sent = 0;
-
-    if (admitted && spdp_handler_) {
-      sent += spdp_handler_->send_to_application_participant(proxy, src_guid, now);
-    }
-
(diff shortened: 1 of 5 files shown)


mitza-oci (approved):

tools/rtpsrelay/RelayHandler.cpp
```diff
@@ -331,64 +331,74 @@ CORBA::ULong VerticalHandler::process_message(const ACE_INET_Addr& remote_addres
...
-          const auto pos = proxy.find(src_guid);
-          if (pos != proxy.end()) {
-            const auto ca_sn = pos->second.identity_info.ca_sn();
-            if (config_.expected_ca_subject_name() == ca_sn) {
-              const auto key = pos->second.identity_info.cert_id();
-              guid_partition_table_.lookup_cert_partitions_cache(to_partitions, key, src_guid);
-            } else if (config_.log_async_discovery()) {
-              HANDLER_WARNING((LM_WARNING, "(%P|%t) WARNING: VerticalHandler::process_message %C Expected CA subject name '%C', but got '%C' from GUID %C\n",
-                name_.c_str(), config_.expected_ca_subject_name().c_str(), ca_sn.c_str(), OpenDDS::DCPS::LogGuid(src_guid).c_str()));
+      from_client = do_normal_processing(proxy, remote_address, src_guid, to, admitted, send_to_application_participant, msg, now, sent);
```
simpsont-oci, line 364: Copying and pasting a quick review from codex:

Medium - Some forwarding still occurs while holding GuidAddrSet lock. In VerticalHandler::process_message, the new scoped lock still calls spdp_handler_->send_to_application_participant(...). That function immediately calls the new send(...), which invokes venqueue_message. The STUN admission path does the same. Additionally, application-participant SPDP/SEDP forwarding inside do_normal_processing still invokes venqueue_message under the lock. Thus the PR moves normal client forwarding outside the lock but doesn’t fully satisfy its stated objective. The cached message/address should be captured under the lock and forwarded afterward, as the new prepare_send path does.

simpsont-oci: We might be OK with just the improvement in this PR though... so maybe we just change the PR description a bit to match actual changes being made.

sonndinh: That's correct -- this change applies to the messages sent from the local client participants. Further improvement could target messages from the application participant. I updated the description to better reflect the status of this change.

simpsont-oci (approved):

PR #5250, opened 2026-06-30
Reduce lock scope in `match_authenticated`

sonndinh:
This attempts to improve the concurrency bounded by the Spdp lock by reducing the duration when the lock is held in `match_authenticated` while it makes calls to security operations.

Specifically, the lock is released while the security calls are made to the security plugins, such as authentication and access control, with data copies from the shared data structures protected by the lock. When the security operations are done, the lock is re-acquired to reconcile with other concurrent changes while the lock was released.

### DIFF
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -134,6 +134,31 @@ namespace {
     const char* const value = prop.value.in();
     return std::strcmp(value, "0") && ACE_OS::strcasecmp(value, "false");
   }
+
+#if OPENDDS_CONFIG_SECURITY
+  // Helper function to cleanup allocated security handles when authentication fails
+  // or completes elsewhere. Called when: participant is removed, crypto fails,
+  // or another thread already completed the match.
+  inline void cleanup_security_handles(
+      DDS::Security::Authentication_var auth,
+      DDS::Security::AccessControl_var access,
+      DDS::Security::CryptoKeyFactory_var key_factory,
+      DDS::Security::SharedSecretHandle_var shared_secret_handle,
+      DDS::Security::PermissionsHandle permissions_handle,
+      DDS::Security::ParticipantCryptoHandle dp_crypto_handle,
+      DDS::Security::SecurityException& se)
+  {
+    if (shared_secret_handle.in() != 0) {
+      auth->return_sharedsecret_handle(shared_secret_handle.in(), se);
+    }
+    if (permissions_handle != DDS::HANDLE_NIL) {
+      access->return_permissions_handle(permissions_handle, se);
+    }
+    if (dp_crypto_handle != DDS::HANDLE_NIL) {
+      key_factory->unregister_participant(dp_crypto_handle, se);
+    }
+  }
+#endif
 }
 
 void Spdp::init(DDS::DomainId_t /*domain*/,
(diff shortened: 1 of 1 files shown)


dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2108,104 +2133,176 @@ Spdp::match_authenticated(const DCPS::GUID_t& guid, DiscoveredParticipantIter& i
...
+        }
+        crypto_ok = false;
+      }
     }
-    return false;
   }
 
-  if (!auth->get_authenticated_peer_credential_token(
-      iter->second.authenticated_peer_credential_token_, iter->second.handshake_handle_, se)) {
+  iter = participants_.find(guid);
```
simpsont-oci, line 2252: Generally, this is fine, but since we've lost the lock and things might have changed, the lookup can always be 'participants_.end()'. Since this assignment is modifying the input reference, it can impact callers as well. If you look above at line 998, you see this:
```
          if (!match_authenticated(guid, iter)) {
            purge_discovered_participant(iter);
            participants_.erase(iter);
```
In particular, passing 'participants_.end()' into 'participants_.erase()' will cause issues (likely crash or corruption).

simpsont-oci (changes requested):

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -134,6 +134,31 @@ namespace {
...
+#if OPENDDS_CONFIG_SECURITY
+  // Helper function to cleanup allocated security handles when authentication fails
+  // or completes elsewhere. Called when: participant is removed, crypto fails,
+  // or another thread already completed the match.
+  inline void cleanup_security_handles(
+      DDS::Security::Authentication_var auth,
+      DDS::Security::AccessControl_var access,
+      DDS::Security::CryptoKeyFactory_var key_factory,
+      DDS::Security::SharedSecretHandle_var shared_secret_handle,
+      DDS::Security::PermissionsHandle permissions_handle,
```
simpsont-oci, line 147: I'm guessing these _var parameters could all be const&, yes?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -134,6 +134,31 @@ namespace {
...
+      DDS::Security::SecurityException& se)
+  {
+    if (shared_secret_handle.in() != 0) {
+      auth->return_sharedsecret_handle(shared_secret_handle.in(), se);
+    }
+    if (permissions_handle != DDS::HANDLE_NIL) {
+      access->return_permissions_handle(permissions_handle, se);
+    }
+    if (dp_crypto_handle != DDS::HANDLE_NIL) {
+      key_factory->unregister_participant(dp_crypto_handle, se);
```
simpsont-oci, line 158: Should we be checking return values here, even if it's just so the first failure's SE doesn't get overwritten by the 2nd and 3rd failure's SE?

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -2108,104 +2133,176 @@ Spdp::match_authenticated(const DCPS::GUID_t& guid, DiscoveredParticipantIter& i
...
+  // Write results back to participant
+  iter->second.shared_secret_handle_ = shared_secret_handle;
+  iter->second.authenticated_peer_credential_token_ = authenticated_peer_credential_token;
+  iter->second.permissions_handle_ = permissions_handle;
+  iter->second.crypto_tokens_ = crypto_tokens;
+
+  handle_registry->insert_remote_participant_permissions_handle(guid, permissions_handle);
+
+  if (dp_crypto_handle != DDS::HANDLE_NIL) {
+    handle_registry->insert_remote_participant_crypto_handle(guid, dp_crypto_handle);
```
simpsont-oci, line 2295: Does this reinsert even when `dp_crypto_handle` was pre-existing (not freshly registered)? I think that's still safe, but it could maybe use a comment saying so or change the guard to `dp_crypto_handle_owned != DDS::HANDLE_NIL`.

PR #5256, opened 2026-07-16
Backwards Compatible Fixes for Spec-compliant Duration_t Encoding

simpsont-oci:
Problem:

PID_DEADLINE, PID_LATENCY_BUDGET, PID_LIVELINESS, PID_RELIABILITY, PID_LIFESPAN, PID_TIME_BASED_FILTER, and PID_DURABILITY_SERVICE embed a DDS::Duration_t (plain DDS-IDL {sec, nanosec}) directly in the RTPS Parameter union used for SEDP wire announcements. The generic DDS::Duration_t CDR operator writes/reads `nanosec` verbatim into the wire's `fraction` field, which RTPS 2.5 Sec 9.3.2 defines as units of 1/2^32 second, not raw nanoseconds. This silently corrupts any non-whole-second Duration_t QoS value for both this process and any spec-compliant peer -- e.g. a 250ms LIFESPAN was announced with fraction=250000000 (the literal nanosecond count) instead of the correct ~1073741824.

Solution:

Unfortunately, simply changing the wire format for Duration_t breaks backwards compatibility for OpenDDS users of the QoS policies and configurations making use of those PID values. So we have added a new OpenDDS participant flag, defaulting to off, whose value determines how Duration_t will be encoded by writers as well as how it will be interpreted by readers of non-OpenDDS participants. Readers of OpenDDS participants will check the value of the participant flag and interpret Duration_t accordingly. This preserves a default behavior of backwards compatibility but allows users to configurable enable spec-compliant duration_t encoding for interoperability. Changing the default value of the participant flag to true may be done at some future date (such as with a major version change).

### DIFF
--- a/dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
+++ b/dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
@@ -1101,14 +1101,15 @@ RtpsDiscoveryConfig::sedp_fragment_reassembly_timeout(const DCPS::TimeDuration& 
 CORBA::ULong
 RtpsDiscoveryConfig::participant_flags() const
 {
-  CORBA::ULong flags = TheServiceParticipant->config_store()->get_uint32(
+  const CORBA::ULong flags = TheServiceParticipant->config_store()->get_uint32(
     config_key("PARTICIPANT_FLAGS").c_str(), PFLAGS_THIS_VERSION);
-  if (use_rtps_duration_fraction()) {
-    flags |= PFLAGS_RTPS_DURATION_FRACTION;
-  } else {
-    flags &= ~PFLAGS_RTPS_DURATION_FRACTION;
-  }
-  return flags;
+  // PFLAGS_RTPS_DURATION_FRACTION is exclusively controlled by
+  // USE_RTPS_DURATION_FRACTION (see use_rtps_duration_fraction()) and always
+  // overrides whatever was stored for that bit via participant_flags(CORBA::ULong).
+  // This is unlike every other PARTICIPANT_FLAGS bit, which is raw, authoritative
+  // storage (see e.g. Spdp::init()'s handling of PFLAGS_REFLECT_HEARTBEAT_COUNT).
+  return use_rtps_duration_fraction()
+    ? (flags | PFLAGS_RTPS_DURATION_FRACTION) : (flags & ~PFLAGS_RTPS_DURATION_FRACTION);
 }
 
 void
--- a/dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
+++ b/dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
@@ -1101,8 +1101,14 @@ RtpsDiscoveryConfig::sedp_fragment_reassembly_timeout(const DCPS::TimeDuration& 
 CORBA::ULong
 RtpsDiscoveryConfig::participant_flags() const
 {
-  return TheServiceParticipant->config_store()->get_uint32(config_key("PARTICIPANT_FLAGS").c_str(),
-                                                           PFLAGS_THIS_VERSION);
+  CORBA::ULong flags = TheServiceParticipant->config_store()->get_uint32(
+    config_key("PARTICIPANT_FLAGS").c_str(), PFLAGS_THIS_VERSION);
+  if (use_rtps_duration_fraction()) {
+    flags |= PFLAGS_RTPS_DURATION_FRACTION;
+  } else {
+    flags &= ~PFLAGS_RTPS_DURATION_FRACTION;
+  }
+  return flags;
 }
 
 void
@@ -1110,6 +1116,20 @@ RtpsDiscoveryConfig::participant_flags(CORBA::ULong participant_flags)
 {
   TheServiceParticipant->config_store()->set_uint32(config_key("PARTICIPANT_FLAGS").c_str(),
                                                     participant_flags);
+}
+
+bool
+RtpsDiscoveryConfig::use_rtps_duration_fraction() const
+{
+  return TheServiceParticipant->config_store()->get_boolean(
+    config_key("USE_RTPS_DURATION_FRACTION").c_str(), false);
+}
+
+void
+RtpsDiscoveryConfig::use_rtps_duration_fraction(bool value)
+{
+  TheServiceParticipant->config_store()->set_boolean(
+    config_key("USE_RTPS_DURATION_FRACTION").c_str(), value);
 }
 
 bool
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -3973,6 +3973,19 @@ ACE_CDR::ULong Spdp::get_participant_flags(const DCPS::GUID_t& guid) const
     ? iter->second.pdata_.participantProxy.opendds_participant_flags.bits : PFLAGS_EMPTY;
 }
 
+bool Spdp::participant_uses_rtps_duration_fraction(const DCPS::GUID_t& guid) const
+{
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_,
+                   (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0);
+  const DiscoveredParticipantMap::const_iterator iter =
+    participants_.find(make_part_guid(guid));
+  if (iter != participants_.end() && is_opendds(iter->second.pdata_.participantProxy)) {
+    return (iter->second.pdata_.participantProxy.opendds_participant_flags.bits &
+            PFLAGS_RTPS_DURATION_FRACTION) != 0;
+  }
+  return (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0;
+}
+
 void
 Spdp::remove_lease_expiration_i(DiscoveredParticipantIter iter)
 {
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -3988,7 +3988,7 @@ bool Spdp::participant_uses_rtps_duration_fraction(const DCPS::GUID_t& guid) con
 {
   // participant_flags_ is const (set once at construction), so reading it
   // here does not depend on lock_ -- lock_ only guards participants_, which
-  // participant_uses_rtps_duration_fraction_i() consults below.
+  // participant_uses_rtps_duration_fraction_i() consults.
   const bool local_default = (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0;
   ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_, local_default);
   return participant_uses_rtps_duration_fraction_i(guid);
--- a/dds/DCPS/RTPS/Spdp.cpp
+++ b/dds/DCPS/RTPS/Spdp.cpp
@@ -3973,10 +3973,8 @@ ACE_CDR::ULong Spdp::get_participant_flags(const DCPS::GUID_t& guid) const
     ? iter->second.pdata_.participantProxy.opendds_participant_flags.bits : PFLAGS_EMPTY;
 }
 
-bool Spdp::participant_uses_rtps_duration_fraction(const DCPS::GUID_t& guid) const
+bool Spdp::participant_uses_rtps_duration_fraction_i(const DCPS::GUID_t& guid) const
 {
-  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_,
-                   (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0);
   const DiscoveredParticipantMap::const_iterator iter =
     participants_.find(make_part_guid(guid));
   if (iter != participants_.end() && is_opendds(iter->second.pdata_.participantProxy)) {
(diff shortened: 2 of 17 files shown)


dds/DCPS/RTPS/RtpsDiscoveryConfig.cpp
```diff
@@ -1090,8 +1090,14 @@ RtpsDiscoveryConfig::sedp_fragment_reassembly_timeout(const DCPS::TimeDuration&
...
-  return TheServiceParticipant->config_store()->get_uint32(config_key("PARTICIPANT_FLAGS").c_str(),
-                                                           PFLAGS_THIS_VERSION);
+  CORBA::ULong flags = TheServiceParticipant->config_store()->get_uint32(
+    config_key("PARTICIPANT_FLAGS").c_str(), PFLAGS_THIS_VERSION);
+  if (use_rtps_duration_fraction()) {
+    flags |= PFLAGS_RTPS_DURATION_FRACTION;
+  } else {
+    flags &= ~PFLAGS_RTPS_DURATION_FRACTION;
+  }
+  return flags;
```
sonndinh:
```suggestion
  return use_rtps_duration_fraction() ? (flags | PFLAGS_RTPS_DURATION_FRACTION) : (flags & ~PFLAGS_RTPS_DURATION_FRACTION);
```
I think `flags` can be `const` if we make this change.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3973,6 +3973,19 @@ ACE_CDR::ULong Spdp::get_participant_flags(const DCPS::GUID_t& guid) const
     ? iter->second.pdata_.participantProxy.opendds_participant_flags.bits : PFLAGS_EMPTY;
 }
 
+bool Spdp::participant_uses_rtps_duration_fraction(const DCPS::GUID_t& guid) const
+{
+  ACE_GUARD_RETURN(ACE_Thread_Mutex, g, lock_,
+                   (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0);
```
mitza-oci: The lock_ is guarding participant_flags_ here, so the in-case-of-error return value shouldn't use it

mitza-oci (changes requested):

docs/news.d/rtps-duration-fraction.rst
```diff
@@ -0,0 +1,8 @@
+.. news-prs: 0
```
simpsont-oci:
```suggestion
.. news-prs: 5256
```

dds/DCPS/RTPS/Spdp.h
```diff
@@ -216,6 +217,14 @@ class OpenDDS_Rtps_Export Spdp
...
   VendorId_t get_vendor_id_i(const GUID_t& guid) const;
 
+  /**
+   * Combines get_vendor_id() and participant_uses_rtps_duration_fraction()
+   * into a single locked lookup, for callers that need both for the same guid.
+   */
+  void get_vendor_id_and_duration_encoding(const GUID_t& guid,
+                                            VendorId_t& vendor_id,
+                                            bool& uses_rtps_duration_fraction) const;
+
```
mitza-oci, line 227: Could `vendor_id` be const?

mitza-oci (changes requested):

dds/DCPS/RTPS/Spdp.h
simpsont-oci: Not here, no. Both `vendor_id` and `uses_rtps_duration_fraction` are out parameters here.

mitza-oci: Got it, makes sense that it's "get vendor id" -- I don't mind this out-parameter pattern but std::pair/tuple may be easier to follow?  I guess the downside there is that the bool wouldn't have a name.

dds/DCPS/RTPS/Spdp.cpp
```diff
@@ -3973,6 +3973,27 @@ ACE_CDR::ULong Spdp::get_participant_flags(const DCPS::GUID_t& guid) const
...
+            PFLAGS_RTPS_DURATION_FRACTION) != 0;
+  }
+  return (participant_flags_ & PFLAGS_RTPS_DURATION_FRACTION) != 0;
+}
+
+bool Spdp::participant_uses_rtps_duration_fraction(const DCPS::GUID_t& guid) const
+{
+  // participant_flags_ is const (set once at construction), so reading it
+  // here does not depend on lock_ -- lock_ only guards participants_, which
+  // participant_uses_rtps_duration_fraction_i() consults below.
```
sonndinh:
```suggestion
  // participant_uses_rtps_duration_fraction_i() consults.
```
`participant_uses_rtps_duration_fraction_i` is defined above this function, so maybe "below" could be removed.

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1670,13 +1679,19 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, decode_writer_deadline)
...
+                              type_info.xtypes_type_info_, RDE_FRACTIONAL_SECONDS));
   EXPECT_TRUE(writer_data.ddsPublicationData.deadline.period.sec ==
               writer_data_out.ddsPublicationData.deadline.period.sec);
-  EXPECT_TRUE(writer_data.ddsPublicationData.deadline.period.nanosec ==
-              writer_data_out.ddsPublicationData.deadline.period.nanosec);
+  // Round-tripping through the wire's 1/2^32-sec fraction unit is not
+  // perfectly bijective at nanosecond granularity (double integer
+  // truncation): 35000ns encodes to fraction=150323, which decodes back to
+  // 34999ns, not 35000ns. This 1ns discrepancy is inherent to the format,
+  // not a bug.
```
sonndinh: This sounds like a bug in the spec. Perhaps it's worth filing a bug report if it hasn't been done already?

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1927,6 +1961,102 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, set_writer_lifespan_default)
               writer_data_out.ddsPublicationData.lifespan.duration.nanosec);
 }
 
+TEST(dds_DCPS_RTPS_ParameterListConverter, encode_decode_writer_reliability_infinite)
+{ // DURATION_INFINITE must round-trip through the wire fraction conversion
```
sonndinh:
```suggestion
{
  // DURATION_INFINITE must round-trip through the wire fraction conversion
```
Some test functions below have a similar formatting.

mitza-oci (approved):

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1927,6 +1961,103 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, set_writer_lifespan_default)
...
+  ParameterList fractional_param_list;
+  EXPECT_TRUE(to_param_list(writer_data, fractional_param_list, true, type_info,
+                            RDE_FRACTIONAL_SECONDS));
+  ASSERT_TRUE(is_present(fractional_param_list, PID_LIFESPAN));
+  EXPECT_EQ(0x40000000u,
+            get(fractional_param_list, PID_LIFESPAN).lifespan().duration.nanosec);
+}
+
+TEST(dds_DCPS_RTPS_ParameterListConverter, writer_lifespan_duration_out_of_range_nanosec)
+{ // Duration_t is only spec-valid for nanosec < 1e9, but not every QoS
```
simpsont-oci:
```suggestion
{
  // Duration_t is only spec-valid for nanosec < 1e9, but not every QoS
```

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -2767,15 +2914,59 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, decode_reader_reliability)
...
               reader_data_out.ddsSubscriptionData.reliability.max_blocking_time.sec);
-  EXPECT_TRUE(reader_data.ddsSubscriptionData.reliability.max_blocking_time.nanosec ==
-              reader_data_out.ddsSubscriptionData.reliability.max_blocking_time.nanosec);
+  // See decode_writer_deadline: fraction round-trip is lossy by 1ns for this
+  // magnitude (100ns -> fraction=429 -> 99ns).
+  EXPECT_TRUE(reader_data_out.ddsSubscriptionData.reliability.max_blocking_time.nanosec == 99);
+}
+
+TEST(dds_DCPS_RTPS_ParameterListConverter, encode_reader_time_based_filter)
+{ // Should encode reader time based filter using the RTPS wire fraction unit,
```
simpsont-oci:
```suggestion
{
  // Should encode reader time based filter using the RTPS wire fraction unit,
```

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -2767,15 +2914,59 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, decode_reader_reliability)
...
+                            RDE_FRACTIONAL_SECONDS, map));
+  EXPECT_TRUE(is_present(param_list, PID_TIME_BASED_FILTER));
+  Parameter param = get(param_list, PID_TIME_BASED_FILTER);
+  EXPECT_TRUE(param.time_based_filter().minimum_separation.sec == 3);
+  // 40000ns -> wire fraction (40000 << 32) / 1e9 = 171798.
+  EXPECT_TRUE(param.time_based_filter().minimum_separation.nanosec == 171798);
+}
+
+TEST(dds_DCPS_RTPS_ParameterListConverter, decode_reader_time_based_filter)
+{ // Should decode reader time based filter back from the RTPS wire fraction
```
simpsont-oci:
```suggestion
{
  // Should decode reader time based filter back from the RTPS wire fraction
```

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
```diff
@@ -1927,6 +1961,103 @@ TEST(dds_DCPS_RTPS_ParameterListConverter, set_writer_lifespan_default)
...
+
+  DiscoveredWriterData writer_data_out;
+  EXPECT_TRUE(from_param_list(param_list, VENDORID_OPENDDS, writer_data_out,
+                              true, type_info.xtypes_type_info_));
+  EXPECT_EQ(static_cast<CORBA::ULong>(DDS::DURATION_INFINITE_NSEC),
+            writer_data_out.ddsPublicationData.reliability.max_blocking_time.nanosec);
+}
+
 TEST(dds_DCPS_RTPS_ParameterListConverter, encode_writer_user_data)
 { // Should encode writer user data
```
simpsont-oci:
```suggestion
{
  // Should encode writer user data
```

tests/unit-tests/dds/DCPS/RTPS/ParameterListConverter.cpp
simpsont-oci: It's an inconsistency in the spec, for sure. DDS::Duration_t (from DDS spec) uses direct nanosecond representation, but RTPS::Duration_t (from RTPS spec) uses the NTP "fraction of a second" representation. So the represenations don't always cleanly map back and forth to one another. But it's an implementation detail how these are stored internally, and much like with floating point values, it's probably a mistake to (like this test) do direct value comparisons for this very reason. I'll relax the test comparison.

sonndinh: I'm okay with comparing against the final value from the round-trip conversion, e.g. 34999 ns in this test case, since that's how OpenDDS converts the values and we're testing for OpenDDS.

Claude suggested to me that rounding, as opposed to truncating, would give back the original value in this round-trip conversion testing. The suggested conversion is:
```#include <stdint.h>

uint32_t ns_to_fraction(uint32_t ns) {
    // ns * 2^32 fits in 64 bits (max ~4.3e18, well under UINT64_MAX)
    uint64_t scaled = (uint64_t)ns * 4294967296ULL;
    return (uint32_t)((scaled + 500000000ULL) / 1000000000ULL); // round-to-nearest
}

uint32_t fraction_to_ns(uint32_t fraction) {
    uint64_t scaled = (uint64_t)fraction * 1000000000ULL;
    return (uint32_t)((scaled + (1ULL << 31)) >> 32); // round-to-nearest
}
```

In this test case, the converted values would be:
fraction = round(35000 × 4294967296 / 1e9) = round(150323.855...) = 150324
back to nanoseconds:     round(150324 × 1e9 / 4294967296) = round(34999.90...) = 35000

I think that would be a change to the current conversion functions: https://github.com/OpenDDS/OpenDDS/blob/master/dds/DCPS/Time_Helper.inl#L287-L297

mitza-oci (approved):

jwillemsen:
Would be nice when the default could be changed at compile time, easier for new systems

dds/DCPS/Time_Helper.inl
```diff
@@ -287,7 +287,16 @@ bool non_negative_duration(const DDS::Duration_t& t)
 ACE_INLINE OpenDDS_Dcps_Export
 ACE_UINT32 uint32_fractional_seconds_to_nanoseconds(ACE_UINT32 fraction)
 {
-  return static_cast<ACE_UINT32>((static_cast<ACE_UINT64>(fraction) * 1000000000) >> 32);
+  const ACE_UINT64 nanoseconds =
+    (static_cast<ACE_UINT64>(fraction) * 1000000000 +
+     (static_cast<ACE_UINT64>(1) << 31)) >> 32;
```
sonndinh, line 292: I may have missed whether this was discussed before -- does the `nanoseconds_to_uint32_fractional_seconds` function below also need to do rounding for the round-trip conversion to return to the original value?

simpsont-oci: No. The number of fractional values (2^32) is much larger than the number of nanosecond values (10^9), so it's only when coming back to nanoseconds that rounding is important. The tests now check for exact values from the round trip conversion.

simpsont-oci:
> Would be nice when the default could be changed at compile time, easier for new systems

Applications can always programmatically force the configuration flag if they'd prefer that behavior. A major version change like 4.0 would have more leeway for forcing a breaking change like this to become the new default behavior.

sonndinh (approved):

PR #5260, opened 2026-07-17
Map Ownership By Topic Name

### LINKED ISSUE #5249: Data reader handles should not be based on the type name as the key
Problem: Data Reader Handle Key Conflict

When Reader QoS ownership is set to **exclusive** and data keys are the same, OpenDDS fails to receive data on multiple topics if the data type is the same. Only one reader can receive data; other readers get “is not elected as owner” error messages. The current implementation uses the type name as the key for handle management, causing conflicts when multiple topics use the same data type and key value.

1. Environment

| Component         | Version/Details                               |
|-------------------|-----------------------------------------------|
| OpenDDS Version   | 3.34.0                                        |
| OS                | Red Hat Enterprise Linux 8.10 (Ootpa), x86-64 |
| Language          | C++                                           |
| Transport         | RTPS                                          |

2. Setup

| Entity  | QoS Ownership |
|---------|---------------|
| Writer  | Exclusive     |
| Reader  | Exclusive     |

 3. IDL

```idl
module MyModule {
  @topic
  struct CommonPayload {
    @key long id;
    string message;
  };
};
```
4 . Test Scenario

### Publishers

| Publisher | Topic  | Data Type     | id Value |
|-----------|--------|--------------|----------|
| WriterA   | TopicA | CommonPayload | 1        |
| WriterB   | TopicB | CommonPayload | 1        |

### Subscribers

| Subscriber | Topic  | Data Type     |
|------------|--------|--------------|
| ReaderA    | TopicA | CommonPayload |
| ReaderB    | TopicB | CommonPayload |

5. Test Results

- ReaderB fails to receive data and gets “is not elected as owner” error messages.

6. Observation

- When data arrives on TopicA:
  - `DataReaderImpl_T.h` (line 1828): handle returns as 0 from `owner_manager->get_instance_map`
  - `DataReaderImpl_T.h` (line 1843): creates a new handle using `get_next_handle(key)`
  - `DataReaderImpl_T.h` (line 1881): stores the handle with `owner_manager->set_instance_map`
- When data arrives on TopicB:
  - `DataReaderImpl_T.h` (line 1828): gets ReaderA handle from `owner_manager->get_instance_map`
  - Since both ReaderA and ReaderB have the same type name, the returned handle belongs to TopicA's DataReader, causing the “is not elected as owner” error message.

7.  Summary

- The current implementation uses the type name as the key for handle management, which causes conflicts when multiple topics use the same data type and key value.
- Handle management should distinguish between topics, not just type names, to avoid ownership conflicts.

simpsont-oci:
Changes:
- Exclusive-ownership shared instance maps are now keyed by topic name instead of type name.
- Renamed TypeInstanceMap and related variables/comments to topic-oriented terminology.
- Updated lookup, insertion, release, and reader-cleanup call sites.
- Added an RTPS regression scenario with:
  - Two topics
    - The same generated type
    - Identical key values
    - Equal ownership strengths
    - One reader and writer per topic
    - Assertions that both readers receive their topic’s samples
  - Preserved the existing same-topic/two-reader coverage.

Fixes: #5249

### DIFF
--- a/dds/DCPS/DataReaderImpl.cpp
+++ b/dds/DCPS/DataReaderImpl.cpp
@@ -160,7 +160,7 @@ DataReaderImpl::cleanup()
 #ifndef OPENDDS_NO_OWNERSHIP_KIND_EXCLUSIVE
   OwnershipManagerPtr owner_manager = this->ownership_manager();
   if (owner_manager) {
-    owner_manager->unregister_reader(topic_servant_->type_name(), this);
+    owner_manager->unregister_reader(topic_servant_->topic_name(), this);
   }
 #endif
 
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -1250,7 +1250,7 @@ protected:
     if (owner_manager) {
       ACE_GUARD(ACE_Recursive_Thread_Mutex, instance_guard, instances_lock_);
 
-      SharedInstanceMap_rch inst = dynamic_rchandle_cast<SharedInstanceMap>(owner_manager->get_instance_map(topic_servant_->type_name(), this));
+      SharedInstanceMap_rch inst = dynamic_rchandle_cast<SharedInstanceMap>(owner_manager->get_instance_map(topic_servant_->topic_name(), this));
       if (inst != 0) {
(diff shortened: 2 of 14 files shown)


mitza-oci (commented):
Please add NEWS, otherwise looks good

mitza-oci (approved):

Issue #5255, opened 2026-07-15
Interest in runtime loadable external DDS Security plugins?


fj-blanco:
Hi,

I am working on external DDS Security plugins for OpenDDS. My use case is a custom Authentication plugin combined with OpenDDS’s built-in Access Control, Cryptographic, and Utility implementations.

Before opening a PR, I wanted to ask whether this functionality and the proposed approach would be of interest to the project.

The implementation adds:

- Loading security plugins from shared libraries through ACE’s service-loading mechanism.
- Named security configurations.
- Per-interface composition of external and built-in implementations.
- Selection of a global security configuration from the OpenDDS configuration file.
- Validation and error handling for invalid configurations, missing libraries, and missing plugin interfaces.

The OpenDDS changes are contained in a single commit here: [External security plugin support](https://github.com/fj-blanco/OpenDDS/commit/ca619729ff)

I am using this work to develop an [external OpenDDS Authentication plugin](https://github.com/qursa-uc3m/pqsec-dds/tree/feature/multi-dds-layout/adapters/opendds).

I also noticed that OpenDDS’s current [`DdsSecurityCore.idl`](https://github.com/OpenDDS/OpenDDS/blob/master/dds/DdsSecurityCore.idl) is based on DDS Security 1.1, while the current specification is [DDS Security 1.2](https://www.omg.org/spec/DDS-SECURITY/1.2). This work exposes OpenDDS’s existing security interfaces and does not claim DDS Security 1.2 compatibility.

I am not sure whether adding external plugin support to the current 1.1-based implementation is useful to the project, or whether the security SPI should first be updated to 1.2. If there is interest, I can prepare a focused PR for the external plugin support. I would also be willing to help with adapting it to a future DDS Security 1.2 interface.

I would appreciate feedback on this.

Thank you.

mitza-oci:
In general, contributions like this are certainly welcome.

Maybe we should discuss the 1.1 to 1.2 update as a separate issue.  Do you have an idea of the scope?

fj-blanco:
@mitza-oci 

Thanks! I'll open the PR for the plugin loading support shortly.

Regarding 1.1 vs 1.2, after going through the 1.2 spec, I think the PR is mostly orthogonal to the SPI changes. It only deals with loading and composing plugin implementations, which doesn't depend on which version of the interfaces they express, when the SPI moves to 1.2, externally loaded plugins would just implement the extended interfaces like the built-in ones.

[...]

PR #5262, opened 2026-07-21
Support runtime-loadable external DDS Security plugins

fj-blanco:
Implements the proposal in #5255 by adding runtime-loadable external DDS Security plugins. This allows a custom implementation of one security interface to be combined with the built-in implementations of the remaining interfaces.

- New `[security_plugin/<name>]` sections declare a plugin shared library and its ACE service loader class. Libraries are loaded on demand through  `ACE_Service_Config`, using the same mechanism as the built-in plugins.
- `[security/<name>]` sections gain `AuthConfig`, `AccessCtrlConfig`,  `CryptoConfig`, and `UtilityConfig` settings that select the implementation used for each interface. Each defaults to `BuiltIn`.
- A new `DCPSGlobalSecurityConfig` common setting selects a process-wide security configuration, analogous to `DCPSGlobalTransportConfig`.
- `SecurityRegistry::load_security_configuration()` existed but was not invoked during normal initialization. It is now called by `Service_Participant`.
- `create_config()` now creates the Utility interface instead of passing a nil handle, and reports an error when a selected plugin does not provide its requested interface.
- Plugin names, loader names, and library paths are validated before being used in ACE service directives. Missing libraries and directives are reported instead of being silently ignored.
- The existing SecurityConfig unit test was not included in the UnitTests build and used an obsolete configuration-loading API. This change enables the test, updates it to use `ConfigStore`, and removes its now-unused INI file.
- The new configuration format is documented in the Developer's Guide, including the property-name canonicalization caveat: exact punctuation-sensitive properties such as `dds.sec.*` should be supplied through participant QoS instead of `[security]` sections.

Tested with:

- the re-enabled SecurityConfig unit tests;
- security-enabled builds of the changed core sources;
- security-disabled compile checks for the `OPENDDS_CONFIG_SECURITY` guards;
- an on-demand external shared-library loading smoke test; and
- an external Authentication plugin combined with the built-in AccessControl,
  Cryptographic, and Utility implementations.

This PR exposes OpenDDS's current DDS Security 1.1-based C++ interfaces. It does not claim DDS Security 1.2 compatibility or a stable plugin ABI across OpenDDS versions.

### Problem report

[PROBLEM-REPORT-FORM.md](https://github.com/user-attachments/files/30238919/PROBLEM-REPORT-FORM.md)


docs/news.d/external-security-plugins.rst
```diff
@@ -0,0 +1,10 @@
+# This will have to be replaced with the PR number after the PR is created:
+.. news-prs: 0
```
simpsont-oci:
```suggestion
.. news-prs: 5262
```

docs/devguide/run_time_configuration.rst
```diff
@@ -2445,6 +2453,91 @@ The second implicit transport configuration is defined whenever an OpenDDS confi
...
+  .. prop:: Loader=<class_name>
+    :required:
+
+    Name of the ``ACE_Service_Object`` loader class. The library must export
+    its ACE factory using ``ACE_FACTORY_DEFINE``.
+
+.. sec:: security/<config_name>
+
+  Configuration names are case-sensitive, including the value supplied to
+  :prop:`DCPSGlobalSecurityConfig`.
```
simpsont-oci: This is causing an error. Typo? Or maybe missing '..' syntax?

dds/DCPS/security/framework/SecurityRegistry.cpp
```diff
@@ -17,20 +17,55 @@
...
+
+bool is_library_path(const DCPS::String& value)
+{
+  if (value.empty()) {
+    return false;
+  }
+  for (size_t i = 0; i != value.size(); ++i) {
+    const char c = value[i];
+    const bool drive_separator = c == ':' && i == 1 &&
+      ACE_OS::ace_isalpha(static_cast<unsigned char>(value[0]));
```
simpsont-oci: Warnings here to clean up. You likely want to use something like ACE_TEXT_CHAR_TO_TCHAR, as DCPS::String is always char-based and TCHAR is variably sized.

simpsont-oci (changes requested):

fj-blanco:
Addressed the review comments in 6c90219. Corrected the PR number and documentation property reference, and updated validation to use ACE TCHAR conversion. Rebased onto the current master. The security/wchar build and all unit tests pass. Ready for review.

PR #5264, opened 2026-07-25
Fix classic C++ mapping code generation for maps with bounded string keys

### LINKED ISSUE #5263: OpenDDS generates invalid C++ for `map<string<8>, long, 4>` in classic IDL-to-C++ mapping
```
This is OpenDDS version 3.34.0 (NOT A RELEASE)

    OpenDDS VERSION: 3.34.0
    TAO VERSION: 2.5.24

    Ubuntu 22.04.5 LTS (x86_64)

    N/A

    g++ (Ubuntu 13.3.0-6ubuntu2~22.04) 13.3.0

    CONTENTS OF $ACE_ROOT/ace/config.h [if you use a link to a platform-
    specific file, simply state which one]:
    config-linux.h (via symbolic link)

    platform_linux.GNU (via symbolic link)

    N/A - using default MPC configuration

    MEDIUM

    AREA/CLASS/EXAMPLE AFFECTED:
[What example failed?  What module failed to compile?]
    OpenDDS IDL-to-C++ classic mapping code generation for IDL maps with bounded string keys.
    Affected generated files:
    - mutatedC.cpp
    - mutatedTypeSupportC.cpp
    - mutatedTypeSupportImpl.cpp

    The compilation failure occurs in the OpenDDS TypeSupport and DynamicData adapter code generated for structs containing `map<string<8>, long, 4>`.

    DOES THE PROBLEM AFFECT:
        COMPILATION?
            If so, what do your $ACE_ROOT/ace/config.h and
            $ACE_ROOT/include/makeinclude/platform_macros.GNU contain?
            YES. config.h contains config-linux.h (symlink). platform_macros.GNU contains platform_linux.GNU (symlink).
        LINKING?
            NO.
        EXECUTION?
            NO.
        OTHER (please specify)?
            Code generation completes successfully (opendds_idl and tao_idl return 0), but the generated C++ code fails to compile.

    SYNOPSIS:
[Brief description of the problem]
    OpenDDS 3.34.0 generates invalid C++ for IDL map with bounded string key (`map<string<8>, long, 4>`). The IDL frontend and TAO code generation complete successfully, but the generated C++ is internally inconsistent and fails to compile. The map key type is inconsistently represented as `TAO::String_Manager` in some places and `OpenDDS::DCPS::String` or invalid `::char *` in others.

    DESCRIPTION:
[Detailed description of problem.  Don't just say "<blah>
doesn't work, here's a fix," explain what your program does
to get to the <blah> state. ]

    The following IDL is processed through the OpenDDS code generation pipeline:

    ```idl
    module test {
      struct S5_B2_StringBound_0000 {
        string<4> name;
        sequence<long, 5> values;
        map<string<8>, long, 4> map_probe;
      };
    };
    ```

    The `opendds_idl` and `tao_idl` stages all complete with exit code 0. However, when the generated C++ files are compiled with g++ -std=c++17, compilation fails with multiple errors.

    The root cause is that different parts of the code generator use inconsistent C++ type mappings for the map key:

    1. The TAO classic mapping generates the map type as:
       ```cpp
       typedef std::map< ::TAO::String_Manager, ::CORBA::Long>
         _tao_map_CORBA_STRING_8_CORBA_Long__4;
       ```

    2. The OpenDDS TypeSupport code attempts to use `OpenDDS::DCPS::String` as the key type:
       ```cpp
       String key3;
       ...
       value.map_probe[key3]  // fails because map key is TAO::String_Manager
       ```

    3. The generated marshal code uses `.size()` on the key type, but `TAO::String_Manager` has no `.size()` method.

    4. The DynamicData adapter generates invalid C++:
       ```cpp
       ::char * key;
       ::char * id_to_key(DDS::MemberId) { return ::char *(); }
       ```

    This was found through differential IDL/codegen fuzzing comparing OpenDDS, Fast DDS, and CycloneDDS behavior.

    REPEAT BY:
[What you did to get the error; include test program or session
transcript if at all possible. ]

    Save the following IDL as `mutated.idl`:

    ```idl
    module test {
      struct S5_B2_StringBound_0000 {
        string<4> name;
        sequence<long, 5> values;
        map<string<8>, long, 4> map_probe;
      };
    };
    ```

    Run the following commands (adjust paths as needed):

    ```bash
    export OPENDDS=/path/to/opendds/install
    export DDS_ROOT="$OPENDDS/share/dds"
    export ACE_ROOT="$OPENDDS/share/ace"
    export TAO_ROOT="$OPENDDS/share/tao"
    export PATH="$OPENDDS/bin:$PATH"

    mkdir -p build
    cp mutated.idl build/

    opendds_idl -o build build/mutated.idl
    tao_idl --idl-version 4 -as -o build build/mutated.idl
    tao_idl --idl-version 4 -as -I. -Ibuild -I"$DDS_ROOT" -o build build/mutatedTypeSupport.idl

    c++ -std=c++17 \
      -I "$OPENDDS/include" -I build \
      build/mutatedC.cpp \
      build/mutatedTypeSupportC.cpp \
      build/mutatedTypeSupportImpl.cpp \
      -L "$OPENDDS/lib" \
      -Wl,-rpath,"$OPENDDS/lib" \
      -lOpenDDS_Dcps -lTAO_Valuetype -lTAO_AnyTypeCode -lTAO -lACE \
      -fPIC -shared \
      -o build/libtypesupport.so
    ```

    The compilation will fail with errors including:

    ```
    mutatedC.cpp:118:55: error: ‘_tc__tao_map_CORBA_STRING_8_CORBA_Long__4’ is not a member of ‘test::S5_B2_StringBound_0000’
    mutatedTypeSupportImpl.cpp:99:53: error: no match for ‘operator[]’ operand types are ‘std::map<TAO::String_Manager_T<char>, int>’ and ‘OpenDDS::DCPS::String’
    mutatedTypeSupportImpl.cpp:409:23: error: ‘const class TAO::String_Manager_T<char>’ has no member named ‘size’
    mutatedTypeSupportImpl.cpp:456:8: error: expected id-expression before ‘char’ ::char * key;
    mutatedTypeSupportImpl.cpp:1017:6: error: expected id-expression before ‘char’ ::char * id_to_key(DDS::MemberId)
    ```

    SAMPLE FIX/WORKAROUND:
[If available ]

    No workaround is currently available for classic IDL-to-C++ mapping. The IDL cannot be used with OpenDDS as-is.

    Suggested fix directions:

    1. Use a single canonical C++ type mapping for map keys across all generated TypeSupport paths. For classic mapping, string/wstring map keys should be handled as `TAO::String_Manager` / `TAO::WString_Manager`.

    2. `serialized_size()` for classic string map keys should use a classic-mapping-safe accessor, for example `.in()` plus `ACE_OS::strlen()`, instead of `.size()`.

    3. Map deserialization should read string keys into a type assignable/comparable with the generated `std::map` key type.

    4. DynamicData adapter generation for maps should generate valid key types or be omitted if maps are unsupported there.

    5. TypeCode generation should not reference an undeclared anonymous map TypeCode.

    If bounded string keys in IDL maps are intentionally unsupported in classic mapping, OpenDDS should reject the IDL during `opendds_idl` processing with a clear diagnostic instead of generating invalid C++ that fails to compile.
```

wuyuqa:
Title:
Fix classic C++ mapping code generation for maps with bounded string keys
Fixes #5263
Summary:
This fixes OpenDDS IDL code generation for classic C++ mapping when a map uses a bounded string as its key, for example:

```idl
module test {
  struct ClassicMapStringKey {
    map<string<8>, long, 4> map_probe;
  };
};
```
Before this change, generated TypeSupport code used C++11/string-style assumptions in classic mapping code paths. This caused generated C++ to fail to compile due to mismatches between OpenDDS::DCPS::String, raw string expressions, and the actual TAO classic mapping type TAO::String_Manager.
Fixes:
Use string_type() for map string key/value temporary types so classic mapping uses TAO::String_Manager / TAO::WString_Manager.
Use .in() for classic string serialized-size calculations instead of .size().
Use .out() with ACE CDR bounded string extraction for classic map string keys/values.
Fix DynamicData adapter map key helper generation for classic string keys.
Add a classic mapping regression test for map<string<8>, long, 4>.
Verification:
Rebuilt opendds_idl.
Regenerated TypeSupport for the new classic map test.
Confirmed generated code now uses TAO::String_Manager and ACE_InputCDR::to_string(key.out(), 8).
Compiled the generated TypeSupport and regression main.
Ran the round-trip regression executable successfully.
Also regenerated and compiled the existing C++11 maps TypeSupport to check for regression.
Note:
The regression test uses TAO_IDL_OPTIONS -St to avoid an independent TAO TypeCode generation issue for IDL maps. This PR is scoped to OpenDDS TypeSupport/codegen for classic map string keys.

### DIFF
--- a/dds/idl/dynamic_data_adapter_generator.cpp
+++ b/dds/idl/dynamic_data_adapter_generator.cpp
@@ -249,8 +249,8 @@ namespace {
         elem = scoped(map_node->key_type()->name());
       }
       const Classification key_cls = classify(map_node->key_type());
-      if ((key_cls & CL_STRING) && be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11) {
-        elem = (key_cls & CL_WIDE) ? "std::wstring" : "std::string";
+      if (key_cls & CL_STRING) {
+        elem = string_type(key_cls);
       }
       RefWrapper key(map_node->value_type(), elem, "key");
       be_global->impl_ <<
--- a/dds/idl/marshal_generator.cpp
+++ b/dds/idl/marshal_generator.cpp
@@ -932,10 +932,13 @@ namespace {
     } else if (cls & CL_STRING) {
       be_global->impl_ << indent <<
         "primitive_serialized_size_ulong(encoding, size);\n";
+      const bool use_cxx11 = be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
+      const string get_size = use_cxx11 ? accessor + ".size()"
+        : "ACE_OS::strlen(" + accessor + ".in())";
       const string strlen_suffix = (cls & CL_WIDE)
         ? " * char16_cdr_size;\n" : " + 1;\n";
       be_global->impl_ << indent <<
-        "size += " << accessor << ".size()" << strlen_suffix;
+        "size += " << get_size << strlen_suffix;
     } else {
       RefWrapper elem_wrapper(type, cxx_elem, accessor);
       elem_wrapper.nested_key_only_ = nested_key_only;
@@ -968,6 +971,20 @@ namespace {
       be_global->impl_ <<
         streamAndCheck("<< " + member_wrapper.ref(), streamIndent);
     }
+  }
+
+  std::string generate_input_for_map(Classification member_cls, AST_Type* member_type,
+                                     const std::string& accessor, bool use_cxx11)
+  {
+    if (member_cls & CL_STRING) {
+      if (member_cls & CL_BOUNDED) {
+        const string args = accessor + (use_cxx11 ? ", " : ".out(), ")
+          + bounded_arg(member_type);
+        return getWrapper(args, member_type, WD_INPUT);
+      }
+      return accessor + (use_cxx11 ? "" : ".out()");
+    }
+    return accessor;
   }
 
   void gen_map_i(UTL_ScopedName* tdname, AST_Map* map, bool nested_key_only, const FieldInfo* anonymous = 0)
@@ -1010,11 +1027,11 @@ namespace {
       val_cxx_elem = scoped(val->name());
     }
     const bool use_cxx11 = be_global->language_mapping() == BE_GlobalData::LANGMAP_CXX11;
-    if ((key_cls & CL_STRING) && use_cxx11) {
-      key_cxx_elem = (key_cls & CL_WIDE) ? "std::wstring" : "std::string";
+    if (key_cls & CL_STRING) {
+      key_cxx_elem = string_type(key_cls);
     }
-    if ((val_cls & CL_STRING) && use_cxx11) {
-      val_cxx_elem = (val_cls & CL_WIDE) ? "std::wstring" : "std::string";
+    if (val_cls & CL_STRING) {
+      val_cxx_elem = string_type(val_cls);
     }
 
     RefWrapper(base_wrapper).done().generate_tag();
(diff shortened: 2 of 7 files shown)


simpsont-oci:
Looks like there are some build errors. @wuyuqa can you take a look?

wuyuqa:
> Looks like there are some build errors. @wuyuqa can you take a look?

Sure. I'll take a look when I have time. I've been quite busy these days.

PR #5266, opened 2026-07-26
Add reproducer for intermittent SHM post-match delivery failure

marble703:
## Problem Report

Related issue: #5274

**OpenDDS version:** 3.35.0-dev, `ac227d5028a64c2250585a28097f49b8912e6985`

**Host and target:** Ubuntu 22.04.5 LTS, Linux 6.8.0-136-generic, x86_64 (same host)

**Compiler:** g++ 12.3.0

**ACE/TAO:** ACE 6.5.24 / TAO 2.5.24, configured with `ace/config-linux.h` and `include/makeinclude/platform_linux.GNU`; OpenDDS configured with `--no-debug --optimize`.

**Urgency:** Medium

**Area affected:** `dds/DCPS/transport/shmem`, association-adjacent delivery with RTPS discovery and best-effort QoS.

**Affects:** Execution of an application using the OpenDDS shared-memory transport.

### Synopsis

After both directions observe matched reader and writer endpoints and a short settle interval completes, a same-host best-effort SHM ping/pong exchange can fail to deliver the request or its echo. The same MWE is stable with RTPS/UDP.

### Description

This PR adds `tests/DCPS/ShmemWakeupRace`, a two-process MWE using only public DDS/OpenDDS application APIs. Ping creates a writer for the ping topic and a reader for the pong topic. Pong creates the inverse pair. Each process waits for its local writer and reader to observe a match, then waits 1.2 seconds before Ping writes exactly one best-effort sample and waits five seconds for the echo.

The runner selects `shmem` or `rtps_udp`; the executable, RTPS discovery configuration, QoS, domain, timing, and process topology are otherwise the same. This eliminates the earlier one-direction-only match barrier as a generic startup confounder.

### Controlled Comparison

One fresh-process 50-trial run on the environment above produced:

| Transport | Trials | Failed |
|---|---:|---:|
| shmem | 50 | 38 |
| rtps_udp | 50 | 0 |

All 50 trials in both arms reported successful bidirectional endpoint matches. Of the 38 SHM failures, Pong timed out before receiving the request in 21 trials. In the remaining 17 failures, Pong echoed the request but Ping did not receive the reply.

A later independent rerun produced 29/50 SHM failures and 0/50 RTPS/UDP failures. All trials again reported successful bidirectional endpoint matches. The exact SHM failure rate varies, but the transport-specific separation was reproduced.

This MWE is intentionally probabilistic and is not presented as a deterministic CI regression test. It also does **not** claim an OMG DDS reliability-contract violation: best-effort QoS does not guarantee delivery.

Related history: #3549 handled an earlier SHM association race. This MWE shows a remaining SHM-specific post-match delivery failure. Separate investigation suggests a wakeup/registration ordering window, but that internal cause is a hypothesis beyond what this public-API MWE alone proves.

### Repeat By

```sh
cd tests/DCPS/ShmemWakeupRace
source ../../../setenv.sh
"$MPC_ROOT/mpc.pl" -type gnuace ShmemWakeupRace.mpc
make -f GNUmakefile.ShmemWakeupRace_Publisher
make -f GNUmakefile.ShmemWakeupRace_Subscriber

perl run_test.pl transport=rtps_udp trials=50
perl run_test.pl transport=shmem trials=50
```

### Expected Result

A healthy trial prints:

```text
Ping: matched pong reader and writer
Pong: matched ping reader and writer
Ping: wrote sample 1
Pong: echoed sample 1
Ping: received pong sample 1
```

The RTPS/UDP control reports `failed trials: 0/N` and exits with status zero.

### Actual Result

On the affected host, SHM intermittently times out after both directions have reported endpoint matches. The runner reports one or more failed trials and exits nonzero.

### Sample Fix/Workaround

No production fix is included in this PR. The MWE is provided to establish the transport-specific failure mode and request guidance on an appropriate transport-level regression strategy.

### DIFF
--- a/tests/DCPS/ShmemWakeupRace/.gitignore
+++ b/tests/DCPS/ShmemWakeupRace/.gitignore
@@ -1,0 +1,17 @@
+/GNUmakefile.ShmemWakeupRace_Publisher
+/GNUmakefile.ShmemWakeupRace_Subscriber
+/ShmemWakeupRaceC.cpp
+/ShmemWakeupRaceC.h
+/ShmemWakeupRaceC.inl
+/ShmemWakeupRaceS.cpp
+/ShmemWakeupRaceS.h
+/ShmemWakeupRaceTypeSupport.idl
+/ShmemWakeupRaceTypeSupportC.cpp
+/ShmemWakeupRaceTypeSupportC.h
+/ShmemWakeupRaceTypeSupportC.inl
+/ShmemWakeupRaceTypeSupportImpl.cpp
+/ShmemWakeupRaceTypeSupportImpl.h
+/ShmemWakeupRaceTypeSupportS.cpp
+/ShmemWakeupRaceTypeSupportS.h
+/publisher
+/subscriber
--- a/tests/DCPS/ShmemWakeupRace/MatchBarrier.h
+++ b/tests/DCPS/ShmemWakeupRace/MatchBarrier.h
@@ -1,0 +1,81 @@
+/*
+ *
+ *
+ * Distributed under the OpenDDS License.
+ * See: http://www.opendds.org/license.html
+ */
+
+#ifndef SHMEM_WAKEUP_RACE_MATCH_BARRIER_H
+#define SHMEM_WAKEUP_RACE_MATCH_BARRIER_H
+
+#include <dds/DCPS/WaitSet.h>
+
+#include <ace/OS_NS_unistd.h>
+#include <ace/Time_Value.h>
+
+namespace ShmemWakeupRace {
+
+inline bool wait_for_match(DDS::DataWriter_ptr writer)
+{
+  DDS::PublicationMatchedStatus match_status;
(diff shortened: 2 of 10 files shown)


marble703:
emm... I don't remember i had enabled copilot

marble703:
Pushed a follow-up addressing the cross-platform build and test-isolation issues:

- Renamed the IDL constant from `DOMAIN` to `DOMAIN_ID` to avoid a macro collision on macOS and Windows.
- Changed the test domain from 113 to the unused domain 228.
- Added explicit `ReadCondition` cleanup and more precise `WaitSet` error reporting.

Local IDL regeneration, build, Perl syntax check, and OpenDDS lint pass.

Revalidation after the isolation and cleanup changes:
- RTPS/UDP: 0/50 failed
- SHM: 35/50 failed
- All trials reported successful bidirectional endpoint matches

The changes did not alter the reproduced transport-specific behavior.

Issue #5232, opened 2026-05-21
Not all CDR extraction results checked


jwillemsen:
Grep `cdr >>` and `cdr <<` in the OpenDDS core, will show multiple places where the return value of the operator>> and operator<< is not checked

SenaDev2023:
Hi @jwillemsen,

I'd like to work on this issue; this would be my first contribution to OpenDDS, so I'd appreciate any guidance on scope or conventions I might be missing.

Starting with `dds/InfoRepo/UpdateManager.cpp` as a first PR.

I audited all 18 `operator>>` call sites in that file. None currently check the extraction's return value; the distinction between sites is in push/extract ordering, not validation:

- **4 sites** (`146`, `176`, `326`, `346`) extract before any use — no corrupted state risk from ordering, but still no validation at all
- **6 sites** in `pushImage()` push a heap-allocated pointer into a `SeqGuard` container *before* the extraction that fills it; if extraction fails, a partially/un-populated object still flows into the resulting `UImage` and reaches `info_->receive_image()`
- **8 sites** extract into stack-local values passed directly as function args (`add_subscription`/`add_publication`/etc.);  different shape again, still unchecked

I'll start with adding proper checks across all 18 and fixing the push-before-validate ordering on the 6 affected sites.

Let me know if `UpdateManager.cpp` first makes sense as a starting point, and if there's a preferred error-handling convention in this file/codebase I should follow (e.g. return-and-log vs. exception vs. something else) before I write the patch.

PR #5275, opened 2026-08-02
Check CDR extraction results in UpdateManager (partial fix for #5232)

SenaDev2023:
## Scope

This is a partial fix for #5232. The audit found 18 total unchecked operator>> sites in this file; the remaining 14 are either structural (a value is pushed into an owned sequence before extraction is validated,
requiring a reorder rather than a bare check) or live in Manager::add(const DActor&), which wasn't covered by the initial audit. Follow-up PR(s) will address those separately to keep this change reviewable.

## Testing

- Full solution build (Debug x64): 0 errors, 0 new warnings
- tools/scripts/lint.pl: no issues in the modified file
- DevGuideExamples/DCPS/Messenger integration test: test PASSED.

### DIFF
--- a/dds/InfoRepo/UpdateManager.cpp
+++ b/dds/InfoRepo/UpdateManager.cpp
@@ -12,7 +12,7 @@
 #include "DCPSInfo_i.h"
 
 #include "tao/CDR.h"
-
+#include "dds/DCPS/debug.h"
 #include <vector>
 
 OPENDDS_BEGIN_VERSIONED_NAMESPACE_DECL
@@ -143,7 +143,14 @@ Manager::pushImage(const DImage& image)
 
     DDS::DomainParticipantQos* qos;
     ACE_NEW_NORETURN(qos, DDS::DomainParticipantQos);
-    in_cdr >> *qos;
+    if (!(in_cdr >> *qos)) {
+      if (OpenDDS::DCPS::log_level >= OpenDDS::DCPS::LogLevel::Error) {
+        ACE_ERROR((LM_ERROR,
+          "(%P|%t) ERROR: UpdateManager::pushImage: failed to demarshal participant qos\n"));
+      }
+      delete qos;
+      continue;
+    }
     part_qos.push_back(qos);
 
     UParticipant* u_part;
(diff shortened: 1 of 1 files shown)


SenaDev2023:
@simpsont-oci @jwillemsen; this is PR #1 of the split we discussed in #5232 (4 safe-order sites, structural reorders to follow in a separate PR). CI's green, ready whenever you have time to take a look. Thank you for your time and consideration.

SenaDev2023:
@jwillemsen  @simpsont-oci Single failing check (test_u22_i0_xer0) is tests/DCPS/Thrasher/run_test.pl high rtps durable, config 3-5-0 — a DCPS-level durability/discovery stress test unrelated to UpdateManager.cpp. All 126 other checks passed, including the same u22 build/test matrix otherwise. Looks like CI flake rather than a regression from this change.

dds/InfoRepo/UpdateManager.cpp
```diff
@@ -12,7 +12,7 @@
 #include "DCPSInfo_i.h"
 
 #include "tao/CDR.h"
-
+#include "dds/DCPS/debug.h"
```
sonndinh:
```suggestion
#include <dds/DCPS/debug.h>
```
I would move this include to line 13 and add a blank line immediately before and after it. So it would look something like:
```
#include "DCPSInfo_i.h"

#include <dds/DCPS/debug.h>

#include <tao/CDR.h>
```

For reference, OpenDDS has development guidelines for header includes: https://opendds.readthedocs.io/en/master/internal/dev_guidelines.html#includes. You can take a look at other files in the codebase for examples of the style. I think the more actively developing files are more likely to adhere to the guidelines, for example, Spdp.cpp (https://github.com/OpenDDS/OpenDDS/blob/master/dds/DCPS/RTPS/Spdp.h).

sonndinh (changes requested):

SenaDev2023:
Thanks, @sonndinh, I appllied fix. Moved dds/DCPS/debug.h above tao/CDR.h, switched both to angle brackets, and added the blank lines between groups per the include ordering guidelines. Rebuilt DCPSInfoRepo_Lib locally to confirm nothing broke; clean build.

SenaDev2023:
@sonndinh, just following up when you get a chance. I pushed the include order/style fix we discussed and rebuilt DCPSInfoRepo_Lib locally to confirm it’s clean. Let me know if you’d like me to re-request review, or if anything else needs adjusting. Thanks!

sonndinh (approved):

sonndinh:
> @sonndinh, just following up when you get a chance. I pushed the include order/style fix we discussed and rebuilt DCPSInfoRepo_Lib locally to confirm it’s clean. Let me know if you’d like me to re-request review, or if anything else needs adjusting. Thanks!

It looks good to me. Thank you!

PR #5278, opened 2026-08-06
Add read_encapsulation_header to tighten encapsulation validity checks

### LINKED ISSUE #5271: OpenDDS accepts `PL_CDR2` encapsulation for an `@appendable` XCDRv2 type instead of rejecting the representation mismatch
## Summary

OpenDDS accepts an XCDRv2 payload whose encapsulation says `PL_CDR2`, even
though the target type is `@appendable` and the body is still the ordinary
delimited appendable layout.

This is a header-only mismatch:

- the body bytes are unchanged
- only the encapsulation identifier changes from `0x0009` to `0x000b`
- the payload is still accepted and decoded as a normal object

Under XCDR2, `@appendable` top-level data should use the delimited form, while
`PL_CDR2` is the parameter-list representation for mutable types.

## Why this matters

The decoder accepts a payload whose declared XCDRv2 representation is
incompatible with the target type. Even if this sample still decodes to the
expected object, the implementation is not enforcing the representation
contract at the point where the encapsulation header is interpreted.

## Affected schema

```idl
module test {
  @appendable
  struct S5_B1_Alignment_0001 {
    boolean flag;
    long counter;
    octet marker;
    double ratio;
  };
};
```

## Payloads

Valid payload (`DELIMIT_CDR2`, little-endian):

```text
00090000140000000100000098ffffffa900000075931804560e1ec0
```

Header-only mismatch payload (`PL_CDR2` header + unchanged appendable body):

```text
000b0000140000000100000098ffffffa900000075931804560e1ec0
```

## Local reproduction

Files in this directory:

- `test.idl`
```text
module test {
  @appendable
  struct S5_B1_Alignment_0001 {
    boolean flag;
    long counter;
    octet marker;
    double ratio;
  };
};
```
- `repro.cpp`
```cpp
#include "build/testTypeSupportImpl.h"

#include <dds/DCPS/EncapsulationHeader.h>

#include <cstring>
#include <iomanip>
#include <iostream>
#include <vector>

static void run_case(
        const char* name,
        const std::vector<unsigned char>& bytes)
{
    ACE_Message_Block mb(bytes.size());
    std::memcpy(mb.wr_ptr(), bytes.data(), bytes.size());
    mb.wr_ptr(bytes.size());

    OpenDDS::DCPS::Serializer ser(&mb, OpenDDS::DCPS::Encoding::KIND_XCDR2);
    OpenDDS::DCPS::EncapsulationHeader encap;
    test::S5_B1_Alignment_0001 sample;

    const bool ok = (ser >> encap) && (ser >> sample);

    std::cout << name << "\n";
    std::cout << "  accepted: " << (ok ? "yes" : "no") << "\n";
    if (ok)
    {
        std::cout << std::boolalpha;
        std::cout << "  decoded : {flag=" << static_cast<bool>(sample.flag)
                  << ", counter=" << sample.counter
                  << ", marker=" << static_cast<unsigned>(sample.marker)
                  << ", ratio=" << std::setprecision(17) << sample.ratio << "}\n";
        std::cout << "  consumed: " << ser.rpos() << "/" << bytes.size() << "\n";
    }
}

int main()
{
    const std::vector<unsigned char> valid = {
        0x00, 0x09, 0x00, 0x00, 0x14, 0x00, 0x00, 0x00,
        0x01, 0x00, 0x00, 0x00, 0x98, 0xff, 0xff, 0xff,
        0xa9, 0x00, 0x00, 0x00, 0x75, 0x93, 0x18, 0x04,
        0x56, 0x0e, 0x1e, 0xc0
    };

    const std::vector<unsigned char> header_only = {
        0x00, 0x0b, 0x00, 0x00, 0x14, 0x00, 0x00, 0x00,
        0x01, 0x00, 0x00, 0x00, 0x98, 0xff, 0xff, 0xff,
        0xa9, 0x00, 0x00, 0x00, 0x75, 0x93, 0x18, 0x04,
        0x56, 0x0e, 0x1e, 0xc0
    };

    run_case("valid DELIMIT_CDR2 payload", valid);
    run_case("issue: header-only mismatch (PL_CDR2 encapsulation, DELIMIT body)", header_only);
    return 0;
}

```
- `build_and_run.sh`
```bash
#!/usr/bin/env bash

if [[ ! -x "$OPENDDS_IDL" ]]; then
  echo "opendds_idl not found or not executable: $OPENDDS_IDL" >&2
  exit 2
fi

if [[ ! -x "$TAO_IDL" ]]; then
  echo "tao_idl not found or not executable: $TAO_IDL" >&2
  exit 2
fi

rm -rf "$BUILD_DIR"
mkdir -p "$BUILD_DIR"

echo "[1/4] opendds_idl"
"$OPENDDS_IDL" -o "$BUILD_DIR" "$SCRIPT_DIR/test.idl"

echo "[2/4] tao_idl for user IDL"
"$TAO_IDL" --idl-version 4 -as -o "$BUILD_DIR" "$SCRIPT_DIR/test.idl"

echo "[3/4] tao_idl for generated TypeSupport IDL"
"$TAO_IDL" --idl-version 4 -as \
  -I"$SCRIPT_DIR" \
  -I"$BUILD_DIR" \
  -I"$DDS_ROOT" \
  -o "$BUILD_DIR" \
  "$BUILD_DIR/testTypeSupport.idl"

echo "[4/4] Build and run"
c++ -std=c++17 \
  -I "$OPENDDS/include" \
  -I "$SCRIPT_DIR" \
  "$SCRIPT_DIR/repro.cpp" \
  "$BUILD_DIR/testC.cpp" \
  "$BUILD_DIR/testTypeSupportC.cpp" \
  "$BUILD_DIR/testTypeSupportImpl.cpp" \
  -L "$OPENDDS/lib" \
  -Wl,-rpath,"$OPENDDS/lib" \
  -lOpenDDS_Dcps -lTAO_Valuetype -lTAO_AnyTypeCode -lTAO -lACE \
  -o "$BUILD_DIR/repro"

"$BUILD_DIR/repro"

```

Run:

```bash
cd pocs/issue1_opendds_header_mismatch
bash build_and_run.sh
```

Observed output on the affected build:

```text
valid DELIMIT_CDR2 payload
  accepted: yes
  decoded : {flag=true, counter=-104, marker=169, ratio=-7.5140000000000002}
  consumed: 28/28
issue: header-only mismatch (PL_CDR2 encapsulation, DELIMIT body)
  accepted: yes
  decoded : {flag=true, counter=-104, marker=169, ratio=-7.5140000000000002}
  consumed: 28/28
```

## Expected behavior

The second payload should be rejected before field decoding, because the
declared XCDRv2 representation is incompatible with the target appendable type.

## Actual behavior

The payload is accepted and decoded as a normal object.

## Other DDS behavior

- FastDDS / Fast-CDR: also accepts this header-only mismatch
- CycloneDDS: not affected on the encapsulation-aware upper path; it rejects the
  pure header-only mismatch

## Code-level analysis

Relevant code locations in OpenDDS:

- `opendds/dds/DCPS/EncapsulationHeader.cpp:81-92`
  - `operator>>(Serializer&, EncapsulationHeader&)` only parses the 4-byte
    header and stores `kind/options`; it does not enforce any expected
    extensibility.
- `opendds/dds/DCPS/EncapsulationHeader.cpp:157-257`
  - `to_encoding_i(...)` and `to_encoding(...)` already contain the logic that
    would reject `PL_CDR2` for an appendable type, but this direct
    deserialization path does not invoke that check before decoding the sample.

In the local PoC, `Serializer ser(..., Encoding::KIND_XCDR2); ser >> encap;
ser >> sample;` accepts a `PL_CDR2` header and then decodes the payload as the
appendable type anyway. That strongly suggests the header is parsed but not
bound to the target type's expected representation.

## Suggested root cause

The deserialization path accepts generic XCDR2 input after parsing the
encapsulation header, but does not apply the existing
`to_encoding(..., expected_extensibility)` validation before decoding the
appendable object.

## Suggested fix

1. After reading the encapsulation header, validate it against the target
   type's expected extensibility-derived representation.
2. Reject `PL_CDR2` when decoding an appendable type that requires the
   delimited XCDR2 form.
3. Add a regression test with the header-only mismatch payload above.

### LINKED ISSUE #5272: OpenDDS accepts invalid appendable XCDRv2 body layout and decodes garbage values
```
This is OpenDDS version 3.34.0 (NOT A RELEASE)

    OpenDDS VERSION: 3.34.0
    TAO VERSION: 2.5.24

    Ubuntu 22.04.5 LTS (x86_64)

    N/A

    g++ (Ubuntu 13.3.0-6ubuntu2~22.04) 13.3.0

    CONTENTS OF $ACE_ROOT/ace/config.h [if you use a link to a platform-
    specific file, simply state which one]:
    config-linux.h (via symbolic link)

    platform_linux.GNU (via symbolic link)

    N/A - using default MPC configuration

    MEDIUM

    AREA/CLASS/EXAMPLE AFFECTED:
[What example failed?  What module failed to compile?]
    OpenDDS XCDRv2 deserialization path for `@appendable` types.
    The issue is a runtime behavioral bug: OpenDDS accepts a malformed
    appendable XCDRv2 body that uses an EMHEADER-style layout (characteristic
    of mutable encoding) and decodes it into a garbage object instead of
    rejecting it.

    This is distinct from the header-mismatch issue (Issue 1A). The
    encapsulation header is correct (`0x0009`, DELIMIT_CDR2), but the body
    structure is invalid for an appendable type.

    Affected component: OpenDDS generated TypeSupport deserialization code for
    `@appendable` structs and the underlying Serializer CDR extraction logic.

    DOES THE PROBLEM AFFECT:
        COMPILATION?
            If so, what do your $ACE_ROOT/ace/config.h and
            $ACE_ROOT/include/makeinclude/platform_macros.GNU contain?
            NO. Code generation and compilation complete successfully.
        LINKING?
            On Unix systems, did you run make realclean first?
            NO.
        EXECUTION?
            YES. OpenDDS decodes a structurally invalid XCDRv2 payload as if
            it were valid, producing a corrupted object without any error
            indication.
        OTHER (please specify)?
            The issue affects runtime deserialization correctness. The decoder
            validates only the DHEADER length, but does not verify that the
            post-DHEADER bytes actually follow appendable layout rules.

    SYNOPSIS:
[Brief description of the problem]
    OpenDDS accepts an XCDRv2 payload for an `@appendable` type where the
    encapsulation header is `0x0009` (`DELIMIT_CDR2`, correct), but the body
    after the DHEADER uses an EMHEADER-style layout that is only valid for
    `PL_CDR2` (mutable types). The decoder consumes the malformed body as
    ordinary fields and produces a garbage object instead of rejecting it.

    This indicates missing body-structure validation in the appendable XCDRv2
    deserialization path.

    DESCRIPTION:
[Detailed description of problem.  Don't just say "<blah>
doesn't work, here's a fix," explain what your program does
to get to the <blah> state. ]

    The following IDL is processed through `opendds_idl`:

    ```idl
    module test {
      @appendable
      struct S5_B1_Alignment_0001 {
        boolean flag;
        long counter;
        octet marker;
        double ratio;
      };
    };
    ```

    Two payloads are constructed:
    1. Valid payload: encapsulation `0x0009` (`DELIMIT_CDR2`), body contains
       fields in appendable declaration order.
    2. Body-only mismatch payload: encapsulation `0x0009` (`DELIMIT_CDR2`),
       but body after DHEADER is rewritten as an EMHEADER-style layout.

    According to OMG DDS-XTypes 1.3 Table 35, `@appendable` + XCDRv2 must use
    `DELIMITED_CDR`. The `DELIMITED_CDR` format requires that fields be encoded
    in declaration order, not with per-member EMHEADERs. The EMHEADER layout
    is specific to `PL_CDR2` (mutable types).

    The decoder should reject the body-only mismatch payload because the body
    bytes do not follow appendable layout rules. However, OpenDDS accepts it,
    consuming the EMHEADER bytes as if they were field data, and produces a
    corrupted object.

    The decoded object is completely wrong:
    - Expected: `{"flag":true, "counter":-104, "marker":169, "ratio":-7.514}`
    - Observed: `{"flag":false, "counter":0, "marker":152, "ratio":6.304e-289}`

    FastDDS/Fast-CDR and CycloneDDS also accept the same body-only mismatch
    on their normal upper paths, indicating this is a common XCDRv2 validation
    weakness across implementations.

    REPEAT BY:
[What you did to get the error; include test program or session
transcript if at all possible. ]

    The PoC directory contains:
    - test.idl
    - repro.cpp
    - build_and_run.sh

    Run the following commands:

    ```bash
    cd pocs/issue1_opendds_body_layout
    bash build_and_run.sh
    ```

    Observed output on the affected build:

    ```text
    valid DELIMIT_CDR2 payload
      accepted: yes
      decoded : {flag=true, counter=-104, marker=169, ratio=-7.5140000000000002}
      consumed: 28/28
    issue: body-only mismatch (DELIMIT encapsulation, EMHEADER-style body)
      accepted: yes
      decoded : {flag=false, counter=0, marker=152, ratio=6.3045695601937706e-289}
      consumed: 32/32
    ```

    The second payload should have been rejected.

    Minimal IDL:

    ```idl
    module test {
      @appendable
      struct S5_B1_Alignment_0001 {
        boolean flag;
        long counter;
        octet marker;
        double ratio;
      };
    };
    ```

    Valid payload bytes:
    ```text
    00090000140000000100000098ffffffa900000075931804560e1ec0
    ```

    Body-only mismatch payload:
    ```text
    0009000018000000000000200000000098ffffffa900000075931804560e1ec0
    ```

    The difference:
    - valid payload DHEADER: `0x14` (20 bytes), then fields in order
    - mismatch payload DHEADER: `0x18` (24 bytes), then EMHEADER `0x20000000`,
      then `0x00000000` (4 bytes as flag data), then remaining bytes consumed
      as other fields

    SAMPLE FIX/WORKAROUND:
[If available ]

    Workaround: None. Applications cannot validate the appendable body
    structure through the normal `Serializer` API because the decoder does not
    expose a structure-validation hook.

    Suggested fix:
    1. In the appendable XCDRv2 deserialization path, after reading the DHEADER
       (`read_delimiter(total_size)`), validate that the body structure matches
       the appendable layout:
       - The body should consist of fields in declaration order.
       - There should be no EMHEADER-style per-member headers.
       - Each field's serialized representation should match the expected type.
    2. Reject bodies that contain EMHEADER-style layout (i.e., bytes that
       resemble mutable member headers) when the target type is appendable.
    3. Add a regression test with the malformed payload above and assert
       deserialization failure.

    Code locations:
    - `opendds/dds/idl/marshal_generator.cpp:2850-3100`: generated appendable
      extraction code calls `read_delimiter(total_size)`, computes
      `end_of_struct`, then reads fields sequentially.
    - `opendds/dds/DCPS/Serializer.inl:874-883`: `read_delimiter(size_t&)`
      reads only the DHEADER value; it does not validate the body layout that
      follows.

    Specification violation:
    OMG DDS-XTypes 1.3 Table 35: `APPENDABLE` + XCDRv2 must use `DELIMITED_CDR`.
    The EMHEADER-style layout is only valid for `PL_CDR2` (mutable types), not
    for `DELIMITED_CDR` (appendable types).
    Section 7.4.3.1: "The encoding format is determined by the encoding version
    and the extensibility kind of the object being serialized."
```

simpsont-oci:
Motivation

read_encapsulation_header() collapses the previous three-step pattern (deserialize the 4-byte encapsulation header, validate its representation against the target type's expected extensibility via to_encoding(), then set the stream's Encoding) into a single call. Previously these were three independent steps a caller had to get right in order; skipping the validation step after reading the header is exactly the misuse pattern described in #5271. All of OpenDDS's actual deserialization paths already performed this validation correctly before this change (each call site already passed the target's expected extensibility into to_encoding()), so this isn't a fix for an observed acceptance of malformed payloads — it's hardening the API so that mistake can't be made by a future call site, and it removes duplicated boilerplate at the same time.

Description of Changes

- Added read_encapsulation_header() to dds/DCPS/EncapsulationHeader.h/.cpp, which reads the header, validates it against an expected extensibility, and sets the stream's encoding in one call. It returns an EncapsulationReadStatus (Ok / HeaderError / ExtensibilityMismatch) rather than bool so callers can still distinguish a truncated/malformed header from a representation mismatch and log accordingly.
- Updated DataReaderImpl_T.h, FilterEvaluator.cpp, and Sedp.cpp to use the new helper, removing the duplicated read/validate/set-encoding boilerplate at each site. Also fixed a pre-existing format-string argument mismatch in DataReaderImpl_T.h's error logging (a %C with no corresponding argument).
- Added regression tests in tests/unit-tests/dds/DCPS/EncapsulationHeader.cpp covering XCDR2 representation-mismatch rejection (e.g. PL_CDR2 header on an @appendable type) and both success/failure paths of read_encapsulation_header().

### DIFF
--- a/dds/DCPS/EncapsulationHeader.h
+++ b/dds/DCPS/EncapsulationHeader.h
@@ -134,7 +134,7 @@ struct EncapsulationReadStatus {
     /// The 4-byte header itself could not be deserialized, e.g. a truncated buffer.
     HeaderError,
     /// The header was read, but its representation doesn't match expected_extensibility.
-    ExtensibilityMismatch
+    ExtensibilityMismatch,
   };
 };
 
--- a/dds/DCPS/EncapsulationHeader.h
+++ b/dds/DCPS/EncapsulationHeader.h
@@ -128,6 +128,35 @@ OpenDDS_Dcps_Export
 bool to_any_encoding(Encoding& encoding,
                      const EncapsulationHeader& header);
 
+struct EncapsulationReadStatus {
+  enum Value {
+    Ok,
+    /// The 4-byte header itself could not be deserialized, e.g. a truncated buffer.
+    HeaderError,
+    /// The header was read, but its representation doesn't match expected_extensibility.
+    ExtensibilityMismatch
+  };
+};
+
+/**
+ * Read encapsulation header from Serializer, validate representation against
+ * expected extensibility, and set the stream's encoding. Leaves the stream's
+ * encoding unchanged unless EncapsulationReadStatus::Ok is returned.
+ */
+OpenDDS_Dcps_Export
+EncapsulationReadStatus::Value read_encapsulation_header(
+  Serializer& s,
+  EncapsulationHeader& header,
+  Extensibility expected_extensibility);
+
+/**
+ * Like the above, for callers that don't need the parsed header.
+ */
+OpenDDS_Dcps_Export
+EncapsulationReadStatus::Value read_encapsulation_header(
+  Serializer& s,
+  Extensibility expected_extensibility);
+
 } // namespace DCPS
 } // namespace OpenDDS
 
--- a/dds/DCPS/DataReaderImpl_T.h
+++ b/dds/DCPS/DataReaderImpl_T.h
@@ -930,8 +930,9 @@ namespace OpenDDS {
       static_cast<Endianness>(sample.header_.byte_order_));
 
     if (encapsulated) {
-      EncapsulationHeader encap;
-      if (!(ser >> encap)) {
+      const EncapsulationReadStatus::Value read_status =
+        read_encapsulation_header(ser, type_support_->base_extensibility());
+      if (read_status == EncapsulationReadStatus::HeaderError) {
         if (DCPS_debug_level > 0) {
           ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
             ACE_TEXT("%CDataReaderImpl::lookup_instance: ")
@@ -939,18 +940,18 @@ namespace OpenDDS {
             TraitsType::type_name()));
         }
         return;
-      }
-      Encoding encoding;
-      if (!to_encoding(encoding, encap, type_support_->base_extensibility())) {
+      } else if (read_status == EncapsulationReadStatus::ExtensibilityMismatch) {
         if (log_level >= LogLevel::Error) {
           ACE_ERROR((LM_ERROR,
                      "(%P|%t) ERROR: %CDataReaderImpl::lookup_instance: "
                      "to_encoding failed writer %C reader %C\n",
+                     TraitsType::type_name(),
                      LogGuid(sample.header_.publication_id_).c_str(),
                      LogGuid(subscription_id()).c_str()));
         }
         return;
       }
+      const Encoding& encoding = ser.encoding();
 
       if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
         if (DCPS_debug_level >= 1) {
@@ -970,8 +971,6 @@ namespace OpenDDS {
           TraitsType::type_name(),
           Encoding::kind_to_string(encoding.kind()).c_str()));
       }
-
-      ser.encoding(encoding);
     }
 
     bool ser_ret = true;
@@ -1106,8 +1105,9 @@ protected:
       static_cast<Endianness>(sample.header_.byte_order_));
 
     if (encapsulated) {
-      EncapsulationHeader encap;
-      if (!(ser >> encap)) {
+      const EncapsulationReadStatus::Value read_status =
+        read_encapsulation_header(ser, type_support_->base_extensibility());
+      if (read_status == EncapsulationReadStatus::HeaderError) {
         if (DCPS_debug_level > 0) {
           ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
             ACE_TEXT("%CDataReaderImpl::dds_demarshal: ")
@@ -1115,18 +1115,18 @@ protected:
             TraitsType::type_name()));
         }
         return;
-      }
-      Encoding encoding;
-      if (!to_encoding(encoding, encap, type_support_->base_extensibility())) {
+      } else if (read_status == EncapsulationReadStatus::ExtensibilityMismatch) {
         if (log_level >= LogLevel::Error) {
           ACE_ERROR((LM_ERROR,
                      "(%P|%t) ERROR: %CDataReaderImpl::dds_demarshal: "
                      "to_encoding failed writer %C reader %C\n",
+                     TraitsType::type_name(),
                      LogGuid(sample.header_.publication_id_).c_str(),
                      LogGuid(subscription_id()).c_str()));
         }
         return;
       }
+      const Encoding& encoding = ser.encoding();
 
       if (decoding_modes_.find(encoding.kind()) == decoding_modes_.end()) {
         if (DCPS_debug_level >= 1) {
@@ -1146,8 +1146,6 @@ protected:
           TraitsType::type_name(),
           Encoding::kind_to_string(encoding.kind()).c_str()));
       }
-
-      ser.encoding(encoding);
     }
 
     const bool key_only_marshaling =
--- a/dds/DCPS/EncapsulationHeader.cpp
+++ b/dds/DCPS/EncapsulationHeader.cpp
@@ -263,6 +263,30 @@ to_any_encoding(Encoding& encoding,
   return to_encoding_i(encoding, header, 0);
 }
 
+EncapsulationReadStatus::Value
+read_encapsulation_header(Serializer& s,
+                          EncapsulationHeader& header,
+                          Extensibility expected_extensibility)
+{
+  if (!(s >> header)) {
+    return EncapsulationReadStatus::HeaderError;
+  }
+  Encoding encoding;
+  if (!to_encoding(encoding, header, expected_extensibility)) {
+    return EncapsulationReadStatus::ExtensibilityMismatch;
+  }
+  s.encoding(encoding);
+  return EncapsulationReadStatus::Ok;
+}
+
+EncapsulationReadStatus::Value
+read_encapsulation_header(Serializer& s,
+                          Extensibility expected_extensibility)
+{
+  EncapsulationHeader header;
+  return read_encapsulation_header(s, header, expected_extensibility);
+}
+
 }
 }
 
--- a/dds/DCPS/FilterEvaluator.cpp
+++ b/dds/DCPS/FilterEvaluator.cpp
@@ -140,23 +140,20 @@ FilterEvaluator::SerializedForEval::lookup(const char* field) const
   Message_Block_Ptr mb(serialized_->duplicate());
   Serializer ser(mb.get(), encoding_);
   if (encoding_.is_encapsulated()) {
-    EncapsulationHeader encap;
-    if (!(ser >> encap)) {
+    const EncapsulationReadStatus::Value read_status = read_encapsulation_header(ser, exten_);
+    if (read_status == EncapsulationReadStatus::HeaderError) {
       ACE_ERROR((LM_ERROR, ACE_TEXT("(%P|%t) ERROR ")
         ACE_TEXT("FilterEvaluator::SerializedForEval::lookup: ")
         ACE_TEXT("deserialization of encapsulation header failed.\n")));
       throw std::runtime_error("FilterEvaluator::SerializedForEval::lookup:"
         "deserialization of encapsulation header failed.\n");
-    }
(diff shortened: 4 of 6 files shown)


dds/DCPS/EncapsulationHeader.h
```diff
@@ -128,6 +128,35 @@ OpenDDS_Dcps_Export
 bool to_any_encoding(Encoding& encoding,
                      const EncapsulationHeader& header);
 
+struct EncapsulationReadStatus {
+  enum Value {
+    Ok,
+    /// The 4-byte header itself could not be deserialized, e.g. a truncated buffer.
+    HeaderError,
+    /// The header was read, but its representation doesn't match expected_extensibility.
+    ExtensibilityMismatch
```
mitza-oci:
```suggestion
    ExtensibilityMismatch,
```
Let future changes extend the list without editing this line

mitza-oci (changes requested):

mitza-oci (approved):

PR #5282, opened 2026-08-20
fix: invalid bool load in `to_boolean`

### LINKED ISSUE #5281: Invalid `bool` load in OpenDDS RTPS `Parameter` deserializer (`to_boolean`) when a boolean byte is not 0 or 1
## Summary

When OpenDDS deserializes an RTPS `Parameter` whose discriminator is `PID_EXPECTS_INLINE_QOS` (`0x0043`, `dds/DCPS/RTPS/RtpsCore.idl:287`) and whose single boolean data byte is not `0` or `1` (e.g. `0xFF`), `OpenDDS::DCPS::operator>>(Serializer&, ACE_InputCDR::to_boolean)` (`dds/DCPS/Serializer.inl:1316`) copies the raw byte directly into a `bool` (`ACE_CDR::Boolean = bool`, `ace/CDR_Base.h:180`) without validating the value.

The invalid load happens inside the library, unconditionally in the deserializer, before `good_bit()` is consulted. The generated caller (`dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp:13828-13829`) trusts `to_boolean` to either yield a valid `bool` or return false; it can do neither for `0xFF`. No caller contract is violated. The root cause is that OpenDDS's `to_boolean` (`Serializer.inl:1316-1320`) omits the value validation/normalization that its own ACE parent, `ACE_InputCDR::read_boolean` (`ace/CDR_Stream.inl`), deliberately performs — ACE reads into an `Octet`, normalizes (`tmp ? true : false`), and rejects any value but `0`/`1`. The bug is network-reachable: `dds/DCPS/RTPS/Spdp.cpp:3254` deserializes a remote peer's SPDP `ParameterList` over `KIND_XCDR1` via the same `ser >> plist` → `Parameter` → `case 67u` → `to_boolean` path.

## Version

```bash
$ git describe --tags
v3.34.0-108-g1335f0d38b
```

Pinned commit: `1335f0d38b150422e3f46308938b639f5753ba9b`

## Description

The generated RTPS `Parameter` deserializer reads the discriminator and length, then dispatches on `disc`. For `PID_EXPECTS_INLINE_QOS` (`case 67u`) it reads one boolean:

```cpp
// RtpsCoreTypeSupportImpl.cpp:13827-13834 (generated)
case 67u: {
    ACE_CDR::Boolean tmp;
    if (strm >> ACE_InputCDR::to_boolean(tmp)) {   // expects a safe read
        uni.expects_inline_qos(tmp);               // <-- UBSan: load of value 255
        uni._d(disc);
        return true;
    }
    return false;
}
```

`to_boolean` copies the raw byte straight into the `bool` reference with no validation:

```cpp
// dds/DCPS/Serializer.inl:1316-1320
operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
{
    s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
    return s.good_bit();
}
```

`buffer_read` only fails if the stream is exhausted — it never inspects the byte value, so `0xFF` is loaded into a `bool` and `good_bit()` stays true. The subsequent `uni.expects_inline_qos(tmp)` then uses the ill-formed `bool`. By contrast, ACE's own `read_boolean` (which OpenDDS's overload supersedes) reads into an `Octet`, normalizes to `true`/`false`, and sets `good_bit_ = !(tmp & ~1)` to reject corrupt booleans — the exact contract the generated code assumes:

```cpp
// ace/CDR_Stream.inl — ACE_InputCDR::read_boolean
ACE_InputCDR::read_boolean(ACE_CDR::Boolean& x)
{
    ACE_CDR::Octet tmp {};
    if (this->read_octet(tmp)) {
        x = tmp ? true : false;
        // ... another value is very likely incorrect or corrupt data
        this->good_bit_ = !(tmp & static_cast<ACE_CDR::Octet>(~1));
        return this->good_bit_;
    }
    return false;
}
```

The trigger is a structurally valid RTPS `Parameter`: discriminator `0x0043` (`PID_EXPECTS_INLINE_QOS`), length `1`, one data byte `0xFF`. The only "malformed" aspect is the boolean value — precisely the kind of input a robust network deserializer must reject, and which ACE does. The same `to_boolean` overload is emitted by the codegen template for every boolean field (`dds/idl/marshal_generator.cpp:746`), so the defect is not specific to this one parameter. 

## PoC Code
```cpp
#include <dds/DCPS/RTPS/RtpsCoreC.h>
#include <dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.h>

using OpenDDS::DCPS::Serializer;
using OpenDDS::DCPS::Encoding;
using OpenDDS::RTPS::Parameter;

int main() {
  static const unsigned char cdr[] = {0x43, 0x00, 0x01, 0x00, 0xFF};

  ACE_Message_Block mb(reinterpret_cast<const char*>(cdr), sizeof cdr);
  mb.wr_ptr(sizeof cdr);

  Serializer ser(&mb, Encoding(Encoding::KIND_XCDR1));

  Parameter param;
  ser >> param;  // UBSan: load of value 255, not a valid bool (case PID_EXPECTS_INLINE_QOS)
  return 0;
}
```

## Stack Trace
```bash
runtime error: load of value 255, which is not a valid value for type 'ACE_CDR::Boolean' (aka 'bool')
    #0 0x5c39bf42bfa6 in OpenDDS::DCPS::operator>>(OpenDDS::DCPS::Serializer&, OpenDDS::RTPS::Parameter&) dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp:13829:30
    #1 0x5c39bda189c2 in main poc.cpp:19:7
    #2 0x7372841e31c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7372841e328a in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x5c39bd92da94 in _start (poc+0x18a2a94) (BuildId: 11c182c423013db473dcde2ef6dc798973658c04)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp:13829:30 
```

## Reproduction Step

```bash
# 1. Checkout OpenDDS at the version above
git clone https://github.com/OpenDDS/OpenDDS.git
cd OpenDDS
git checkout 1335f0d38b150422e3f46308938b639f5753ba9b

# 2. Install deps the security build needs (XercesC, OpenSSL)
apt-get update && apt-get install -y --no-install-recommends \
    clang clang++ libxerces-c-dev libssl-dev cmake make

# 3. Build the library with ASan+UBSan (static libs + headers)
export CC=clang CXX=clang++ LD=clang++
export CFLAGS="-fsanitize=address,undefined -fno-sanitize=function \
  -fsanitize-address-use-after-scope -g -O0 -fstandalone-debug -w -fPIC \
  -fno-inline -fno-omit-frame-pointer -fno-optimize-sibling-calls \
  -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types \
  -Wno-error=int-conversion -Wno-error=deprecated-declarations \
  -Wno-error=implicit-function-declaration -Wno-error=implicit-int \
  -Wno-error=vla-cxx-extension -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export CXXFLAGS="$CFLAGS" LDFLAGS="$CXXFLAGS"
export WORK="$PWD/build-sanitizer"
export SRC="$PWD"

cmake -DCMAKE_INSTALL_PREFIX="$WORK" \
  -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
  -DCMAKE_C_FLAGS="$CFLAGS" -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
  -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DBUILD_SHARED_LIBS=OFF \
  -DBUILD_TESTING=OFF -DOPENDDS_BUILD_TESTS=OFF -DOPENDDS_BUILD_EXAMPLES=OFF \
  -DOPENDDS_SECURITY=ON -DOPENDDS_XERCES3=ON -DOPENDDS_BUILT_IN_TOPICS=ON \
  -DOPENDDS_CONTENT_SUBSCRIPTION=ON -DOPENDDS_OBJECT_MODEL_PROFILE=ON \
  -DOPENDDS_PERSISTENCE_PROFILE=ON -DOPENDDS_OWNERSHIP_PROFILE=ON \
  -DOPENDDS_TAO_IIOP=ON "$SRC"
make -j$(nproc) && make install
export ASAN_OPTIONS="new_delete_type_mismatch=0:detect_leaks=0"

# 4. Build the standalone PoC against the sanitizer build.
ACE_TAO="$WORK/ace_tao"
clang++ -g -O0 -fno-omit-frame-pointer -ftrivial-auto-var-init=zero \
    -fsanitize=address,undefined \
    -I"$WORK/include" -I"$ACE_TAO" -I"$ACE_TAO/TAO" \
    poc.cpp -o poc \
    -Wl,--start-group "$WORK"/lib/lib*.a "$ACE_TAO"/lib/lib*.a -Wl,--end-group \
    -lxerces-c -lssl -lcrypto -lpthread -ldl -lrt -lz -lm

# 5. Run
UBSAN_OPTIONS=halt_on_error=1 ./poc
```

The crash is deterministic and reproduces on every run.
On a non-UBSan build, 0xFF is silently accepted as the boolean value; good_bit() remains true.

## Submission Statement

This report was produced by FuzzAnything's AI-assisted library fuzzer and manually verified by a team member. We reviewed the PoC against the upstream API documentation — call order, parameters, and memory ownership — and found no API misuse.

> Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

yvonnelxxxx:
This PR attempts to fix: [#5281](https://github.com/OpenDDS/OpenDDS/issues/5281#issue-5201714565)

## Problem

When OpenDDS deserializes an RTPS `Parameter` whose discriminator is `PID_EXPECTS_INLINE_QOS` (`0x0043`) and whose single boolean data byte is not `0` or `1` (e.g. `0xFF`), `operator>>(Serializer&, ACE_InputCDR::to_boolean)` (`dds/DCPS/Serializer.inl:1316`) copies the raw byte directly into a `bool` (`ACE_CDR::Boolean = bool`, `ace/CDR_Base.h:180`) with no validation:

```cpp
operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
{
  s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
  return s.good_bit();
}
```

`buffer_read` only fails on a short stream — it never inspects the byte value — so `0xFF` is loaded into a `bool`. The invalid load happens *inside the library*, unconditionally, before `good_bit()` is consulted. The generated caller (`RtpsCoreTypeSupportImpl.cpp:13828-13829`, and every boolean field emitted via `marshal_generator.cpp:746`) trusts `to_boolean` to either yield a valid `bool` or return `false`; it can do neither for `0xFF`, so the ill-formed `bool` flows into setters like `uni.expects_inline_qos(tmp)`.

This diverges from OpenDDS's own ACE parent. `ACE_InputCDR::read_boolean` (`ace/CDR_Stream.inl`) reads into an `Octet`, normalizes (`tmp ? true : false`), and rejects non-`0`/`1` values via `good_bit_ = !(tmp & ~1)`. OpenDDS's overload skips exactly that contract.

The bug is network-reachable: `dds/DCPS/RTPS/Spdp.cpp:3254` deserializes a remote peer's SPDP `ParameterList` over `KIND_XCDR1` via the same `ser >> plist` -> `Parameter` -> `case 67u` -> `to_boolean` path.

## Fix

Mirror ACE's `read_boolean`: read the byte into a stack `Octet` (safe — `Octet` is `unsigned char`, any bit pattern is valid), normalize it into the `bool` reference, and — if the value is not `0`/`1` — mark construction as failed and return `false` so the generated caller takes its existing failure branch.

```cpp
ACE_INLINE bool
operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
{
  ACE_CDR::Octet tmp = 0;
  s.buffer_read(reinterpret_cast<char*>(&tmp), boolean_cdr_size, s.swap_bytes());
  if (!s.good_bit()) {
    return false;
  }
  if (tmp & ~ACE_CDR::Octet(1)) {
    x.ref_ = tmp != 0;
    s.set_construction_status(Serializer::ElementConstructionFailure);
    return false;
  }
  x.ref_ = tmp != 0;
  return true;
}
```

Why this shape:

- **Read into an `Octet`, not the `bool`.** The UB is the load of `0xFF` into a `bool`. Routing through an `unsigned char` makes the load always-defined.
- **Normalize even on failure** (`x.ref_ = tmp != 0`). Callers that ignore the return value never see a trap representation. Matches ACE's `x = tmp ? true : false`.
- **Return `false` + set `ElementConstructionFailure`.** The generated code at `RtpsCoreTypeSupportImpl.cpp:13828-13834` already handles a `false` return by setting `ElementConstructionFailure` and returning `false` from the `Parameter` deserializer; setting it here too is belt-and-suspenders and mirrors how `to_string` reports `BoundConstructionFailure` for its own validation failures. No new API surface is needed — `set_construction_status` (`Serializer.h:662`) and the `ConstructionStatus` enum (`Serializer.h:654-657`) already exist.
- **`~ACE_CDR::Octet(1)` mask** is the exact check ACE uses, so behavior matches the upstream parent.

This overload is the single chokepoint for *every* boolean deserialization in OpenDDS (the IDL compiler emits `>> to_boolean` for every boolean field), so fixing it here fixes all boolean sites; no generated-code regeneration is required.

### Before

```
$ UBSAN_OPTIONS=halt_on_error=1 ./poc
RtpsCoreTypeSupportImpl.cpp:13829:30: runtime error: load of value 255, which is not a valid value for type 'ACE_CDR::Boolean' (aka 'bool')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior
RtpsCoreTypeSupportImpl.cpp:13829:30 in ... expects_inline_qos
```

### After

```
$ UBSAN_OPTIONS=halt_on_error=1 ./poc
$ echo $?
0
```

The boolean read returns `false`, the `Parameter` deserializer returns `false` from `case 67u`, `ParameterList` deserialization fails gracefully, and the UBSan diagnostic disappears (clean exit, no sanitizer errors).

In-range booleans (`0x00`, `0x01`) are unchanged: `buffer_read` succeeds, the mask check passes, and `x.ref_` is set to `false`/`true` exactly as before.

### Test results

Rebuilt the library with the OpenDDS security/sanitizer configuration (`OPENDDS_SECURITY=ON`, ASan+UBSan, clang) and reran the reproducer against the patched `Serializer.inl`. The UBSan crash no longer fires; valid booleans deserialize identically.

### DIFF
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -1315,8 +1315,18 @@ operator>>(Serializer& s, long double& x)
 ACE_INLINE bool
 operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
 {
+  ACE_CDR::Octet tmp = 0;
   s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
-  return s.good_bit();
+  if (!s.good_bit()) {
+    return false;
+  }
+  if (tmp & ~ACE_CDR::Octet(1)) {
+    x.ref_ = tmp != 0;
+    s.set_construction_status(Serializer::ElementConstructionFailure);
+    return false;
+  }
+  x.ref_ = tmp != 0;
+  return true;
 }
 
 ACE_INLINE bool
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -1316,7 +1316,7 @@ ACE_INLINE bool
 operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
 {
   ACE_CDR::Octet tmp = 0;
-  s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
+  s.buffer_read(reinterpret_cast<char*>(&tmp), boolean_cdr_size, s.swap_bytes());
   if (!s.good_bit()) {
     return false;
   }


dds/DCPS/Serializer.inl
```diff
@@ -1315,8 +1315,18 @@ operator>>(Serializer& s, long double& x)
 ACE_INLINE bool
 operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
 {
+  ACE_CDR::Octet tmp = 0;
   s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
-  return s.good_bit();
+  if (!s.good_bit()) {
+    return false;
+  }
+  if (tmp & ~ACE_CDR::Octet(1)) {
```
sonndinh, line 1323: `tmp` doesn't contain the value from the stream and is always 0 at this point. The value is still read directly from the stream into `x.ref_` using `buffer_read`, then line 1328 reassigns it using `tmp` which hasn't been set yet.

yvonnelxxxx:
@sonndinh Thanks for pointing that out. I've updated `buffer_read` to deserialize into `tmp` instead of `x.ref_` so the validation check operates on the actual stream data.

PR #5283, opened 2026-08-20
Honor ignore_member_names for unions

simpsont-oci:
### Motivation

Union assignability incorrectly rejected compatible unions with different member IDs even when ignore_member_names was enabled.

### Description of changes

Remove the redundant unconditional union member-name/ID check (the earlier check at lines 802-826 is correctly gated on ignore_member_names) and add regression coverage for assignability with ignore_member_names enabled and disabled.

### DIFF
--- a/dds/DCPS/XTypes/TypeAssignability.cpp
+++ b/dds/DCPS/XTypes/TypeAssignability.cpp
@@ -847,12 +847,6 @@ bool TypeAssignability::assignable_union(const MinimalTypeObject& ta,
       for (unsigned k = 0; k < label_seq_b.length(); ++k) {
         for (unsigned t = 0; t < label_seq_a.length(); ++t) {
           if (label_seq_b.members[k] == label_seq_a.members[t]) {
-            if (name_hash_equal(ta.union_type.member_seq[j].detail.name_hash,
-                                tb.union_type.member_seq[i].detail.name_hash) &&
-                ta.union_type.member_seq[j].common.member_id !=
-                tb.union_type.member_seq[i].common.member_id) {
-              return false;
-            }
             const TypeIdentifier& ti_a = ta.union_type.member_seq[j].common.type_id;
             const TypeIdentifier& ti_b = tb.union_type.member_seq[i].common.type_id;
             if (!assignable(ti_a, ti_b)) {
--- a/tests/unit-tests/dds/DCPS/XTypes/TypeAssignability.cpp
+++ b/tests/unit-tests/dds/DCPS/XTypes/TypeAssignability.cpp
@@ -3818,6 +3818,40 @@ TEST(dds_DCPS_XTypes_TypeAssignability, UnionTypeTest_Assignable)
   EXPECT_TRUE(test_imn.assignable(TypeObject(MinimalTypeObject(a_imn)),
                                   TypeObject(MinimalTypeObject(b_imn))));
 
+  // Members with matching names and labels but different IDs are assignable
+  // only when member names are ignored.
+  MinimalUnionType a_reordered, b_reordered;
+  a_reordered.union_flags = IS_APPENDABLE;
+  b_reordered.union_flags = IS_APPENDABLE;
+  a_reordered.discriminator.common.type_id = TypeIdentifier(TK_UINT32);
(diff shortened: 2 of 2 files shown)


dds/DCPS/XTypes/TypeAssignability.cpp
```diff
@@ -847,12 +847,6 @@ bool TypeAssignability::assignable_union(const MinimalTypeObject& ta,
       for (unsigned k = 0; k < label_seq_b.length(); ++k) {
         for (unsigned t = 0; t < label_seq_a.length(); ++t) {
           if (label_seq_b.members[k] == label_seq_a.members[t]) {
-            if (name_hash_equal(ta.union_type.member_seq[j].detail.name_hash,
-                                tb.union_type.member_seq[i].detail.name_hash) &&
-                ta.union_type.member_seq[j].common.member_id !=
-                tb.union_type.member_seq[i].common.member_id) {
-              return false;
-            }
```
sonndinh, line 855: Does this need to incorporate the value of `ignore_member_names` so that the name check is only ignored when it's `true`?

simpsont-oci: No, though I see why you're asking. The name checking logic is already implemented up above this line on line 803. This PR is fixing a bug where the names were being checked twice, once when ignore_member_names was false (above) and then also here, regardless of whether this flag was set.

tests/unit-tests/dds/DCPS/XTypes/TypeAssignability.cpp
```diff
@@ -3818,6 +3818,40 @@ TEST(dds_DCPS_XTypes_TypeAssignability, UnionTypeTest_Assignable)
   EXPECT_TRUE(test_imn.assignable(TypeObject(MinimalTypeObject(a_imn)),
                                   TypeObject(MinimalTypeObject(b_imn))));
 
+  // Members with matching names and labels but different IDs are assignable
+  // only when member names are ignored.
```
sonndinh, line 3822: I might be wrong but my understanding is that, for assignability check, the member ID is the main factor and the member name is the secondary factor which can be ignored in the check. So I think two members, each from a union, with a matching label still need to have the same ID, but can have different names depending on `ignore_member_names` value, in order for them to be assignable. The spec, however, seems not clear on this.

simpsont-oci: For unions, I think the emphasis is actually type assignability between matching labels, not matching IDs. The spec's union rule finds the "corresponding member" between T1 and T2 purely by shared case label. It doesn't require the two members it connects to share an ID. So if a label is common to both types, the members it points to just need to be type-assignable; they can have different names, different IDs, and even different (but still assignable) types.

Separately, the spec has its own global rule: any members in T1 and T2 that have the same name must have the same ID, and vice versa, unless ignore_member_names is set.

Because these are two separate rules, you can get a genuinely weird case: two members sharing a name and type across T1/T2, where the label-to-ID mapping is swapped between the two types (e.g., in T1 label 1 → id 1 → "m1", but in T2 label 1 → id 2 → "m1"). Type-wise this is perfectly assignable via the label rule. But it's only actually accepted overall when ignore_member_names is true. With the flag off, the name/ID consistency rule catches and rejects the mismatch (same name "m1", different ID). With the flag on, that check is skipped and the swap goes through.

This is exactly what the a_reordered/b_reordered test case already in this PR (~line 3821). EXPECT_FALSE with default settings, EXPECT_TRUE under test_imn (ignore_member_names=true). So I believe the current code (after this fix) is correctly modeling the spec: label matching drives type assignability, and the name/ID consistency rule is a separate check on top of it — not a requirement that label-matched members share an ID.

sonndinh: Recap on what has been discussed: the XTypes assignability rules for union are driven by matching labels -- member IDs don't seem to be required to match. Howerver, `ignore_member_names` description seems to say member IDs check should always be done, and whether member names should also be checked is controlled by `ignore_member_names` value. The spec could provide more clarity on that. For now, no further change to the implementation is needed.

sonndinh (approved):

Issue #5288, opened 2026-08-28
OpenDDS accepts an XCDR1 payload whose `sequence<long, 5>` length is zero while trailing serialized elements remain in the sample


wuyuqa:
```
This is OpenDDS version 3.35.0-dev (NOT A RELEASE)

    OpenDDS VERSION: 3.35.0-dev
    TAO VERSION: (built from OpenDDS 3.34.0/ACE+TAO bundle)

    HOST MACHINE and OPERATING SYSTEM:
    Ubuntu 24.04.4 LTS (Linux x86_64)

    TARGET MACHINE and OPERATING SYSTEM, if different from HOST:
    (same as HOST)

    COMPILER NAME AND VERSION (AND PATCHLEVEL):
    GCC (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0

    CONTENTS OF $ACE_ROOT/ace/config.h [if you use a link to a platform-
    specific file, simply state which one]:
    #include "ace/config-linux.h"

    CONTENTS OF $ACE_ROOT/include/makeinclude/platform_macros.GNU (unless
    this isn't used in this case, e.g., with Microsoft Visual C++):
    include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU
    (or as generated by OpenDDS configure script)

    CONTENTS OF $ACE_ROOT/bin/MakeProjectCreator/config/default.features
    (used by MPC when you generate your own makefiles):
    (default OpenDDS 3.35.0-dev configuration)

    LEVEL OF URGENCY (LOW, MEDIUM, or HIGH):
    MEDIUM

    AREA/CLASS/EXAMPLE AFFECTED:
[What example failed?  What module failed to compile?]
    DDS XTypes / XCDR1 deserialization; RTPS receive path; no compilation failure.

    DOES THE PROBLEM AFFECT:
        COMPILATION?
            NO.
        LINKING?
            NO.
        EXECUTION?
            YES — OpenDDS subscriber accepts and delivers malformed XCDR1 payloads
            where sequence length declares 0 but trailing serialized elements remain.
        OTHER (please specify)?
            DDS interoperability / specification compliance.
[Please indicate whether DDS, your application, or both are affected.]
    Both DDS and application are affected: the DDS layer accepts invalid wire data,
    and the application receives a sample whose serialized form does not match the
    deserialized object state.

    SYNOPSIS:
[Brief description of the problem]
[...]

Issue #3133, opened 2021-11-03
Redundant variable generated in deserialize operation, deserialize result not checked


jwillemsen:
See generated `dds/DCPS/RTPS/RtpsCoreTypeSupportImpl.cpp`, it contains the following operation, see that `tempvar` is generated on the stack, the length is set, but not used at all. Also the return value of `strm >> tempvar_9tmp_9var_forany` is not checked

```
bool operator>>(Serializer& strm,  ::OpenDDS::RTPS::FilterSignatureSeq& seq)
{
  ACE_UNUSED_ARG(strm);
  ACE_UNUSED_ARG(seq);
  const Encoding& encoding = strm.encoding();
  ACE_UNUSED_ARG(encoding);
  size_t total_size = 0;
  if (encoding.xcdr_version() == Encoding::XCDR_VERSION_2) {
    if (!strm.read_delimiter(total_size)) {
      return false;
    }
  }
  const size_t end_of_seq = strm.rpos() + total_size;
  CORBA::ULong length;
  if (!(strm >> length)) {
    return false;
  }
  if (length > strm.length()) {
    if (DCPS_debug_level >= 8) {
      ACE_DEBUG((LM_DEBUG, ACE_TEXT("(%P|%t) Invalid sequence length (%u)\n"), length));
    }
    return false;
  }
  CORBA::ULong new_length = length;
  seq.length(new_length);
  for (CORBA::ULong i = 0; i < new_length; ++i) {
     ::OpenDDS::RTPS::FilterSignature_t_var seq_i_tmp_var=  ::OpenDDS::RTPS::FilterSignature_t_alloc();
     ::OpenDDS::RTPS::FilterSignature_t_forany seq_9i_9tmp_9var_forany = seq_i_tmp_var.inout();
     if (!(strm >> seq_9i_9tmp_9var_forany)) {
      strm.set_construction_status(Serializer::ElementConstructionFailure);
      if (encoding.xcdr_version() == Encoding::XCDR_VERSION_2) {
        strm.skip(end_of_seq - strm.rpos());
      } else {
         ::OpenDDS::RTPS::FilterSignatureSeq tempvar;
        tempvar.length(1);
        for (CORBA::ULong j = i + 1; j < length; ++j) {
           ::OpenDDS::RTPS::FilterSignature_t_var tempvar_tmp_var=  ::OpenDDS::RTPS::FilterSignature_t_alloc();
[...]

PR #5290, opened 2026-09-08
XCDR Deserialization Hardening

### LINKED ISSUE #5281: Invalid `bool` load in OpenDDS RTPS `Parameter` deserializer (`to_boolean`) when a boolean byte is not 0 or 1

### LINKED ISSUE #5287: OpenDDS accepts an XCDRv2 `DELIMIT_CDR2` payload whose top-level `DHEADER` is smaller than the actual serialized object size
```plaintext
PROBLEM-REPORT-FORM

This is OpenDDS version 3.35.0-dev (NOT A RELEASE)

    OpenDDS VERSION: 3.35.0-dev (commit 1a9f3ccca65c8933f2698a7cd45a1cef54c64780)
    TAO VERSION: (as shipped with OpenDDS 3.35.0-dev)

    x86_64, Ubuntu 24.04.4 LTS (Linux kernel 6.8.0)

    Same as host (tested on loopback)

    g++ (Ubuntu 13.2.0-23ubuntu4) 13.2.0  (or other, please fill actual)

    CONTENTS OF $ACE_ROOT/ace/config.h:
    #include "ace/config-linux.h"
    (or equivalent, as generated by OpenDDS build)

    CONTENTS OF $ACE_ROOT/include/makeinclude/platform_macros.GNU:
    (standard for Linux, e.g.,)
    include $(ACE_ROOT)/include/makeinclude/platform_linux.GNU

    CONTENTS OF $ACE_ROOT/bin/MakeProjectCreator/config/default.features:
    (default MPC features; not modified)

    MEDIUM

    AREA/CLASS/EXAMPLE AFFECTED:
    DDS XTypes serialization/deserialization (XCDRv2)
    Affects the receive path for @appendable structs with DELIMIT_CDR2 encoding.
    Tested with the user-provided IDL and subscriber/publisher.

    DOES THE PROBLEM AFFECT:
        COMPILATION? No
        LINKING? No
        EXECUTION? Yes – the OpenDDS subscriber accepts and delivers a sample whose top-level DHEADER (delimiter length) is smaller than the actual serialized payload.
        OTHER? The issue is related to XCDRv2 validation; it does not cause a crash, but allows malformed packets to be processed as valid.

    SYNOPSIS:
    OpenDDS accepts an XCDRv2 `DELIMIT_CDR2` payload whose top-level `DHEADER` is decremented by 1, while the object bytes remain unchanged. The sample is delivered to the application even though the declared size under-reports the actual serialized object size, contrary to the XTypes specification.

    DESCRIPTION:
    According to OMG DDS-XTypes 1.3, for an `@appendable` struct encoded as `DELIMITED_CDR`, the top-level `DHEADER` MUST represent the exact serialized size of the object that follows. Appendable evolution permits extra trailing data (to be ignored) only when the delimiter correctly reports the base object size. It does not allow the delimiter to be smaller than the actual object size.

    In this reproducer, we take a real RTPS packet carrying a serialized sample of the `@appendable` struct `S4_B2_TypeObject_0000`. We locate the top-level DHEADER and change its value from `27` to `26` (i.e., minus one) without modifying any of the object bytes. The mutated packet is replayed to a real OpenDDS subscriber. The subscriber accepts the sample and prints the received data (id and string) to the console, indicating that the malformed payload was successfully deserialized.

    This behavior is observed with a real OpenDDS subscriber and publisher, using real RTPS/UDP communication, not merely a unit test. The mutation is applied only to the DHEADER field, and the rest of the packet remains untouched.

    Cross-implementation comparison:
    - OpenDDS (3.35.0-dev): accepts
    - Fast DDS: also accepts (observed)
    - CycloneDDS: rejects with a deserialization failure

    This indicates a divergence from the specification and from at least one other OMG DDS implementation.

    REPEAT BY:
    The following steps reproduce the issue on a clean Ubuntu 24.04 system with OpenDDS 3.35.0-dev built from source.

    Preparation:
    1. Check out OpenDDS at commit 1a9f3ccca65c8933f2698a7cd45a1cef54c64780 and build it (using ACE/TAO as provided).
    2. Create the IDL file `test.idl` with:
       ```
       module test {
         @topic
         @appendable
         struct S4_B2_TypeObject_0000 {
           long id;
           string<24> type_name_probe;
         };
       };
       ```
    3. Build the subscriber and publisher applications using the commands provided in the original report (see below). Ensure the subscriber and publisher are built with the same OpenDDS.

    Reproduction commands (from the user's environment):
    ```
    cd /home/wuyuqa/DDS_empirical/pocs/opendds_xcdr2_dheader_length_validation
    python3 -m unittest test_build_remote_poc.py test_mutate_and_replay.py
    bash build_remote_poc.sh
    REPLAY_MODE=passthrough bash run_packet_replay_poc.sh
    bash run_packet_replay_poc.sh
    ```

    Key steps in the script:
    - Start the subscriber (using `timeout 30s`).
    - Start the publisher with a UDP capture preload (LD_PRELOAD) to record packets.
    - The capture tool saves packets to a directory.
    - The mutation script `mutate_and_replay.py` locates the packet containing the marker string "packet-replay-dh01" and modifies only the top-level DHEADER from its original value (27) to 26.
    - The modified packet is replayed to the subscriber.
    - The subscriber output shows a line `REMOTE_SAMPLE_RECEIVED id=4015 type_name_probe=packet-replay-dh01`, proving acceptance.

    Control run (without mutation) shows that the original packet is accepted. The mutated run produces the same acceptance.

    Relevant log output:
    Control:
    ```
    CONTROL_PACKET_REPLAY_ACCEPTED
    publisher_status=143
    replay_status=0
    subscriber_status=0
    captured_packets=12
    ```
    Mutated:
    ```
    PACKET_REPLAY_ACCEPTED_MALFORMED_DHEADER
    publisher_status=143
    replay_status=0
    subscriber_status=0
    captured_packets=12
    ```
    Subscriber output (mutated):
    ```
    REMOTE_SAMPLE_RECEIVED
    id=4015 type_name_probe=packet-replay-dh01
    ```

    The full code (subscriber.cpp, benign_publisher.cpp, udp_capture_preload.cpp, mutate_and_replay.py, build scripts) is available in the user's local repository and can be provided upon request.

    SAMPLE FIX/WORKAROUND:
    No workaround is currently known for end users, but a potential fix would be to add validation in the XCDRv2 deserialization path that checks whether the actual number of bytes consumed from the payload exceeds the value declared in the top-level DHEADER. If it does, the deserialization should fail and the sample should be rejected (e.g., by throwing an exception or marking the sample as invalid).

    In the OpenDDS code, this would likely affect the `DCPS::XCDR2::deserialize` or related methods for appendable types, where the delimiter length is used to bound the parsing. The parser should ensure that after reading all required fields, the total bytes consumed do not exceed the DHEADER value; if they exceed, it should indicate an error. Similarly, if the DHEADER indicates a larger size than the actual object, the extra bytes could be ignored (as per appendable semantics), but the under-reporting case is invalid and must be rejected.

```

### LINKED ISSUE #5289: OpenDDS accepts an XCDR1 payload whose `sequence<long, 5>` length exceeds the remaining serialized elements and still delivers the sample to application code
## OpenDDS Problem Report

**This is OpenDDS version 3.34.0 (NOT A RELEASE)**

- **OpenDDS VERSION:** 3.34.0  
- **TAO VERSION:** Not provided (built with OpenDDS 3.34.0)  
- **HOST MACHINE and OPERATING SYSTEM:** Ubuntu 24.04.4 LTS  
- **TARGET MACHINE and OPERATING SYSTEM, if different from HOST:** Same as host  
- **COMPILER NAME AND VERSION (AND PATCHLEVEL):** Not provided (system compiler on Ubuntu 24.04.4 LTS)  
- **CONTENTS OF $ACE_ROOT/ace/config.h:** Not provided  
- **CONTENTS OF $ACE_ROOT/include/makeinclude/platform_macros.GNU:** Not provided  
- **CONTENTS OF $ACE_ROOT/bin/MakeProjectCreator/config/default.features:** Not provided  
- **LEVEL OF URGENCY (LOW, MEDIUM, or HIGH):** HIGH  
- **AREA/CLASS/EXAMPLE AFFECTED:**  
  Serialization/deserialization of bounded sequences in XCDR1 encoding. The generated sequence reader in the OpenDDS serialization layer accepts malformed payloads where the sequence length field claims more elements than the payload contains.  

- **DOES THE PROBLEM AFFECT:**  
  - **COMPILATION?** No.  
  - **LINKING?** No.  
  - **EXECUTION?** Yes. A real OpenDDS subscriber accepts the malformed sample and delivers it to application code.  
  - **OTHER?** N/A  

- **Please indicate whether DDS, your application, or both are affected.**  
  Both the DDS infrastructure and the application are affected. The malformed sample is accepted by the DDS layer and delivered to the application callback.  

- **SYNOPSIS:**  
  OpenDDS accepts an XCDR1 payload where a `sequence<long, 5>` field declares length 5, even though the payload contains bytes for only 3 serialized `long` elements after the length field. The mutated sample is still delivered to application code, and the subscriber prints `values_size=5` even though only 3 values were present on the wire.  

- **DESCRIPTION:**  
  A real OpenDDS publisher and subscriber are built from the following IDL:  

  ```idl
  module test {
    struct S5_B2_StringBound_0000 {
      string<8> name;
      sequence<long, 5> values;
    };
    #pragma DCPS_DATA_TYPE "test::S5_B2_StringBound_0000"
  };
  ```

  The publisher sends a sample with `name="ockgknze"` and `values = [63, 64, 65]`. The subscriber receives it correctly.  

  A UDP packet capture preload library captures outgoing RTPS datagrams from the publisher. The original benign user‑data datagram containing the target payload is dropped. The captured datagram is then replayed to the real subscriber in two modes:  

  - **Control (passthrough):** original datagram replayed unchanged → subscriber prints `values_size=3`  
  - **Mutated:** same datagram with only the sequence‑length field changed from `3` to `5` → subscriber prints `values_size=5`  

  **Original payload body (hex):**  
  ```
  090000006f636b676b6e7a6500000000030000003f0000004000000041000000
  ```

  **Mutated payload body (hex):**  
  ```
  090000006f636b676b6e7a6500000000050000003f0000004000000041000000
  ```

  **Field interpretation:**  
  - `09000000` → string length 9  
  - `6f636b676b6e7a65 00` → `"ockgknze\0"`  
  - `000000` → alignment padding  
  - `03000000` → original `values.length = 3`  
  - `05000000` → mutated `values.length = 5`  
  - `3f000000 40000000 41000000` → only three serialized `long` values: 63, 64, 65  

  The mutated payload claims a sequence of 5 elements, but only 12 bytes remain after the length field. Decoding 5 `long` elements would require 20 bytes. Despite this, the subscriber accepts the sample and delivers it to application code, which prints:  
  ```
  REMOTE_SAMPLE_RECEIVED name=ockgknze values_size=5 values_prefix=63,64,65
  ```

  **Suspected root cause:**  

  The generated OpenDDS sequence reader performs a pre‑check:  
  ```cpp
  if (length > strm.length()) {
    return false;
  }
  ```

  For this case:  
  - declared `length = 5`  
  - remaining bytes after the length field = 12  
  - `5 > 12` is false, even though decoding 5 `long` elements would require 20 bytes  

  The function then performs the array read without checking the return value:  
  ```cpp
  resize_bounded_seq_no_init(seq, new_length);
  strm.read_long_array(seq.get_buffer(), new_length);
  return true;
  ```

  The `read_long_array()` call fails (not enough bytes), but its return value is ignored, so the sequence extraction returns success.  

  **Cross‑implementation note:**  
  The same mutated payload was tested against CycloneDDS and Fast DDS. Both rejected the malformed sample during deserialization and did not deliver it to application code. This issue appears to be OpenDDS‑specific.  

- **REPEAT BY:**  
  The complete self‑contained PoC is available at:  
  `/home/wuyuqa/DDS_empirical/pocs/opendds_xcdr1_sequence_length_exceeds_payload`  

  To reproduce:  
  ```bash
  cd /home/wuyuqa/DDS_empirical/pocs/opendds_xcdr1_sequence_length_exceeds_payload
  python3 -m unittest test_mutate_and_replay.py test_build_remote_poc.py test_run_packet_replay_poc.py
  bash build_remote_poc.sh
  REPLAY_MODE=passthrough bash run_packet_replay_poc.sh   # control: prints values_size=3
  REPLAY_MODE=mutated bash run_packet_replay_poc.sh       # mutated: prints values_size=5
  ```

  **Observed control output:**  
  ```
  CONTROL_PACKET_REPLAY_ACCEPTED
  publisher_status=143
  replay_status=0
  subscriber_status=0
  captured_packets=11
  REMOTE_SAMPLE_RECEIVED name=ockgknze values_size=3 values_prefix=63,64,65
  ```

  **Observed mutated output:**  
  ```
  PACKET_REPLAY_ACCEPTED_SEQUENCE_LENGTH_EXCEEDS_PAYLOAD
  publisher_status=143
  replay_status=0
  subscriber_status=0
  captured_packets=11
  REMOTE_SAMPLE_RECEIVED name=ockgknze values_size=5 values_prefix=63,64,65
  ```

- **SAMPLE FIX/WORKAROUND:**  
  A possible fix would be to correct the pre‑check in the generated sequence reader to compare the required byte size against remaining bytes, rather than comparing element count against remaining bytes:  

  ```cpp
  // Current (incorrect):
  if (length > strm.length()) {
    return false;
  }

  // Suggested fix:
  if (length * sizeof(element_type) > strm.length()) {
    return false;
  }
  ```

  Additionally, the return value of `read_long_array()` (and similar read functions) should be checked before returning success from the sequence extraction operator.

simpsont-oci:
### Motivation

Harden XCDR1 and XCDR2 deserialization against malformed or inconsistent wire data that could cause excessive allocation, boundary over-reads, invalid Boolean representations, or sanitizer failures.

### Description of changes

- Normalize deserialized Boolean values so a wire byte other than 0 or 1 cannot produce an invalid `bool`.
- Validate sequence and map element counts against the remaining serialized size — accounting for element size and alignment — before allocating any container, and check the result of `read_<primitive>_array()` in the generated bounded-sequence reader.
- Enforce XCDR2 `DHEADER` sizes and XCDR1/XCDR2 mutable-member sizes as hard read boundaries through a `Serializer::ScopedReadLimit` helper; a delimited object or member that under-reports its size is now rejected instead of over-reading into later data. Bound the `read_string` pre-allocation check by the active limit.
- Reject malformed XCDR1 extended parameter headers, XCDR2 EMHEADER length codes, mutable-union discriminator IDs, and non-empty parameter-list-end sentinels.
- Preserve forward compatibility for unknown bitmask bits and, when member names are ignored, for compatible unions whose members use different IDs (see #5283).

Cleanup carried alongside the hardening:

- Add a one-argument `ScopedReadLimit(Serializer&)` that reads the XCDR2 delimiter and installs the limit; collapse ~65 near-identical `operator>>` prologues in `TypeObject.cpp` and drop the manual skip-to-end blocks it makes redundant.
- Add `ScopedReadLimit::reset()` and use it to scope the `DynamicData` mutable-member read limits with RAII.
- Regression tests for the malformed-input cases; correct two `DynamicDataXcdrReadImpl` test fixtures whose `DHEADER` values did not match their payloads (previously accepted only because the delimiter was unvalidated).

### Issues

Fixes #5281
Fixes #5287
Fixes #5289

Related, partially addressed: #5232 and #3133 (unchecked CDR extraction results — this PR fixes the generated bounded-primitive-sequence path, not the general sweep). #5288 (XCDR1 trailing bytes after a short sequence length) is intentionally out of scope: XCDR1 has no object framing, so it is a candidate for a future opt-in strict-decoding mode.

### DIFF
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -266,6 +266,10 @@ Serializer::doread(char* dest, size_t size, bool swap, size_t offset)
 ACE_INLINE void
 Serializer::buffer_read(char* dest, size_t size, bool swap)
 {
+  if (size > length()) {
+    good_bit_ = false;
+    return;
+  }
   size_t offset = 0;
 
   while (size > offset) {
@@ -398,17 +402,65 @@ Serializer::good_bit() const
 ACE_INLINE size_t
 Serializer::length() const
 {
-  return good_bit_ && current_ ? current_->total_length() : 0;
+  if (!good_bit_ || !current_ || rpos_ > read_limit_) {
+    return 0;
+  }
+  return (std::min)(current_->total_length(), read_limit_ - rpos_);
+}
+
+ACE_INLINE bool
+Serializer::check_size(size_t count, size_t element_size, size_t element_alignment)
+{
+  const size_t max = (std::numeric_limits<size_t>::max)();
+  if (count && element_size > max / count) {
+    good_bit_ = false;
+    return false;
+  }
+
+  size_t padding = 0;
+  if (alignment() && element_alignment > 1 && current_) {
+    element_alignment = (std::min)(element_alignment, encoding().max_align());
+    padding = (element_alignment - reinterpret_cast<size_t>(current_->rd_ptr()) +
+      align_rshift_) % element_alignment;
+  }
+  const size_t bytes = count * element_size;
+  if (padding > length() || bytes > length() - padding) {
+    good_bit_ = false;
+    return false;
+  }
+  return true;
+}
+
+ACE_INLINE bool
+Serializer::set_read_limit(size_t size)
+{
+  const size_t max = (std::numeric_limits<size_t>::max)();
+  if (!good_bit_ || size > length() || size > max - rpos_) {
+    good_bit_ = false;
+    return false;
+  }
+  read_limit_ = rpos_ + size;
+  return true;
 }
 
 ACE_INLINE bool
 Serializer::skip(size_t n, int size)
 {
+  if (size < 0 || (n && static_cast<size_t>(size) >
+      (std::numeric_limits<size_t>::max)() / n)) {
+    good_bit_ = false;
+    return false;
+  }
   if (size > 1 && !align_r((std::min)(size_t(size), encoding().max_align()))) {
     return false;
   }
 
-  for (size_t len = n * static_cast<size_t>(size); len;) {
+  const size_t total = n * static_cast<size_t>(size);
+  if (total > length()) {
+    good_bit_ = false;
+    return false;
+  }
+  for (size_t len = total; len;) {
     if (!current_) {
       good_bit_ = false;
       return false;
@@ -425,7 +477,7 @@ Serializer::skip(size_t n, int size)
   }
 
   if (good_bit_) {
-    rpos_ += n * static_cast<size_t>(size);
+    rpos_ += total;
   }
   return good_bit();
 }
@@ -440,6 +492,10 @@ ACE_INLINE void
 Serializer::read_array(char* x, size_t size,
                        ACE_CDR::ULong length, bool swap)
 {
+  if (length && size > (std::numeric_limits<size_t>::max)() / length) {
+    good_bit_ = false;
+    return;
+  }
   if (!swap || size == 1) {
     //
     // No swap, copy direct.  This silently corrupts the data if there is
@@ -494,8 +550,15 @@ Serializer::write_array(const char* x, size_t size,
 ACE_INLINE bool
 Serializer::read_boolean_array(ACE_CDR::Boolean* x, ACE_CDR::ULong length)
 {
-  read_array(reinterpret_cast<char*>(x), boolean_cdr_size, length);
-  return good_bit();
+  for (ACE_CDR::ULong i = 0; i < length; ++i) {
+    ACE_CDR::Octet value = 0;
+    buffer_read(reinterpret_cast<char*>(&value), boolean_cdr_size, false);
+    if (!good_bit()) {
+      return false;
+    }
+    x[i] = value != 0;
+  }
+  return true;
 }
 
 ACE_INLINE bool
@@ -876,6 +939,10 @@ bool Serializer::read_delimiter(size_t& size)
   if (encoding().xcdr_version() == Encoding::XCDR_VERSION_2) {
     ACE_CDR::ULong dheader;
     if (*this >> dheader) {
+      if (dheader > length()) {
+        good_bit_ = false;
+        return false;
+      }
       size = dheader;
       return true;
     }
@@ -1327,8 +1394,13 @@ operator>>(Serializer& s, long double& x)
 ACE_INLINE bool
 operator>>(Serializer& s, ACE_InputCDR::to_boolean x)
 {
-  s.buffer_read(reinterpret_cast<char*>(&x.ref_), boolean_cdr_size, s.swap_bytes());
-  return s.good_bit();
+  ACE_CDR::Octet value = 0;
+  s.buffer_read(reinterpret_cast<char*>(&value), boolean_cdr_size, false);
+  if (s.good_bit()) {
+    x.ref_ = value != 0;
+    return true;
+  }
+  return false;
 }
 
 ACE_INLINE bool
@@ -1734,7 +1806,7 @@ void Serializer::set_construction_status(ConstructionStatus cs)
 ACE_INLINE
 Serializer::RdState Serializer::rdstate() const
 {
-  RdState state(align_rshift_, rpos_);
+  RdState state(align_rshift_, rpos_, read_limit_);
   return state;
 }
 
@@ -1743,6 +1815,7 @@ void Serializer::rdstate(const RdState& state)
 {
   align_rshift_ = state.align_rshift;
   rpos_ = state.rpos;
+  read_limit_ = state.read_limit;
 }
 
 } // namespace DCPS
--- a/dds/DCPS/Serializer.inl
+++ b/dds/DCPS/Serializer.inl
@@ -417,6 +417,8 @@ Serializer::check_size(size_t count, size_t element_size, size_t element_alignme
     return false;
   }
 
+  // Bytes of alignment padding align_r() would insert before the first element;
+  // this must use the same formula as align_r() to stay in sync with it.
   size_t padding = 0;
   if (alignment() && element_alignment > 1 && current_) {
     element_alignment = (std::min)(element_alignment, encoding().max_align());
--- a/dds/idl/marshal_generator.cpp
+++ b/dds/idl/marshal_generator.cpp
@@ -262,6 +262,30 @@ namespace {
     }
   }
 
+  string getPrimitiveCdrSize(AST_Type* type)
+  {
+    switch (dynamic_cast<AST_PredefinedType*>(type)->pt()) {
+    case AST_PredefinedType::PT_long: return "int32_cdr_size";
+    case AST_PredefinedType::PT_ulong: return "uint32_cdr_size";
+    case AST_PredefinedType::PT_short: return "int16_cdr_size";
+    case AST_PredefinedType::PT_ushort: return "uint16_cdr_size";
+#if OPENDDS_HAS_EXPLICIT_INTS
+    case AST_PredefinedType::PT_int8: return "int8_cdr_size";
+    case AST_PredefinedType::PT_uint8: return "uint8_cdr_size";
+#endif
+    case AST_PredefinedType::PT_longlong: return "int64_cdr_size";
+    case AST_PredefinedType::PT_ulonglong: return "uint64_cdr_size";
+    case AST_PredefinedType::PT_float: return "float32_cdr_size";
+    case AST_PredefinedType::PT_double: return "float64_cdr_size";
+    case AST_PredefinedType::PT_longdouble: return "float128_cdr_size";
+    case AST_PredefinedType::PT_char: return "char8_cdr_size";
+    case AST_PredefinedType::PT_wchar: return "char16_cdr_size";
+    case AST_PredefinedType::PT_boolean: return "boolean_cdr_size";
+    case AST_PredefinedType::PT_octet: return "byte_cdr_size";
+    default: return "1u";
+    }
+  }
+
   string nameOfSeqHeader(AST_Type* elem)
   {
     string ser = getSerializerName(elem);
@@ -403,7 +427,7 @@ namespace {
         "  if (!(strm >> length)) {\n"
         "    return false;\n"
         "  }\n"
-        "  if (length > strm.length()) {\n"
+        "  if (!strm.check_size(length, 1u)) {\n"
         "    return false;\n"
         "  }\n"
         "  seq.length(length);\n"
@@ -698,7 +722,7 @@ namespace {
         marshal_generator::generate_dheader_code(
           "    if (!strm.read_delimiter(total_size)) {\n"
           "      return false;\n"
-          "    }\n", !primitive);
+          "    }\n", !primitive, true, "  ", true);
 
         if (!primitive) {
           be_global->impl_ << "  const size_t end_of_seq = strm.rpos() + total_size;\n";
@@ -706,9 +730,21 @@ namespace {
         be_global->impl_ <<
           "  CORBA::ULong length;\n"
           << streamAndCheck(">> length");
-        // The check here is to prevent very large sequences from being allocated.
+        // Validate a conservative lower bound before allocating the sequence.
+        // Primitive sequences use their exact fixed wire size.
+        std::string min_element_size = "1u";
+        std::string min_element_alignment = "1u";
+        if (elem_cls & CL_PRIMITIVE) {
+          min_element_size = min_element_alignment = getPrimitiveCdrSize(elem);
+        } else if (elem_cls & CL_ENUM) {
+          min_element_size = min_element_alignment = "int32_cdr_size";
+        } else if ((elem_cls & CL_STRING) || elem->node_type() == AST_Decl::NT_sequence ||
+                   elem->node_type() == AST_Decl::NT_map) {
+          min_element_size = min_element_alignment = "uint32_cdr_size";
+        }
         be_global->impl_ <<
-          "  if (length > strm.length()) {\n"
+          "  if (!strm.check_size(length, " << min_element_size << ", "
+          << min_element_alignment << ")) {\n"
           "    if (DCPS_debug_level >= 8) {\n"
           "      ACE_DEBUG((LM_DEBUG, ACE_TEXT(\"(%P|%t) Invalid sequence length (%u)\\n\"), length));\n"
           "    }\n"
@@ -1087,10 +1123,22 @@ namespace {
         "  DDS::UInt32 size = 0;\n" << // in bytes for DHeader, or in elements for Length
         streamAndCheck(">> size");
 
+      if (both_primitive) {
+        be_global->impl_ <<
+          "  if (!strm.check_size(size, " << getPrimitiveCdrSize(key) << " + "
+          << getPrimitiveCdrSize(val) << ")) {\n"
+          "    return false;\n"
+          "  }\n";
+      }
+
       if (!both_primitive) {
         be_global->impl_ <<
           "  const bool size_is_bytes = strm.encoding().xcdr_version() == Encoding::XCDR_VERSION_2;\n"
-          "  const std::size_t start = size_is_bytes ? strm.rpos() : 0u;\n";
+          "  const std::size_t start = size_is_bytes ? strm.rpos() : 0u;\n"
+          "  Serializer::ScopedReadLimit read_limit(strm, size, size_is_bytes, true);\n"
+          "  if (!read_limit.valid()) {\n"
+          "    return false;\n"
+          "  }\n";
       }
 
       const bool bounded = !map->unbounded(),
@@ -1342,7 +1390,7 @@ namespace {
       marshal_generator::generate_dheader_code(
         "    if (!strm.read_delimiter(total_size)) {\n"
         "      return false;\n"
-        "    }\n", !primitive);
+        "    }\n", !primitive, true, "  ", true);
 
       if (!primitive && (try_construct != tryconstructfailaction_use_default)) {
         be_global->impl_ << "  const size_t end_of_arr = strm.rpos() + total_size;\n";
@@ -2846,7 +2894,7 @@ namespace {
       marshal_generator::generate_dheader_code(
                                                "    if (!strm.read_delimiter(total_size)) {\n"
                                                "      return false;\n"
-                                               "    }\n", not_final);
+                                               "    }\n", not_final, true, "  ", true);
 
       if (not_final) {
         be_global->impl_ <<
@@ -2880,6 +2928,10 @@ namespace {
           "      }\n"
           "      if (encoding.xcdr_version() == Encoding::XCDR_VERSION_1 && member_id == Serializer::pid_list_end) {\n"
           "        return true;\n"
+          "      }\n"
+          "      Serializer::ScopedReadLimit member_limit(strm, field_size, true, true);\n"
+          "      if (!member_limit.valid()) {\n"
+          "        return false;\n"
           "      }\n"
           "      const size_t end_of_field = strm.rpos() + field_size;\n"
           "      ACE_UNUSED_ARG(end_of_field);\n"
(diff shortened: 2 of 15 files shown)


dds/DCPS/Serializer.inl
```diff
@@ -398,17 +402,67 @@ Serializer::good_bit() const
...
+Serializer::check_size(size_t count, size_t element_size, size_t element_alignment)
+{
+  const size_t max = (std::numeric_limits<size_t>::max)();
+  if (count && element_size > max / count) {
+    good_bit_ = false;
+    return false;
+  }
+
+  // Bytes of alignment padding align_r() would insert before the first element;
+  // this must use the same formula as align_r() to stay in sync with it.
```
sonndinh, line 421: Should it return early and skip the remaining operations when `count` is 0?

It looks to me that when `count` is 0, the `padding` can be non-zero depending on the current alignment of the stream. So if `padding` is greater than the remaining length, it's marked as an error. Could it be a false error indication? For example, if `check_size` is called while it's reading an empty sequence of long long from an XCDR1 stream. So the `count` passed to `check_size` is 0. `padding` can be as large as 4 since XCDR1 max alignment is 8 and long long size is also 8. If the remaining length is less than 4, for example, there's only a boolean value left in the stream, then looks like `check_size` would return false while the input data is valid?

dds/idl/marshal_generator.cpp
```diff
@@ -698,17 +728,29 @@ namespace {
...
           be_global->impl_ << "  const size_t end_of_seq = strm.rpos() + total_size;\n";
         }
         be_global->impl_ <<
           "  CORBA::ULong length;\n"
           << streamAndCheck(">> length");
-        // The check here is to prevent very large sequences from being allocated.
+        // Validate a conservative lower bound before allocating the sequence.
+        // Primitive sequences use their exact fixed wire size.
+        std::string min_element_size = "1u";
+        std::string min_element_alignment = "1u";
```
mitza-oci: Are there cases where these have different values?

docs/news.d/5290.rst
```diff
@@ -0,0 +1,16 @@
+.. news-prs: 5290
+
+.. news-start-section: Security
+- Hardened XCDR1 and XCDR2 deserialization against malformed or inconsistent wire data that could otherwise cause excessive allocation, out-of-bounds reads, or undefined behavior:
+
+  - Deserialized Boolean values are normalized, so a Boolean wire byte other than 0 or 1 no longer produces an invalid ``bool`` (:ghissue:`5281`).
+  - Sequence and map element counts are validated against the remaining serialized size, accounting for element size and alignment, before any container is allocated (:ghissue:`5289`).
+  - XCDR2 ``DHEADER`` sizes, and XCDR1 and XCDR2 mutable member sizes, are enforced as hard read boundaries, so a delimited object or member that under-reports its size is rejected instead of over-reading into later data (:ghissue:`5287`).
```
mitza-oci:
```suggestion
  - XCDR2 ``DHEADER`` sizes, and XCDR1 and XCDR2 mutable member sizes, are enforced as read boundaries, so a delimited object or member that under-reports its size is rejected instead of reading into later data (:ghissue:`5287`).
```

docs/news.d/5290.rst
```diff
@@ -0,0 +1,16 @@
...
+  - Malformed XCDR1 extended parameter headers, XCDR2 EMHEADER length codes, mutable-union discriminator IDs, and non-empty parameter-list-end sentinels are rejected.
+
+  Forward compatibility is preserved for unknown bitmask bits and, when member names are ignored, for compatible unions whose members use different IDs.
+
+  *Backwards Compatibility Note*: Readers will now reject some malformed samples and RTPS parameter lists that earlier versions accepted and delivered.
+  Well-formed data produced by OpenDDS or another conformant implementation is unaffected.
```
mitza-oci: I'm not sure this is worthy of a note if there are no known instances of OpenDDS producing invalid output

mitza-oci (changes requested):

PR #5293, opened 2026-09-18
Trim whitespace in comma-separated config list values

### LINKED ISSUE #1470: Spaces in comma separated list of transports creates invalid configuration load
A simple multi transport network configuration  is listed below. When parsing the "transports=" element of the 'config/config1' block, the individual elements of the comma separated list are gathered in the TransportRegistry class implementation. The subsequent calls to process the individual network configurations ( net1 and net2) in this same file results in the code finding 'net1' but does not find 'net2' because it is attempting to include the leading space shown in the "transports=" line (see below).

Tests show that a trailing space after 'net2' does work and does not seem to be taken into consideration.  A trailing space after 'net1' and before the comma also fails to work. So any spaces surrounding the commas appear to be the fragile points in the parsing. Using a ^ to show spaces in these examples shows what works and what doesn't currently.

transports=^net1,net2     #this works
transports=net1^,net2     #this does not work
transports=net1,^net2     #this does not work
transports=net1,net2^     #this works

This behavior seems to introduce a degree of fragility that human readable/edited files might encounter more often, so a decision to trim the offending spaces might ease possible subtle errors.

Sample full configuration file:
```
[common]
DCPSDebugLevel=10
DCPSTransportDebugLevel=5
DCPSInfoRepo=corbaloc::192.168.1.100:5000/DCPSInfoRepo

# local IP-address for Built-In-Topics
DCPSBitTransportIPAddress=192.168.1.20

# Sets the global transport configuration (used by default in the
# process to config1, defined below
DCPSGlobalTransportConfig=config1

# Transport configuration
[config/config1]
transports=net1, net2           <------------- Space before net2 is an issue

# Transport instance named net1, of type "tcp".
[transport/net1]
transport_type=tcp
local_address=192.168.1.20:

# Transport instance named net2, of type "tcp".
[transport/net2]
transport_type=tcp
local_address=192.168.1.21:
```

devinplauche:
## Summary

`ConfigStoreImpl::get(key, StringList)` split comma-separated values but never trimmed the elements, so `transports=net1, net2` in a `[config/...]` section failed with "The inst ( net2) in [config/...] section is undefined." This change trims leading/trailing whitespace around each element using `ACE_OS::ace_isspace`, matching the style already used in that file. The fix is in the central `StringList` getter, so it also covers the other caller (`PARTITION_NAME` in StaticDiscovery).

Fixes #1470.

## Testing

Built ACE/TAO 6.5.24 + OpenDDS from scratch and ran the unit tests:
- `dds_DCPS_ConfigStoreImpl.set_get_StringList` **passes**, including new assertions that `"net1, net2 ,net3"` yields `{"net1","net2","net3"}`.
- The new test was verified to **fail without the source fix** (untrimmed `" net2"`) and pass with it, so it genuinely guards the fix.
- Full `dds_DCPS_ConfigStoreImpl` suite: 27/28 pass. The one failure, `get_NetworkAddressSet`, is **pre-existing on master** (verified on a pristine checkout) and unrelated — that code path goes through `split()` + `parse_optional_port` and never calls the modified getter.
- `tools/scripts/lint.pl` is clean on the changed files; a news fragment was added per the contributor guidelines.

## Checklist

- [x] Target: master
- [x] Bugfix covered by extending the existing `set_get_StringList` test
- [x] News fragment added (`docs/news.d/fix-1470-trim-transport-list.rst`)


sonndinh (approved):

devinplauche:
Note on the CI failure in `test_u22_w1a1_sec`: the failing subtest is `outer_union_appendable_XCDR1` in `tests/DCPS/DynamicTypes`, where the publisher timed out in `WaitSet::wait` waiting for writer/reader matching (see the stuck-stack dump in the "check results" step). This looks like a timing flake rather than something caused by this PR: the change only trims whitespace around elements in `ConfigStoreImpl::get(key, StringList)`, and that test's config uses `transports=rtps` (single name, no whitespace), so the trim is a no-op on its config path. The other 113 jobs are green.

PR #5297, opened 2026-09-22
Check multitopic ownership in DomainParticipantImpl::delete_multitopic

### LINKED ISSUE #2059: domainparticipantimpl::delete_multitopic doesn’t check dp
The domainparticipantimpl::delete_multitopic doesn’t check whether the multitopic passed belongs to this dp, looks there is no test for that

devinplauche:
## Summary

`DomainParticipantImpl::delete_multitopic` looked the multitopic up by name only (`topic_descrs_.find(mt_name)`), so passing a multitopic with a matching name that belongs to a *different* participant deleted this participant's own multitopic. This change checks `get_participant()` on the passed multitopic first — mirroring the ownership check already in `delete_topic_i` — and returns `RETCODE_PRECONDITION_NOT_MET` when it belongs elsewhere. (`delete_publisher` and `delete_subscriber`, also mentioned in the issue thread, already had membership checks.)

Fixes #2059.

## Testing

- Extended `tests/DCPS/MultiTopic/MultiTopicTest.cpp` (`run_test`): two participants each create a multitopic with the same name; `dp->delete_multitopic(other_dp's_mt)` must return `RETCODE_PRECONDITION_NOT_MET`, and each participant's own multitopic must still be deletable afterwards.
- Built `libOpenDDS_Dcps` and the `MultiTopic` test (classic and C++11 mappings) from this branch with no new warnings/errors.
- Verified the new test **fails without the source fix** (`ERROR: delete_multitopic deleted another participant's multitopic`) and gets past the new block with the fix. The full pub/sub portion of that test can't complete in this sandbox (RTPS discovery matching needs multicast, which the container blocks), so CI is the full check for the untouched remainder.
- `tools/scripts/lint.pl` is clean on the changed files; a news fragment was added per the contributor guidelines.

## Checklist

- [x] Target: master
- [x] Bugfix covered by a regression test in the existing `MultiTopic` test
- [x] News fragment added (`docs/news.d/fix-2059-delete-multitopic-ownership.rst`)


dds/DCPS/DomainParticipantImpl.cpp
```diff
@@ -966,6 +966,18 @@ DDS::ReturnCode_t DomainParticipantImpl::delete_multitopic(
   ACE_GUARD_RETURN(ACE_Recursive_Thread_Mutex, guard, topics_protector_,
                    DDS::RETCODE_OUT_OF_RESOURCES);
   DDS::MultiTopic_var mt = DDS::MultiTopic::_duplicate(a_multitopic);
+  DDS::DomainParticipant_var dp = mt->get_participant();
+  DomainParticipantImpl* const dp_servant =
+    dynamic_cast<DomainParticipantImpl*>(dp.in());
```
sonndinh: A small point -- could `dp_servant` be `const DomainParticipantImpl* const` as the domain participant object isn't changed via the pointer?

devinplauche: Good catch — done. `dp_servant` is only used for the ownership comparison, so it's now `const DomainParticipantImpl* const` (commit f0a9f9c).

devinplauche: Good catch — done. `dp_servant` is only used for the ownership comparison, so it's now `const DomainParticipantImpl* const` (commit f0a9f9c).

